PR #2991, opened 2019-04-02
Sign up with email

powerful23:
*Issue #, if available:*
To make it easier to review as the whole pr would be large.
Still in developing.

Currently please focus on the React part. Changes in other categories will be changed later on.
Using `signUpWith` temporarily.
*Description of changes:*


packages/aws-amplify-react-native/src/Auth/SignUp.js
```diff
@@ -50,7 +50,16 @@ export default class SignUp extends AuthPiece {
         this.sortFields = this.sortFields.bind(this);
         this.getDefaultDialCode = this.getDefaultDialCode.bind(this);
         this.checkCustomSignUpFields = this.checkCustomSignUpFields.bind(this);
-        this.defaultSignUpFields = defaultSignUpFields;
+
+        const { signUpConfig={} } = this.props || {};
```
undefobj: Could we move the config into a type that we can start checking for safety? Example:

```
 enum signUpWith {
  EMAIL
  PHONE
}
```

packages/aws-amplify-react/src/Auth/AuthPiece.jsx
```diff
@@ -27,6 +34,77 @@ export default class AuthPiece extends Component {
...
+                            )}
+                        </select>
+                        <Input
+                            placeholder={I18n.get("Please enter your phone number")}
+                            theme={theme}
+                            type="tel"
+                            id="phone_line_number"
+                            key="phone_line_number"
+                            name="phone_line_number"
+                            onChange={this.handleInputChange}
```
undefobj: Have we investigated adding a validation function for forms that customers can control a few things (such as the password complexity) through props, while other things like phone patterns & email validity are standard regex? For example: https://goshakkk.name/instant-form-fields-validation-react/

packages/aws-amplify-react-native/src/Auth/SignUp.js
powerful23: Good idea. I created a `types.ts` in `aws-amplify-react` to add this enum. Also I sent a pr: https://github.com/aws-amplify/amplify-js/pull/3001 to allow adding typescript file into `aws-amplify-react-native`

packages/aws-amplify-react/src/Auth/AuthPiece.jsx
powerful23: Not yet. We do have some simple validations like showing error toasts or disabling the button if one of the input fields is blank. But I think it's definitely a good story we can add to improve the UI experience and provide some props to let the customer to customize the validation themselves.

undefobj: Ok cool lets capture this in Pivotal for a future sprint.

Issue #2983, opened 2019-04-02
Android app crashes by throwing [TypeError: @@for is not a function (evaluating '(typeof Symbol === "function" ? Symbol.for : "@@for")('amplify_default')')]


sgovindan:
![IMG_0203](https://user-images.githubusercontent.com/34024493/55422376-2162ef80-5599-11e9-85f5-0da8ac7c1478.JPG)
On configuring Amplify with latest version as _Amplify.configure(config);_ throws the following error:

I/ReactNativeJS: '[DEBUG] 37:33.234 AWSPinpointProvider - Failed to generate endpointId', { [TypeError: @@for is not a function (evaluating '(typeof Symbol === "function" ? Symbol.for : "@@for")('amplify_default')')]

and this issue happens only in Android phones, with the same package iOS builds are working fine.
Attaching the crash error screenshot.

elorzafe:
@sgovindan thanks for pointing this issue, which version of react-native are you using? I sent a PR to fix this problem, so we can publish a release soon.

malithjkmt:
I'm also facing the same issue. 

Following conbination works fine.
 ```
 "aws-amplify": "^1.0.8",
 "aws-amplify-react-native": "^1.0.2",
 "react": "16.4.1",
 "react-native": "0.56.0"
```
But, when I upgrade aws packages to following, the exact error occurs!
```
"aws-amplify": "^1.1.23",
"aws-amplify-react-native": "^2.1.9",
```
@sgovindan @elorzafe 
More to understand the issue:
![Screenshot_20190403-044440](https://user-images.githubusercontent.com/9573056/55442228-656ee800-55cb-11e9-8ca1-2e5643915508.jpg)

1. Above happens when `react-native run-android`
2. When "Debug JS Remotely" the above error doesn't occur (work around to do development until a fix comes around)
3. When building the **release version**, app builds successfully but when it opens, it crashes with `AndroidRuntime: java.lang.RuntimeException: Unable to load script from assets 'index.android.bundle'. Make sure your bundle is packaged correctly or you're running a packager server.`

Hope AWS will fix the issue soon  :confused:

logivadivelu:
@elorzafe  - Also change the condition at **_Hub.ts_**  **_HubClass.prototype.dispatch_**
as below: 
**_var hasAccess = ampSymbol === AMPLIFY_SYMBOL;_**

PR #2992, opened 2019-04-02
Adding compatibility for Symbol

elorzafe:
*Issue #, if available:*
Fixes: #2983 
*Description of changes:*
Adding checking for Symbol support. I didn't want to export `AMPLIFY_SYMBOL` because is an implementation detail

### DIFF
--- a/packages/analytics/src/Analytics.ts
+++ b/packages/analytics/src/Analytics.ts
@@ -25,8 +25,11 @@ import { PageViewTracker, EventTracker, SessionTracker } from './trackers';
 
 const logger = new Logger('AnalyticsClass');
 
+const AMPLIFY_SYMBOL = ((typeof Symbol !== 'undefined' && typeof Symbol.for === 'function') ?
+    Symbol.for('amplify_default') : '@@amplify_default') as Symbol;
+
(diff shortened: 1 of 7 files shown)


packages/analytics/src/Analytics.ts
```diff
@@ -25,8 +25,11 @@ import { PageViewTracker, EventTracker, SessionTracker } from './trackers';
 
 const logger = new Logger('AnalyticsClass');
 
+const AMPLIFY_SYMBOL = ((typeof Symbol !== 'undefined' && typeof Symbol.for === 'function') ?
```
powerful23: Can we have this `APMLIFY_SYMBOL` as one of the static member of `Hub` module so we don't need to define it multiple times?

elorzafe: I didnt want to exported because is an implementation detail. What to do you think @undefobj

undefobj: I think we should define it in the originating classes since having it exported in Hub would make it easy for customers to discover and use it, then write to one of the protected channels which is a behavior we're trying to discourage. Having it a private member of the calling class helps us keep the symbol out of sight as much as possible.

PR #3001, opened 2019-04-03
feat(aws-amplify-react-native): enable typescript

powerful23:
*Issue #, if available:*
The `aws-amplify-react-native` is not written in Typescript. This pr is the first step to transform it into a Typescript project.
*Description of changes:*

### DIFF
--- a/packages/aws-amplify-react-native/package.json
+++ b/packages/aws-amplify-react-native/package.json
@@ -5,8 +5,8 @@
   "main": "dist/index.js",
   "scripts": {
     "test": "jest --coverage --maxWorkers 2",
-    "build": "npm run clean && babel src --out-dir dist --copy-files",
-    "watch": "npm run clean && babel src --out-dir dist --copy-files --watch",
+    "build": "npm run clean && tsc -p .",
+    "build:watch": "npm run clean && tsc -p . --watch",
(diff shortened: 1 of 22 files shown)


buggy:
> This is a good step, I'd love to see some actual type checking (even if basic) in this PR.

@manueliglesias Having spent almost a year trying to get PR's merged to convert aws-amplify-react to TypeScript I'd strongly recommend against this. @powerful23 seems to be following the approach I took with aws-amplify-js. There's a lot of work and you want some really simple PR's like this one to convert the build process first so you can make sure that it's working correctly before you start introducing other possible bugs. It's the same approach that I took with aws-amplify-js.

Ideally the second step would be to rename all of the `.js` files to `.ts` and use `any` almost everywhere. This gives you time to get the interfaces right.

I combined renaming `.js` to .`ts` and creating interfaces in a single PR with `aws-amplify-js` and it's been painful. The PR is huge and keeping the branch alive (it's been going for almost 6 months) is difficult. You really want to be able to make the changes and get them merged within a few days which smaller PR's help with. I think I've re-written the TypeScript conversion PR for aws-amplify-js 2 or 3 times now because it really gets away from you when you have people actively developing a code base you're converting.

PR #3023, opened 2019-04-05
Use input instead of prompt for password

### LINKED ISSUE #3014: newPasswordRequired prompt versus form on a new page
I am using AWS Cognito with a User Pool configured such that only an administrator can create users. As such, when a new user is created, the administrator assigns a temporary password and the user is configured into the FORCE_CHANGE_PASSWORD status requiring the user to change their password after the first login attempt.

I have successfully implemented this workflow based on the documentation available in AWS repository and github. In particular, I am using the exact representation of the authenticateuser example given in this link:
https://docs.aws.amazon.com/cognito/latest/developerguide/using-amazon-cognito-identity-user-pools-javascript-example-authenticating-admin-created-user.html

However, my currently implementation makes use of the browser prompt as described in Case 12 in the following link to grab the new password from the user:
https://github.com/aws-amplify/amplify-js/tree/master/packages/amazon-cognito-identity-js

The problem with this approach is that the browser prompt input cannot unfortunately be masked and the password is therefore out there for everyone to see. The next obvious choice is to redirect the user to a change password page, but that will reset the context and I will no longer have the parameters sent to the callback function (by that I mean I will have to recreate the arguments userAttributes, requiredAttributes being passed to the newPasswordRequired callback handler), something that is nicely being handled for me.

A solution comes to mind that in such a case, I can redirect to the reset password page while passing the username and password as parameters and repeat the login process all over again, then as the newPassword callback is called, pass the password entered by the user to complete the process.

Is this really the best way to handle this workflow or is there a better method to this madness I wonder?

powerful23:
*Issue #, if available:*
fixes #3014 
*Description of changes:*

### DIFF
--- a/packages/amazon-cognito-identity-js/README.md
+++ b/packages/amazon-cognito-identity-js/README.md
@@ -444,6 +444,18 @@ Note: this method is now deprecated. Please use `setUserMfaPreference` instead.
 
 **Use case 12.** Starting and completing a forgot password flow for an unauthenticated user.
 
+For example: 
+```html
+<body>
+    <label for="#code">Code: </label>
+    <input id="code"></input>
+    </br>
+    <label for="#new_password">New Password: </label>
+    <input id="new_password" type="password"></input>
+    <br/>
+</body>
+```
+
 ```javascript
     cognitoUser.forgotPassword({
         onSuccess: function (data) {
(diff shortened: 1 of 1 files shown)


VinayakVenkataraman:
Can you also suggest a similar fix for Case 23? Updated #3014 also with this question.

Issue #3021, opened 2019-04-05
aws-amplify-react-native - withOAuth HOC - Loading State


jessedoyle:
**Is your feature request related to a problem? Please describe.**
We've implemented a custom OAuth provider login in our react native application with the functionality introduced in https://github.com/aws-amplify/amplify-js/pull/2665 and it's working great!

One issue we've noticed is that it's difficult to render a "loading" state when using the `withOAuth` HOC as there is no ideal state field to query to determine if amplify is loading the user session.

We'd like to be able to render an animation to our users so that it's clear that the application state is loading.

**Describe the solution you'd like**
Ideally, a `loading` prop that defaults to `true` could be passed down to the wrapped component in `withOAuth` [here](https://github.com/aws-amplify/amplify-js/blob/76cde598c30ee8049798e5988261db9165cc775e/packages/aws-amplify-react-native/src/Auth/withOAuth.js#L119-L128).

**Describe alternatives you've considered**
I haven't considered alternatives, but I'd be open to suggestions.

**Additional context**
N/A

~I'd be happy to submit a PR to implement this functionality - it should be pretty straightforward!~ (PR: https://github.com/aws-amplify/amplify-js/pull/3024)

PR #3024, opened 2019-04-05
feat(aws-amplify-react-native): withOAuth Loading

jessedoyle:
*Issue #, if available:*

https://github.com/aws-amplify/amplify-js/issues/3021

*Description of changes:*

* Add a `loading` prop to the `withOAuth` higher order component
  that defaults to `true`.
* Once the amplify session has been created, the `loading` prop
  is set to `false`.
* This allows wrapped components to make decisions and display
  animations to users when the HOC is loading.
* Resolves https://github.com/aws-amplify/amplify-js/issues/3021

### DIFF
--- a/packages/aws-amplify-react-native/src/Auth/withOAuth.js
+++ b/packages/aws-amplify-react-native/src/Auth/withOAuth.js
@@ -40,6 +40,7 @@ export default (Comp) => {
             this.state = {
                 user: null,
                 error: null,
+                loading: false,
             };
 
             listeners.forEach(listener => Hub.remove('auth', listener));
@@ -50,11 +51,13 @@ export default (Comp) => {
 
         componentDidMount() {
             this._isMounted = true;
-            Auth.currentAuthenticatedUser().then(user => {
-                this.setState({ user })
-            }).catch(error => {
-                logger.debug(error);
-                this.setState({ user: null });
+            this.setState({ loading: true }, () => {
+                Auth.currentAuthenticatedUser().then(user => {
+                    this.setState({ user, loading: false })
+                }).catch(error => {
+                    logger.debug(error);
+                    this.setState({ user: null, loading: false });
+                });
             });
         }
         componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


jessedoyle:
It looks like https://github.com/aws-amplify/amplify-js/pull/3005 was just merged and changed OAuth behaviour. I'll see if I can update this PR shortly!

jessedoyle:
Okay - I think I've fixed up this PR. I'm a little uncertain about how to handle the loading state change when initiating logout (or if this is necessary).

jessedoyle:
@manueliglesias - I've rebased master into this PR. Any news on the review? Thanks!

packages/aws-amplify-react-native/src/Auth/withOAuth.js
```diff
@@ -40,6 +40,7 @@ export default (Comp) => {
             this.state = {
                 user: null,
                 error: null,
+                loading: true,
```
jordanranz: Can we set the initial state of loading to false, then move the loading state change into componentDidMount() so it encapsulates state change into where the async function is actually made?

jordanranz:
Looks good to me besides one minor change. Thanks for contributing!

packages/aws-amplify-react-native/src/Auth/withOAuth.js
jessedoyle: Yup - absolutely!

jessedoyle: @jordanranz - I'm just curious how you would prefer the handling of `setState`? 

For now I'm going to implement `setState` with a callback function in `componentDidMount`. But we could use a different pattern with `componentDidUpdate` instead.

jessedoyle:
@jordanranz - I amended my last commit to include your requested change.

jordanranz:
LGTM. Thanks for making that change.

packages/aws-amplify-react-native/src/Auth/withOAuth.js
jordanranz: The callback function looks good! Thanks for making this change. Approving.

Issue #2541, opened 2019-01-15
PushNotification.onRegister() not called at "@aws-amplify/pushnotification": "^1.0.20"


RobertNg1125:
Hi, 

I'd want to use method `PushNotification.onNotificationOpened()` which seems to be available only at 1.0.20 or later. However, after upgrading "@aws-amplify/pushnotification" to 1.0.20, my `PushNotification.onRegister()` no longer got called. I no longer can print out the device token. 

I followed instructions at https://aws-amplify.github.io/docs/js/push-notifications and could confirm that the system was working with "@aws-amplify/pushnotification@1.0.19"

Anyone has successfully implemented `PushNotification.onNotificationOpened()` on react-native application? 

Thanks for your input,

My dependencies: 

```
"dependencies": {
    "@aws-amplify/analytics": "^1.2.10",
    "@aws-amplify/pushnotification": "^1.0.20",
    "aws-amplify": "^1.1.19",
    "react": "^16.7.0",
    "react-native": "^0.57.8"
  },
```

RobertNg1125:
I seems to be a false alarm. 

The `PushNotification.onRegister()` does get called but in different pattern: 
- the iOS app calls `.onRegister()` all the time
- the Android app calls `.onRegister()` once only, when the app/device is completely new. The calls afterwards seems to load from cached? 

Is it an expected behaviour? 

I'm trying to dig deeper into the `@aws-amplify/pushnotification` trying to understand this...

powerful23:
@RobertNg1125 It is an expected behavior for Android because the device token should only be registered to Pinpoint service when:
1. The first time device token received from FCM
2. When the device token is refreshed by FCM

Unfortunately there is no public API to get the device token manually. A work around is to read the token from the AsyncStorage as it's cached by this line: https://github.com/aws-amplify/amplify-js/blob/master/packages/pushnotification/src/PushNotification.ts#L236

RobertNg1125:
@elorzafe @powerful23 Thanks for replying. 

Yes, my workaround has been storing the device token using AsyncStorage. 
It was a bit confusing to me the way it got call multiple times on iOS and once on Android, wish it'll be documented somewhere.

elorzafe:
@RobertNg1125 I have created a new app from scratch (iOS 9.0) `onRegister` is called only the first time the app runs on the device.

There are my dependencies:
```json
{
    "@aws-amplify/pushnotification": "^1.0.22",
[...]

PR #3060, opened 2019-04-10
Update firebase-core and firebase-messaging

dusan-dragon:
*Issue #, if available:*
No issue was created

*Description of changes:*

Updating firebase core and firebase messaging for pushnotifications as it is using quite old version of it and it is kind of conflicting with others projects which also use firebase but with up to date versions of it....


jordanranz:
Related to https://github.com/aws-amplify/amplify-js/issues/2541#issuecomment-479875853

dusan-dragon:
Also note that with new version of gradle google is enforcing strict version checks, so most of the apps using amplify are going to have broken builds.... the not pleasant workaround for this is just disabling all version checks :(

powerful23:
@dusan-dragon about the strict version, is it related to: https://github.com/OneSignal/OneSignal-Gradle-Plugin/issues/37 ?
And isn't this a breaking change because if firebase is upgraded in @aws-amplify/pushnotification, then you also need to upgrade your firebase version in your app accordingly, https://aws-amplify.github.io/docs/js/push-notifications#setup-for-android -> step 7

dusan-dragon:
@powerful23 kind of it is related in the way that if you use gradle plugin 'com.google.gms.google-services' the new version is automatically checking strict version...
We do not use OneSignal at all... Also yes, you need to upgrade the firebase deps but if the app use it already (for example ours) then we want to stick to new up-to date version and not to historic 12.0 version

AndruByrne:
Looking forward to this PR getting reviewed...

elorzafe:
@dusan-dragon the team is addressing this problem in the Push Notifications milestone. We will keep you posted on the repo with the updates.

PR #3069, opened 2019-04-10
fix(@aws-amplify/auth): Valide OAuth state only when generated by Amlify

### LINKED ISSUE #3054: How to avoid error "Invalid state in OAuth flow"?
** Which Category is your question related to? **
Auth

** What AWS Services are you utilizing? **
Cognito User Pools, AWS Amplify

** Provide additional details e.g. code snippets **

Hi AWS Amplify team! Just today we started seeing a new error in fresh npm installs of `@aws-amplify/auth`:

> Invalid state in OAuth flow

Which we narrowed down to this line of code that looks like it was published to npm about a day ago: https://github.com/aws-amplify/amplify-js/blob/76cde598c30ee8049798e5988261db9165cc775e/packages/auth/src/OAuth/OAuth.ts#L216

It seems that our call to `Auth.configure({...our config})` from our cognito hosted ui CallbackURL is now hitting a code path that invokes this `_validateState`, and fails because [`window.sessionStorage['oauth_state']`](https://github.com/aws-amplify/amplify-js/blob/76cde598c30ee8049798e5988261db9165cc775e/packages/auth/src/OAuth/oauthStorage.ts#L15) is coming back as `null`.

Noteworthy is that **our app does not invoke any aws-amplify functions to initiate login with the cognito hosted UI**. Instead, our code:

1. Invokes `Auth.configure({...our config...})`
2. Invokes `Auth.currentAuthenticatedUser()` within a try/catch block
3. In the aforementioned catch block, manually constructs a url in the format `https://{our-cognito-hosted-ui-subdomain}.auth.{aws-region}.amazoncognito.com/oauth2/authorize?response_type=code&idp_identifier={idpId}&redirect_uri={callbackUrl}&clientId={clientId}&scope=&state={ourOwnCustomSamlRelayState}}`

Our user pool only has a single IdP configured, and it is a SAML provider. We do not use any other authn providers. I imagine this is why we are now seeing this error, because we do nothing to set a `window.sessionStorage` value with the key `oauth_state`, yet when our app lands back on our user pool app client's `CallbackURL` and we invoke `Auth.configure({...our config...})`, it seems to now require that the expected URL state matches.

For now we have temporarily worked around this issue by changing our npm dependency for `@aws-amplify/auth` to `1.2.20`. It is version `1.2.21` that is a breaking change for our scenario. 

**What do you recommend we do?**

Is there an aws amplify function we can invoke that will automatically write the expected oauth flow state to `window.sessionStorage['oauth_state']`? We would need one that constructs and redirects to a URL in the above format, **especially with the ability to pass in our own custom SAML relay state. (This is why we have been constructing a manual URL until now, because when the code was initially written, AWS Amplify offered no way in the API surface to pass a custom `state` querystring parameter in the `/oauth2/authorize` endpoint.)**

Or, if there is still no way for us to invoke an AWS Amplify cognito hosted UI login initiation function that accepts a custom `state` argument, should we just write our expected state to `window.sessionStorage['oauth_state']` before we do our custom redirect to the `/oauth2/authorize` endpoint? And if this is on the right track, should we file a bug in this repo?

Or, is there some third option we're not thinking of that would be better?

### LINKED ISSUE #3055: Invalid state in OAuth flow
When I log in through the hosted UI, I am redirected to my domain with a code. All was working fine with version 1.1.23, when I upgraded to 1.1.25 I got an exception "Invalid state in OAuth flow"

The exception is thrown from 
```
OAuth.prototype._validateState = function (urlParams) {
        if (!urlParams) {
            return;
        }
        var savedState = oAuthStorage.getState();
        var returnedState = urlParams.state;
        if (savedState !== returnedState) {
           throw new Error('Invalid state in OAuth flow');
        }
    };
```
When I check savedState and returnedState, savedState is null and returnedState is undefined

 - Device: MacBook Pro
 - OS: 10.13.6
 - Browser Chrome
 - Version 73

[DEBUG] 36:20.15 AuthClass - getting current authenticted user
ConsoleLogger.js:78 [DEBUG] 36:20.77 AuthClass - cannot load federated user from auth storage
ConsoleLogger.js:78 [DEBUG] 36:20.78 AuthClass - get current authenticated userpool user
ConsoleLogger.js:78 [DEBUG] 36:20.78 AuthClass - Failed to get user from user pool
ConsoleLogger.js:88 [DEBUG] 36:20.79 AuthClass - The user is not authenticated by the error

manueliglesias:
*Issue #, if available:*
Fixes #3054
Fixes #3055

*Description of changes:*

### DIFF
--- a/packages/auth/src/OAuth/OAuth.ts
+++ b/packages/auth/src/OAuth/OAuth.ts
@@ -212,6 +212,7 @@ export default class OAuth {
     const savedState = oAuthStorage.getState();
     const { state: returnedState } = urlParams;
 
+    // This is because savedState only exists if the flow was initiated by Amplify
     if (savedState && savedState !== returnedState) {
       throw new Error('Invalid state in OAuth flow');
     }
--- a/packages/auth/src/OAuth/OAuth.ts
+++ b/packages/auth/src/OAuth/OAuth.ts
@@ -127,7 +127,7 @@ export default class OAuth {
       code,
       client_id,
       redirect_uri,
-      code_verifier
+      ...(code_verifier ? { code_verifier } : {})
     };
 
     logger.debug(`Calling token endpoint: ${oAuthTokenEndpoint} with`, oAuthTokenBody);
@@ -212,7 +212,7 @@ export default class OAuth {
     const savedState = oAuthStorage.getState();
     const { state: returnedState } = urlParams;
 
-    if (savedState !== returnedState) {
+    if (savedState && savedState !== returnedState) {
       throw new Error('Invalid state in OAuth flow');
     }
   }
--- a/packages/auth/src/OAuth/OAuth.ts
+++ b/packages/auth/src/OAuth/OAuth.ts
@@ -66,7 +66,7 @@ export default class OAuth {
     domain: string,
     redirectSignIn: string,
     clientId: string,
-    provider: CognitoHostedUIIdentityProvider | string) {
+    provider: CognitoHostedUIIdentityProvider | string = CognitoHostedUIIdentityProvider.Cognito) {
 
     const state = this._generateState(32);
     oAuthStorage.setState(state);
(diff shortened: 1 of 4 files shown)


CaryLandholt:
Nice job, thanks!

PR #598, opened 2018-04-05
Allow custom storage to be passed in via configuration

shark404:
*Issue #, if available:*

*Description of changes:*
Changes to allow custom Auth storage to be passed in via configuration. This is functionality that exists when using the `amazon-cognito-identity-js` library directly, but has not been added to Amplify.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3076, opened 2019-04-12
added isMounted to check if the component has been unmounted before s…

### LINKED ISSUE #3041: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
**Describe the bug**
I am using Amplify in a React-Native project. After signing in with Facebook or Google, it will show this warning. I have removed the listener to avoid memory leaks. And I also have tried to add a set of "_isMounted" states as some of the instructions say, but either of them worked. 

**To Reproduce**
Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.

**Expected behavior**

**Screenshots**
image

**Sample code**
image

inh556:
…etState, related to issue #598

*Issue #, if available:*
issue #598
fixes #3041 
*Description of changes:*
added _isMounted states to check if the component has been unmounted before setState


packages/aws-amplify-react-native/src/Auth/withOAuth.js
```diff
@@ -49,15 +49,19 @@ export default (Comp) => {
...
         componentDidMount() {
+            this._isMounted = true;
             Auth.currentAuthenticatedUser().then(user => {
                 this.setState({ user })
             }).catch(error => {
                 logger.debug(error);
-
-                this.setState({ user: null });
+                if (this.this._isMounted) {
+                    this.setState({ user: null });
```
alebiavati: @inh556 I don't think this line is the problem you are facing. Based on our conversations on #3041, I thought that the `setState` that causes the warning is on line 72:

https://github.com/aws-amplify/amplify-js/blob/fb826d09cfd962af8ccd4660d15da38552756172/packages/aws-amplify-react-native/src/Auth/withOAuth.js#L72

inh556: @robbevan I am so sorry. I just corrected it to the right place.

packages/aws-amplify-react-native/src/Auth/withOAuth.js
```diff
@@ -49,15 +49,17 @@ export default (Comp) => {
...
+                logger.debug(error)；
                 this.setState({ user: null });
             });
         }
-
+        componentWillUnmount() {
+            this._isMounted = false;
+        }
         onHubCapsule(capsule) {
             // The Auth module will emit events when user signs in, signs out, etc
```
powerful23: You can add this check here so if the component is not mounted, then just directly return.

inh556: @powerful23 Hi, I corrected the typo and move the check to the componentWillUnmount method, please check if this makes it work. Thank you.

packages/aws-amplify-react-native/src/Auth/withOAuth.js
```diff
@@ -49,15 +49,18 @@ export default (Comp) => {
...
             });
         }
-
+        componentWillUnmount() {
+            this._isMounted = false;
+            return;
+        }
         onHubCapsule(capsule) {
             // The Auth module will emit events when user signs in, signs out, etc
             const { channel, payload } = capsule;
```
powerful23:
```suggestion
            if (!this._isMounted) return;
            const { channel, payload } = capsule;
```

PR #3102, opened 2019-04-15
fix(aws-amplify-react-native): Use more compatible default export syntax

ProdigySim:
Relevant issue #2686 

Users trying to use `aws-amplify-react-native` in an Expo application have issues getting the Jest runtime to import files from this library. Specifically, two issues block their usage:

1. The default jest configuration used in the expo application will not transpile `aws-amplify-react-native`
2. Even when transpiling, the babel-based toolchain chokes on the `export default IDENTIFIER = EXPR` syntax used in `aws-amplify-react-native`.

The first issue is reasonably solved by configuration on the users' end, but the second is more easily solved by aws-amplify-react-native. This PR fixes the second issue by simply removing the esoteric syntax.


GeorgeBellTMH:
Can we get an update on this, it is causing me problems in my expo project...

PR #3104, opened 2019-04-15
fix(aws-amplify-react-native): catch the error when verifying contact

### LINKED ISSUE #3072: Authenticator React Native offline behavior
**Describe the bug**

I have wrapped my root RN Component in `withAuthenticator`. It all works fine while online, but if I start the app after previously logging in when I don't have a network connection, I just get a white screen. Looking into the source code of `aws-amplify-react-native`, it looks like on mount, it will `checkUser` (and since I logged in previously, there will be a user), and then `checkContact`. It looks like that method is checking online whether the user has verified contact details, but when offline it will just throw a network error. Ultimately, the authState will stay in `loading` and the app just shows a blank screen without being able to get to the login even after the network connection is restored.

(`checkContact` also doesn't have a `catch` so we get an unhandled promise rejection / network error.)

**To Reproduce**

1. Use Authenticator in a React Native app.
2. Login once.
3. Stop the app.
4. Disable network (airplane mode / disable wifi / etc.)
5. Start the app again.

**Expected behavior**

Authenticator shows the app or at least is configurable, so the app can specify what it wants to happen.

 - Device: tested on Nokia 3.1 and Pixel 2 emulator
 - OS: Android

powerful23:
*Issue #, if available:*
fixes #3072 
*Description of changes:*

### DIFF
--- a/packages/aws-amplify-react-native/src/Auth/Authenticator.js
+++ b/packages/aws-amplify-react-native/src/Auth/Authenticator.js
@@ -107,17 +107,20 @@ export default class Authenticator extends React.Component {
         }
     }
 
-    checkContact(user) {
-        Auth.verifiedContact(user)
-            .then(data => {
-                logger.debug('verified user attributes', data);
-                if (!JS.isEmpty(data.verified)) {
-                    this.handleStateChange('signedIn', user);
-                } else {
-                    user = Object.assign(user, data);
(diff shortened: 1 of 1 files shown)


jordanranz:
Is the handled in aws-amplify-react

powerful23:
@jordanranz no it's only in aws-amplify-react-native.

PR #3156, opened 2019-04-23
fix(@aws-amplify/analytics): (pinpoint provider) resolve/reject correctly when calling record

powerful23:
*Issue #, if available:*
`Analytics.record` will resolve itself which will swallow the actual result of sending events.
*Description of changes:*
By passing down the resolve and reject handlers the `record function` could resolve correctly. Also refactored the retry part.


packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -94,50 +95,34 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
             logger.debug('exceed analytics events buffer size');
-            return Promise.reject(false);
+            return handlers.reject(new Error('Exceed the size of analytics events buffer'));
         }
     }
 
-    private async _sendFromBuffer(params) {
-        const { event, config } = params;
+    private async _sendFromBuffer(params, handlers) {
+        // const { event, config } = params;
```
jordanranz: Remove commented code

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -382,31 +355,29 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
+                    this._removeUnusedEndpoints(appId, request.User.UserId)
+                    .then(() => {
+                        logger.debug('Remove the unused endpoints successfully');
+                    }).catch(e => {
+                        logger.warn(`Failed to remove unused endpoints with error: ${e}`);
+                        logger.warn(`Please ensure you have updated your Pinpoint IAM Policy ` +
+                            `with the Action: "mobiletargeting:GetUserEndpoints" ` +
+                            `in order to get endpoints info of the user`);
+                    }).finally(() => {
+                        return handlers.reject(err);
```
jordanranz: Do you mean to return reject here. This will occur even if the promise succeeds

packages/analytics/src/Providers/AWSPinpointProvider.ts
powerful23: Good catch, done.

packages/analytics/src/Providers/AWSPinpointProvider.ts
powerful23: Good catch. I changed the logic so if unused endpoints are removed successfully, will call retry functions. Otherwise reject.

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -288,82 +249,84 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
+                                    [eventId]: {
+                                        StatusCode,
+                                        Message
+                                    }
+                                }
+                            } 
+                        } 
+                    } 
+                } = data;
+                if ( ACCEPTED_CODES.find(c => c === StatusCode)) {
```
jordanranz: Curious if you can use .includes() here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

It improves the readability

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -288,82 +249,84 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
                     this._endpointGenerating = false;
                     logger.debug('record event success. ', data);
-                    res(true);
+                    return handlers.resolve(data);
+                } else {
+                    if (RETRYABLE_CODES.find(c => c === StatusCode)) {
```
jordanranz: Possibly .includes() here as well

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -288,82 +249,84 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
-        // credentials updated
-        const { event, timestamp, config, credentials } = params;
-        this._initClients(config, credentials);
-
-        logger.debug('record session stop');
-        const eventParams = this._generateBatchItemContext(params);
-        return this._pinpointPutEvents(eventParams);
+    private _retry(params, handlers) {
+        const { config : { resendLimit } } = params;
+        params.resendLimit = typeof params.resendLimit === 'number' ?
```
jordanranz: This line and the lines below are a bit hard to understand. Could you add a comment here to explain what is happening here.

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -288,82 +249,84 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
+            params.resendLimit : resendLimit;
+        if (params.resendLimit-- > 0) {
```
jordanranz: What is the decrement doing here?

powerful23: To reduce the resend times. Equals to
```js
if (params.resendLimit > 0) {
    params.resendLimit -= 1;
}
```

packages/analytics/src/Providers/AWSPinpointProvider.ts
powerful23: It's mainly for backward compatibility. Has added a comment.

packages/analytics/src/Providers/AWSPinpointProvider.ts
powerful23: @jordanranz Done.

packages/analytics/src/Providers/AWSPinpointProvider.ts
powerful23: @jordanranz Done.

Issue #3142, opened 2019-04-22
My react app enclosed with withAuthenticator is showing Loading dialog often


Private-SO:
** Which Category is your question related to? **

I had a react-app i enclosed the app with `withAuthenticator` and also added some few components to the app and pushed to cloudfront. Whenever i click on a page after `signin` it is showing `loading dialog` but once i `refresh` or `go back to previous page and come in` its works properly. I want to know why it is rendering `loading dialog` every time whenever i am `signing in` . I believe it is taking time to render the components.

BTW: I didn't find any logs in the console when it is rendering `loading dialog` .

** What AWS Services are you utilizing? **
`amplify-cli - 0.1.36` and its integrated features like `add hosting with cloudfront`.

** Provide additional details e.g. code snippets **

Loading screenshot
![image](https://user-images.githubusercontent.com/26573898/56485242-de899d00-64f0-11e9-9571-48d21426b8f9.png)

jordanranz:
I found the root cause. I will put a PR up as soon as I have fixed this.

YaswanthC:
Is there any temporary workaround for this right now ? Because we are bit out of time..

Thanks

jordanranz:
@YaswanthC We merged the fix, once published to npm it will be available on the unstable release `aws-amplify-react@unstable`

jordanranz:
Fix is published to npm: `aws-amplify-react@2.3.7-unstable.3`

Closing this issue. Please reopen if there are any related questions or concerns.

YaswanthC:
Thanks @jordanranz . Then I need to import  `withAuthenticator` from `aws-amplify-react@unstable` right ??

When it will be published to npm ? Any ETA on that ?

Thanks

Private-SO:
Hi Jordanranz,

`Loading issue` got resolved. Now, I am facing a new issue in this release `aws-amplify-react@2.3.7-unstable.3` . And i am not seeing this issue when i use the 'amplify 2.3.0`.

Reproduce:
1) create-react-app
2)amplify init followed by amplify add auth
3)In that go for custom and enable mfa SMS
4)Create an account
5)Sigin into your account and then enter SMS code

Expected Behaviour
It need to signin me into the app .

Problem
It is routing me again into the login screen .

Screenshot
 
![loadingfixcode](https://user-images.githubusercontent.com/26573898/56900312-23926e00-6ab3-11e9-9ad9-0806edcebd5f.PNG)

[...]

PR #3158, opened 2019-04-24
Use boolean type instead of string for local storage value

jordanranz:
*Issue #, if available:*
#3142 

*Description of changes:*
This local storage value is retrieved [here](https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react/src/Auth/Authenticator.jsx#L64) then evaluated [here](https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react/src/Auth/Authenticator.jsx#L66)

The evaluation will be incorrect because local storage was setting false as a string.

Modify the check to be evaluate string value instead of boolean.


powerful23:
@jordanranz Seems like the localStorage only accepts string value. I think it would be safer to change the way of checking the value instead of changing it to boolean. Could you change this line instead?

Issue #3117, opened 2019-04-17
User may not be fully signed out


matsev:
**Describe the bug**
Not awaiting the promise when calling [Auth.signOut()](https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/Auth.ts#L1252) implies that the user may not be fully signed out when the promise returned by the function resolves.

**To Reproduce**
 - The call to [this._oAuthHandler.signOut(); on line 1226](https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/Auth.ts#L1226) returns a promise which is not resolved before the new promise is resolved on [line 1228](https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/Auth.ts#L1228).

- Similarly, the call to [this._oAuthHandler.signOut(); on line 1240](https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/Auth.ts#L1240) returns a promise which is not resolved before the new promise is resolved on [line 1242](https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/Auth.ts#L1242).

**Expected behavior**
All promises should be resolved before the wrapping promise is resolved.

**Version**
- "@aws-amplify/auth": "^1.2.22

PR #3179, opened 2019-04-27
fix(@aws-amplify/auth): Handle case when `signOut()` promise is rejected

matsev:
This change forwards the returned promise so that the caller gets notified both if  `this._oAuthHandler.signOut()` is resolved as well as rejected.

*Issue #, if available:*
#3117 (see also #3172)

*Description of changes:*

Previously, a rejection of the `this._oAuthHandler.signOut()` method was left unhandled. Consequently, the promise created on [line 1233](https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/Auth.ts#L1233) was neither rejected nor resolved if `this._oAuthHandler.signOut()` failed.


packages/auth/src/Auth.ts
```diff
@@ -1259,7 +1259,7 @@ export default class AuthClass {
                 logger.debug('user sign out', user);
                 user.signOut();
                 if (isSignedInHostedUI) {
-                    return this._oAuthHandler.signOut().then(() => res());
+                    return res(this._oAuthHandler.signOut());
```
powerful23: Should we also add `return` before this line https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/OAuth/OAuth.ts#L244 ?

matsev: First reaction is "yes", since the [signOut()](https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/OAuth/OAuth.ts#L221) function is `async` so it makes sense. However, after I looked into what `this._urlOpener(oAuthLogoutEndpoint)` does it is not obvious what is the correct way forward.

- `_this._urlOpener` can be assigned two different values [OAuth.ts#L58 ](https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/OAuth/OAuth.ts#L58):
 ```this._urlOpener = config.urlOpener || launchUri;```

- In case of `config.urlOpener`, it is either an [AwsCognitoOAuthOpts](https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/types/Auth.ts#L104) or an [Auth0OAuthOpts](https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/types/Auth.ts#L119). Regardless, `urlOpener` is a declared as a function that returns a promise, e.g.
```urlOpener?: (url:string, redirectUrl: string) => Promise<any>```

- In case of `launchUri` it does _not_ return a promise, [urlOpener.ts#L16](https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/OAuth/urlOpener.ts#L16)
```export const launchUri = (url) => window.open(url, SELF);```
The `windows.open()` function returns [A WindowProxy object](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#Return_value) in case of success or `null` if it fails. 

One suggestion is to let the `launchUri` function return a promise, e.g.

```
export const launchUri = (url) => {
    const windowProxy = window.open(url, SELF);
    if (windowProxy) {
        return Promise.resolve(windowProxy);
    } else {
        return Promise.reject();
   });
}
```

However, I do not know if this change has other implications elsewhere.

manueliglesias: @matsev great analysis, thanks!

I like the idea of explicitly returning a promise from `launchUri`.

For context, the `urlOpener` exists to give developers some control on how to initiate the oath flow for their platform (e.g. using *chrome custom tabs* in react native on android, or [Expo's `WebBrowser`](https://docs.expo.io/versions/latest/sdk/webbrowser/))

Could you add that change to this PR? The react native one is already returning a promise:

https://github.com/aws-amplify/amplify-js/blob/7e90705afd21798b5c79069d637621f0c53b70da/packages/auth/src/OAuth/urlOpener.native.ts#L16

matsev: @manueliglesias done, see updates below

matsev:
@powerful23 Any news about this PR? The stale bot marked it as stale, but AFAIK the ball is in your court.

Issue #2460, opened 2019-01-04
Could not find a declaration file for module '@aws-amplify/ui'


samuelcastro:
**Describe the bug**
When trying to build a typescript project I'm getting: **Could not find a declaration file for module '@aws-amplify/ui'**:

```
node_modules/aws-amplify/lib/index.d.ts:8:16 - error TS7016: Could not find a declaration file for module '@aws-amplify/ui'. '/Users/samuelcastro/Projects/Control4/react-sdk/node_modules/@aws-amplify/ui/dist/aws-amplify-ui.js' implicitly has an 'any' type.
  Try `npm install @types/aws-amplify__ui` if it exists or add a new declaration (.d.ts) file containing `declare module '@aws-amplify/ui';`

8 import UI from '@aws-amplify/ui';
```

**To Reproduce**

1. In a typescript (3.1.2) project run: `./node_modules/.bin/tsc`

**Expected behavior**
Successfully build

jordanranz:
Hey @samuelcastro, how are you including amplify into your project?

samuelcastro:
Yeah, an easy workaround would be just include it, I just wanted to see if a better solution is being take, like a define typescript definition of it.

Thanks anyway.

ffxsam:
Just stumbled upon this too. Everything else has a proper type declaration file in `@aws-amplify`. So this seems like just a tiny oversight. Other folders such as `@aws-amplify/cache` have a `lib` folder full of `.d.ts` files. But the UI folder does not.

pl4yradam:
This should not be marked as a feature request, it completely breaks compiling

vuongnguyen-agilityio:
Add `"@aws-amplify/ui": "1.0.18",` to package.json, below `"aws-amplify": "1.1.26",` and good luck :))

willkara:
Any updates on this for Angular 8 support?

rjjlucy:
I added 
`    "noImplicitAny": false,`
in my tsconfig.json and seems like the error is gone.

sammartinez:
As the above callouts and merged PRs, this is resolved. Closing this issue

half2me:
@sammartinez I still have this issue on the latest version. How do I solve it?

kevin4dhd:
same problem

rekomerio:
Problem is here also. I'm trying to just use `withAuthenticator` in React and it crashes everything

RobinBeekhof:
I have this problem to with the latest version

PR #3180, opened 2019-04-28
Add type definitions for @aws-amplify/ui

buggy:
*Issue #, if available:*

#2460 

*Description of changes:*

- Rebuilt the build process using Webpack 4. A lot of unnecessary packages were removed and other packages updated.
- Converted the project to TypeScript.
- Used dts-css-modules-loader to automatically generate the type definitions from the .css files during the build.

Also, removed a lot of unnecessary packages in the process.


buggy:
Bump

powerful23:
@buggy Hi thanks for the pr. We will review it ASAP.

packages/amplify-ui/package.json
```diff
@@ -8,30 +8,21 @@
   "scripts": {
     "test": "webpack",
     "start": "webpack-dev-server",
-    "clean": "rimraf dist",
-    "build": "npm run clean && webpack"
+    "clean": "rimraf dist lib",
+    "build": "npm run clean && webpack && copyfiles -u 1 ./src/*.css.d.ts ./lib"
```
powerful23: Can you give some explanations about why copying those files from src to lib? Also can we auto generate those `.css.d.ts` files during the building process?

I tried to remove those and rebuild but get errors like ` TS2307: Cannot find module './Nav.css'`. Is it related to : https://github.com/s-panferov/awesome-typescript-loader/issues/146 ?

buggy: @powerful23  Good questions and all related.

The `.css.d.ts` files in `src` are automatically generated by dts-css-modules-loader during the webpack build but they are generated after `tsc` runs. This requires some unpacking:

1. dts-css-modules-loader automatically generates the `.css.d.ts` next to the original `.css` files in `src` and I can't change that.
2. If you remove the `.css.d.ts` files from `src` the first build will fail because those files are missing but they will be created again and the next build will pass.
3. I've checked the `.css.d.ts` files into version control because point 2 means that each build TypeScript is actually using the `.css.d.ts` from the previous build. Similar projects like [css-modules-typescript-loader](https://github.com/seek-oss/css-modules-typescript-loader) recommend checking in the `.css.d.ts` files too.
4. The only time point 3 becomes an issue is when you add a new `.css` file because you need to run build twice. The first build will fail due to the missing file but create it and the second build will succeed.
5. I copy the `.css.d.ts` from `src` to `lib` because that's where all of the compiled code and emitted TypeScript definitions files go. TypeScript won't add those automatically so you need to copy them.
6. Points 3, 4 and 5 may make it sound like stale `.css.d.ts` might be getting used but in `lib` (which is what other packages import) you basically have an `index.d.ts` that just imports a bunch of `.css` files and the most recent `.css.d.ts` copied from `src` after webpack finishes generating them.

> I tried to remove those and rebuild but get errors like...

The build for `amplify-ui` was failing because `rimraf` wasn't included as a development dependency. I thought it was in the `package.json` in the `amplify-js` root but I can't find it now. Perhaps I had it installed globally on my machine?? Either way it's now a dev dependency. Can you try removing the `.css.d.ts` files from `src` then running build twice? It should fail but generate them then succeed.

I've also added a banner in the `.css.d.ts` files in `src` saying they are automatically generated.

powerful23: @buggy thanks for the explanation. The point 4 you mentioned seems to be a concern for us as building twice is not a ideal process. Can we try to find a way to generate those `.css.d.ts` files before starting the `webpack` building process?

buggy: @powerful23 I couldn't find a solution. While it's not ideal I do want to stress how rare the double build is. It's once on the machine of the developer who adds it.

jordanranz: I think the benefits we receive from this PR outweigh the friction imposed by the rare double-build

buggy:
@jordanranz @powerful23 I think this needs to be reverted. When I wrote the PR everything tested ok but now it's been merged into master I can't see the styles. I think it's a public holiday in the US on Monday so that gives me an extra day to investigate.

PR #3222, opened 2019-05-04
fix(@aws-amplify/storage): private by default

### LINKED ISSUE #3207: Storage using private by default
**Describe the bug**
It seems that since #2898 was merged, the default storage level is now `private`. I _think_ that the issue is not making the change on a copy, but directly on the configuration coming from `StorageClass` (see "Addition context" for details).

**To Reproduce**
1. Configure Amplify with storage enabled
2. Get or store something with the Storage object.

**Expected behavior**
With no special configuration, public level should be used (unless `Storage.vault` is used).

**Actual behavior**
The private level is used.

 - OS: Linux
 - Browser Chrome
 - Version 1.1.26 (Storage object is gotten through the Amplify-Angular service, version 2.1.12 but I don't think that's really related)

**Additional context**
The relevant code seems to be here (from packages/storage/src/index.ts)
``` TypeScript
        logger.debug('storage configure called');
        const vaultConfig = old_configure.call(_instance, options);

        // set level private for each provider for the vault
        Object.keys(vaultConfig).forEach((providerName) => {
            if (typeof vaultConfig[providerName] !== 'string') {
                vaultConfig[providerName] = { ...vaultConfig[providerName], level: "private" };
            }
        });
        logger.debug('storage vault configure called');
        _instance.vault.configure(vaultConfig);
```

When stepping through , it's pretty clear that `_instance._config` is updated along with the one used for the vault. AFAIK, _instance is the actual Storage object (which would explain why everything is loaded using the private level since I updated to the latest version).

LosD:
*Issue #, if available:*
Fixes #3207

*Description of changes:*
This fixes a problem where the global Storage instance (the one
gotten from the Amplify registry) would be private by default, and adds
a test for it.

### DIFF
--- a/packages/storage/__tests__/Storage-unit-test.ts
+++ b/packages/storage/__tests__/Storage-unit-test.ts
@@ -238,7 +238,23 @@ describe('Storage', () => {
             storage.configure({ Storage: { level: "public"} });
         });
 
-        test('backwards compatible issue, third configure call track', () => {
+        test('normal storage level is public by default', () => {
+          const storage = StorageCategory;
+
+          storage.configure({
+            region: 'region',
+            bucket: 'bucket',
+          });
+
+          expect(storage['_config']).toEqual({
+            AWSS3: {
+              bucket: 'bucket',
+              region: 'region'
+            }
+          });
+        });
+
+      test('backwards compatible issue, third configure call track', () => {
             const storage = new Storage();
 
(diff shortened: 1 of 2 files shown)


LosD:
@elorzafe That's odd, my tests failed first time. Maybe some kind of timing? Oh well, thanks for merging! :)

PR #3223, opened 2019-05-04
fix(Vue-SignUp): check fields are non-empty prior to submitting them

### LINKED ISSUE #3174: Optional attributes cannot be empty during sign-up in Vue
**Describe the bug**
When using the SignUp component in Vue and leaving optional parameters with a blank value, an error is raised.

**To Reproduce**
1. Use `<amplify-sign-up>` component
2. Add in the `signUpConfig` the following:
```
{
  label: this.$Amplify.I18n.get('Middle Name'),
  key: 'middle_name',
  required: false,
  type: 'string',
  displayOrder: 10
}
```
3. Leave the value empty and sign-up
4. See error **Attribute value for middle_name must not be null**

**Expected behavior**
Since the parameter is optional, no error should be raised

**Additional context**
The validation on the component seems to be working properly and no error is raised, Error seems to come from the cognito server.

I believe it would be solved by not adding elements in `user.attributes` that are empty (the fact that they are optional is already taken care of by the `validate` part of the form.

borisdayma:
Cognito does not accept empty attribute values so corresponding attributes cannot be submitted to avoid any error from returned API.

This pull request is fixes #3174.

*Description of changes:*
In Vue <amplify-sign-up> component, we now check that attributes are non-empty prior to submitting them.
No change on the validator has been made so rules for checking correct formatting and required values are not affected.

### DIFF
--- a/packages/aws-amplify-vue/src/components/authenticator/SignUp.vue
+++ b/packages/aws-amplify-vue/src/components/authenticator/SignUp.vue
@@ -210,7 +210,7 @@ export default {
           user.password = e.value
         } else if (e.key === 'phone_number' && e.value) {
           user.attributes.phone_number = `+${this.countryCode}${e.value}`
-        } else {
+        } else if (e.value) {
           const newKey = `${this.needPrefix(e.key) ? 'custom:' : ''}${e.key}`;
           user.attributes[newKey] = e.value;
         };


borisdayma:
I've not been able to test it locally after spending much time on it.
I followed this guide (no so easy to find, maybe should be referenced in CONTRIBUTING): https://github.com/aws-amplify/amplify-js/wiki/Local-Development

 Commands:
```
$ cd amplify-js
$ yarn config set workspaces-experimental true
$ yarn
$ yarn bootstrap
$ yarn build
```

Then I tried either to install or link but would run into following issues:
```
./plugins/amplify.client.js 6:8-21
"export 'AmplifyPlugin' was not found in 'aws-amplify-vue'
./plugins/amplify.client.js 5:0-7
"export 'default' (imported as 'Amplify') was not found in 'aws-amplify'
./plugins/amplify.client.js 9:20-27
"export 'default' (imported as 'Amplify') was not found in 'aws-amplify'
```

The export statement is definitely present in the source, not sure if I'm missing something on the building side.

When I just install the public packages from npm depositories, I don't get these errors.

PR #3229, opened 2019-05-06
Adding cypress data-test attributes, add a spread props for the toast component and match text across components

sammartinez:
*Description of changes:*
Removed Cypress from amplify-js as it will be used in another repo, added data-test attributes in order to utilize [best practices of Cypress](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements), adjusted the Toast component to accept a props spread in order to add an error data test attribute for cypress and ended up adjusting naming of fields within the authentication workflow to be consistent across frameworks.


packages/aws-amplify-react/src/Amplify-UI/Amplify-UI-Components-React.jsx
```diff
@@ -383,7 +383,7 @@ export const Toast = (props) => {
     const theme = props.theme || AmplifyTheme;
 
     return (
-        <div theme={theme} className={AmplifyUI.toast} style={theme.toast}>
+        <div {...props} theme={theme} className={AmplifyUI.toast} style={theme.toast}>
```
elorzafe: Use specific `data-test` instead of passing the props from top level component

packages/aws-amplify-angular/src/components/authenticator/confirm-sign-in-component/confirm-sign-in-component.core.ts
```diff
@@ -19,9 +19,9 @@ import { AuthState } from '../../../providers/auth.state';
 
 const template = `
 <div class="amplify-container" *ngIf="_show">
-  <div class="amplify-form-container">
-    <div class="amplify-form-body">
-    <div class="amplify-form-header">{{ this.amplifyService.i18n().get('Confirm Sign in') }}</div>
+  <div class="amplify-form-container" data-test="confirm-sign-in-section">
```
elorzafe: I would have a single source for all the `data-test` values, and import those from each file. Ideally should have a bidirectional checking. All of the `data-test` values are being tested and you don't have `undefined` on the components.

packages/aws-amplify-react/src/Amplify-UI/Amplify-UI-Components-React.jsx
sammartinez: Sounds good to me, I can see what I can do.

packages/aws-amplify-angular/src/components/authenticator/confirm-sign-in-component/confirm-sign-in-component.core.ts
sammartinez: The only place that these attributes are used will be within the cypress integration tests themselves. Within the tests are the selectors that will be used in order to find the predefined attributes, so I don't believe it would lead to having the attribute be undefined on the component itself because it is already predetermined. I also have a doc I can share with you offline for all of the available attributes we have for it. I just feel it can get slightly messy with having one source file available to us and we would need to import, parse and structure the attributes into our respected frameworks we support. WDYT?

sammartinez: Offline conversation with you and Jordan, I'll see about creating a map and importing it across all of our frameworks.

packages/aws-amplify-react/src/Amplify-UI/Amplify-UI-Components-React.jsx
sammartinez: In talking with Jordan on this, we do want to have a spread props where if the customer wanted to use their own props we wanted to give them the ability to do so, so this is a valid change

sammartinez:
closed in favor of https://github.com/aws-amplify/amplify-js/pull/3252

Issue #3057, opened 2019-04-09
aws-amplify-react-native version 2.1.10 breaks withOAuth Custom Provider Authentication


jessedoyle:
**Describe the bug**

After updating `aws-amplify` from version `1.1.24` to `1.1.25` and `aws-amplify-react-native` from `2.1.9` to `2.1.10`, our React Native application's authentication with a custom provider on a Cognito User Pool no longer works using the `withOAuth` HOC.

The error our app receives is (see screenshot 1):

```
NotAuthorizedException: Token is not from a supported provider of this identity pool.
```

At first glance, it looks like this PR introduced the bug: https://github.com/aws-amplify/amplify-js/pull/3005. I have a [comment](https://github.com/aws-amplify/amplify-js/pull/3005#issuecomment-481460171) on that PR that provides some initial details as well.

The upgrade also introduces a YellowBox warning in our app (see screenshot 2):

```
Possible Unhandled Promise Rejection (id: 0)
TypeError: undefined is not an object (evaluating 'window.location.href')
```

**To Reproduce**
Here is a code example similar to our code using the `withOAuth` HOC (generalized):

```js
/*
 * @flow
 */

import React from 'react';
import {
  Button,
  StyleSheet,
  Text
  View,
} from 'react-native';
import { withOAuth } from 'aws-amplify-react-native';
import { Authenticated } from '../../views';

const styles = StyleSheet.create({
  container: {
    flex: 1,
  },
});

type Props = {
  oAuthError: ?string,
  oAuthUser: ?{},
  customProviderSignIn: Function,
  signOut: Function,
};

const SignIn = (props: Props) => {
  if (props.oAuthUser) {
    return (<Authenticated {...props} />);
  }

  return (
    <View style={styles.container}>
      <Text>Login</Text>
      <Text>{ props.oAuthError }</Text>
      <Button
        onPress={() => props.customProviderSignIn('azure')}
        title="Proceed to Login"
      />
    </View>
  );
};

export default withOAuth(SignIn);
```

**Expected Behavior**
After clicking the "Proceed to Login" button the user is redirected using the device's browser to the custom provider authentication page.

After the custom provider authenticates, the browser redirects to a deep link that opens our application and provides an  OAuth2 `code` parameter and grant.

This process worked as expected with `aws-amplify-react-native == 2.1.9` and `aws-amplify == 1.1.24`.

**Current Behavior**
[...]

PR #3230, opened 2019-05-06
fix(@aws-amplify/auth): react-native - guard for window reference

jessedoyle:
*Issue #, if available:* 

- resolves: https://github.com/aws-amplify/amplify-js/issues/3057

*Description of changes:*

In a non-browser environment, the `window` global object will likely not be available. Currently, the latest versions of `aws-amplify` throw a YellowBox warning when using authentication in React Native:

```
Possible Unhandled Promise Rejection (id: 0)
TypeError: undefined is not an object (evaluating 'window.location.href')
```
 
- Use the `browserOrNode` JS helper function to determine the runtime environment. Guard against the case when not running in a browser by providing a `null` value.
- My text editor caught some trailing whitespace throughout the file - I've committed these changes as well.

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -185,8 +185,8 @@ export default class AuthClass {
 
         // initiailize cognitoauth client if hosted ui options provided
         // to keep backward compatibility:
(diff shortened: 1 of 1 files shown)


davidfarinha:
Hi, is this fix available on npm already?

powerful23:
@davidfarinha , yes, it has been published.

Issue #3232, opened 2019-05-07
Auto Login with OAuth without clicking on a button


marioaae:
** Which Category is your question related to? **
Authentication

** What AWS Services are you utilizing? **
Cognito User Pools

** Provide additional details e.g. code snippets **
I'm developing a react js app in a corporate environment, that will authenticate users with cognito which ultimately is linked to a Azure AD application. I manage to do it with no problem using  a button that looks like this:

```
<Button variant="outline-secondary" onClick={this.props.OAuthSignIn();}>
</Button>
```

The snippet above works fine because I'm importing:

`import { withOAuth } from 'aws-amplify-react';`

I'm opening this question because I would like to avoid a button. My application has no meaning if the user is not authenticated instead of pressing a button I want the user to be signed in automatically when the page loads.

So far I tried to do it in the componentDidMount without success:

```
componentDidMount = () => {
    Auth.currentAuthenticatedUser()
      .then(user => {
        // Do nothing because the user is authenticated already
        }})
      .catch(err => {
        // Authenticate the user because the user is not authenticated yet
        this.props.OAuthSignIn()
        
      });
  }
```

The problem with the above way of doing it is that after the redirections triggered by signing in, and reloading this component, the Auth.currentAutheticatedUser method will catch the "not authenticated" error again even when the SignIn process has been started but probably hasn't finished yet.

I would like to know if there is a way to find out if the SignIn process has been started even when it hasn't finished yet?

Or, is there a smarter way/location to call the OAuthSignIn() without the need of clicking on a button?

Thanks!

powerful23:
@marioaae Hi, as a work around you can set a flag in your local storage which indicating whether the user is logging in or not, for example: 
```js
componentDidMount = () => {
    const isSigningIn = localStorage.getItem('flag');
    Auth.currentAuthenticatedUser()
      .then(user => {
        // Do nothing because the user is authenticated already
        }})
      .catch(err => {
        // Authenticate the user because the user is not authenticated yet
        if (isSigningIn !== 'true') {
          this.props.OAuthSignIn();
          localStorage.setItem('flag', 'true');
         }
      });
  }
```
 And you can reset that flag based on the `oAuthUser` or `oAuthError` callback props.

sammartinez:
@marioaae Does the above work for you?

marioaae:
> 
> 
> @marioaae Hi, as a work around you can set a flag in your local storage which indicating whether the user is logging in or not, for example:
> 
[...]

Issue #3214, opened 2019-05-03
Auth: Delay of Hub events using federatedSignIn 


tinymarsracing:
I am referring to the full react sample code from [this section of the docs](https://aws-amplify.github.io/docs/js/authentication#react-components):

```
// App.js
import React, { Component } from 'react';
import logo from './logo.svg';
import './App.css';
import OAuthButton from './OAuthButton';
import Amplify, {Auth, Hub} from 'aws-amplify';
import awsmobile from './aws-exports'; // your Amplify configuration

// your Cognito Hosted UI configuration
const oauth = {
  domain: 'your_cognito_domain',
  scope: ['phone', 'email', 'profile', 'openid', 'aws.cognito.signin.user.admin'],
  redirectSignIn: 'http://localhost:3000/',
  redirectSignOut: 'http://localhost:3000/',
  responseType: 'code' // or token
};

Amplify.configure(awsmobile);
Auth.configure({ oauth });

class App extends Component {
  constructor(props) {
    super(props);
    this.signOut = this.signOut.bind(this);
    // let the Hub module listen on Auth events
    Hub.listen('auth', (data) => {
        switch (data.payload.event) {
            case 'signIn':
                this.setState({authState: 'signedIn'});
                this.setState({authData: data.payload.data});
                break;
            case 'signIn_failure':
                this.setState({authState: 'signIn'});
                this.setState({authData: null});
                this.setState({authError: data.payload.data});
                break;
            default:
                break;
        }
    });
    this.state = {
      authState: 'loading',
      authData: null,
      authError: null
    }
  }

  componentDidMount() {
    console.log('on component mount');
    // check the current user when the App component is loaded
    Auth.currentAuthenticatedUser().then(user => {
      console.log(user);
      this.setState({authState: 'signedIn'});
    }).catch(e => {
      console.log(e);
      this.setState({authState: 'signIn'});
    });
  }

  signOut() {
    Auth.signOut().then(() => {
      this.setState({authState: 'signIn'});
    }).catch(e => {
      console.log(e);
    });
  }

  render() {
    const { authState } = this.state;
    return (
      <div className="App">
        {authState === 'loading' && (<div>loading...</div>)}
        {authState === 'signIn' && <OAuthButton/>}
        {authState === 'signedIn' && <button onClick={this.signOut}>Sign out</button>}
      </div>
    );
  }
}

export default App;

// OAuthButton.js
import { withOAuth } from 'aws-amplify-react';
[...]

PR #3258, opened 2019-05-11
dispatch an event when parsing the url and change the authentictor

powerful23:
*Issue #, if available:*
Related to #3214 
Related to #3232 

When users login using Hosted UI with code flow and get redirected back, there is a middle state that the code is sent to Cognito service to get the auth tokens back. 

By adding the `parsingRedirectedBackUrl` event in the `_handleAuthResponse` function, we can take action by listen on that event, for example, rendering a loading screen or a spinner and eventually either `signIn` or `signIn_failure` event will be emitted to tell whether the user is logged in or not.

And in the `Authenticator` we will listen on that event and set a flag in the local storage indicating the user is signing in via Hosted UI so that it will not check the current user until receive the auth events.


packages/auth/src/Auth.ts
```diff
@@ -1503,6 +1503,12 @@ export default class AuthClass {
             throw new Error(`OAuth responses require a User Pool defined in config`);
         }
 
+        dispatchAuthEvent(
+            'parsingRedirectedBackUrl', 
+            { url: URL },
+            `The Auth category has been configured successfully`
```
manueliglesias: I think this message is not accurate, we already set that message above, see:

```ts
        dispatchAuthEvent(
            'configured',
            null,
            `The Auth category has been configured successfully`
        );
```

packages/aws-amplify-react/src/Auth/common/constants.js
```diff
@@ -4,6 +4,6 @@ const constants = {
     GOOGLE: 'google',
     FACEBOOK: 'facebook',
     AMAZON: 'amazon',
-    SIGN_IN_WITH_HOSTEDUI_KEY: 'amplify-signin-with-hostedUI'
+    SIGNING_IN_WITH_HOSTEDUI_KEY: 'amplify-signing-in-with-hostedUI'
```
manueliglesias: This change to the name and value of this constant is not backwards compatible, when devs migrate to the new version, their users might get undesired behavior if they logged in with the hosted ui with an old version.

If the constant is not reachable form a top-level export (e.g. from the package `index.js`), we could fix the typo in the name of the constant only, not the value. 

My preference would be to leave the typo as is and live with it in the sake of not breaking users.

packages/auth/src/Auth.ts
powerful23: Good catch. I have changed the message and also I have changed the name of the event because `callbackUrl` is the phase used by Cognito.

packages/aws-amplify-react/src/Auth/common/constants.js
powerful23: The key is not exposed from the top-level export so I only changed the value back.

dementiev:
@powerful23 is it already available? If yes, in which amplify version?

Issue #3233, opened 2019-05-07
Can't find Variable: Symbol


Samykills:
**Describe the bug**
Using the library 1.1.27

**Expected behavior**
No error should occur, since this was working fine untill now.

**Screenshots**
![image](https://user-images.githubusercontent.com/18477982/57312693-afd40f00-710b-11e9-84ff-605fb5be0686.png)

**Smartphone (please complete the following information):**
 - Device: Redmi note 5 pro
 - OS: android

**Sample code**
Adding the sample code here
https://github.com/Samykills/AwsAmplifyExp

awsAmplify.js
`import Amplify, { Analytics } from "aws-amplify";
import Config from "react-native-config";
import { Platform } from "react-native";
import { AppContext } from "react-native-ss-app-core";
class AwsAmplify {
  static initalize() {
    const amplifyConfig = {
      Auth: {
        identityPoolId: Config.AWS_COGNITO_IDENTITY_POOL_ID,
        region: "us-east-1"
      }
    };
    //Initialize Amplify
    Amplify.configure(amplifyConfig);

    const analyticsConfig = {
      AWSPinpoint: {
        // Amazon Pinpoint App Client ID
        appId: Config.AWS_PINPOINT_APP_ID,
        // Amazon service region
        region: "us-east-1",
        mandatorySignIn: false
      }
    };

    Analytics.configure(analyticsConfig);
  }

  /**
   *
   * @param {String} eventName - name of the event or module - "traffic"
   * @param {String} type - Type of event- ("entry"/"exit") or anything
   * @param {String} value - Type of event- can be null or anything
   */
  static recordEvent(eventName, action, title) {
    let attributesJson = {};
    let appContext = AppContext.getAppContext();
    attributesJson["action"] = JSON.stringify(action);
    attributesJson["title"] = title;
    attributesJson["zoneSelected"] = appContext.parentZoneName;
    attributesJson["os"] = Platform.OS;
    attributesJson["version"] = Config.CODE_PUSH_VERSION;
    attributesJson["eventTimeStamp"] = new Date().getTime().toString();
    if (appContext.sessionData) {
      let sessionData = JSON.parse(appContext.sessionData);
      attributesJson["userId"] = sessionData.userId.toString();
      sessionData.tenantId
        ? (attributesJson["tenantId"] = sessionData.tenantId.toString())
        : null;
      sessionData.tenantDto
        ? (attributesJson["tenantname"] = sessionData.tenantDto.name.toString())
[...]

PR #3286, opened 2019-05-15
fix(@aws-amplify/core): guard for Symbol reference

jessedoyle:
*Issue #, if available:*

resolves: https://github.com/aws-amplify/amplify-js/issues/3233

*Description of changes:*

* Some JS interpreters don't natively support `Symbol` yet. An example is the version of JavaScriptCore that ships with React Native 0.58 on Android.
* Add a guard clause that checks for the existence of `Symbol`. If it's not available, substitute the value for a string prefixed with `@@`.
* I used the same pattern used in https://github.com/aws-amplify/amplify-js/pull/2992.

### DIFF
--- a/packages/core/src/constants.ts
+++ b/packages/core/src/constants.ts
@@ -12,7 +12,10 @@
  */
 
 /**
- * This Symbol is used to reference an internal-only PubSub provider that 
+ * This Symbol is used to reference an internal-only PubSub provider that
  * is used for AppSync/GraphQL subscriptions in the API category.
  */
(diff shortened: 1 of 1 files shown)


dinukasal:
in version aws-amplify@1.1.29-unstable.1 this change is not there

jessedoyle:
> in version aws-amplify@1.1.29-unstable.1 this change is not there, it's there only in 1.1.29-unstable.0

I'm noticing the same issue.

@sammartinez - Are there plans to cut a stable release anytime soon?

PR #3301, opened 2019-05-17
fix(@amazon-cognito-identity-js): Adds sendMFASelectionAnswer to types

gcox:
*Issue #, if available:*
None

*Description of changes:*
This method was missing from the index.d.ts file, leading to errors trying to use the library with TypeScript.


packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -110,6 +110,13 @@ declare module "amazon-cognito-identity-js" {
             }): void;
         public verifySoftwareToken(totpCode: string, friendlyDeviceName: string, callbacks: {onSuccess: (session: CognitoUserSession) => void, onFailure: (err: Error) => void}): void;
         public setUserMfaPreference(smsMfaSettings: IMfaSettings | null, softwareTokenMfaSettings: IMfaSettings | null, callback: NodeCallback<Error, string>): void;
+        public sendMFASelectionAnswer(answerChallenge: string,
+                                      callbacks: {
```
powerful23:
```suggestion
                                      callback: {
```

gcox: @powerful23 I believe `callbacks` is actually the correct name in this case. The argument isn't a callback. Instead, it is a hash containing multiple callbacks. The `callbacks` name is consistent with the other functions accepting a hash of callbacks rather than a function.

powerful23: I see. Make sense.

PR #3325, opened 2019-05-21
fix(aws-amplify-angular): fixes #3115 and addresses Chrome web api gesture issue

### LINKED ISSUE #3115: Angular lex issue with Chatbox component aws-lex-audio.js loading blob for 5-7 second even though i am not using it
Amplify angular is loading chatbox component which loading aws-lex-audio.js that is requesting 
blob:http://localhost:8777/215a834f-395d-4d2a-bbb9-b65723611 took 5-7 second in general.

**To Reproduce**
1. install amplify angular
2. import amplify module
3. start website 
4. open website
5. aws-lex-audio.js is loading blob for 5-7 seconds

**Expected behavior**
not loading chatbox component as i am not using it 

**Screenshots**
image

image

 - OS: Windows
 - Chrome
 - Version 73.0.3683.103

haverchuck:
*Issue #, if available:*
#3115 

*Description of changes:*
1. Moves 'worker' variable instantiation to within the recorder function so that blob isn't loaded when component is not in use.
2. Creates the lex audio control when user clicks the mic button, so that Chrome's web sound api is satisfied.  This also required a check in the lex-audio file to make sure the audio_stream is instatiated and do so if it isn't.  This in turn required some async/await usage.


elorzafe:
Nice!! It will be awesome if we can do the same for react and vue?

haverchuck:
@elorzafe It appears that we won't have to fix the unneeded blob loading issue for react and vue, but will need to fix the Chrome audiocontext issue.  Working on it now.

cmbrehm:
so this got auto-closed and not merged?  running into issue #3115

PR #3335, opened 2019-05-22
fix(aws-amplify-react): confirmSignIn bugfix

haverchuck:
*Description of changes:*
Adds e.preventDefault to confirm function in ConfirmSignIn component to prevent page reload


packages/aws-amplify-react/src/Auth/ConfirmSignIn.jsx
```diff
@@ -63,7 +63,8 @@ export default class ConfirmSignIn extends AuthPiece {
             });
     }
 
-    confirm() {
+    confirm(e) {
```
sammartinez: Can we change `e` to `event`, it makes it clearer as we do use `e` for catch blocks in some of our code.

elorzafe: +1

haverchuck: Done!

sammartinez: Thanks!

Issue #3348, opened 2019-05-27
CSS should be prepackaged or not included


shicholas:
**Describe the bug**
This comment on https://github.com/aws-amplify/amplify-js/issues/2523 has this comment: 

```
Any plans to fix this? Im using next.js for my app and this https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react/src/Amplify-UI/Amplify-UI-Components-React.jsx#L5 is causing me to get that error. Would be nice to override to exclude the styles or have them be build as part of the distribution so I didn't have to custom configure webpack for it.
```

**To Reproduce**

1. Use Amplify-React in any repo with webpack. You will need a `css` processor in order for everything to work.

**Expected behavior**

I should not need to require a `css` webpack loader because I expect this repo to only include javascript files.

**Screenshots**
n/a
**Desktop (please complete the following information):**
all

**Smartphone (please complete the following information):**
 all

**Additional context**
n/a

**Sample code**
n/a

jordanranz:
This is being refactored in the UI Component refactor:
https://github.com/aws-amplify/amplify-js/issues/3279

aldegoeij:
@shicholas could you point me to the right direction on how to workaround this using webpack css preprocessor?

shicholas:
```
      {
        test: /\.css$/,
        use: [
          {
            loader: 'style-loader',
          },
          {
            loader: 'css-loader',
          },
        ],
      },
```


I've just been processing css referred to by my app

Issue #3279, opened 2019-05-14
RFC: Amplify UI Component Library Refactor


jordanranz:
This issue is a Request For Comments (RFC).  It is intended to elicit community feedback regarding a proposed change to the Amplify JS UI components.  Please feel free to post comments or questions here.

## Current state of Amplify UI Components

Amplify JS currently provides a library of UI components intended to help developers leverage Amplify categories such as API, Auth, Storage, Interactions and XR.  These UI components are implemented in different modules based on a particular frontend framework.  Currently these frameworks include:

* React: [aws-amplify-react](https://github.com/aws-amplify/amplify-js/tree/master/packages/aws-amplify-react)
* React Native: [aws-amplify-react-native](https://github.com/aws-amplify/amplify-js/tree/master/packages/aws-amplify-vue)
* Angular/Ionic: [aws-amplify-angular](https://github.com/aws-amplify/amplify-js/tree/master/packages/aws-amplify-angular)
* Vue: [aws-amplify-vue](https://github.com/aws-amplify/amplify-js/tree/master/packages/aws-amplify-vue)

The components provided by these modules are implemented in each respective JavaScript framework with duplicated logic. UI components are loosely styled using the Amplify UI style guide with some shared and some duplicated CSS. UI components are customizable but this functionality is limited and lacks parity between frameworks.

## The Problem

This framework-specific approach has led to a variety of maintenance and extensibility challenges, including but not limited to:

* Lack of feature-parity across frameworks
* Duplication of code logic and developer efforts
* Divergent design patterns
* Ramp up for learning and maintaining multiple frameworks
* Inconsistent interaction and styling across frameworks
* Decreased developer velocity in creating new components
* Limited customizability of components

### Overview

Currently, we plan to implement an Amplify UI design system using [Web Components](https://developer.mozilla.org/en-US/docs/Web/Web_Components) generated by [StencilJS](https://stenciljs.com/). We plan to refactor the current framework-specific component logic to alternatively use the logic implemented inside of web components from this design system. 

Additionally, we will utilize CSS-in-JS to create a more flexible customization and theming strategy for components. We are evaluating the use of [Emotion](https://emotion.sh/docs/introduction) to build this strategy into our web components.

This approach has a number of benefits associated with it, including but not limited to:

* Reduced code base
* Faster iteration on enhancements, new features and new components
* Greater accessibility for Amplify users working with frameworks other than React, Angular, Ionic, Vue, or those using no framework at all
[...]

PR #3358, opened 2019-05-28
remove import of CSS files in JSX

shicholas:
The css can be prepackaged and explicitly imported via Webpack or
another way. The explicit inclusion of css styles requires a CSS
processor in dependant aplications that use the amplify-react library.
This PR removes that hard dependency for people that do not wish to use
the CSS processor (e.g. CSS in JS)

*Issue #, if available:* #3348

*Description of changes:*

This removes the inclusion of a CSS file in a `jsx` file, thereby removing the need for dependant applications that use the `amplify-react` package of a CSS processor in their webpack config.

### DIFF
--- a/packages/aws-amplify-react/src/Amplify-UI/Amplify-UI-Components-React.jsx
+++ b/packages/aws-amplify-react/src/Amplify-UI/Amplify-UI-Components-React.jsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 
 import { JS } from '@aws-amplify/core';
 
-import "@aws-amplify/ui/dist/style.css";
 import * as AmplifyUI from '@aws-amplify/ui';
 
 import AmplifyTheme from './Amplify-UI-Theme';


jordanranz:
This will be a breaking change for many users. The React library will be refactored in the current UI Refactor work that is happening: https://github.com/aws-amplify/amplify-js/issues/3279 and this should no longer be a problem after that.

In the meantime, I believe prepackaging the css may be a better solution to removing it.

shicholas:
thanks for the info @jordanranz - if this will be a moot point in the future, I can wait. It's not a blocker for what I'm trying to do.

Olliebaba:
Does this mean that we will be able to manually omit this package's style sheet or import it at a specific point in our code?

My goal is to import this repo's styles first so I can easily override the custom-properties in another CSS file after that. Currently, I cannot figure out when/where your styles are being pulled in so I'm not able to leverage this cascade.

Issue #3368, opened 2019-05-30
Storage: Allow global configuration


aldo-roman:
**Is your feature request related to a problem? Please describe.**
Our S3 bucket requires to upload using Server Side Encryption (SSE) using KMS. This was made supported on Amplify on #2171.

Given this is configured on the provider, current approach requires to include the `serverSideEncryption` on **every request**:
```typescript
return Storage.put(key, file, {
    serverSideEncryption: 'aws:kms',
    contentType: file.type
});
```

This may lead to human errors, forgetting to encrypt a file on S3 if bucket is not strict.

Other AWS S3 configurations present the same problem, such as `cacheControl`. See [AWSS3Provider.ts](https://github.com/aws-amplify/amplify-js/blob/master/packages/storage/src/Providers/AWSS3Provider.ts#L184-L186).

**Describe the solution you'd like**
Prior to #2898, we could pass this configuration flag on Storage configuration and it would be used on all `put` requests. This is because configuration was extended from default config to request-specific config. See [change](https://github.com/aws-amplify/amplify-js/pull/2898/files#diff-c5937922d4eec993fdcdf55bacbe2003L105).

I would like to see Amplify to allow global configuration of attributes, as in:
```typescript
Storage.configure({
    bucket: BUCKET,
    region: REGION,
    // ...
    serverSideEncryption: 'aws:kms',
});
```

**Describe alternatives you've considered**
Continue configuring SSE KMS on every `put` request. This error only happens on runtime, when a user is already trying to upload a file, so it is hard to catch.

**Additional context**
This was a **breaking change** introduced on 1.0.27. Not published on [Changelog](https://github.com/aws-amplify/amplify-js/blob/master/packages/storage/CHANGELOG.md#1027-2019-03-28) AFAIK.

elorzafe:
@aldo-roman this is already merge and published on npm under unstable tag. We will close this issue, but feel free to open a new one or let us know if you have any issues.

PR #3378, opened 2019-05-31
feat: allow SSE to be passed in global configure of storage

sammartinez:
*Issue #, if available:*
addresses #3368
*Description of changes:*
* Fixed minor typos
* Added serverSideEncryption to be passable from Storage.configure


packages/storage/src/Storage.ts
```diff
@@ -100,10 +100,15 @@ export default class StorageClass {
         if (!config) return this._config;
 
         const amplifyConfig = Parser.parseMobilehubConfig(config);
-        const { bucket, region, level, track, customPrefix } = amplifyConfig.Storage as any;
-        // Update DEFAULT_PROVIDER with defined attributes bucket, region, level, track, customPrefix if exists 
-        // on amplifyConfig.Storage, backwards compatible issue 
-        if ((bucket || region || level || track || customPrefix) && !amplifyConfig.Storage[DEFAULT_PROVIDER]) {
+        const { bucket, region, level, track, customPrefix, serverSideEncryption } = amplifyConfig.Storage as any;
```
elorzafe: I think we need to add more properties to solve that issue.

`SSECustomerAlgorithm, SSECustomerKey, SSECustomerKeyMD5, SSEKMSKeyId` and maybe this others as well
`cacheControl, expires`

The problem we cannot just copy all the attributes is because some of them can be another provider.

The same logic should be updated down below. I was thinking on maybe centralize all the attributes that can be configured globally and look for those on `amplifyConfig.Storage` instead of checking the same logic for destructuring and assigning.

@sammartinez what do you think?

sammartinez: Seems right to me. When i cut the PR I was thinking about this as well. Ill see what I can do

aldo-roman: IMHO, the very nature of Javascript allows configurations to be easily extended.
It seems fairly safe to just pass the complete options object down to the provider.
Is there any reason to opt for the current whitelist approach?

sammartinez: @aldo-roman Just as @elorzafe stated above, other providers could potentially have the same. Hence doing this approach

packages/storage/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "scripts": {
     "test": "tslint 'src/**/*.ts' && jest --coverage",
+    "test:watch": "tslint 'src/**/*.ts' && jest --watch",
```
sammartinez: Adding in order for use when were developing

packages/storage/tsconfig.json
```diff
@@ -9,7 +9,7 @@
             "es2015",
             "dom",
             "esnext.asynciterable",
-            "es2017.object"
+            "es2017"
```
sammartinez: Added es2017 so we can have all newer methods be exposed to from es2017

sammartinez: https://www.typescriptlang.org/docs/handbook/compiler-options.html For reference @elorzafe

jordanranz:
Does this need a documentation change for the Storage section? If so, can you change it and link the PR

sammartinez:
> Does this need a documentation change for the Storage section? If so, can you change it and link the PR

@jordanranz I can look to add a docs PR around this.

sammartinez:
> Does this need a documentation change for the Storage section? If so, can you change it and link the PR

@jordanranz  [The PUT section](https://aws-amplify.github.io/docs/js/storage#put) does have the list broken out of what I am adding for this PR. I believe this is ready to go. I can have a last check with @elorzafe on this when he is back next week.

packages/storage/tsconfig.json
elorzafe: be sure that this is not breaking anything

packages/storage/src/Storage.ts
```diff
@@ -100,32 +100,34 @@ export default class StorageClass {
...
+            'SSECustomerKey',
+            'SSECustomerKeyMD5',
+            'SSEKMSKeyId'
+        ];
+
+        const isInStorageArrayKeys = (k: any) => storageArrayKeys.includes(k);
+
+        const isDefinedNonObject = (v: any) => typeof v !== 'object' && v !== undefined;
+
+        if ((Object.values(amplifyConfig.Storage)) && !amplifyConfig.Storage[DEFAULT_PROVIDER]) {
```
elorzafe: I think this is only when at least one of `storageArrayKeys` is on `amplifyConfig.Storage` and `amplifyConfig.Storage[DEFAULT_PROVIDER]` is null

packages/storage/tsconfig.json
sammartinez: i dont see this as a break change as it will expand of what is allowed to be parsed by the typescript compiler. If anything it allows us to use more methods from es2017

packages/storage/src/Storage.ts
```diff
@@ -100,32 +100,34 @@ export default class StorageClass {
...
+            'track',
+            'customPrefix',
+            'serverSideEncryption',
+            'SSECustomerAlgorithm',
+            'SSECustomerKey',
+            'SSECustomerKeyMD5',
+            'SSEKMSKeyId'
+        ];
+
+        const isInStorageArrayKeys = (k: any) => storageArrayKeys.includes(k);
```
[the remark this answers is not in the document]
sammartinez: I believe it was complaining about the string in the complier, I will double check on this.

packages/storage/src/Storage.ts
sammartinez: With our offline conversation on this, I will look to add the `isInStorageArrayKeys` method as a check for this.

packages/storage/tsconfig.json
sammartinez: In talking with @manueliglesias on this, this would lead to adding more polyfills that we may not want to use, I went ahead and made the adjustments needed in my code and reverted this tsconfig change

PR #3412, opened 2019-06-06
Update circleci config to use new cypress wf, adjust data test attrs

sammartinez:
*Description of changes:*
 * Adjust the cypress integration tests in the circle ci config in order to use the new integration workflow.
* Adjusted data test attributes to match the newly added ones from @powerful23


.circleci/config.yml
```diff
@@ -18,84 +17,145 @@ defaults: &defaults
...
+            cd amplify-js-samples-staging
+            yarn cypress:vue
       - store_artifacts:
-          path: /root/app/packages/aws-amplify-react/cypress/videos
+          path: amplify-js-samples-staging/cypress/videos
       - store_artifacts:
-          path: /root/app/packages/aws-amplify-react/cypress/screenshots
+          path: amplify-js-samples-staging/cypress/screenshots
   deploy:
     <<: *defaults
```
jordanranz: I am a bit worried about the deploy step. Maybe we can apply the same change we did to the unit_test workflow

sammartinez: yeah I can make the adjustment for this then

Issue #2095, opened 2018-11-13
SSO implementation


koladilip:
** Which Category is your question related to? **
Authentication
** What AWS Services are you utilizing? **
Cognito User pools
** Provide additional details e.g. code snippets **
I want to implement SSO using this package. We have several websites and we want each of those websites to redirect to our Universal login website which is an Angular website uses aws-amplify for authentication against Cognito user pool, once login is successful we want to pass tokens to the requested website. Also requested website may not implement aws-amplify.
What is the correct way to implement this? I think we can't pass all the token using URL Query parameters may not be possible as the length of URL will exceed its limits right?

D2KX:
I have spent way too much time on this.

There are a few things Cognito is missing: support for the prompt=none parameter for their authorize endpoint and iframe-support for embedding their endpoints. This is how e.g. Auth0 achieves SSO with Auth0's checkSession() method. These have been frequently requested for a long time, but unfortunately they do not want to implement them it seems.

The only way for a a working (silent) SSO implementation is the one you are trying to do. Create your own auth page, get the tokens, and redirect back to a site of yours with the tokens as fragments/hashes in the URL (this is how OAuth Implicit Grant works aswell). Just make sure that no one can abuse this/redirect to a foreign site.

For a fully silent SSO (check whether a user is signed in the first time a user visits one of your domains), you can use something like Zendesk Cross-Storage and let it embed an invisible iframe to your auth site to check whether a user is already logged in/get the tokens without redirecting and basically emulate what other Auth providers like Auth0 are doing for silent SSO.

I wish they would just support this officially though.

koladilip:
Thanks @D2KX for your reply, this is really good information to get started.

koladilip:
We are not using iframe approach instead we just load the login page in the browser and that will redirect back original page once authenticated.

koladilip:
I have created a pull request for passing refresh token to another site and that site should be able to refresh directly: https://github.com/aws-amplify/amplify-js/pull/2439

koladilip:
I found an easy way to get the ID and access tokens using Cognito refresh token using node-fetch and calling Cognito REST API directly and here is the code for it.
`
fetch("https://cognito-idp.us-east-1.amazonaws.com/", {
            headers: {
                "X-Amz-Target": "AWSCognitoIdentityProviderService.InitiateAuth",
                "Content-Type": "application/x-amz-json-1.1",
            },
            mode: 'cors',
            cache: 'no-cache',
[...]

PR #3466, opened 2019-06-16
feat(@aws-amplify/auth): Add create session using refresh token for SSO

koladilip:
*Issue #, if available:*
#2095
*Description of changes:*
We have a requirement of passing refresh token to a different app hosted in a different domain and that app should be able to create CognitoUserSession using the refreshToken.

### DIFF
--- a/packages/auth/__tests__/auth-unit-test.ts
+++ b/packages/auth/__tests__/auth-unit-test.ts
@@ -2595,6 +2595,34 @@ describe('auth unit test', () => {
             spyon.mockClear();
         });
     });
+
+    describe('createSession test', () => {
+        test('happy case', async () => {
+            const spyon = jest.spyOn(CognitoUser.prototype, 'refreshSession')
+                .mockImplementation((refreshToken, callback) => {
+                    return callback(null, 'cognitoSession');
+                });
+            const auth = new Auth(authOptions);
+            expect.assertions(1);
+            expect(await auth.createSession('username', 'refreshToken')).toBe('cognitoSession');
+            spyon.mockClear();
+        });
+
+        test('error case: token expired', async () => {
+            const spyon = jest.spyOn(CognitoUser.prototype, 'refreshSession')
+                .mockImplementation((refreshToken, callback) => {
+                    return callback('token expired');
+                });
+            const auth = new Auth(authOptions);
+            expect.assertions(1);
+            try {
+                await auth.createSession('username', 'refreshToken');
+            } catch (e) {
+                expect(e).toBe('token expired');
+            }
+            spyon.mockClear();
+        });
+    });
 });
 
 
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -1517,6 +1517,28 @@ export default class AuthClass {
     }
 
     /**
+     * Creates the user session using refreshToken.
+     * In case of SSO, we can pass the refresh token to another app to create new session in that app.
+     * @param {String} username 
+     * @param {String} refreshToken 
+     * @return - A promise resolves to session object if success
+     */
+    public createSession(username: string, refreshToken: string): Promise<CognitoUserSession | any> {
+        const cognitoUser = this.createCognitoUser(username);
+        const cognitoRefreshToken = new CognitoRefreshToken({ RefreshToken: refreshToken });
+        return new Promise((res, rej) => {
+            cognitoUser.refreshSession(cognitoRefreshToken, (error, data) => {
+                if (error) {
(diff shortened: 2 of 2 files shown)


koladilip:
> Thank you for the PR. Could you give more specifics on your use case and what you're trying to build? Adding a new API is something that would take larger review, and most likely isn't something we'd merge without a major release. The code you reference seems to just be a helper wrapper of a couple calls. Is there a reason this cannot be done in app code?

Thanks for quick reply @undefobj, there are few classes and methods not exposed in Auth class when using it in the app like CognitoRefreshToken, createCognitoUser (private method) so I have implemented this new API without changing much of the internal classes.

undefobj:
> > Thank you for the PR. Could you give more specifics on your use case and what you're trying to build? Adding a new API is something that would take larger review, and most likely isn't something we'd merge without a major release. The code you reference seems to just be a helper wrapper of a couple calls. Is there a reason this cannot be done in app code?
> 
> Thanks for quick reply @undefobj, there are few classes and methods not exposed in Auth class when using it in the app like **CognitoRefreshToken**, **createCognitoUser** (private method) so I have implemented this new API without changing much of the internal classes.
> 
> We can refresh from the app also but not using amplify-js and it is a very cumbersome process:
> https://gist.github.com/koladilip/7d3ca1e0d060ed42965eb41744d12333#file-aws-cognito-refresh-tokens-from-brower-js

I see. You can still access them you just don't have the typings.

For full transparency, this most likely will not be merged. We have plans to do a major version of Amplify in a few months which is largely driven by a rewrite of auth for reasons such as this. I understand short term you have to do a little more app code, but we also have too much surface area in the Auth API today due to many convenience methods being added without review or design.

undefobj:
> We can refresh from the app also but not using amplify-js and it is a very cumbersome process:
> https://gist.github.com/koladilip/7d3ca1e0d060ed42965eb41744d12333#file-aws-cognito-refresh-tokens-from-brower-js

BTW Amplify automatically refreshes tokens at the appropriate time. Why do you need to do this manually?

undefobj:
cc @haverchuck

koladilip:
> BTW Amplify automatically refreshes tokens at the appropriate time. Why do you need to do this manually?

Yes, it does, in the same app using local storage but we want to pass the refresh token to a second app and it should refresh on its own without calling the first app, the second app only calls the first app only if the refresh token itself expired.

undefobj:
> > BTW Amplify automatically refreshes tokens at the appropriate time. Why do you need to do this manually?
> 
> Yes, it does, in the same app using local storage but we want to pass the refresh token to a second app and it should refresh on its own without calling the first app, the second app only calls the first app only if the refresh token itself expired.

This sounds like a security concern to me. Why would you pass refresh tokens outside of the current application domain? Perhaps you could writeup a more in-depth explanation/feature request so we could evaluate it?

koladilip:
> This sounds like a security concern to me. Why would you pass refresh tokens outside of the current application domain? Perhaps you could writeup a more in-depth explanation/feature request so we could evaluate it?

Right now also it is possible to do refresh token outside the app using https://gist.github.com/koladilip/7d3ca1e0d060ed42965eb41744d12333#file-aws-cognito-refresh-tokens-from-brower-js  and AWS exposed an API to do the same: https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_InitiateAuth.html. 
I don't see any security issues with this, this is how OpenID connect should work.

D2KX:
> This sounds like a security concern to me. Why would you pass refresh tokens outside of the current application domain? Perhaps you could writeup a more in-depth explanation/feature request so we could evaluate it?

We have similar use case. We are building our own login site that we want to use for multiple apps, so users don't have to log in again and again for every app and we don't have to reimplement the login logic again. Basically, we are building our own version of the Cognito Hosted UI because it misses features we need and is English only with no support for other languages.

undefobj:
Could you please open an issue with your use case so we could evaluate this as a feature request? And to be very specific, please write it as your **use case** so we can evaluate the requirements. "I'm trying to build an app that does X, Y, Z with the following scenario/architecture and here is how I would like to do this in my app code". The replies above are all implementations, not use cases. Thank you.

koladilip:
@undefobj https://github.com/aws-amplify/amplify-js/issues/2095 please refer this

01ank1998:
I've understood your point of sharing the refresh token to the application requesting but how can you verify that requesting application is trusted source to send across the refresh token. How will you verify that part. Can you please explain that part . It has become really difficult for me to find a solution

mbsimonovic:
@01ank1998 by whitelisting valid redirect urls

PR #3575, opened 2019-07-02
fix(@aws-amplify/core): reject promise on error

BeniKrauss:
*Description of changes:*
Reject promise on error,
e.g. if browser asks for permission to enable push notifications but user declines this permission.
Right now the error is logged but the promise is never settled. Therefore one has no possibility to react to this error.

### DIFF
--- a/packages/core/src/ServiceWorker/ServiceWorker.ts
+++ b/packages/core/src/ServiceWorker/ServiceWorker.ts
@@ -126,6 +126,7 @@ class ServiceWorkerClass {
                                 resolve(subscription);
                             }).catch((error) => {
                                 this._logger.error(error);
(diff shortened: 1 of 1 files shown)


greimela:
This small fix would help us a lot!
Currently we are not able to react to users declining the notification popup...

powerful23:
@BeniKrauss thanks for sending this pr but unfortunately we can't merge this because it would potentially break other people's app. However, I think there is another work around to catch that error which is by sending it via `Hub` module: https://aws-amplify.github.io/docs/js/hub so that you can listen on this event in your app and react based on that. What do you think?

sammartinez:
Resolving as there has been no feedback from the customer

PR #3593, opened 2019-07-04
fix(aws-amplify-react-native): Allow non USA dial codes on react native

millarm:
countryDialCodes that was used in getDefaultDialCode() of SignUp wasn't imported, so it was impossible to use a non USA dial code.

The following change has been successfully tested with the following configuration to work for UK phones

const signUpConfig = {
  header: 'My Customized Sign Up',
  defaultCountryCode: '44',
};

export default withAuthenticator(App, { signUpConfig });


millarm:
Bump!   Please fix this for the rest of the world outside those with +1 country code

PR #3560, opened 2019-06-29
Allow theme customizations to username field.

ptahdunbar:
*Issue #, if available:*
custom theme doesn't apply input changes to the username field.

*Description of changes:*
```
<Authenticator theme={CustomTheme} />
```

Should allow the username field to be customized.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3561, opened 2019-06-29
Allow theme customizations to username field.

ptahdunbar:
Same issue as #3560

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3599, opened 2019-07-05
Allow theme customizations to username field.

### LINKED ISSUE #3645: Username input on SignIn not applying custom theme (aws-amplify-react-native)
**Describe the bug**
The username input is not using the custom theme.

This is because in the showComponent method on the SignIn component, [renderUsernameField() is called](https://github.com/aws-amplify/amplify-js/blob/aws-amplify%401.1.30/packages/aws-amplify-react-native/src/Auth/SignIn.js#L78) without without passing the theme.

The fix should be just to pass the theme variable when it's called.

I have fixed this in my app by extending the SignIn component and replacing the renderUsernameField method with the following.

```typescript
  renderUsernameField(theme = this.props.theme) {
    // @ts-ignore
    return super.renderUsernameField(theme);
  }
```

**To Reproduce**
Use the following theme on an Authenticator

```typescript
const theme = StyleSheet.create({
  input: {
    borderRadius: 32,
  },
});
```

**Expected behavior**
The borders should be rounded, but only the password input will be rounded.

ptahdunbar:
Same issue as #3560

*Issue #, if available:*
fixes #3645
*Description of changes:*


ptahdunbar:
@haverchuck @wlee221 I merge in the changes from #3561. Both Sign In and Forgot Password were missing the theme.

PR #3608, opened 2019-07-08
query string parsing and customChallenge Cognito response case

haverchuck:
*Description of changes:*
Changes to support Lambda triggers.  Watches for username param in the query string, as well as adds a custom case for 'challengeName' Cognito response.

### DIFF
--- a/packages/aws-amplify-angular/src/components/authenticator/sign-in-component/sign-in.component.core.ts
+++ b/packages/aws-amplify-angular/src/components/authenticator/sign-in-component/sign-in.component.core.ts
@@ -156,6 +156,8 @@ export class SignInComponentCore implements OnInit {
           this.amplifyService.setAuthState({ state: 'confirmSignIn', user });
         } else if (user['challengeName'] === 'NEW_PASSWORD_REQUIRED') {
           this.amplifyService.setAuthState({ state: 'requireNewPassword', user });
+        } else if (user['challengeName'] === 'CUSTOM_CHALLENGE') {
+          this.amplifyService.setAuthState({ state: 'customConfirmSignIn', user });
         } else {
           this.amplifyService.setAuthState({ state: 'signedIn', user });
         }
--- a/packages/aws-amplify-angular/src/components/authenticator/sign-in-component/sign-in.component.core.ts
+++ b/packages/aws-amplify-angular/src/components/authenticator/sign-in-component/sign-in.component.core.ts
@@ -32,7 +32,9 @@ const template = `
         (usernameFieldChanged)="onUsernameFieldChanged($event)"
       ></amplify-auth-username-field-core>
       <div class="amplify-form-row amplify-signin-password">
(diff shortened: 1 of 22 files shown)


packages/aws-amplify-angular/src/components/authenticator/sign-in-component/sign-in.component.core.ts
```diff
@@ -42,11 +44,13 @@ const template = `
...
         />
-        <span class="amplify-form-action" *ngIf="!shouldHide('ForgotPassword')">{{ this.amplifyService.i18n().get('Forgot Password?') }}
-        <a class="amplify-form-link"
+        <span class="amplify-form-action" *ngIf="!shouldHide('ForgotPassword')">
+          {{ this.amplifyService.i18n().get('Forgot Password?') }}
+          <a class="amplify-form-link"
             (click)="onForgotPassword()"
             data-test="${auth.signIn.forgotPasswordLink}"
-          >{{ this.amplifyService.i18n().get('Reset your password') }}</a></span>
+          >{{ this.amplifyService.i18n().get('Reset your password') }}</a>
```
sammartinez: NON BLOCKING: Just want a note of this. To keep consistent for all frameworks the text should read `Reset Password`. This will need to be a change in the future.

packages/aws-amplify-vue/__mocks__/Amplify.mocks.js
```diff
@@ -15,7 +15,10 @@ module.exports = {
     resendSignUp: jest.fn(() => Promise.resolve({})),
     setPreferredMFA: jest.fn(() => Promise.resolve({})),
     setupTOTP: jest.fn(() => Promise.resolve('gibberish')),
-    signIn: jest.fn(() => Promise.resolve({})),
+    signIn: jest.fn(() => Promise.resolve({
```
sammartinez: Does this affect any other sign in unit tests?

packages/aws-amplify-vue/src/components/authenticator/UsernameField.vue
```diff
@@ -67,6 +67,14 @@ export default {
             auth,
         }
     },
+    mounted: function() {
```
sammartinez: Couldnt this be short handed ? So it would just be `mounted() {...}`

haverchuck: I believe it could be, but I don't think it's a major issue to leave this.

packages/aws-amplify-vue/__mocks__/Amplify.mocks.js
haverchuck: No - and that's actually a problem (but not one that needs to be fixed with this PR).  We need a way to have a consistent mock amplify object that can return dynamic values - and I think it should be common and stored outside of the individual frameworks.

PR #3629, opened 2019-07-11
XR data-test attributes

bluezebragames:
Adding data-test= attributes to the Sumerian Scene components in Angular and React, to match the ones in Vue.  There's a lot of lines changed for not much of a feature difference, so I could really use some thorough checking to make sure that I didn't accidentally break anything...  Another thing to consider: should there be files with lists of data-test attribute strings that can just be imported in?  That's probably more maintainable, but I'd like feedback on whether or not it's the right decision.


packages/aws-amplify-react/src/XR/Loading.jsx
```diff
@@ -15,7 +15,7 @@ import * as AmplifyUI from '@aws-amplify/ui';
 
 const Loading = (props) => {
   return (
-    <div className={AmplifyUI.loadingOverlay}>
+    <div className={AmplifyUI.loadingOverlay} data-test="sumerian-scene-loading">
       <div className={AmplifyUI.loadingContainer}>
         <div className={AmplifyUI.loadingLogo}>
```
jordanranz: Can you add a data-test element to the loading logo here.

jordanranz:
LGTM after you add the extra elements I commented about and fix the merge conflicts.

PR #3667, opened 2019-07-16
fixes #3654 and bug in react url search params check

### LINKED ISSUE #3654: `AmplifyService.authStateChange$` emitting state change twice in Angular
**Describe the bug**
Using `amplify-authenticator` UI component, I use `authStateChange$` as described in docs for Angular is emitting state change twice when clicking `Sign In` button.

**To Reproduce**
Subscribe to `authStateChange$`

**Expected behavior**
To emit value only once when user clicks `Sign in button`.

haverchuck:
*Issue #, if available:*
#3654 

*Description of changes:*
Prevents authStateChange from firing twice.  It does this be removing the setAuthState call from the signIn component, which isn't needed because of the decorateSignIn decorator.  Also fixes checks on URLSearchParams in react.

### DIFF
--- a/packages/aws-amplify-angular/src/components/authenticator/sign-in-component/sign-in.component.core.ts
+++ b/packages/aws-amplify-angular/src/components/authenticator/sign-in-component/sign-in.component.core.ts
@@ -169,8 +169,6 @@ export class SignInComponentCore implements OnInit {
           user.challengeParam.trigger === 'true'
         ) {
           this.amplifyService.setAuthState({ state: 'customConfirmSignIn', user });
-        } else {
-          this.amplifyService.setAuthState({ state: 'signedIn', user });
         }
       })
       .catch((err) => {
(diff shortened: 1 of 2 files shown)


packages/aws-amplify-angular/src/components/authenticator/sign-in-component/sign-in.component.core.ts
```diff
@@ -169,8 +169,6 @@ export class SignInComponentCore implements OnInit {
           user.challengeParam.trigger === 'true'
         ) {
           this.amplifyService.setAuthState({ state: 'customConfirmSignIn', user });
-        } else {
-          this.amplifyService.setAuthState({ state: 'signedIn', user });
```
sammartinez: Im just curious if we will want an `else` block or if we want to change the `else if`s to just `if` blocks WDTY?

haverchuck: I think what we really need to do is get rid of duplicate code between the signIn component and the decorateSignIn decorator.  For this PR, though, I was just trying to make a minimal change to eleminate the initial bug.

PR #3671, opened 2019-07-16
Tests and docs for amplify-text-field

bluezebragames:
This PR adds Stencil and Storybook integration (with Stencil end-to-end tests) to the `amplify-text-field` component.

Storybook component

Unfortunately all the `yarn version`-ing created a bunch of useless commits.


packages/amplify-ui-components/src/components/amplify-button/amplify-button.stories.tsx
```diff
@@ -1,9 +1,8 @@
 import { storiesOf } from '@storybook/html';
```
bluezebragames: Changes due to ESLint

packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
```diff
@@ -14,7 +14,7 @@ export class AmplifyButton {
```
bluezebragames: Changes due to ESLint

packages/amplify-ui-components/src/components/amplify-examples/example5.tsx
```diff
@@ -1,8 +1,6 @@
 import { h } from '@stencil/core';
```
bluezebragames: Changes due to ESLint

packages/amplify-ui-components/src/components/amplify-examples/scene-example.tsx
```diff
@@ -1,8 +1,6 @@
 import { h } from '@stencil/core';
```
bluezebragames: Changes due to ESLint

packages/amplify-ui-components/src/components/amplify-scene-loading/amplify-scene-loading.style.ts
```diff
@@ -1,10 +1,10 @@
 import { css } from 'emotion';
```
bluezebragames: Changes due to ESLint

packages/amplify-ui-components/src/components/amplify-scene-loading/amplify-scene-loading.tsx
```diff
@@ -1,5 +1,13 @@
 import { Component, Element, Prop, h } from '@stencil/core';
```
bluezebragames: Changes due to ESLint

packages/amplify-ui-components/src/components/amplify-scene/amplify-scene.style.ts
```diff
@@ -24,4 +24,4 @@ export const sceneActions = css`
   margin-left: auto;
```
bluezebragames: Changes due to ESLint

packages/amplify-ui-components/src/components/amplify-scene/amplify-scene.tsx
```diff
@@ -32,7 +32,7 @@
```
bluezebragames: Changes due to ESLint

packages/amplify-ui-components/src/components/amplify-text-field/amplify-text-field.e2e.ts
```diff
@@ -0,0 +1,68 @@
...
+    const page = await newE2EPage();
+
+    await page.setContent(`<amplify-text-field></amplify-text-field>`);
+
+    const func = jest.fn();
+    await page.exposeFunction('exposedfunc', func);
+    await page.$eval('amplify-text-field', (element: any) => {
+      element.inputProps.onInput = this.exposedfunc;
+      element.label = 'adding a label so that the component rerenders';
+    });
```
bluezebragames: So the above block of code is kinda counterintuitive.  I snagged it from https://github.com/ionic-team/stencil/issues/1174, and it does work.  However in the [Stencil docs](https://stenciljs.com/docs/end-to-end-testing#set-a-prop-on-a-component-using-an-external-reference) it suggests the following:

```ts
await page.$eval(
  'amplify-text-field',
  (element: any, callback: Function) => {
    element.inputProps.onInput = callback;
    callback();
    element.label = 'adding a label so that the component rerenders';
  },
  func,
);
```

which in fact _doesn't_ work.  So I'm going with this for now, and maybe we should consider submitting some sort of change to the docs?

bluezebragames: I'll take this opportunity to also discuss why this is necessary in the first place.  One would expect that we can just pass the function into the JSX which is being rendered via `page.setContent()`, on line 36 there.  But since this is in a string, we can't pass objects.  This might be obvious here, but [down below](https://github.com/aws-amplify/amplify-js/pull/3671/files#diff-3a16ffd363461e2f35156b2fece0cb8f) the reasoning is less clear (so I'll go into more detail there too).

packages/amplify-ui-components/src/components/amplify-text-field/amplify-text-field.stories.tsx
```diff
@@ -0,0 +1,8 @@
+import { storiesOf } from '@storybook/html';
+import { text as textKnob } from '@storybook/addon-knobs';
+
+storiesOf('Text field', module).add('with text', () => {
+  const label = textKnob('Label', 'Label placeholder');
+  const description = textKnob('Description', 'Description placeholder');
+  return `<amplify-text-field fieldId="id" label="${label}" description="${description}"/>`;
```
bluezebragames: I really would've loved to just be able to have an object knob here to set the `inputProps` on the `amplify-text-field`, but unfortunately I could not get it to work.  The problem here again is that we are returning the component as a string instead of raw JSX.  It might seem that all that is required is to instead return the component itself, as is done in React `render()` functions, but that runs into trouble.  [These Storybook docs](https://storybook.js.org/docs/configurations/typescript-config/) describe how to set up webpack with Storybook to be able to compile .tsx files, but eventually it seems to want the `h()` function from @stencil/core, which makes everything compile but then Storybook is unable to find the component because it has been compiled to a different name or something.

packages/amplify-ui-components/src/components/amplify-text-field/amplify-text-field.tsx
```diff
@@ -27,14 +27,11 @@ export class AmplifyTextField {
...
         <div>
-          {/* <input
-            
-          /> */}
           <amplify-text-input
             id={this.fieldId}
             aria-describedby={this.fieldId && this.description ? `${this.fieldId}-description` : null}
-            type="text"
-            {...this.inputProps}
+            type={this.inputProps.hasOwnProperty('type') ? this.inputProps.type : 'text'}
```
bluezebragames: I'll be the first to admit that this is not very elegant.  Here's the situation, and four solutions to it.

First of all, we want to either use `type=inputProps.type` if type exists on the prop, or use `type="text"` if not.  The issue is that it's not standard to have two attributes with the same name on a single component.

Solution 1: disregard the fact that it's not standard -- every browser just disregards the second instance of a prop.  The solution, then, is to just use
```ts
{...this.inputProps}
type="text"
```
instead of the other way around.  It's brittle, though -- if a browser ever decides to parse in a different way, functionality breaks again.

Solution 2: the one in this commit.  The `type=` attribute still appears twice in the component, it's just with the same value each time, so it's slightly more resistant to change.  Of course, if a browser decides to just refuse to render the page or something in this scenario, this solution will not work.

Solution 3: a little bit more work.  Just delete the `type: ` key-value pair from `inputProps` (or alternatively just list out all of the props in `inputProps`, which is just `onInput=this.inputProps.onInput` for now, but could change.  I'm not a fan of this sub-option because it makes extending the component more difficult -- you have to make the change in two places instead of one.)  It's a little code-smell-y because we'd presumably need a temporary variable to store the `inputProps` object without the `type: ` key.

Another quick thing: if we go with any of these first three, we should probably change that ternary to `{this.inputProps.type || 'text'}`, that's much cleaner.

Solution 4: I just came up with this solution while writing this comment, and I _think_ it's the best one, but I'd like some other opinions.  Above the return, just test if `type` exists on `inputProps`, and if not, set it to `'text'`.  It's simple for sure, and you get to keep the nice clean `{...this.inputProps}`.

packages/amplify-ui-components/src/components/amplify-text-field/amplify-text-field.tsx
```diff
@@ -27,14 +27,11 @@ export class AmplifyTextField {
...
+            inputProps={{ ...this.inputProps }}
```
bluezebragames: This is another weird one.  Originally, the code was `{...this.inputProps}`, which isn't correct because the  `<amplify-text-input>` component is also looking to receive `inputProps={type?: str, onInput?: Function}`.  I figured there could conceivably be some reason why just saying `inputProps={this.inputProps}` would mess up the component, but does anyone know if it actually will?

packages/amplify-ui-components/src/components/amplify-text-input/amplify-text-input.tsx
```diff
@@ -18,7 +18,7 @@ export class AmplifyTextInput {
       <input
         id={this.fieldId}
         aria-describedby={this.fieldId && this.description ? `${this.fieldId}-description` : null}
-        type="text"
+        type={this.inputProps.hasOwnProperty('type') ? this.inputProps.type : 'text'}
```
bluezebragames: See above.  I'm definitely leaning towards option 4.

packages/amplify-ui-components/stencil.config.ts
```diff
@@ -3,15 +3,13 @@ import nodePolyfills from 'rollup-plugin-node-polyfills';
 
 export const config: Config = {
   namespace: 'amplify-ui-components',
-  plugins: [
```
bluezebragames: Changes due to ESLint

packages/amplify-ui-components/src/components/amplify-text-field/amplify-text-field.tsx
```diff
@@ -14,6 +14,8 @@ export class AmplifyTextField {
   } = {};
 
   render() {
+    this.inputProps.type = this.inputProps.type || 'text';
```
bluezebragames: See [this comment](https://github.com/aws-amplify/amplify-js/pull/3671#discussion_r304138732) for explanation.  Another possibility to implement this exact behavior: [Stencil default props](https://stenciljs.com/docs/properties#prop-default-values-and-validation), but for some reason that seems to break specifically on properties within an object, which is exactly what we have.

More specifically, this code
```ts
@Prop() fieldId: string = "this works";
@Prop() inputProps: {
  type?: string;
  onInput?: (Event) => void;
} = {};
```
works, but this code
```ts
@Prop() fieldId: string;
@Prop() inputProps: {
  type?: string = "this doesn't work";
  onInput?: (Event) => void;
} = {};
```
is not valid TS, failing with the error `A type literal property cannot have an initializer.`

bluezebragames:
One more comment, I have no clue why the CircleCI build is failing...  Is this the problem people were talking about earlier, or what?  [People online](https://github.com/facebook/jest/issues/5239#issuecomment-355867359) were saying that running jest with `-w=2` fixed the problem, but I have no clue how or why

bluezebragames:
Yet another comment: we should decide what tests to put in --spec and what tests to put in --e2e, I kinda just lumped them all into e2e when some of them should probably be --spec.

packages/amazon-cognito-identity-js/package.json
```diff
@@ -81,6 +81,7 @@
     "eslint-plugin-node": "^5.2.0",
     "eslint-plugin-promise": "^3.6.0",
     "eslint-plugin-standard": "^3.0.1",
+    "is-my-json-valid": "^2.20.0",
```
jordanranz: what is this dependency for?

packages/amplify-ui-components/package.json
```diff
@@ -25,27 +25,33 @@
     "build:watch": "npm run clean && npm run stencil:watch",
     "clean": "rimraf dist .stencil"
   },
-  "dependencies": {
-    "emotion": "^10.0.9"
```
jordanranz: oo nice, does emotion not have to be an actual dependency?

packages/amplify-ui-components/src/components/amplify-scene-loading/amplify-scene-loading.tsx
```diff
@@ -17,14 +25,16 @@ export class AmplifySceneLoading {
...
-            : <div class={loadingBar}>
-                <div class={loadingBarFill} style={{ width: this.loadPercentage + '%'}} />
-              </div> // TODO: Make loading-bar component
+          {this.sceneError ? (
+            <div class={sceneErrorText}>{this.sceneError}</div>
+          ) : (
+            <div class={loadingBar}>
+              <div class={loadingBarFill} style={{ width: this.loadPercentage + '%' }} />
+            </div>
+          ) // TODO: Make loading-bar component
```
jordanranz: Agree on the loading bar component.

packages/amplify-ui-components/src/components/amplify-text-field/amplify-text-field.e2e.ts
jordanranz: Can you transfer these comments to a Quip document so we can discuss on a more permanent forum

packages/amplify-ui-components/src/components/amplify-text-field/amplify-text-field.stories.tsx
jordanranz: As mentioned above let's move these comments to a Quip doc

packages/amplify-ui-components/src/components/amplify-text-field/amplify-text-field.tsx
jordanranz: last comment for moving this to a Quip doc

packages/amazon-cognito-identity-js/package.json
bluezebragames: ah I think that was for attempting to let Storybook return raw JSX, but since I abandoned that route I should probably remove the dependency

packages/amplify-ui-components/package.json
bluezebragames: hmm that's odd... it's down in dev dependencies, but I don't know how to test if it works without it or not

bluezebragames: it's *conceivable* that it isn't actually necessary for it to be a real dependency, since the emotion might just compile away, but I think it's pretty likely that this just got removed accidentally, thank you for the catch!

packages/amplify-ui-components/src/components/amplify-scene-loading/amplify-scene-loading.tsx
bluezebragames: good point, although that's very much not my comment lol

package.json
```diff
@@ -35,14 +35,17 @@
   },
   "homepage": "https://aws-amplify.github.io/",
   "devDependencies": {
-    "@types/jest": "^20.0.7",
+    "@types/jest": "24.0.15",
```
sammartinez: What is the difference of removing the `^` and just hardcoding to a specific types of jest? Wouldnt we get the latest version if there was no lock file?

packages/amazon-cognito-identity-js/package.json
sammartinez: can we remove it if were not using it :)

packages/amplify-ui-components/package.json
sammartinez: This be great if this is the case!

packages/amplify-ui-components/package.json
```diff
@@ -25,27 +25,33 @@
...
+    "@typescript-eslint/eslint-plugin": "^1.12.0",
+    "@typescript-eslint/parser": "^1.12.0",
+    "babel-loader": "^8.0.6",
+    "emotion": "^10.0.14",
     "eslint": "^5.16.0",
     "eslint-config-prettier": "^4.3.0",
     "eslint-plugin-prettier": "^3.1.0",
     "husky": "^2.4.1",
-    "jest": "23.6.0",
+    "is-my-json-valid": "^2.20.0",
```
sammartinez: Same callout as above, if its no longer needed we should remove this.

packages/amplify-ui-components/src/components/amplify-text-field/amplify-text-field.e2e.ts
```diff
@@ -0,0 +1,68 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('amplify-text-field', () => {
+  it('renders', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(
+      `<amplify-text-field field-id="id" label="Label test" description="Description test"></amplify-text-field>`,
+    );
+    const element = await page.find('amplify-text-field');
```
sammartinez: Can we rename this to `textElement`? This will keep it consistent with what variables are defined below e.g. `labelElement, descriptionElement, etc.`

packages/amplify-ui-components/src/components/amplify-text-field/amplify-text-field.e2e.ts
```diff
@@ -0,0 +1,68 @@
...
+    expect(element).toHaveProperty('type');
```
sammartinez: What is the purpose of this check? Can we be a little more specific of what type we expect this element to have?

packages/amplify-ui-components/src/components/amplify-text-field/amplify-text-field.tsx
sammartinez: Have you tried it ? :)

jordanranz:
> One more comment, I have no clue why the CircleCI build is failing... Is this the problem people were talking about earlier, or what? [People online](https://github.com/facebook/jest/issues/5239#issuecomment-355867359) were saying that running jest with `-w=2` fixed the problem, but I have no clue how or why

I believe there was a change to Stencil that reserves the word 'change'. I'll change this reserved word so the build succeeds.

PR #3698, opened 2019-07-20
fix(@aws-amplify/auth): Improve OAuth flow in react native

### LINKED ISSUE #3681: Hub Auth event, doesnt trigger after signIn with facebook/google for the second time
**Describe the bug**
I'm having some issues with withOauth HOC, I'm using facebook and google authentication via provider directly passing through hostedUI. The first time i can complete and login without any problems, but if i logout and then i log in again, the hub doesnt trigger, and the user is not authenticated. Am i missing something?

 - Device: Iphone 8
 - OS: IOS 12.2

### LINKED ISSUE #3707: Signing in after signing out does not work
**Describe the bug**
Using withOAuth here with react-native and Expo. Signing in then signing out then signing in again does not work. Last sign is the only part not working. It seems this has been fixed for withAuthenticator but not withOAuth.

```
  "dependencies": {
    "amazon-cognito-identity-js": "^3.0.13",
    "aws-amplify": "^1.1.31-unstable.5",
    "aws-amplify-react-native": "^2.1.14",
    "expo": "^33.0.7",
    "expo-facebook": "^5.0.1",
    "expo-image-picker": "^5.0.2",
    "react": "16.8.3",
    "react-dom": "^16.8.6",
    "react-native": "https://github.com/expo/react-native/archive/sdk-33.0.0.tar.gz",
    "react-native-fbsdk": "^0.10.1",
    "react-native-web": "^0.11.4"
  },
```

**To Reproduce**
1. Sign in using facebookSignIn
2. Sign out using signOut
3. Sign in again using facebookSignIn
4. Check the logs and see that :
```
[DEBUG] 11:00.987 Hub - Dispatching to auth with  Object {
  "data": Object {
    "url": "exp://10.0.2.59:19000?code=65def4c0-f901-4557-bb4d-73d05089bf81&state=c9WuZeu2SXWqs2nrGGO584nTTwtwZ6GE",
  },
  "event": "parsingCallbackUrl",
  "message": "The callback url is being parsed",
}
``` 
doesnt appear to be exectued at the second sign in.

**Expected behavior**
Signing in after signing out should be expected

 - iPhone 7

manueliglesias:
Fixes #3681
Fixes #3707

Tested with

- https://github.com/manueliglesias/hosted-ui-react-native-cli
  - RN cli app - no Expo
  - "react-native": "0.58.4"
  - both iOS and Android
- https://github.com/manueliglesias/hosted-ui-react-native
  - Expo 31
  - both iOS and Android

### DIFF
--- a/packages/auth/src/urlListener.native.ts
+++ b/packages/auth/src/urlListener.native.ts
@@ -24,31 +24,23 @@ export default async callback => {
 
     let Linking: any;
     let AppState: any;
-    let Platform: any;
 
     try {
-        ({ Linking, AppState, Platform } = require('react-native'));
+        ({ Linking, AppState } = require('react-native'));
     } catch (error) { /* Keep webpack happy */ }
 
 
     handler = handler || (({ url, ...rest }: { url: string }) => {
-        logger.debug('addEventListener', { url, ...rest });
+        logger.debug('urlListener', { url, ...rest });
         callback({ url });
-
-        AppState.removeEventListener('change', handler);
-        Linking.removeEventListener('url', handler);
     });
 
     Linking.removeEventListener('url', handler);
     Linking.addEventListener('url', handler);
-    if (Platform.OS === 'android') {
-        AppState.removeEventListener('change', handler);
-        AppState.addEventListener('change', async newAppState => {
-            if (newAppState === 'active') {
-                const initialUrl = await Linking.getInitialURL();
-                handler({ url: initialUrl });
-            }
-        });
-    }
-
+    AppState.addEventListener('change', async newAppState => {
+        if (newAppState === 'active') {
+            const initialUrl = await Linking.getInitialURL();
+            handler({ url: initialUrl });
+        }
+    });
 };


packages/auth/src/urlListener.native.ts
```diff
@@ -24,31 +24,23 @@ export default async callback => {
...
-        AppState.addEventListener('change', async newAppState => {
-            if (newAppState === 'active') {
-                const initialUrl = await Linking.getInitialURL();
-                handler({ url: initialUrl });
-            }
-        });
-    }
-
+    AppState.addEventListener('change', async newAppState => {
+        if (newAppState === 'active') {
```
undefobj: The [React Native docs](https://facebook.github.io/react-native/docs/appstate) show an extra check 

```
if (
      this.state.appState.match(/inactive|background/) &&
      nextAppState === 'active'
    )
```

Should we add this in as well for safety?

manueliglesias: I think they show that check because in their example they only care for the scenario where the app was in the background and then brought to the foreground, they don't care about the first launch.

> However, currentState will be null at launch while AppState retrieves it over the bridge.

We don't want that check to, e.g. be able to handle *[Universal Links](https://developer.apple.com/ios/universal-links/) / [Android App Links](https://developer.android.com/training/app-links)*

Issue #1028, opened 2018-06-14
Typescript support for aws-amplify-react


buggy:
Do you want to request a feature or report a bug?

Feature

What is the current behavior?

The aws-amplify-react package is written in Javascript instead of TypeScript which is used by aws-amplify and aws-amplify-angular. This prevents the TypeScript definitions from being automatically generated (see #305).

What I'm proposing is to convert the code base from Javascript to TypeScript so the definitions can be generated automatically. I think this process can be completed comfortably with 4 PR's:

1. Add webpack. This changes the build process while continuing to use Babel as the compiler.
2. Replace Babel with TypeScript. TypeScript will be configure to allow JS so there are no code changes.
3. Rename .js(x) files to .ts(x), ts-ignore any "errors", disable allowJs and enable type definition generation.
4. Remove the ts-ignore's and fix the underlying "errors".

There should probably be a gap between PR 3 and 4 to minimize the impact on any PR's that other people are working on.

buggy:
@powerful23 I think you're the code owner for aws-amplify-react. I'm happy to work on the PR's but I realize you might have other plans so I want to check first.

powerful23:
@buggy thanks but can we just add a `index.d.ts` file under that package instead of changing those js files like what `amazon-cognito-identity-js` does?

buggy:
@powerful23 It's possible to write and maintain definitions independent of the code base but in my experience the definition files fall behind pretty quickly and never catch up. While converting the code base sounds daunting most of the work is in changing the build tools because TypeScript is a superset of Javascript so once that tools have been changed the rest is virtually just renaming file extensions. You then get the benefits of static types as well as the ability to automatically generate the .d.ts files

buggy:
@ossareh Some interfaces and types will get fixed in step 4 but not all. My goal is to get it on par with aws-amplify which still makes heavy use of `any` but with a focus on having the public interfaces defined.

powerful23:
@buggy That's a good point. But do we really need to use `webpack`, to generate a bundle? As I know we can generate  by only `tsc`. Also I would prefer working on `aws-amplify` first for the typescript enhancement. @manueliglesias @mlabieniec what do you think?

buggy:
[...]

PR #2436, opened 2019-01-01
Convert aws-amplify-react to TypeScript

buggy:
*Issue #, if available:*

#1028

*Description of changes:*

Converts the code base for aws-amplify-react from JavaScript to TypeScript.

Most of the components have interfaces and the few that don't have been typed as `any` so they will allow anything. Where an interface has been defined I'm confident all of the properties are included but I've only provided a type for the property where I'm confident it is correct. This means many properties will be typed as `any`.

I think the best approach is to merge this PR and use subsequent PR's to improve the interface definitions as I can't dedicate enough time completing it all in one hit.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3715, opened 2019-07-23
Convert aws-amplify-react to TypeScript

buggy:
Issue #, if available:

#1028

Description of changes:

Converts the code base for aws-amplify-react from JavaScript to TypeScript.

Most of the components have interfaces and the few that don't have been typed as any so they will allow anything. Where an interface has been defined I'm confident all of the properties are included but I've only provided a type for the property where I'm confident it is correct. This means many properties will be typed as any.

I think the best approach is to merge this PR and use subsequent PR's to improve the interface definitions as I can't dedicate enough time completing it all in one hit.


buggy:
This is a replacement for #2436. After 6 months there were just too many changes to the aws-amplify-react package to be able to keep that PR going.

I'll try to get the branch test this week so I can remove the WIP.

jordanranz:
Thank you for consolidating this into a new PR. I'll go ahead and pull down the branch and test once you remove WIP.

buggy:
@jordanranz I've removed the WIP after testing it with a new create-react-app application. If there are any problems I'll try to resolve them within 24 hours. If we need to co-ordinate a time for both of us to be online then DM me via Twitter.

It would be really good if we can merge this before a lot of changes happen to aws-amplify-react as the PR tends to become more risky the longer it stays open.

jordanranz:
I will focus on testing this and merging it before the end of the week.

packages/aws-amplify-react/tsconfig.json
```diff
@@ -10,10 +10,10 @@
         "esnext.asynciterable",
         "es2017.object"
       ], /* Specify library files to be included in the compilation. */
-      "allowJs": true, /* Allow javascript files to be compiled. */
+      // "allowJs": true, /* Allow javascript files to be compiled. */
```
jordanranz: I think commented values can probably be removed from this file unless you have a good argument to keep them.

buggy: I can't see any reason to keep it. All of the JS has gone now.

jskrt:
Hi, I'm wondering if there's any update on this?

packages/aws-amplify-react/src/Auth/SignUp.tsx
```diff
@@ -34,20 +34,37 @@ import {
...
 import { valid } from 'semver';
 import { PhoneField } from './PhoneField';
 
 const logger = new Logger('SignUp');
 
 
-export default class SignUp extends AuthPiece {
-    constructor(props) {
+export interface ISignUpConfig {
+  defaultCountryCode?: number;
```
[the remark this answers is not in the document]
buggy: I'll change the definition to allow strings because that actually makes more sense for telephone numbers (example "911" is 9... 1... 1... and not 911). In this case it doesn't matter because you need to enter a number without the + prefix. The only places they're used expect 1 or "1" and convert it to "+1"

elorzafe:
Yeah!

PR #3736, opened 2019-07-25
Add screenshot testing

bluezebragames:
Screenshot testing via Stencil + Jest.  Also fixes the jest-config problem, although there might be a better way to do it.


packages/amplify-ui-components/screenshot/.gitignore
```diff
@@ -0,0 +1,3 @@
+images
```
bluezebragames: This .gitignore is removing everything in the /screenshot/ folder.  I can't quite decide if that's a good or a bad idea -- it will certainly decrease the size of the repo, among other things, but is it really worth the downsides?  For example, if you've run tests on your computer (creating screenshots), but someone remotely updates the components, when you pull and re-run the tests they'll fail the screenshot comparison and you'll need to `--update-screenshot`.  Or for another example, suppose you pull the repo fresh and don't run the tests before you change the components -- then the screenshots aren't doing their job at all, since they're not stopping you from accidentally breaking the styling of the components.  All in all, I think there's a reason why snapshots are normally committed...

jordanranz: This is a good analysis. I think the 'baseline' screenshot should be committed so that everyone is comparing against the same image. The latest images generated to compare against the baseline image should be ignored. Is this possible with Stencil screenshot testing?

packages/amplify-ui-components/src/components/amplify-text-field/amplify-text-field.e2e.ts
```diff
@@ -7,6 +7,9 @@ describe('amplify-text-field', () => {
     await page.setContent(
       `<amplify-text-field field-id="id" label="Label test" description="Description test"></amplify-text-field>`,
     );
+    const screenshot = await page.compareScreenshot('Amplify Text Field', {fullPage: true});
```
jordanranz: Curious if there is a good reason to use `fullPage: true`?

bluezebragames: My thinking was that if we ever had a component that went off the side or bottom of the screen it would be a good idea to just pass this argument consistently so it's not hard to track down in the future.  It didn't make a difference on this component, I tried with and without it -- so we certainly could take it out.

packages/amplify-ui-components/screenshot/.gitignore
bluezebragames: Doesn't look like it.  The file names are generated via just a hash of the contents, so there's no easy way to know which one is the master.  Moreover there's not even a function for deleting a file, let alone removing all of the previous versions whenever a new master is set.

packages/amplify-ui-components/package.json
```diff
@@ -29,7 +30,7 @@
     "emotion": "^10.0.9"
   },
   "devDependencies": {
-    "@stencil/core": "^1.1.9",
+    "@stencil/core": "^1.2.0",
```
jordanranz: Bump this to 1.2.2

jordanranz:
Add a husky pre-push hook

Issue #3759, opened 2019-07-29
Storage global configuration does not accept `customPrefix`


aldo-roman:
**Describe the bug**
A check on configuration is preventing `customPrefix` attribute to be included in Storage global configuration:

https://github.com/aws-amplify/amplify-js/blob/799a1cfe59f7244f1cc51bdc0220a6c4ebf81455/packages/storage/src/Storage.ts#L121

According to `storageArrayKeys`, `customPrefix` is expected, and is an object according to docs. See also #3368

**To Reproduce**
1. Config Amplify as:
```ts
Storage.configure({
    ...
    level: 'public',
    customPrefix: {
        public: ''
    }
});
```
2. Run app
3. Upload a file to S3
4. File gets uploaded to `/public` dir

**Expected behavior**
File gets uploaded to root dir

**Desktop (please complete the following information):**
 - OS: All
 - Browser: All
 - Version: `"@aws-amplify/storage": "^1.0.33"`

jordanranz:
Related: https://github.com/aws-amplify/amplify-js/issues/3752

sammartinez:
@aldo-roman  Working on a fix now, should be up shortly

Issue #3752, opened 2019-07-29
Updating @aws-amplify/storage from 1.0.31 to 1.0.33 gives an "AccessDenied: Access Denied" error on `.put`


philiiiiiipp:
**Describe the bug**

It seems as #3051 is happening again.

The custom prefixes are getting ignored.

sammartinez:
Seems to be a bug related issue, will look into this and report back

sammartinez:
@philiiiiiipp Just want to make sure, from the [customization section in Storage](https://aws-amplify.github.io/docs/js/storage#customization) are your IAM policies up to date?

jordanranz:
Closing in favor of https://github.com/aws-amplify/amplify-js/issues/3759

PR #3760, opened 2019-07-29
Remove object check & value is undefined in order to resolve custom prefix issue

sammartinez:
*Issue #, if available:*
Fixes: #3759 
Fixes: #3752 

*Description of changes:*
Removing the check of an object for custom prefix to be added correctly to the AWSS3 provider. Added an additional test for custom prefix


packages/storage/__tests__/Storage-unit-test.ts
```diff
@@ -401,6 +401,35 @@ describe('Storage', () => {
                 }
             });
         });
+
+        test('should add customPrefix global if is defined', () => {
```
jordanranz: Can we add a failing test to accompany this.

sammartinez: absolutely

PR #3761, opened 2019-07-29
Sync angular dev-dependency with main package

wlee221:
*Issue #, if available:*

*Description of changes:* Dev dependency was out of sync and caused CircleCI fails on some branches.


sammartinez:
This is correct that we need to have this point to the correct version of aws amplify, going to merge as this was tested by Francisco and William both

PR #3767, opened 2019-07-30
AmplifyTextField & AmplifyTextInput -> AmplifyFormField & AmplifyInput *as well as* Amplify UI test change

bluezebragames:
- Change AmplifyTextField to AmplifyFormField and AmplifyTextInput to AmplifyInput
- Add Storybook integration and Stencil .e2e and .spec tests
  - including screenshot and snapshot tests
- Use Emotion for styling
- Add slots
- Add input placeholder prop

### AmplifyFormField:
AmplifyFormField

### AmplifyInput:
AmplifyInput


packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.tsx
```diff
@@ -15,14 +15,16 @@ export class AmplifyTextField {
   } = {};
 
   render() {
+    const type = this.inputProps.type || "text";
+
     return (
       <div class={formField}>
         {this.label && <amplify-label htmlFor={this.fieldId}>{this.label}</amplify-label>}
-        <amplify-text-input
+        <amplify-input
```
jordanranz: I think we can use a `<slot />` here instead of hardcoding `<amplify-input>`. This will allow the user to pass in whatever content they want here. It also makes the Field props more flexible.

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.tsx
```diff
@@ -15,14 +15,16 @@ export class AmplifyTextField {
   } = {};
 
   render() {
+    const type = this.inputProps.type || "text";
+
     return (
       <div class={formField}>
```
jordanranz: Probably out of scope for this PR, but we should decide to use either formField or Field thoughout.

packages/amplify-ui-components/src/components/amplify-field/amplify-field.tsx
```diff
@@ -1,20 +1,18 @@
...
-  @Prop() inputProps: {
-    type?: string;
-    onInput?: (Event) => void;
-  } = {};
+  @Prop() type?: string = "text";
+  @Prop() onInput?: (arg0: Event) => void;
 
   render() {
-    this.inputProps.type = this.inputProps.type || 'text';
+    this.type = this.type || 'text';
```
jordanranz: Not sure if this is necessary. Based off of the @Prop definition above, it should use the passed in string value or "text".

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.tsx
bluezebragames: wait... something's fishy here, right?  Like we already have an AmplifyTextField -> AmplifyField, in `components/amplify-field/amplify-field.tsx`...  I think this was supposed to be `AmplifyFormField` but maybe it just didn't get changed when it migrated over to this package?

bluezebragames: If my suspicion is right, I do want to put the fix in this PR

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.tsx
bluezebragames: ooh good call!  That is kind of a sizable change because it affects the `.e2e.ts` and `.stories.ts` files, so I might make that into a separate PR if that makes sense

packages/amplify-ui-components/src/components/amplify-field/amplify-field.tsx
bluezebragames: You're absolutely right, that completely slipped my mind.  I'll make the change in `amplify-input/amplify-input.tsx` too.

packages/amplify-ui-components/src/components/amplify-field/amplify-field.e2e.ts
```diff
@@ -57,12 +58,7 @@ describe('amplify-text-field', () => {
   it('can have multiple input types', async () => {
```
jordanranz: I'd consolidate this to test specific input types. For this one it would be 'checkbox'. You can additionally write a test for a different type like 'number'. This way if we want to expand the test it is focused on that type.

packages/amplify-ui-components/src/components/amplify-field/amplify-field.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+          <div>
+            <amplify-input type="text">
+            </amplify-input>
+          </div>
+        </div>
+      </amplify-field>`
+    );
+  });
+
+  it('renders a label and description, if they are provided, and no id', async () => {
```
jordanranz: I'd use this spec file to test more granular functionality of the component methods. Checking for HTML or DOM composition from the fields passed in should be covered in the e2e tests.

packages/amplify-ui-components/src/components/amplify-field/amplify-field.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+          <div>
+            <amplify-input type="text">
+            </amplify-input>
+          </div>
+        </div>
+      </amplify-field>`
+    );
+  });
+
+  it('renders with an id, if it is provided', async () => {
```
jordanranz: Same as above.

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.stories.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { text as textKnob } from '@storybook/addon-knobs';
+
+storiesOf('amplify-form-field', module).add('with knobs', () => {
+  const label = textKnob('Label', 'Label placeholder');
+  const description = textKnob('Description', 'Description placeholder');
+  const type = textKnob('Type', 'text');
+  return `<amplify-form-field field-id="id" label="${label}" description="${description}" type="${type}"/>`;
+});
+
+storiesOf('amplify-form-field', module).add('password example', () => {
```
jordanranz: I like using multiple examples but lets give access to the knobs for all stories if possible.

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.tsx
```diff
@@ -1,33 +1,48 @@
 import { Component, Prop, h } from '@stencil/core';
-import { formField } from './amplify-form-field.style';
+import { TextFieldTypes } from '../../common/types';
 
 @Component({
   tag: 'amplify-form-field',
-  shadow: false,
```
jordanranz: `shadow: false` is the default but let's be explicit here and keep `shadow: false`

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.tsx
```diff
@@ -1,33 +1,48 @@
 import { Component, Prop, h } from '@stencil/core';
-import { formField } from './amplify-form-field.style';
```
jordanranz: Keep this import and attach it to the root dom element's class.

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.tsx
```diff
@@ -1,33 +1,48 @@
...
+import { TextFieldTypes } from '../../common/types';
 
 @Component({
   tag: 'amplify-form-field',
-  shadow: false,
+  styleUrl: 'amplify-form-field.css',
```
jordanranz: Remove this for now.

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.tsx
```diff
@@ -1,33 +1,48 @@
...
+  /** The text of a hint to the user as to how to fill out the input.  Goes just below the input. */
   @Prop() hint: string | null;
-  @Prop() inputProps: {
-    type?: string;
-    onInput?: (Event) => void;
-  } = {};
+  /** The input type.  Can be any HTML input type. */
+  @Prop() type?: TextFieldTypes = "text";
+  /** The callback, called when the input is modified by the user. */
+  @Prop() onInput?: (arg0: Event) => void;
```
jordanranz: can we change `arg0` to be a bit more descriptive.

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.tsx
```diff
@@ -1,33 +1,48 @@
...
-        {this.label && <amplify-label htmlFor={this.fieldId}>{this.label}</amplify-label>}
-        <amplify-text-input
-          id={this.fieldId}
-          aria-describedby={this.fieldId && this.hint ? `${this.fieldId}-description` : null}
-          type="text"
-          {...this.inputProps}
-        />
+      <div>
+        {this.label && (
+          <label htmlFor={this.fieldId} class="label">
```
jordanranz: Let's export these from the emotion file.

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.tsx
```diff
@@ -1,33 +1,48 @@
...
+            {this.label}
+          </label>
+        )}
+        {this.description && (
+          <div id={`${this.fieldId}-description`} class="description">
```
jordanranz: Same as above.

packages/amplify-ui-components/src/components/amplify-form-field/readme.md
```diff
@@ -1,4 +1,4 @@
-# amplify-form-field
+# amplify-field
```
jordanranz: How did this get autogenerated?

packages/amplify-ui-components/src/components/amplify-input/amplify-input.e2e.ts
```diff
@@ -0,0 +1,58 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('amplify-input', () => {
+  it('renders with a different test name', async () => {
```
jordanranz: different test name? maybe write just name this "renders"

packages/amplify-ui-components/src/components/amplify-input/amplify-input.e2e.ts
```diff
@@ -0,0 +1,58 @@
...
+    const page = await newE2EPage();
+
+    await page.setContent(
+      `<amplify-input field-id="id" description="Description test"></amplify-input>`,
+    );
+    const screenshot = await page.compareScreenshot('Amplify Input', {fullPage: true});
+    expect(screenshot).toMatchScreenshot({ allowableMismatchedPixels: 10 });
```
jordanranz: Let's move `10` to a const or a common test file

packages/amplify-ui-components/src/components/amplify-input/amplify-input.e2e.ts
```diff
@@ -0,0 +1,58 @@
...
+    await page.exposeFunction('exposedfunc', func);
+
+    await page.$eval('amplify-input', (inputElement: any) => {
+      inputElement.onInput = this.exposedfunc;
+      inputElement.label = 'adding a label so that the component rerenders';
+    });
+    await page.waitForChanges();
+
+    const input = await page.find('input');
+    await input.press('8');
```
jordanranz: Let's test a few types of values here. Like character 'a' or special character '$'

packages/amplify-ui-components/src/components/amplify-input/amplify-input.e2e.ts
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it('can have a number input', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(`<amplify-input type='number'></amplify-input>`);
+    await page.waitForChanges();
+
+    const input = await page.find('input');
+    expect(input).toEqualAttribute('type', 'number');
```
jordanranz: Test happy case that a number can be entered and unhappy case that a character cannot be entered here.

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -0,0 +1,29 @@
...
+})
+export class AmplifyInput {
+  /** The ID of the field.  Should match with its corresponding input's ID. */
+  @Prop() fieldId: string;
+  /** The text of the description.  Goes just below the label. */
+  @Prop() description: string | null;
+  /** The input type.  Can be any HTML input type. */
+  @Prop() type?: TextFieldTypes = "text";
+  /** The callback, called when the input is modified by the user. */
+  @Prop() onInput?: (arg0: Event) => void;
```
jordanranz: Maybe a more descriptive arg name here.

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  render() {
+    return (
+      <input
+        id={this.fieldId}
+        aria-describedby={this.fieldId && this.description ? `${this.fieldId}-description` : null}
+        type={this.type}
+        onInput={this.onInput}
+        class={textInput}
```
jordanranz: this could be just `input`

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.spec.ts
```diff
@@ -0,0 +1,40 @@
+import { newSpecPage } from '@stencil/core/testing';
+import { AmplifyFormField } from './amplify-form-field.tsx'
```
jordanranz: You shoudn't need `.tsx` at the end here

packages/amplify-ui-components/src/components/amplify-input/amplify-input.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { newSpecPage } from '@stencil/core/testing';
+import { AmplifyInput } from './amplify-input.tsx'
```
jordanranz: You can remove `.tsx` at the end

packages/amplify-ui-components/src/components/amplify-form-field/readme.md
bluezebragames: I still don't know, but re-running `$ stencil build` didn't fix it; the only workaround was to change it manually for some reason

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.tsx
```diff
@@ -1,34 +1,53 @@
...
       <div class={formField}>
-        {this.label && <amplify-label htmlFor={this.fieldId}>{this.label}</amplify-label>}
-        <amplify-text-input
-          id={this.fieldId}
-          aria-describedby={this.fieldId && this.hint ? `${this.fieldId}-description` : null}
-          type="text"
-          {...this.inputProps}
-        />
+        {this.label && (
+          <amplify-label htmlFor={this.fieldId} class={formLabel}>
```
jordanranz: The `amplify-label` component should handle it's own class assignment

bluezebragames: My thinking was just that it's out of scope for this pr.  Definitely eventually but do you think that now is the right time for it?  It's a small change, `amplify-label` already uses Emotion styling so it's just moving the style over

bluezebragames: The other thing to consider is whether or not this particular label is special because it's a part of this form field.  i.e. `amplify-label` has a default style but we have additional style on top of it because it's in `amplify-form-field`

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.style.ts
```diff
@@ -1,5 +1,32 @@
 import { css } from 'emotion';
 
 export const formField = css`
-  margin-bottom: 22px;
+  input {
```
jordanranz: This should probably live in the `amplify-input` style file

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.e2e.ts
```diff
@@ -1,48 +1,50 @@
...
-    expect(textElement).not.toBeNull();
+    const fieldElement = await page.find('amplify-form-field');
+    expect(fieldElement).not.toBeNull();
 
     const labelElement = await page.find('label');
     expect(labelElement).toEqualText('Label test');
     expect(labelElement).toEqualAttribute('for', 'id');
 
-    const descriptionElement = await page.find('.description');
+    const descriptionElement = await page.find(`.${formDescription}`);
```
bluezebragames: will add data-test attributes

packages/amplify-ui-components/.storybook/webpack.config.js
```diff
@@ -12,5 +12,12 @@ module.exports = ({ config, mode }) => {
     config || {}
   );
 
+  config.module.rules.push({
```
sammartinez: wouldnt this just live under [rules](https://webpack.js.org/concepts/loaders/#example)?

packages/amplify-ui-components/package.json
```diff
@@ -14,8 +14,9 @@
     "lint": "eslint --ext .js,.ts,.tsx .",
     "format": "eslint --fix --ext .js,.ts,.tsx .",
     "pretest": "npm run lint",
-    "test": "stencil test --spec --e2e --screenshot",
-    "test:update-screenshot": "stencil test --spec --e2e --screenshot --update-screenshot",
+    "test": "stencil test --spec --e2e",
+    "test:screenshot": "stencil test --spec --e2e --screenshot",
+    "test:update-screenshot": "rm -f screenshot/builds/master.json && rm -f screenshot/images/* && stencil test --spec --e2e --screenshot --update-screenshot",
```
sammartinez: 🥇

packages/amplify-ui-components/src/common/testing.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const pixelThreshold = 10;
+
+export const selectors = {
+    // amplify-form-field
+    formFieldDescription: '[data-test=form-field-description]',
+}
+
+export const knobs = {
+    labelKnob: function (placeholder: string) {
```
sammartinez: since we are using es6, we may be able to short hand these functions. 
`labelKnob(placeholder: string) {...}`

WDYT?

sammartinez: Also would the parameter for the function be called label? and would have a string type? Same callout as the below methods so it doesnt look like I went off on you :)

packages/amplify-ui-components/src/common/types.ts
```diff
@@ -0,0 +1 @@
+export type TextFieldTypes = 'date' | 'email' | 'number' | 'password' | 'search' | 'tel' | 'text' | 'url' | 'time';
```
sammartinez: The [tel](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel) is fine but we may want to think about an optional parameter for a pattern then that a customer can pass in if that is gonna be one of the accepting types for an input. Doesnt have to be solved this PR but a note to write down at least

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.e2e.ts
```diff
@@ -0,0 +1,78 @@
+import { newE2EPage } from '@stencil/core/testing';
+import { pixelThreshold, selectors } from '../../common/testing';
+
+describe('amplify-form-field', () => {
+  it('renders', async () => {
```
sammartinez: renders what? Can we be slightly more descriptive for this

packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.e2e.ts
```diff
@@ -0,0 +1,78 @@
...
+    const page = await newE2EPage();
```
sammartinez: Im noticing that alot of the below is using a new page instances. We may want to have this at least in a beforeEach. Can leave it up to you.

packages/amplify-ui-components/.storybook/webpack.config.js
bluezebragames: good point, I just stole this from @jordanranz's PR

packages/amplify-ui-components/src/common/testing.ts
bluezebragames: Yeah it's a placeholder for that the label knob will be, but maybe `labelPlaceholder` is a better way of conveying that -- although I shudder to think about `placeholderPlaceholder`

packages/amplify-ui-components/.storybook/webpack.config.js
bluezebragames: actually that's really weird, Storybook isn't working if I move this up into the actual config object, but if I leave it where it is now, it works

bluezebragames: I'm very confused

jordanranz:
I like the screenshots in the PR description, nice work

packages/amplify-ui-components/.storybook/webpack.config.js
jordanranz: This is exactly what I ran into as well. Out of scope of this PR to simplify

packages/amplify-ui-components/src/common/testing.ts
```diff
@@ -0,0 +1,30 @@
+import { select as selectKnob, text as textKnob } from '@storybook/addon-knobs';
```
jordanranz: Nice work consolidating this into a common file.

Issue #3620, opened 2019-07-10
Error when building with AOT enabled when using Angular


jeremyputeaux:
**Describe the bug**
I'm running a basic ng build --prod and I'm getting the following error:

```
ERROR in node_modules/aws-amplify-angular/dist/src/components/authenticator/confirm-sign-in-component/confirm-sign-in-component.core.d.ts.ConfirmSignInComponentCore.html(7,55): : Property 'amplifyService' is protected and only accessible within class 'ConfirmSignInComponentCore' and its subclasses.
node_modules/aws-amplify-angular/dist/src/components/authenticator/confirm-sign-in-component/confirm-sign-in-component.core.d.ts.ConfirmSignInComponentCore.html(16,11): : Property 'amplifyService' is protected and only accessible within class 'ConfirmSignInComponentCore' and its subclasses.
node_modules/aws-amplify-angular/dist/src/components/authenticator/confirm-sign-in-component/confirm-sign-in-component.core.d.ts.ConfirmSignInComponentCore.html(46,42): : Property 'amplifyService' is protected and only accessible within class 'ConfirmSignInComponentCore' and its subclasses.
node_modules/aws-amplify-angular/dist/src/components/authenticator/confirm-sign-in-component/confirm-sign-in-component.core.d.ts.ConfirmSignInComponentCore.html(5,81): : Property 'amplifyService' is protected and only accessible within class 'ConfirmSignInComponentCore' and its subclasses.
node_modules/aws-amplify-angular/dist/src/components/authenticator/confirm-sign-in-component/confirm-sign-in-component.core.d.ts.ConfirmSignInComponentCore.html(27,16): : Property 'amplifyService' is protected and only accessible within class 'ConfirmSignInComponentCore' and its subclasses.
node_modules/aws-amplify-angular/dist/src/components/authenticator/confirm-sign-in-component/confirm-sign-in-component.core.d.ts.ConfirmSignInComponentCore.html(37,14): : Property 'amplifyService' is protected and only accessible within class 'ConfirmSignInComponentCore' and its subclasses.
node_modules/aws-amplify-angular/dist/src/components/authenticator/confirm-sign-up-component/confirm-sign-up.component.core.d.ts.ConfirmSignUpComponentCore.html(68,40): : Property 'amplifyService' is protected and only accessible within class 'ConfirmSignUpComponentCore' and its subclasses.
[...]

PR #3793, opened 2019-08-01
fixes(#3620)

haverchuck:
*Issue #, if available:*
fixes(#3620)

*Description of changes:*
This will allow users to use the strict template check option for Angular AOT compilation.


bluezebragames:
The changes definitely look good.  Just recording my testing experience here before I go...

It took me a while, but I got the changes to be included correctly and it looks like it does in fact fix the issue with `protected` variables not being accessible.  However, I was still unable to get the angular project to compile with AOT for two reasons:

1.  `rxjs`.  `error TS2307: Cannot find module 'rxjs/Subject'`, in `import { Subject } from 'rxjs/Subject';` -- it looks like in `rxjs` version 6, it's just `import { Subject } from 'rxjs';`.  I tried pinning `rxjs` to a lower version, but Angular requires at least `rxjs 5.6.0-forward-compat.0`, which still gives the `Cannot find module 'rxjs/Subject'` errors (plus some more, to boot).

2.  `@ViewChild`.  The error here was `An argument for 'opts' was not provided` -- in Angular 8, `@ViewChild` requires two parameters instead of one, also passing an object like `{static: false}`.  I feel like the previous problem might just be because of a versioning error or something, but this one I'm pretty sure will need to be fixed before Angular 8 AOT support at least.

antoinebrault:
any ETA for the merge ?

dsoegijono:
@haverchuck Any chance of updating this?

sammartinez:
Closing pull request as issue was addressed

PR #3806, opened 2019-08-05
feat: Add support for local testing of api and storage

Ashish5591:
*Description of changes:*
This code adds the functionality necessary to support local testing of api and storage once enabled with the cli

### DIFF
--- a/packages/core/src/Parser.ts
+++ b/packages/core/src/Parser.ts
@@ -35,7 +35,9 @@ export default class Parser {
             storageConfig = {
                 AWSS3: {
                     bucket: config['aws_user_files_s3_bucket'],
-                    region: config['aws_user_files_s3_bucket_region']
+                    region: config['aws_user_files_s3_bucket_region'],
+                    dangerouslyConnectToHttpEndpointForTesting:
+                    config['aws_user_files_s3_dangerously_connect_to_http_endpoint_for_testing']
                 }
             };
         } else {
--- a/packages/pubsub/__tests__/PubSub-unit-test.ts
+++ b/packages/pubsub/__tests__/PubSub-unit-test.ts
@@ -7,6 +7,10 @@ import * as Paho from '../src/vendor/paho-mqtt';
 
 const pahoClientMockCache = {};
 
+const mockConnect = jest.fn((options) => {
+    options.onSuccess();
+});
+
 const pahoClientMock = jest.fn().mockImplementation((host, clientId) => {
     if (pahoClientMockCache[clientId]) {
         return pahoClientMockCache[clientId];
@@ -14,9 +18,7 @@ const pahoClientMock = jest.fn().mockImplementation((host, clientId) => {
 
     var client = {} as any;
 
-    client.connect = jest.fn((options) => {
-        options.onSuccess();
-    });
+    client.connect = mockConnect;
     client.send = jest.fn((topic, message) => {
         client.onMessageArrived({ destinationName: topic, payloadString: message });
     });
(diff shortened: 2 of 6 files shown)


packages/core/src/Parser.ts
```diff
@@ -35,7 +35,9 @@ export default class Parser {
             storageConfig = {
                 AWSS3: {
                     bucket: config['aws_user_files_s3_bucket'],
-                    region: config['aws_user_files_s3_bucket_region']
+                    region: config['aws_user_files_s3_bucket_region'],
+                    dangerouslyConnectToHttpEndpointForTesting:
```
sammartinez: reading `dangerouslyConnectToHttpEndpointForTesting` is this gonna be the name of the variable? Should it not be `dangerouslyConnectToHttpEndpointForLocalTesting` ?

packages/pubsub/__tests__/PubSub-unit-test.ts
```diff
@@ -213,6 +215,32 @@ describe('PubSub', () => {
         });
     });
 
+    describe('MqttOverWSProvider local testing config', () => {
+        test('ssl should be disabled in the case of local testing', async () => {
+            mockConnect.mockClear();
```
sammartinez: This may be better suited to sit in a `beforeEach(...)` if Im reading this right. WDYT?

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
```diff
@@ -102,6 +101,26 @@ describe('StorageProvider test', () => {
...
+
+            const config = storage.configure(aws_options);
+            expect(config).toEqual({
+                AWSS3: {
+                    "bucket": "bucket",
+                    "region": "region",
+                    "dangerouslyConnectToHttpEndpointForTesting": true
+                }
+            });
+        });
```
sammartinez: Can we add a fail case as well?

packages/core/src/Parser.ts
Ashish5591: I'm not opposed to it, but this is the name that was decided for the flag in the design review, so decided to stick to that.

sammartinez: Got it !

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
sammartinez: Awe didnt see the case above, as per our offline convo, I am fine if we adjust the name of the `happy case` is all

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
```diff
@@ -87,7 +86,7 @@ describe('StorageProvider test', () => {
     });
 
     describe('configure test', () => {
-        test('happy case', () => {
+        test('standard configuration', () => {
```
sammartinez: 🚀

packages/storage/src/Providers/AWSS3Provider.ts
```diff
@@ -39,6 +39,8 @@ const dispatchStorageEvent = (track:boolean, event:string, attrs:any, metrics:an
     }
 };
 
+const localTestingStorageEndpoint = 'http://localhost:20005';
```
elorzafe: I think this should be part of the config, not hardcoded.

Ashish5591: I agree. However this was an explicit decision made by the CLI team in their design review. They may add support for providing the url in the config in the future.

lukeramsden: Is there any public explanation as to why they made this decision? I'd like to proxy my local mocking endpoint so I can run it over HTTPS, but because this is hardcoded for some reason, I can't.

Issue #3753, opened 2019-07-29
Auth: Loader while waiting for authentication response


hlynurlogi:
**Context**
I am using NextJS with `aws-amplify`, also using `aws-amplify-react`. 

**Is your feature request related to a problem? Please describe.**
I want to be able to display a loader indicating the process of authentication is going on, loader to be shown as soon as user presses **Sign In** or **Sign up** and disappears when response has been received from server.

At the moment nothing is shown in the UI that indicates that the user has successfully submitted his credentials and that the user should be waiting for a response/to be rerouted to another page.

**Describe the solution you'd like**
An ability to render my own loading component wherever I want on the screen while the authentication process is ongoing. Or at least get some out of the box loaders which I can change the colors of.

**Describe alternatives you've considered**
Nothing

**Additional context**
![image](https://user-images.githubusercontent.com/47002112/62049288-696be900-b1fe-11e9-963d-e5b7126c237f.png)

Above I have just pressed sign in, how am I supposed to know that my data was submitted successfully?

PR #3820, opened 2019-08-08
Change the signIn button text when loading

Anenth:
*Issue #, if available:*
Right now there is no indication about after user click on signin

*Description of changes:*
Changing the text button text to 'loading...' will help the user understand there is something happening.
This also helps here #3753 

### DIFF
--- a/packages/aws-amplify-react/src/Auth/SignIn.jsx
+++ b/packages/aws-amplify-react/src/Auth/SignIn.jsx
@@ -161,7 +161,7 @@ export default class SignIn extends AuthPiece {
                             disabled={this.state.loading}
                             data-test={auth.signIn.signInButton}
                             >
-                            {I18n.get('Sign In')}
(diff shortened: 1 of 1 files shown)


jordanranz:
I like this improvement but we will need to add this improvement to each of the other frameworks we support to maintain consistency for the UX. Is this something you are willing to add to the PR?

Anenth:
@jordanranz that make sense, I don't really have much experience other frameworks, maybe we could create issues for assigned to people who are good at respective frameworks?

jordanranz:
I think we may want to use a spinner icon instead of actually rendering text on the button here. We are in the process of doing a UI component refactor to utilize web components under the covers. This is definitely a feature we would like to improve on for that project. I'll add the milestone to this PR.

Anenth:
@jordanranz do you have any discussion about the "utilize web components under the covers" you mentioned here?

PR #3821, opened 2019-08-08
fills out defaultValue for username when querystring present

haverchuck:
*Description of changes:*
Uses defaultValue instead of 'value' in username field when username param is present in querystring.


packages/aws-amplify-react/src/Auth/AuthPiece.jsx
```diff
@@ -49,6 +49,16 @@ export default class AuthPiece extends React.Component {
         this.onPhoneNumberChanged = this.onPhoneNumberChanged.bind(this);
     }
 
+    componentDidMount() {
+        if (window && window.location && window.location.search) {
+            if (!this.props.authData || !this.props.authData.username) {
+                const searchParams = new URLSearchParams(window.location.search);
+                let username = searchParams ? searchParams.get('username') : undefined;
+                this.setState({username})
```
sammartinez: fix typo `;`

haverchuck: done

packages/aws-amplify-react/src/Auth/AuthPiece.jsx
```diff
@@ -87,16 +97,11 @@ export default class AuthPiece extends React.Component {
...
-            if (window && window.location && window.location.search) {
-                const searchParams = new URLSearchParams(window.location.search);
-                value = searchParams ? searchParams.get('username') : undefined
-            }
             return (
                 <FormField theme={theme}>           
                     <InputLabel theme={theme}>{I18n.get(this.getUsernameLabel())} *</InputLabel>
                     <Input
-                        value={value}
+                        defaultValue={this.state.username}
```
sammartinez: I believe we can keep this as `value` but just use the state `value={this.state.username}`

packages/aws-amplify-react/src/Auth/AuthPiece.jsx
```diff
@@ -49,6 +49,16 @@ export default class AuthPiece extends React.Component {
         this.onPhoneNumberChanged = this.onPhoneNumberChanged.bind(this);
     }
 
+    componentDidMount() {
+        if (window && window.location && window.location.search) {
+            if (!this.props.authData || !this.props.authData.username) {
+                const searchParams = new URLSearchParams(window.location.search);
+                let username = searchParams ? searchParams.get('username') : undefined;
```
sammartinez: Can we change this to a `const` since the next line is the location we will be using it.

packages/aws-amplify-react/src/Auth/AuthPiece.jsx
```diff
@@ -49,6 +49,16 @@ export default class AuthPiece extends React.Component {
...
+                this.setState({username});
```
sammartinez: NIT- NON-BLOCKING: can we add spaces around username? Just will make it more readable

packages/aws-amplify-react/src/Auth/AuthPiece.jsx
haverchuck: I think 'value' won't let them change it (which in 99% of cases they won't anyway, but still).

packages/aws-amplify-react/src/Auth/AuthPiece.jsx
haverchuck: done

packages/aws-amplify-react/src/Auth/AuthPiece.jsx
haverchuck: done

PR #3897, opened 2019-08-22
feat(@aws-amplify/auth): Error messaging

iartemiev:
*Issue #, if available:*
https://www.pivotaltracker.com/story/show/159083701

*Description of changes:*

* Created custom classes to standardize error handling in Auth (`auth/src/Errors.ts`)
The class consists of `message` and `log` properties. The former gets displayed in the header to the end-user and `log` gets printed to the console.
* Added a dict containing error messages. 
* Detecting 2 different causes of the noUserpool error and offering solutions on how to fix them (in the console). 
* Will be adding a couple of additional noUserpool scenarios to the docs (couldn't detect them programmatically) 

Notes:
* I had prettier enabled, so there are some formatting changes (sorry!)
* docs will be done a bit later today 

### DIFF
--- a/packages/auth/__tests__/auth-unit-test.ts
+++ b/packages/auth/__tests__/auth-unit-test.ts
@@ -179,6 +179,8 @@ import Cache from '@aws-amplify/cache';
 import { CookieStorage, CognitoUserPool, CognitoUser, CognitoUserSession, CognitoIdToken, CognitoAccessToken } from 'amazon-cognito-identity-js';
 import { CognitoIdentityCredentials } from 'aws-sdk';
 import { Credentials, GoogleOAuth, StorageHelper } from '@aws-amplify/core';
+import { AuthError, NoUserPoolError } from '../src/Errors';
+import { AuthErrorTypes } from '../src/types/Auth';
 
 const authOptions : AuthOptions = {
     userPoolId: "awsUserPoolsId",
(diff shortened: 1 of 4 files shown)


packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -427,27 +415,22 @@ describe('auth unit test', () => {
             spyon.mockClear();
         });
 
-        test('no user pool', async () => {
-            // @ts-ignore
+        test('no user pool in config', async () => {
```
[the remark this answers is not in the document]
sammartinez: Wonder if we should adjust the describe blocks to link up more with the `resendSignUp` `confirmSignUp` and so on.

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -870,12 +854,11 @@ describe('auth unit test', () => {
                 Pool: userPool
             });
 
-            expect.assertions(1);
-            try {
-                await auth.completeNewPassword(user, null, {});
-            } catch (e) {
-                expect(e).toBe('Password cannot be empty');
-            }
+            const error = new AuthError(AuthErrorTypes.EmptyPassword);
```
sammartinez: `error` or `errorMessage`? Everything above with this pattern seems to be an `errorMessage` just want to have consistency if we can :)

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -1322,13 +1303,11 @@ describe('auth unit test', () => {
                 Username: 'username',
                 Pool: userPool
             });
+            const error = new AuthError(AuthErrorTypes.EmptyCode);
```
sammartinez: Same callout as above

packages/auth/src/Errors.ts
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright 2017-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sammartinez: Can you update this to `2019-2020`

packages/auth/src/Errors.ts
```diff
@@ -0,0 +1,106 @@
...
+    noConfig: {
+        message: DEFAULT_MSG,
+        log: `
+            Error: Amplify has not been configured correctly.
+            This error is typically caused by one of the following scenarios:
+            
+            1. Make sure you're passing the awsconfig object to Amplify.configure() in your app's entry point
+                See https://aws-amplify.github.io/docs/js/authentication#configure-your-app for more information
+            
+            2. There might be multiple conflicting versions of aws-amplify or amplify packages in your node_modules.
```
sammartinez: 🔥

packages/auth/src/Errors.ts
```diff
@@ -0,0 +1,106 @@
...
+            Did you run \`amplify push\` after adding auth via \`amplify add auth\`?
+            See https://aws-amplify.github.io/docs/js/authentication#amplify-project-setup for more information
+        `
+    },
+    emptyUsername: {
+        message: 'Username cannot be empty'
+    },
+    invalidUsername: {
+        message:
+            'The username should either be a string or one of the sign in types'
```
sammartinez: Just a callout, non-blocking, may want to have a link for all the available sign in types

packages/auth/src/types/Auth.ts
```diff
@@ -11,122 +11,139 @@
  * and limitations under the License.
  */
 
- import { ICookieStorageData, ICognitoStorage, CognitoUserAttribute } from "amazon-cognito-identity-js";
+import {
+    ICookieStorageData,
+    ICognitoStorage,
+    CognitoUserAttribute
+} from 'amazon-cognito-identity-js';
```
sammartinez: **PRETTIER!**

packages/auth/src/types/Auth.ts
```diff
@@ -138,27 +155,49 @@ export interface Auth0OAuthOpts {
...
+    EmptyCode = 'emptyCode',
+    SignUpError = 'signUpError',
+    NoMFA = 'noMFA',
+    InvalidMFA = 'invalidMFA',
+    EmptyChallengeResponse = 'emptyChallengeResponse',
+    NoUserSession = 'noUserSession',
+    Default = 'default'
+}
+
+export type AuthErrorMessages = { [key in AuthErrorTypes]?: AuthErrorMessage };
```
sammartinez: Why would this be optional for the `...[key in AuthErrorTypes]?:..`

packages/auth/src/Errors.ts
iartemiev: Fair. Probably makes more sense to add that to the log, rather than the message, right?

packages/auth/src/types/Auth.ts
iartemiev: Good catch

packages/auth/src/Errors.ts
sammartinez: yup yup :) If there was a log for this bad boy but yeah makes sense to me. Just wanted to be a callout. Im not positive if we have that list today

packages/auth/__tests__/auth-unit-test.ts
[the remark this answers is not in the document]
iartemiev: It is in a separate `describe` but I don't think there's any value in running the same test multiple times regardless. I'll remove the redundant ones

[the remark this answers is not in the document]
iartemiev: They are in separate descibes. Same name because it's testing the same functionality but for different methods

PR #3900, opened 2019-08-23
Hint component

bluezebragames:
*Issue #, if available:*

*Description of changes:*


packages/amplify-ui-components/src/index.html
```diff
@@ -11,7 +11,7 @@
     </script>
     <script type="module" src="build/amplify-ui-components.esm.js"></script>
     <script nomodule src="build/amplify-ui-components.js"></script>
-    <link href="/build/amplify-ui-components.css" rel="stylesheet">
+    <link href="build/amplify-ui-components.css" rel="stylesheet">
```
bluezebragames: This change is here only so that I can throw out the callout that the directory where these files live is dist/amplify-ui-components/ and not build/...

packages/amplify-ui-components/src/components/amplify-hint/amplify-hint.e2e.ts
```diff
@@ -0,0 +1,19 @@
+import { E2EPage, newE2EPage } from '@stencil/core/testing';
+import { pixelThreshold } from '../../common/testing';
+
+describe('amplify-hint', () => {
```
sammartinez: can we add the `e2e` call out, so it should read

`describe('amplify-hint e2e:'`

packages/amplify-ui-components/src/components/amplify-hint/amplify-hint.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { newSpecPage } from '@stencil/core/testing';
+import { AmplifyHint } from './amplify-hint';
+
+describe('amplify-input', () => {
```
sammartinez: Same callout as above but with `spec:`

sammartinez: Also are we going to follow the structure for Component Logic and Render logic describe blocks?

packages/amplify-ui-components/src/index.html
sammartinez: can we adjust to point to the dist then ?

bluezebragames: that's what I was thinking, but I'm not sure if it's supposed to be this way or not

packages/amplify-ui-components/src/components/amplify-hint/amplify-hint.e2e.ts
bluezebragames: right, good reminder

packages/amplify-ui-components/src/components/amplify-hint/amplify-hint.spec.ts
bluezebragames: yep yep

packages/amplify-ui-components/src/index.html
bluezebragames: also, is this PR the right place for that change?  I can rebase these two components yet again if necessary

Issue #876, opened 2018-05-21
Not usable in NodeJS


sallar:
**Do you want to request a feature or report a bug?**
Bug Report

**What is the current behavior?**
Throws an Error when using in NodeJS

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than AWS Amplify.**
1. Require this module in NodeJS
2. Try to use it

**What is the expected behavior?**
The codebase should not have direct references to `window` object at all. This is considered bad practice. Authentication for example, is something one might want to use in NodeJS instead of browser or RN. There should be getters instead of accessing `window`, so they can be polyfilled when necessary.

Also if you're not going to support NodeJS, please kindly specify this in the docs or README, right now the docs are very vague about this matter.

Thank you ♥️

mlabieniec:
@sallar, we have included some [polyfills](https://github.com/aws/aws-amplify/blob/master/packages/aws-amplify/src/Common/Polyfills/Polyfills.ts) that polyfill the window object so that it is usable in node. The only other thing you would need to do is include a fetch module i.e. [node-fetch](https://www.npmjs.com/package/node-fetch) for the fetch requests done by the amazon cognito sdk, see here: https://github.com/aws/aws-amplify/issues/403

sallar:
@mlabieniec thanks. But the polyfill doesn’t seem to work. As soon as you `import` the module it throws an error. Is there any example of usage in node?

sallar:
The errors appear to come from importing everything from RNComponents folder

wzup:
Same error
https://github.com/aws/aws-amplify/issues/1066

elorzafe:
Hi @flaviocordova I check on node.js an App that uses Auth and is working if you install node-fetch 
and then add `global.fetch = require('node-fetch');` before importing Amplify.

flaviocordova:
Hi @elorzafe, thanks for your prompt reply...

My problem (or anyone with a vuejs ssr application) is not that nodejs should mock the browser environment better but instead should not do that at all..

The point here is that amplify should *not* create a window object because it's making the server-side code to misbehave.

take a look at https://github.com/vuejs/vue/blob/3eb37acf98e2d9737de897ebe7bdb7e9576bcc21/src/core/util/env.js, line 7 and you'll find it:

```
export const inBrowser = typeof window !== 'undefined'
```

[...]

PR #3902, opened 2019-08-23
NodeJS support for amazon-cognito-identity-js

exoego:
Closes https://github.com/aws-amplify/amplify-js/issues/876

*Description of changes:*
1. Use string `"nodejs"` as fallback value for `navigator.userAgent` if no global `navigator` object.
1. Use [cross-fetch](https://github.com/lquixada/cross-fetch) to support both browsers and node.js

### DIFF
--- a/packages/amazon-cognito-identity-js/src/Client.js
+++ b/packages/amazon-cognito-identity-js/src/Client.js
@@ -1,4 +1,6 @@
+import fetch from 'cross-fetch';
 import UserAgent from './UserAgent';
+
 /** @class */
 export default class Client {
 	/**
--- a/packages/amazon-cognito-identity-js/src/Client.js
+++ b/packages/amazon-cognito-identity-js/src/Client.js
@@ -1,6 +1,4 @@
-import fetch from 'cross-fetch';
 import UserAgent from './UserAgent';
-
 /** @class */
 export default class Client {
 	/**
--- a/packages/amazon-cognito-identity-js/src/CognitoUser.js
+++ b/packages/amazon-cognito-identity-js/src/CognitoUser.js
@@ -70,6 +70,9 @@ import StorageHelper from './StorageHelper';
  * @param {bool=} userConfirmationNecessary User must be confirmed.
  */
 
+const isBrowser = typeof navigator !== "undefined";
+const userAgent = isBrowser ? navigator.userAgent : "nodejs";
+
 /** @class */
 export default class CognitoUser {
 	/**
@@ -544,7 +547,7 @@ export default class CognitoUser {
 						DeviceKey: newDeviceMetadata.DeviceKey,
 						AccessToken: this.signInUserSession.getAccessToken().getJwtToken(),
 						DeviceSecretVerifierConfig: deviceSecretVerifierConfig,
-						DeviceName: navigator.userAgent,
+						DeviceName: userAgent,
 					},
 					(errConfirm, dataConfirm) => {
 						if (errConfirm) {
@@ -926,7 +929,7 @@ export default class CognitoUser {
 									.getAccessToken()
 									.getJwtToken(),
 								DeviceSecretVerifierConfig: deviceSecretVerifierConfig,
-								DeviceName: navigator.userAgent,
+								DeviceName: userAgent,
 							},
 							(errConfirm, dataConfirm) => {
 								if (errConfirm) {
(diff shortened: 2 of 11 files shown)


corollari:
You should also modify the following instructions in README.md to be in line with your changes:
```
**Note:** This library uses the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). For [older browsers](https://caniuse.com/#feat=fetch) or in Node.js, you may need to include a polyfill. For example.

```javascript
global.fetch = require('node-fetch');
var AmazonCognitoIdentity = require('amazon-cognito-identity-js');
```

exoego:
Hello, how should I proceed this PR ?

exoego:
Could someone review this, since CI now passed except optional `Label (pull_request)` ?

exoego:
@ericclemmons
I assume you have knowledge and experience wiht  `amazon-cognito-identity-js` package from commit logs.

Thanks in advance.

packages/amazon-cognito-identity-js/src/Client.js
```diff
@@ -1,4 +1,6 @@
+import fetch from 'cross-fetch';
```
ericclemmons: The `cross-fetch` library is pretty cool!  I believe it's desirable to defer the `fetch` dependencies to the user to avoid adding to bundle size.

So, would it make more sense for the README to say:

```js
// Using ES6 modules
import 'cross-fetch/polyfill';

// Using CommonJS modules
require('cross-fetch/polyfill');
```

Currently, it has:

```js
global.fetch = require('node-fetch');`
```

My hope is that this change to `Client.js` won't be necessary and we won't have a direct dependency on `cross-fetch` (since the majority of this library's usage is within a browser environment with overwhelming `fetch` support).

packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -926,7 +929,7 @@ export default class CognitoUser {
 									.getAccessToken()
 									.getJwtToken(),
 								DeviceSecretVerifierConfig: deviceSecretVerifierConfig,
-								DeviceName: navigator.userAgent,
+								DeviceName: userAgent,
```
ericclemmons: Was this creating an error in Node, or was it `undefined` and you'd prefer to have it send `nodejs`?

We'll need to do some investigation to ensure that sending `nodejs` where it wasn't before doesn't trickle into core metrics or underlying behavior. 🤞

exoego: Yes, this was creating an error so prohibit users using some functions in amazon-cognito-identity-js.
It is because `navigator` object is `undefined` in node.js land.

packages/amazon-cognito-identity-js/src/Client.js
exoego: OK, I reverted adding `cross-fetch` in https://github.com/aws-amplify/amplify-js/pull/3902/commits/f9cdbab16c2b166997789efaebd7d4579b7b1128 to remove it from direct dependency.
Then updated README in https://github.com/aws-amplify/amplify-js/pull/3902/commits/44e82f900d44a7398677a117ea32df78ee50a7d7 to metnion `cross-fetch` as polyfil for Node.js and old browsers.

packages/amazon-cognito-identity-js/src/CognitoUser.js
ericclemmons: Thanks for the info! This should be pretty safe then.

I'mg going to double-check with the Cognito team, but I'm sure this is fine.

Thanks again @exoego!

Jaystified:
Any updates on this?

packages/amazon-cognito-identity-js/src/CognitoUser.js
ericclemmons: @Amplifiyer With your SSR work, this change may be pertinent to your efforts.  @exoego's changes solve a `ReferenceError` that was happening within Node.

Amplifiyer: The changes look good to me as long as Cognito team is ok with the string 'nodejs'

ericclemmons: Confirmed with Yuchen Qian of the Cognito team that this is safe to merge. 👍

PR #3910, opened 2019-08-24
Remove json-loader in amazon-cognito-identity-js 

corollari:
According to [webpack's documentation](https://webpack.js.org/loaders/json-loader/), the json-loader is not longer needed when using the latest version (which is the version that will be installed when the instructions on the README are followed):
```
⚠️ Since webpack >= v2.0.0, importing of JSON files will work by default. You might still want to use this if you use a custom file extension. See the v1.0.0 -> v2.0.0 Migration Guide for more information
```

Given that this is no longer needed and it leads to errors with other packages (in my case, using this json loader lead to build errors due to the inclusion of the package `cryptr`) it's better to remove it. This PR does that.


corollari:
@ericclemmons hey, saw that you added a label yesterday and fixed the merge conflict (it was only a formatting changed caused by the application of prettier in an earlier commit).

ericclemmons:
Thanks! I had  to add `webpack-cli`, so moving your code to https://github.com/aws-amplify/amplify-js/pull/5749.

corollari:
Glad to see it was useful :+1:

PR #3945, opened 2019-08-30
Tooltip component and tests

bluezebragames:
- Tooltip implementation
- unit tests and a screenshot test
- Storybook playground

After first round of comments:
- added `styleBranch()` to helpers file instead of using `cx()` directly
- now with e2e tests!!


packages/amplify-ui-components/src/common/testing.ts
```diff
@@ -55,4 +55,8 @@ export const knobs = {
     formSectionSubmitKnob (submitPlaceholder: string) {
         return textKnob('Submit button text', submitPlaceholder);
     },
+
+    tooltipAutoShowKnob (autoShowDefault: boolean) {
```
sammartinez: I like that we are adding these in order to use in all, but I want to think about if we can consolidate these into a few methods instead of one broken out for each since they are using a `booleanKnob` or `textKnob`. Not a blocker for this PR just want to place it in your ear

packages/amplify-ui-components/src/components/amplify-tooltip/amplify-tooltip.e2e.ts
```diff
@@ -0,0 +1,41 @@
...
+
+describe('amplify-tooltip e2e:', () => {
+  let page: E2EPage;
+
+  beforeEach(async () => {
+    page = await newE2EPage();
+  });
+
+  it('renders a tooltip correctly', async () => {
+    await page.setContent('<div style="margin-bottom:60px"></div> <amplify-tooltip text="Tooltip">FOO</amplify-tooltip>');
```
sammartinez: why is the `margin-bottom` needed?

packages/amplify-ui-components/src/components/amplify-tooltip/amplify-tooltip.e2e.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  it(`renders an always-visible tooltip if 'shouldAutoShow' is true`, async () => {
+    await page.setContent('<amplify-tooltip text="Tooltip" should-auto-show="true">FOO</amplify-tooltip>');
+
+    const tooltip = await page.find('amplify-tooltip');
+    console.log((await tooltip.getComputedStyle(":after")).opacity);
+    expect(await tooltip.isVisible()).toBe(true);
+
+    // and then hover and do it again
```
sammartinez: Do we need this comment?

packages/amplify-ui-components/src/components/amplify-tooltip/amplify-tooltip.e2e.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  it(`renders a tooltip that is only visible on hover if 'shouldAutoShow' is false`, async () => {
+    await page.setContent('<amplify-tooltip text="Tooltip" should-auto-show="false">FOO</amplify-tooltip>');
+
+    const tooltip = await page.find('amplify-tooltip');
+
+    await page.mouse.move(500, 500);
+    console.log(tooltip.classList);
+
+    console.log((await tooltip.getComputedStyle(":after")).opacity);
```
sammartinez: Can we please remove any `console.log(..)`

packages/amplify-ui-components/src/components/amplify-tooltip/amplify-tooltip.style.ts
```diff
@@ -0,0 +1,69 @@
+import { css } from 'emotion';
+
+export const tooltip = css`
+  display: inline;
+  position: relative;
+  font-size: 12px;
+  font-family: 'Amazon Ember';
```
sammartinez: there is a `var(...)` for `--font-family` can you look to use that please? Not all browsers may have Amazon Ember so we want to make sure there is a fallbackl

packages/amplify-ui-components/src/components/amplify-tooltip/amplify-tooltip.tsx
```diff
@@ -0,0 +1,30 @@
...
+export class AmplifyTooltip {
+  /** (Required) The text in the tooltip */
+  @Prop() text: string;
+  /** (Optional) Override default styling */
+  @Prop() overrideStyle: boolean = false;
+  /** (Optional) Whether or not the tooltip should be automatically shown, i.e. not disappear when not hovered */
+  @Prop() shouldAutoShow: boolean = false;
+
+  render() {
+    const emotionTooltipClass = this.shouldAutoShow ? cx(tooltip, autoShowTooltip) : tooltip;
```
sammartinez: This may be worth wild to move to the `helpers.ts` since cx is already defined there

packages/amplify-ui-components/src/components/amplify-tooltip/amplify-tooltip.e2e.ts
bluezebragames: It was reminding me of what to put there lol

packages/amplify-ui-components/src/components/amplify-tooltip/amplify-tooltip.e2e.ts
bluezebragames: Ah yes, it's because the tooltip shows up above the DOM elements it surrounds, so if there's no margin here the button appears at the top of the screen and then the tooltip goes off the top and the screenshot is useless.  The same thing is in the Storybook file for the same reason

packages/amplify-ui-components/src/components/amplify-tooltip/amplify-tooltip.e2e.ts
bluezebragames: "It's a draft!"

bluezebragames: haha

packages/amplify-ui-components/src/components/amplify-tooltip/amplify-tooltip.tsx
bluezebragames: Yeah, that s a solid idea!  I think there's another component where I do the same sort of thing, so let me consolidate them both into `helpers`

packages/amplify-ui-components/src/common/testing.ts
bluezebragames: Yeah, I totally get where you're coming from.  After a bit of thought, I've decided that there's not much one could do as a sort of middle ground between what we have now (one knob for every single use case of knobs) and no knobs at all (just have every `.stories` file import `textKnob`s and `booleanKnobs` as needed).  This is because essentially the only thing these knobs are doing is providing a knob name specific to the use case -- if we combine these methods into smaller groups, we'll need to differentiate them by their knob name, so what we'll be left with is something like
```
combinedTextKnob (knobName: string, placeholder: string) {
    return textKnob(knobName, placeholder);
}
```
which is of course literally just a `textKnob`.

The only possibility for a middle ground I could come up with is to only have the highly shared knobs be in this testing file and have every `.stories` file just import `textKnob`, etc.  However, the only knob that would fit that first requirement is `overrideStyleKnob`, so it seems like a pretty ineffectual middle ground.  Also, if we pursue that path, quite a few components will be including both `booleanKnob` and `overrideStyleKnob`, which seems like a waste for sure...

packages/amplify-ui-components/src/components/amplify-tooltip/amplify-tooltip.e2e.ts
```diff
@@ -18,24 +18,44 @@ describe('amplify-tooltip e2e:', () => {
...
+    // Move mouse away from div with tooltip -- should still be visible
+    await page.mouse.move(500, 500);
+    await page.waitFor(1500);
+    const tooltipInnerDiv = await tooltip.find('div');
+    expect((await tooltipInnerDiv.getComputedStyle(':after')).opacity).toBe("1");
+
+    // Hover div with tooltip -- should still be visible
+    await page.hover('#toHover');
+    await page.waitFor(1500);
+    expect((await tooltipInnerDiv.getComputedStyle(':after')).opacity).toBe("1");
```
sammartinez: Can we move what is in the expect into a variable? I see you using it a couple of times so may be worth to place in a variable. 

e.g.
```
const toolTipOpacity = (await tooltipInnerDiv.getComputedStyle(':after')).opacity;

expect(toolTipOpacity).toBe("1");
```

Also believe we can make the `toBe` a `toEqual`. Usually only use `toBe` with `true` or `false` statements.

packages/amplify-ui-components/src/components/amplify-tooltip/amplify-tooltip.e2e.ts
```diff
@@ -30,12 +30,14 @@ describe('amplify-tooltip e2e:', () => {
     await page.mouse.move(500, 500);
     await page.waitFor(1500);
     const tooltipInnerDiv = await tooltip.find('div');
-    expect((await tooltipInnerDiv.getComputedStyle(':after')).opacity).toBe("1");
+    const opacityNoHover = (await tooltipInnerDiv.getComputedStyle(':after')).opacity;
```
sammartinez: There is no change correct? between the variables `opacityNoHover` and `opacityHover`? Can we use the same variable?

bluezebragames: I mean they'd better be different since we're expecting it to be a different value between when it's hovered and not...  But more to the point, yes but it couldn't be `const` -- do you think that's a worthwhile tradeoff?

bluezebragames: to clarify about that last part, it's either
```
const opacityNoHover = ...;
...
const opacityHover = ...;
```
or
```
let opacity = ...;
...
opacity = ...;
```
right?  I feel like there's pros and cons to both of them -- for the `const` one, you get a little more clarity on the purpose at the cost of having to use two variables, and vice versa...  Neither option seems dramatically better than the other though, and I'd like to hear your thoughts

sammartinez: After conversation offline, I am good with both.

PR #3988, opened 2019-09-06
Added icon unit tests and additional storybook

bluezebragames:
- Left the `snapshotOverrideTestIcon` if you want to take the time to fix the [shared nodes problem](https://stenciljs.com/docs/templating-jsx#avoid-shared-jsx-nodes) and add it into the inner `describe()` block
- Another Storybook playground to show that the shared nodes thing isn't an issue in the real world


jordanranz:
LGTM, sorry for the late merge.

Issue #3932, opened 2019-08-28
Unhandled Rejection (Error): when oauth flow return server error


hp0210:
**Describe the bug**
We are currently on the process of upgrading our aws-amplify library from 1.1.9 to 1.1.36 and aws-amplify-react from 1.0.19 to 2.3.12 . We found the following issue where we will get **Unhandled Rejection (Error)** error from aws-amplify library when our user is redirected back to the app with OAuth flow error (e.g https://localhost:3000/?error_description=Error+in+SAML+response&error=server_error)

**Additional context:**
* aws-amplify version 1.1.36
* aws-amplify-react version 2.3.12
* We use Cognito preSignUp trigger to make sure that only federated users that have been provisioned can access our application. So with any federated user that logs in via SSO flow (SAML provider via Cognito hosted UI) that doesn't have an account in our system, we will throw an exception in Cognito preSignUp trigger to prevent the user logging in the system. This worked fine with before we upgrading the library

**To Reproduce**
Steps to reproduce the behaviour:
1. Configure your aws-amplify with oauth config
2. Navigate to your app with error params (e.g https://localhost:3000/?error_description=Error+in+SAML+response&error=server_error) 
3. You will see the Unhandled Rejection (Error)

**Expected behaviour**
React app will be freezed due to Unhandled Rejection (Error)

**Our code**
```javascript
let amplifyConfig = {
  Auth: {
    identityPoolId: Config.cognito.IDENTITY_POOL_ID,
    region: Config.cognito.REGION,
    userPoolId: Config.cognito.USER_POOL_ID,
    userPoolWebClientId: Config.cognito.APP_CLIENT_ID,
    mandatorySignIn: true,
    cookieStorage: {
      domain: Config.reactApp.COOKIE_DOMAIN,
      path: '/',
      expires: 1,
      secure: Config.reactApp.SECURE_PROTOCOL
    },
    oauth: {
      domain : Config.cognito.USERPOOL_HOSTED_DOMAIN+'.auth.'+Config.cognito.REGION+'.amazoncognito.com',
      scope : [ 'openid', 'email' ],
      redirectSignIn : Config.cognito.USERPOOL_ADFS_CALLBACK_URL,
      redirectSignOut : Config.cognito.USERPOOL_ADFS_LOGOUT_URL,
      responseType: 'code',
    }
  },
  Analytics: {
    disabled: true,
    autoSessionRecord: false,
  }
};

Amplify.configure(amplifyConfig)
```
[...]

PR #3993, opened 2019-09-09
fix(#3932)

haverchuck:
*Issue #, if available:*
#3932 

*Description of changes:*
Adds try/catch block to handleAuthResponse.

### DIFF
--- a/packages/auth/src/OAuth/OAuth.ts
+++ b/packages/auth/src/OAuth/OAuth.ts
@@ -183,29 +183,33 @@ export default class OAuth {
   }
 
   public async handleAuthResponse(currentUrl?: string) {
-    const urlParams = currentUrl ? {
-      ...(parse(currentUrl).hash || '#').substr(1)
-        .split('&')
-        .map(entry => entry.split('='))
-        .reduce((acc, [k, v]) => (acc[k] = v, acc), {}),
-      ...(parse(currentUrl).query || '')
-        .split('&')
-        .map(entry => entry.split('='))
-        .reduce((acc, [k, v]) => (acc[k] = v, acc), {})
-    } as any : {};
-    const { error, error_description } = urlParams;
-
-    if (error) {
-      throw new Error(error_description);
-    }
-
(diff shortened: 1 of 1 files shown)


packages/auth/src/OAuth/OAuth.ts
```diff
@@ -183,29 +183,33 @@ export default class OAuth {
...
+          .map(entry => entry.split('='))
+          .reduce((acc, [k, v]) => (acc[k] = v, acc), {}),
+        ...(parse(currentUrl).query || '')
+          .split('&')
+          .map(entry => entry.split('='))
+          .reduce((acc, [k, v]) => (acc[k] = v, acc), {})
+      } as any : {};
+      const { error, error_description } = urlParams;
+  
+      if (error) {
```
sammartinez: Is this needed now since we will have the catch for this?

haverchuck: @sammartinez -  I believe so?  The 'error' in the if test appears to be a destructured value from the a JSON response, which would not necessarily result in an error in the stack.  It's just a value.  So I would think we still need to manually throw the error.  In any case, I don't think there is harm in leaving it.

packages/auth/src/OAuth/OAuth.ts
```diff
@@ -183,29 +183,33 @@ export default class OAuth {
...
+        throw new Error(error_description);
```
ericclemmons: Based on #3932, I couldn't tell if this line is what's causing the development-mode error vs. Line 207/209.

Is there a risk in wrapping _too much_ logic (e.g. `urlParams` creation, throwing with `error_description`, `await this._handleCodeFlow`, `await this._handleImplicitFlow`), and losing visibility into other bugs as a result?

navilan:
@sammartinez @haverchuck - An unfortunate side effect of this PR is that errors are now swallowed. We use Hub.listen to listen on errors and handle them, this change has prevented that.  Can we revert this change please or remove lines 199 - 201?

Issue #3983, opened 2019-09-06
Error thrown if only using Identity Pool


justinwaite:
**Describe the bug**
When using Amplify Auth with only an identity pool ID (no user pool), the following error is thrown:

**To Reproduce**

**Expected behavior**
It is expected that the user pool is optional and that simply an identity pool Id would suffice.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser chrome
 - Version 76

**Additional context**
Looking at the source, it seems that this line needs to be updated:
https://github.com/aws-amplify/amplify-js/blob/632064e165656bafe224c67d99d91a7a069c632b/packages/auth/src/Auth.ts#L1700

According to the docs, only an identity pool ID is required. ~Therefore, the line should probably read:~
```javascript
if (!config.userPoolId && !config.identityPoolId) {
``` 
~(updated from `||` to `&&`).~

Edit: Looking into the source further this is not the solution.

jordanranz:
@justinwaite, Can you post a link to the documentation you are referring to. This does indeed seem odd.

justinwaite:
Sure thing, it's all over the docs, but here's one example in particular:
https://aws-amplify.github.io/docs/js/authentication#manual-setup

justinwaite:
To remedy right now I simply put in a fake `userPoolId` and `userPoolWebClientId` so that the error isn't thrown.

justinwaite:
The error seems to be gone now, and I'm not sure why... I wasn't calling any auth methods that would have thrown this error. It could have been a change in the way I did some stuff with the Storage API (maybe storage API was trying to sign in automatically?).

justinwaite:
The error isn't thrown until I use the Storage API to try and upload a file. Maybe the Storage API is trying to call one of the authentication methods that would throw this error?

elorzafe:
Hi @justinwaite I reproduced the problem, we are going to send a fix for this soon. Thanks for pointing out!

justinwaite:
Thanks!

justinwaite:
Looks like that did it. Thank you for the quick update!

PR #4000, opened 2019-09-11
Prevent console error when using guest creds with Storage

iartemiev:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-js/issues/3983

*Description of changes:*
Return empty `Promise.reject()` to prevent error from showing up in the console

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -1075,7 +1075,8 @@ export default class AuthClass {
     public currentSession(): Promise<CognitoUserSession> {
         const that = this;
         logger.debug('Getting current session');
-        if (!this.userPool) { return this.rejectNoUserPool(); }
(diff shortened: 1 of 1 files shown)


packages/auth/src/Auth.ts
```diff
@@ -1075,7 +1075,8 @@ export default class AuthClass {
     public currentSession(): Promise<CognitoUserSession> {
         const that = this;
         logger.debug('Getting current session');
-        if (!this.userPool) { return this.rejectNoUserPool(); }
+        // Purposely not calling the reject method here because we don't need a console error
+        if (!this.userPool) { return Promise.reject(); }
```
Woodz, line 1079: This results in silent failure if the user pool was accidentally misconfigured. I don't think that this is expected behaviour - can you clarify why we don't want a console error when the user pool is not configured @iartemiev ?

Issue #4003, opened 2019-09-11
Federated Sign In (Google) with Cognito is not working on Android and Expo Standalone


oriharel:
**Describe the bug**
I'm using withOAuth() HOC. User is directed to the Google (code) flow, comes back to the app. Linking url event is triggered (with code parameter in the url) thus causing a call to Amplify token endpoint (with code_verifier). However, multiple additional Linking url events are fired, causing a throttle of token endpoint calls to trigger (all of them without code_verifier) that ends up with invalid_request and invalid_grant and eventually crashes the app. 
This happens only on Android (iOS works fine) and in Expo standalone (for some reason, when running the app in the Expo Client, it works fine because the url is stripped from the code this the token endpoint is not called).

**To Reproduce**
User Expo standalone Android app with Amplify withOAuth HOC to call google.

**Expected behavior**
Avoid invalid calls to the token endpoint.

**Smartphone (please complete the following information):**
 - Device: Any Android device
 - OS: Any Android version
 - Expo Standalone

Here is the relevant config of Auth.configure:

```
const urlOpener = async (url, redirectUrl) => {
	const { type, url: newUrl } = await WebBrowser.openAuthSessionAsync(
		url,
		redirectUrl
	);

	if (Platform.OS === 'ios') {
		if (type === 'success') {
			await WebBrowser.dismissBrowser();
			return Linking.openURL(newUrl);
		}
	}
};
export const oauth = {
	domain: 'some.domain.com',
	scope: [
		'phone',
		'email',
		'openid',
		'profile',
		'aws.cognito.signin.user.admin'
	],

	redirectSignIn: Linking.makeUrl('/signIn'),
	redirectSignOut: Linking.makeUrl('/signOut'),
	responseType: 'code',

	urlOpener: urlOpener
};
```

elorzafe:
@oriharel thanks for your feedback. Can you share your package.json and expo-cli version?

oriharel:
I've opened a PR #4005

oriharel:
So this issue is actually not an Amplify issue, but an expo one. For some reason, expo standalone apps fires multiple events on deep linking on Android. There is an issue on this matter [here](https://github.com/expo/expo/issues/2128).
Perhaps @brentvatne can take a look. However, Amplify can provide a hook to workaround this as suggested in the PR (#4005).

Thanks.

schellack:
[...]

PR #4005, opened 2019-09-11
Adding urlListener to the configuration object.

oriharel:
This can come handy for customising 'url' event handling.

*Issue #, if available:*
Issue #4003 

*Description of changes:*
Recently the withOAuth API was added to support Expo users. that was awesome, however I think this is also needed. and also a way to handle the issue (for managed workflow of Expo)

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -94,6 +94,7 @@ export default class AuthClass {
     private _oAuthHandler: OAuth;
     private _storage;
     private _storageSync;
+    private _urlListener;
 
     /**
      * Initialize Auth with AWS configurations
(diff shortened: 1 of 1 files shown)


jdcas89:
This would be very helpful, as now amplify clashes with other Oauth services that also redirect to your app with the query param `code`.

evertson90:
Any progress on this issue?

Ashish5591:
Hi, thanks for opening the PR. We are looking into it.
As you mentioned here: https://github.com/aws-amplify/amplify-js/issues/4003#issuecomment-530685218 its not exactly an Amplify issue, but rather an expo one. We are looking into the issue  and considering the benefits of adding this workaround outside the specific issue on Android Expo standalone apps.

loganwedwards:
I have reason to believe this may be an issue on iOS as well. I am using Expo 33 with amplify (partly due to the promoted integration) and am seeing multiple requests to validate the code. I have mixed success with the current setup due to, what appears to be a race condition on which callback from the oauth endpoint executes first.

Issue #3786, opened 2019-08-01
Session Stops Not Always Logged


AbiSri:
**Describe the bug**
Pinpoint does not always send the session stop event when the browser is closed.

**To Reproduce**
1. Open console in Safari. Right click on the app at the bottom of the screen and click quit.
Note: Have not tried on other browsers.

iartemiev:
Hey @AbiSri we published a fix for this that's currently in the `1.1.41-unstable.2
` version of aws-amplify and will be part of the next release! Please let me know if this has resolved this issue.

jordanranz:
Closing due to inactivity and a fix has been released.

Issue #3042, opened 2019-04-09
Add option to specify method to send AWS Pinpoint events


Dilatorily:
**Is your feature request related to a problem? Please describe.**
AWS Pinpoint users are currently using AWS Amplify to send analytics events through XHR POST requests. Some useful analytics metrics are the time spent on page, which would require an event to be sent when the browser is closed. However, most browsers will not send XHR POST requests when a page is closed. This makes AWS Amplify JS useless for those kind of analytics events.

**Describe the solution you'd like**
AWS Amplify could provide the option to configure the method used to send the analytics metrics through a callback function, and provide a default callback that uses the current POST XHR method. It would allow customers to customize how their AWS Pinpoint events are being sent.

**Describe alternatives you've considered**
Currently, customers has to either overload the XHR method, or to proxy every request to their own API.

**Additional context**
N/A

roborourke:
Just about to open a similar request so +1.

In browsers that support it `Navigator.sendBeacon()` should be used to send events to Pinpoint to avoid the drawbacks of XHR.

@Dilatorily if it helps in the meantime I've implemented my own custom pageView event using `Analytics.record()` on a `beforeunload` event listener. It's annoying that there's a small delay in navigation but it's the only way I've managed to collect time on page and scroll depth metrics so far.

iartemiev:
Hey @Dilatorily and @roborourke , we have published a fix for this that's currently in the `1.1.41-unstable.2` version of aws-amplify and will be part of the next release! Please let me know if this resolves the issue you were experiencing.

Dilatorily:
@iartemiev I think #4012 only solves `_session.stop` events.
I'm also looking to use navigator.sendBeacon for other custom events when the page/browser is closed.

iartemiev:
@Dilatorily you're correct, we're currently only using it for _session.stop. I misread your feature request. I will look into using it for custom events as well.

PR #4012, opened 2019-09-13
fix: Pinpoint analytics stopSession event

iartemiev:
utilize `navigator.sendBeacon()` to reliably send _session.stop events on `beforunload`

closes issues #3786 and #3042

### DIFF
--- a/packages/analytics/package.json
+++ b/packages/analytics/package.json
@@ -10,8 +10,10 @@
   },
   "scripts": {
     "test": "tslint 'src/**/*.ts' && jest --coverage",
+    "test:watch": "tslint 'src/**/*.ts' && jest --watch",
     "build-with-test": "npm run clean && npm test && tsc && webpack",
     "build": "npm run clean && tsc && webpack",
+    "build:watch": "npm run clean && tsc --watch",
     "clean": "rimraf lib lib-esm dist",
     "format": "echo \"Not implemented\"",
     "lint": "tslint 'src/**/*.ts'"
@@ -79,6 +81,9 @@
     },
     "coveragePathIgnorePatterns": [
       "/node_modules/"
+    ],
+    "setupFiles": [
+      "<rootDir>/src/setupTests.ts"
     ]
   }
 }
--- a/packages/analytics/src/Providers/AWSPinpointProvider.ts
+++ b/packages/analytics/src/Providers/AWSPinpointProvider.ts
@@ -27,7 +27,6 @@ import Cache from '@aws-amplify/cache';
 
 import { AnalyticsProvider, PromiseHandlers } from '../types';
 import { v1 as uuid } from 'uuid';
-import { RequestParams } from './AmazonPersonalizeHelper/DataType';
 
 const AMPLIFY_SYMBOL = ((typeof Symbol !== 'undefined' && typeof Symbol.for === 'function') ?
     Symbol.for('amplify_default') : '@@amplify_default') as Symbol;
--- a/packages/analytics/src/Providers/AWSPinpointProvider.ts
+++ b/packages/analytics/src/Providers/AWSPinpointProvider.ts
@@ -16,7 +16,6 @@ import {
     ClientDevice,
     Platform,
     Credentials,
-    Constants,
     Signer,
     JS,
     Hub
@@ -28,8 +27,6 @@ import Cache from '@aws-amplify/cache';
 
 import { AnalyticsProvider, PromiseHandlers } from '../types';
 import { v1 as uuid } from 'uuid';
-import { SSL_OP_TLS_BLOCK_PADDING_BUG } from 'constants';
-import { resolve } from 'url';
 
 const AMPLIFY_SYMBOL = ((typeof Symbol !== 'undefined' && typeof Symbol.for === 'function') ?
     Symbol.for('amplify_default') : '@@amplify_default') as Symbol;
--- a/packages/analytics/src/Providers/AWSPinpointProvider.ts
+++ b/packages/analytics/src/Providers/AWSPinpointProvider.ts
@@ -331,7 +331,7 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
             method,
         };
 
-        const serviceInfo = { region, service: SERVICE_NAME };
+        const serviceInfo = { region, service: MOBILE_SERVICE_NAME };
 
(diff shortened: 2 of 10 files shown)


packages/analytics/package.json
```diff
@@ -10,8 +10,10 @@
   },
   "scripts": {
     "test": "tslint 'src/**/*.ts' && jest --coverage",
+    "test:watch": "tslint 'src/**/*.ts' && jest --watch",
```
sammartinez: thank you :)

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -37,6 +38,7 @@ const dispatchAnalyticsEvent = (event, data) => {
 const logger = new Logger('AWSPinpointProvider');
 const RETRYABLE_CODES = [429, 500];
 const ACCEPTED_CODES = [202];
+const SERVICE_NAME = "mobiletargeting";
```
sammartinez: This seems to generic for the constant, can we update this to something like `MOBILE_TARGETING_SERVICE_NAME`?

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -306,6 +309,40 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
         });
     }
 
+    private _pinpointSendStopSession(params, handlers): Promise<string> {
+        const eventParams = this._generateBatchItemContext(params);
+        
+        const {region} = this._config;
```
sammartinez: NIT: can we add spaces around the destructs?

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -306,6 +309,40 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
+        const serviceInfo = { region, service: SERVICE_NAME };
+
+        const requestUrl: string = Signer.signUrlWithBody(reqOptions, accessInfo, serviceInfo, null, 'POST');
+
+        const success: boolean = navigator.sendBeacon(requestUrl, body);
+
+        if (success) {
+            return handlers.resolve('sendBeacon success');
+        }
+        return handlers.reject("sendBeacon failure");
```
sammartinez: NIT: can we adjust to single quotes or double quotes? I just see the success has single and the reject double. Just would like to have consistency

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -445,6 +482,7 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
         logger.debug('init clients with credentials', credentials);
         this.mobileAnalytics = new MobileAnalytics({ credentials, region });
         this.pinpointClient = new Pinpoint({ region, credentials });
+
```
sammartinez: ### S P A C E S

packages/analytics/src/setupTests.ts
```diff
@@ -0,0 +1,4 @@
+const anyGlobal = (global as any);
```
sammartinez: nice!

packages/core/src/Signer.ts
```diff
@@ -395,6 +395,85 @@ const signUrl = function(urlToSign: String, accessInfo: any, serviceInfo?: any,
     return result;
 };
 
+const signUrlWithBody = 
+  (request: any, accessInfo: any, serviceInfo?: any, expiration?: Number, method: String = 'POST') => {
+      
+    const now = new Date().toISOString().replace(/[:\-]|\.\d{3}/g, '');
+    const today = now.substr(0, 8);
+    // Intentionally discarding search
+    const {search, ...parsedUrl} = url.parse(request.url, true, true);
```
sammartinez: NIT: can we add spaces around the destructs

packages/core/src/Signer.ts
```diff
@@ -395,6 +395,85 @@ const signUrl = function(urlToSign: String, accessInfo: any, serviceInfo?: any,
...
+        service
+    );
+
+    // IoT service does not allow the session token in the canonical request
+    // https://docs.aws.amazon.com/general/latest/gr/sigv4-add-signature-to-request.html
+    const sessionTokenRequired = accessInfo.session_token && service !== IOT_SERVICE_NAME;
+    const queryParams = {
+        'X-Amz-Algorithm': DEFAULT_ALGORITHM,
+        'X-Amz-Credential': [accessInfo.access_key, credentialScope].join('/'),
+        'X-Amz-Date': now.substr(0, 16),
```
sammartinez: shouldnt this be `today` ?

iartemiev: I copy-pasted this from the function above `signUrl`, so I'm assuming that's how it should be

iartemiev: Actually, now that I've looked, `today` is `substr(0, 8)` so that wouldn't work here

packages/core/src/Signer.ts
```diff
@@ -434,7 +367,7 @@ const signUrlWithBody =
             },
         }),
         headers: signedHeaders,
-        data: request.body
+        data: body
```
elorzafe: is this ok to be part of the canonical_request even empty?

iartemiev: Yep, it's fine. `canonical_request` passes that value to `hash` which handles `undefined` safely.

packages/core/src/Signer.ts
```diff
@@ -232,176 +232,182 @@ const get_authorization_header = function(algorithm, access_key, scope, signed_h
...
+        request.headers['Authorization'] = authorization_header;
+    
+        return request;
+    }
+    
+    static signUrl(urlToSign: string, accessInfo: any, serviceInfo?: any, expiration?: number): string;
+    static signUrl(request: any, accessInfo: any, serviceInfo?: any, expiration?: number): string;
+    static signUrl(urlOrRequest: string | any, accessInfo: any, serviceInfo?: any, expiration?: number): string {
+
+        const urlToSign: string = typeof urlOrRequest === 'string' ? urlOrRequest : urlOrRequest.url;
```
elorzafe: I would check if `urlOrRequest` is not `undefined` before getting `url`, the same goes the L337, L338. Maybe setting default to and empty object is enough. What do you think?

iartemiev: Good catch - fixed

PR #4038, opened 2019-09-18
feat: Added Prettier formatting

iartemiev:
Formatted all `.ts`, `.js`, `.jsx`, and `.json` files (except some config json files)
Added `.prettierignore` and `.prettierrc` with default settings
Adjusted tslint configs to prevent conflicts with Prettier

No issues with tests or build after formatting


.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+    "trailingComma": "es5",
+    "tabWidth": 4,
```
ericclemmons: In yesterday's conversation, zero-config sounded like the way to go.

With fewer options, would there be fewer files affected?

iartemiev: https://prettier.io/docs/en/configuration.html

That is the default config, only with semicolons enabled because I figured that would be too wild of a change to leave them off :)

iartemiev: The main reason for including a prettierrc is to override whatever one might have defined in their vscode settings

ericclemmons: That's interesting for sure.  Generally I've used `.pretterrc` with `{}`, which uses only the defaults.

Have you given it a shot with `{}`? It may result in less changes (primarily due to `tabWidth` defaulting to `2`).

Even `create-react-app` has _some_ configuration in their codebase, likely to avoid controversy as well (e.g. `semi`, `singleQuote`):
> https://github.com/facebook/create-react-app/blob/master/.prettierrc

.vscode/extensions.json
```diff
@@ -0,0 +1,9 @@
+{
+    // See http://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+    // Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+    // List of extensions which should be recommended for users of this workspace.
+    "recommendations": ["esbenp.prettier-vscode"],
```
ericclemmons: For times when you _don't_ want to apply prettier (e.g. API response or something), how are you handling these situations?

Previously, I would have to change my preferences, save, then turn them back on. 

Currently, this little toggle in the toolbar has been great: https://marketplace.visualstudio.com/items?itemName=tombonnike.vscode-status-bar-format-toggle?

Screen Shot 2019-09-18 at 12 55 29 PM
Screen Shot 2019-09-18 at 12 55 41 PM

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+  "editor.formatOnSave": true,
+  "editor.tabSize": 2
```
sammartinez: Why 2? I believe we should be doing 4. I do want weigh in from people on this

.vscode/extensions.json
iartemiev: I like that. Added. 

Previously, I would either do what you described above or save and then undo changes with `git checkout -- [file]` or add the file or dir to prettierignore

.vscode/settings.json
iartemiev: 2 is the Prettier default. I'm open to going with 4, but what's your reasoning behind it?

pravgupt: +1. It does come down to personal preference and we can go with the one that most people prefer?

iartemiev: I agree. 

I set the editor tab size to 2 just to keep it in sync with what Prettier formats it to but we can override Prettier to 4 if that's what most people would rather have.

iartemiev: Personally, I think 2 is slightly better because then the total width is less affected by nesting, i.e. you can see more horizontally. But I could go either way

pravgupt:
What is the guidance for PRs/branches that are work in progress? Should they only take the prettier config and run the prettier before doing a merge from master/release?

iartemiev:
> What is the guidance for PRs/branches that are work in progress? Should they only take the prettier config and run the prettier before doing a merge from master/release?

@pravgupt yes, I think that would be easiest from a merge conflict standpoint. They don't even need to use the config though, I'll provide a single CLI command that'll format the codebase.

.vscode/settings.json
sammartinez: From @manueliglesias point, should use the default of prettier to avoid this argument. I am fine with the default of 2 for prettier. Would prefer 4 but I dont have a clear argument of why minus I believe reading is easier for myself

sammartinez:
> > What is the guidance for PRs/branches that are work in progress? Should they only take the prettier config and run the prettier before doing a merge from master/release?
> 
> @pravgupt yes, I think that would be easiest from a merge conflict standpoint. They don't even need to use the config though, I'll provide a single CLI command that'll format the codebase.

I want to consider the bigger PRs like the Modularization and the TS React one. To resolve that may be a lot of conflicts

iartemiev:
@sammartinez There shouldn't be if the PR author runs prettier locally and commits/pushes up the result

.vscode/settings.json
jordanranz: I prefer 2

iartemiev:
> LGTM

@manueliglesias I saw that but I'm not sure how to resolve that last alert without accidentally breaking something. It's not an issue that Prettier introduced, I think it just hadn't been caught by LGTM until the formatting changed in that file and it saw the issue in the commit

iartemiev:
For all open PRs, the authors should run the following command from the root of the project before pulling latest `master` in order to avoid conflicts: 

```
yarn global add prettier && prettier --single-quote --trailing-comma es5 --write "packages/**/*.{ts,js,jsx,md}"
```

(This will also format the files in dist but that's inconsequential)

manueliglesias:
> For all open PRs

It would be nice if we could do it for them if the author left the "Allow edits from maintainers" checkbox checked when submitting the PR

iartemiev:
@manueliglesias good point. I'm happy to do that after this gets merged

.vscode/settings.json
undefobj: There actually is an accessibility issue where it's not just a preference: https://www.reddit.com/r/javascript/comments/c8drjo/nobody_talks_about_the_real_reason_to_use_tabs/

iartemiev: @undefobj That's fair. I'll switch it to 4

iartemiev: @undefobj scratch that. I misunderstood. Will just add the setting `useTabs`

Issue #4049, opened 2019-09-20
Application crashes when signing up without touching the form fields


ricardoribas:
**Describe the bug**
If a user trieds to sign up a form **containing only mandatory fields (standard and custom ones)** without touching any of the form fields, the application will crash. The application will not crash anymore after filling the fields. **However**, the validation will pass even if i put custom attributes that are marked as mandatory.

Another thing that i've noticed and i don't know if make sense or not to you guys is the fact that you are passing `isValid` method (inside signUp) as a reference to a button (named as `AmplifyButton`) that is expecting a boolean prop.

**To Reproduce**
#### Case one
1. Add signUpConfig to have three standard fields (username as phone number, name and password), but also two custom fields (for instance age and city)-
2. Go to 'sign up' screen
3. Click on 'sign up'

#### Case two
1. Add signUpConfig to have three standard fields (username as phone number, name and password), but also two custom fields (for instance age and city)-
2. Go to 'sign up' screen
3. Fill the name, username and password fields
4. Click on 'sign up'
5. The form will pass, but there are two extra custom fields (custom:age and custom:city)

**Expected behavior**
The application should not crash on standard fields and should not pass if the custom fields are not filled.

***You can turn on the debug mode to provide more info for us by setting window.LOG_LEVEL = 'DEBUG'; in your app.***

ricardoribas:
The fix for this issue is to add two changes in the `SignUp.js`:
1 - in the isValid method 
2 - make the **AmplifyButton** to execute the method and not passing it as reference.
How can i create a PR for this issue? I followed all the steps to contribute to this project [here](https://github.com/aws-amplify/amplify-js/blob/master/CONTRIBUTING.md) but i am not allowed to push my code (😄). Maybe i am doing something wrong. The commit hooks are executing successfully.

ricardoribas:
@jordanranz i wonder if there is any progress on this topic. 🤓

ericclemmons:
@ricardoribas Can you share a snippet of your component code?  I'm unsure if you're using `aws-amplify-react`'s `<Authenticator>`, `withAuthenticator`, or another library.

sammartinez:
@ricardoribas As Erics ask above, can you please share your code snippets of your implementations? This will help us further narrow down this issue for you. Thanks ahead of time!

PR #4050, opened 2019-09-20
Fix/feat aws amplify react native required fields

ricardoribas:
*Issue #4049*

*Description of changes:*
1 - Fixed the access to the signUpConfig form fields properties inside the `isValid` method on the SignUp class
2 - Passed the `isValid` method return value to the **AmplifyButton** submit button.


packages/aws-amplify-react-native/src/Auth/SignUp.js
```diff
@@ -263,7 +267,7 @@ export default class SignUp extends AuthPiece {
 							text={I18n.get('Sign Up').toUpperCase()}
 							theme={theme}
 							onPress={this.signUp}
-							disabled={!this.isValid}
+							disabled={!this.isValid()}
```
ricardoribas: Before proceeding with the unit tests, i would like to know your opinion about it 🤓

ricardoribas:
@jordanranz here's the PR about the issue that i reported last week. Let me know if there is anything that i need to change.

ricardoribas:
I wonder if there is any update on this. We are hoping to remove our workaround from the codebase ❤️

ricardoribas:
Going to close this one as the code that is this PR's code is the same as the one in master. Problem is solved already

PR #4055, opened 2019-09-20
Adding data-test attributes to ChatBot react component

elorzafe:
_Issue #, if available:_

_Description of changes:_
Adding `data-test` attributes to ChatBot on react component


packages/aws-amplify-react/src/Amplify-UI/data-test-attributes.tsx
```diff
@@ -143,6 +143,13 @@ export const genericAttrs = {
 	dialCodeSelect: 'dial-code-select',
 };
 
+export const chatBot = {
```
sammartinez: General question: Is Chat Bot only available in React? We may need to add these to Angular and Vue data test attrs files as well. Even if its a placeholder

packages/aws-amplify-react/src/Interactions/ChatBot.tsx
```diff
@@ -441,7 +455,7 @@ export class ChatBot extends Component<IChatBotProps, IChatBotState> {
 						micButtonDisabled={this.state.micButtonDisabled}
 						handleMicButton={this.micButtonHandler}
 						currentVoiceState={this.state.currentVoiceState}
-					></ChatBotInputs>
+					/>
```
sammartinez: Cleaner code!

packages/aws-amplify-react/src/Amplify-UI/data-test-attributes.tsx
jordanranz: Yea, can we add the Angular and Vue data test attributes as well.

elorzafe: Cool, I can add those as well

Issue #3365, opened 2019-05-30
RFC: Amplify Library Modularization and Bundle Size Improvement


powerful23:
This issue is a Request For Comments (RFC). It is intended to elicit community feedback regarding a proposed change to the library. Please feel free to post comments or questions here.

## Situation

The bundle size of amplify-js has always been a concern to the customers.  At the very beginning we put all the categories in a single package which leads to a huge bundle size. Then we decided to split the whole package into several ones based on the category so we can import each category on need.

## The Problem

Although the bundle size can be largely reduced by using scoped packages instead of directly using `aws-amplify`, there are still many things that we can do to make it better:

- The way of using scoped packages is tricky and not intuitive.
- When using `aws-amplify-react`, all the Amplify library will be imported and bundled regardless of the actual usage.
- Using default export is not the best practice.
- Graphql will be imported when using API module which is unnecessary if only calling REST APIs.
- Version conflicting when using scoped packages which will lead to multiple versions of same package under `node_modules`
- The `aws-sdk` takes a lot of the space in the bundle size but we are only using part of it.

## Proposed Solutions

- Generate ES6 modules so that bundlers like Webpack and Rollup can do the tree shaking when importing something from `aws-amplify`. So instead of using scoped packages like `import Auth from '@aws-amplify/auth'` you can now just use `import { Auth } from 'aws-amplify'` and the bundler will do the rest work for you.
- Update the library code and the documentation to stop using default import/export but also keep backward compatibility.
- Split the API module into two, one will be only for REST APIs, the other one will be for GraphQL APIs. For example, `import { RestAPI, GraphQLAPI } from 'aws-amplify'`
- Use peer dependencies and loose version number to avoid version conflicting.
- Build a http client to replace `aws-sdk`. The client should be able to generate the request context, sign the request with AWS credentials, validate the input parameters, configure the retry policy etc.

## Related Issues

#122 
#2781 
#3097 
#3096 
#2309 
#2199
#1831 
#701 
#2894 
#2058
#488

## Reference

- [Tree shaking](https://webpack.js.org/guides/tree-shaking/)
- [Static module structure](http://exploringjs.com/es6/ch_modules.html#static-module-structure)
- [Why we have banned default exports in Javascript and you should do the same
](https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad)
[...]

PR #4077, opened 2019-09-24
[Delivers #168673137] Migrate core category to aws sdk V3

Amplifiyer:
_Issue #, if available:_ https://github.com/aws-amplify/amplify-js/issues/3365

_Description of changes:_ Migrate core category to use V3 javascript SDK ofr getting credentials from cognito service and signer

[Delivers #168673137]


packages/core/src/Credentials.ts
```diff
@@ -316,69 +301,50 @@ export class CredentialsClass {
...
-								token,
-								user,
-								expires_at,
-								identity_id,
-							})
-						);
-					} catch (e) {
-						logger.debug('Failed to put federated info into auth storage', e);
+			credentials
+				.catch(err => {
```
iartemiev: Just making sure I understand: is the reason why the `catch` precedes the `then` here because you only want the `catch` to get called if something throws an error in `credentials` specifically, as opposed to also catching errors thrown inside the `then`?

Since there's nothing in the `then` that could throw an error, it doesn't matter either way. But I'm just curious if the ordering of `catch` and `then` chaining here is deliberate.

pravgupt: This wasn't intentional. I guess I screwed up during some merge. Flipped them back.

packages/auth/src/Auth.ts
```diff
@@ -106,12 +105,6 @@ export class AuthClass {
 
 		this.currentUserCredentials = this.currentUserCredentials.bind(this);
 
-		if (AWS.config) {
-			AWS.config.update({ customUserAgent: Constants.userAgent });
```
sammartinez: Does v3 offer userAgents? I believe we still want to track this. Is that correct @elorzafe

packages/core/package.json
```diff
@@ -1,88 +1,92 @@
...
+		"access": "public"
+	},
+	"scripts": {
+		"test": "tslint 'src/**/*.ts' && jest --coverage",
+		"build-with-test": "npm test && npm run build",
+		"build:cjs": "node ./build es5 && webpack && webpack --config ./webpack.config.dev.js",
+		"build:esm": "node ./build es6",
+		"build": "npm run clean && npm run build:esm && npm run build:cjs",
+		"clean": "rimraf lib-esm lib dist",
+		"format": "tsfmt --useTsfmt tsfmt.json -r src/**/*.ts",
```
sammartinez: I thought this was getting removed?

pravgupt: Michael mentioned that it uses typescript internal formatter, but I don't think anybody is using this command. Will remove from the base branch (not this PR)

packages/core/src/Credentials.ts
```diff
@@ -396,30 +362,9 @@ export class CredentialsClass {
 	}
 
 	public async clear() {
-		const { identityPoolId, region } = this._config;
```
haverchuck: @pravgupt We've verified that this workaround is no longer needed?

packages/auth/src/Auth.ts
elorzafe: talked to @pravgupt it requires a middleware change to update UA

packages/core/src/Credentials.ts
```diff
@@ -113,7 +112,6 @@ export class CredentialsClass {
 				provider,
 				token,
 				user,
-				identity_id,
```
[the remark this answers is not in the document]
pravgupt: There is no identity_id received in the federation flow from the federated providers. This identity_id was a big source of confusion in our code base.

packages/core/src/Credentials.ts
```diff
@@ -127,14 +125,12 @@ export class CredentialsClass {
 					.then(data => {
 						logger.debug('refresh federated token sucessfully', data);
 						token = data.token;
-						identity_id = data.identity_id;
```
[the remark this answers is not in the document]
pravgupt: Same as above.

packages/core/src/Credentials.ts
```diff
@@ -127,14 +125,12 @@ export class CredentialsClass {
...
 						expires_at = data.expires_at;
 
 						return that._setCredentialsFromFederation({
 							provider,
 							token,
 							user,
-							identity_id,
```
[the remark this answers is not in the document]
pravgupt: Same as above.

packages/core/src/Credentials.ts
```diff
@@ -189,65 +188,41 @@ export class CredentialsClass {
...
-					);
-					return this._loadCredentials(newCredentials, 'guest', false, null);
-				} else {
-					return e;
-				}
+				return e;
 			});
 	}
 
-	private _setCredentialsFromAWS() {
```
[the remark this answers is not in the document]
pravgupt: Interesting, I didn't know about this feature. @haverchuck Do we want to keep this feature parity in Auth V2?

packages/interactions/src/Providers/AWSLexProvider.ts
```diff
@@ -18,7 +18,7 @@ import {
 	InteractionsResponse,
 } from '../types';
 import * as LexRuntime from 'aws-sdk/clients/lexruntime';
```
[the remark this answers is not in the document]
pravgupt: @jordanranz's PR will take care of it.

packages/core/src/Signer.ts
```diff
@@ -12,28 +12,26 @@
  */
 
 import { ConsoleLogger as Logger } from './Logger';
-import { AWS } from './Facet';
+import { Sha256 as browserSha256 } from '@aws-crypto/sha256-browser';
```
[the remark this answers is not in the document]
pravgupt: I don't fully understand, will sync up with you offline.

packages/core/src/Credentials.ts
```diff
@@ -189,65 +188,41 @@ export class CredentialsClass {
...
-						},
-						{
-							region,
-						}
-					);
-					return this._loadCredentials(newCredentials, 'guest', false, null);
-				} else {
-					return e;
-				}
+				return e;
```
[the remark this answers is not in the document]
pravgupt: I believe it was dead code. loadCredentials was not throwing such exceptions. Moreover we are handling "identityId" not found above where we switch between fromCognitoIdentity when identityId is present or fromCognitoIdentityPool when it isn't.

packages/core/package.json
```diff
@@ -1,88 +1,92 @@
...
+		"prepend-file": "^1.3.1",
+		"prettier": "^1.7.4",
+		"typescript": "^3.2.2"
+	},
+	"dependencies": {
+		"@aws-crypto/sha256-browser": "^0.1.0-preview.1",
+		"@aws-sdk/client-cognito-identity-browser": "^0.1.0-preview.2",
+		"@aws-sdk/credential-provider-cognito-identity": "^0.1.0-preview.2",
+		"@aws-sdk/util-hex-encoding": "^0.1.0-preview.1",
+		"aws-sdk": "2.518.0",
```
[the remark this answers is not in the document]
pravgupt: yes, will remove.

packages/interactions/src/Providers/AWSLexProvider.ts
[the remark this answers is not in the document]
jordanranz: https://github.com/aws-amplify/amplify-js/pull/4103 ^

packages/auth/src/Auth.ts
jordanranz: We should probably track a task for this. If you haven't already created it, I can create one

pravgupt: https://www.pivotaltracker.com/story/show/168889299

packages/core/src/Signer.ts
```diff
@@ -295,7 +291,7 @@ export class Signer {
     *
     * @returns {object} Signed HTTP request
     */
-	static sign(request, access_info, service_info = null) {
+	static async sign(request, access_info, service_info = null) {
```
jordanranz: Customers can be using this API. Do we have a document for breaking changes?

pravgupt: I'm keeping a list of all the incompatible changes made in the V3 and modularization branches. Is there a better place or way I can document them for now?

packages/core/src/Credentials.ts
```diff
@@ -189,65 +188,41 @@ export class CredentialsClass {
...
+		// because https://github.com/aws/aws-sdk-js-v3/issues/354
+		const cognitoClient = new CognitoIdentityClient({
+			region,
+			credentials: () => Promise.resolve({} as any),
+			signer: {} as any,
+		});
+		cognitoClient.middlewareStack.remove('SIGNATURE');
+
+		let credentials = undefined;
+		if (identityId) {
```
jordanranz: This will bug due to the nature of `this._storage.getItem` above. `identityId` at `identityId = this._storage.getItem('CognitoIdentityId-' + identityPoolId);` may return the string "undefined". This will oddly enough will evaluate to true and eventually throw a validation error from `Credentials.get()`. Found this while testing scene loading for XR.

pravgupt: Interesting, was this not a problem before? Since earlier the code was 

```
const credentials = new AWS.CognitoIdentityCredentials(
	{
		IdentityPoolId: identityPoolId,
		IdentityId: identityId ? identityId : undefined,
	},
	{
		region,
	}
);```

It might have been going to the fallback logic where identityId is always passed undefined. I'll add additional condition to check for string "undefined"

File: packages/analytics/src/Providers/AWSPinpointProvider.ts [hunk 15]
```diff
@@ -344,7 +344,7 @@ export class AWSPinpointProvider implements AnalyticsProvider {
 
 		const serviceInfo = { region, service: MOBILE_SERVICE_NAME };
 
-		const requestUrl: string = Signer.signUrl(
+		const requestUrl: string = await Signer.signUrl(
```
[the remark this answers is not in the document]
iartemiev: Good catch @elorzafe. It really depends on how long `Signer.singUrl` takes to execute. Anything > 10ms seems to be problematic in FF.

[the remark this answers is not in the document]
iartemiev: Actually testing this locally and `sendBeacon` doesn't fire on `window.beforeunload` in this branch. But it doesn't seem to be due to this change, i.e. taking out the async/await keywords in this method and commenting out the Signer.signUrl call still doesn't get it to work. Seems like something else is causing a delay in the event handler invocation. I'm investigating it further.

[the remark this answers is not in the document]
elorzafe: Thanks @iartemiev!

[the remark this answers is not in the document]
elorzafe: @pravgupt any news from aws-sdk team to solve this?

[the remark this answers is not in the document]
pravgupt: They said it's possible to provide a sync method. Here is the issue created https://github.com/aws/aws-sdk-js-crypto-helpers/issues/6

packages/core/src/Credentials.ts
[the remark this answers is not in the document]
pravgupt: As discussed offline, we will create a [backlog item](https://www.pivotaltracker.com/story/show/169187682) to explore use cases that would need this functionality but as of now, we believe that this was a hack and not a best practice way to handle credentials.

packages/core/src/Signer.ts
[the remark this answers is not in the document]
pravgupt: Confirmed with a signed Rest API call that there are no performance impact on react native platform. I tested on an app with Expo, is there any difference expected with react-scripts that I should test that too?

packages/analytics/src/Providers/AWSPinpointProvider.ts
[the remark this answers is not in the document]
Amplifiyer: Addressed in the last commit and using the sync method from the crypto library.

packages/core/src/Credentials.ts
```diff
@@ -156,18 +152,18 @@ export class CredentialsClass {
...
 		}
-		logger.debug('is this credentials expired?', credentials);
-		const ts = new Date().getTime();
-		const delta = 10 * 60 * 1000; // 10 minutes
-		const { expired, expireTime } = credentials;
-		if (!expired && expireTime > ts + delta) {
+		logger.debug('are these credentials expired?', credentials);
+		const ts = Math.floor(Date.now() / 1000);
+		const delta = 10 * 60; // 10 minutes in seconds
+		const { expiration } = credentials; // returns unix time stamp
```
elorzafe: If this is in seconds, can we confirm is not changing before GA?

packages/core/src/Credentials.ts
```diff
@@ -294,16 +275,21 @@ export class CredentialsClass {
...
-			},
-			{
-				region,
-			}
-		);
 
+		// Removing the signature middleware and passing empty credentials and signer
+		// because https://github.com/aws/aws-sdk-js-v3/issues/354
+		const cognitoClient = new CognitoIdentityClient({
+			region,
```
elorzafe: What happen if region is `undefined`?

packages/core/src/index.ts
```diff
@@ -31,7 +28,7 @@ export { Credentials } from './Credentials';
 export { ServiceWorker } from './ServiceWorker';
 export { ICredentials } from './types';
 export { StorageHelper, MemoryStorage } from './StorageHelper';
-export { Platform } from './Platform';
+export { Platform, appendAmplifyUserAgent } from './Platform';
```
elorzafe: where is `appendAmplifyUserAgent` being used?

Amplifiyer: @elorzafe, this is a utility function exported by core package so that other packages who are using aws-sdk clients can call this and append amplify user agent. This will be take care by [pivotal story](https://www.pivotaltracker.com/story/show/168889299) to add user agents once this is merged in with other aws-sdk migration PRs.

packages/core/src/Credentials.ts
Amplifiyer: Will confirm.

packages/core/src/Credentials.ts
Amplifiyer: How is this handled elsewhere? and is this a supported use case to have undefined region?

Amplifiyer: Talked offline. Added static validations for region in credentials.

docs/amplify-theme/assets/js/scripts.min.js
```diff
@@ -1 +1,101 @@
...
+!(function(e) {
```
sammartinez: can we look to exclude docs from prettier config? Dont believe we need `.min` files prettiered

packages/core/package.json
```diff
@@ -1,88 +1,93 @@
...
+		"prettier": "^1.7.4",
+		"typescript": "^3.2.2"
+	},
+	"dependencies": {
+		"@aws-crypto/sha256-js": "^0.1.0-preview.2",
+		"@aws-sdk/client-cognito-identity-browser": "^0.1.0-preview.2",
+		"@aws-sdk/credential-provider-cognito-identity": "^0.1.0-preview.2",
+		"@aws-sdk/util-hex-encoding": "^0.1.0-preview.1",
+		"@aws-sdk/util-user-agent-browser": "^0.1.0-preview.2",
+		"aws-sdk": "2.518.0",
```
sammartinez: Is there a reason why this is pinned? Can we not use the latest versions?

docs/amplify-theme/assets/js/scripts.min.js
Amplifiyer: Excluded and vscode directory as well. Don't know why prettier is not running on these directories on master branch and increasingly changing more and more files.

packages/core/package.json
Amplifiyer: Yes, as this PR still has other aws-sdk clients (there are different PRs for each category). This can be only removed once all the PRs are merged to the base branch and no trace of aws-sdk clients exist, until then have to keep it to keep the builds successful.

PR #4078, opened 2019-09-24
[Delivers #168673124] Migrate S3 client to aws sdk V3

Amplifiyer:
_Issue #, if available:_ https://github.com/aws-amplify/amplify-js/issues/3365

_Description of changes:_ This PR upgrades the S3 client from aws sdk V2 to V3 client to achieve modularization at the AWS clients level.

[Delivers #168673124]

### DIFF
--- a/packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
+++ b/packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
@@ -22,9 +22,9 @@ import { ListObjectsCommand } from '@aws-sdk/client-s3-browser/commands/ListObje
  * actually be using dispatchStorageEvent from Storage
  */
 
-S3Client.prototype.send = jest.fn((command, callback) => {
+S3Client.prototype.send = jest.fn(async command => {
 	if (command instanceof ListObjectsCommand) {
-		callback(null, {
+		return {
 			Contents: [
 				{
 					Key: 'public/path/itemsKey',
@@ -33,10 +33,10 @@ S3Client.prototype.send = jest.fn((command, callback) => {
 					Size: 'size',
 				},
 			],
-		});
+		};
 	}
-	callback(null, 'data');
-}) as any;
+	return 'data';
+});
 
 S3RequestPresigner.prototype.presignRequest = jest.fn((request, expires) => {
 	return new Promise<any>((res, rej) => {
@@ -185,8 +185,8 @@ describe('StorageProvider test', () => {
 			storage.configure(options_with_download);
 			const spyon = jest
 				.spyOn(S3Client.prototype, 'send')
-				.mockImplementationOnce((params, callback) => {
-					callback(null, { Body: [1, 2] } as any);
+				.mockImplementationOnce(async params => {
+					return { Body: [1, 2] };
 				});
 
 			expect.assertions(2);
@@ -210,8 +210,8 @@ describe('StorageProvider test', () => {
 			storage.configure(options);
 			jest
 				.spyOn(S3Client.prototype, 'send')
-				.mockImplementationOnce((params, callback) => {
-					callback('err', null);
+				.mockImplementationOnce(async params => {
+					throw 'err';
 				});
 
 			expect.assertions(1);
@@ -474,8 +474,8 @@ describe('StorageProvider test', () => {
 			storage.configure(options);
 			const spyon = jest
 				.spyOn(S3Client.prototype, 'send')
-				.mockImplementationOnce((params, callback) => {
-					callback('err', null);
+				.mockImplementationOnce(async params => {
+					throw 'err';
 				});
 
 			expect.assertions(1);
@@ -596,8 +596,8 @@ describe('StorageProvider test', () => {
 			storage.configure(options);
 			const spyon = jest
 				.spyOn(S3Client.prototype, 'send')
-				.mockImplementationOnce((params, callback) => {
-					callback('err', null);
+				.mockImplementationOnce(async params => {
+					throw 'err';
 				});
 
 			expect.assertions(1);
@@ -727,8 +727,8 @@ describe('StorageProvider test', () => {
 			storage.configure(options);
 			const spyon = jest
 				.spyOn(S3Client.prototype, 'send')
-				.mockImplementationOnce((params, callback) => {
-					callback('err', null);
+				.mockImplementationOnce(async params => {
+					throw 'err';
 				});
 
 			expect.assertions(1);
--- a/packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
+++ b/packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
@@ -39,9 +39,7 @@ S3Client.prototype.send = jest.fn(async command => {
 });
 
 S3RequestPresigner.prototype.presignRequest = jest.fn((request, expires) => {
-	return new Promise<any>((res, rej) => {
-		res();
-	});
+	return (Promise as any).resolve();
 });
 
 const credentials = {
@@ -414,9 +412,9 @@ describe('StorageProvider test', () => {
 			const spyon = jest.spyOn(S3Client.prototype, 'send');
 
 			expect.assertions(2);
-			expect(await storage.put('key', 'obejct', {})).toEqual({ key: 'key' });
+			expect(await storage.put('key', 'object', {})).toEqual({ key: 'key' });
 			expect(spyon.mock.calls[0][0].input).toEqual({
-				Body: 'obejct',
+				Body: 'object',
 				Bucket: 'bucket',
 				ContentType: 'binary/octet-stream',
 				Key: 'public/key',
@@ -436,11 +434,11 @@ describe('StorageProvider test', () => {
 			const spyon2 = jest.spyOn(Hub, 'dispatch');
 
 			expect.assertions(3);
-			expect(await storage.put('key', 'obejct', { track: true })).toEqual({
+			expect(await storage.put('key', 'object', { track: true })).toEqual({
 				key: 'key',
 			});
 			expect(spyon.mock.calls[0][0].input).toEqual({
-				Body: 'obejct',
+				Body: 'object',
 				Bucket: 'bucket',
 				ContentType: 'binary/octet-stream',
 				Key: 'public/key',
@@ -480,7 +478,7 @@ describe('StorageProvider test', () => {
 
 			expect.assertions(1);
 			try {
-				await storage.put('key', 'obejct', {});
+				await storage.put('key', 'object', {});
 			} catch (e) {
 				expect(e).toBe('err');
 			}
@@ -501,13 +499,13 @@ describe('StorageProvider test', () => {
 
 			expect.assertions(2);
 			expect(
-				await storage.put('key', 'obejct', {
+				await storage.put('key', 'object', {
 					level: 'private',
 					contentType: 'text/plain',
 				})
 			).toEqual({ key: 'key' });
 			expect(spyon.mock.calls[0][0].input).toEqual({
-				Body: 'obejct',
+				Body: 'object',
 				Bucket: 'bucket',
 				ContentType: 'text/plain',
 				Key: 'private/id/key',
--- a/packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
+++ b/packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
@@ -12,43 +12,37 @@
  */
 import StorageProvider from '../../src/providers/AWSS3Provider';
 import { Hub, Credentials } from '@aws-amplify/core';
-import * as S3 from 'aws-sdk/clients/s3';
+import * as formatURL from '@aws-sdk/util-format-url';
+import { S3Client } from '@aws-sdk/client-s3-browser/S3Client';
+import { S3RequestPresigner } from '@aws-sdk/s3-request-presigner';
+import { ListObjectsCommand } from '@aws-sdk/client-s3-browser/commands/ListObjectsCommand';
 
 /**
  * NOTE - These test cases use Hub.dispatch but they should
  * actually be using dispatchStorageEvent from Storage
  */
 
-S3.prototype.getSignedUrl = jest.fn((key, params) => {
-	return 'url';
-});
-
-S3.prototype.getObject = jest.fn((params, callback) => {
+S3Client.prototype.send = jest.fn((command, callback) => {
+	if (command instanceof ListObjectsCommand) {
+		callback(null, {
+			Contents: [
+				{
+					Key: 'public/path/itemsKey',
+					ETag: 'etag',
+					LastModified: 'lastmodified',
+					Size: 'size',
+				},
+			],
+		});
+	}
 	callback(null, 'data');
-});
+}) as any;
 
-S3.prototype.deleteObject = jest.fn((params, callback) => {
-	callback(null, 'data');
-});
-
-S3.prototype.listObjects = jest.fn((params, callback) => {
-	callback(null, {
-		Contents: [
-			{
-				Key: 'public/path/itemsKey',
-				ETag: 'etag',
-				LastModified: 'lastmodified',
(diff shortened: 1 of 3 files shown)


packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -127,7 +134,8 @@ export class AWSS3Provider implements StorageProvider {
 
 		if (download === true) {
 			return new Promise<any>((res, rej) => {
-				s3.getObject(params, (err, data) => {
+				const getObjectCommand = new GetObjectCommand(params);
+				s3.send(getObjectCommand, (err, data) => {
```
iartemiev: Just wondering why we're sticking with callbacks with `s3.send` when v3 supports promises?

For example:

```javascript
if (download === true) {
  try {
    const getObjectCommand = new GetObjectCommand(params);
    const data = await s3.getSignedUrl(getObjectCommand);
    dispatchStorageEvent(
      track,
      'download', {
        method: 'get',
        result: 'success'
      }, {
        fileSize: Number(data.Body['length'])
      },
      `Download success for ${key}`
    );
    return data;
  } catch (err) {
    dispatchStorageEvent(
      track,
      'download', {
        method: 'get',
        result: 'failed',
      },
      null,
      `Download failed with ${err.message}`
    );
    return err;
  }
}
```

Or:

```javascript
if (download === true) {
  const getObjectCommand = new GetObjectCommand(params);
  return s3
    .send(getObjectCommand)
    .then(data => {
      dispatchStorageEvent(
        track,
        'download', {
          method: 'get',
          result: 'success'
        }, {
          fileSize: Number(data.Body['length'])
        },
        `Download success for ${key}`
      );
      return data;
    })
    .catch(err => {
      dispatchStorageEvent(
        track,
        'download', {
          method: 'get',
          result: 'failed',
        },
        null,
        `Download failed with ${err.message}`
      );
      return err;
    });
}
```

jordanranz: I think the reason this was not migrated away from the callback function was to change the least code as possible for the migration. On the flip side we should probably take this opportunity to migrate to try/catch async/await as a refactor. I'll go ahead and add this to my PR as well.

pravgupt: Discussed with Jordan and Sam, we will go ahead with the first approach and update all PRs

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
```diff
@@ -12,44 +12,38 @@
...
-	return 'url';
-});
-
-S3.prototype.getObject = jest.fn((params, callback) => {
-	callback(null, 'data');
-});
-
-S3.prototype.deleteObject = jest.fn((params, callback) => {
-	callback(null, 'data');
+S3Client.prototype.send = jest.fn(async command => {
```
sammartinez: NIT: Can you wrap `command` in parenthesis, looks slightly strange without it.

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
```diff
@@ -129,45 +128,36 @@ describe('StorageProvider test', () => {
...
 			expect.assertions(2);
 			expect(await storage.get('key', { downloaded: false })).toBe('url');
-			expect(spyon).toBeCalledWith('getObject', {
-				Bucket: 'bucket',
-				Key: 'public/key',
-			});
-
-			spyon.mockClear();
-			curCredSpyOn.mockClear();
+			expect(spyon.mock.calls[0][0].path).toEqual('/bucket/public/key');
```
sammartinez: Can i check with you on this when you're in today?

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
```diff
@@ -425,52 +402,44 @@ describe('StorageProvider test', () => {
...
 			storage.configure(options);
-			const spyon = jest.spyOn(S3.prototype, 'upload');
+			const spyon = jest.spyOn(S3Client.prototype, 'send');
 
 			expect.assertions(2);
-			expect(await storage.put('key', 'obejct', {})).toEqual({
-				key: 'path/itemsKey',
-			});
-			expect(spyon).toBeCalledWith({
+			expect(await storage.put('key', 'obejct', {})).toEqual({ key: 'key' });
```
sammartinez: typo `obejct`

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
```diff
@@ -638,27 +583,21 @@ describe('StorageProvider test', () => {
...
 		});
 
 		test('remove object failed', async () => {
-			const curCredSpyOn = jest
-				.spyOn(Credentials, 'get')
-				.mockImplementationOnce(() => {
-					return new Promise((res, rej) => {
-						res({});
-					});
+			jest.spyOn(Credentials, 'get').mockImplementationOnce(() => {
```
sammartinez: we may want to consider moving these `spyOns` towards the top of the file, I am seeing some dupes and im wondering if we can centralize them

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -126,61 +133,61 @@ export class AWSS3Provider implements StorageProvider {
...
+			)) as any);
+			dispatchStorageEvent(
+				track,
+				'getSignedUrl',
+				{ method: 'get', result: 'success' },
+				null,
+				`Signed URL: ${url}`
+			);
+			return url;
+		} catch (e) {
```
sammartinez: Can we be consistent on either `e` or `err` or `error`? From above its `err`, I am totally cool with whatever, just want to be consistent

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
iartemiev: Prettier controls that. It doesn't wrap single params (unless they have a type), i.e. `async (command: someType) => {}`

sammartinez: gotcha. I had a feeling but was curious what we decided on that for it

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
```diff
@@ -12,44 +12,38 @@
...
-		Contents: [
-			{
-				Key: 'public/path/itemsKey',
-				ETag: 'etag',
-				LastModified: 'lastmodified',
-				Size: 'size',
-			},
-		],
+S3RequestPresigner.prototype.presignRequest = jest.fn((request, expires) => {
+	return new Promise<any>((res, rej) => {
```
jordanranz: Can this just be Promise.resolve()?

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
pravgupt: Fixed!

packages/storage/src/providers/AWSS3Provider.ts
pravgupt: Changed all to error

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
pravgupt: As discussed not all spyons implementation is same. We can try to refactor these tests as paramaterized tests, but that is outside the scope of this PR.

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
pravgupt: Changed

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
```diff
@@ -129,45 +126,36 @@ describe('StorageProvider test', () => {
...
 			expect.assertions(2);
 			expect(await storage.get('key', { downloaded: false })).toBe('url');
-			expect(spyon).toBeCalledWith('getObject', {
-				Bucket: 'bucket',
-				Key: 'public/key',
-			});
-
-			spyon.mockClear();
-			curCredSpyOn.mockClear();
+			expect(spyon.mock.calls[0][0].path).toEqual('/bucket/public/key');
```
elorzafe: Why you prefer `(spyon.mock.calls[0][0]` instead of `toBeCalledWith`?

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -257,46 +264,42 @@ export class AWSS3Provider implements StorageProvider {
 				params.SSEKMSKeyId = SSEKMSKeyId;
 			}
 		}
-
 		try {
-			const upload = s3.upload(params).on('httpUploadProgress', progress => {
-				if (progressCallback) {
```
elorzafe: This is something we need to add. We can create a feature-request [here](https://github.com/aws/aws-sdk-js-v3)

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -126,61 +133,61 @@ export class AWSS3Provider implements StorageProvider {
...
+					`Download failed with ${error.message}`
 				);
-				rej(e);
+				throw error;
 			}
-		});
+		}
+
+		params.Expires = expires || 900; // Default is 15 mins as defined in V2 AWS SDK
+		params.Expires = new Date(Date.now() + params.Expires * 1000); // expires is in secs
```
elorzafe: Why is required to use the timestamp?, I think it should be just time in seconds

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -257,46 +264,42 @@ export class AWSS3Provider implements StorageProvider {
...
 				track,
 				'upload',
 				{ method: 'put', result: 'success' },
 				null,
 				`Upload success for ${key}`
 			);
-
 			return {
-				key: data.Key.substr(prefix.length),
+				key,
```
elorzafe: This is something we need to discuss, that will impact the way UI components use this. 
The origin of that is to get the key without prefix in order to use it later for `get` because that automatically appends the prefix depending on the `level` and `identityId`. Get and list remain the same in that case it should be consistent with those.

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -257,46 +264,42 @@ export class AWSS3Provider implements StorageProvider {
...
 				'upload',
 				{ method: 'put', result: 'failed' },
 				null,
 				`Error uploading ${key}`
 			);
-
-			throw e;
+			throw error;
 		}
+		// This functionality is not available in V3 SDK. Amplify needs to implement it.
```
elorzafe: We should create a task to keep track of remaining pieces before release

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -479,10 +479,7 @@ export class AWSS3Provider implements StorageProvider {
 			};
 		}
 
-		return new S3({
-			apiVersion: '2006-03-01',
-			params: { Bucket: bucket },
-			signatureVersion: 'v4',
+		return new S3Client({
```
elorzafe: Is `localTestingConfig` compatible the v3 client?

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
```diff
@@ -129,45 +126,36 @@ describe('StorageProvider test', () => {
...
+			jest.spyOn(formatURL, 'formatUrl').mockReturnValueOnce('url');
+
 			expect.assertions(2);
 			expect(await storage.get('key', { downloaded: false })).toBe('url');
-			expect(spyon).toBeCalledWith('getObject', {
-				Bucket: 'bucket',
-				Key: 'public/key',
-			});
-
-			spyon.mockClear();
```
jordanranz: Curious why we are removing mock clearing here? Or we could move this to an afterEach?

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
pravgupt: I found this to be cleaner when I want to assert only some part of the input object rather than constructing the full mock object. For instance this test only concerns the path field of the first input rather than the full object or other parameters.

packages/storage/src/providers/AWSS3Provider.ts
pravgupt: Yes, there is already a task [here](https://www.pivotaltracker.com/story/show/168994501) which needs to be implemented by Amplify and not in the SDK as discussed in the V3 meeting.

packages/storage/src/providers/AWSS3Provider.ts
pravgupt: V3 S3PreSigne requires a date object as input.

packages/storage/src/providers/AWSS3Provider.ts
pravgupt: Task is in our [backlog](https://www.pivotaltracker.com/story/show/168994501)

packages/storage/src/providers/AWSS3Provider.ts
pravgupt: I'm not sure if we relied on the SDK support for local testing feature, good call out, I'll test it out.

packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
pravgupt: all mocks reset has been moved to an afterEach() method. I find this incredibly helpful since it avoids polluting state of mocks between tests explicitly and doesn't rely on individual tests to call clear/reset.

packages/storage/src/providers/AWSS3Provider.ts
pravgupt: Clarified offline. I'll also test this once again to ensure there are no difference pre and post migration.

pravgupt: Verified that it's the same response.

packages/storage/src/providers/AWSS3Provider.ts
pravgupt: Tested and it works as expected.

packages/storage/src/providers/AWSS3Provider.ts
elorzafe: this could be a problem if the device clock is skewed. Ping me when you can to discuss about it

pravgupt: SDK is actually creating a difference between this date and the current date so the clock skew should not be an issue since it will impact both the dates equally and the diff (interval) which is sent to S3 service won't be impacted. See https://github.com/aws/aws-sdk-js-v3/blob/master/packages/signature-v4/src/SignatureV4.ts#L136-L139

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -16,13 +16,20 @@ import {
 	Credentials,
 	Parser,
 } from '@aws-amplify/core';
-import * as S3 from 'aws-sdk/clients/s3';
+import { S3Client } from '@aws-sdk/client-s3-browser/S3Client';
```
elorzafe: are sub  `@aws-sdk/client-s3-browser` package published independently in case customers dont have tree shaking?

packages/storage/src/providers/AWSS3Provider.ts
elorzafe: Nice

packages/storage/src/providers/AWSS3Provider.ts
Amplifiyer: yes `@aws-sdk/client-s3-browser` is an independent package. This will be changed with a universal package in coming weeks which would have a browser field in package.json to only get the browser specific stuff.

PR #4092, opened 2019-09-26
fix: lerna linking for unstable and beta

iartemiev:
This is a workaround for the following issue when using lerna with the `--canary` option:
https://github.com/lerna/lerna/issues/2060


iartemiev:
Big thanks to @pravgupt for finding the solution!

PR #4103, opened 2019-09-30
[Delivers #168789986] Migrate Interactions category to use aws-sdk v3 clients

jordanranz:
Related to #3365

Description of changes:
This PR upgrades the aws-sdk v2 clients used in the Interactions category to aws-sdk V3 client. This helps achieve modularization at the AWS clients level in Interactions.

[Delivers #168789986]

### DIFF
--- a/packages/interactions/__tests__/Interactions-unit-test.ts
+++ b/packages/interactions/__tests__/Interactions-unit-test.ts
@@ -1,66 +1,70 @@
-jest.mock('aws-sdk/clients/lexruntime', () => {
-	const LexRuntime = () => {};
-	LexRuntime.prototype.postText = (params, callback) => {
-		if (params.inputText === 'done') {
-			callback(null, {
-				message: 'echo:' + params.inputText,
+import { LexRuntimeServiceClient } from '@aws-sdk/client-lex-runtime-service-browser/LexRuntimeServiceClient';
+import Interactions from '../src/Interactions';
+import { findInterfacesAddedToObjectTypes } from 'graphql/utilities/findBreakingChanges';
+import { AWSLexProvider, AbstractInteractionsProvider } from '../src/Providers';
+import { Credentials } from '@aws-amplify/core';
+import { PostContentCommand } from '@aws-sdk/client-lex-runtime-service-browser/commands/PostContentCommand';
+import { PostTextCommand } from '@aws-sdk/client-lex-runtime-service-browser/commands/PostTextCommand';
+
+LexRuntimeServiceClient.prototype.send = jest.fn((command, callback) => {
+	if (command instanceof PostTextCommand) {
+		if (command.input.inputText === 'done') {
+			const result = {
+				message: 'echo:' + command.input.inputText,
 				dialogState: 'ReadyForFulfillment',
 				slots: {
 					m1: 'hi',
 					m2: 'done',
 				},
-			});
+			};
+			return Promise.resolve(result);
 		} else {
-			callback(null, {
-				message: 'echo:' + params.inputText,
+			const result = {
+				message: 'echo:' + command.input.inputText,
 				dialogState: 'ElicitSlot',
-			});
+			};
+			return Promise.resolve(result);
 		}
-	};
-
-	LexRuntime.prototype.postContent = (params, callback) => {
-		if (params.contentType === 'audio/x-l16; sample-rate=16000') {
-			if (params.inputStream === 'voice:done') {
-				callback(null, {
-					message: 'voice:echo:' + params.inputStream,
+	} else if (command instanceof PostContentCommand) {
+		if (command.input.contentType === 'audio/x-l16; sample-rate=16000') {
+			if (command.input.inputStream === 'voice:done') {
+				const result = {
+					message: 'voice:echo:' + command.input.inputStream,
 					dialogState: 'ReadyForFulfillment',
 					slots: {
 						m1: 'voice:hi',
 						m2: 'voice:done',
 					},
-				});
+				};
+				return Promise.resolve(result);
 			} else {
-				callback(null, {
-					message: 'voice:echo:' + params.inputStream,
+				const result = {
+					message: 'voice:echo:' + command.input.inputStream,
 					dialogState: 'ElicitSlot',
-				});
+				};
+				return Promise.resolve(result);
 			}
 		} else {
-			if (params.inputStream === 'done') {
-				callback(null, {
-					message: 'echo:' + params.inputStream,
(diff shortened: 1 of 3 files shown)


packages/interactions/__tests__/Interactions-unit-test.ts
```diff
@@ -1,66 +1,70 @@
-jest.mock('aws-sdk/clients/lexruntime', () => {
-	const LexRuntime = () => {};
-	LexRuntime.prototype.postText = (params, callback) => {
-		if (params.inputText === 'done') {
-			callback(null, {
-				message: 'echo:' + params.inputText,
+import { LexRuntimeServiceClient } from '@aws-sdk/client-lex-runtime-service-browser/LexRuntimeServiceClient';
+import Interactions from '../src/Interactions';
+import { findInterfacesAddedToObjectTypes } from 'graphql/utilities/findBreakingChanges';
```
elorzafe: Why is this needed? I know I added lol, but it seems it shouldn't be there

packages/interactions/src/Providers/AWSLexProvider.ts
```diff
@@ -12,153 +12,151 @@
...
-	InteractionsOptions,
-	InteractionsMessage,
-	InteractionsResponse,
-} from '../types';
-import * as LexRuntime from 'aws-sdk/clients/lexruntime';
+import { InteractionsOptions, InteractionsMessage } from '../types';
+import { LexRuntimeServiceClient } from '@aws-sdk/client-lex-runtime-service-browser/LexRuntimeServiceClient';
+import { PostTextCommand } from '@aws-sdk/client-lex-runtime-service-browser/commands/PostTextCommand';
+import { PostContentCommand } from '@aws-sdk/client-lex-runtime-service-browser/commands/PostContentCommand';
 import { ConsoleLogger as Logger, AWS, Credentials } from '@aws-amplify/core';
```
elorzafe: `AWS` object won't be needed here

packages/interactions/src/Providers/AWSLexProvider.ts
```diff
@@ -12,153 +12,151 @@
...
+        if (!this._config[botname]) {
+            return Promise.reject('Bot ' + botname + ' does not exist');
+        }
+        const credentials = await Credentials.get();
+        if (!credentials) {
+            return Promise.reject('No credentials');
+        }
+        AWS.config.update({
+            credentials,
+        });
```
elorzafe: This is not needed

packages/interactions/src/Providers/AWSLexProvider.ts
```diff
@@ -12,153 +12,151 @@
...
+                    userId: credentials.identityId,
+                    accept: 'audio/mpeg',
+                };
+            }
+
+            logger.debug('postContent to lex', message);
+
+            try {
+                const postContentCommand = new PostContentCommand(params);
+                const data = this.lexRuntimeServiceClient.send(postContentCommand);
```
elorzafe: Is this synchronous?

packages/interactions/src/Providers/AWSLexProvider.ts
```diff
@@ -12,153 +12,151 @@
...
+                this.responseCallback(data, botname);
```
elorzafe: `this.responseCallback` is asynchronous, completion callback can be called after the response is sent

packages/interactions/__tests__/Interactions-unit-test.ts
jordanranz: I do not believe it is needed, I can remove it.

packages/interactions/src/Providers/AWSLexProvider.ts
jordanranz: Awesome, I can remove.

packages/interactions/src/Providers/AWSLexProvider.ts
jordanranz: Good catch

packages/interactions/src/Providers/AWSLexProvider.ts
jordanranz: what do you mean by "completion callback" here?

packages/interactions/src/Providers/AWSLexProvider.ts
jordanranz: Removing this

packages/interactions/src/Providers/AWSLexProvider.ts
```diff
@@ -12,153 +12,148 @@
...
+                userId: credentials.identityId,
+            };
+
+            logger.debug('postText to lex', message);
+
+            try {
+                const postTextCommand = new PostTextCommand(params);
+                const data = await this.lexRuntimeServiceClient.send(postTextCommand);
+                this.responseCallback(data, botname);
+                return data;
```
pravgupt: Shouldn't we return the response from the responseCallBack() as was before in case the callback changes/updates the data or conditionally return the response? Also we should rename the method as it isn't technically callback anymore.

packages/interactions/src/Providers/AWSLexProvider.ts
```diff
@@ -12,153 +12,148 @@
...
-					typeof this._config[botname].onComplete === 'function'
-				) {
-					setTimeout(
-						() => this._config[botname].onComplete(null, { slots: data.slots }),
-						0
-					);
-				}
-			}
-
-			res(data);
```
pravgupt: Is it logically same to move the resolve from here in the middle of this method to skip it to the end?

jordanranz: This will logically be the same.

packages/interactions/src/Providers/AWSLexProvider.ts
jordanranz: I can change the name of the function here. I mainly left it around to preserve the content but I think it is fair to change the naming of it to better describe the functionality.

packages/interactions/src/Providers/AWSLexProvider.ts
jordanranz: Spoke offline about this. Seems the responseCallback is just doing some error logging to the chatbot. I will change the name of it to better describe the functionality.

packages/interactions/src/Providers/AWSLexProvider.ts
elorzafe: `responseCallBack` is invoking onComplete callbacks that are configured. I think `reportBotStatus` is accurate

Issue #3950, opened 2019-08-30
aws-amplify-react Picker component input overflows


rogerchi:
**Describe the bug**
The Picker component has an input field which overflows because the parent div does not have "position: relative" in its styling.  A previous commit removed this styling.

**To Reproduce**
Use <Picker> component.  See that the clickable input area overflows.

**Expected behavior**
The clickable area is restrained to the button area.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Web
 - Browser: Chrome
 - Version: 76.0.3809.132

rahulbaisla:
I found the same behavior.

Added following to my  project's theme  and exporting it works for me .

--------------------------------------------------------
 export const pickerPicker = {
    position: 'relative'
 };

const AppTheme = {
    pickerPicker: pickerPicker
}

export default AppTheme;
--------------------------------------------------------

ashika01:
Closing the issue on back of the merged PR. Thank you for brining the issue up.

PR #4135, opened 2019-10-04
fix(#3950): picker style issue

ashika01:
_Issue #, if available:_ #3950 

_Description of changes:_ Just a line of styling that got lost over history.

### DIFF
--- a/packages/aws-amplify-react/src/Widget/Picker.tsx
+++ b/packages/aws-amplify-react/src/Widget/Picker.tsx
@@ -66,10 +66,15 @@ export default class Picker extends Component<IPickerProps, {}> {
 		const accept = this.props.accept || '*/*';
 
 		const theme = this.props.theme || AmplifyTheme;
+		const pickerStyle = Object.assign(
+			{},
+			{ position: 'relative' },
+			theme.pickerPicker
+		);
 		const inputStyle = Object.assign({}, PickerInput, theme.pickerInput);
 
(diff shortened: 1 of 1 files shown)


ashika01:
There was no real way of testing this because the change lies more on styling side of things. Any ideas welcome.

sammartinez:
> There was no real way of testing this because the change lies more on styling side of things. Any ideas welcome.

Can you provide a screenshot of the change? With and without the style change

packages/aws-amplify-react/src/Widget/Picker.tsx
```diff
@@ -66,10 +66,15 @@ export default class Picker extends Component<IPickerProps, {}> {
 		const accept = this.props.accept || '*/*';
 
 		const theme = this.props.theme || AmplifyTheme;
+		const pickerStyle = Object.assign(
+			{},
+			{ position: 'relative' },
```
jordanranz: We should move this to a css module file so it can be shared between frameworks.

ashika01: I didn't know we are using cssm. I just returned it to its old state. I will go add and do that.

jordanranz: Spoke offline, I think we will consolidate this style into our new Stencil component and skip doing the work to move this to a CSS module for now.

PR #4149, opened 2019-10-08
feat(@aws-amplify/auth): clientMetadata

haverchuck:
do not merge yet

_Description of changes:_
Passes clientMetadata through Auth to amazon-cognito-identity-js to Cognito/Lambda.
Merge with [docs PR](https://github.com/aws-amplify/docs/pull/966/).


packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -256,11 +261,16 @@ export default class CognitoUser {
 				authParameters.CHALLENGE_NAME = 'SRP_A';
 			}
 
+			const clientMetaData =
```
sammartinez: What is the difference of this `clientMetaData` and the one listed above (lines 157-161)? Do we need to define these twice or can we use one of them?

haverchuck: @sammartinez They are separate functions - although they will often be called in sequence, I think it's better to decouple them by reinitializing this variable rather than passing it through.

packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -34,6 +34,7 @@ declare module 'amazon-cognito-identity-js' {
 		Username: string;
 		Password?: string;
 		ValidationData?: { [key: string]: any };
+		ClientMetadata?: any;
```
undefobj: Why are we defining this as `any`? Can we not be a little more strict?

packages/auth/src/types/Auth.ts
```diff
@@ -25,6 +25,7 @@ export interface SignUpParams {
 	password: string;
 	attributes?: object;
 	validationData?: CognitoUserAttribute[];
+	clientMetadata?: any;
```
undefobj: Same question as above, is there a better option than `any` for this feature?

spencersteers:
The [docs](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_SignUp.html#API_SignUp_RequestSyntax) specify `clientMetadata` type as 'string to string map'.

packages/amazon-cognito-identity-js/index.d.ts
haverchuck: Went ahead and added a ClientMetadata type for { [key: string]: string }.
This will throw some errors (including in the UI) that only a developer will understand, if they don't strucuture their metadata right.  I'm assuming that's okay.

When initializing parameters with defaults in some of the function signatures, I'm doing `ClientMetadata | undefined`; this is to avoid throwing an error if they do not have any default metadata in their config object.  The alternative would be more verbose assignment with the body of the function.

packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -361,4 +368,6 @@ declare module 'amazon-cognito-identity-js' {
 		removeItem(key: string): void;
 		clear(): void;
 	}
+
+	export type ClientMetadata = { [key: string]: string };
```
Amplifiyer: Can we create this type earlier in the file and use it later in the file?

packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -87,32 +88,38 @@ declare module 'amazon-cognito-identity-js' {
...
-			onFailure: (err: Error) => void;
-			inputVerificationCode?: (data: any) => void;
-		}): void;
+		public forgotPassword(
+			callbacks: {
+				onSuccess: (data: any) => void;
+				onFailure: (err: Error) => void;
+				inputVerificationCode?: (data: any) => void;
+			},
+			authDetails?: { [key: string]: string } | undefined
```
Amplifiyer: any benefit of specifying `| undefined` since the variable is optional anyways?

packages/amazon-cognito-identity-js/src/AuthenticationDetails.js
```diff
@@ -60,4 +67,11 @@ export default class AuthenticationDetails {
 	getAuthParameters() {
 		return this.authParameters;
 	}
+
+	/**
+	 * @returns {Object} the clientMetadata for a Lambda trigger
```
Amplifiyer: I think we can add the ClientMetadata type here instead of Object.

packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -565,9 +581,15 @@ export default class CognitoUser {
 	 * @param {customChallenge} callback.customChallenge Custom challenge
 	 *         response required to continue.
 	 * @param {authSuccess} callback.onSuccess Called on success with the new session.
+	 * @param {object} clientMetadata object which is passed from client to Cognito Lambda trigger
```
Amplifiyer: We can add the correct type here.

packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -565,9 +581,15 @@ export default class CognitoUser {
...
 	 * @returns {void}
 	 */
-	completeNewPasswordChallenge(newPassword, requiredAttributeData, callback) {
+	completeNewPasswordChallenge(
+		newPassword,
+		requiredAttributeData,
+		callback,
+		clientMetadata
+	) {
```
Amplifiyer: I think this method(and some others down later) got missed in the declarations file.

packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -87,32 +88,38 @@ declare module 'amazon-cognito-identity-js' {
 		public confirmRegistration(
 			code: string,
 			forceAliasCreation: boolean,
-			callback: NodeCallback<any, any>
+			callback: NodeCallback<any, any>,
+			clientMetadata?: { [key: string]: string }
 		): void;
```
Amplifiyer: Do we need to update any documentation with these API changes, e.g. https://github.com/aws-amplify/amplify-js/blob/master/packages/amazon-cognito-identity-js/README.md

packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -1932,6 +1984,7 @@ export default class CognitoUser {
 	 * @param {string} totpCode The MFA code entered by the user.
 	 * @param {string} friendlyDeviceName The device name we are assigning to the device.
 	 * @param {nodeCallback<string>} callback Called on success or error.
+	 * @param {object} clientMetadata object which is passed from client to Cognito Lambda trigger
```
Amplifiyer: It doesn't seem to be added to this method.

packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -87,32 +88,38 @@ declare module 'amazon-cognito-identity-js' {
...
 		public sendCustomChallengeAnswer(
 			answerChallenge: any,
 			callback: IAuthenticationCallback
 		): void;
 		public resendConfirmationCode(
-			callback: NodeCallback<Error, 'SUCCESS'>
+			callback: NodeCallback<Error, 'SUCCESS'>,
+			authDetails?: { [key: string]: string } | undefined
```
Amplifiyer: The method definition uses the second argument as clientMetadata, should these all be clientMetadata in the declaration file?

packages/auth/src/types/Auth.ts
```diff
@@ -201,6 +204,12 @@ export interface AuthErrorMessage {
 // We can extend this in the future if needed
 export type SignInOpts = UsernamePasswordOpts;
 
+export type ClientMetaData =
+	| {
```
Amplifiyer: Is the leading pipe required?

packages/auth/src/types/Auth.ts
```diff
@@ -156,6 +158,7 @@ export type OAuthOpts = AwsCognitoOAuthOpts | Auth0OAuthOpts;
 
 export interface ConfirmSignUpOptions {
 	forceAliasCreation?: boolean;
+	clientMetadata?: any;
```
Amplifiyer: can we use the type here?

packages/auth/src/types/Auth.ts
```diff
@@ -25,6 +25,7 @@ export interface SignUpParams {
 	password: string;
 	attributes?: object;
 	validationData?: CognitoUserAttribute[];
+	clientMetadata?: { [key: string]: string };
```
Amplifiyer: can we use the defined type here?

packages/auth/src/Auth.ts
```diff
@@ -359,7 +381,10 @@ export default class AuthClass {
 	 * @param {String} username - The username to be confirmed
 	 * @return - A promise resolves data if success
```
Amplifiyer: can we add new param to jsDoc (and to rest other API changes in this file). Do we need any other documentation update for this public api?

packages/auth/src/Auth.ts
```diff
@@ -387,14 +412,16 @@ export default class AuthClass {
 	 */
 	public signIn(
 		usernameOrSignInOpts: string | SignInOpts,
-		pw?: string
+		pw?: string,
+		clientMetadata: ClientMetaData | undefined = this._config.clientMetadata
```
Amplifiyer: probably can use ? instead of | undefined

packages/auth/src/Auth.ts
```diff
@@ -359,7 +381,10 @@ export default class AuthClass {
 	 * @param {String} username - The username to be confirmed
 	 * @return - A promise resolves data if success
 	 */
-	public resendSignUp(username: string): Promise<string> {
+	public resendSignUp(
+		username: string,
+		clientMetadata: any = this._config.clientMetadata
```
Amplifiyer: type can be added here

packages/amazon-cognito-identity-js/src/CognitoUserPool.js
```diff
@@ -90,13 +90,21 @@ export default class CognitoUserPool {
...
+		callback,
+		clientMetadata
+	) {
 		const jsonReq = {
 			ClientId: this.clientId,
 			Username: username,
 			Password: password,
 			UserAttributes: userAttributes,
 			ValidationData: validationData,
+			ClientMetadata: clientMetadata,
```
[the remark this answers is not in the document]
haverchuck: Because for this CognitoUserPool.signUp, authDetails is not passed in.

packages/amazon-cognito-identity-js/index.d.ts
haverchuck: Yes, good catch :)

packages/amazon-cognito-identity-js/src/AuthenticationDetails.js
haverchuck: Done.

packages/amazon-cognito-identity-js/src/CognitoUser.js
haverchuck: Done.

packages/amazon-cognito-identity-js/src/CognitoUser.js
haverchuck: Done.

packages/auth/src/types/Auth.ts
haverchuck: Yes.

packages/auth/src/types/Auth.ts
haverchuck: Fixed.

packages/auth/src/Auth.ts
haverchuck: I believe that optional parameters cannot be initializer (`[ts] Parameter cannot have question mark and initializer`), meaning that would would need to do something like...
`let myClientMetadata = clientMetadata || this._config.clientMetadata`
...in the body of the function.  I personally thought this way was cleaner.

packages/auth/src/Auth.ts
haverchuck: done.

packages/amazon-cognito-identity-js/src/CognitoUser.js
haverchuck: Intentionally removed this... forgot to revert the jsdoc.  It's fixed now.

packages/auth/src/Auth.ts
haverchuck: Done.

packages/auth/src/types/Auth.ts
haverchuck: Added a type.

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -252,12 +289,74 @@ describe('auth unit test', () => {
...
+			expect(await CognitoUserPool.prototype.signUp).toBeCalledWith(
+				attrs.username,
+				attrs.password,
+				[
+					{ Name: 'email', Value: 'email' },
+					{ Name: 'phone_number', Value: 'phone_number' },
+					{ Name: 'otherAttrs', Value: 'otherAttrs' },
+				],
+				null,
+				jasmine.any(Function),
```
sammartinez: Does [expect.any(Function)](https://jestjs.io/docs/en/expect#expectanyconstructor) similar to this?

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -1769,19 +2148,74 @@ describe('auth unit test', () => {
 	});
 
 	describe('forgotPasswordSubmit', () => {
+		// 	test('happy case', async () => {
```
sammartinez: Do we need this test?

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -1769,19 +2148,74 @@ describe('auth unit test', () => {
...
+				{
+					onFailure: jasmine.any(Function),
+					onSuccess: jasmine.any(Function),
+				},
+				{ foo: 'bar' }
+			);
+			spyon.mockClear();
+		});
+
+		test('happy case clientMetadata default', async () => {
```
sammartinez: This test compared to the one above seems to be similar? Is this one just for the custom changes? May want to rename the tests themselves

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -2051,6 +2485,42 @@ describe('auth unit test', () => {
 
 			spyon.mockClear();
 		});
+
+		test('happy case clientMetadata default', async () => {
```
sammartinez: same callout as above

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -2051,6 +2485,42 @@ describe('auth unit test', () => {
...
+
+			expect(await CognitoUser.prototype.updateAttributes).toBeCalledWith(
+				[],
+				jasmine.any(Function),
+				{ foo: 'bar' }
+			);
+			spyon.mockClear();
+		});
+
+		test('happy case clientMetadata default', async () => {
```
sammartinez: same callout as above

packages/auth/src/Auth.ts
```diff
@@ -343,23 +354,38 @@ export default class AuthClass {
...
 	/**
 	 * Resend the verification code
 	 * @param {String} username - The username to be confirmed
+	 * @param {ClientMetadata} clientMetadata - Metadata to be passed to Cognito Lambda triggers
 	 * @return - A promise resolves data if success
 	 */
-	public resendSignUp(username: string): Promise<string> {
+	public resendSignUp(
+		username: string,
+		clientMetadata: ClientMetaData = this._config.clientMetadata
```
sammartinez: Do we want this to be an optional value? Or is resendSignUp going to be that it requires a lambda trigger

packages/auth/src/Auth.ts
```diff
@@ -853,7 +882,8 @@ export default class AuthClass {
 	public confirmSignIn(
 		user: CognitoUser | any,
 		code: string,
-		mfaType?: 'SMS_MFA' | 'SOFTWARE_TOKEN_MFA' | null
+		mfaType?: 'SMS_MFA' | 'SOFTWARE_TOKEN_MFA' | null,
+		clientMetadata: ClientMetaData = this._config.clientMetadata
```
sammartinez: same callout as above

packages/auth/__tests__/auth-unit-test.ts
haverchuck: I'm not sure... I've never used expect.any.  Maybe one is a mock and another more of a stub?

packages/auth/__tests__/auth-unit-test.ts
haverchuck: Yes.  It ensures that stuff is still working if clientMetadata isn't present.

packages/auth/__tests__/auth-unit-test.ts
haverchuck: It's different, but I pasted the same name.  Fixed now.

packages/auth/__tests__/auth-unit-test.ts
haverchuck: same error as above.

packages/auth/__tests__/auth-unit-test.ts
haverchuck: same

packages/auth/__tests__/auth-unit-test.ts
sammartinez: yeah true true

PR #4185, opened 2019-10-15
fix(aws-amplify-angular): pass signout options object in auth decorator

LanderN:
_Description of changes:_
aws-amplify-angular does not pass the signout options object to @aws-amplify/auth.
This means it is impossible to perform a global signout using the angular package.
This pull request fixes this.


LanderN:
Can someone take a look at this? Global sign out is still not working without this fix.

Issue #4188, opened 2019-10-16
Can we refresh token sooner


hoang-innomizetech:
** Which Category is your question related to? **
Auth
** What AWS Services are you utilizing? **
E.g. Cognito
** Provide additional details e.g. code snippets **
We are using Amplify for authentication to our Angular app with Cognito User Pool, sometimes our user is getting access token has expired error. We are trying to figure out a solution to allow refresh token easier instead of extract 1h. Let's say if we logged at 10 am then the token is refreshing at 11 am, we saw that **driftLock** stored in the storage is 0.

Is there any way to allow refreshing access token easier i.e. before 10mins?

We are using the latest version of Angular and Amplify

jkeys-ecg-nmsu:
This doesn't directly answer your question, but I believe if you add this line:

```
//before your code, inside an async function
await Auth.currentSession();
```

It will refresh the tokens if they have expired. Then your code can proceed as normal with fresh tokens. I'm sure an Amplify dev will correct me if I'm wrong :P

hoang-innomizetech:
Yeah, we are doing likes this since we are using Angular interceptor to get the access token and add to HTTP request header, but access token has expired error still occur. Also, that error is rarely happening and hard to reproduce. 

Basically, Amplify does have a job to auto refresh token, but right now it seems doesn't refresh ahead of expiration time

jkeys-ecg-nmsu:
Oh I see.

I checked the AuthClass and didn't see a method for forcing a token refresh before the expiry, so the Amplify team will probably have to add a method for that or you'd have to manually send the refresh token to the TOKENS endpoint and grab new tokens, then inject them into a new service client and execute your request. I don't know if you can inject them into the global AWS client directly without interfering with the auto-token refrsh functionality.

You can probably temporarily ameliorate this problem just by increasing the max session time for the authenticated role in your Cognito IdP though.

hoang-innomizetech:
Well, access token and id token valid within 1h and cannot be changed. Your solution might work, but it add complexity, amplify does refresh, so we just need to adjust function to refresh it on time, maybe add a new prop

jkeys-ecg-nmsu:
@hoang-innomizetech are you sure that the max role duration is 1 hour? I think that's the default. I don't know if Amplify supports setting the DurationsSecond parameter for the internal authentication logic. 

@elorzafe does Amp support DurationSeconds, and if not is there any way to monkey patch OP's code to get them unblocked?

[...]

PR #4205, opened 2019-10-18
Refresh credentials on 50 minutes max

elorzafe:
_Issue #, if available:_
#4188

_Description of changes:_
Forcing Cognito credentials to refreshed on 50 minutes max since it was retrieved from the service.


Christilut:
Can this be merged asap? Currently my application is logging users out at seemingly random times and it's even happening once or twice every day during development.

packages/core/__tests__/Credentials-test.ts
```diff
@@ -70,10 +64,12 @@ describe('Credentials test', () => {
 			Amplify.register(cacheClass);
 			const credentials = new Credentials(null);
 
+			const expireTime = new Date().getTime() + 20 * 60 * 1000;
```
ericclemmons: @elorzafe  Should this line be updated to use `DateUtils.getDateWithOffset().getTime()` from #4844?

elorzafe: I think is not necessary because the time is only local, not sent or compared to something external

ericclemmons: Excellent point, since this is just for comparison down below.

PR #4212, opened 2019-10-18
S3Image passes className prop to <img>

### LINKED ISSUE #1272: Passing down className prop in S3Image component
**Feature request**

What is the current behavior?
In _aws-amplify-react_ Passing `className` prop to S3Image components has no effect. The prop is ignored. I want to be able to set CSS class on the S3Image component.

```
<S3Image className="thumbnail" imgKey={props.image} />
```

ericclemmons:
Fixes #1272

`S3Image` supports `className` prop.

I was able to test this via:

```shell
npx lerna exec --scope aws-amplify-react jest -- --watch
```

---


packages/aws-amplify-react/__tests__/Storage/__snapshots__/S3Image-test.tsx.snap
```diff
@@ -22,6 +22,29 @@ exports[`S3Image render test render null if no test and no picker 1`] = `
...
+  style={Object {}}
+>
+  <div
+    onClick={[Function]}
+    style={
+      Object {
+        "position": "relative",
+      }
+    }
+  >
```
ericclemmons: I can't speak to why these wrapping containers exist, but it's quite possible that even with the introduction of ``, users will still have issues with the resulting output.

It would be a backwards-incompatible change, but if we have UI examples to test for regressions, I'd recommend that `<S3Image>` returns only an ``. (And when `picker` is supplied, it returns a fragment)

ericclemmons:
@jordanranz @sammartinez We have some new UI work taking place, but some 👀 would be appreciated to resolve #1272.

packages/aws-amplify-react/src/Storage/S3Image.tsx
```diff
@@ -194,11 +195,12 @@ export default class S3Image extends Component<IS3ImageProps, IS3ImageState> {
 			return null;
 		}
 
-		const { selected } = this.props;
+		const { className, selected } = this.props;
 		const containerStyle: React.CSSProperties = { position: 'relative' };
 		return (
 			<div style={containerStyle} onClick={this.handleClick}>
 				<img
+					className={className}
```
sammartinez: What if there is no `className` prop passed?

ericclemmons: `className={undefined}` removes the attribute from the rendered HTML, effectively making it a no-op.

The other tests that _weren't_ passing `className` have no regressions:
> https://github.com/aws-amplify/amplify-js/blob/1a726e824b8a3c09955cec3f5514545bd57c1ea0/packages/aws-amplify-react/__tests__/Storage/__snapshots__/S3Image-test.tsx.snap

PR #2680, opened 2019-02-08
Include src/ files in npm release (for resolving source maps)

justingrant:
Fixes #2234

*Description of changes:*
Currently it's very hard to debug Amplify apps because no original source is available to the debugger because the source files are stripped out when publishing to npm. Intstead, you need to debug into the transpiled source which (especially for async methods like in the API package!) is incredibly difficult.  

This PR removes `src/**` from `.npmignore` in 13 packages that previously excluded the `src` folder and its subdirectories.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4216, opened 2019-10-19
fix: sourcemaps for CSS, weird paths, /predictions

justingrant:
_Description of changes:_
 
This PR fixes a remaining handful of sourcemap issues remaining after my previous sourcemap-related PRs #3059 and #2680:
* CSS files in UI library - Currently, there's no source maps emitted for the CSS files in the amplify-ui package. This PR updates the amplify-ui package's webpack config to emit sourcemaps.
* CSS file paths - CSS files' sourcemap paths (both existing as well as the new ones added above) were invalid because they were relative to the package root, not to the /dist or /lib subdirectory where the sourcemap files were. This PR fixes these paths by prepending a `.` to these paths, e.g. `./src/foo.css` -> `../src/foo.css`
* Vue - enables source maps for the vue package via a `configureWebpack` setting in its config
* Weird webpath paths - Webpack emits some oddball paths like `webpack/universalModuleDefinition` or `webpack/bootstrap 21c2cca6cf7a65b395b7` or `fs (ignored)`. The logic in #3059 would pre-pend `webpack:///` to these paths which would trigger warnings in some sourcemap-consuming tools like https://github.com/Volune/source-map-loader/tree/fixes. This PR removes that prefix if a path is already a weird webpack path, which removes those warnings.
* relative URLs - The same code that fixes the "weird path" fix above also fixes a problem with #3059 where it would pre-pend `webpack:///` as a fall-through case, even when the path was already a valid relative URL and didn't need that prefix.  Also, according to @loganfsmyth (the most knowledgeable sourcemap expert I know), OSS library sourcemaps should always use relative URLs and should never use the webpack:/// prefix. Therefore, this PR removes that prefix.
* adds support for the new `predictions` library in this repo

### DIFF
--- a/packages/amplify-ui/webpack.config.js
+++ b/packages/amplify-ui/webpack.config.js
@@ -33,6 +33,7 @@ module.exports = {
 					{
 						loader: MiniCssExtractPlugin.loader,
 						options: {
+							sourceMap: true,
 							hmr: process.env.NODE_ENV === 'development',
 						},
 					},
@@ -48,10 +49,16 @@ module.exports = {
 						options: {
 							modules: true,
 							importLoaders: 1,
+							sourceMap: true,
 							localIdentName: '[name]__[local]___[hash:base64:5]',
 						},
 					},
-					'postcss-loader',
+					{
+						loader: 'postcss-loader',
+						options: {
+							sourceMap: true,
+						},
+					},
 				],
 			},
 		],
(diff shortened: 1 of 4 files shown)


justingrant:
I'm not sure why CircleCI builds haven't completed. Let me know if there's something I need to do to unblock them.

amhinson:
Thanks for the work here, and I'm sorry we couldn't get to this sooner 😕 

With that said, we've rewritten our UI components so that they can be shared across all supported frameworks, so these changes are no longer necessary. 

If you haven’t already, please see our migration guide and updated documentation for using `@aws-amplify/ui-vue` in your Vue projects:
> https://docs.amplify.aws/ui/auth/authenticator/q/framework/vue

If there are any features or functionality missing, please open an issue & we’ll get it resolved:
> https://github.com/aws-amplify/amplify-js/issues/new/choose

PR #4227, opened 2019-10-21
fixes #4181 upgrade react-native-vector-icons package to avoid memory…

### LINKED ISSUE #4181: Memory Leak.
**Describe the bug**

aws-amplify-react-native 2.2.2 (current) is using react-native-vector-icons 5.0.0, which is using yargs 8.0.2, which is using os-locale 2.0.0, which is using mem 1.1.0 which has a memory leak. 

Please upgrade to the most recent react-native-vector-icons version to remove this memory leak.

Amplifiyer:
… leak from mem-1.1.0

_Issue #, if available:_ #4181

_Description of changes:_ Upgrade react-native-vector-icons package to avoid memory leak from mem-1.1.0


packages/aws-amplify-react-native/package.json
```diff
@@ -24,7 +24,7 @@
     "babel-preset-es2015": "^6.24.1",
     "buffer": "^5.2.1",
     "react-native-elements": "^0.19.1",
-    "react-native-vector-icons": "^5.0.0"
+    "react-native-vector-icons": "^6.6.0"
```
sammartinez: Does this break anything backwards compatiable? @Ashish5591  Is this fine to do?

Amplifiyer:
> I was trying to find where we use this package, and looks like we don't (at least directly). Looks like it was added when react-native-elements was added
> 
> I think the package should be removed from the dependencies.
> Look at the note at the bottom of:
> https://react-native-elements.github.io/react-native-elements/docs/0.19.1/getting_started.html

You are right, I can't find any usage either. I'll remove the package dependency (somehow I'm not able to view your requested changed and accept them :( )

Ashish5591:
Agree with removing it since its a peer dependency. I think we will need to update our docs to ensure we add the step to install `react-native-vector-icons` though

packages/aws-amplify-react-native/package.json
```diff
@@ -23,8 +23,7 @@
   "dependencies": {
     "babel-preset-es2015": "^6.24.1",
     "buffer": "^5.2.1",
-    "react-native-elements": "^0.19.1",
-    "react-native-vector-icons": "^5.0.0"
+    "react-native-elements": "^0.19.1"
```
sammartinez: [This](https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react-native/src/AmplifyUI.js#L28) is the only location I can find that we are using `react-native-elements`

Amplifiyer:
@sammartinez @Ashish-Nanda @manueliglesias 
👀 on this please

sammartinez:
> @sammartinez @Ashish-Nanda @manueliglesias
> 👀 on this please

@Amplifiyer To the reference I called out that there is one location where we are importing and using the Icon component from react-native-elements. Please see the above.

packages/aws-amplify-react-native/package.json
Amplifiyer: I didn't understand it. We are only removing `react-native-vector-icons` and not `react-native-elements`

sammartinez: Understood, sorry I thought that is what you meant on just verifying this. 😄

Ashish-Nanda:
If we are removing it from dependencies, should we add the following to peer dependencies?:
`'"react-native-vector-icons": "^6.6.0"`

This is because `react-native-vector-icons` is a peer dependency of `react-native-elements` and previously users  did not need to install vector icons themselves since it was included by `aws-amplify-react-native`. 

We should definitely update documentation in the relevant places to add the installation step so that users do not run into issues. This should be done at the same time as merging the PR.

packages/aws-amplify-react-native/package.json
Ashish-Nanda: If we are removing it from dependencies, should we add the following to peer dependencies?:
`'"react-native-vector-icons": "^6.6.0"`

This is because `react-native-vector-icons` is a peer dependency of `react-native-elements` and previously users  did not need to install vector icons themselves since it was included by `aws-amplify-react-native`. 

We should definitely update documentation in the relevant places to add the installation step so that users do not run into issues. This should be done at the same time as merging the PR.

Amplifiyer: Not as per https://react-native-elements.github.io/react-native-elements/docs/0.19.1/getting_started.html as mentioned by @manueliglesias in previous comment.

Amplifiyer: It won't work with RN CLI. Since it's already a peer dependency of react-native-elements, I don't think we need to specify another peer dependency of aws-amplify-react-native. I'll update the docs to accompany this PR.

sammartinez: @Amplifiyer On the non expo tab, [React Native CLI](https://react-native-elements.github.io/react-native-elements/docs/0.19.1/getting_started.html), there is a callout to install [vectors](https://react-native-elements.github.io/react-native-elements/docs/0.19.1/getting_started.html).

Ashish-Nanda: We need it in the case of React Native CLI but not expo. Check instructions under the React Native CLI tab.
https://react-native-elements.github.io/react-native-elements/docs/0.19.1/getting_started.html#step-2-install-react-native-vector-icons

Amplifiyer: Talked with @Ashish-Nanda offline. Adding the `react-native-vector-icons` dependency back to avoid any breaking changes (i.e. having users to install an additional package).

PR #4229, opened 2019-10-21
fixes #3861 : improper promise resolution in TotpSetupComp

### LINKED ISSUE #3861: Asynchronous Auth.setPreferredMFA in verifyTotpToken
**Describe the bug**
Here the original code in aws-amplify-eact
```
verifyTotpToken() {
        if (!this.inputs) {
            logger.debug('no input');
            return;
        }
        const user = this.props.authData;
        const { totpCode } = this.inputs;
        if (!Auth || typeof Auth.verifyTotpToken !== 'function' || typeof Auth.setPreferredMFA !== 'function') {
            throw new Error('No Auth module found, please ensure @aws-amplify/auth is imported');
        }
        Auth.verifyTotpToken(user, totpCode)
            .then(() => {
                // set it to preferred mfa
                Auth.setPreferredMFA(user, 'TOTP');
                this.setState({setupMessage: 'Setup TOTP successfully!'});
                logger.debug('set up totp success!');
                this.triggerTOTPEvent('Setup TOTP', 'SUCCESS', user);
            })
            .catch(err => {
                this.setState({setupMessage: 'Setup TOTP failed!'});
                logger.error(err);
            });
    }
```
It's obvious that `Auth.setPreferredMFA(user, 'TOTP');` did not call synchronously, then it will change state immediately after the call and not care about the result even errors.

**To Reproduce**
I wrote my own `withAuthenticator` function and used TOTP.
```
export function withAuthenticator(RenderedComponent) {
    handleAuthStateChange = (state, data) => {
        Auth.getPreferredMFA(data, { bypassCache: true }).then(mfa => {
            console.log(mfa) // NOMFA because setPrefferedMFA not effect yet
        }.catch(console.log)
    }
}
```

**Expected behavior**
The `verifyTotpToken` should be changed to synchronous by waiting for `setPreferredMFA` returning the result.

**Screenshots**

 - OS: MacOS
 - Browser: Chrome
 - Version any

 - Device: any
 - OS: any
 - Browser any
 - Version: any

**Additional context**
None.

**Sample code**
```
verifyTotpToken() {
        if (!this.inputs) {
            logger.debug('no input');
            return;
        }
        const user = this.props.authData;
        const { totpCode } = this.inputs;
        if (!Auth || typeof Auth.verifyTotpToken !== 'function' || typeof Auth.setPreferredMFA !== 'function') {
            throw new Error('No Auth module found, please ensure @aws-amplify/auth is imported');
        }
        Auth.verifyTotpToken(user, totpCode)
            .then(() => {
                // set it to preferred mfa
                return Auth.setPreferredMFA(user, 'TOTP');
           }).then(()=>{
                this.setState({setupMessage: 'Setup TOTP successfully!'});
                logger.debug('set up totp success!');
                this.triggerTOTPEvent('Setup TOTP', 'SUCCESS', user);
           }) .catch(err => {
                this.setState({setupMessage: 'Setup TOTP failed!'});
                logger.error(err);
            });
    }
`
```

***You can turn on the debug mode to provide more info for us by setting window.LOG_LEVEL = 'DEBUG'; in your app.***

ashika01:
_Issue #, if available:_ #3861 

_Description of changes:_ This  PR resolves the promise properly.


packages/aws-amplify-react/src/Widget/TOTPSetupComp.tsx
```diff
@@ -131,10 +131,11 @@ export default class TOTPSetupComp extends Component<
 		Auth.verifyTotpToken(user, totpCode)
 			.then(() => {
 				// set it to preferred mfa
-				Auth.setPreferredMFA(user, 'TOTP');
-				this.setState({ setupMessage: 'Setup TOTP successfully!' });
-				logger.debug('set up totp success!');
-				this.triggerTOTPEvent('Setup TOTP', 'SUCCESS', user);
+				Auth.setPreferredMFA(user, 'TOTP').then(() => {
```
sammartinez: NIT: Do you believe it would be worth wild to make this a try catch?

packages/aws-amplify-react/src/Widget/TOTPSetupComp.tsx
```diff
@@ -131,10 +131,11 @@ export default class TOTPSetupComp extends Component<
...
+					this.setState({ setupMessage: 'Setup TOTP successfully!' });
+					logger.debug('set up totp success!');
+					this.triggerTOTPEvent('Setup TOTP', 'SUCCESS', user);
+				});
```
elorzafe: In case `Auth.setPreferredMFA` is rejected is not going to be catched. it requieres a catch here
```suggestion
				}).catch(err => this.setState({ setupMessage: 'Setup TOTP failed!' }));
```

packages/aws-amplify-react/__tests__/Widget/TOTPSetupComp-test.tsx
```diff
@@ -143,14 +143,16 @@ describe('TOTPSetupComp test', () => {
 			const spyon2 = jest
 				.spyOn(Auth, 'setPreferredMFA')
 				.mockImplementationOnce(() => {
-					return;
+					return new Promise((res, rej) => {
+						res();
+					});
```
Amplifiyer: Can just be return Promise.resolve();

packages/aws-amplify-react/__tests__/Widget/TOTPSetupComp-test.tsx
```diff
@@ -143,14 +143,16 @@ describe('TOTPSetupComp test', () => {
...
 				});
 			const spyon3 = jest.spyOn(instance, 'triggerTOTPEvent');
 
 			await instance.verifyTotpToken();
 
 			expect(spyon).toBeCalled();
-			expect(spyon2).toBeCalled();
+			await expect(spyon2).toBeCalled();
```
Amplifiyer: I thought we can only use await for expect...rejects or expect...resolves. I don't think it's needed here since toBeCalled() is probably not an async method.

packages/aws-amplify-react/src/Widget/TOTPSetupComp.tsx
Amplifiyer: It is wild! but +1 for async/await and the one enclosing it as well.

Amplifiyer: > NIT: Do you believe it would be worth wild to make this a try catch?

It is wild!

packages/aws-amplify-react/src/Widget/TOTPSetupComp.tsx
sammartinez: I dont believe this is missing correct? Cause the below code has the `.catch`

packages/aws-amplify-react/__tests__/Widget/TOTPSetupComp-test.tsx
ashika01: await here is for the spyon2 to resolve. need to synchronize that for checking the spyon3 being called.

packages/aws-amplify-react/__tests__/Widget/TOTPSetupComp-test.tsx
ashika01: That would work too. Just going with this for consistency.

packages/aws-amplify-react/src/Widget/TOTPSetupComp.tsx
ashika01: We need to catch that separately.

packages/aws-amplify-react/__tests__/Widget/TOTPSetupComp-test.tsx
ashika01: Rewrote the tests. Let me know if it looks good :)

ashika01:
@Amplifiyer @elorzafe This PR is ready to review. 
One major issue breaking different test implementation was using `mockClear` which only clears the mock info and that caused leaks. Changing it to `mockRestore` solved the leak as they force restore to original implementation.

Issue #3558, opened 2019-06-29
The "Create Account" button is not disabled while a request is pending


all-iver:
**Describe the bug**
During sign up, the "Create Account" button is not disabled while a request is pending.  Additionally there is no indication that it is doing anything.  In particular, if there is a lambda trigger involved then the request may take several seconds.  This leads to the user frustratedly pressing the button multiple times and results in a new Cognito user for every button press.

**To Reproduce**
1. Make a react app using withAuthenticator() and try to create a new user.
2. Quickly press the "Create Account" a bunch of times.
3. Observe that multiple new users have been created in your User Pool.

**Expected behavior**
The "Create Account" button should disable while a request is pending.

haverchuck:
@all-iver We have a PR open which fixes this in aws-amplify-react.  We will merge it as soon as we are able.

BardiaN:
This issue is back or it hasn't been resolved completely.
So I have a pre signup lambda that checks for a secret and the button stays enabled during the lambda runtime, so user is able to press the button so many times. (And Cognito has a known bug that let the same user being created if requests are too close to each other)
@all-iver Was your problem fixed?

all-iver:
> 
> 
> This issue is back or it hasn't been resolved completely.
> So I have a pre signup lambda that checks for a secret and the button stays enabled during the lambda runtime, so user is able to press the button so many times. (And Cognito has a known bug that let the same user being created if requests are too close to each other)
> @all-iver Was your problem fixed?

I just tried to replicate and it seems to be fixed for me, but I haven't updated in quite a while so there could have been a regression.  I'm still on aws-amplify-react 4.1.7.

BardiaN:
> > This issue is back or it hasn't been resolved completely.
> > So I have a pre signup lambda that checks for a secret and the button stays enabled during the lambda runtime, so user is able to press the button so many times. (And Cognito has a known bug that let the same user being created if requests are too close to each other)
> > @all-iver Was your problem fixed?
> 
> I just tried to replicate and it seems to be fixed for me, but I haven't updated in quite a while so there could have been a regression. I'm still on aws-amplify-react 4.1.7.

Thanks for the reply @all-iver 
So I tested my scenario and I guess I found my problem, but I think I have to open a new issue for it. It's related to the user being in `Unconfirmed` status and it happens because it takes a while for my user to be created due to existence of the PreSignUp trigger. 

[...]

PR #4247, opened 2019-10-24
fix(aws-amplify-react) disables signup button while request is in flight

haverchuck:
_Issue #, if available:_
Fixes: #3558 

_Description of changes:_
 Disables signup button while request is in flight by using a requestPending property in state.  It is implemented at the AuthPiece level, so should be reusable in other Auth components (but making this PR just for signup since the bug is more than just a nuissance but actually results in multiple user pool users).

To test, trying creating a pre-signup lambda trigger with a timeout wrapping the response.


packages/aws-amplify-react/__tests__/Auth/SignUp-test.tsx
```diff
@@ -125,6 +125,39 @@ describe('signUp without signUpConfig prop', () => {
 			spyon_changeState.mockClear();
 		});
 
+		test('state.requestPending shoudl be true when signUp execution begins', () => {
```
Amplifiyer: Nit: typo

haverchuck: fixed :)

haverchuck:
@sammartinez - I created the tasks and wrote some tests

packages/aws-amplify-react/src/Auth/SignUp.tsx
```diff
@@ -327,6 +333,7 @@ export default class SignUp extends AuthPiece<ISignUpProps, IAuthPieceState> {
 				<SectionFooter theme={theme} data-test={auth.signUp.footerSection}>
 					<SectionFooterPrimaryContent theme={theme}>
 						<Button
+							disabled={this.state.requestPending}
```
ashika01: Just a NIT: we could destructure the state like we have done with props :)

PR #4248, opened 2019-10-24
feat(@aws-amplify/ui-components): Add inputProps and update Stencil

jordanranz:
_Description of changes:_
- Add `inputProps` to components using `input`
- Update Stencil
 

packages/amplify-ui-components/src/components/amplify-code-field/amplify-code-field.tsx
```diff
@@ -17,7 +17,9 @@ export class AmplifyCodeField {
   /** The callback, called when the input is modified by the user. */
   @Prop() handleInputChange?: (inputEvent: Event) => void;
   /** The value of the content inside of the input field */
-  @Prop() value: string;
+  @Prop() value?: string;
+  /** Attributes places on the input element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Attributes */
+  @Prop() inputProps?: object;
```
ericclemmons: This works for via JSX, but how will customers or our `.spec` tests pass this?

(See test below)

packages/amplify-ui-components/src/components/amplify-input/__snapshots__/amplify-input.spec.ts.snap
```diff
@@ -12,6 +12,12 @@ exports[`amplify-input spec: Render logic -> renders with a name for the input 1
 </amplify-input>
 `;
 
+exports[`amplify-input spec: Render logic -> renders with inputProps spread 1`] = `
+<amplify-input inputprops="[object Object]">
```
ericclemmons: `[object Object]` isn't desirable, right?  Vanilla HTML only supports string attributes, so this `inputprops` would need to be provided differently...?

jordanranz: Good callout here. The input props should properly spread on the `<input>` element

jordanranz: Also, I believe the prop should not be camel case from the perspective of the customer and spec.

packages/amplify-ui-components/src/components/amplify-code-field/amplify-code-field.tsx
jordanranz: This can be added programmatically in both the `.spec` and when customers are using this.

packages/amplify-ui-components/src/components/amplify-input/__snapshots__/amplify-input.spec.ts.snap
jordanranz: Fixed this by programmatically adding the complex prop

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -41,6 +43,7 @@ export class AmplifyInput {
         name={this.name}
         class={styleNuker(this.overrideStyle, STATIC_INPUT_CLASS_NAME, input)}
         value={this.value}
+        {...this.inputProps}
         disabled={this.disabled}
```
ericclemmons: Note that if the the user attempts to set `inputProps.disabled = true`, we're overriding it explicitly.

Could be a gotcha down the road, but there's not a valid scenario I can think of off-hand that we'd see.

packages/amplify-ui-components/src/components/amplify-input/__snapshots__/amplify-input.spec.ts.snap
```diff
@@ -12,6 +12,12 @@ exports[`amplify-input spec: Render logic -> renders with a name for the input 1
 </amplify-input>
 `;
 
+exports[`amplify-input spec: Render logic -> renders with complex prop: inputProps spread on the input element 1`] = `
+<amplify-input>
+  <input class="amplify-ui--input css-g9dmxw" placeholder="" required="" type="text">
```
sammartinez: can we double check the `disabled` field for this. Seems like it may be missing from the snapshot, and I am seeing the test as having it.

jordanranz: Updated this with new 'complex' props

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
jordanranz: I'll change the order for consistency

PR #4249, opened 2019-10-24
feat(@aws-amplify/ui-components): amplify-federated-sign-in + a…

ericclemmons:
[Delivers #168413082]

This migrates https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react/src/Auth/FederatedSignIn.tsx to a Stencil component.

The process is re-using the tests in https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react/__tests__/Auth/FederatedSignIn-test.tsx, capturing the HTML output, converting them to stories, and validating the Stencil version outputs visually similar output to the React version.

- [x] Fix HMR experience with Storybook.
- [x] Address rendering React components within Storybook.
- [x] Snapshot `FederatedButtons` rendered component.
- [x] Snapshot `FederatedSignIn` rendered component.
- [x] Convert https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react/__tests__/Auth/FederatedSignIn-test.tsx to stories.
- [x] Create matching `amplify-federated-sign-in` component stories.
- [x] Create matching `amplify-federated-buttons` component stories.
- [x] Create Google/Facebook/etc. Provider buttons.
- [x] Migrate styling to Stencil components.
- [x] Create spec test (ideally re-use story).
- [x] ~Create e2e test.~

> Screen Shot 2019-10-24 at 11 53 51 AM


packages/amplify-ui-components/package.json
```diff
@@ -23,25 +23,27 @@
     "stencil": "stencil build --docs-readme",
     "stencil:watch": "stencil build --dev --watch",
     "start": "stencil build --dev --watch --serve",
-    "storybook": "start-storybook -s ./www",
+    "storybook": "concurrently 'start-storybook -s ./www' 'yarn:stencil:watch' --raw",
```
jordanranz: nice improvement

ericclemmons:
Trying to figure out why `yarn build` (via stencil) is returning a TypeScript error:

```
[ ERROR ]  TypeScript: ../../node_modules/@types/react/index.d.ts:107:31
           Type 'T' does not satisfy the constraint 'string | ((props: any) => ReactElement<any, string | ... | (new
           (props: any) => Component<any, any, any>)>) | (new (props: any) => Component<any, any, any>)'.Type 'string |
           number | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>)
           | (new (props: any) => Component<any, any, any>)' is not assignable to type 'string | ((props: any) =>
           ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>) | (new (props: any) =>
           Component<any, any, any>)'.Type 'number' is not assignable to type 'string | ((props: any) =>
           ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>) | (new (props: any) =>
           Component<any, any, any>)'.Type 'T' is not assignable to type 'string'.Type 'string | number | ((props: any)
           => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>) | (new (props: any) =>
           Component<any, any, any>)' is not assignable to type 'string'.Type 'number' is not assignable to type
           'string'.

    L106:          P = Pick<ComponentProps<T>, Exclude<keyof ComponentProps<T>, 'key' | 'ref'>>
    L107:      > extends ReactElement<P, T> { }
```

If I were to guess, it's due to TypeScript or Stencil running against stories, which includes `jsx2dom` and a custom JSX definition.

Ideally, the stories could use `h` from `@stencil/core`, but rendering that is a bit tough right now.

packages/aws-amplify-react/src/Auth/FederatedSignIn.tsx
```diff
@@ -27,11 +27,14 @@ export interface IFederatedButtonsProps {
...
+	theme?: any;
 }
 
-export class FederatedButtons extends React.Component<IFederatedButtonsProps, {}> {
-		google(google_client_id) {
+export class FederatedButtons extends React.Component<
+	IFederatedButtonsProps,
+	{}
+> {
+	google(google_client_id) {
```
ericclemmons: prettier

packages/aws-amplify-react/src/Auth/FederatedSignIn.tsx
```diff
@@ -27,11 +27,14 @@ export interface IFederatedButtonsProps {
 	federated: any;
 	onAuthEvent?: any;
 	onStateChange: any;
-	theme: any;
+	theme?: any;
```
ericclemmons: `FederatedSignIn` defaults to `AmplifyTheme`, so it's not required.

packages/aws-amplify-react/__tests__/Auth/__snapshots__/FederatedSignIn-test.tsx.snap
```diff
@@ -122,17 +219,72 @@ exports[`FederatedButtons test render test render with correct authState and onl
...
+  </div>
+</div>
+`;
+
+exports[`FederatedButtons test render test render with correct authState and only google id 1`] = `
+<div>
   <div>
     <class_1
-      facebook_app_id="facebook_app_id"
+      google_client_id="google_client_id"
```
ericclemmons: Correcting a typo from the previous test.

packages/auth/src/types/Auth.ts
```diff
@@ -117,6 +117,7 @@ export interface FederatedResponse {
 export interface FederatedUser {
 	name: string;
 	email?: string;
+	picture?: string;
```
ericclemmons: Google provides an image URL for this value (no other services do), but it's being passed nonetheless.

Adding it to the interface, since TypeScript was giving me errors otherwise.

packages/aws-amplify-react/__tests__/Auth/FederatedSignIn-test.tsx
```diff
@@ -22,6 +22,7 @@ describe('FederatedSignIn test', () => {
 				onStateChange: jest.fn(),
 			});
 			expect(wrapper).toMatchSnapshot();
+			expect(render(wrapper)).toMatchSnapshot();
```
ericclemmons: Rendering the HTML (instead of the React component tree) helps with matching parity between React & Stencil.

packages/amplify-ui-components/package.json
ericclemmons: Thanks!  This has been pulled out into #4347 for others to benefit from.

packages/amplify-ui-components/.storybook/stencil-jsx.js
```diff
@@ -0,0 +1,45 @@
...
+ * convert JSX to a DOM node with Array/Object/Function support.
+ *
+ * (Other `h` utilities create vDOM trees.)
+ *
+ * @param tag string
+ * @param props object | null
+ * @param children array
+ */
+
+export const h = (tag, props, ...children) => {
```
ericclemmons: This utility makes it possible to re-use stories' JSX within spec tests.

`import { h} from "@stencil/core"` is actually just a no-op (`function() {}`), unless run through Stencil's build process.

packages/amplify-ui-components/.storybook/webpack.config.js
```diff
@@ -5,19 +5,23 @@ module.exports = ({ config, mode }) => {
...
     test: /\.(ts|tsx)$/,
     include: path.resolve(__dirname, '../src'),
-    loader: require.resolve('ts-loader')
+    loader: require.resolve('ts-loader'),
   });
+
   config.resolve.extensions.push('.ts', '.tsx');
 
+  // Stencil requires a build step, stories using JSX will create DOM nodes instead.
+  config.resolve.alias['@stencil/core$'] = require.resolve('./stencil-jsx');
```
ericclemmons: `import { h } from "@stencil/core"` returns a custom `h` that converts JSX into a valid DOM element.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.stories.tsx
```diff
@@ -1,7 +1,23 @@
...
+export default {
+  title: 'amplify-authenticator',
+};
 
-authenticatorStories.add('default', () => {
-  return `<amplify-authenticator></amplify-authenticator>`;
-});
+export const defaults = () => <amplify-authenticator />;
+
+export const withFederated = () => (
```
ericclemmons: I converted this story into the recommended CSF format to add this federated story.

> https://storybook.js.org/docs/formats/component-story-format/

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.stories.tsx
```diff
@@ -1,7 +1,23 @@
...
+  <amplify-authenticator
+    federated={{
+      amazon_client_id: knobs.text('Amazon client ID', 'amazon_client_id'),
+      auth0: {
+        clientID: knobs.text('Auth0 client ID', 'auth0_client_id'),
+        domain: knobs.text('Auth0 account domain', 'example.auth0.com'),
+      },
+      facebook_app_id: knobs.text('Facebook app ID', 'facebook_app_id'),
```
ericclemmons: I tested this with values from my sample's `amplify/team-provider-info.json` file.

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.spec.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const components = [AmplifyAmazonButton];
+
+describe(title, () => {
+  describe('stories', () => {
+    Object.entries(specs).forEach(([name, spec]) => {
+      it(name, async () => {
+        const page = await newSpecPage({ components, html: spec() });
+
+        expect(page.root).toMatchSnapshot();
```
ericclemmons: There is some repetition with re-using stories for structural testing (i.e. testing DOM structure), which could be abstracted into an internal/OSS package like so:

```js
import { toSpecTests } from 'storybook-stencil';
import * as stories from './amplify-amazon-button.stories'; 

describe('amplify-amazon-button stories', () => {
  toSpecTests(stories)
});
```

packages/amplify-ui-components/src/components/amplify-federated-buttons/amplify-federated-buttons.spec.tsx
```diff
@@ -0,0 +1,23 @@
...
+  ...templates
+} = stories;
+
+const components = [AmplifyFederatedButtons];
+
+describe(title, () => {
+  describe('stories', () => {
+    Object.entries(templates).forEach(([name, template]) => {
+      it(name, async () => {
+        const page = await newSpecPage({ components, template });
```
ericclemmons: This is the best format, IMO, as it allows for providing a `template` (with JSX) compared to a static `html` string.

This would allow _every_ story to use JSX, which is much easier compared to manual DOM manipulation.

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import * as classNames from './amplify-sign-in-button.style';
+
+@Component({ tag: 'amplify-sign-in-button' })
+export class AmplifySignInButton {
+  @Prop() provider: 'amazon' | 'auth0' | 'facebook' | 'google' | 'oauth';
+
+  render() {
+    return (
+      <Host class={cx(this.provider, classNames.Host)}>
```
ericclemmons: I initially had multiple exports from `classNames`, but this ended up looking simpler in `.style`.

If we're concerned about CSS specificity and preventing on-page `.icon` or `.content` classes affecting the appearance, then it'd make sense to let Emotion generate the classes here.

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
```diff
@@ -0,0 +1,98 @@
...
+export class AmplifyAmazonButton {
+  /** App-specific client ID from Google */
+  @Prop() amazon_client_id: string;
+  /** Passed from the Authenticatior component in order to change Authentication state
+   * e.g. SignIn -> 'Create Account' link -> SignUp
+   */
+  @Prop() handleAuthStateChange: (nextAuthState: AuthState, data?: object) => void;
+
+  constructor() {
+    this.handleClick = this.handleClick.bind(this);
```
jordanranz: Is this required?

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
```diff
@@ -0,0 +1,98 @@
...
+      const authenticatedUser = await Auth.currentAuthenticatedUser();
+
+      this.handleAuthStateChange(AuthState.SignedIn, authenticatedUser);
+    });
+  };
+
+  /**
+   * @see https://developer.amazon.com/docs/login-with-amazon/install-sdk-javascript.html
+   */
+  @Listen('click')
```
jordanranz: Curious why this was preferred over using the onClick functionality on the `amplify-sign-in-button`?

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
```diff
@@ -0,0 +1,98 @@
...
+              id="Fill-6"
+              fill="#FFF"
+            />
+          </g>
+        </svg>
+
+        {I18n.get('Sign In with Amazon')}
```
jordanranz: Let's go ahead and move this to `constants` for now. I think we will eventually migrate all string in constants to utilize I18n in a centralized place so we are importing i18n in every component.

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
ericclemmons: Yep, `handleClick` calls `this.federatedSignIn(response)`, and decorators cannot be used around property member declarations:

```js
@Listen('click')
handleClick = async () => {...}
// ! Invalid syntax
```

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
ericclemmons: Could you clarify? `amplify-sign-in-button` doesn't have an `onClick` prop.

Stencil encourages the use of events, which has reduced prop-drilling and noisy child props:
> https://stenciljs.com/docs/events

packages/amplify-ui-components/src/common/constants.ts
```diff
@@ -47,6 +49,11 @@ export const CREATE_ACCOUNT_TEXT = 'Create account';
 export const NO_ACCOUNT_TEXT = 'No account?';
 export const CONFIRM_SMS_CODE = 'Confirm SMS Code';
 export const CONFIRM_TOTP_CODE = 'Confirm TOTP Code';
+export const SIGN_IN_WITH_AMAZON = 'Sign In with Amazon';
+export const SIGN_IN_WITH_AUTH0 = 'Sign In with Auth0';
+export const SIGN_IN_WITH_FACEBOOK = 'Sign In with Facebook';
+export const SIGN_IN_WITH_GOOGLE = 'Sign In with Google';
+export const SIGN_IN_WITH_OAUTH = 'Sign in with AWS';
```
jordanranz: Can we set this to `SIGN_IN_WITH_AWS` so we don't confusing the constant for a string when doing translations

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
```diff
@@ -0,0 +1,25 @@
+import { Component, h, Prop, Host } from '@stencil/core';
+import { cx } from 'emotion';
+
+import * as classNames from './amplify-sign-in-button.style';
+
+@Component({ tag: 'amplify-sign-in-button' })
```
jordanranz: We should add `shadow: false` here for consistency.

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
```diff
@@ -0,0 +1,25 @@
...
+export class AmplifySignInButton {
+  @Prop() provider: 'amazon' | 'auth0' | 'facebook' | 'google' | 'oauth';
+
+  render() {
+    return (
+      <Host class={cx(this.provider, classNames.Host)}>
+        <button>
+          <span class="icon">
```
jordanranz: Can we use `amplify-icon` here and add the federated icons to the list of icons we support.

packages/amplify-ui-components/src/components/amplify-google-button/amplify-google-button.tsx
```diff
@@ -0,0 +1,116 @@
...
+      {
+        email: profile.getEmail(),
+        name: profile.getName(),
+        picture: profile.getImageUrl(),
+      },
+    );
+
+    const authenticatedUser = await Auth.currentAuthenticatedUser();
+
+    this.handleAuthStateChange(AuthState.SignedIn, authenticatedUser);
```
jordanranz: Can we add a try/catch here

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
jordanranz: Looks like the `amplify-sign-in-button` uses an `onClick` function in the `amplify-google-button`. 

I don't think there would need to be much prop drilling here. It would look like:
`<amplify-sign-in-button onClick={event => this.signInWithAmazon(event)} provider="amazon">`
Where `signInWithAmazon` is the function below (instead of handle click). I think this is a bit more readable.

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
```diff
@@ -0,0 +1,98 @@
...
+              id="Fill-6"
+              fill="#FFF"
+            />
+          </g>
+        </svg>
+
+        {I18n.get(SIGN_IN_WITH_AMAZON)}
```
jordanranz: I think `I18n` can also be removed here (and in the other parts of this PR) as the plan is to import and use it in a single file.

packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
```diff
@@ -0,0 +1,36 @@
+import { I18n } from '@aws-amplify/core';
+import { Component, h, Listen, Prop } from '@stencil/core';
+
+import { SIGN_IN_WITH_AUTH0 } from '../../common/constants';
+
+@Component({ tag: 'amplify-auth0-button' })
+export class AmplifyAuth0Button {
+  /** See: https://auth0.com/docs/libraries/auth0js/v9#available-parameters */
+  @Prop() auth0: any;
```
jordanranz: Should this be a `string` type?

jordanranz: I see this prop is not being used. Curious if this is needed?

packages/amplify-ui-components/src/components/amplify-facebook-button/amplify-facebook-button.tsx
```diff
@@ -0,0 +1,119 @@
...
+  };
+
+  /**
+   * @see https://developers.facebook.com/docs/javascript/reference/FB.init/v5.0
+   */
+  @Listen('click')
+  handleClick(event) {
+    event.preventDefault();
+
+    // Initialize before usage for latest knob value
```
jordanranz: Is this comment still relevant?

packages/amplify-ui-components/src/components/amplify-federated-sign-in/amplify-federated-sign-in.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+@Component({
+  tag: 'amplify-federated-sign-in',
+  shadow: false,
+})
+export class AmplifyFederatedSignIn {
+  /** The current authentication state. */
+  @Prop() authState: 'signIn' | 'signedOut' | 'signedUp' = 'signIn';
+  /** Federated credentials & configuration. */
+  @Prop() federated: any = {};
```
jordanranz: Can we defined an interface for this

packages/amplify-ui-components/src/components/amplify-federated-buttons/amplify-federated-buttons.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { Component, h, Prop } from '@stencil/core';
+
+import { AuthState } from '../../common/types/auth-types';
+
+@Component({ tag: 'amplify-federated-buttons' })
+export class AmplifyFederatedButtons {
+  /** The current authentication state. */
+  @Prop() authState: 'signIn' | 'signedOut' | 'signedUp' = 'signIn';
+  /** Federated credentials & configuration. */
+  @Prop() federated: any = {};
```
jordanranz: Can we defined an interface for this?

packages/amplify-ui-components/src/components/amplify-federated-sign-in/amplify-federated-sign-in.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const logger = new Logger('amplify-federated-sign-in');
+
+@Component({
+  tag: 'amplify-federated-sign-in',
+  shadow: false,
+})
+export class AmplifyFederatedSignIn {
+  /** The current authentication state. */
+  @Prop() authState: 'signIn' | 'signedOut' | 'signedUp' = 'signIn';
```
jordanranz: There is an AuthState enum you can use here:
https://github.com/aws-amplify/amplify-js/blob/ui-components/authenticator/packages/amplify-ui-components/src/common/types/auth-types.ts#L3

jordanranz:
Looks like the Sign in with Amazon button's svg is a little broken:
Screen Shot 2019-11-07 at 11 31 25 AM

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
jordanranz: Just tested this on your branch and it seems to work^

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
jordanranz: Curious why `Host` is being used here?

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.style.ts
```diff
@@ -0,0 +1,96 @@
...
+    padding: 10px 14px;
+  }
+
+  &.google .icon {
+    background-color: var(--white);
+    border-radius: 4px 0 0 4px;
+    height: 28px;
+    width: 28px;
+    padding: 12px;
+  }
```
jordanranz: Styling specific to the provider here can be probably be moved to the individual button components `.style` files

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
```diff
@@ -0,0 +1,98 @@
+import { Auth } from '@aws-amplify/auth';
+import { ConsoleLogger as Logger, I18n } from '@aws-amplify/core';
+import { Component, h, Listen, Prop } from '@stencil/core';
+
+import { AUTH_SOURCE_KEY, SIGN_IN_WITH_AMAZON } from '../../common/constants';
+import { AuthState } from '../../common/types/auth-types';
+
+const logger = new Logger('amplify-amazon-button');
+
+@Component({ tag: 'amplify-amazon-button' })
```
sammartinez: I remember that you brought it up on my phone field PR but we do want to be explicit that we don't want shadow dom on `shadow: false`. Can we look to adjust this for all components.

packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
sammartinez: Do we know if there is an interface available we can use for this?

packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
```diff
@@ -0,0 +1,36 @@
...
+  handleClick(event) {
+    event.preventDefault();
+  }
+
+  render() {
+    return (
+      <amplify-sign-in-button provider="auth0">
+        <script src="https://cdn.auth0.com/js/auth0/9.11/auth0.min.js"></script>
+
+        <svg slot="icon" id="artwork" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 193.7 216.6">
```
sammartinez: Could you utilize [amplify-icon](https://github.com/aws-amplify/amplify-js/tree/ui-components/authenticator/packages/amplify-ui-components/src/components/amplify-icon) here?

packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  @Listen('click')
+  handleClick(event) {
+    event.preventDefault();
+  }
+
+  render() {
+    return (
+      <amplify-sign-in-button provider="auth0">
+        <script src="https://cdn.auth0.com/js/auth0/9.11/auth0.min.js"></script>
```
sammartinez: Do we really need to have a script tag here for this?

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -17,6 +17,8 @@ export class AmplifyAuthenticator {
   @State() authState: AuthState = AuthState.Loading;
 
   @State() authData: CognitoUserInterface;
+  /** Federated credentials & configuration. */
+  @Prop() federated: any = {};
```
sammartinez: Do we know the shape by chance? I want to avoid using `any` if we can.

packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  @Listen('click')
+  handleClick(event) {
+    event.preventDefault();
+  }
+
+  render() {
+    return (
+      <amplify-sign-in-button provider="auth0">
```
sammartinez: For the `provider` prop, can we make these constants?

packages/amplify-ui-components/src/components/amplify-federated-buttons/amplify-federated-buttons.tsx
```diff
@@ -0,0 +1,87 @@
...
+          </div>
+        )}
+
+        {auth0 && (
+          <div>
+            <amplify-auth0-button auth0={auth0} />
+          </div>
+        )}
+
+        <amplify-strike>{I18n.get('or')}</amplify-strike>
```
jordanranz: The `strike` will show up if the user adds a `federated` object but leaves the content empty or have a client_id whose value is undefined.

packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
ericclemmons: This implementation hasn't been completed yet, as to not hold up the PR further.

(The React implementation has it as a prop as well)

packages/amplify-ui-components/src/components/amplify-facebook-button/amplify-facebook-button.tsx
ericclemmons: It is.  The rationale is that I originally initialized when the component mounted, but the social providers don't allow changing the client IDs _after_ they've been initialized.

Initializing upon use allowed for testing with storybook for validation.

Is there a better place to put it?  I want to avoid any changes in the future breaking the behavior because there wasn't a comment indicating why it is this way...

packages/amplify-ui-components/src/components/amplify-federated-buttons/amplify-federated-buttons.tsx
ericclemmons: It isn't typed since I'm unsure what the fully supported structure is yet, especially with Auth0 not being fully implemented:
> https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react/src/Auth/Authenticator.tsx#L47

At the very least, I could type `{ google_client_id, ... }` since those seem well defined, but I haven't found a source of truth for the full structure.

Do you know of it?

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
```diff
@@ -0,0 +1,98 @@
...
+export class AmplifyAmazonButton {
+  /** App-specific client ID from Google */
+  @Prop() amazon_client_id: string;
```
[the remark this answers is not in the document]
ericclemmons: I was keeping it consistent with the existing implementation, but I'd like to normalize all of these to `clientID`, since `<amplify-amazon-button amazonClientID />` is redundant.

packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
ericclemmons: Is there a reason not to?  It worked out of the box and seemed so much simpler than manual script tag creation (which has been a week spot for React since the beginning).

packages/amplify-ui-components/src/components/amplify-federated-buttons/amplify-federated-buttons.tsx
ericclemmons: That's today's behavior & come over as part of the parity tests, but I'll fix it here. 👍

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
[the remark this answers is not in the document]
jordanranz: in the context of `amplify-amazon-button` this can just be `clientId`

packages/amplify-ui-components/src/common/constants.ts
```diff
@@ -56,6 +58,11 @@ export const CREATE_ACCOUNT_TEXT = 'Create account';
 export const NO_ACCOUNT_TEXT = 'No account?';
 export const CONFIRM_SMS_CODE = 'Confirm SMS Code';
 export const CONFIRM_TOTP_CODE = 'Confirm TOTP Code';
+export const SIGN_IN_WITH_AMAZON = 'Sign In with Amazon';
+export const SIGN_IN_WITH_AUTH0 = 'Sign In with Auth0';
+export const SIGN_IN_WITH_FACEBOOK = 'Sign In with Facebook';
+export const SIGN_IN_WITH_GOOGLE = 'Sign In with Google';
+export const SIGN_IN_WITH_AWS = 'Sign in with AWS';
```
ericclemmons: @jordanranz This has been fixed, but now I remember why I called it `SIGN_IN_WITH_OAUTH`: because the config value is `federated.oauthConfig` for the `amplify-oauth-button`.

The new constant is is consistent with the label text now.

ericclemmons:
@jordanranz No idea why the Amazon button is looking this way.  I compared the diff of the SVG and they were identical.
> Screen Shot 2019-11-07 at 11 31 25 AM

ericclemmons:
Any other changes requested can be done in a smaller, follow-up PRs.

Planning on merging this tomorrow so we can avoid merge conflicts & keep progress going. 💪

packages/amplify-ui-components/src/common/types/auth-types.ts
```diff
@@ -21,6 +21,20 @@ export enum AuthState {
   VerifyContact = 'verifycontact',
 }
 
+export interface FederatedConfig {
```
jordanranz: Nice!

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
```diff
@@ -0,0 +1,100 @@
...
+      const authenticatedUser = await Auth.currentAuthenticatedUser();
+
+      this.handleAuthStateChange(AuthState.SignedIn, authenticatedUser);
+    });
+  };
+
+  /**
+   * @see https://developer.amazon.com/docs/login-with-amazon/install-sdk-javascript.html
+   */
+  handleClick = event => {
```
jordanranz: Can we name this `amazonSignIn` or `amazonFederatedSignIn` and call it from the `amplify-sign-in-button` in the render. Like `<amplify-sign-in-button onClick={event => amazonFederatedSignIn(event)} provider="amazon">`

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -17,6 +17,8 @@ export class AmplifyAuthenticator {
   @State() authState: AuthState = AuthState.Loading;
 
   @State() authData: CognitoUserInterface;
+  /** Federated credentials & configuration. */
+  @Prop() federated: FederatedConfig = {};
```
jordanranz: This may be better using a default of `null` or `undefined`. What do you think?

packages/amplify-ui-components/src/components/amplify-facebook-button/amplify-facebook-button.tsx
```diff
@@ -0,0 +1,120 @@
...
+          this.federatedSignIn(response.authResponse);
+        }
+      },
+      { scope },
+    );
+  };
+
+  render() {
+    return (
+      <amplify-sign-in-button onClick={this.handleClick} provider="facebook">
```
jordanranz: Same comment from the `amazon-sign-in-button` above.

packages/amplify-ui-components/src/components/amplify-federated-sign-in/amplify-federated-sign-in.tsx
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    if (!Object.values(AuthState).includes(this.authState)) {
+      return null;
+    }
+
+    logger.debug('federated Config is', this.federated);
+
+    return (
+      <amplify-form-section data-test="federated-sign-in-section">
```
jordanranz: We should make a common file for these data-test attributes

ericclemmons:
Made notes of the follow-up items. 👀 Thanks everyone for the feedback in getting this thing functional & tested!

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
ericclemmons: Can you clarify this?  Are you asking for `const amazonFederatedSignIn` outside the scope of `this`, or are you asking that `handleClick` be a less ambiguous name?

`onClick={this.amazonFederatedSignIn}` is equivalent to `onClick={(event) => this.amazonFederatedSignIn(event)`, but _without_ extraneous closure that's generated per-render.

Issue #3719, opened 2019-07-23
Cognito Signature expired


mezalejandro:
**Describe the bug**
After login with cognito i have the following error:
{"message":"Signature expired: 20190723T172834Z is now earlier than 20190723T201503Z (20190723T203003Z - 15 min.)","code":"InvalidSignatureException","time":"2019-07-23T17:28:34.586Z","requestId":"6GQQCIH6BKQ8UK7035CLUUI44FVV4KQNSO5AEMVJF66Q9ASUAAJG","statusCode":400,"retryable":true,"line":122002,"column":42,"sourceURL":"http://10.0.2.2:8081/index.delta?platform=android&dev=true&minify=false"}

**To Reproduce**
1. LogIn user and password

**Additional context**
React Native 0.59.9

haverchuck:
@mezalejandro - In the entry point of your application, can you do:
`import AWS from 'aws-sdk'`
...and then, after you call Amplify.configure, do
`AWS.config.correctClockSkew = true`?

Please respond with whether or not that has any effect.

mezalejandro:
> @mezalejandro - In the entry point of your application, can you do:
> `import AWS from 'aws-sdk'`
> ...and then, after you call Amplify.configure, do
> `AWS.config.correctClockSkew = true`?
> 
> Please respond with whether or not that has any effect.

Hello,

for now it is working fine with AWS.config.correctClockSkew = true

Thank you

willdady:
@haverchuck your solution requires explicitly adding `aws-sdk` as a dependency yes? It would be great if `correctClockSkew` was exposed via Amplify's config rather than having to install aws-sdk since it's already a dependency of Amplify. Personally I feel this should simply be **true by default**. My team sees this error logged to Sentry several times a day for our React Native app. We've had to resort to explicitly asking our users to check their device time in our error messages because of this.

ericclemmons:
With the latest release of Amplify, we leverage https://github.com/aws/aws-sdk-js-v3 which automatically accounts for clock drift by default: https://github.com/aws/aws-sdk-js-v3/pull/459 

Can you share if you're still experiencing issues and, if so, steps to reproduce?

ericclemmons:
I tested this flow with a +5 hour clock drift and confirmed that username + password logins work, thanks to aws-sdk-v3's auto-drift correction.

For anyone affected, be sure to upgrade to the latest `aws-amplify` (v3) and let me know if you still experience any issues!

[...]

Issue #3699, opened 2019-07-21
Analytics.record throws  ExpiredTokenException and InvalidSignatureException


libeanim:
**Which Category is your question related to?**
Amplify Analytics in combination with React Native (Expo)

**What AWS Services are you utilizing?**

- Amplify Analytics (Pinpoint)
- Amplify Auth (Cognito)
- Amplify Api (Appsync)

**Provide additional details e.g. code snippets**
Hi guys, thanks for your work! :)
I recently added analytics to our productions system and since then I receive errors via our bug tracking service that seem to come from the `Analytics.record` function when I record custom events in our app.
I received following two errors which seem to be outside of my control:
- `AWSPinpointProvider - record event failed.  ExpiredTokenException: The security token included in the request is expired`.
As far as I understand, keeping the authentication token (which is issued by cognito) up to date is the responsibility of the sdk? There haven't been any issues with AppSync which we use as backend via 'Amplify Api'.

- `[ERROR] 16:45.276 AWSPinpointProvider - record event failed.  InvalidSignatureException: Signature not yet current: 20190721T181644Z is still later than 20190721T180702Z (20190721T180202Z + 5 min.)` Well maybe one of the users changed the clock on their phone manually? Why does this cause a signature exception?

Did I miss something in the documentation that would prevent this from happening or is that expected behaviour and I have to wrap the `Analytics.record` function calls in a `try-catch-block`?

_I am a bit scared that these exceptions actually cause the main thread of the app to crash._

libeanim:
Ok I can confirm that the `InvalidSignatureException` happens when someone changes the phone clock to a time in the future. Also it causes a react-native app to crash when this error occurs.

- Wrapping the `Analytics.record` function in a try-catch-block doesn't change anything, it still crashes.
- Setting `autoSessionRecord` to `false` doesn't have an effect.
- Disabling the Analytics entirely  doesn't have an effect either.

Only removing the entire analytics part from the app solves it. As a consequence I cannot use Amplify Analytics and Pinpoint. This seems like a serious bug.

I could not reproduce the ` ExpiredTokenException` but I assume it also causes the app to crash.

haverchuck:
@libeanim Are you seeing this on both iOS and Android?  I have been unable to reproduce on an Android emulator.

haverchuck:
@libeanim In the entry point of your application, can you do:
import AWS from 'aws-sdk'
...and then, after you call Amplify.configure, do
[...]

Issue #2014, opened 2018-10-30
correctClockSkew for API


eduardogspereira:
**Which Category is your question related to?**

AWS Amplify API.

**What AWS Services are you utilizing?**

Cognito and API Gateway.

**Provide additional details e.g. code snippets**

Hi,

I notice that on aws-sdk there is an option for `correctClockSkew` (https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/APIGateway.html), that is better explained here https://github.com/aws/aws-sdk-js/issues/1174#issuecomment-253600875 .

Is this or a similar option exists on AWS Amplify? The only approach for me achieve it is similar to here https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/APIGateway.html ?

My main goal on this is that when some machine has a clock that for some reason is not correctly synchronized, then, s/he stills be able to make a request. 

I was getting this error with machines where the time was not correctly set:

```
GET :: "Signature expired: 20181029T181641Z is now earlier than 20181029T191126Z (20181029T191626Z - 5 min.)"
```

Thanks for the help and sorry for the few details.

mhelgeson:
Having a similar problem: users with clocks that are not set correctly are able to authenticate with cognito (amplify/auth) but requests to api gateway are rejected (amplify/api). 

Is there a way to correct the clock skew through some configuration setting?

jayair:
I'm having this issue as well. Is there a fix for this? Is it being worked on?

AWS SDK `correctClockSkew` setting does not work anymore. And this issue is incredibly easy to replicate. Just set your system clock ahead by more than 5 minutes all API requests will fail.

willdady:
Same here. Using Amplify in React Native contacting API Gateway we get **several** of these errors reported daily in Sentry.

`Signature not yet current: 20190115T080011Z is still later than 20190115T075748Z (20190115T075248Z + 5 min.)`

What can be done to fix this? On version 1.1.10 if that helps.

jayair:
Hey guys is this being worked on? We keep getting this error.

This was an issue that the AWS SDK has a fix for. But we aren't able to use the setting through Amplify.

taschik:
Hi,

[...]

PR #4251, opened 2019-10-24
fix(@aws-amplify/core): AWS.config.systemClockOffset for signing requ…

thiagohirata:
…ests

_Issue #, if available:_
#2014  #3719 #3699

_Description of changes:_
Signer calls `AWS.util.date.getDate()` instead of `new Date()` - making it possible to correct the signature and headers for the correct date when the system time is wrong by:

```
import { AWS } from '@aws-amplify/core'
AWS.config = new AWS.Config({
    systemClockOffset: CLOCK_OFFSET
})
```

### DIFF
--- a/packages/core/__tests__/Signer-test.ts
+++ b/packages/core/__tests__/Signer-test.ts
@@ -1,5 +1,5 @@
 jest.mock('../src/Facet', () => {
-	let ret = { util: { crypto: { lib: {} } } };
+	let ret = { util: { crypto: { lib: {} }, date: {} } };
 	ret['util']['crypto']['lib']['createHmac'] = () => {
 		const update = () => {
 			return {
@@ -20,13 +20,15 @@ jest.mock('../src/Facet', () => {
 		};
 		return { update };
 	};
+	ret['util']['date']['getDate'] = () => new Date();
+
 	return {
 		AWS: ret,
 	};
 });
 
 import Signer from '../src/Signer';
-import AWS from '../src';
+import { AWS } from '../src/Facet';
 
 describe('Signer test', () => {
 	describe('sign test', () => {
@@ -45,6 +47,8 @@ describe('Signer test', () => {
 				.spyOn(Date.prototype, 'toISOString')
 				.mockReturnValueOnce('0');
 
+			const getDateSpy = jest.spyOn(AWS['util']['date'], 'getDate');
+
 			const res = {
 				headers: {
 					Authorization:
@@ -56,6 +60,7 @@ describe('Signer test', () => {
 				url: url,
 			};
 			expect(Signer.sign(request, access_info)).toEqual(res);
+			expect(getDateSpy).toHaveBeenCalledTimes(1);
 
 			spyon.mockClear();
 		});
--- a/packages/core/src/Signer.ts
+++ b/packages/core/src/Signer.ts
@@ -299,7 +299,7 @@ export default class Signer {
 		request.headers = request.headers || {};
 
 		// datetime string and date string
-		const dt = new Date(),
+		const dt = AWS['util']['date'].getDate(),
 			dt_str = dt.toISOString().replace(/[:\-]|\.\d{3}/g, ''),
 			d_str = dt_str.substr(0, 8);
 
@@ -370,7 +370,10 @@ export default class Signer {
 		const body: any =
 			typeof urlOrRequest === 'object' ? urlOrRequest.body : undefined;
 
-		const now = new Date().toISOString().replace(/[:\-]|\.\d{3}/g, '');
+		const now = AWS['util']['date']
+			.getDate()
+			.toISOString()
+			.replace(/[:\-]|\.\d{3}/g, '');
 		const today = now.substr(0, 8);
 		// Intentionally discarding search
 		const { search, ...parsedUrl } = url.parse(urlToSign, true, true);


thiagohirata:
Hi, @Amplifiyer! 
Any chance checking this out?

Amplifiyer:
> Hi, @Amplifiyer!
> Any chance checking this out?

@thiagohirata thanks for submitting the PR. I have a few questions about the implementation and the intended usage.
* How are customers going to calculate CLOCK_OFFSET and pass it to AWS.Config?
* Would it be better to calculate this offset in the signer itself? Such as by reading the data header in the response and calculating the clock drift and setting it again on the retry?

We are also looking to migrate towards aws-sdk V3 which is in preview right now. V3 version doesn't export a global AWS variable or currently has any ways to provide a clockSkew value. This change will then need to be migrated as well when we migrate to V3 of aws-sdk.

thiagohirata:
Hi, @Amplifiyer 

I am not sure if the current version of aws-sdk provides an automatic way to recalculate itself the clock skew. I will check.

Making this process automatic sounds good, but I was expecting that amplify-js would follow the same behaviour as aws-sdk.

This is a current issue here in Brazil - we had in two consecutive years changes in daylight saving time rules (the current rule is that there is no daylight saving time), and there is a bunch of devices with outdated databases. With users manually adjusting the time, but not changing the internal timezone offset, all signatures are automatically expired.

I made a webservice myself to calculate the skew of the client. The requests made with aws-sdk work when I provide this option, but there is no way to make amplify-js work.

Thanks!

dorsegal:
@Amplifiyer 
We applied the fix manually and it worked great.
Any change this merge will happen?

dorsegal:
We found a way to calculate clock offset using aws amplify:
```
Auth.currentAuthenticatedUser({
      bypassCache: true // this will get the latest clockDrift
    }).then(user => {
        AWS.config = new AWS.Config({
          systemClockOffset: -(user.signInUserSession.clockDrift * 1000)
      })
}
```

thiagohirata:
Hi @Amplifiyer!

> @thiagohirata thanks for submitting the PR. I have a few questions about the implementation and the intended usage.
> 
> * How are customers going to calculate CLOCK_OFFSET and pass it to AWS.Config?

I created a webservice that returns the server timestamp, and the client compares it with the browser's date. @dorsegal provided another way.

> * Would it be better to calculate this offset in the signer itself? Such as by reading the data header in the response and calculating the clock drift and setting it again on the retry?

Maybe, but I personally think that it is better to give more alternatives to the user.

> We are also looking to migrate towards aws-sdk V3 which is in preview right now. V3 version doesn't export a global AWS variable or currently has any ways to provide a clockSkew value. This change will then need to be migrated as well when we migrate to V3 of aws-sdk.

Yes, but I don't think this is a blocking reason for not merging this PR, as this solves a problem with current aws-sdk verson - don't you agree?

houmark:
Dear AWS team,

Instead of searching for the perfect long term solution and consider eight different options, when you cannot come up with a solution that solves this issue better than the already ready PR, please respect your customers and their valuable time, by merging the already ready PR. Then you can always improve on the solution later when you come up with a better solution long term.

This issue is potentially affecting thousands of users in the wild and developers using Amplify can spend hours individually searching for why a rare bug like this happens, while you could solve this with a click of a button. There's nothing complex about this PR in any way, so the risk of regression is about non-existent.

LGTM!!!

computationalcore:
@houmark is absolutely correct.  In my case, it is literally affecting thousands of users in a published application. Please, merge it and release ASAP.

Amplifiyer:
I'll look into this PR tomorrow and update.

computationalcore:
> I'll look into this PR tomorrow and update.

Ok, thanks @Amplifiyer

Jun711:
This potentially affect a lot of AWS customers who use AWS_IAM for API authorization.
Some paid users may report to us but new users may just leave our website because the website is not working as expected.

Jun711:
@Amplifiyer @thiagohirata 
I wonder if you could explain what the possible cause of this case. We have asked the user to turn on` set time automatically` on a windows machine but there is still drift issue. Maybe `set time automatically` doesn't set time correctly immediately? 

Error message is 
```
Signature expired: 20200204T203226Z is now earlier than 
20200204T203234Z (20200204T203734Z - 5 min.) .  
```

in har file

**Request**
"startedDateTime": "2020-02-04T20:32:26.801Z",
"time": 134.58600000012666,

header
```
"name": "x-amz-date",
"value": "20200204T203226Z"
```

**Response**
```
 "name": "date",
 "value": "Tue, 04 Feb 2020 20:37:34 GMT"
```

Amplifiyer:
We discussed internally how to ensure that future changes (integrating with [aws-sdk-js-v3](https://github.com/aws/aws-sdk-js-v3)) doesn't break this, we are making small changes(#4844) to this PR to abstract out the `AWS` global variable and exposing a util to handle the clock skew.

We are working on releasing this as soon as possible.

ramanasak:
Hi, @Amplifiyer 

How to give a customized header to Sign-in screen, I have used signUpConfig for sign up screen its working fine but for the sign-in screen, it's not working, 

const signUpConfig = {
  //header: 'My Customized Sign Up',
  header: <HeaderLogo/>,
  hideAllDefaults: true,
  defaultCountryCode: '1',
  Label:{color:'red'},
 
};
const SigninConfig = {
  //header: 'My Customized Sign Up',
  header: <HeaderLogo/>,
  hideAllDefaults: true,
  defaultCountryCode: '1',
  Label:{color:'red'},
};

export default withAuthenticator((connect(mapStateToProps, mapDispatchToProps)(App)),{ signUpConfig },true,[],null,MyTheme);


I am stuck here please help thanks in advance.

PR #4258, opened 2019-10-25
fixes #4225 decouple local storage key amplify-signin-with-hostedUI

### LINKED ISSUE #4225: amplify-signin-with-hostedUI is loosing value when you navigate to other components in React JS
**Describe the bug**
As soon as logged in, amplify-signin-with-hostedUI is going to be true, when you navigate to other components, this key is getting removed from the localstorage, hence Auth.logout() is not working.

**To Reproduce**
create multiple Single Page applications using ReactJS. And and see if you navigate from one app to other app, from the local storage variable amplify-signin-with-hostedUI is getting remove.

refer the ticket AWS case 6508461831 for more information.

Amplifiyer:
_Issue #, if available:_
Fixes: #4225 

_Description of changes:_
Key `amplify-signin-with-hostedUI` was overloaded at two places for apparently disconnected reasons.
* Created a new key `amplify-redirected-from-hosted-ui` to be used in the react package to specify that user has just landed/redirected from the hosted UI.
* This key will help to not call checkUser() right away as Auth.currentAuthenticatedUser may not return credentials right away.
* Updated comments around this workaround and made it clearer.
* Set this new key in `Auth:_handleAuthResponse` as soon as it is determined that the redirected URL was from hostedUI (or OAuth flow) (This has to happen before Authenticator component is mounted)
* As before, the key is removed on the first load (mount) of Authenticator component.


jordanranz:
Curious if the following items are affected by this PR: https://github.com/aws-amplify/amplify-js/blob/94b7181d0c37fa9dbfaec8d09b93c0c9f1246ec3/packages/auth/src/Auth.ts#L119-L125

Amplifiyer:
> Curious if the following items are affected by this PR:
> 
> https://github.com/aws-amplify/amplify-js/blob/94b7181d0c37fa9dbfaec8d09b93c0c9f1246ec3/packages/auth/src/Auth.ts#L119-L125

Yes, the reason for this PR is to decouple react package's workaround (not doing checkUser() only on the first page load after redirected from cognito hosted ui) from this functionality which is to ensure that we always sign out from cognito hosted ui if the user had signed in from hosted ui

packages/aws-amplify-react/src/Auth/Authenticator.tsx
```diff
@@ -152,9 +153,6 @@ export default class Authenticator extends React.Component<
 				case 'customGreetingSignOut':
 					this.handleStateChange('signIn', null);
 					break;
-				case 'parsingCallbackUrl':
```
ashika01: Looks like there is an event dispatched for this case from [here](https://github.com/aws-amplify/amplify-js/pull/4258/files#diff-1cf5b10c87e919f81d5f1ea503f15074L1729) . We might wanna consider why it is there in first place and/or consider putting back this block.

Amplifiyer: yes it has been dispatched from there, bu there are two issues to keep this case
* The event is dispatched even when the URL callback is not from hosted UI but just a page load. This effectively prevents react component to determine if the component is loaded right after the hosted UI
* Even if fix the dispatch such that the event is only for hosted UI redirects, this component is too late to set the hub listener and is always missing the parsingCallbackUrl event locally.
Since this event only was setting the key and we removed that, I saw no point in keeping the switch case here especially because of point 2 above.

Issue #1989, opened 2018-10-25
Amplify Analytics does not batch submit events


lambrospetrou:
**Describe the bug**
The AWS Amplify Analytics do not batch the events, even when using the buffering capabilities offered, but instead submit each event individually (two requests per event, OPTIONS+submission).

I was using the old [amazon mobile analytics](https://github.com/aws/aws-sdk-mobile-analytics-js) and it was batching the events just fine, but upgraded to amplify for the new features and due to the fact that the AMA bundled the whole AWS SDK.

It's unfortunate that this shiny new library does not batch submit the events, resulting in hundreds of API calls in my application, hence restricting me on how many things to record.

**To Reproduce**
1. Enable Amplify analytics using the [default guide](https://aws-amplify.github.io/docs/js/analytics#record-a-custom-event-with-attributes)
2. Call `Analytics.record(...)` many times (10+) in the same second to add events to the buffer
3. Open the **network** section in your web developer tools
4. See the numerous submissions to the API with each submission containing a single event

**Expected behavior**
I expected a submission to contain multiple recorded events, not necessarily to the local buffer size, but definitely not 1 event per API call.

**Desktop (please complete the following information):**
 - OS: Linux, MacOS
 - Browser: Chrome, Firefox, Safari

**Issue in the code**

I believe the problem is [this line of the code](https://github.com/aws-amplify/amplify-js/blob/master/packages/analytics/src/Providers/AWSPinpointProvider.ts#L76) which executes `this._sendFromBuffer(params);` for each individual event from the buffer, which in turn calls `_recordCustomEvent`, which in turn calls [`_generateBatchItemContext(params)`](https://github.com/aws-amplify/amplify-js/blob/master/packages/analytics/src/Providers/AWSPinpointProvider.ts#L341-L343). Note the misleading name, which is not a **BatchItemContext** since it only includes a **single** event.

In my opinion, [this looping over the **buffered** events](https://github.com/aws-amplify/amplify-js/blob/master/packages/analytics/src/Providers/AWSPinpointProvider.ts#L74-L77) should create a batch context containing several events and then doing one API call for each of those batches.

mlabieniec:
The methods you are referencing here are part of a larger update/refactor to the Pinpoint provider to actually support an updated batch events Pinpoint service API that did not exist prior. This update on the Amplify Analytics side should in-turn be complete in the next couple sprints and is in our backlog up next now that the API is available. Will track the release of that update here.

[...]

PR #4289, opened 2019-10-29
feat(analytics): Allow sending batch events

iartemiev:
_Issue #, if available:_
Fixes: #1989

_Description of changes:_
**Enables batch event sending.**

I took out the other features I'm working on as part of the Analytics Refactor because I need to switch over to another task for a little bit and wanted to get this out sooner rather than later.


packages/analytics/__tests__/Providers/AWSPinpointProvider-unit-test.ts
```diff
@@ -501,7 +501,8 @@ describe('AnalyticsProvider test', () => {
 				await analytics.record(params, { resolve, reject });
 
 				const expectedUrl =
-					'https://pinpoint.region.amazonaws.com/v1/apps/appId/events?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=accessKeyId%2FisoStrin%2Fregion%2Fmobiletargeting%2Faws4_request&X-Amz-Date=isoString&X-Amz-Security-Token=sessionToken&X-Amz-SignedHeaders=host&X-Amz-Signature=c43336d302010144ca492cc6f42d5545e0ef01c00bc71d99dc7597351f027810';
+					'https://pinpoint.region.amazonaws.com/v1/apps/appId/events/legacy?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=accessKeyId%2FisoStrin%2Fregion%2Fmobiletargeting%2Faws4_request&X-Amz-Date=isoString&X-Amz-Security-Token=sessionToken&X-Amz-SignedHeaders=host&X-Amz-Signature=9dfa2a29782d344c56a9ab99fe58db6d1748e097ae418c398b26ab372a23f22f';
```
sammartinez: Can we confirm that `Amz-Credential=accessKeyId%2FisoStrin%2Fregion` is not a typo on the `Strin`? Im just seeing `isoString` in other places

packages/analytics/src/Analytics.ts
```diff
@@ -254,7 +254,7 @@ export default class AnalyticsClass {
 	}
 
 	public async updateEndpoint(attrs, provider?) {
-		const event = Object.assign({ name: '_update_endpoint' }, attrs);
+		const event = { ...attrs, name: '_update_endpoint' };
```
sammartinez: Will this work? We may not have this compile down just want to double check

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -415,36 +619,35 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
+			logger.debug('updateEndpoint success', data);
+			this._endpointGenerating = false;
+			return handlers.resolve(data);
+		} catch (err) {
+			logger.debug('updateEndpoint failed', err);
+			if (err.message.startsWith('Exceeded maximum endpoint per user count')) {
+				try {
+					await this._removeUnusedEndpoints(appId, request.User.UserId);
+
+					logger.debug('Remove the unused endpoints successfully');
```
sammartinez: Remove or Removed?

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -644,12 +847,26 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
+	private _bufferToMap(buffer: EventBuffer) {
+		return buffer.reduce((acc, curVal) => {
+			const [[key, value]] = Object.entries(curVal);
+			acc[key] = value;
+			return acc;
+		}, {});
+	}
+
+	private _eventError(err: any) {
+		logger.error('record event failed. ', err);
```
sammartinez: Is this space needed after `record event failed. `?

packages/core/src/Credentials.ts
```diff
@@ -158,7 +158,7 @@ export class Credentials {
 		}
 		logger.debug('is this credentials expired?', credentials);
 		const ts = new Date().getTime();
-		const delta = 10 * 60 * 1000; // 10 minutes
+		const delta = 10 * 60 * 1000 - 5000; // 10 minutes minus 5 seconds
```
sammartinez: why the change? Just curious is all. Related to the batch of events?

packages/analytics/src/Analytics.ts
iartemiev: It'll work. We use object prop destructuring elsewhere as well. 
For example: https://github.com/aws-amplify/amplify-js/blob/master/packages/api/src/API.ts#L68

packages/core/src/Credentials.ts
iartemiev: It's not, it's related to the Creds / Retry stuff I'm also working on

Amplifiyer: @elorzafe is also working on something similar (if not same). Let's sync up

iartemiev: I'll take it out for now though

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -79,34 +90,73 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
+			const endpointBatchSize =
+				that._endpointBuffer.length < flushSize
+					? that._endpointBuffer.length
+					: flushSize;
+
+			if (!batchSize && !endpointBatchSize) {
+				return;
+			}
+
+			// TODO: figure out if this is still relevant
```
jordanranz: Is this TODO still needed?

iartemiev: Probably not haha. I'll remove it

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -79,34 +90,72 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
+			}
+
+			if (that._endpointGenerating && endpointBatchSize) {
 				// If this is the first request sent by Analytics module, we should stop sending remaining requests
 				// to prevent race condition of updating one endpoint when it's being created in the backend
-				if (this._endpointGenerating) break;
+				const firstEvent = that._endpointBuffer.shift();
+				const [{ params, handlers }] = Object.values(firstEvent);
+				that._send(params, handlers);
+				return;
```
elorzafe: that will wait for `flushInterval` to do the next one right?, what happen if customer set that to low value by default. How long should it wait, in case the app sends an immediate event it should wait for that, or maybe there should an exponential retry for an error because of waiting for first update endpoint event?

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -389,15 +565,39 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
+
+		const batchEndpointUpdates = this._batchEndpoints(endpointsMap);
+		logger.debug('batchEndpointUpdates', batchEndpointUpdates);
+
+		try {
+			const data = await this.pinpointClient
+				.updateEndpointsBatch(batchEndpointUpdates)
+				.promise();
+
+			// TODO: handle response
```
elorzafe: I think you are missing something here ;)

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -79,34 +90,72 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
 		const that = this;
-		this._timer = setInterval(() => {
-			const size =
-				this._buffer.length < flushSize ? this._buffer.length : flushSize;
-			for (let i = 0; i < size; i += 1) {
-				const { params, handlers } = this._buffer.shift();
-				that._send(params, handlers);
+
+		this._timer = setInterval(async () => {
+			await that._getCredentials(); // TODO: set class creds here
```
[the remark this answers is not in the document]
elorzafe: There might be a refactor on `_initClients` method, because the credentials are not store on this provider

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -369,6 +544,7 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
 	}
 
 	private _retry(params, handlers) {
+		logger.debug('_retry', params, handlers);
```
elorzafe: I think this retry logic has a bug since a long time, is putting the retryable events on the end of the buffer, or maybe it shouldn't be an issue. What do you think?

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -79,34 +90,72 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
 		if (this._buffer.length < bufferSize) {
-			this._buffer.push({ params, handlers });
+			const {
+				event: { eventId, name },
+			} = params;
+
+			const bufferElement = { [eventId]: { params, handlers } };
+
+			if (name === UPDATE_ENDPOINT) {
+				this._endpointBuffer.push(bufferElement);
```
elorzafe: this could be an issue with `updateEndpoint`, for retry it will put it on the end of the buffer an can override other updateEndpoints (basically processing in sequence it will be broken)

iartemiev:
Closed this PR and re-opened as #4377 due to unintentionally getting my local branch too far out of sync with this one through rebasing latest master, etc.

PR #4323, opened 2019-11-01
feat(@aws-amplify/storage): get supports Response headers

### LINKED ISSUE #3563: How do you override response header values with Storage.get
I've got an S3 bucket where I'm storing mp3 files.

When using Storage.put(), I can set (or not set) the Content-Type and Content-Disposition of the files.  

If I set the Content-Type to "audio/mpeg" and Content-Disposition to "attachment" I can use Storage.get() to retrieve a pre-signed URL, which will prompt an OS download window to save the file.

If I only set the Content-Type to "audio/mpeg", leaving Content-Disposition at default ("inline"), Storage.get() will return a pre-signed URL that will begin streaming/playing the file in the browser.

This seems to be an either/or road.

What I'm ultimately trying to do is allow for both streaming and/or downloading in the UI, like this, with the play and download buttons having the appropriate behaviours:
image

While I can set the Content-Type and Content-Disposition on PUT with something like this:

**For Dowloading**
```js
await Storage.put(fileName, file, {
  contentType: "audio/mpeg",
  contentDisposition: "attachment",
});
```
**For Streaming**
```js
await Storage.put(fileName, file, {
  contentType: "audio/mpeg"
});
```

I can't seem to figure out how to do this with the GET method, so as to override the Content-Disposition.

I've tried the following with the file uploaded to S3 using a Content-Type: "audio/mpeg" and no Content-Disposition, trying to override the Content-Disposition to allow downloads, using the signed-urls:

```js
const url = await Storage.get(fileName, {
  contentDisposition: "attachment"
});
```
and
```js
const url = await Storage.get(fileName, {
  response-content-disposition: "attachment",   
});
```

But the download vs. stream behavior will always default to whatever the Content-Disposition is set on the file in S3 and no override is happening. 

This site I'm working on was initialized via the CLI using `amplify init`, S3 storage was provisioned using `amplify add storage` (S3), and the (create-react) app was published via a connected repo.

Any thoughts on this would be appreciated.

ericclemmons:
Fixes #3563 

`Storage.get` now supports the following response headers when getting signed URLs:

- `cacheControl` as `ResponseCacheControl`
- `contentDisposition` as `ResponseContentDisposition`
- `contentEncoding` as `ResponseContentEncoding`
- `contentLanguage` as `ResponseContentLanguage`
- `contentType` as `ResponseContentType`
- `expires` as `ResponseExpires` **instead of `Expires`**

    > https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#getObject-property

---


packages/storage/__tests__/Providers/AWSS3Provider-unit-test.ts
```diff
@@ -333,7 +368,7 @@ describe('StorageProvider test', () => {
 			expect(spyon).toBeCalledWith('getObject', {
 				Bucket: 'bucket',
 				Key: 'public/key',
-				Expires: 1200,
+				ResponseExpires: 1200,
```
ericclemmons: This is potentially a backwards compatible break, but based on the API documentation, `Expires` doesn't exist:
> https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#getObject-property

packages/storage/src/Providers/AWSS3Provider.ts
```diff
@@ -154,9 +164,13 @@ export default class AWSS3Provider implements StorageProvider {
...
-		if (expires) {
-			params.Expires = expires;
-		}
+		if (expires) params.ResponseExpires = expires;
+		if (cacheControl) params.ResponseCacheControl = cacheControl;
+		if (contentDisposition)
+			params.ResponseContentDisposition = contentDisposition;
+		if (contentEncoding) params.ResponseContentEncoding = contentEncoding;
+		if (contentLanguage) params.ResponseContentLanguage = contentLanguage;
+		if (contentType) params.ResponseContentType = contentType;
```
Amplifiyer: Are we able to verify that `param.ResponseExpires` works while `param.Expires` doesn't? Looking at the SDK code, especially for getSignedURL API, it seems it looks at the `Expires`. Also I believe getSignedURL() doesn't actually hit S3 API to return the signed URL, it's generated by the SDK with proper signature so that client can then call GET on it to download.

We also may have to move the param headers before the if condition so that we can pass it in the `s3.getObject()` API

ericclemmons: @Amplifiyer That's a good point.  The safest thing would be to revert 6ad86e2234c1e43bd51f80f0df86b82a59619f44, as that was added in bulk with `ResponseCacheControl` to fix #3567 based on the documentation.

Amplifiyer: Yes and we should probably move updating `params` before the [if condition ](https://github.com/aws-amplify/amplify-js/pull/4323/files/27cc500aad236def8eb97fb03f77abc779ba0a22#diff-c7c081aa778dae8f92ac9eee463ebaffR138-R142) so that getObject() call can use it.

Issue #2466, opened 2019-01-04
Expose profile picture for federated authentication providers


christophgysin:
**Is your feature request related to a problem? Please describe.**
Using withAuthenticator with federated authentication, only name and email are exposed.

**Describe the solution you'd like**
Additionally, the profile picture url should also be exposed.

**Describe alternatives you've considered**
It's possible to fetch the image url manually after the user has logged in.

**Additional context**
Should be supported for
- [x] Google (#2467)
- [ ] Facebook (#4206)
- [x] Auth0 (#4329)

jordanranz:
This is a valid ask. Thanks for the PR. Would you like to PR for Facebook and Auth0 as well?

christophgysin:
Yes, I have commits for Facebook and Auth0, as well as exposing the email for the Amazon provider. These are not fully tested yet and lack tests. Will try to create PRs for those too.

evertson90:
Will this no longer be supported?

ericclemmons:
Finally closed via https://github.com/aws-amplify/amplify-js/pull/4206.  Thanks everone!

PR #4329, opened 2019-11-02
Auth0 profile image

christophgysin:
Issue #2466

Description of changes:
Return profile image url for Auth0 authentication provider.


alkurop:
picture should go as one of the attributes.
it's done in Cognito console / User Pools / Federation / Attribute mapping.
Hardcoding it does not seem like a good idea

christophgysin:
@alkurop Thanks for your feedback. Did you check the linked issue? This is equivalent to how Google authentication provider is implemented. If you disagree with the issue itself, please add a comment on the issue instead.

PR #4348, opened 2019-11-05
feat(@aws-amplify/ui-components): amplify-phone-field

sammartinez:
_Issue #, if available:_
[Delivers #169219518]
_Description of changes:_
* Add a phone field component
* Adjusted the Sign Up component to support the new phone field comp
* Add unit tests / e2e tests


packages/amplify-ui-components/src/components/amplify-country-dial-code/amplify-country-dial-code.tsx
```diff
@@ -4,17 +4,23 @@ import { CountryCodeDialOptions } from './amplify-country-dial-code-interface';
...
   @Prop() options: CountryCodeDialOptions = countryDialCodes;
+  /** The callback, called when the input is modified by the user. */
+  @Prop() handleInputChange?: (inputEvent: Event) => void;
 
   render() {
     return (
-      <amplify-select options={this.options} override-style={this.overrideStyle} />
+      <amplify-select
+        options={this.options}
+        override-style={this.overrideStyle}
```
ericclemmons: For JSX this should be `overrideStyle`.

packages/amplify-ui-components/src/components/amplify-examples/authenticator.tsx
```diff
@@ -2,7 +2,7 @@ import { h } from '@stencil/core';
 
 const Example1 = () => (
   <div>
-    <amplify-authenticator />
+    <amplify-authenticator initial-auth-state="signup" />
```
ericclemmons: `initialAuthState` when using JSX.

packages/amplify-ui-components/src/components/amplify-phone-field/amplify-phone-field.tsx
```diff
@@ -0,0 +1,49 @@
+import { Component, Prop, FunctionalComponent, h } from '@stencil/core';
+import { PHONE_SUFFIX, PHONE_LABEL, PHONE_PLACEHOLDER } from '../../common/constants';
+import { phoneField } from './amplify-phone-field.style';
+
+@Component({
+  tag: 'amplify-phone-field',
+  shadow: false,
```
ericclemmons: `shadow` defaults to `false`, so this can be removed:
> https://stenciljs.com/docs/component#component-options

sammartinez: For our components, we want to be explicit that have shadow set to false. Mainly to make it clear to customers that we don't have shadow set to true or false. Maybe can be a bigger convo with @jordanranz on this since this is in our components else where.

packages/amplify-ui-components/src/components/amplify-country-dial-code/amplify-country-dial-code.tsx
sammartinez: will adjust

packages/amplify-ui-components/src/components/amplify-examples/authenticator.tsx
sammartinez: strange that when I was trying to use `initialAuthState` prior it wasn't liking it. Let me adjust and see if its fine

ericclemmons: If Stencil renders both correctly, and the camelCase version is just the difference of TypeScript getting in the way, let's call this resolved and unblock the PR.

There are further improvements that can be done in a holistic audit, since several patterns need to be uplifted across the board.

packages/amplify-ui-components/src/components/amplify-phone-field/amplify-phone-field.tsx
ericclemmons: If customers have an expectation that web components = shadowed by default, this totally makes sense.

packages/amplify-ui-components/src/components/amplify-examples/authenticator.tsx
sammartinez: ended up removing this. Part of the reason was related to us using enums as part of the `initialAuthState`. So technically, for this example, we could import in the enum and utilize it. Since I was using this for just testing the sign up flow. Im gonna remove it for now.

ericclemmons: You're right, enums can be annoying with TypeScript: they tend to require casting or referencing anywhere they're used.

Going to mark this as resolved, since the most important piece is that the component renders correctly whether kebab-case or camelCase.

jordanranz:
Can you start using the title format:
`feat(@aws-amplify/ui-components): component-name`

packages/amplify-ui-components/src/components/amplify-phone-field/amplify-phone-field.tsx
jordanranz: If for some reason the default behavior was to change this would keep us safe.

packages/amplify-ui-components/src/components/amplify-select/amplify-select.style.ts
```diff
@@ -1,34 +1,29 @@
...
   box-sizing: border-box;
-  margin: 10px 0 10px 0;
   -webkit-appearance: none;
   -moz-appearance: none;
   appearance: none;
   flex-basis: auto;
-  width: 7%;
-  flex: 1;
+  width: fit-content;
+  height: 51px;
```
jordanranz: The general `select` should probably not have a static height. Curious why this is explicitly set

packages/amplify-ui-components/src/components/amplify-examples/authenticator.tsx
jordanranz: You can remove the initial auth state in the example.

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
```diff
@@ -60,37 +83,10 @@ export class AmplifySignUp {
   @State() username: string;
   @State() password: string;
   @State() email: string;
+  @State() country_dial_code_value: string = '+1';
```
jordanranz: I think we should migrate anything using `_` to camel case for consistency, unless there is a good reason for this.

packages/amplify-ui-components/src/components/amplify-country-dial-code/amplify-country-dial-code.tsx
```diff
@@ -4,17 +4,23 @@ import { CountryCodeDialOptions } from './amplify-country-dial-code-interface';
...
   @Prop() overrideStyle: boolean = false;
   /** The options of the country dial code select input. */
   @Prop() options: CountryCodeDialOptions = countryDialCodes;
+  /** The callback, called when the input is modified by the user. */
+  @Prop() handleInputChange?: (inputEvent: Event) => void;
 
   render() {
     return (
-      <amplify-select options={this.options} override-style={this.overrideStyle} />
+      <amplify-select
```
jordanranz: Let's have the fieldId for this set to a constant related to country code.

packages/amplify-ui-components/src/components/amplify-select/amplify-select.style.ts
```diff
@@ -1,34 +1,29 @@
...
+    linear-gradient(135deg, gray 50%, transparent 50%), linear-gradient(to right, #ccc, #ccc);
+  background-position: calc(100% - 20px) calc(1em + 8px), calc(100% - 15px) calc(1em + 8px), calc(100% - 2.5em) 0.5em;
+  background-size: 6px 5px, 6px 5px, 0px 1.5em;
   background-repeat: no-repeat;
-`;
\ No newline at end of file
+
+  :focus {
+    outline: none;
+    border-color: orange;
```
jordanranz: This should use a css var.

packages/amplify-ui-components/src/components/amplify-select/amplify-select.style.ts
```diff
@@ -1,34 +1,29 @@
 import { css } from 'emotion';
 
 export const select = css`
-  display: flex;
   padding: 14px;
   font-size: 14px;
   color: var(--deep-squid-ink);
   background-color: #fff;
```
jordanranz: Can we use a css var here.

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
jordanranz: Also, we should make `+1` a constant

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
```diff
@@ -104,7 +100,25 @@ export class AmplifySignUp {
...
+
+    if (name === SELECT_SUFFIX) {
+      this.country_dial_code_value = value;
+    }
+
+    if (name === PHONE_SUFFIX) {
+      this.phone_number_value = value;
+    }
+
+    /** By default, `+1` is set on the `country-dial-code` this will most likely need to be adjusted when a customer passes a different default on the `country-dial-code` */
```
jordanranz: Can we solve for this now. Maybe allow for a default dial code to be passed in as a prop.

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
jordanranz: We should set this to a default dial code prop and also make "+1" a constant.

packages/amplify-ui-components/src/components/amplify-examples/authenticator.tsx
sammartinez: Already removed

packages/amplify-ui-components/src/components/amplify-select/amplify-select.style.ts
sammartinez: This was pulled over from pre-exisiting css in the mono repo, let me look to remove the height

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
sammartinez: Sounds good, thought `State()` was the other way but can update as needed

packages/amplify-ui-components/src/components/amplify-select/amplify-select.style.ts
sammartinez: sounds good

packages/amplify-ui-components/src/components/amplify-select/amplify-select.style.ts
sammartinez: si

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
sammartinez: For parity, this is the state it is in when passing from SignUp in react. Can look into it at least

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
sammartinez: yup

packages/amplify-ui-components/src/components/amplify-country-dial-code/amplify-country-dial-code.tsx
sammartinez: So the default fieldId for the `select` component is set to select or the `SELECT_SUFFIX`. It is the similar pattern we have for other components like the `username`, `password`, etc. components. In the dom on the inputs themselves this leads to having the `fieldId` be the `name` on the input component. Im not sure changing to listen to the constant makes sense in this case especially if that'll change over time. WDTY?

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
jordanranz: I am saying just add default dial code as a prop and set the state to the prop. The default prop value will be "+1"

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
```diff
@@ -104,7 +101,25 @@ export class AmplifySignUp {
...
-    this.phone_number = event.target.value;
+    const name = event.target.name;
+    const value = event.target.value;
+
+    /** Cognito expects to have a string be passed when signing up. Since the Select input is separate
+     * input from the phone number input, we need to first capture both components values and combined
+     * them together.
+     */
+
+    if (name === SELECT_SUFFIX) {
```
jordanranz: Talked offline about a small rework of this. Just making a note.

packages/amplify-ui-components/src/components/amplify-select/amplify-select.tsx
```diff
@@ -17,20 +22,29 @@ export class AmplifySelect {
   /** The options of the select input. Must be an Array of Objects with an Object shape of {label: string, value: string|number} */
   @Prop() options: SelectOptionsString | SelectOptionsNumber = DEFAULT_SELECT_OPTION;
   /** Used for id field */
-  @Prop() fieldId: string;
+  @Prop() fieldId: string = SELECT_SUFFIX;
```
jordanranz: I don't think this should have a default.

sammartinez: Fair call, Ill remove

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
```diff
@@ -104,7 +100,30 @@ export class AmplifySignUp {
...
+
+    if (name === PHONE_SUFFIX) {
+      this.phoneNumber.phoneNumberValue = value;
+    }
+
+    return this.phoneNumber;
+  }
+
+  composePhoneNumberInput(phoneNumber: PhoneNumberInterface) {
+    return `${phoneNumber.countryDialCodeValue || COUNTRY_DIAL_CODE_DEFAULT}${phoneNumber.phoneNumberValue.replace(
```
jordanranz: I understand this is what was in the React package but can we clean this up. Also curious what happens if `phoneNumberValue` is null

packages/amplify-ui-components/src/components/amplify-select/amplify-select.style.ts
jordanranz: Can we achieve this with padding?

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
sammartinez: I believe this will be on more of the validation. If phone is null, Cognito will return an error stating the at the format of the phone number is invalid. WDTY about a TODO to add validation in for this?

sammartinez: adjusted based on our feedback

packages/amplify-ui-components/src/common/constants.ts
```diff
@@ -29,6 +29,11 @@ export const CODE_PLACEHOLDER = 'Enter code';
 export const SEND_CODE = 'Send Code';
 export const CONFIRM = 'Confirm';
 
+// Phone common constants
+export const PHONE_SUFFIX = 'phone';
+export const PHONE_LABEL = 'Phone Number *';
+export const PHONE_PLACEHOLDER = '5555555555';
```
jordanranz: Can we make the placeholder (555) 555-1212

PR #4358, opened 2019-11-07
feat: enable watch mode for builds

Amplifiyer:
_Issue #, if available:_ NA

### New watch mode
Enable building the library with watch mode. Following options are possible now
* Build and watch a sub package. Run command `npx lerna exec --scope @aws-amplify/<package> yarn build:esm:watch` OR `yarn run build:esm:watch --scope @aws-amplify/<package>` 
* Build and watch the whole library. Run command `yarn build:esm:watch` from root workspace. This enables packages that export ES modules to enter in watch mode, for other packages, the existing mechanism to watch the builds remains. This will also ensure that if a downstream package is changed, all it's upstream dependencies will be built incrementally.

### Other improvements
  * Stream the log output of builds and test of sub packages in real time when executing commands from root.
  * Enable incremental build to improve the build times and 

### Caveats
 * Running `yarn build:esm:watch` ignores the topological order to build the packages, hence a full build is required before running this command from the root.
 * Running `yarn build:cjs:watch` from the root will build and start watch mode for all the packages, however there won't be cascading builds for consumers as is the case in `yarn build:esm:watch`.
 * aws-amplify-angular doesn't get built when a downstream package is changed when running the watch mode from root. Possibly because of https://github.com/angular/angular-cli/issues/13941
 * Watch mode only works for modules output and not for webpack bundles. Hence only modules in lib and lib-esm for cjs and esm respectively will be built and updated in the watch mode and not the dist or bundles.


scripts/build.js
```diff
@@ -116,8 +169,8 @@ async function buildES5() {
 
 	compilerOptions = ts.convertCompilerOptionsFromJson(compilerOptions);
 	const include = [pkgSrcDir];
-	console.log(`Using the typescript compiler options:`);
-	console.log(compilerOptions);
+	console.debug(`Using the typescript compiler options:`);
+	console.debug(compilerOptions);
```
ericclemmons: When running, there's a significant amount of noise from these statements.  Here's a fraction of what I'm seeing:

```
@aws-amplify/storage: Using the typescript compiler options:
@aws-amplify/storage: {
@aws-amplify/storage:   options: {
@aws-amplify/storage:     noImplicitAny: false,
@aws-amplify/storage:     lib: [
@aws-amplify/storage:       'lib.dom.d.ts',
@aws-amplify/storage:       'lib.es2017.d.ts',
@aws-amplify/storage:       'lib.es2018.asynciterable.d.ts'
@aws-amplify/storage:     ],
@aws-amplify/storage:     jsx: undefined,
@aws-amplify/storage:     sourceMap: true,
@aws-amplify/storage:     target: 1,
@aws-amplify/storage:     module: 5,
@aws-amplify/storage:     moduleResolution: 2,
@aws-amplify/storage:     declaration: true,
@aws-amplify/storage:     noEmitOnError: true,
@aws-amplify/storage:     incremental: true,
@aws-amplify/storage:     tsBuildInfoFile: '/Users/ecclemm/Projects/ericclemmons/amplify-js/packages/storage/lib-esm/.tsbuildinfo',
@aws-amplify/storage:     typeRoots: [
@aws-amplify/storage:       '/Users/ecclemm/Projects/ericclemmons/amplify-js/node_modules/@types',
@aws-amplify/storage:       '/Users/ecclemm/Projects/ericclemmons/amplify-js/packages/storage/node_modules/@types'
@aws-amplify/storage:     ],
@aws-amplify/storage:     types: [ 'node' ],
@aws-amplify/storage:     outDir: '/Users/ecclemm/Projects/ericclemmons/amplify-js/packages/storage/lib-esm'
@aws-amplify/storage:   },
@aws-amplify/storage:   errors: []
@aws-amplify/storage: }
@aws-amplify/storage: Files to be transpiled by tsc:
@aws-amplify/storage: [
@aws-amplify/storage:   '/Users/ecclemm/Projects/ericclemmons/amplify-js/packages/storage/src/Storage.ts',
@aws-amplify/storage:   '/Users/ecclemm/Projects/ericclemmons/amplify-js/packages/storage/src/index.ts',
@aws-amplify/storage:   '/Users/ecclemm/Projects/ericclemmons/amplify-js/packages/storage/src/Providers/AWSS3Provider.ts',
@aws-amplify/storage:   '/Users/ecclemm/Projects/ericclemmons/amplify-js/packages/storage/src/Providers/index.ts',
@aws-amplify/storage:   '/Users/ecclemm/Projects/ericclemmons/amplify-js/packages/storage/src/types/Provider.ts',
@aws-amplify/storage:   '/Users/ecclemm/Projects/ericclemmons/amplify-js/packages/storage/src/types/Storage.ts',
@aws-amplify/storage:   '/Users/ecclemm/Projects/ericclemmons/amplify-js/packages/storage/src/types/index.ts'
@aws-amplify/storage: ]
@aws-amplify/storage: message TS6031: Starting compilation in watch mode...
```

I'd recommend removing this, or leveraging [debug](https://www.npmjs.com/package/debug) to log it behind an environment variable. (e.g. `DEBUG=amplify:*`)

scripts/build.js
```diff
@@ -95,7 +107,46 @@ function tsc(fileNames, options) {
...
+		ts.sys,
+		createProgram,
+		reportErrorDiagnostic,
+		reportWatchStatusChanged,
+		null
+	);
+
+	// `createWatchProgram` creates an initial program, watches files, and updates
+	// the program over time.
+	ts.createWatchProgram(host);
```
ericclemmons: This is good!  It's not often you see TypeScript being used programmatically 💪

scripts/build.js
```diff
@@ -95,7 +107,46 @@ function tsc(fileNames, options) {
...
+	console.error(
+		'Error',
+		diagnostic.code,
+		':',
+		ts.flattenDiagnosticMessageText(
+			diagnostic.messageText,
+			formatHost.getNewLine()
+		)
+	);
+}
```
ericclemmons: Is there more information available in the `diagnostic` object?  (Ideally an equivalent to `error.stacktrace`)

I introduced a syntax error to `Storage.ts` and only got:

```
@aws-amplify/storage: message TS6032: File change detected. Starting incremental compilation...
@aws-amplify/storage: Error 2304 : Cannot find name 'howdy'.
@aws-amplify/storage: Error 1005 : ';' expected.
@aws-amplify/storage: Error 2304 : Cannot find name 'there'.
@aws-amplify/storage: message TS6194: Found 4 errors. Watching for file changes.
```

TypeScript's linter in VS Code seems aware of both the file & the line/col numbers:

```js
{
	"resource": "/Users/ecclemm/Projects/ericclemmons/amplify-js/packages/storage/src/Storage.ts",
	"owner": "typescript",
	"code": "2304",
	"severity": 8,
	"message": "Cannot find name 'howdy'.",
	"source": "ts",
	"startLineNumber": 20,
	"startColumn": 1,
	"endLineNumber": 20,
	"endColumn": 6
}
```

packages/auth/package.json
```diff
@@ -16,6 +16,8 @@
     "build-with-test": "npm test && npm run build",
     "build:cjs": "node ./build es5 && webpack && webpack --config ./webpack.config.dev.js",
     "build:esm": "node ./build es6",
+    "build:cjs:watch": "node ./build es5 --watch",
+    "build:esm:watch": "node ./build es6 --watch",
```
ericclemmons: There's an opportunity in the future to introduce a `"private": true"` package (e.g. `amplify-scripts`) such that we don't need to reference `../../scripts/build` within all of our packages, but instead run `amplify-scripts build es6 --watch`.

(Think of it like `react-scripts`, but for internal use)

Not needed for this PR, but will help hoist dependencies & build complexity out of individual packages. ✨

Amplifiyer:
> @Amplifiyer This is great! A fantastic DX improvement that I was able to validate locally:
> 
>     1. `hub pr checkout 4358`
> 
>     2. `yarn clean`
> 
>     3. `yarn build:esm:watch`
> 
>     4. 🎉
> 
> 
> There are some DX opportunities that can be made within this PR or as a follow-on.
> 
> For merging, I would like to see https://github.com/aws-amplify/amplify-js/wiki/Local-Development updated to include these updated steps (including changes to `yarn link`, if any). (In addition, I believe that wiki page should be folded into `CONTRIBUTING.md`, to keep with open-source conventions.)
> 
> Again, fantastic work @Amplifiyer! 🌮 (Vegetarian with tempeh)

Thanks @ericclemmons, yes my next follow up is to update the wiki and good point on updating the `CONTRIBUTING.md`. Loved the tempeh 🌮!

scripts/build.js
Amplifiyer: I agree, I tried to change the log to `console.debug()` but that didn't work and I somehow thought something was wrong with my console. I'll look into debug.

scripts/build.js
Amplifiyer: Unfortunately I couldn't find a stack trace in the diagnostic object. there is a start and the length for the token that is invalid which is what vscode uses to mark the error in IDE, but without a lot of parsing, it's difficult to get the line number and column from that.

iartemiev:
Great job, Praveen! I can't wait to start using this!

I am having some trouble with utilizing the resulting package in a sample app though, maybe I'm missing a step?

Here are the steps I took:

1. `hub pr checkout 4358`, made sure I'm in the root
2. `yarn clean`
3. `yarn build:esm:watch` --> all packages build successfully and are watching for changes
4. From another terminal tab: `yarn link-all` --> success


Sample app:

Contains App.js with
```
import Amplify, { Analytics } from 'aws-amplify';
...
```
Running commands:
1. `yarn remove aws-amplify` to remove the package that was installed originally via npm
2. `yarn link aws-amplify`
3. `yarn start`

Result:

> Failed to compile.
> 
> Error in /Users/ivan/dev/work/amplify-js/packages/aws-amplify/index.js
> Module not found: ./dist/aws-amplify.js in /Users/ivan/dev/work/amplify-js/packages/aws-amplify
> 
>  @ /Users/ivan/dev/work/amplify-js/packages/aws-amplify/index.js 6:18-50

scripts/build.js
ericclemmons: You're right, apparently the properties aren't immediately available on the `diagnostic`.

Based on the wiki, would this work?

```js
      let message = ts.flattenDiagnosticMessageText(diagnostic.messageText, "\n");
      if (diagnostic.file) {
        let { line, character } = diagnostic.file.getLineAndCharacterOfPosition(
          diagnostic.start!
        );
        console.log(
          `  Error ${diagnostic.file.fileName} (${line + 1},${character +
            1}): ${message}`
        );
      } else {
        console.log(`  Error: ${message}`);
      }
```

scripts/build.js
ericclemmons: `console.debug` is an alias for `console.log`:
> https://github.com/nodejs/node/blob/master/lib/internal/console/constructor.js#L582-L585

(IIRC, its for parity with the browsers' `console` object)

We could use our internal `Logger`, but it would need to a way to support changing level threshold from `process.env`:
> https://aws-amplify.github.io/docs/js/logger#setting-logging-levels

Amplifiyer: I tried with the `winston` logger, so far it's working good, but changing the log level would require changing it in the script or perhaps I could pass it as a command line switch to build command. With our internal logger, is there a way to change the log level on command line so node picks it up?

scripts/build.js
Amplifiyer: yes, I realized it later that I was using this for regular builds. I refactored to use this for both types of build.

scripts/build.js
ericclemmons: @Amplifiyer Since this is a CLI script, we can probably do it safely in `scripts/build.js`:

```js
const logger = new Logger('build', process.env.LOG_LEVEL)
```

`Logger` defaults to `WARN`, so this should clean up logs by default, and give us an escape hatch for seeing debug messages via `LOG_LEVEL=DEBUG yarn build`:

> https://github.com/aws-amplify/amplify-js/blob/f37cff7a099b9f2463239b971e6bbefef7d204c9/packages/core/src/Logger/ConsoleLogger.ts#L36

Honestly, this PR is such an improvement already I'm totally cool with shipping & adding to it as DX optimization!

Amplifiyer: The reason I didn't end up using our own logger was because it's in core package which is most likely not built when we are using it in the script and our src is in typescript and our script is in nodejs :(

But good point on using process.env.LOG_LEVEL variable, I added it in the winston logger.

PR #4372, opened 2019-11-09
chore:Prettier 1.19.0

manueliglesias:
_Issue #, if available:_
N/A

_Description of changes:_
Upgrade to prettier 1.19.0 so we can start using TypeScript 3.7 features without prettier complaining.

Tested with a small usage of `??` and verifying that the pre-commit hook handled it fine

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -31,9 +31,7 @@
     "url": "https://github.com/aws/aws-amplify/issues"
   },
   "homepage": "https://aws-amplify.github.io/",
-  "devDependencies": {
-    "prettier": "^1.7.4"
-  },
+  "devDependencies": {},
   "dependencies": {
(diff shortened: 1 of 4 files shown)


packages/api/package.json
```diff
@@ -31,9 +31,7 @@
     "url": "https://github.com/aws/aws-amplify/issues"
   },
   "homepage": "https://aws-amplify.github.io/",
-  "devDependencies": {
-    "prettier": "^1.7.4"
-  },
+  "devDependencies": {},
```
Amplifiyer: can we just delete this?

manueliglesias: Sure, I just left it the way yarn left it after running a `yarn remove prettier` from the packages with:

```
npx lerna exec yarn remove prettier
```

PR #4399, opened 2019-11-14
fix(@aws-amplify/auth): throw error on handleAuthResponse catch block

### LINKED ISSUE #4004: Wrong error reported from signIn_failure on 1.1.39-unstable.4
Hi amplify team, I participated in this issue #3932 a few days ago.
I created this new issue because the previous one was closed by this pr # 3993.

**Describe the bug**
I'm testing the 1.1.39-unstable.4 version of the aws-amplify package and the bug is kind of fixed, because there is no longer a react white screen for unhandled rejection, but the catch on OAuth/OAuth.ts on handleAuthResponse method is not throwing the error, just logging. 
What happens there is that the method returns undefined on Auth.ts line 1564 and that triggers the catch on private _handleAuthResponse method on Auth.ts with a new error (because javascript can't get accessToken from undefined). That's the error we get on signIn_failed event, so we lose the proper error_description that comes in the URL.

**To Reproduce**
1. Enter the amplify react app with a URL like: http://localhost:3000/somePage#error_description=PreSignUp+failed+with+error+SOME_CUSTOM_ERROR_FROM_PRE_SIGNUP_TRIGGER.+&state=someStateHash&error=invalid_request
2. Catch signIn_failure event, the data.message there is: "Cannot read property 'accessToken' of undefined"
3. The console logs the correct error, the same as error_description.

**Expected behavior**
I think the proper behavior should be to log the error on OAuth.ts handleAuthResponse but also throw it, so Auth.ts private _handleAuthResponse gets it in the catch and triggers the failure events with the proper error, that way I can catch it and give some feedback to the user regarding that custom error. That would also prevent the undefined return.

**Additional context**
Same as #3932 but with aws-amplify@1.1.39-unstable.4

rabellino12:
_Issue #4004 

_Description of changes:_

This pr closes #4004, it contains a quick fix for a bug preventing the correct error from being catched
somewhere else in the application, because handleAuthResponse returned undefined in case of error. Now with the `throw e` line, the correct error is sent.


HMilbradt:
Is something holding up merging this PR?  We're currently running into this

ribeirojose:
Does anyone suggest a workaround for this before the PR is merged?

rabellino12:
Thank you too @ericclemmons ! :+1:

PR #4400, opened 2019-11-14
chore: (aws-amplify@2.0.0) Moving AWS Amplify JS to 2.0.0

sammartinez:
_Description of changes:_
Moving packages to 2.0.0 as follows:
 - @aws-amplify/analytics@2.0.0
 - @aws-amplify/api@2.0.0
 - @aws-amplify/auth@2.0.0
 - aws-amplify-angular@4.0.0
 - aws-amplify-react-native@3.0.0
 - aws-amplify-react@3.0.0
 - aws-amplify-vue@1.0.0
 - aws-amplify@2.0.0
 - @aws-amplify/cache@2.0.0
 - @aws-amplify/core@2.0.0
 - @aws-amplify/interactions@2.0.0
 - @aws-amplify/predictions@2.0.0
 - @aws-amplify/pubsub@2.0.0
 - @aws-amplify/pushnotification@2.0.0
 - @aws-amplify/storage@2.0.0
 - @aws-amplify/xr@1.0.0

### DIFF
--- a/packages/xr/package.json
+++ b/packages/xr/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@aws-amplify/xr",
-  "version": "0.2.4",
+  "version": "1.0.0",
   "description": "XR category of aws-amplify",
   "main": "./index.js",
   "module": "./lib-esm/index.js",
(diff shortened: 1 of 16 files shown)


packages/xr/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@aws-amplify/xr",
-  "version": "0.2.4",
+  "version": "1.0.0",
```
jordanranz: Just making a note that we may not need to update this if we don't have to.

sammartinez: Looks to be that even from doing this from individual packages it does force a major version for all

Issue #4089, opened 2019-09-26
support RN-0.60.+ for @aws-amplify/pushnotification


bernhardt1:
**Describe the bug**
There are many issues with this library when installing it in a RN0.60.+ project. The worst part is, none of these are mentioned in the installation docs. 

The vast majority of other libraries provide detailed instructions for installation depending on the projects RN version. When using RN0.60.+ they almost all provide support for autolinking. For those using versions less than 0.59 they provide different instructions.

It is unclear how to incorporate push notifications into a project using RN-0.60.+ or if it is even possible.

It is very frustrating to follow a guide that leads to a failed state. RN-0.60.+ has been released for months. Please update the docs to let people know of the libraries limitations.

**To Reproduce**
Try to add push notifications to a RN0.60.+ project.

**Expected behavior**
- Autolinking for both iOS and android
- compatibility with RN0.60.+
- documentation of libraries limitations

**Screenshots**
NA

**Desktop (please complete the following information):**
NA

**Smartphone (please complete the following information):**
NA

**Additional context**
NA

**Sample code**
NA

**_You can turn on the debug mode to provide more info for us by setting window.LOG_LEVEL = 'DEBUG'; in your app._**

bernhardt1:
Update: Android is working without explicitly linking the package.

I was able to skip much of the documentation to get notifications working on android.
parts of step 3, 6,  and 7.

The iOS documentation seems to have similar problems.

bneigher:
Yea amplify-js is getting very antiquated regarding it's support on modern react native.  I'm about ready to remove it from my app.   Feels like the web is treated more like a first class citizen and react native is an afterthought

bernhardt1:
@bneigher 

I've thought about that. AWS is pretty integrated into my app at this point though. Authentication, storage, push notifications. Not to mention appsync and lambda.

If you decide to remove it let me know. Would be interesting to talk about alternative options. @ me here or twitter @JasonBernhardt1

MaffooBristol:
The library forces the use of the bundled RN push notifications library and every single workaround I've tried has not worked. It seems it simply needs to be updated to allow the `@react-native-community` version of the library to work. Hopefully there can be some solution soon because I'm trying to integrate amplify with a new RN-0.61 project and I'd rather not have to downgrade RN just for this. I'd probably go with firebase for both ios and android if it can't be resolved soon 😞

MaffooBristol:
[...]

Issue #3995, opened 2019-09-10
Push notification tokens are not generated for react-native version 0.60.5 with production certs


sandaruny:
**Describe the bug**
I experienced that the PushNotification.onRegister is not called when using react-native version 60.5 with production Apple Push Services certs. This might be an issue arising with the deprecation of the PushNotificationIOS native module. ([https://facebook.github.io/react-native/docs/pushnotificationios](https://facebook.github.io/react-native/docs/pushnotificationios))

**To Reproduce**
1. Create env with the react-native 0.60.5 version. 
2. Check PushNotification.onRegister for a production build

**Expected behavior**
The token should be generated

This was working for the previous react-native version (0.55.0) and generating token successfully.

But it was able to fix by patching the module. When the library code was modified to use the "@react-native-community/push-notification-ios", the tokens were generating properly.

elorzafe:
@sandaruny there was a bunch of breaking changes on react-native 0.60, we will take this a feature-request that we want to support soon.

Thanks for pointing this out!

nikhil-dabhade:
@Ashish5591 : Any insights in this issue?

matamicen:
@elorzafe 
It works for us in TestFlight ... we didn't launch production. @sandaruny  @nikhil-dabhade  it works for you on TestFlight also? 

Thanks for the info.
Matt.

sammartinez:
Resolving as the reference issue above from @Ashish5591 resolves this.

Ashish5591:
Hi  @sandaruny and @matamicen  
We just did a release to the @unstable tag with the changes.
Please take a look at this [comment](https://github.com/aws-amplify/amplify-js/issues/4089#issuecomment-559333746) for the instructions and let us know if it works for you.

PR #4401, opened 2019-11-14
Adding support for push notifications in iOS for React Native 0.60+(#…

Ashish5591:
…4215)

_Issue #:_
#3995
#4089

_Description of changes:_
Adding support for push notifications in iOS for React Native 0.60+

### DIFF
--- a/packages/pushnotification/package.json
+++ b/packages/pushnotification/package.json
@@ -27,6 +27,7 @@
   },
   "homepage": "https://aws-amplify.github.io/",
   "devDependencies": {
+    "@react-native-community/push-notification-ios": "^1.0.2",
     "@types/jest": "^20.0.8",
     "@types/node": "^8.10.15",
     "awesome-typescript-loader": "^3.2.2",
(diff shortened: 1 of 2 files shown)


bernhardt1:
@Ashish5591 

Are push notifications going to be getting proper attention in the near future?

I'll preface what I'm about to say by telling you that I don't want this to come across as negative complaining...

Support for this library has been terrible for the last 5 months. It is incredibly unreliable yet it is advertised as a reliable push notification service to be used in production.

There are quite a few issues around supporting RN 0.60.+. Some of them are surprisingly old, yet they are not getting any attention:
- **the lag time for fixing issues has been in the order of months (RN-0.60.+ came out 5 months ago!**
- the documentation hasn't been updated in many months (maybe even a year? - it is incredibly old and misleading)
- the documentation/code doesn't even support auto-linking. Literally, every other package I use (20-30 packages), most of which **do not cost money**, have fully supported RN0.60.+ for months. Yet this library which is used as part of a monetized service does not.

My main question is, will this library get proper attention in the coming months. I would prefer to keep my using AWS for push notifications but if the library is not going to be supported properly then I need to know so I can take appropriate steps for my project.

dantasfiles:
@bernhardt1 One thing that would help is a sample app. There's good sample apps for Auth, API, etc. but I haven't been able to find any for Push Notifications. Especially considering the documentation for Push Notifications is currently very sparse.

cinder92:
@bernhardt1 Proper steps, its, 

1 .- Fork this git
2 .- Fix the issues by yourself (with your own time)
3 .- Integrate it into your existing app
4 .- Keep calm and code

:)

Issue #4259, opened 2019-10-25
S3 is not a constructor & Observable is not a constructor ERRORS


orcunorcun:
### Problem:
I was using Storage and subscription well on React Native, but now it doesn't work with same codes. It rejects with 

> _S3 is not a constructor_

 for Storage,  

> _Observable is not a constructor_

for Subscription (Queries and mutations working well.)

### Packages:
```
  "react": "16.9.0",
  "react-native": "0.61.2",
  "amazon-cognito-identity-js": "3.1.3",
  "aws-amplify": "1.2.3",
  "aws-amplify-react-native": "2.2.3",
```
### Codes:
```
import awsconfig from './aws-exports';
import Amplify, {Storage} from 'aws-amplify';

Amplify.configure(awsconfig);

export const ImageDownloader = fileName => {
  return Storage.get(fileName, {track: true});
};
```

Nitheesh-u:
Same error for me "Object is not a constructor (evaluating 'new S3')"

rarira:
same error @aws-amplify/storage.. v1.2.2 is fine.

Amplifiyer:
@orcunorcun @Nitheesh-u @rarira I pushed a quick fix for that which is now published to unstable, can you please test it with this version(unstable tag) and let us know if that solved this issue?

itsgitz:
@Amplifiyer well, I've changed the version with unstable tag:
 `"aws-amplify": "^1.2.3-unstable.0"`

And then new warning has occured:
`[Unhandled promise rejection: Error: No credentials, applicationId or region]`

So, I disable `Analytics` like this:
`Amplify.configure({
  Analytics: {
    disabled: true
  }
})`

And it works for both S3 and Subscriptions.

adamsythe:
why is this closed?

itsgitz:
@adamsythe  I don't why, but I'm still waiting for update. My application can't `subscribe` using `aws-amplify` with unstable version

Amplifiyer:
@itsgitz @adamsythe, can you please post your error messages and code samples here that is causing the issue? Please also verify the unstable version of aws-amplify packages you are consuming in the node_modules folder.

itsgitz:
@Amplifiyer for now, there's no problem with S3 but I got this error/warning messages, `and` I can't publish message using MQTT:
![ss](https://user-images.githubusercontent.com/20938898/67733314-995e9e80-fa30-11e9-8fd0-ecad3b418cbb.png)

[...]

PR #4404, opened 2019-11-15
Remove require statements

jordanranz:
_Description of changes:_
- Remove require statements due to issues with mixing import/require in Rollup
- Use genversion dev module to generate module exporting package version number
- Prettier to core package.json


packages/auth/src/OAuth/OAuth.ts
```diff
@@ -23,8 +23,8 @@ import {
 
 import { ConsoleLogger as Logger, Hub } from '@aws-amplify/core';
 
-const SHA256 = require('crypto-js/sha256');
-const Base64 = require('crypto-js/enc-base64');
+import * as sha256 from 'crypto-js/sha256';
+import Base64 from 'crypto-js/enc-base64';
```
Amplifiyer: Do we need to change for base64 import as well? Also the installation guide https://github.com/brix/crypto-js suggests to just use default import and not wild card import. I'm worried if it will cause issues in react native.

jordanranz: Using the default export was what caused the unit tests to break. Something like `sha256_1.default is undefined`. I can change Base 64 for consistency. I'll also time-box a bit more time to investigate the sha256 error I was getting.

Have you seen wild card imports break our React Native package in the past?

Amplifiyer: Yes, this is what caused the recent react native regression when publishing ES6 modules https://github.com/aws-amplify/amplify-js/issues/4259

jordanranz: ok. I was able to revert back to using the default import `import sha256 from 'crypto-js/sha256`. Do you forsee a problem with this as well or just the wildcard import?

Amplifiyer: this is great, and I think the preferred way. Thanks!

PR #4412, opened 2019-11-15
feat(@aws-amplify/ui-components): amplify-greetings/amplify-nav

jordanranz:
_Description of changes:_
- `amplify-greetings` component
- `amplify-nav` component
- Rebase from #4412

[Delivers: #167797513]

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-greetings/amplify-greetings.spec.ts
+++ b/packages/amplify-ui-components/src/components/amplify-greetings/amplify-greetings.spec.ts
@@ -9,12 +9,16 @@ describe('amplify-greetings spec:', () => {
 			greetings = new AmplifyGreetings();
 		});
 
-		it('should render `Submit` for the button text', () => {
+		it('should have user prop set to null by default', () => {
 			expect(greetings.user).toBe(null);
 		});
 
-		it('should have style override prop set to false by default', () => {
+		it('should have logo prop set to null by default', () => {
 			expect(greetings.logo).toBe(null);
+		});
+
+		it('should have overrideStyle prop set to false by default', () => {
+			expect(greetings.overrideStyle).toBe(false);
 		});
 	});
 	describe('Render logic ->', () => {
--- a/packages/amplify-ui-components/src/components/amplify-greetings/amplify-greetings.spec.ts
+++ b/packages/amplify-ui-components/src/components/amplify-greetings/amplify-greetings.spec.ts
@@ -1,0 +1,30 @@
+import { newSpecPage } from '@stencil/core/testing';
+import { AmplifyGreetings } from './amplify-greetings';
+
+describe('amplify-greetings spec:', () => {
+	describe('Component logic ->', () => {
+		let greetings;
+
+		beforeEach(() => {
+			greetings = new AmplifyGreetings();
+		});
+
+		it('should render `Submit` for the button text', () => {
+			expect(greetings.user).toBe(null);
+		});
+
+		it('should have style override prop set to false by default', () => {
+			expect(greetings.logo).toBe(null);
+		});
+	});
+	describe('Render logic ->', () => {
+		it('should render', async () => {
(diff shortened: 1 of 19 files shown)


packages/amplify-ui-components/src/components/amplify-greetings/amplify-greetings.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+
+		beforeEach(() => {
+			greetings = new AmplifyGreetings();
+		});
+
+		it('should render `Submit` for the button text', () => {
+			expect(greetings.user).toBe(null);
+		});
+
+		it('should have style override prop set to false by default', () => {
```
sammartinez: Looking at the test the spec for the `it` doesnt make sense, can we update please

packages/amplify-ui-components/src/components/amplify-greetings/amplify-greetings.style.ts
```diff
@@ -0,0 +1,10 @@
+import { css } from 'emotion';
+
+export const greetings = css`
+  position: relative;
+  border: 1px solid transparent;
+  border-color: #e7e7e7;
```
sammartinez: I believe we have css variables for this, can we update please

packages/amplify-ui-components/src/components/amplify-greetings/amplify-greetings.tsx
```diff
@@ -0,0 +1,37 @@
...
+  watchHandler(newUserValue: CognitoUserInterface) {
+    // If the user passed in has a username, show username greeting message
+    if (newUserValue && newUserValue.username) this.showUsernameGreeting = true;
+  }
+
+  @State() showUsernameGreeting: boolean = false;
+
+  render() {
+    return (
+      <header class={greetings}>
```
sammartinez: Are we not wanting to use style nuker? The snapshot only has the emotion css hash class so that is why I was curious

jordanranz: I will bring that in :D Was focusing on getting the out

packages/amplify-ui-components/src/components/amplify-nav/amplify-nav.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { nav } from './amplify-nav.style';
+
+@Component({
+  tag: 'amplify-nav',
+  shadow: false,
+})
+export class AmplifyNav {
+  render() {
+    return (
+      <nav class={nav} role="navigation">
```
ericclemmons: No need for `role="navigation"`, since you're using the `nav` tag:
> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role

packages/amplify-ui-components/src/components/amplify-greetings/amplify-greetings.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+@Component({
+  tag: 'amplify-greetings',
+  shadow: false,
+})
+export class AmplifyGreetings {
+  /** Used for the username to be passed to resend code */
+  @Prop() user: CognitoUserInterface = null;
+  /** Logo displayed inside of the header */
+  @Prop() logo: FunctionalComponent | null = null;
```
ericclemmons: I assume this is so that `<amplify-greetings logo={null} />` is valid? (vs. being declared as `@Prop() logo?: FunctionalComponent`)

jordanranz:
@ericclemmons, I will add stories for these components and play around with the changes you made. Thanks for the reminder.

packages/amplify-ui-components/src/components/amplify-greetings/amplify-greetings.tsx
jordanranz: Correct

PR #4421, opened 2019-11-18
feat(@aws-amplify/ui-components): amplify-require-new-password component

jordanranz:
_Description of changes:_
- `<amplify-require-new-password>` component

[Delivers: #168413541]


packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -0,0 +1,109 @@
...
+  @Prop() headerText: string = CHANGE_PASSWORD;
+  /** The text displayed inside of the submit button for the form */
+  @Prop() submitButtonText: string = CHANGE;
+  /** The form fields displayed inside of the forgot password form */
+  @Prop() formFields: FormFieldTypes;
+  /** (Optional) Overrides default styling */
+  @Prop() overrideStyle: boolean = false;
+  /** The function called when submitting a new password */
+  @Prop() handleSubmit: (event: Event) => void = event => this.change(event);
+  /** Passed from the Authenticatior component in order to change Authentication state */
```
ericclemmons:
```suggestion
  /** Passed from the Authenticator component in order to change Authentication state */
```

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -49,18 +49,14 @@ export class AmplifyAuthenticator {
         return <amplify-sign-in federated={this.federated} handleAuthStateChange={this.onAuthStateChange} />;
       case AuthState.ConfirmSignIn:
         return <amplify-confirm-sign-in handleAuthStateChange={this.onAuthStateChange} user={this.authData} />;
-      case AuthState.SignOut:
-        // TODO: add sign out component
-        return <div>Sign Out Component</div>;
```
ericclemmons: This disappeared because of the greetings component, correct?

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -0,0 +1,109 @@
...
+  @Prop() handleAuthStateChange: (nextAuthState: AuthState, data?: object) => void;
```
ericclemmons: Do you think it's time we abstract this into one of the auth types?

```ts
type AuthStateHandler = (nextAuthState: AuthState, data?: object) => void;
```

```suggestion
  @Prop() handleAuthStateChange: AuthStateHandler;
```

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -0,0 +1,109 @@
...
+  @State() username: string;
+  @State() password: string;
+  @State() code: string;
+
+  componentWillLoad() {
+    this.formFields = [
+      {
+        type: 'password',
+        required: true,
+        handleInputChange: event => this.handlePasswordChange(event),
```
ericclemmons: I finally understand why there are so many anonymous, scoped closures in these components:


```suggestion
        handleInputChange: this.handlePasswordChange,
```

The handler functions themselves aren't scoped:

```js
handlePasswordChange(event) {
  this.password = event.target.value;
}

// vs.
handlePasswordChange = (event) => {
  this.password = event.target.value;
}
```

If they were, then almost all extraneous closures could be removed, except for situtations where the event & the handler don't share 1-to-1 positional parameters:

```js
handleInputChange: event => this.updatePassword(event.target.value)
```

ericclemmons: This is a common scenario for JSX/React users, and even has a mention regarding preferences & performance here:
> https://reactjs.org/docs/handling-events.html

Nothing specific to change: this is primarily subjective unless performance data shows otherwise.  Personally, I'm disappointed that `class` methods aren't automatically scoped to `this`, which is a major stumbling point for newcomers to JavaScript.

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -0,0 +1,109 @@
...
+  checkContact(user) {
+    if (!Auth || typeof Auth.verifiedContact !== 'function') {
+      throw new Error(NO_AUTH_MODULE_FOUND);
+    }
+    Auth.verifiedContact(user).then(data => {
+      if (!isEmpty(data.verified)) {
+        this.handleAuthStateChange(AuthState.SignedIn, user);
+      } else {
+        user = Object.assign(user, data);
+        this.handleAuthStateChange(AuthState.VerifyContact, user);
```
ericclemmons: Mutating data can often have unknown side-effects, especially in the future, when other methods may depend on `user` in the originating promise-chain.  (`user` is passed by reference)

Is there a reason the reference is being mutated, or would the following do?

```suggestion
        this.handleAuthStateChange(AuthState.VerifyContact, { ...user, ...data });
```

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { CHANGE_PASSWORD, CHANGE, BACK_TO_SIGN_IN, NO_AUTH_MODULE_FOUND } from '../../common/constants';
+// import { CodeDeliveryType } from './amplify-forgot-password-interface';
+
+import { Auth } from '@aws-amplify/auth';
+import { Logger, isEmpty } from '@aws-amplify/core';
+
+const logger = new Logger('amplify-require-new-password');
+
+@Component({
+  tag: 'amplify-require-new-password',
```
ericclemmons: What was the rationale for renaming to `require-new-password` vs. common phrases such as `reset-password`, or `password-confirmation`?

packages/amplify-ui-components/src/common/constants.ts
```diff
@@ -35,6 +35,8 @@ export const PHONE_LABEL = 'Phone Number *';
 export const PHONE_PLACEHOLDER = '(555) 555-1212';
 
 // Password common constants
+export const CHANGE_PASSWORD = 'Change Password';
+export const CHANGE = 'Change';
```
ericclemmons: `CHANGE` is ambiguous. I'd recommend `CHANGE_PASSWORD_ACTION` or something that explicitly differentiates changing email/password/name/etc.

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -0,0 +1,109 @@
+import { Component, Prop, State, h } from '@stencil/core';
+import { FormFieldTypes } from '../amplify-auth-fields/amplify-auth-fields-interface';
+import { AuthState, ChallengeName, CognitoUserInterface } from '../../common/types/auth-types';
+import { CHANGE_PASSWORD, CHANGE, BACK_TO_SIGN_IN, NO_AUTH_MODULE_FOUND } from '../../common/constants';
+// import { CodeDeliveryType } from './amplify-forgot-password-interface';
```
ericclemmons: Is this a `TODO`, or something to be removed?

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
jordanranz: correct :D

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
jordanranz: I think this is a good idea. I can make a follow-up PR for this.

packages/amplify-ui-components/src/common/constants.ts
jordanranz: Fair, I can make this change.

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
jordanranz: `require-new-password` is a specific state from Cognito when an admin creates an account for a user. When the user logs in for the first time, they are required to submit a new password. This component encompasses that use case. In the primary action function you can see the use of `completeNewPassword`

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
jordanranz: Changed (this seems to be misspelled in many other files. I'll add this to a follow-up task.

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
jordanranz: This was pulled directly from the React component, let me see if I can execute your suggested change. I think your solution is definitely cleaner.

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
jordanranz: agreed. I will add a note to investigate the performance difference here. Do you have a strong concern against the current use of anonymous, scoped closures?

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
jordanranz: Removed (draft PR ftw)

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
jordanranz: The anon function usage we have in our code is based off of this guide:
https://stenciljs.com/docs/forms#advanced-forms

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
jordanranz: The `checkContact` function is actually used in a few components and can be abstracts into a common place. I'll add this to the list of follow-up items.

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
ericclemmons: No concerns without explicit data to justify the change. 

Just pointing out TIL moments for me :)

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -0,0 +1,111 @@
...
+        user = Object.assign(user, data);
+        this.handleAuthStateChange(AuthState.VerifyContact, user);
+      }
+    } catch (error) {
+      logger.error(error);
+      throw new Error(error);
+    }
+  }
+
+  async change(event) {
```
sammartinez: `change` seems to generic, can we rename this to be more specific?

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -0,0 +1,111 @@
...
+      const user = await Auth.completeNewPassword(this.user, this.password, attrs);
+      logger.debug('complete new password', user);
+      if (user.challengeName === ChallengeName.SMSMFA) {
+        this.handleAuthStateChange(AuthState.ConfirmSignIn, user);
+      } else if (user.challengeName === ChallengeName.MFASetup) {
+        logger.debug('TOTP setup', user.challengeParam);
+        this.handleAuthStateChange(AuthState.TOTPSetup, user);
+      } else {
+        this.checkContact(user);
+      }
```
sammartinez: Could we adjust this to a switch?

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
jordanranz: yep, good idea.

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -0,0 +1,116 @@
...
+  tag: 'amplify-require-new-password',
+  shadow: false,
+})
+export class AmplifyRequireNewPassword {
+  /** The header text of the forgot password section */
+  @Prop() headerText: string = CHANGE_PASSWORD;
+  /** The text displayed inside of the submit button for the form */
+  @Prop() submitButtonText: string = CHANGE_PASSWORD_ACTION;
+  /** (Optional) Overrides default styling */
+  @Prop() overrideStyle: boolean = false;
```
sammartinez: NIT: can we make this optional? just missing the `overrideStyle?:`

jordanranz: Doesn't having a default make it optional?

jordanranz: I think we can make this change across the board. I'll it to the follow-up if we need it.

PR #4422, opened 2019-11-18
Fix for negative BigIntegers (toString())

killdash9:
Without this fix, a negative bigint will always end up throwing `new Error('Only radix 2, 4, 8, 16, 32 are supported')` because b is undefined.  This same fix is present in the original code from which this file was derived: http://www-cs-students.stanford.edu/~tjw/jsbn/jsbn.js

_Issue #, if available:_
None, trivial change.

_Description of changes:_
The first line of `bnToString` calls toString() with no radix parameter, when the number is negative.  The toString function is set equal to bnToString, which requires a radix parameter.  This fix adds the radix parameter.

### DIFF
--- a/packages/amazon-cognito-identity-js/src/BigInteger.js
+++ b/packages/amazon-cognito-identity-js/src/BigInteger.js
@@ -219,7 +219,7 @@ function bnpClamp() {
 
 // (public) return string representation in given radix
 function bnToString(b) {
-	if (this.s < 0) return '-' + this.negate().toString();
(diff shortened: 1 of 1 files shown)


killdash9:
A pull request identical to this one was submitted independently 1 day after this one: #4427

KesleyDavid:
Same problem with me, any predictions for the update?

cam8001:
Bump - would love this to be committed.

ericclemmons:
This PR has been pulled into #4520 & is getting ready for release this afternoon.

Stay tuned!

Issue #4387, opened 2019-11-13
@aws-amplify/api: export 'GraphQLError' was not found in 'graphql'


piraka9011:
**Describe the bug**
When I try to build my typescript react app, I get the following error:
```
ERROR in ./node_modules/@aws-amplify/api/lib-esm/API.js 543:40-52
"export 'GraphQLError' was not found in 'graphql'
 @ ./node_modules/@aws-amplify/api/lib-esm/index.js
 @ ./node_modules/aws-amplify/lib/index.js
 @ ./src/pages/ForgotPassword/index.tsx
 @ ./src/routes.ts
 @ ./src/components/Routes.tsx
 @ ./src/App.tsx
 @ ./src/client/index.tsx
 @ multi regenerator-runtime/runtime ./src/client/index.tsx

ERROR in ./node_modules/@aws-amplify/api/lib-esm/API.js 560:41-53
"export 'GraphQLError' was not found in 'graphql'
 @ ./node_modules/@aws-amplify/api/lib-esm/index.js
 @ ./node_modules/aws-amplify/lib/index.js
 @ ./src/pages/ForgotPassword/index.tsx
 @ ./src/routes.ts
 @ ./src/components/Routes.tsx
 @ ./src/App.tsx
 @ ./src/client/index.tsx
 @ multi regenerator-runtime/runtime ./src/client/index.tsx

ERROR in ./node_modules/@aws-amplify/api/lib-esm/API.js 607:65-77
"export 'GraphQLError' was not found in 'graphql'
 @ ./node_modules/@aws-amplify/api/lib-esm/index.js
 @ ./node_modules/aws-amplify/lib/index.js
 @ ./src/pages/ForgotPassword/index.tsx
 @ ./src/routes.ts
 @ ./src/components/Routes.tsx
 @ ./src/App.tsx
 @ ./src/client/index.tsx
 @ multi regenerator-runtime/runtime ./src/client/index.tsx
```
It looks like aws-amplify can't find graphql components.

**To Reproduce**

1. `npm run install`
2. `npm run dev` (`ts-node -T -O '{\"module\": \"commonjs\"}' --transpileOnly internal/scripts/build -- --optimize`)

RyanWarner:
Hey @piraka9011 have you made any progress figuring this out? I've been blocked by this for a few days and nothing I've tried has helped.

piraka9011:
[...]

PR #4425, opened 2019-11-19
fix(@aws-amplify/api): fix for issue #4387, graphql version 14.5.8

BennettDixon:
_Issue #4387 , if available:_

_Description of changes:_ fixed bug that was causing GraphQLError to not be imported properly for some prior releases of GraphQL


BennettDixon:
I ultimately realized that a bug in my web-pack configuration was causing this issue. I'm not sure this should be merged because of this. However, being more explicit in the import syntax is not necessarily a bad thing.

PR #4427, opened 2019-11-19
fix(@aws-amplify/amazon-cognito-identity): Add missing required param to string conversion fn for negative big integers

antweer:
The bnToString() function throws an exception when called without a radix value or given a radix outside of [2, 4, 8, 16, 32]. 

When bnToString() is called on a negative big int value, it's negated value is used to call bnToString() which then gets concatenated to a '-'. However, in the bnToString() call for the negated value, no radix value is passed in so it always throws an exception. 

This change will use the radix value argument from the initial bnToString() call on the negative big integer value when calling bnToString() for it's negated value so that it works as intended.

_Issue #, if available:_

_Description of changes:_
BigInteger.bnToString() throws errors on negative values. Include the radix param when calling bnToString() recursively for negative big ints.

### DIFF
--- a/packages/amazon-cognito-identity-js/src/BigInteger.js
+++ b/packages/amazon-cognito-identity-js/src/BigInteger.js
@@ -219,7 +219,7 @@ function bnpClamp() {
 
 // (public) return string representation in given radix
 function bnToString(b) {
-	if (this.s < 0) return '-' + this.negate().toString();
+	if (this.s < 0) return '-' + this.negate().toString(b);
 	var k;
 	if (b == 16) k = 4;
 	else if (b == 8) k = 3;


ofirm93:
Issue is fixed the same as in jsbn lib.
Please merge it.

killdash9:
This fix is identical to my pull request submitted one day earlier #4422

KesleyDavid:
Same problem with me, any predictions for the update?

antweer:
@Amplifiyer @ericclemmons Could y'all review this changeset whenever possible? Thanks

ericclemmons:
@antweer Thanks for the ping! I'm looking at this PR this morning, which is simple, but going through some tests to validate it.

How did you validate it on your end?

antweer:
@ericclemmons 
For end-to-end validation:
Cognito auth returns a negative salt string when you attempt to authenticate with a user email that doesn't exist in the specified user pool which causes the exception noted above to be raised instead of calling the auth validation fn that would return the intended error message describing why auth failed.

For comparison, you can attempt to use the cognito auth library to authenticate for a user that does exist in the user pool with an invalid password. The auth validation fails gracefully and provides a consumable error log to the client stating the user or pass is invalid. 

This should be the expected result for attempting to authenticate with an invalid user email as well where Cognito's response to the initial request in the auth sequence has a negative salt string. 

For functional validation: 
Using the BigInteger script from the changeset, create bigint objects with both positive and negative value. Then call their toString(radix) fn with a valid radix value (2,4,8,16,32). 

Without changeset:
The positive big integer returns a string value as expected and the negative big integer raises an exception.

With changeset:
Both big int objects return their string values as expected.

ericclemmons:
Thanks for the info @antweer!

I created a PR for this here: https://github.com/aws-amplify/amplify-js/pull/4520, including both your fix & @killdash9's (so that your efforts are counted in helping to resolve this).

Let me know if those tests match your experience, and we can get this resolved and hopefully prevent regressions in the future!

antweer:
@ericclemmons Looks good to me 👍 Thanks

ericclemmons:
This PR has been pulled into #4520 & is getting ready for release this afternoon.

Stay tuned!

antweer:
Changeset included in #4520

PR #4443, opened 2019-11-21
chore(@aws-amplify/ui-components): Use amplify-icon for federated button SVGs

ericclemmons:
[Delivers: #169785387]


packages/amplify-ui-components/src/components/amplify-icon/icons.tsx
```diff
@@ -1,58 +1,123 @@
...
+          />
         </g>
       </g>
     </svg>
   ),
-}
\ No newline at end of file
+};
+
+export type IconNameType = keyof typeof icons;
```
sammartinez: I like this. @jordanranz WDTY?

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
```diff
@@ -14,9 +15,11 @@ export class AmplifySignInButton {
     return (
       <Host class={cx(this.provider, classNames.Host)}>
         <button>
-          <span class="icon">
-            <slot name="icon" />
-          </span>
+          {this.provider in icons && (
```
jordanranz: Curious if we can rely on `amplify-icon` to handle this logic rather than importing the `IconNameType` and list of icons?

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
```diff
@@ -2,6 +2,7 @@ import { Component, h, Prop, Host } from '@stencil/core';
 import { cx } from 'emotion';
 
 import * as classNames from './amplify-sign-in-button.style';
```
jordanranz: Can we individually import the class names from this file instead of `import *`

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
```diff
@@ -14,9 +15,11 @@ export class AmplifySignInButton {
     return (
       <Host class={cx(this.provider, classNames.Host)}>
         <button>
-          <span class="icon">
-            <slot name="icon" />
-          </span>
+          {this.provider in icons && (
+            <span class="icon">
```
jordanranz: Is this class used to style the `span` here?

ericclemmons: Yes.

packages/amplify-ui-components/src/components/amplify-icon/icons.tsx
jordanranz: yea, nice improvement

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
ericclemmons: Ok. There's only one rule now (`Host`, see line 16), but I can split it up.

(This is pre-existing from the previous PR)

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
ericclemmons: It's necessary here because the `<span class="icon">` and `<slot />` needs to be short-circuited when the provider doesn't have an icon (e.g. the hosted UI OAuth button).

This is part of the trade-off with centralizing icons behind a single `icons` export: using a single icon means bundling _every_ icon.

Personally, I prefer the previous implementation since it avoids exactly this, but I assume there are benefits to using `amplify-icon`.

Let me know what you'd like to do, and we'll wrap up this PR 💪

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
jordanranz: With emotion shouldn't this be a javascript variable rather than a string?

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
jordanranz: Ok, we can revisit the icon discussion later! I was chatting with Harry about it as well. We may need a better way to encapsulate and modularize icons.

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
jordanranz: Right, I think we use `componentName` (or related part of the component) as the class name and import it like `import { componentName } from './component-name.style';` in other files. What do you think of that?

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
ericclemmons: `icon` isn't a separate `css` literal in the referenced `.style.js` file:
https://github.com/aws-amplify/amplify-js/blob/a76154a88da915d6326f1f85961b8c424499769b/packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.style.ts#L53-L56

Originally, I did start splitting them, but having the provider name (e.g. `google`) referenced in the markup seemed helpful.

Do you want all classes going through Emotion's `css`? 

This came from a near-direct port of https://github.com/aws-amplify/amplify-js/blob/ui-components/master/packages/amplify-ui/src/Button.css

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
ericclemmons: I'd love to be included on it!  I have experience with tree-shaking & bundle sizes, so I'm interested to understand the icon problems that exist today (or foreseeable challenges).

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
jordanranz: Got it, I see where you are coming from now. I think the class name as a string `icon` might conflict with user app classes. I wonder if we can just get rid of the `span` altogether and just have the selector use > amplify-icon

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
ericclemmons: I'll follow the model from the others then, thanks!

This is related to the discussion ~400px below, but it sounds like that's the next move to resolution here: break apart the CSS declarations into individual, `css` literals for unique classNames.

(Plus, having those class names being static often leads to unsanctioned API dependencies by users, where they use CSS to target them out of necessity vs. intentional design)

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
ericclemmons: Oh man, good call!  I keep forgetting that this renders an HTML element you can target!

I'll give that a shot.

Great feedback @jordanranz  💪 💅

ericclemmons: (Oh, and even if that doesn't work, all CSS conversation has led to not having static class names for explicit isolation)

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
ericclemmons: Updated!  The styles are pretty nested (e.g. `button` gets additional styles added form `button.google`), but they're all scoped behind the `signInButton` `css` literal, so now `styleNuker` works out of the box with it!

I even added a Story to validate it and, yep, all styles are 🍄💨 (mushroom cloud).

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
ericclemmons: See the comment here: https://github.com/aws-amplify/amplify-js/pull/4443#discussion_r349850542

I believe this is resolved, as evidence with `overrideStyle` support.

If we truly want to hide things like `google` and `icon`, we'll need to update `styleNuker` to run `cx` across multiple classes, or change the style declarations entirely to better fit the `css` format.

I think this strikes a pretty good balance without spending a ton of time rewriting it.

Issue #4389, opened 2019-11-13
An accessor cannot be declared in an ambient context - @aws-amplify - 1.2.5-unstable.5


vc-ishver:
ERROR in ·[96m../node_modules/@aws-amplify/core/lib-esm/ServiceWorker/ServiceWorker.d.ts·[0m:·[93m23·[0m:·[93m9·[0m - ·[91merror·[0m·[90m TS1086: ·[0mAn accessor cannot be declared in an ambient context. 
 
·[7m23·[0m     get serviceWorker(): ServiceWorker; 
·[7m  ·[0m ·[91m        ~~~~~~~~~~~~~·[0m 
·[96m../node_modules/@aws-amplify/pubsub/lib-esm/PubSub.d.ts·[0m:·[93m13·[0m:·[93m17·[0m - ·[91merror·[0m·[90m TS1086: ·[0mAn accessor cannot be declared in an ambient context. 
 
·[7m13·[0m     private get awsAppSyncProvider(); 
·[7m  ·[0m ·[91m                ~~~~~~~~~~~~~~~~~~·[0m 
·[96m../node_modules/@aws-amplify/pubsub/lib-esm/Providers/PubSubProvider.d.ts·[0m:·[93m9·[0m:·[93m19·[0m - ·[91merror·[0m·[90m TS1086: ·[0mAn accessor cannot be declared in an ambient context. 
 
·[7m9·[0m     protected get options(): ProvidertOptions; 
·[7m ·[0m ·[91m                  ~~~~~~~·[0m 
·[96m../node_modules/@aws-amplify/pubsub/lib-esm/Providers/MqttOverWSProvider.d.ts·[0m:·[93m12·[0m:·[93m9·[0m - ·[91merror·[0m·[90m TS1086: ·[0mAn accessor cannot be declared in an ambient context. 
 
·[7m12·[0m     get allClients(): string[]; 
·[7m  ·[0m ·[91m        ~~~~~~~~~~·[0m 
·[96m../node_modules/@aws-amplify/pubsub/lib-esm/Providers/MqttOverWSProvider.d.ts·[0m:·[93m18·[0m:·[93m19·[0m - ·[91merror·[0m·[90m TS1086: ·[0mAn accessor cannot be declared in an ambient context. 
 
·[7m18·[0m     protected get clientId(): any; 
·[7m  ·[0m ·[91m                  ~~~~~~~~·[0m 
[...]

PR #4452, opened 2019-11-22
fix: Downgrade TS to ~3.6

iartemiev:
_Issue #, if available:_
Fixes: https://github.com/aws-amplify/amplify-js/issues/4389

_Description of changes:_
TypeScript 3.7 introduced a feature change to declaration files which broke support with Angular versions supported by AmplifyJS. 

Downgrading the version to `~3.6.4` thereby floating the patch version but pinning the major and minor version to 3.6.

Reverting change to `packages/api/src/API.ts` that utilized nullish coalescing operator `??`

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
 		"tslint": "^5.7.0",
 		"tslint-config-airbnb": "^5.8.0",
 		"typedoc": "^0.11.0",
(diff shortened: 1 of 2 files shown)


iartemiev:
Tested this locally using Verdaccio and was able to confirm that the issue has been resolved

PR #4455, opened 2019-11-23
chore(@aws-amplify/ui-components): federated prop defaults to u…

ericclemmons:
[Delivers #169785203]

Removes default value of `@Prop() federated: {}` where possible.

Some places require it for backwards compatibility reasons, where `federated.oauth` is filled in on the user's behalf.

The good news is that this logic is entirely within federated components, vs. parent components.

---


packages/amplify-ui-components/src/components/amplify-sign-in/amplify-sign-in.tsx
```diff
@@ -170,7 +170,7 @@ export class AmplifySignIn {
       >
         <amplify-federated-buttons handleAuthStateChange={this.handleAuthStateChange} federated={this.federated} />
 
-        {!isEmpty(this.federated) && <amplify-strike>or</amplify-strike>}
+        {this.federated && <amplify-strike>or</amplify-strike>}
```
ericclemmons: The problem here is that https://github.com/aws-amplify/amplify-js/blob/c28851a6379f3ba05c4fef529bb5be7f18ebe316/packages/core/src/JS.ts#L53-L55 **is not** equivalent to `_.isEmpty`: this version fails on any non-object (e.g. `undefined`).

(`isEmpty` shouldn't exist in the library IMO, but that's another matter)

jordanranz: I agree that it shouldn't exist in the core lib.

If `this. federated` is passed in as an empty object the `strike` will still show up here correct?

ericclemmons: That's correct: if a user supplies `<amplify-authenticator federated={} />` that will be propagate through and trigger most of the federated UI (except the buttons themselves).

jordanranz: what do you think about using a proper isEmpty function maybe from [lodash](https://www.npmjs.com/package/lodash.isempty) or adding `instanceof Object` to the utility function

ericclemmons: I'd prefer to use `_.isEmpty`, but an easy fix without the bundle-size bloat could be to update our `static isEmpty(obj)` check to use `static isEmpty(object = {})`.  That way it works for both `undefined` and user-provided `{}` values.

(I mention bundle-size bloat because vanilla `lodash.isempty` does a ton more checks, and [it adds ~3k](https://bundlephobia.com/result?p=lodash.isempty@4.4.0)).

ericclemmons:
@jordanranz This is ready for re-review now.

ericclemmons:
@sammartinez @jordanranz Ready for review again.

packages/core/src/JS.ts
```diff
@@ -49,7 +49,7 @@ const MIME_MAP = [
 	{ type: 'message/rfc822', ext: 'eml' },
 ];
 
-export const isEmpty = obj => Object.keys(obj).length === 0;
+export const isEmpty = (obj = {}) => Object.keys(obj).length === 0;
```
jordanranz: Nice

PR #4463, opened 2019-11-25
React native predictions

dabit3:
_Issue #, if available:_

_Description of changes:_


ericclemmons:
Thanks for the PR @dabit3! It looks like this is adding some duck-typing to support files chosen within the React Native file picker, is that correct?

What's the best way for me to test this out for validation? Is there a sample/expo app I can reference?

PR #4478, opened 2019-11-27
Adding support for push notifications in iOS for React Native 0.60+

Ashish5591:
_Issue #:_
#3995
#4089

_Description of changes:_
Adding support for push notifications in iOS for React Native 0.60+

Tested these changes on an iPhone running iOS 12.x and a React Native 0.61.x app. Was able to receive push notifications through a pinpoint campaign and log the notification data in the onNotification handler.

### DIFF
--- a/packages/pushnotification/src/PushNotification.ts
+++ b/packages/pushnotification/src/PushNotification.ts
@@ -15,10 +15,10 @@ import {
 	NativeModules,
 	DeviceEventEmitter,
 	AsyncStorage,
-	PushNotificationIOS,
 	Platform,
 	AppState,
 } from 'react-native';
+import PushNotificationIOS from '@react-native-community/push-notification-ios';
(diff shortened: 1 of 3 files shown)


packages/pushnotification/src/PushNotification.ts
```diff
@@ -15,10 +15,10 @@ import {
 	NativeModules,
 	DeviceEventEmitter,
 	AsyncStorage,
-	PushNotificationIOS,
 	Platform,
 	AppState,
 } from 'react-native';
+import * as PushNotificationIOS from '@react-native-community/push-notification-ios';
```
sammartinez: Can we just verify with @Amplifiyer that this is ok? I know we were trying to go towards a different approach

Ashish5591: Updated the code and tsconfig to use default import

PR #3589, opened 2019-07-03
fix(@aws-amplify/analytics): use startsWith to filter the updateEndpoint error message

powerful23:
*Issue #, if available:*
Related #1378 
*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1378, opened 2018-08-03
Pinpoint Analytics Exceeded maximum endpoint per user count 10


Jun711:
Do you want to request a feature or report a bug?
seems like a misconfiguration issue? 

What is the current behavior?

Amplify analytics configuration
```
Analytics: {
    appId: '***',
    region: 'us-east-1',
    // endpointId: '***'
  }
```
When I had endpointId as a constant, I could only see one endpoint at Pinpoint. Thus, I got advice in gitter that I should let Amplify set the endpointId. Thus, I did that and I see this `Exceeded maximum endpoint per user count 10` error for some users:
```
400 Bad Request with pinpoint saying that
{"RequestID":"*","Message":"Exceeded maximum endpoint per user count 10"
```

What is the expected behavior?

Which versions of Amplify, and which browser / OS are affected by this issue? Did this work in previous versions?
0.4.3

You can turn on the debug mode to provide more info for us by setting ```window.LOG_LEVEL = 'DEBUG';``` in your app.
yes

powerful23:
The issue happens if your localStorage are cleaned too many times or you don't have localStorage so that the Amplify will generate a new endpointId every time.

Jun711:
@powerful23 
yea, I have it clean the localStorage specifically 2 cognito keys when I can't log in due to
Linked logins disabled on Federated Identities
https://github.com/aws-amplify/amplify-js/issues/1015

Jun711:
Hi @powerful23, for a website, should we have a fix endpoint? or let Amplify assigns the endpoint? Thanks

powerful23:
@Jun711 the endpoint stands for the browser in the device. When using amplify, it will automatically generate an Id and store it in your browser's localStorage. It will update the endpoint details like the browser version etc. If you don't want Amplify to assign the endpoint, you can configure it by yourself: https://aws-amplify.github.io/amplify-js/media/analytics_guide#manual-setup

Jun711:
@powerful23 ok, I got it. Previously, I assigned only one fixed ID in the Amplify config and thus, I saw only one endpoint at PinPoint console. Because of the limit of 10, users can see a lot of PinPoint errors if they have the console open.

TheVaporTrail:
@powerful23 I don't understand your previous comment: "...so I don't think the endpoint ." Is it missing a word? 

[...]

PR #4485, opened 2019-11-29
Handle undefined `message`

mariotsi:
From our remote logs we saw errors coming from this line `TypeError: Cannot read property 'startsWith' of undefined`. We are unable to check more deeply what was the response because we do not not log payloads. I'm making this change in order to make the code more failsafe

_Description of changes:_

Default the error message to empty string in order to make `.startsWith` not throwing in any case

### DIFF
--- a/packages/analytics/src/Providers/AWSPinpointProvider.ts
+++ b/packages/analytics/src/Providers/AWSPinpointProvider.ts
@@ -447,7 +447,7 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
 		failureData: EndpointFailureData
 	) {
 		const { err, update_params, endpointObject } = failureData;
-		const { message } = err;
+		const { message = '' } = err;
 		const { ApplicationId, EndpointRequest } = update_params;
 
 		if (!message.startsWith('Exceeded maximum endpoint per user count')) {
--- a/packages/analytics/src/Providers/AWSPinpointProvider.ts
+++ b/packages/analytics/src/Providers/AWSPinpointProvider.ts
@@ -445,10 +445,12 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
 		failureData: EndpointFailureData
 	) {
 		const { err, update_params, endpointObject } = failureData;
-		const { message = '' } = err;
+		const { message } = err;
 		const { ApplicationId, EndpointRequest } = update_params;
 
-		if (!message.startsWith('Exceeded maximum endpoint per user count')) {
+		if (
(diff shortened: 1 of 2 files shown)


mariotsi:
I see a Unit Test failing but I cannot see the link with my PR. Any help appreciated

ericclemmons:
@mariotsi Thanks for your contribution!  I'm going to do some validation around why `error.message` is `undefined`.

In either case, you're right that we shouldn't be calling `.startsWith` non a non-String.

ericclemmons:
Doing some research into this to determine why `err.message` would be empty to begin with: this `.startsWith` fix came from #3589 to address #1378.

We may need to update the test in https://github.com/aws-amplify/amplify-js/blob/master/packages/analytics/__tests__/Providers/AWSPinpointProvider-unit-test.ts#L737 to include an error without `message` set, to match your scenario.

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -447,7 +447,7 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
 		failureData: EndpointFailureData
 	) {
 		const { err, update_params, endpointObject } = failureData;
-		const { message } = err;
+		const { message = '' } = err;
 		const { ApplicationId, EndpointRequest } = update_params;
 
 		if (!message.startsWith('Exceeded maximum endpoint per user count')) {
```
ericclemmons: We're going to review this change and confirm the desired logic in the event an `err` comes back without a `message`.

One option is to not default `message` at all, and do a false-y check like so:

```suggestion
		if (!(message && message.startsWith('Exceeded maximum endpoint per user count'))) {
```

Going to mark this for review by the team & get this resolved.

Thanks for sharing your logs: I haven't seen an `err` without a `message` myself 🤞

ericclemmons: I'm going to update this branch today with a test to validate the changes, then we should be good.

Thanks again for bringing this to our attention @mariotsi.  We've identified this is a `400`, but don't know the full shape of the response.

If you ever _do_ see the shape of the error (e.g. `{ Debug: "Whoopsie!" }`), we can add that into the test.

ericclemmons:
@mariotsi I wasn't able to push directly to your branch, but I _did_ pull in your commits into a new PR that includes tests:

https://github.com/aws-amplify/amplify-js/pull/4507

I'd appreciate your 👀 on it: the only difference is that, since we can't guarantee the presence of `message` or that it's a string (vs. `null` or even a number), I had to be a bit more defensive.

PR #4501, opened 2019-12-03
Extracts the theme from the signUpConfig

Domino987:
This adds the missing theme from the signUpConfig to the Wrapper to be passed to the child components

_Description of changes:_

### DIFF
--- a/packages/aws-amplify-react-native/src/Auth/index.js
+++ b/packages/aws-amplify-react-native/src/Auth/index.js
@@ -63,7 +63,8 @@ export function withAuthenticator(
 			this.authConfig = {};
 
 			if (typeof includeGreetings === 'object' && includeGreetings !== null) {
-				this.authConfig = Object.assign(this.authConfig, includeGreetings);
+				this.authConfig = Object.assign(this.authConfig, includeGreetings);				
+				theme = includeGreetings.theme;
 			} else {
 				this.authConfig = {
 					includeGreetings,


ericclemmons:
@Domino987 This seems simple enough to land, but could you provide more detail on the behavior before/after this change? (Screenshots would be a huge help)

Domino987:
Of course, the theme is not applied at all, if added with the config object So without it, it just looks like the default:
photo5427132010100534360
and after the change, the theme is applied:
photo5429493327285300247

using the config object:
`
const config = {
  signUpConfig: {
    hideAllDefaults: true,
    signUpFields: [
      {
        custom: false,
        displayOrder: 1,
        key: 'username',
        label: 'Email',
        required: true,
        type: 'string',
      },
      {
        custom: false,
        displayOrder: 2,
        key: 'password',
        label: 'Password',
        required: true,
        type: 'password',
      },
    ],
  },
  theme: mergeDeep(AmplifyTheme, theme),`

packages/aws-amplify-react-native/src/Auth/index.js
```diff
@@ -63,7 +63,8 @@ export function withAuthenticator(
 			this.authConfig = {};
 
 			if (typeof includeGreetings === 'object' && includeGreetings !== null) {
-				this.authConfig = Object.assign(this.authConfig, includeGreetings);
+				this.authConfig = Object.assign(this.authConfig, includeGreetings);				
+				theme = includeGreetings.theme;
```
ericclemmons: Is there a reference for `includeGreetings` being an `object`? This seems to have been introduced in #2099 (via 6181247557f5343e5a8db75ba51ead333cd4268b), but I can't find the use-case for it:

The docs mention a boolean value here:
> https://aws-amplify.github.io/docs/js/authentication#using-withauthenticator-hoc

Such as:

```js
export default withAuthenticator(App, {includeGreetings: true, theme: MyTheme});
```

@haverchuck Can you weigh in on this?

ericclemmons:
@Domino987 Thanks for the screenshots! That helps a ton.  What I want to figure out is why `includeGreetings` is being converted to an `object` in our code, since you're config doesn't set it explicitly to `true`:

```js
const config = {
  signUpConfig: {
    hideAllDefaults: true,
    signUpFields: [
      {
        custom: false,
        displayOrder: 1,
        key: "username",
        label: "Email",
        required: true,
        type: "string"
      },
      {
        custom: false,
        displayOrder: 2,
        key: "password",
        label: "Password",
        required: true,
        type: "password"
      }
    ]
  },
  theme: mergeDeep(AmplifyTheme, theme)
};
```


It's great that it fixes it for you. I want to be able to understand it & possibly add a test for this behavior since I don't understand _why_ it works yet :)

Domino987:
As mentioned in the react [AWS docs](https://aws-amplify.github.io/docs/js/react#signup-configuration), the withAuthentificator HOC accepts either the parameters or an object which gets parsed to the parameters.
That's why it checks if isGreating is used as an object, if it is an object.

amhinson:
@Domino987 Would you be able to update the conflicts here?

eddiekeller:
This functionality is now present in our code. Thank you for the submission! Closing this PR.

https://github.com/aws-amplify/amplify-js/blob/main/packages/aws-amplify-react-native/src/Auth/index.tsx#L84

PR #4507, opened 2019-12-03
Handle Pinpoint error without "message"

ericclemmons:
_Issue #, if available:_ #4485

@mariotsi discovered in #4485:
> From our remote logs we saw errors coming from this line `TypeError: Cannot read property 'startsWith' of undefined`. We are unable to check more deeply what was the response because we do not not log payloads. I'm making this change in order to make the code more failsafe

This PR fixes a unit test that was incorrectly asserting that a rejection happened, _not what got rejected_, and adds a test case for the behavior @mariotsi is seeing.

The primary difference between aa8e3de and this PR is that `message` is cast as a `String` to defensively catch `undefined`, `null`, booleans, numbers, or any other valid, non-String JSON responses.  It also avoids defaulting `message` to `"", so that the original value of `undefined` is preserved for future changes.

---

### DIFF
--- a/packages/analytics/__tests__/Providers/AWSPinpointProvider-unit-test.ts
+++ b/packages/analytics/__tests__/Providers/AWSPinpointProvider-unit-test.ts
@@ -749,6 +749,28 @@ describe('AnalyticsProvider test', () => {
 				expect(reject).toBeCalledWith(mockError);
 				spyon.mockRestore();
 			});
+
+			test('BAD_REQUEST_CODE without message rejects error', async () => {
+				const analytics = new AnalyticsProvider();
+				const mockError = { debug: 'error', statusCode: 400 };
+
+				analytics.configure(options);
+				const spyon = jest
+					.spyOn(Pinpoint.prototype, 'updateEndpoint')
+					.mockImplementationOnce(params => ({
+						promise: jest.fn().mockRejectedValue(mockError),
+					}));
+
+				jest.spyOn(Credentials, 'get').mockImplementationOnce(() => {
+					return Promise.resolve(credentials);
+				});
+
+				const params = { event: { name: '_update_endpoint', immediate: true } };
+
+				await analytics.record(params, { resolve, reject });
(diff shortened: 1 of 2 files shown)


packages/analytics/__tests__/Providers/AWSPinpointProvider-unit-test.ts
```diff
@@ -730,12 +730,14 @@ describe('AnalyticsProvider test', () => {
 
 			test('error case', async () => {
 				const analytics = new AnalyticsProvider();
+				const mockError = { message: 'error' };
+
 				analytics.configure(options);
 				const spyon = jest
 					.spyOn(Pinpoint.prototype, 'updateEndpoint')
-					.mockImplementationOnce((params, callback) => {
-						callback({ message: 'error' }, null);
```
ericclemmons: This test was rejecting with `callback is not a function` and still passing, since it wasn't mocking this behavior correctly:
https://github.com/aws-amplify/amplify-js/blob/master/packages/analytics/src/Providers/AWSPinpointProvider.ts#L405-L407

packages/analytics/__tests__/Providers/AWSPinpointProvider-unit-test.ts
```diff
@@ -744,7 +746,29 @@ describe('AnalyticsProvider test', () => {
...
 
 				await analytics.record(params, { resolve, reject });
-				expect(reject).toBeCalled();
+				expect(reject).toBeCalledWith(mockError);
+				spyon.mockRestore();
+			});
+
+			test('BAD_REQUEST_CODE without message rejects error', async () => {
+				const analytics = new AnalyticsProvider();
+				const mockError = { debug: 'error', statusCode: 400 };
```
ericclemmons: Prior to the fix in a1b13da7c0fd6ff217fd66ebf00bfb712d8cae25, this fails.

mariotsi:
LGTM. I agree that preserving  `undefined` is more future proof. 
Thanks for you time investigating this.

Issue #4506, opened 2019-12-03
Does Amplify require zen-observable?


TriangularCube:
** Which Category is your question related to? **
Auth

** What AWS Services are you utilizing? **
Just Auth

After updating from 1.5 to ~~2.2~~ 2.1.1, it seems Auth is now missing a dependency of zen-observable that must be manually installed. This doesn't seem to be documented anywhere. Is the documentation simply lagging behind or am I missing something?

ericclemmons:
Hi @TriangularCube,

Can you share your `package.json` and the exact error you're getting?  

I was testing this out in a sandbox (https://codesandbox.io/s/httpsgithubcomaws-amplifyamplify-jsissues4506-sr9l1), and `zen-observable` appears to come in from a sub-dependency of `aws-amplify`:

```console
yarn why zen-observable
yarn why v1.16.0
warning package.json: No license field
[1/4] Why do we have the module "zen-observable"...?
[2/4] Initialising dependency graph...
warning https-githubcom-aws-amplify-amplify-js-issues-4506@1.0.0: No license field
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "zen-observable@0.8.15"
info Reasons this module exists
   - "aws-amplify#@aws-amplify#pubsub" depends on it
   - Hoisted from "aws-amplify#@aws-amplify#pubsub#zen-observable"
   - Hoisted from "aws-amplify#@aws-amplify#api#zen-observable"
```

TriangularCube:
It is a sub-dependency of aws-amplify, but this is not documented anywhere, and is not automatically installed when I use `npm install`.

I can't duplicate this in codesandbox, and I have no idea how it's installing dependencies, so here's a simple package.json I threw together.

`{
  "name": "amplify-test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "start": "parcel index.js --open"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@aws-amplify/auth": "^2.1.1"
  },
  "devDependencies": {
    "parcel-bundler": "^1.12.4"
  }
}`

Simply run `npm i`, then add `import Auth from "@aws-amplify/auth";` to the top of index.js, and parel should fail to compile.

I notice you're using Yarn though, so maybe it's an NPM issue?

ericclemmons:
[...]

PR #4494, opened 2019-12-02
Fix #4342: move @types/zen-observable to devDependencies

ericclemmons:
_Issue #, if available:_ #4342

Moves `@types/zen-observable` to `devDependencies` and updates to `^0.8.0`.

This is a regression of #3529 when #4007 was merged.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4511, opened 2019-12-04
fix(@aws-amplify/core): add zen-observable dependency

ericclemmons:
_Issue #, if available:_ #4506

`core` uses `zen-observable`:

https://github.com/aws-amplify/amplify-js/blob/8d2ba6e85031a7880d2b573e1f68108d22a7de54/packages/core/src/Util/Reachability.ts#L1

But it isn't declared as a dependency:
https://github.com/aws-amplify/amplify-js/blob/8d2ba6e85031a7880d2b573e1f68108d22a7de54/packages/core/package.json#L39-L46

Users who had multiple packages installed (such as API/PubSub) would get this as a  transitive dependency, which hid the error.

See this demo repo: https://github.com/ericclemmons/amplify-js-issues-4506

### DIFF
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -80,5 +80,8 @@
     "coveragePathIgnorePatterns": [
       "/node_modules/"
     ]
+  },
+  "devDependencies": {
+    "rimraf": "^3.0.0"
   }
 }
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -80,8 +80,5 @@
     "coveragePathIgnorePatterns": [
       "/node_modules/"
     ]
-  },
-  "devDependencies": {
-    "rimraf": "^3.0.0"
   }
 }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -43,7 +43,7 @@
   "dependencies": {
     "aws-sdk": "2.518.0",
     "url": "^0.11.0",
-    "zen-observable": "^0.8.15"
+    "zen-observable": "^0.8.6"
   },
   "jest": {
     "globals": {
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -42,7 +42,8 @@
   },
   "dependencies": {
     "aws-sdk": "2.518.0",
-    "url": "^0.11.0"
+    "url": "^0.11.0",
+    "zen-observable": "^0.8.15"
   },
   "jest": {
     "globals": {


ericclemmons:
@Amplifiyer How would you test that `@aws-amplify/core`'s `zen-observable` dependency installs via `@aws-amplify/auth` in a project like https://github.com/ericclemmons/amplify-js-issues-4506?

`yarn link` is normally the way to go for local development, but there must be a command or two I'm missing to do this successfully.

Amplifiyer:
> @Amplifiyer How would you test that `@aws-amplify/core`'s `zen-observable` dependency installs via `@aws-amplify/auth` in a project like https://github.com/ericclemmons/amplify-js-issues-4506?
> 
> `yarn link` is normally the way to go for local development, but there must be a command or two I'm missing to do this successfully.

If I understand your question correctly, you can publish locally using verdaccio and then import the auth package (which should now bring the verdaccio published core package and would have zen-observable dependency). Let me know if I misunderstood your question.

Also shouldn't the rimraf dependency be part of the monorepo dependencies?

ericclemmons:
@Amplifiyer I see.  Verdaccio isn't mentioned for public contributors in https://github.com/aws-amplify/amplify-js/wiki/Local-Development or https://github.com/aws-amplify/amplify-js/blob/master/CONTRIBUTING.md, so I didn't know it was a sanctioned workflow.

`rimraf` is a root-level dependency, as well as in a few package's `devDependencies`: https://github.com/aws-amplify/amplify-js/search?utf8=%E2%9C%93&q=rimraf&type=

It was added here because `yarn workspace @aws-amplify/core build` didn't work initially (likely due to hoisted dependencies failing to install prior).

Let me know if you'd like me to revert a60ce75, and we can discuss monorepo build tooling another time 🙏

Amplifiyer:
> It was added here because yarn workspace @aws-amplify/core build didn't work initially (likely due to hoisted dependencies failing to install prior).
>
> Let me know if you'd like me to revert a60ce75, and we can discuss monorepo build tooling another time 🙏

Yeah, we moved all the common devdependencies to the root level in all the packages where we added ES6 builds. I believe the idea is to run yarn bootstrap from the root level first before building the individual packages. I think this part is missing from the local development wiki as well. Maybe we can create a story to update this wiki?

ericclemmons:
@Amplifiyer Updated & ready for 👀

ashika01:
Ship it 🚢

mtraynham:
Any word on this getting merged?  Seems rather important if the project is missing dependencies.

ericclemmons:
@mtraynham Sorry for the delay!  We merged #4494 and didn't bring this with it by mistake.

hisham-blackbuck:
Hi guys, any word on when this patch would get released?

PR #4519, opened 2019-12-05
chore(@aws-amplify/ui-components): Check for user before authenticator loads

jordanranz:
_Description of changes:_
- Check user before Authenticator loads to determine if they are signed in or not

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
@@ -1,7 +1,7 @@
 import { Component, State, Prop, h } from '@stencil/core';
 import { AuthState, CognitoUserInterface, FederatedConfig } from '../../common/types/auth-types';
 import { AuthStateTunnel } from '../../data/auth-state';
-import { NO_AUTH_MODULE_FOUND, SIGNING_IN_WITH_HOSTEDUI_KEY } from '../../common/constants';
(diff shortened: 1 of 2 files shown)


packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -20,8 +21,36 @@ export class AmplifyAuthenticator {
...
+
+  async checkUser() {
+    if (!Auth || typeof Auth.currentAuthenticatedUser !== 'function') {
+      throw new Error(NO_AUTH_MODULE_FOUND);
+    }
+
+    try {
+      const user = await Auth.currentAuthenticatedUser();
+      this.onAuthStateChange(AuthState.SignedIn, user);
+    } catch (error) {
```
sammartinez: Can we double check this? Maybe a boolean factor if the user is authenticated then we move to signedIn if not we show sign out?

ericclemmons: From what I remember when using Auth, it always throws when there's not an error:
> https://github.com/aws-amplify/amplify-js/blob/8d2ba6e85031a7880d2b573e1f68108d22a7de54/packages/auth/src/Auth.ts#L1220

I don't believe I've had it return `null`, but I could be wrong since I primarily tested social providers...

jordanranz: Spoke offline about this with @sammartinez to explain the flow here

Issue #4517, opened 2019-12-05
Analytics(session tracking): Add option not to trigger new session on visibilityChange


germain-receeve:
**Is your feature request related to a problem? Please describe.**
Yes, in fact we want to track only when the user entered our page and closed it. At the moment, `_session.start` and ``_session.close` event are also emitted when switching tab for example because there is a watcher on page visibility (hidden or not)

**Describe the solution you'd like**
An option for the `autotrack` feature that would allow us to disable this behaviour.
Could be for example `trackVisibilityChange` which would 

**Describe alternatives you've considered**
Tracking the session event ourselves. Sadly, it's not feasible because the `immediate` event firing is only available internally in the lib, therefore it's not possible to send an event during `beforeUnload` event when user leaves the page.

**Additional context**
Line responsible for tracking `visibilityChange` events
https://github.com/aws-amplify/amplify-js/blob/8d2ba6e85031a7880d2b573e1f68108d22a7de54/packages/analytics/src/trackers/SessionTracker.ts#L166

ericclemmons:
You're right, the `_session.stop` event is called when visibility is lost.

https://github.com/aws-amplify/amplify-js/blob/8d2ba6e85031a7880d2b573e1f68108d22a7de54/packages/analytics/src/trackers/SessionTracker.ts#L81-L87

Is the problem that you can't differentiate the visibility events vs. close events? (For example, `_session.pause` and `_session.resume` vs. `_session.start` and `_session.stop`)

Or is the problem that it's creating noise and you'd like to filter these out entirely?

germain-receeve:
I would say it's more that it's making noise since in our case we are mainly interested by the user starting to use the site and effectively quitting it.
If we could differentiate those events like you suggested, that would work for us as well 👍

ericclemmons:
Great feedback @germain-receeve!

I'd like to proceed with getting some internal feedback with differentiating those events vs. filtering, and can even put together a PR for you.

Here's my rationale:

- Adding more API options would require more consideration & a higher bar for testing & support.
- Differentiating these events, in my opinion, has a lot of value in data-slicing: you'll be able to understand why there was a lull in events vs. seeing nothing for a long period of time.

ericclemmons:
@germain-receeve Alright, I quickly put together https://github.com/aws-amplify/amplify-js/pull/4525 pretty easily, so we can have an internal discussion to ensure nothing blows up 🤞 

[...]

PR #4525, opened 2019-12-06
Track visibility via _session.pause & _session.resume

ericclemmons:
_Issue #, if available:_ #4517

@germain-receeve has been receiving noisy `_session.start` and `_session.stop` data, due to page visibility also sending those same events.

The initial suggestion was to introduce an option (e.g. `trackVisibilityChange`) to filter these events out entirely:

> An option for the autotrack feature that would allow us to disable this behaviour.
Could be for example trackVisibilityChange option which would not add the event listener if set to false.
> ...
> I would say it's more that it's making noise since in our case we are mainly interested by the user starting to use the site and effectively quitting it.
If we could differentiate those events like you suggested, that would work for us as well 👍

This PR opts for differentiating the `_session.start` / `_session.stop` with **new** `_session.pause` `_session.resume` visbility events, rather then removing data entirely.

These event names seem to be based on convention vs. a contract:
> https://docs.aws.amazon.com/mobileanalytics/latest/ug/managing-sessions.html

**However, if reporting depends on these events (e.g. for calculating DAU), then this would be a backwards incompatible change**. If that's the case, then adding an option for users to opt into is the best course of action.

---

### DIFF
--- a/packages/analytics/__tests__/trackers/SessionTracker-test.ts
+++ b/packages/analytics/__tests__/trackers/SessionTracker-test.ts
@@ -114,7 +114,7 @@ describe('SessionTracker test', () => {
 
 			expect(tracker).toBeCalledWith(
 				{
-					name: '_session.stop',
+					name: '_session.pause',
 					attributes: {},
 				},
 				'AWSPinpoint'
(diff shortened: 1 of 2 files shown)


ericclemmons:
I'm engaging the Pinpoint service team to determine the impact of this change 🤞

ericclemmons:
Unfortunately, this change cannot be accepted.  It would be a potential breaking change for customer reporting and would require a major release and multi-platform support.

PR #4532, opened 2019-12-09
@aws-amplify/auth: fix typos in OAuth field names

sprohaska:

Signed-off-by: Steffen Prohaska <prohaska@zib.de>

### DIFF
--- a/packages/auth/src/OAuth/oauthStorage.ts
+++ b/packages/auth/src/OAuth/oauthStorage.ts
@@ -22,16 +22,16 @@ export const getState = () => {
 };
 
 export const setPKCE = (private_key: string) => {
-	window.sessionStorage.setItem('ouath_pkce_key', private_key);
+	window.sessionStorage.setItem('oauth_pkce_key', private_key);
 };
 
 export const getPKCE = () => {
-	const ouath_pkce_key = window.sessionStorage.getItem('ouath_pkce_key');
-	window.sessionStorage.removeItem('ouath_pkce_key');
-	return ouath_pkce_key;
+	const oauth_pkce_key = window.sessionStorage.getItem('oauth_pkce_key');
+	window.sessionStorage.removeItem('oauth_pkce_key');
+	return oauth_pkce_key;
 };
 
 export const clearAll = () => {
-	window.sessionStorage.removeItem('ouath_pkce_key');
+	window.sessionStorage.removeItem('oauth_pkce_key');
 	window.sessionStorage.removeItem('oauth_state');
 };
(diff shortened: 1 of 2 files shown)


Amplifiyer:
@sprohaska, thanks for catching the typo and submitting the PR to fix it. Since the local storage key name is getting changed, will it not cause issues for customers when they update their app? I'm guessing that all such customers will get logged out?

sprohaska:
>  I'm guessing that all such customers will get logged out?

Active sessions probably get logged out. If this is a concern, I'd suggest to either schedule the fix for a major revision or keep the typo and add a comment in the source code that it must not be fixed.

packages/auth/src/OAuth/oauthStorage.ts
```diff
@@ -22,16 +22,16 @@ export const getState = () => {
...
 export const setPKCE = (private_key: string) => {
-	window.sessionStorage.setItem('ouath_pkce_key', private_key);
+	window.sessionStorage.setItem('oauth_pkce_key', private_key);
 };
 
 export const getPKCE = () => {
-	const ouath_pkce_key = window.sessionStorage.getItem('ouath_pkce_key');
-	window.sessionStorage.removeItem('ouath_pkce_key');
-	return ouath_pkce_key;
+	const oauth_pkce_key = window.sessionStorage.getItem('oauth_pkce_key');
```
ericclemmons: One approach is maintaining backwards-compatibility for reading from the incorrect key as a fall-back, while future reads/writes use the correct key:

```suggestion
	const oauth_pkce_key = window.sessionStorage.getItem('oauth_pkce_key') || window.sessionStorage.getItem('ouath_pkce_key');
	window.sessionStorage.removeItem('oauth_pkce_key');
	window.sessionStorage.removeItem('ouath_pkce_key');
```

Thoughts?

Amplifiyer: @ericclemmons, that's awesome and clever 😉

sprohaska:
I've looked into backward compatibility in more detail. I now believe that customers won't get signed out. `setPKCE()` is used in `OAuth/OAuth.ts:oauthSignIn()` to set a code verifier. The code is only temporarily relevant during the ongoing OAuth flow until `_handleCodeFlow()` calls `getPKCE()`.

packages/auth/src/OAuth/oauthStorage.ts
ericclemmons: @Amplifiyer We may not need this based on @sprohaska's comment:
> https://github.com/aws-amplify/amplify-js/pull/4532#issuecomment-571058126

PR #4546, opened 2019-12-10
fix(@aws-amplify/auth): Adjust Cognito error message checking for dis…

### LINKED ISSUE #4544: Missing "." in /auth/src/Auth.ts
**Describe the bug**
The currentUserPoolUser method does not throw because its checking the err string incorrectly!

```err.message === 'User is disabled'```

should be

```err.message === 'User is disabled.'``` (notice the fullstop/period)

This means that  `Auth.currentAuthenticatedUser` never throws is a user has been disabled from the console.

manueliglesias:
…abled user.

Fixes #4544

_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/packages/auth/__tests__/auth-unit-test.ts
+++ b/packages/auth/__tests__/auth-unit-test.ts
@@ -500,12 +500,11 @@ describe('auth unit test', () => {
 
 			await auth.confirmSignUp('username', code);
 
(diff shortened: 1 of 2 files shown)


sammartinez:
@manueliglesias Seems to be a broken unit test, can you please take a look when you get a chance?

PR #4548, opened 2019-12-10
feat: (@aws-amplify/ui-components) TOTP / Select MFA Type Components

sammartinez:
_Issue #, if available:_
[Delivers: #169738491]

_Description of changes:_
In order to reach feature parity, this work is incorporating the TOTP component and the Select MFA Type component (UC - Utility Component). On the UC, there is an action item to combine the TOTP component with the Select MFA Type. This will need to be done after* the toast component has been completed.

* Unit tests
* Storybook integrations
* Components from React to WC Stencil
* TOTP Example added

Steps to test TOTP: 
1. Use internal `aws-exports` file, ping me if need be
3. Go through the `Amplify Authenticator` under this Pull Request and follow steps provided offline.
3a. QR Code will appear to scan. You can use `Google Authenticator`, link for [iOS](https://apps.apple.com/us/app/google-authenticator/id388497605) or [Android](https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2&hl=en_US), scan the QR Code and follow the prompts on `Google Authenticator` to receive the number needed to finish the TOTP Authentication process. 
3. Once you have successfully configured TOTP, sign out of the application and re-sign in. After the initial log in, you will see a component with `Confirm TOTP code`. Using the `Google Authenticator` application, find the user and type in the code generated.


ericclemmons:
What's the best way I can validate this behavior (step-by-step is ideal)?

The story doesn't have enough states to validate correctness:
> Screen Shot 2019-12-10 at 2 03 21 PM

And running `amplify pull` doesn't yield projects that I recognize. Is there an Isengard account/team profile I should be using?

> Screen Shot 2019-12-10 at 2 43 54 PM

I can review the code, but that'd be backwards without being able to review the functionality behavior.

ericclemmons:
As I went through the flow & confirmed my SMS code, the next screen was:
> Screen Shot 2019-12-10 at 4 13 16 PM

Upon refresh, I correctly got the barcode & scanned with my authenticator client.

However, when pressing "Verify security token", it doesn't visibly do anything:
> Screen Shot 2019-12-10 at 4 15 31 PM

Clicking again only returns the error:
> [DEBUG] 16:16.996 TOTP - No TOTP Code provided

Is this correct?

sammartinez:
@ericclemmons 
> As I went through the flow & confirmed my SMS code, the next screen was:
> 
> > Screen Shot 2019-12-10 at 4 13 16 PM
> 
> Upon refresh, I correctly got the barcode & scanned with my authenticator client.
> 
> However, when pressing "Verify security token", it doesn't visibly do anything:
> 
> > Screen Shot 2019-12-10 at 4 15 31 PM
> 
> Clicking again only returns the error:
> 
> > [DEBUG] 16:16.996 TOTP - No TOTP Code provided
> 
> Is this correct?

Currently yes this is correct since there is no Toast component in place yet to show the success. We will be implementing that shortly

ericclemmons:
Thanks for the response.  I'll do some code review to wrap this up.

The part that seemed wrong was that the state didn't transition successfully to showing a QR code, but upon refresh it did.  If the toast integration fixes that, then 👍

packages/amplify-ui-components/src/components/amplify-radio-button/amplify-radio-button.tsx
```diff
@@ -10,12 +10,16 @@ const STATIC_RADIO_BUTTON_CLASS_NAME = `${AMPLIFY_UI_PREFIX}--radio-button`;
...
+  /** The callback, called when the input is modified by the user. */
+  @Prop() handleInputChange?: (inputEvent: Event) => void;
   /** (Optional) Overrides default styling */
   @Prop() overrideStyle: boolean = false;
   /** (Optional) Name of radio button */
   @Prop() name?: string;
   /** (Optional) Value of radio button */
   @Prop() value?: string;
+  /** (Optional) The placeholder for the input element.  Using hints is recommended, but placeholders can also be useful to convey information to users. */
+  @Prop() placeholder?: string = '';
```
ericclemmons: Is there different behavior between `''` as a default vs `undefined`? (As in, how does it affect users?)

For example, with `undefined`, the snapshot won't have the `placholder` attribute at all, while they do now `placeholder=""`.

(Looking to understand if this is preference vs. intentional for users)

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+    it('should render authData to null by default', () => {
+      expect(selectMFAType.authData).toEqual(null);
+    });
+
+    it('should render `MFATypes` to undefined by default', () => {
+      expect(selectMFAType.MFATypes).toBeUndefined();
+    });
+  });
+  describe('Render logic ->', () => {
+    it('should render a `less than 2 mfa types available` message by default', async () => {
```
ericclemmons: I don't understand what this message means, or when it happens in the wild.

Is it an error state due to configuration, or a user is in some impossible state?

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.stories.tsx
```diff
@@ -0,0 +1,7 @@
+import { storiesOf } from '@storybook/html';
+
+const selectMfaTypeStories = storiesOf('amplify-select-mfa-type', module);
+
+selectMfaTypeStories.add('default', () => {
+  return `<amplify-select-mfa-type />`;
+});
```
ericclemmons: For new stories, we should consider using this format:
> https://storybook.js.org/docs/formats/component-story-format/

The immediate benefit of this is that I was able to snapshot all of my stories without much effort:
https://github.com/aws-amplify/amplify-js/blob/47aeaa7a2fafb5a81cd05cf8f48093a0376d14d6/packages/amplify-ui-components/src/components/amplify-federated-buttons/amplify-federated-buttons.spec.tsx#L4-L19

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
```diff
@@ -0,0 +1,162 @@
...
+  /** Fires when Verify is clicked */
+  @Prop() handleSubmit: (submitEvent: Event) => void = () => this.verify();
+
+  @State() TOTPSetup: boolean = false;
+  @State() selectMessage: string = null;
+  @State() mfaMethod: any = null;
+
+  @State() isTotp: boolean = false;
+  @State() isNoMfa: boolean = false;
+  @State() isSMS: boolean = false;
```
ericclemmons: The casing is different between `MFAType`, `TOTOPSetup`, `mfaMethod`, `isTotp`, `isNoMfa`, `isSMS`.

Sure, if you're using something like `_.camelCase` to generate it, we'll get `isTotp`.  

However, if you'd prefer to keep the initialisms in tact (e.g. `isTOTP`, `isNoMFA`, `MFAMethod`, to match `isSMS`, etc.), I'm on board. 💪

At the end of the day, we have typings to help us a bit, but these are internal `state` variables, so they don't warrant as much scrutiny as the `@Prop`s above.

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
```diff
@@ -0,0 +1,162 @@
...
+    this.TOTPSetup = false;
+    this.selectMessage = null;
+
+    // Reseting state values to default
+    this.isNoMfa = false;
+    this.isTotp = false;
+    this.isSMS = false;
+
+    const { value, type, checked } = event.target;
+    const checkType = ['radio', 'checkbox'].includes(type);
```
ericclemmons: The `handleRadioButtonChange` event is bound specifically to `<amplify-radio-button>`, so it appears that `checkType` will always be `true`.

What else am I missing?

packages/amplify-ui-components/src/components/amplify-totp/amplify-totp.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+@Component({
+  tag: 'amplify-totp',
+  shadow: false,
+})
+export class AmplifyTOTP {
+  /** Used in order to configure TOTP for a user */
+  @Prop() authData: CognitoUserInterface = null;
+  /** Triggers an TOTP Event after handleSubmit method has been called */
+  @Prop() onTOTPEvent?: (event: TOTPSetupEventType, data: any, user: CognitoUserInterface) => void;
```
ericclemmons: What do you think about typing this, similar to:
https://github.com/aws-amplify/amplify-js/blob/47aeaa7a2fafb5a81cd05cf8f48093a0376d14d6/packages/amplify-ui-components/src/common/types/auth-types.ts#L68

packages/amplify-ui-components/stencil.config.ts
```diff
@@ -6,6 +6,7 @@ export const config: Config = {
   plugins: [nodePolyfills()],
   commonjs: {
     namedExports: {
+      '@aws-sdk/client-cognito-identity-browser': ['CognitoIdentityClient', 'GetIdCommand'],
```
ericclemmons: What error do you get that is fixed with this change?

jordanranz: @sammartinez This change is in ui-components/master, so you can rebase and it will get rid of this.

@ericclemmons This is the error:
```
[ ERROR ]  Rollup: Missing Export: ../core/lib-esm/Credentials.js:54:9
           'CognitoIdentityClient' is not exported by
           ../../node_modules/@aws-sdk/client-cognito-identity-browser/index.js

     L53:  import { fromCognitoIdentity, fromCognitoIdentityPool, } from '@aws-sdk/credential-provider-cognito-identity';
     L54:  import { CognitoIdentityClient, GetIdCommand, } from '@aws-sdk/client-cognito-identity-browser';
     L55:  var logger = new Logger('Credentials');

[29:26.0]  build failed in 23.41 s
```

packages/amplify-ui-components/src/components/amplify-examples/auth-totp-example.tsx
```diff
@@ -0,0 +1,22 @@
...
+let authData;
+
+Auth.currentAuthenticatedUser().then(data => {
+  authData = data;
+  console.log('Auth data set', authData);
+});
+
+const TOTPAuthenticator = () => (
+  <amplify-authenticator>
+    <amplify-totp authData={authData}></amplify-totp>
```
jordanranz: `amplify-totp` should be built into the `amplify-authenticator`

sammartinez: As per our offline discussion, working on this now

packages/amplify-ui-components/src/common/types/auth-types.ts
```diff
@@ -73,6 +73,12 @@ export enum MfaOption {
   NOMFA = 'NOMFA',
 }
 
+export interface MFATOTPOptions {
```
jordanranz: These may be better described as `MFATypes` instead of `MFATOTPOptions`

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -84,6 +84,8 @@ export class AmplifyAuthenticator {
         return <amplify-forgot-password handleAuthStateChange={this.onAuthStateChange} />;
       case AuthState.ResetPassword:
         return <amplify-require-new-password handleAuthStateChange={this.onAuthStateChange} user={this.authData} />;
+      case AuthState.TOTPSetup:
+        return <amplify-totp handleAuthStateChange={this.onAuthStateChange} user={this.authData} />;
```
jordanranz: I think this changed to `amplify-totp-setup`

sammartinez: Updating.. Seems like my rebase had some issues. will circle back

packages/amplify-ui-components/src/common/types/auth-types.ts
sammartinez: Let me double check, Im not sure it likes `MFATypes` as the type for `@Prop(): MFATypes` in `select-mfa-type` hence the change but let me see if it throws an error. May just end up doing an interface tag at the end.

packages/amplify-ui-components/stencil.config.ts
ericclemmons: Thanks!  I haven't gotten these rollup errors yet, so this helps me recognize it for fixing it myself when the time comes. 💪

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+@Component({
+  tag: 'amplify-select-mfa-type',
+  shadow: false,
+})
+export class AmplifySelectMFAType {
+  /** Types of MFA options */
+  @Prop() MFATypes: MFATypesInterface;
+  /** Current authenticated user in order to sign requests properly for TOTP */
+  @Prop() authData: CognitoUserInterface = null;
```
jordanranz: Can this be `undefined` instead of defaulting to `null`?

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
sammartinez: The OG code had this in place, I wanted to make sure that if in the future someone were to change the behavior this would be caught because we do need the `checked` value in order to set it to true and/or false

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
```diff
@@ -0,0 +1,167 @@
...
+  /** Fires when Verify is clicked */
+  @Prop() handleSubmit: (submitEvent: Event) => void = () => this.verify();
```
jordanranz: You should pass an `event` variable into the `verify` default function call

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    if (value === MfaOption.NOMFA && checkType) {
+      this.isNoMfa = checked;
+    }
+  }
+
+  verify() {
+    // avoid submitting the form
+    if (event) {
```
jordanranz: `event` is not passed into the function.

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
sammartinez: Fair callout and as you have pointed out these are only present to the state logic itself. If there are other things to adjust I can make this adjustment.

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup-interface.ts
```diff
@@ -0,0 +1,7 @@
+export enum MfaMethod {
```
jordanranz: There is an `enum` for this in the common `auth-types` file

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup-interface.ts
```diff
@@ -0,0 +1,7 @@
...
+  TOTP = 'TOTP',
+  SMS = 'SMS',
+  NOMFA = 'NOMFA',
+}
+
+export type TOTPSetupEventType = 'SETUP_TOTP';
```
jordanranz: Is this needed? I am confused as to how Typescript helps here. shouldn't this just be a `const`?

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.stories.tsx
```diff
@@ -0,0 +1,7 @@
+import { storiesOf } from '@storybook/html';
+
+const totpStories = storiesOf('amplify-totp', module);
+
+totpStories.add('default', () => {
+  return `<amplify-totp></amplify-totp>`;
```
jordanranz: `totp-setup`

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+@Component({
+  tag: 'amplify-totp-setup',
+  shadow: false,
+})
+export class AmplifyTOTPSetup {
+  /** Used in order to configure TOTP for a user */
+  @Prop() user: CognitoUserInterface = null;
+  /** Used to set autoFocus to true when TOTP Component has loaded */
+  @Prop() inputProps: object = {
```
jordanranz: I don't think we should have `inputProps` as a prop of this component. Can we just set this on the input element?

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -0,0 +1,164 @@
...
+  setup() {
+    this.setupMessage = null;
+    const user = this.user;
+    const issuer = encodeURI('AWSCognito');
+
+    if (!Auth || typeof Auth.setupTOTP !== 'function') {
+      throw new Error(NO_AUTH_MODULE_FOUND);
+    }
+
+    Auth.setupTOTP(user)
```
jordanranz: Can we use `async/await` here?

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -0,0 +1,164 @@
...
+  /** Passed from the Authenticator component in order to change Authentication state */
+  @Prop() handleAuthStateChange: AuthStateHandler;
+
+  @State() code: string | null = null;
+  @State() setupMessage: string | null = null;
+  @State() qrCodeImageSource: string;
+  @State() qrCodeInput: string | null = null;
+
+  componentDidLoad() {
+    this.setup();
```
jordanranz: Can `setup` be called on `componentWillLoad` instead?

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    const user = this.user;
+
+    if (!Auth || typeof Auth.verifyTotpToken !== 'function' || typeof Auth.setPreferredMFA !== 'function') {
+      throw new Error(NO_AUTH_MODULE_FOUND);
+    }
+
+    Auth.verifyTotpToken(user, this.qrCodeInput)
+      .then(() => {
```
jordanranz: Can we use `async/await` here?

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -0,0 +1,164 @@
...
+    if (event === SETUP_TOTP && data === SUCCESS) {
+      this.checkContact(user);
+    }
+  }
+
+  triggerTOTPEvent(event: TOTPSetupEventType, data: any, user: CognitoUserInterface) {
+    if (this.onTOTPEvent) {
+      this.onTOTPEvent(event, data, user);
+    }
+  }
```
jordanranz: Can we merge these functions?

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+        this.generateQRCode(this.code);
+      })
+      .catch(error => {
+        logger.debug(TOTP_SETUP_FAILURE, error);
+      });
+  }
+
+  verifyTotpToken() {
+    if (event) {
```
jordanranz: `event` is not being passed into this function.

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.stories.tsx
sammartinez: im never rebasing againnn....

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
sammartinez: adjusted

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -0,0 +1,164 @@
...
+  componentDidLoad() {
+    this.setup();
+  }
+
+  checkContact(user) {
+    if (!Auth || typeof Auth.verifiedContact !== 'function') {
+      throw new Error(NO_AUTH_MODULE_FOUND);
+    }
+    Auth.verifiedContact(user)
+      .then(data => {
```
jordanranz: Can we use `async/await` here?

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
sammartinez: I can double check to see, I made a similar callout on your PR, I was under the impression it may not be safe so

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
sammartinez: From the onSubmit, there is still an event that is triggered from this, I can see about adjusting this.

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const issuer = encodeURI('AWSCognito');
+
+    if (!Auth || typeof Auth.setupTOTP !== 'function') {
+      throw new Error(NO_AUTH_MODULE_FOUND);
+    }
+
+    Auth.setupTOTP(user)
+      .then(secretKey => {
+        logger.debug('secret key', secretKey);
+        const code = `otpauth://totp/${issuer}:${user.username}?secret=${secretKey}&issuer=${issuer}`;
```
jordanranz: Can this just be:
this.code = `otpauth://totp/${issuer}:${user.username}?secret=${secretKey}&issuer=${issuer}`;

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
sammartinez: I feel for the `async/await` callouts are out of scope for this PR as these were the implementations from the React components. If I circle back I feel like this is gonna bleed into the Toast work that needs to be done

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
sammartinez: What do you mean? The string literal is there in order to have `${value}` work. On the second point, it may be over kill unless it would be used else where. WDYT?

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
sammartinez: sure

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
sammartinez: There is still an event that comes back from the onChange call

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup-interface.ts
sammartinez: Ill adjust

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup-interface.ts
sammartinez: Its mainly hardcoding the event to be this. Similar to `export type inputType = 'string' | 'number' | 'tel'`

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -0,0 +1,164 @@
...
+    try {
+      this.qrCodeImageSource = await QRCode.toDataURL(codeFromTotp);
+    } catch (error) {
+      throw new Error(error);
+    }
+  }
+
+  setup() {
+    this.setupMessage = null;
+    const user = this.user;
```
jordanranz: Can we just use `this.user` instead of reassigning it?

ericclemmons:
The flow seems to be working better with the provided `aws-exports.js`, but the token continues to fail:
> Screen Shot 2019-12-16 at 4 59 14 PM

ericclemmons:
Used a new user and it works! 👍

ericclemmons:
Used a new user and it works! 👍

sammartinez:
> Used a new user and it works! 👍

Cool! yup the test for this to work would have to be that a new user is created.

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
sammartinez: sure

ericclemmons:
The mistake I made was initially using `eric_totp`, as I thought it was deleted 🤦‍♂️

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.spec.ts
sammartinez: From the OG code, this is in place in order to have customers pass to the utility component at least two values being passed to `MFATypes` prop. This is a cutover from the React Component

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
sammartinez: done

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+    it('should render authData to undefined by default', () => {
+      expect(selectMFAType.authData).toBeUndefined();
+    });
+
+    it('should render `MFATypes` to undefined by default', () => {
+      expect(selectMFAType.MFATypes).toBeUndefined();
+    });
+  });
+  describe('Render logic ->', () => {
+    it('should render a `less than 2 mfa types available` message by default', async () => {
```
jordanranz: Can we include a snapshot test for when 2+ MFA types are passed in.

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
```diff
@@ -0,0 +1,167 @@
...
+  /** Types of MFA options */
+  @Prop() MFATypes: MFATypesInterface;
+  /** Current authenticated user in order to sign requests properly for TOTP */
+  @Prop() authData: CognitoUserInterface;
+  /** Fires when Verify is clicked */
+  @Prop() handleSubmit: (submitEvent: Event) => void = submitEvent => this.verify(submitEvent);
+
+  @State() TOTPSetup: boolean = false;
+  @State() selectMessage: string = null;
+  @State() MFAMethod: any = null;
```
jordanranz: Can we change this from `any` to a more specific type?

sammartinez: sure

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.spec.ts
sammartinez: sure

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup-interface.ts
jordanranz: ahh ok

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+@Component({
+  tag: 'amplify-totp-setup',
+  shadow: false,
+})
+export class AmplifyTOTPSetup {
+  /** Used in order to configure TOTP for a user */
+  @Prop() user: CognitoUserInterface;
+
+  inputProps: object = {
```
jordanranz: NIT: We should move this above Props and State

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+    try {
+      const secretKey = await Auth.setupTOTP(this.user);
+
+      logger.debug('secret key', secretKey);
+      this.code = this.buildOtpAuthPath(this.user, issuer, secretKey);
+
+      this.generateQRCode(this.code);
+    } catch (error) {
+      logger.debug(TOTP_SETUP_FAILURE, error);
```
jordanranz: Should we `throw` here?

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
sammartinez: done

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
sammartinez: done

PR #4559, opened 2019-12-11
feat(@aws-amplify/ui-components): amplify-auth0-button

ericclemmons:
[Delivers #169928296]

A placeholder was implemented in #4249, but wasn't functional.  This PR adds that functionality:

> auth0

(View commits for overview)

1. Followed instructions at:

	> https://aws-amplify.github.io/docs/js/authentication#federated-with-auth0
	and
	> https://auth0.com/docs/integrations/integrating-auth0-amazon-cognito-mobile-apps

1. Setup Identity Provider + Identity Pool to match existing `aws-exports.js` file
1. Updated `amplify-examples/authenticator-example.tsx` to look like:

```diff
diff --git a/packages/amplify-ui-components/src/components/amplify-examples/authenticator-example.tsx b/packages/amplify-ui-components/src/components/amplify-examples/authenticator-example.tsx
index 8858123e..84a7a540 100644
--- a/packages/amplify-ui-components/src/components/amplify-examples/authenticator-example.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-examples/authenticator-example.tsx
@@ -1,7 +1,19 @@
 import { h } from '@stencil/core';

 const Authenticator = () => (
-  <amplify-authenticator>
+  <amplify-authenticator
+    federated={{
+      auth0Config: {
+        domain: 'aws-amplify-test.auth0.com',
+        clientID: '•••••••••••••••••••••••••',
+        redirectUri: window.location.origin,
+        responseType: 'token id_token',
+        scope: 'openid profile email',
+      },
+    }}
+  >
     <h1>My App</h1>
   </amplify-authenticator>
 );
```

---

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
@@ -35,13 +35,6 @@ export class AmplifyAuth0Button {
 
     if (!this._auth0) {
       this._auth0 = new window['auth0'].WebAuth(config);
-      // TODO Do we need this?
-      window['auth0_client'] = this._auth0;
-    }
-
-    if (!this._auth0) {
-      logger.error('Auth0 instance was not initialized');
-      return;
     }
 
     this._auth0.parseHash((err, authResult) => {
--- a/packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
@@ -1,7 +1,11 @@
+import { Auth } from '@aws-amplify/auth';
+import { ConsoleLogger as Logger } from '@aws-amplify/core';
 import { Component, h, Prop } from '@stencil/core';
 
-import { SIGN_IN_WITH_AUTH0 } from '../../common/constants';
-import { FederatedConfig } from '../../common/types/auth-types';
+import { AUTH_SOURCE_KEY, SIGN_IN_WITH_AUTH0 } from '../../common/constants';
+import { AuthStateHandler, FederatedConfig, AuthState } from '../../common/types/auth-types';
+
+const logger = new Logger('amplify-auth0-button');
 
 @Component({
   tag: 'amplify-auth0-button',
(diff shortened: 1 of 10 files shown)


packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
```diff
@@ -12,11 +16,94 @@ export class AmplifyAuth0Button {
...
+    if (!this._auth0) {
+      this._auth0 = new window['auth0'].WebAuth(config);
+      // TODO Do we need this?
+      window['auth0_client'] = this._auth0;
+    }
+
+    if (!this._auth0) {
+      logger.error('Auth0 instance was not initialized');
+      return;
+    }
```
jordanranz: Does there need to be 2 separate blocks for `if (!this._auth0)`?

ericclemmons: You're right.  It was a bit of a mess in there, so I'll clean this up 😅

packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
```diff
@@ -12,11 +16,94 @@ export class AmplifyAuth0Button {
   @Prop() config: FederatedConfig['auth0Config'];
   /** (Optional) Override default styling */
   @Prop() overrideStyle: boolean = false;
+  @Prop() handleAuthStateChange: AuthStateHandler;
+
+  _auth0: any;
```
sammartinez: What is this being used for? Is it more of a `@State()` or `@Prop()`?

packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
sammartinez: > You're right. It was a bit of a mess in there, so I'll clean this up 😅

Please and thank you :D

packages/amplify-ui-components/src/components/amplify-auth0-button/amplify-auth0-button.tsx
ericclemmons: It's not renderable state (since it doesn't change) and not a prop (since it doesn't come from the the consumer).

It's an instance variable who's reference is shared between Auth0 load & button click.  (Reference: https://stenciljs.com/docs/state#when-to-use-)

I marked it as `any` since it's a 3rd party that we rely on at run-time.

PR #4560, opened 2019-12-12
CredentialsProvider was returning an object instead of a Promise. Fix…

jordanranz:
_Description of changes:_
- CredentialProvider returned an object instead of a Promise. loadCredentials expects a Promise.
- We were missing this part of the code from `fromCognitoIdentityPool `:
https://github.com/aws/aws-sdk-js-v3/blob/04592b775d715c1f0313954a3f09a01ba5d229bd/packages/credential-provider-cognito-identity/src/fromCognitoIdentityPool.ts#L59-L66

### DIFF
--- a/packages/core/src/Credentials.ts
+++ b/packages/core/src/Credentials.ts
@@ -316,7 +316,7 @@ export class CredentialsClass {
 
 			Note: Retreive identityId from CredentialsProvider once aws-sdk-js v3 supports this.
 		*/
-		let credentialsProvider: CredentialProvider = async () => {
+		const credentialsProvider: CredentialProvider = async () => {
 			const { IdentityId } = await cognitoClient.send(
 				new GetIdCommand({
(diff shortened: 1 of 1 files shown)


packages/core/src/Credentials.ts
```diff
@@ -331,11 +331,15 @@ export class CredentialsClass {
 				identityId: IdentityId,
 			};
 
-			credentials = fromCognitoIdentity(cognitoIdentityParams);
+			credentialsProvider = fromCognitoIdentity(cognitoIdentityParams);
```
Amplifiyer: Can we name this variable something other than `credentialsProvider` since the outer variable is named the same and there is  confusion about the difference between these two.

jordanranz: Yea, it seems like the code I am mimicking in aws-sdk v3 reused the function variable like this. I think it is an improvement to make the change you are suggesting.

PR #4564, opened 2019-12-13
feat: append amplify user agent to all V3 SDK calls

Amplifiyer:
_Issue #, if available:_

_Description of changes:_ append amplify user agent to all V3 SDK calls

### DIFF
--- a/packages/analytics/src/Providers/AWSPinpointProvider.ts
+++ b/packages/analytics/src/Providers/AWSPinpointProvider.ts
@@ -521,27 +521,7 @@ export class AWSPinpointProvider implements AnalyticsProvider {
 		const { region } = config;
 		logger.debug('init clients with credentials', credentials);
 		this.pinpointClient = new PinpointClient({ region, credentials });
-
-		if (Platform.isReactNative) {
-			const appendUserAgentMiddleware = next => args => {
-				const { request } = args;
-				if (!request.headers['User-Agent']) {
-					request.headers['User-Agent'] = Platform.userAgent;
-				} else {
-					request.headers['User-Agent'] += Platform.userAgent;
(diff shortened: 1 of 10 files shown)


packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -521,27 +521,7 @@ export class AWSPinpointProvider implements AnalyticsProvider {
 		const { region } = config;
 		logger.debug('init clients with credentials', credentials);
 		this.pinpointClient = new PinpointClient({ region, credentials });
-
```
sammartinez: Whats the reason for this being removed?

Amplifiyer: Because it is replaced with the utility function appendAmplifyUserAgent() from core.

packages/predictions/src/Providers/AmazonAIIdentifyPredictionsProvider.ts
```diff
@@ -129,7 +133,9 @@ export class AmazonAIIdentifyPredictionsProvider extends AbstractIdentifyPredict
 			} = {},
 		} = this._config;
 		this.rekognitionClient = new RekognitionClient({ region, credentials });
+		appendAmplifyUserAgent(this.rekognitionClient);
 		this.textractClient = new TextractClient({ region, credentials });
+		appendAmplifyUserAgent(this.textractClient);
```
jordanranz: Is there a reason we append when the client is instantiated instead of when a `send` command is called?

jordanranz:
LGTM overall. 1 question about why we append after client instantiation instead of before `send` commands

PR #4568, opened 2019-12-13
fix: Use paulfantom/periodic-labeler for labeling PRs

ericclemmons:
[Delivers #170288921]

The labeler runs successfully on PRs from `aws-amplify`, but **not on forks** with this error:

> 70322537-a38f6c00-182a-11ea-9d7e-6fe8b7d54933
> https://github.com/actions/labeler/issues/16#issuecomment-562550042

This is known, but surprising behavior:
> https://github.com/yogstation13/Yogstation/pull/7203

The fix is to switch to `periodic-labeler`, which runs on a `cron` within the scope of `aws-amplify`, vs. a PR event which runs in the restricted context of the fork:
> https://github.com/actions/labeler/issues/12#issuecomment-555746868

https://github.com/marketplace/actions/periodic-labeler

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -6,12 +6,16 @@
 # https://github.com/actions/labeler/blob/master/README.md
 
 name: Labeler
-on: [pull_request]
+on:
+  schedule:
+    - cron: '*/30 * * * *' # Every 1st and 30th minute
 
 jobs:
   label:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/labeler@v2
+    - uses: paulfantom/periodic-labeler@v0.0.1
       with:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
+        GITHUB_REPOSITORY: ${{ github.repository }}
+        LABEL_MAPPINGS_FILE: .github/labeler.yml


.github/workflows/labeler.yml
```diff
@@ -6,12 +6,16 @@
...
+on:
+  schedule:
+    - cron: '*/30 * * * *' # Every 1st and 30th minute
 
 jobs:
   label:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/labeler@v2
+    - uses: paulfantom/periodic-labeler@v0.0.1
```
sammartinez: This makes sense to me on what is needed to have this fix our labeling of PRs, Im not sure on using this repo as we would be possibly exposing secrets in order to label. @mlabieniec what do you think of using this?

sammartinez:
Are we worried about https://github.com/marketplace/actions/periodic-labeler not be certified by github? @mlabieniec

ericclemmons:
It's not an unreasonable concern: the image referenced (owned by GitHub or otherwise) should still be reviewed and referencing an explicit tag.

I generally regard build tooling with the same scrutiny as NPM dependencies (though they have access to the same, if not more, secrets).

This PR is the "quick fix" based on my research, but we have several options to resolve it if this user's fork doesn't meet our standards!

ericclemmons:
There _is_ another option, and that's to own it ourselves:
- https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-a-javascript-action#example-using-a-private-action

I hesitate to do this because _we already own this workflow_, but in a very manual way.

1. Close this, keeping the status quo of manually keeping PRs & labels up-to-date. (This will always exist in _some_ capacity IMO).
2. Merge this automated solution, delegating the problem to a dependency. (Not unlike the purpose of NPM packages)
3. Create a new PR for an automated, private workflow. More effort, and I'd rather have Option 1 if Option 2 isn't desirable for whatever reason.

ericclemmons:
Making the 🕴️_executive_🕴️ decision to close this.

Our PRs overwhelmingly go unlabeled, which makes triaging more difficult (whether by category or "needs review"):
> https://github.com/aws-amplify/amplify-js/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc

Improving our issue -> PR pipeline (per my open-source doc) will also mitigate this, as tooling can couple issues with PRs so that they are collated in the pipeline to completion.

ericclemmons:
Also, no idea how a "levitating man in suit" is a thing.

PR #4583, opened 2019-12-16
feat(@aws-amplify/ui-components): amplify-verify-contact

ericclemmons:
[Delivers #170288336]

**Note: When developing stories & updating `auth-types.ts`, I often have to run `build` to force updated type definitions. I've committed the build artifacts in this PR, but it looks like `ui-components/master` may need `build` ran against it as well**.

- Styles weren't added, as I was able to use preexisting styled components 💪.
- Updated `CognitoUserInterface` to support the `user.unverified` response.
- Story snapshots to validate behavior when `unverified` is/isn't set.
- Re-build definitions & docs.
- Used constants for labels.

### React's `<VerifyContact />`

> verify react

### Skipping validation with `<amplify-verify-contact>`

> verify skip

### Phone validation with `<amplify-verify-contact>`

> verify wc phone

And it worked!

> Screen Shot 2019-12-17 at 10 57 19 AM

---


packages/amplify-ui-components/.storybook/stencil-jsx.js
```diff
@@ -22,11 +22,7 @@ export const h = (tag, props, ...children) => {
         }
       }
 
-      if (['number', 'string'].includes(typeof value)) {
-        node.setAttribute(key, value);
-      } else {
-        node[key] = value;
-      }
+      node[key] = value;
```
ericclemmons: There was a bug where `<amplify-verify-contact authState="verifyContact">` wasn't picked up as a `@Prop()` by Stencil.

This prop has since been removed, but this fixes it so that not just objects & functions are treated as props.

packages/amplify-ui-components/src/common/types/auth-types.ts
```diff
@@ -18,7 +18,7 @@ export enum AuthState {
   SettingMFA = 'settingMFA',
   TOTPSetup = 'TOTPSetup',
   CustomConfirmSignIn = 'customConfirmSignIn',
-  VerifyContact = 'verifycontact',
+  VerifyContact = 'verifyContact',
```
ericclemmons: This was changed to match the formatting from React's implementation.

Since `authState` is no longer passed as a prop by users, I should be able to safely revert it if we want to change these values to _not_ match `<VerifyContact />`.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -84,10 +84,16 @@ export class AmplifyAuthenticator {
...
+      case AuthState.VerifyContact:
+        return <amplify-verify-contact handleAuthStateChange={this.onAuthStateChange} user={this.authData} />;
       case AuthState.TOTPSetup:
         return <amplify-totp-setup handleAuthStateChange={this.onAuthStateChange} user={this.authData} />;
       case AuthState.Loading:
         return <div>Loading...</div>;
+      case AuthState.SignedIn:
+        return <amplify-greetings handleAuthStateChange={this.onAuthStateChange} user={this.authData} />;
+      default:
+        throw new Error(`Unhandled auth state: ${authState}`);
```
ericclemmons: I added this in 5843962a44d7fa2583c294c01830dc6dd222b389 because I kept missing the case difference of `verifycontact` and `verifyContact`: the explicit error helped explain why nothing was rendering.

As a result, I noticed that `AuthState.SignedIn` _wasn't_ handled, and has been returning `null`.

I moved `<amplify-greetings>` up in 24ea71d63fa0cfee90f2bc950c8c18d61e16eb36, which seems more consistent with how the rest of the states are rendered.

packages/amplify-ui-components/src/components/amplify-verify-contact/amplify-verify-contact.react.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+      unverified: {
+        email: 'email@amazon.com',
+        phone_number: '+12345678901',
+      },
+    }}
+    authState="verifyContact"
+    onAuthEvent={(...args) => console.info('onAuthEvent', ...args)}
+    onStateChange={(...args) => console.info('onStateChange', ...args)}
+  />
+);
```
ericclemmons: I can remove this story if we don't want to keep it for comparison. ☝️

packages/amplify-ui-components/src/components/amplify-verify-contact/amplify-verify-contact.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+
+describe(title, () => {
+  describe('stories', () => {
+    Object.entries(templates).forEach(([name, template]) => {
+      it(name, async () => {
+        const page = await newSpecPage({ components, template });
+
+        expect(page.root).toMatchSnapshot();
+      });
+    });
```
ericclemmons: The additional tests I'd like to add may be better suited for Cypress in the future:

- Clicking "Skip" should switch to `AuthState.SignedIn` with `user`.
- Clicking "Verify" should switch to the "submit" view.

packages/amplify-ui-components/src/components/amplify-verify-contact/amplify-verify-contact.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+    if (!user) {
+      logger.debug('no user for verify');
+      return null;
+    }
+
+    const { unverified } = user;
+
+    if (!unverified) {
+      logger.debug('no unverified on user');
```
sammartinez: This seems slightly strange wording. WDYT about the below? 

```suggestion
      logger.debug('Unverified variable does not exist on user');
```

packages/amplify-ui-components/src/common/types/auth-types.ts
ericclemmons: @jordanranz Should we keep this or ✂️ ☝️ out?

packages/amplify-ui-components/src/components/amplify-verify-contact/amplify-verify-contact.tsx
```diff
@@ -0,0 +1,154 @@
...
+import {
+  VERIFY_CONTACT_VERIFY_LABEL,
+  VERIFY_CONTACT_SUBMIT_LABEL,
+  VERIFY_CONTACT_HEADER_TEXT,
+  CODE_PLACEHOLDER,
+  VERIFY_CONTACT_EMAIL_LABEL,
+  VERIFY_CONTACT_PHONE_LABEL,
+} from '../../common/constants';
+
+const logger = new Logger('amplify-verify-contact');
```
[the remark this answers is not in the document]
ericclemmons: My thought was two-fold:

- `<amplify-verify-contact>` is generating these logs, so consistency seems helpful when searching through the firehose of logs when debugging is enabled.

- The existing implementation uses `VerifyContact` to match the component name:

	https://github.com/aws-amplify/amplify-js/blob/867412030de57fd74078b609252de6f7f81ad331/packages/aws-amplify-react/src/Auth/VerifyContact.tsx#L35-L41

If you'd still like this change made, let me know!

packages/amplify-ui-components/src/components/amplify-verify-contact/amplify-verify-contact.tsx
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    Auth.verifyCurrentUserAttributeSubmit(attr, code)
+      .then(data => {
+        logger.debug(data);
+
+        this.handleAuthStateChange(AuthState.SignedIn, this.user);
+        this.verifyAttr = null;
+      })
+      .catch(logger.error);
```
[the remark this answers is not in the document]
ericclemmons: I did a screen share yesterday around this, but I like documenting it nonetheless:

```js
Promise.catch(logger.error)
```

is equivalent to the following, **but without the extra anonymous closure!**

```js
Promise.catch(error => logger.error(error))
```



Here's what goes into considering this:

- [ ] Do I need to force the scope of `this`?

	For class components, this could be helpful for class properties without `this` bound:
	
	```js
	class Foo {
		// ⚠️ Warning! `this` doesn't reference the instance by default.
		handleChange(event) {...}

		// ✅ Better! `this` is always scoped to the instance.
		handleChange = (event) => {
	```

	In the first example, `onClick={(event) => handleChange(event)}` forces `this` to be bound to the instance.  However, this leaks the unbound nature of `handleChange` to the renderer to scope correctly :(

- [ ] Are the supplied arguments the same as those as the function?

	```js
    // ❌ Incorrect! `map`'s arguments aren't the same as `parseInt`'s.
	["1", "2", "3"]).map(parseInt) // [1, null, null]
	// ✅ Correct!
    ["1", "2", "3"]).map((n) => parseInt(n, 10)) // [1, 2, 3]
    ```

- [ ] Does the anonymous closure handle events or transform data?

	```js
	onChange={event => {
		event.preventDefault();
		setInputValue(event.target.value)
	}}
	```

	This can still be cleaned up with a handler function, but it's a matter of preference:

	```js
	const handleChange = (event) => {
		event.preventDefault();
		setInputValue(event.target.value);
	}
	...
	onClick={handleChange}
    ```

- [ ] The wastes a line in the error trace.  Is it worth it?

packages/amplify-ui-components/src/common/types/auth-types.ts
sammartinez: I believe its a typo on myself for moving this enum over from React 😄 so I believe we should keep it and just adjust this string.

packages/amplify-ui-components/src/components/amplify-verify-contact/amplify-verify-contact.tsx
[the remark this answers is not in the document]
sammartinez: Gotcha, we may need a follow up to follow suit with other Logger implementations in other applications as well then.

packages/amplify-ui-components/src/components/amplify-verify-contact/amplify-verify-contact.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+      logger.debug(data);
+      this.handleAuthStateChange(AuthState.SignedIn, this.user);
+      this.verifyAttr = null;
+    } catch (error) {
+      logger.error(error);
+    }
+  }
+
+  async verify(contact: 'email' | 'phone') {
```
sammartinez: should `phone` be `tel`?

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
jordanranz: I think this behavior is ok. Thanks for reworking this.

packages/amplify-ui-components/src/components/amplify-verify-contact/amplify-verify-contact.tsx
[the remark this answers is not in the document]
jordanranz: Yea, just need to agree on which format we would like. This Logger will be used in all supported frameworks so the actual component name will change. For ex: `amplify-verify-contact` and `AmplifyVerifyContact` will both be component names used depending on the framework you use but the logger should represent a common 'shared' name for the component.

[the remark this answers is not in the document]
ericclemmons: I was mainly calling out that this is a different class than what powers `<VerifyContact />`, so using `VerifyContact` for this logger would create confusion.

I'll rename to `AmplifyVerifyContact` instead of `amplify-verify-contact`, since the logs are generated by the class instance.

ericclemmons: This should do it:

```suggestion
const logger = new Logger('AmplifyVerifyContact');
```

packages/amplify-ui-components/src/components/amplify-verify-contact/amplify-verify-contact.tsx
```diff
@@ -0,0 +1,150 @@
...
+      const data = await Auth.verifyCurrentUserAttribute(contact);
+
+      logger.debug(data);
+      this.verifyAttr = contact;
+    } catch (error) {
+      logger.error(error);
+    }
+  }
+
+  render() {
```
jordanranz: Can we have the main `render` at the bottom of the file for consistency?

sammartinez: I agree with that as well

PR #4596, opened 2019-12-18
feat: (@aws-amplify/ui-components) Toast Component

sammartinez:
_Issue #, if available:_
[Delivers #167797443]

_Description of changes:_
* Added Toast Component to be ready to use for any workflows that will need it.
* Added Unit tests

### Screenshots
With Error Message:
with_error

Default Toast:
default_toast


jordanranz:
Can you write a story for this.

packages/amplify-ui-components/src/components/amplify-toast/amplify-toast.tsx
```diff
@@ -0,0 +1,21 @@
+import { Component, Prop, h } from '@stencil/core';
+
+import { toast, toastClose } from './amplify-toast.style';
+
+@Component({
+  tag: 'amplify-toast',
+  shadow: false,
+})
+export class AmplifyToast {
+  /** Used in order to add a clickable `exit` button for the Toast component */
```
jordanranz: This comment is confusing. Does the onClose add the button? or is it just a callback for when the 'X' is clicked?

sammartinez:
> Can you write a story for this.

I can look into it but itll be probably with something of "foo" to get an idea. But let me work with the new mocks I have now

packages/amplify-ui-components/src/components/amplify-toast/amplify-toast.tsx
sammartinez: Yeah i need to adjust this comment as it was just a 'click' for a callback. Ill adjust

packages/amplify-ui-components/src/global/variables.css
```diff
@@ -25,6 +25,8 @@
   --red: #dd3f5b;
   --white: #ffffff;
   --light-blue: #00a1c9;
+  --amazon-yellow-warning: #f8d047;
```
jordanranz: instead of `amazon` can we use `amplify`?

sammartinez: yeah thats fine Ill make the change

packages/amplify-ui-components/src/components/amplify-toast/amplify-toast.tsx
```diff
@@ -0,0 +1,27 @@
...
+  shadow: false,
+})
+export class AmplifyToast {
+  /** Used in order to add a dismissable `x` for the Toast component */
+  @Prop() onClose: () => void;
+
+  /* 
+  TODO #170365145: Work on a helper function that will populate and 
+  update class colors for success / warning / failure messages 
+  */
```
ericclemmons: I'm good to ship this scaffolding so that the other states from the mockup happen in another task.

It'll break the snapshot tests, so they'll need to be updated (since we wouldn't want the default to be `warning`).

🚢

Issue #4599, opened 2019-12-13
Amplify/Auth "Not Supported" (Typescript-Nodejs)


AdanJSuarez:
Related to Auth, Cognito. 
I'm trying to connect my backend (typescript-nodejs) with the user pool. I had the project done before I `amplify init` in the root folder (I don't know if this detail is relevant or not). I followed all steps of the video-instruction.

I included `Amplify.configure(awsexports)` in my main module (the entry point for the project)

I had to do two hacks to make it built: 
1.- Create a `index.d.ts` file with this inside : `declare module "graphql/language/ast" { export type DocumentNode = any }` to avoid a ts compile error with graph/language/ast that I don't even use.
2.- I had to rename `aws-exports.js` for `aws-exports.ts` and modify `awsmobile` inside it adding `awsmobile: any` to avoid again ts compile errors.

After all of that when I build the project, it gives me the following error.

Any help here? Thanks!

/Users/adanjsuarez/Projects/CDIOmega/cdiomega-backend-inputdata/node_modules/@aws-amplify/auth/dist/aws-amplify-auth.js:13322
    throw new Error('Not supported');
    ^

Error: Not supported
    at ./lib-esm/urlListener.js.__webpack_exports__.default (/Users/adanjsuarez/Projects/CDIOmega/cdiomega-backend-inputdata/node_modules/@aws-amplify/auth/dist/aws-amplify-auth.js:13322:11)
    at AuthClass.configure (/Users/adanjsuarez/Projects/CDIOmega/cdiomega-backend-inputdata/node_modules/@aws-amplify/auth/dist/aws-amplify-auth.js:10265:67)
    at /Users/adanjsuarez/Projects/CDIOmega/cdiomega-backend-inputdata/node_modules/@aws-amplify/core/dist/aws-amplify-core.js:2548:12
    at Array.map (<anonymous>)
    at Function.Amplify.configure (/Users/adanjsuarez/Projects/CDIOmega/cdiomega-backend-inputdata/node_modules/@aws-amplify/core/dist/aws-amplify-core.js:2547:22)
    at Object.<anonymous> (/Users/adanjsuarez/Projects/CDIOmega/cdiomega-backend-inputdata/out/src/Main.js:12:23)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
[...]

Issue #4576, opened 2019-12-14
SSR error "Not supported"


powerfulj:
Hi,

I got an error when run the build for Server Side Rendering, error is "Not Supported".

The source code has a browser check which cause this issue.

package version is:

"@aws-amplify/auth": "2.1.1"

Thanks
CJ

Amplifiyer:
@powerfulj can you try this https://github.com/aws-amplify/amplify-js/issues/4243#issuecomment-564553889 and see if that helps?

I believe the issue is that nextJs apps are using the `main` field of pacakge.json for entry point rather than `module` and so consumes the webpack bundles which are not compatible with nodejs (webpack does polyfill for nodejs process and doesn't have process.versions.node set). Hence overriding the browserOrNode() in your app (before executing the library) skips the pollyfills and return the right results i.e. isNode = true.

powerfulj:
> @powerfulj can you try this [#4243 (comment)](https://github.com/aws-amplify/amplify-js/issues/4243#issuecomment-564553889) and see if that helps?
> 
> I believe the issue is that nextJs apps are using the `main` field of pacakge.json for entry point rather than `module` and so consumes the webpack bundles which are not compatible with nodejs (webpack does polyfill for nodejs process and doesn't have process.versions.node set). Hence overriding the browserOrNode() in your app (before executing the library) skips the pollyfills and return the right results i.e. isNode = true.

I changed my code by following the comment which works great! Tons of thanks mate.

Actually I'm not using NextJS but using Express that customized all the SSR settings.

Thanks again!

PR #4611, opened 2019-12-20
fixes #4243 #4576 #4599 Correctly identify the nodejs environment.

### LINKED ISSUE #4243: "WINDOW NOT DEFINED"... again :-(
**Describe the bug**
Importing the latest aws-amplify into a nuxt application crashes with a "Window not defined" error

**To Reproduce**
In a Nuxt Js application (and probably any SSR framework), try to import the latest aws-amplify

**Expected behavior**
Should not crash

**Additional context**
Most certainly a regression for #3477 

My latest working config is with aws-amplify version 1.1.36
with the following resolutions in package.json:
```json
  "resolutions": {
    "@aws-amplify/analytics": "1.2.23",
    "@aws-amplify/api": "1.0.42",
    "@aws-amplify/auth": "1.2.31",
    "@aws-amplify/cache": "1.0.32",
    "@aws-amplify/core": "1.1.0",
    "@aws-amplify/interactions": "1.0.35",
    "@aws-amplify/predictions": "1.0.3",
    "@aws-amplify/pubsub": "1.1.0",
    "@aws-amplify/storage": "1.1.0"
 }
```

Amplifiyer:
_Issue #, if available:_#4243 #4576 #4599

_Description of changes:_ 

fixes #4243 #4576 #4599 Correctly identify the nodejs environment.

Webpack adds node shim to the bundle overriding the `process` variable
 causing issues detecting nodejs environment. Creating a new function will
 help to avoid incorrect environment detection as new function will have
 it's `this` binded to global scope. Credit: https://stackoverflow.com/a/31090240


packages/core/src/JS.ts
```diff
@@ -165,17 +165,26 @@ export default class JS {
...
-		const isNode =
-			typeof process !== 'undefined' &&
-			process.versions != null &&
-			process.versions.node != null;
+		// function to check if the global scope is "window"
+		const isBrowser = new Function(
+			'try {return this===window;}catch(e){ return false;}'
+		);
+
+		// function to test if global sope is binded to "global"
```
haverchuck: should this be global 'scope'?

haverchuck:
@Amplifiyer How will these checks on this === window or this ==== global be affected if they are using some sort of polyfill for either?

Amplifiyer:
> @Amplifiyer How will these checks on this === window or this ==== global be affected if they are using some sort of polyfill for either?

@haverchuck That's where new Function() helps since it binds to the global scope while polyfills etc would only impact the inner scope. This PR is for this exact reason since webpack was creating polyfills in the bundles and nodejs environments were not able to detect that it's running in Nodejs and causing errors "Not supported"

packages/core/src/JS.ts
Amplifiyer: yes, it's a typo, will update it.

Amplifiyer:
> This is pretty clever. Differentiating the environment at runtime within webpack bundles has been a pain-point for a while since `webpack.DefinePlugin` started polyfilling `process.env.NODE_ENV`.
> 
> I don't see anything that sticks out to warrant blocking, but have you considered how this could be tested?
> 
> Jest has support for making the test environment run as `jsdom` (default) or `node`, so it may be possible to have two different test files running within each environment to validate & prevent future regressions:
> 
> > https://jestjs.io/docs/en/configuration#testenvironment-string

@ericclemmons , that's a very good point. I'm struggling to figure out how to test the webpack built bundle (which is where the polyfills are causing issues) with jest. Currently Jest tests the actual code and not the built code. Maybe import from dist if it's the recommended way?

ericclemmons:
@Amplifiyer I know React has success testing against the built code (vs. the source), specifically for this reason.

With the combination of using the `dist` output and setting `@jest-environment jsdom` and `@jest-environment node`, this could be our best option.

Do you think that'd correctly match the scenarios in the issues?

Amplifiyer:
> @Amplifiyer I know React has success testing against the built code (vs. the source), specifically for this reason.
> 
> With the combination of using the `dist` output and setting `@jest-environment jsdom` and `@jest-environment node`, this could be our best option.
> 
> Do you think that'd correctly match the scenarios in the issues?

Yes exactly the scenarios we want to test. I'll look into react tests.

Amplifiyer:
Added unit tests and verified that the node test fails with the current master and passes with this PR.

PR #4617, opened 2019-12-21
fix(@aws-amplify/datastore) subscription for owner authorization

elorzafe:
_Issue #, if available:_
https://github.com/aws-amplify/amplify-cli/issues/3048

_Description of changes:_
Adding subscription variables for owner authorization use case https://aws-amplify.github.io/docs/cli-toolchain/graphql#owner-authorization
small fix related to AppSyncRealTimeProvider (`observer undefined`)

Tested with this schema
```graphql
type Post @model @auth(rules: [
  {allow: owner, ownerField: "editors", provider: userPools, operations: [update]},
  {allow: owner, ownerField: "owner", identityClaim: "sub", provider: userPools, operations: [create]},
]) {
  id: ID!
  title: String! 
  owner: String
}

type Note @model {
  id: ID!
  title: String!
  content: String!
}
```


packages/datastore/src/sync/utils.ts
```diff
@@ -107,6 +107,42 @@ function getConnectionFields(modelDefinition: SchemaModel): string[] {
...
+		.concat(modelDefinition.attributes)
+		.find(attr => attr && attr.type === 'auth');
+
+	const { properties: { rules = [] } = {} } = authConfig || {};
+	const ownerRules = rules.filter(rule => rule.allow === 'owner');
+
+	let result: [boolean, string, string, string] = [false, '', '', ''];
+
+	// Multiple rules can be declared for allow: owner
+	ownerRules.forEach(ownerRule => {
```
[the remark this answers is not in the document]
yuth: Does this mean only the last owner rule wins? The lib should account all the owner rules, and send the best matching owner rules instead of the last one
```gql
type Post
  @model
  @auth(
    rules: [
      {allow: owner, ownerField: "owner", identityClaim: "cognito:username"},
      {allow: owner, ownerField: "owner", identityClaim: "custom:username"},
    ])
{
  id: ID!
  title: String!
  owner: String
}
```

In the above example, if the JWT token does not have `custom:username`, but has `cognito:username`, this subscription would fail. We probably collect all the rules and see what is available in the token and send that appropritately.

packages/datastore/src/sync/utils.ts
```diff
@@ -107,6 +107,42 @@ function getConnectionFields(modelDefinition: SchemaModel): string[] {
...
+		// setting defaults for backwards compatibility with old cli
```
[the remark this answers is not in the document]
yuth: The updated CLI would always have these fields in the owner rule. So this should not cause any problems

packages/datastore/src/sync/utils.ts
```diff
@@ -107,6 +107,42 @@ function getConnectionFields(modelDefinition: SchemaModel): string[] {
...
+	modelDefinition: SchemaModel,
+	transformerOpType: TransformerMutationType
+): [boolean, string, string, string] {
+	// Searching for owner authorization on attributes
+	const authConfig = []
+		.concat(modelDefinition.attributes)
+		.find(attr => attr && attr.type === 'auth');
+
+	const { properties: { rules = [] } = {} } = authConfig || {};
+	const ownerRules = rules.filter(rule => rule.allow === 'owner');
```
yuth: It looks like we are interested in only `allow:owner` rules and are completely ignoring `allow:groups`. Say if you're in the allowed groups, group rule would allow you to get more result from subscription and we should if we want to do group or owner rule based on whats available in the token

dconlon-monigo: +1 on this, any solution needs to handle mixed group/owner auth rules. I'm hitting up against that precise case.

packages/datastore/src/sync/utils.ts
```diff
@@ -107,6 +107,42 @@ function getConnectionFields(modelDefinition: SchemaModel): string[] {
 	return result;
 }
 
+export function isOwnerAuthorization(
+	modelDefinition: SchemaModel,
+	transformerOpType: TransformerMutationType
+): [boolean, string, string, string] {
```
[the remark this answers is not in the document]
elorzafe: done, thanks

packages/datastore/src/sync/utils.ts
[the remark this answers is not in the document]
elorzafe: fixed

packages/datastore/src/sync/utils.ts
elorzafe: I updated the PR considering groups and also added iam and apiKey as well

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -1,16 +1,32 @@
+import '@aws-amplify/pubsub';
+
+import Observable from 'zen-observable-ts';
+
 import API, { GraphQLResult } from '@aws-amplify/api';
+import { GRAPHQL_AUTH_MODE } from '@aws-amplify/api/lib-esm/types';
```
Amplifiyer: we shouldn't be exporting from generated source

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -23,20 +39,190 @@ class SubscriptionProcessor {
 	][] = [];
 	private dataObserver: ZenObservable.Observer<any>;
 
-	constructor(private readonly schema: InternalSchema) {
-		this.generateQueries();
+	constructor(private readonly schema: InternalSchema) {}
+
+	private buildSubscription(
+		model: SchemaModel,
```
Amplifiyer: Just for my understanding, why don't we need to build the subscription in the constructor anymore?

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -23,20 +39,190 @@ class SubscriptionProcessor {
...
+				oidcTokenPayload
+			) || {};
+
+		const [opType, opName, query] = buildSubscriptionGraphQLOperation(
+			model,
+			transformerMutationType,
+			isOwner,
+			ownerField
+		);
+		return { authMode, opType, opName, query, isOwner, ownerField, ownerValue };
```
Amplifiyer: Can we create types for these?

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -23,20 +39,190 @@ class SubscriptionProcessor {
...
+			rule => rule.authStrategy === 'public' && rule.provider === 'apiKey'
+		);
+
+		if (apiKeyAuth) {
+			result = { authMode: GRAPHQL_AUTH_MODE.API_KEY, isOwner: false };
+		}
+
+		if (result) {
+			return result;
+		}
```
Amplifiyer: we can just return after line 98 in the first `if`

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -23,20 +39,190 @@ class SubscriptionProcessor {
...
+				);
+
+			if (iamPrivateAuth) {
+				result = { authMode: GRAPHQL_AUTH_MODE.AWS_IAM, isOwner: false };
+			}
+		}
+
+		if (result) {
+			return result;
+		}
```
Amplifiyer: same as above, just return when you have a positive result. this will also avoid checking for iamPrivateAuth when it's confirmed that it's iamPublishAuth.

On the second thought, can there be both unauth and auth rules? If yes, is there is a priority?

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -23,20 +39,190 @@ class SubscriptionProcessor {
...
+						authMode: GRAPHQL_AUTH_MODE.AMAZON_COGNITO_USER_POOLS,
+						isOwner: false,
+					};
+				}
+			});
 		});
+
+		if (result) {
+			return result;
+		}
```
Amplifiyer: Same (and later in the file as well), return when you have the first result to make executing a bit faster.

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -60,85 +246,153 @@ class SubscriptionProcessor {
...
+			let userCredentials = USER_CREDENTIALS.none;
+			(async () => {
+				try {
+					// retrieving current AWS Credentials
+					const credentials = await Auth.currentCredentials();
+					userCredentials = credentials.authenticated
+						? USER_CREDENTIALS.auth
+						: USER_CREDENTIALS.unauth;
+				} catch (err) {
+					// best effort to get AWS credentials
```
Amplifiyer: Shouldn't we log something here?

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -60,85 +246,153 @@ class SubscriptionProcessor {
...
+									ownerValue,
+									authMode,
+								}) => {
+									const marker = {};
+
+									if (isOwner) {
+										if (!ownerValue) {
+											// Check if there is an owner field, check where this error should be located
+											observer.error(
+												'Owner field required, sign in is need in order to perform this operation'
```
Amplifiyer: nit in the user error message: `...sign in is needed in order...`

packages/datastore/src/sync/utils.ts
```diff
@@ -118,20 +198,9 @@ export function buildGraphQLOperation(
...
-		case 'SUBSCRIBE':
-			subscriptions = [
-				TransformerMutationType.CREATE,
-				TransformerMutationType.UPDATE,
-				TransformerMutationType.DELETE,
-			].map(op => {
-				const opName = `on${op}${typeName}`;
-				return [op, opName];
-			});
-			break;
```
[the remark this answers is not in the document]
elorzafe: this is for internal use only, I also removed it from the original enum

packages/datastore/src/sync/processors/subscription.ts
elorzafe: agreed, I will fix that

packages/datastore/src/sync/processors/subscription.ts
elorzafe: Before it wasn't checking Auth status on runtime to validate tokens and sign-in sign-out status. Now is done on runtime when the subscriptions is going to be established. The problem now will be how to stop/start this when sign-in/sign-out occurs, but I think that should be in a separate PR. What do you think?

packages/datastore/src/sync/processors/subscription.ts
elorzafe: I don't really see the value of those because are used only there. In case is used on multiple places I totally agree to create a type

packages/datastore/src/sync/processors/subscription.ts
elorzafe: Is a best effort operation just to check whether or not is available, to find which rule fits better

packages/datastore/src/sync/processors/subscription.ts
elorzafe: Thanks

packages/datastore/src/sync/processors/subscription.ts
elorzafe: The problem with this one is that it returns on the callback inside `forEach` that is why I have to return it outside.

packages/datastore/src/sync/processors/subscription.ts
elorzafe: I agree, thanks!

packages/datastore/src/sync/processors/subscription.ts
elorzafe: I agree, my concern was on the following return statement but this are ok to return immediately.

It can be both rules, so in case is `public` is ok to be <unauth> or <auth> user (if you have credentials), less restrictive case, otherwise if is `private` it must be <auth> user

packages/datastore/src/sync/processors/subscription.ts
Amplifiyer: I didn't fully understand. If this is not regressing from the current feature/functionality, we can wait for a separate PR.

packages/datastore/src/sync/processors/subscription.ts
Amplifiyer: yeah makes sense. I guess we can just break in the for loop once we found one?

elorzafe: Yeah, I agree I can use something different rather than foreach. It seems foreach is not clean to break (https://stackoverflow.com/questions/2641347/short-circuit-array-foreach-like-calling-break)

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -78,14 +77,15 @@ class SubscriptionProcessor {
...
-		oidcTokenPayload: { [field: string]: any } | undefined
+		cognitoTokenPayload: { [field: string]: any } = {},
+		oidcTokenPayload: { [field: string]: any } = {}
 	): {
 		authMode: GRAPHQL_AUTH_MODE;
 		isOwner: boolean;
 		ownerField?: string;
 		ownerValue?: string;
 	} {
+		debugger;
```
Amplifiyer: Do we need this in production code?

PR #4624, opened 2019-12-23
feat(aws-amplify-react-native): Upgrade react-native-elements to v1

artursvonda:
Upgraded react-native-elements to v1 in aws-amplify-react-native. And removed react-native-vector-icons from dependencies as that is peer dependency.
The react-native-elements is used in only one view and no changes were required.

As a side-note, maybe it makes sense to remove the dependency altogether as it pulls large framework just to be used for one Icon in one component.


artursvonda:
Bump

artursvonda:
Resolved conflicts

amhinson:
Thanks for the PR @artursvonda! But we are removing react-native-elements & react-native-vector-icons as dependencies, so this will no longer be needed: #6817

PR #4647, opened 2019-12-27
fix(aws-amplify-react): Fix peer dependencies

artursvonda:
Also fix peer dependencies for `aws-amplify-react-native` and `aws-amplify-angular`.

Matched packages used in `aws-amplify-react` to peerDependencies in order to more precisely express dependencies.

Also, looks like some peer dependencies were not updated when migrating versions to 2.x.

### DIFF
--- a/packages/aws-amplify-react/package.json
+++ b/packages/aws-amplify-react/package.json
@@ -95,6 +95,12 @@
     "regenerator-runtime": "^0.11.1"
   },
   "peerDependencies": {
-    "aws-amplify": "^1.x"
+    "@aws-amplify/analytics": "^2.0.0",
+    "@aws-amplify/api": "^2.0.0",
+    "@aws-amplify/auth": "^2.0.0",
+    "@aws-amplify/core": "^2.0.0",
+    "@aws-amplify/interactions": "^2.0.0",
+    "@aws-amplify/storage": "^2.0.0",
+    "@aws-amplify/ui": "^2.0.0"
   }
 }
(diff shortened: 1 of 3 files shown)


packages/aws-amplify-react/package.json
```diff
@@ -95,6 +95,12 @@
...
   },
   "peerDependencies": {
-    "aws-amplify": "^1.x"
+    "@aws-amplify/analytics": "^2.0.0",
+    "@aws-amplify/api": "^2.0.0",
+    "@aws-amplify/auth": "^2.0.0",
+    "@aws-amplify/core": "^2.0.0",
+    "@aws-amplify/interactions": "^2.0.0",
+    "@aws-amplify/storage": "^2.0.0",
+    "@aws-amplify/ui": "^2.0.0"
```
Amplifiyer: Just curious, why not update the `aws-amplify` package itself instead of adding the child packages as in `aws-amplify-angular` and `aws-amplify-react-native` packages?

artursvonda: The code actually uses these packages directly, instead of using `aws-amplify` package. So this better expresses the actual dependencies. And it doesn't pull anything unrelated. And this doesn't pull unrelated dependencies.

Amplifiyer:
@artursvonda changes look good. There is a minor conflict from merging master, can you quickly resolve that so I can merge this?

artursvonda:
@Amplifiyer Done

helloPiers:
I just raised #5224 as there is no version 2.0.0 of @aws-amplify/ui package.

Issue #4311, opened 2019-10-31
GraphQLError: adapter is not a function Next.js getInitialProps


ryanmalesic:
**Describe the bug**
GraphQLError: adapter is not a function
      at APIClass.<anonymous> (/Users/ryanmalesic/Projects/tdee-online/node_modules/@aws-amplify/api/dist/aws-amplify-api.js:5565:24)
      at step (/Users/ryanmalesic/Projects/tdee-online/node_modules/@aws-amplify/api/dist/aws-amplify-api.js:4733:19)
      at Object.throw (/Users/ryanmalesic/Projects/tdee-online/node_modules/@aws-amplify/api/dist/aws-amplify-api.js:4663:14)
      at rejected (/Users/ryanmalesic/Projects/tdee-online/node_modules/@aws-amplify/api/dist/aws-amplify-api.js:4625:32)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (interna

**To Reproduce**

Query in getinitialProps of a nextjs app

**Expected behavior**
It should query, not throw

ryanmalesic:
Doing more stack tracing, this seems to be an axios issue. I have put a discussion in there. I will close this if an axios fix can correct the issue

jordanranz:
Hey @ryanmalesic, is there an issue open to track the axios discussion?

ryanmalesic:
#https://github.com/axios/axios/issues/2507

weizliu:
Any updates? I ran into this issue too

weizliu:
Is the change in 1.2.4 already? I ran into the same issue while using 1.2.4 today.

aliankarali:
Still having this issue with `@aws-amplify/api@2.1.1` on server side. Axios issue is closed, I don't know if it is related to amplify or axios.

simpleshadow:
Getting the same error when running an `aws-amplify/api` query with Next.js (not in `getInitialProps`).

simpleshadow:
I got rid of the error using what I needed from. https://github.com/aws-amplify/amplify-js/issues/4305#issuecomment-552193872

aliankarali:
@simpleshadow,  thanks for the tip, I got it working by locking down the versions with resolutions as well.

lambda0xff:
Getting the same error with a Nuxt.js application after trying to upgrade aws-amplify to the latest (^2.2.1)

[...]

Issue #4499, opened 2019-12-03
After updating @aws-amplify/auth to latest version(1.5.0) test cases is failing because of failed to mock Auth module.


anis-shaikh:
After updating @aws-amplify/auth to latest version(1.5.0) test cases is failing because of failed to mock Auth module.

1. Import Auth API by
import Auth from '@aws-amplify/auth';

2. mock it by jest
jest.mock('@aws-amplify/auth');

3. See error
Failed to get mock metadata: /Users/an***/Documents/s***/node_modules/@aws-amplify/auth/index.js

` Test suite failed to run

    Failed to get mock metadata: /Users/an***/Documents/s***/node_modules/@aws-amplify/auth/index.js

    See: https://jestjs.io/docs/manual-mocks.html#content

    > 1 | import Auth from '@aws-amplify/auth';`

ericclemmons:
@anis-shaikh Thanks for bringing this to your attention, and your clear steps to reproduce 👏!

I was able to reproduce this here:

> https://codesandbox.io/s/httpsgithubcomaws-amplifyamplify-jsissues4499-sr9l1

It looks like the regression was introduced in #4007 and the bump either in 233dd1c56cb637193e4b44437c38853359d0de59 or c0fa439b28672a02662bc160058e2a0fb2bf4c13.

We'll need to investigate the difference in what's being installed so that Jest is able to mock it successfully...

### `@aws-amplify/auth@1.3.3`

<img width="1327" alt="1 3 3" src="https://user-images.githubusercontent.com/15182/70078714-0fb66980-15b8-11ea-844e-4dec9ad9bcda.png">

### `@aws-amplify/auth@1.4.2`

<img width="1330" alt="1 4 2" src="https://user-images.githubusercontent.com/15182/70078716-104f0000-15b8-11ea-9aab-4398c8136650.png">

### `@aws-amplify/auth@1.4.3`

<img width="1331" alt="1 4 3" src="https://user-images.githubusercontent.com/15182/70078717-104f0000-15b8-11ea-8ccf-bceb7e95e661.png">

### `@aws-amplify/auth@1.5.0`
<img width="1327" alt="1 5 0" src="https://user-images.githubusercontent.com/15182/70078718-10e79680-15b8-11ea-94e0-d0ee33e5414f.png">

ericclemmons:
For an easy way to compare the differences between what's published, I've been using:

1. https://unpkg.com/browse/@aws-amplify/auth@1.4.2/
[...]

PR #4678, opened 2020-01-07
fix: #4311 Update main entry field to point to CJS builds instead of …

Amplifiyer:
…webpack bundles

_Issue #, if available:_ fixes: #4311 #4499 

_Description of changes:_

### DIFF
--- a/packages/analytics/package.json
+++ b/packages/analytics/package.json
@@ -2,7 +2,7 @@
   "name": "@aws-amplify/analytics",
   "version": "2.2.1",
   "description": "Analytics category of aws-amplify",
-  "main": "./index.js",
+  "main": "./lib/index.js",
   "module": "./lib-esm/index.js",
   "typings": "./lib-esm/index.d.ts",
   "publishConfig": {
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -2,7 +2,7 @@
   "name": "@aws-amplify/api",
   "version": "2.1.2",
   "description": "Api category of aws-amplify",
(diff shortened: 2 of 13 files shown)


Amplifiyer:
#### Jest's auto-mocking bug (#4499)

Verified locally that this PR fixes the mocking issue with @aws-amplify/auth in #4499 

#### Confirm that a production build of CRA with latest vs. #4678 yields the same size.

CRA app with latest amplify version
```console
  500.75 KB          build/static/js/2.0357ba12.chunk.js
  3.65 KB            build/static/css/2.a1f9020b.chunk.css
  1.1 KB             build/static/js/main.4d3758ea.chunk.js
  762 B              build/static/js/runtime~main.a8a9905a.js
  523 B              build/static/css/main.f5532a38.chunk.css
```

CRA app with this PR changes
```console
  500.82 KB (+63 B)  build/static/js/2.aae62737.chunk.js
  3.65 KB            build/static/css/2.a1f9020b.chunk.css
  1.09 KB (-2 B)     build/static/js/main.d877e69f.chunk.js
  762 B              build/static/js/runtime~main.a8a9905a.js
  523 B              build/static/css/main.f5532a38.chunk.css
```

#### Since the ./index.js entry-point is no longer referenced, does that mean we lose the process.env.NODE_ENV === "production") behavior?
Yes, it'll only be applicable to users who are explicitly depending on the root index.js. If we feel it's no longer important we can even delete the index.js and figure out a better way to publish the webpack bundles.

PR #4692, opened 2020-01-09
Fixes #4338 a11y support for React and Angular anchors

### LINKED ISSUE #4338: Authenticator Form is not accessible
**Describe the bug**
The links in the React's Authenticator form are not accessible. The user cannot navigate in the form links using the keyboard's `tab` button.
For example, on the Sign In, the links `Reset password` or `Create account` are not accessible.

**Expected behavior**
The user should be able to access all the links in the form with the `tab` button.

**Screenshots**
Screenshot 2019-11-04 at 16 53 55

**Additional context**
Added an `href=""` to the links will make them accessible.

ashika01:
_Issue #, if available:_ Fixes #4338 

_Description of changes:_ This PR adds tab support for React and Angular anchors by changing the anchor tags to button. As per [W3 org specification](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element), `<a>` without href becomes more of a placeholder. Since our functionality lean more towards button and since button also gives preoper tab ordering, I have moved to `<button>`. 

Reference: https://stackoverflow.com/questions/10510191/valid-to-use-a-anchor-tag-without-href-attribute/10510353


packages/aws-amplify-react/src/Amplify-UI/Amplify-UI-Components-React.tsx
```diff
@@ -281,9 +281,9 @@ export const Link = props => {
 	const style = propStyle(props, theme.a);
 	const p = objectLessAttributes(props, 'theme');
 	return beforeAfter(
-		<a {...p} className={AmplifyUI.a} style={style}>
+		<button {...p} className={AmplifyUI.linkButton} style={style}>
```
sawyerh, line 284: All of these `button` elements should likely have a `type` attribute set to `"button"` since otherwise, [the default value is `"submit"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type), which is likely not what you want.

ashika01: good call out. It should be button. I will push that in.

PR #4732, opened 2020-01-17
feat(@aws-amplify/ui-components): amplify-authenticator data-test attributes

jordanranz:
_Description of changes:_
- `amplify-authenticator` first pass data-test attributes
- Clean up constants to gain parody with legacy UI libs
- Clean up css variables for input styles


packages/amplify-ui-components/src/common/constants.ts
```diff
@@ -54,9 +54,9 @@ export const COUNTRY_DIAL_CODE_DEFAULT = '+1';
 
 // Sign In common constants
 // TODO: Change to SIGN_IN_HEADER_TEXT
-export const HEADER_TEXT = 'Sign into your account';
+export const HEADER_TEXT = 'Sign in to your account';
```
sammartinez: For the changes in these constants, these should be directly aligned with what we inside of other framework packages. Is this not the case? Just don't want to break integration testing

jordanranz: This is to match other frameworks/integration testing

packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.tsx
```diff
@@ -48,9 +49,15 @@ export class AmplifyFormSection {
   @Prop() headerText: string = 'Amplify';
   /** (Optional) Overrides default styling */
   @Prop() overrideStyle?: boolean = false;
+  /** String prefix for the data-test attributes in this component primarily used for testing purposes */
+  @Prop() testDataPrefix?: string = 'form-section';
```
ericclemmons: Everything else is looking good!  Can you help me understand why a `data-` attribute is bleeding into a `@Prop`?

From what I see, it's basically prop-drilling so that `data-test="sign-in-sign-in-button"` (https://github.com/aws-amplify/amplify-js/pull/4732/files#diff-7ac37451969a9c429f5c9f7dd947f1ddR5) instead of the default `data-test="form-section-form-section-button"`.

Is this to solve a specificity problem in a cypress test?  Would you be in favor of a solution that would remove the need for `testDataPrefix`?

jordanranz: In addition to `data-test="sign-in-sign-in-button"` this is also used for `data-test="sign-in-header-section"`, `data-test="sign-up-header-section"`, and `data-test="forgot-password-header-section"`. The purpose of this prop was to give context to the general `form-section` component.

I was a little hesitant to implement this approach but this was the simplest approach to get to Cypress tests at parody. I am in favor of removing it but we would need to come up with a way to give context to `form-section` to generate the proper `data-test` attributes.

ericclemmons: Can you point me to the tests that were failing until this change was made?

jordanranz: `data-test="sign-in-header-section"`:
https://github.com/aws-amplify/amplify-js-samples-staging/blob/master/cypress/integration/auth/authenticator.spec.js#L20

`data-test="sign-up-header-section"`:
https://github.com/aws-amplify/amplify-js-samples-staging/blob/master/cypress/integration/auth/authenticator.spec.js#L77

`data-test="forgot-password-header-section"`:
https://github.com/aws-amplify/amplify-js-samples-staging/blob/master/cypress/integration/auth/authenticator.spec.js#L117

Issue #4666, opened 2020-01-04
1.2.3+ versions are breaking plain javascript/html imports


ajhool:
**Describe the bug**
Up to version 1.2.2, it was possible to import Amplify as a script in an html page by retrieving the built file from /node_modules/aws-amplify/dist/amplify.js . After version 1.2.2, the script fails to load Tested in 1.2.3, 1.2.4, 2.2.1 and all fail with same error.

Error: 
```
TypeError: _aws_amplify_core__WEBPACK_IMPORTED_MODULE_0__ is undefinedPredictions.js:20:13
    js Predictions.js:20
    Webpack 9
        __webpack_require__
        js
        __webpack_require__
        js
        __webpack_require__
        <anonymous>
        <anonymous>
        webpackUniversalModuleDefinition
        <anonymous>
```

**To Reproduce**
Steps to reproduce the behavior (Linux):

1. Create a new node project which will be used to host the static website
```
npm init
npm install -g http-server
npm install aws-amplify@1.2.2 --save --save-exact
cp ./node_modules/aws-amplify/dist/amplify.js ./amplify.js

# create an index.html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>title</title>
    <link rel="stylesheet" href="style.css">
    <script src="script.js"></script>
  </head>
  <body>
    <!-- page content -->
    <script type="text/javascript" src="amplify.js"></script> 
  </body>
</html>

http-server

# Visit localhost:8080 and inspect the console, there should not be an amplify error

npm install aws-amplify@1.2.3 --save --save-exact
cp ./node_modules/aws-amplify/dist/amplify.js ./amplify.js

# Visit localhost:8080 and inspect the console, there should be an amplify error
```

**Expected behavior**
Amplify is the recommended javascript aws sdk and it must be able to load in plain html / javascript.

<details>

N/A

</details>

**Smartphone (please complete the following information):**

- Device: Thinkpad
- OS: Ubuntu
- Browser Mozilla

**Additional context**

**Sample code**
Include additional sample code or a sample repository to help us reproduce the issue. (Be sure to remove any sensitive data)

**_You can turn on the debug mode to provide more info for us by setting window.LOG_LEVEL = 'DEBUG'; in your app._**

Amplifiyer:
[...]

PR #4745, opened 2020-01-21
fixes: #4666 Remove externals from amplify's root webpack build

Amplifiyer:
_Issue #, if available:_ #4666

_Description of changes:_ fixes: #4666 Remove externals from amplify's root webpack build


Amplifiyer:
> Can you share how you validated?
> 
> I assume the error in #4666 went away and import worked within a plain HTML app?

@ericclemmons Yes, verified that the Amplify variable was present in the window scope and no errors.

PR #4767, opened 2020-01-23
fix(@aws-amplify/ui-components): Move Auth & Core to externals

ericclemmons:
When working on https://github.com/aws-amplify/amplify-js-samples-staging/pull/42, we found that Stencil + Rollup was including `@aws-amplify/auth` and `@aws-amplify/core` in the bundle, bloating file-size and breaking `Amplify.configure` from working at all.

- [x] Explicitly add these packages to `externals` via a rollup plugin
- [x] Add a test for tracking bundle-sizes, since this is both an indicator of performance and possible bundling issues.
- [x] Pre-build before running the test.

Stencil will run a `dev` build prior to tests, but `bundle-size.spec.js` relies on `dist/` being present, meaning a production build must be ran for those prod-specific files to exist.

Otherwise, there will be several dozen files missing, causing the test & snapshot to fail.

1. `yarn workspace @aws-amplify/ui-components build`
1. `yarn workspace @aws-amplify/ui-components test` (or `test:watch`) and run `bundle-size.spec.js`.

---


packages/amplify-ui-components/__tests__/bundle-size.spec.js
```diff
@@ -0,0 +1,30 @@
...
+  const i = Math.floor(Math.log(bytes) / Math.log(k));
+
+  return parseFloat((bytes / Math.pow(k, i)).toFixed(dm)) + ' ' + sizes[i];
+}
+
+describe('@amplify/ui-components bundle size', () => {
+  it('should match snapshot', async () => {
+    const stats = await globby('**', { cwd: dist, stats: true });
+    const sizes = stats.reduce((acc, stat) => {
+      acc[stat.path] = formatBytes(stat.size);
```
ericclemmons: Exact `stat.size` differs too much between builds/tweaks.  Rounding to Bytes/KB/MB works well enough to reduce noise.

packages/amplify-ui-components/stencil.config.ts
```diff
@@ -1,12 +1,19 @@
...
 import nodePolyfills from 'rollup-plugin-node-polyfills';
 
 export const config: Config = {
   excludeSrc: ['**/*.e2e.*', '**/*.spec.*', '**/*.stories.*'],
   namespace: 'amplify-ui-components',
-  plugins: [nodePolyfills()],
+  plugins: [
+    externals({
+      // deps to include in externals (default: [])
+      include: ['@aws-amplify/auth', '@aws-amplify/core'],
```
ericclemmons: Possibly re-use https://github.com/aws-amplify/amplify-js/blob/master/scripts/rollup-externals.js?

packages/amplify-ui-components/__tests__/bundle-size.spec.js
```diff
@@ -0,0 +1,30 @@
...
+  const k = 1024;
+  const dm = decimals < 0 ? 0 : decimals;
+  const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+  const i = Math.floor(Math.log(bytes) / Math.log(k));
+
+  return parseFloat((bytes / Math.pow(k, i)).toFixed(dm)) + ' ' + sizes[i];
+}
+
+describe('@amplify/ui-components bundle size', () => {
+  it('should match snapshot', async () => {
```
ericclemmons: Considering a `before` script to execute `build`, to ensure the test runs against the correct artifacts.

Thoughts?

packages/amplify-ui-components/stencil.config.ts
```diff
@@ -1,12 +1,19 @@
 import { Config } from '@stencil/core';
 import { angularOutputTarget } from '@stencil/angular-output-target';
 import { reactOutputTarget } from '@stencil/react-output-target';
+import externals from 'rollup-plugin-node-externals';
```
jordanranz: I believe this is specifically for NodeJS built-ins to be external. I'm curious if you tested just using `external` highlighted in the [rollup docs here](https://rollupjs.org/guide/en/#peer-dependencies)?

packages/amplify-ui-components/stencil.config.ts
jordanranz: If we are looking to migrate into a separate repo we probably want to track this on our own

packages/amplify-ui-components/stencil.config.ts
ericclemmons: Since Stencil has it's own config (https://stenciljs.com/docs/config), we can't access Rollup's config directly.

PR #4781, opened 2020-01-27
feat(@aws-amplify/ui-components): User agent tracking for UI component packages

jordanranz:
_Description of changes:_
- Export function to append to the Amplify user agent
- Append to user agent before `amplify-authenticator` loads
- Append to user agent in framework bindings packages


packages/amplify-ui-components/__tests__/__snapshots__/bundle-size.spec.js.snap
```diff
@@ -143,34 +187,64 @@ Object {
...
   "esm-es5/index.mjs": "0 Bytes",
   "esm-es5/loader.mjs": "8 KB",
-  "esm/amplify-amazon-button_45.entry.js": "245 KB",
+  "esm/amplify-amazon-button_5.entry.js": "19 KB",
+  "esm/amplify-button.entry.js": "2 KB",
   "esm/amplify-checkbox.entry.js": "1 KB",
-  "esm/amplify-federated-sign-in.entry.js": "2 KB",
+  "esm/amplify-code-field_9.entry.js": "17 KB",
+  "esm/amplify-confirm-sign-in_5.entry.js": "14 KB",
+  "esm/amplify-examples.entry.js": "3 KB",
```
ericclemmons: I'm wondering why the test generated so many new entries 🤔

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
```diff
@@ -1,6 +1,6 @@
-import { Auth } from '@aws-amplify/auth';
 import { ConsoleLogger as Logger } from '@aws-amplify/core';
 import { Component, h, Prop } from '@stencil/core';
+import { Auth } from '@aws-amplify/auth';
```
ericclemmons: Was this moved based on preference, or because the userAgent behavior differed?

packages/amplify-ui-components/stencil.config.ts
```diff
@@ -24,6 +24,12 @@ export const config: Config = {
       '@aws-sdk/eventstream-marshaller': ['EventStreamMarshaller'],
     },
   },
+  bundles: [
+    { components: ['amplify-amazon-button', 'amplify-auth-fields', 'amplify-auth0-button', 'amplify-authenticator', 'amplify-sign-in'] },
+    { components: ['amplify-confirm-sign-in', 'amplify-confirm-sign-up', 'amplify-facebook-button', 'amplify-federated-buttons', 'amplify-federated-sign-in'] },
+    { components: ['amplify-google-button', 'amplify-forgot-password', 'amplify-select-mfa-type', 'amplify-sign-out', 'amplify-sign-up'] },
+    { components: ['amplify-totp-setup', 'amplify-verify-contact'] },
+  ],
```
ericclemmons: I assume this is what created the new entry points from above?

(I'm aware of https://stenciljs.com/docs/config#bundles, but was unsure how it manifested in practice.)

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
jordanranz: I thought I needed to move it because the userAgent behavior differed, but they theory ended up being invalid, ill change this back

packages/amplify-ui-components/__tests__/__snapshots__/bundle-size.spec.js.snap
jordanranz: Not sure, any idea?

packages/amplify-ui-components/stencil.config.ts
jordanranz: This was an artifact of me attempting to solve the bundling problem on this branch, removing.

packages/amplify-ui-angular/src/index.ts
```diff
@@ -1,3 +1,6 @@
+import { appendToAmplifyUserAgent } from '@aws-amplify/core';
+appendToAmplifyUserAgent('@aws-amplify/ui-angular');
```
ericclemmons: This looks like a registration pattern that will modify the `userAgent` based on what exists in the bundle vs. where a network call originated (which would differentiate direct calls to `API.*` by the user vs. calls our component makes).

Can you help me understand how this addresses these questions (copied from our meeting notes):

- We want to track adoption metrics for `@aws-amplify/ui-components`, but until `@aws-amplify/ui-components` has full coverage, users may keep the previous library (e.g. `aws-amplify-react`).  Does this support this scenario?

- We want to gather data on where calls originate (successful or not) to a service (e.g. Cognito).  Does this allow us to differentiate calls from our components vs. the user's?

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
ericclemmons: Let's see what happens when you revert the `bundles` change.  

If it's still weird, I'll dig in. 👀

We want this snapshot to not be noisy, but instead a "smell" that bundling may be affecting users. (More network requests, larger payloads, etc.) 👃

packages/amplify-ui-components/__tests__/__snapshots__/bundle-size.spec.js.snap
jordanranz: It must be because of the `bundles` change as mentioned below

packages/amplify-ui-angular/src/index.ts
jordanranz: If this file is executed in the user's browser, it will append `@aws-amplify/ui-angular` which is more like a registration pattern. To track adoption metrics we append an additional bit of content to the user agent when the `amplify-authenticator` component will load (see `amplify-authenticator.tsx`). This does _not_ track granular `API.*` usage but will at least give us basic adoption metrics. Let's chat offline about this some more.

ericclemmons: That may be good enough: registering additional userAgent bits when components are invoked correlates strongly with the API calls. 👍

Issue #4770, opened 2020-01-24
undefined is not an object (evaluating 'o.accessToken')


HMilbradt:
**Describe the bug**
This bug occurs when a "Pre Sign Up" lambda trigger returns an error when using `Auth.federateSignIn({ provider: social_provider })`.  Just before it is thrown, another error is thrown: `OAuth - Error handling auth response. Error: PreSignUp+failed+with+error-Error-message-here` internally.

**To Reproduce**
Steps to reproduce the behaviour:

1. Add a Pre sign up trigger that returns an error
2. Attempt to sign in with a Social provider
3. See error

**Expected behaviour**
The error message from the pre sign up trigger should be emitted on the event bus, while the other error should not be thrown.

elorzafe:
@HMilbradt I reproduced the problem you mention, we will be working on fix. Thanks for your feedback.

PR #4794, opened 2020-01-29
fix unhandled urlListener for OAuth with `error_description`

elorzafe:
_Issue #, if available:_
Fixes: #4770 
_Description of changes:_

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -225,12 +225,8 @@ export default class AuthClass {
 			});
 
 			// **NOTE** - Remove this in a future major release as it is a breaking change
-			urlListener(async ({ url }) => {
-				try {
-					await this._handleAuthResponse(url);
-				} catch (err) {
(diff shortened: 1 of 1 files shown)


packages/auth/src/Auth.ts
```diff
@@ -225,8 +225,12 @@ export default class AuthClass {
 			});
 
 			// **NOTE** - Remove this in a future major release as it is a breaking change
-			urlListener(({ url }) => {
-				this._handleAuthResponse(url);
+			urlListener(async ({ url }) => {
+				try {
+					await this._handleAuthResponse(url);
+				} catch (err) {
+					logger.debug('Error found while handling url response', err);
```
manueliglesias: Can we give customers a way to handle/get notified about this error?  (e.g. with `Hub`)?

elorzafe: Updated the Pr to not throw the error instead

Issue #4723, opened 2020-01-16
browserOrNode() breaks pages with strict Content-Security-Policy headers


MHarutunian:
**Describe the bug**
The `JS.browserOrNode()` function uses eval-like expressions that prevent amplify-js from running when a strict Content-Security-Policy is in place. The eval-like expressions I'm talking about are `new Function("...")`, which are not allowed in a Content-Security-Policy, that doesn't allow `eval()`.

This leads to the following error message in the browser (in this case Firefox):
> Content Security Policy: The page’s settings blocked the loading of a resource at eval (“default-src”). vendor.js

This prevents us from using any of the functionality provided by amplify-js. I believe the issue was introduced with commit 52bd2877bee66b934885e6726b68c8a926c28c3d, which changed the functionality in `browserOrNode()` to use `new Function()`.

**To Reproduce**

1. Have a page with a Content-Security-Policy that defines `default-src 'self';` (`script-src 'self';` should also work), see `Sample Code` for the one we are using.
2. Include any `amplify-js` package that uses `JS.browserOrNode()` (in our case `@aws-amplify/auth`).
3. See error in the browser's console (and possibly on the page), the package will not be working as expected

**Expected behavior**
`amplify-js` (or rather `browserOrNode()`) should not break any Content-Security-Policy that forbids `eval()`

**Screenshots**
Error in Firefox console:
![image](https://user-images.githubusercontent.com/481294/72535055-2c60e500-3878-11ea-83fb-43b5600cbdb8.png)

Error in Google Chrome console:
![image](https://user-images.githubusercontent.com/481294/72535173-69c57280-3878-11ea-968d-2c6b5f83a354.png)

<details>
  <summary><strong>Environment</strong></summary>

```
  System:
    OS: Linux 5.3 Ubuntu 19.10 (Eoan Ermine)
    CPU: (4) x64 Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
    Memory: 141.61 MB / 7.64 GB
    Container: Yes
    Shell: 5.0.3 - /bin/bash
  Binaries:
    Node: 12.14.1 - /usr/bin/node
    Yarn: 1.21.1 - /usr/bin/yarn
    npm: 6.13.4 - /usr/bin/npm
  Browsers:
    Chrome: 79.0.3945.117
    Firefox: 72.0.1
  npmPackages:
    @aws-amplify/auth: ^2.1.3 => 2.1.3 
[...]

Issue #4708, opened 2020-01-11
undefined is not an object (evaluating 'window.location.href')


dimonchik-com:
**Describe the bug**
I updated react-native from 0.61.4 to 0.61.5 and I got warning 

```
WARN  Possible Unhandled Promise Rejection (id: 0):
TypeError: undefined is not an object (evaluating 'window.location.href')
http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:167369:89
step@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:165674:25
```

It's warning arise when i run next code

```
import Amplify, { Auth } from 'aws-amplify';
import API, {graphqlOperation} from "@aws-amplify/api";
import awsconfig from './../aws-exports';
Amplify.configure(awsconfig); // when i disable this row, warning disappear
```

dimonchik-com:
I don't know where the problem is and at the moment just changed this code
```
	static browserOrNode() {
		// function to check if the global scope is "window"
		const isBrowser =  false;

		// function to test if global scope is binded to "global"
		const isNode = new Function(
			'try {return this===global;}catch(e){return false;}'
		);

		return {
			isBrowser: isBrowser(),
			isNode: isNode(),
		};
	}
```

manueliglesias:
Hi @dimonchik-com 

Did changing the code help with the warning?

if so, the original code was 

```js
		// function to check if the global scope is "window"
		const isBrowser = new Function(
			'try {return this===window;}catch(e){ return false;}'
		);
```

Could it be possible you have another import that could be defining the global `window` object? (e.g. some polyfill or similar)

sasitha:
Having the exact same issue. What would be the cause for this?.

IronLuffy55:
@manueliglesias 
I ran into the same warning on a bare bones (i.e. react-native init) project running v0.59.10.

It looks like react-native wants to setup some global variables on start up. 

See https://stackoverflow.com/questions/49911424/what-does-the-variable-window-represent-in-react-native

List of globals provided by react-native: https://github.com/facebook/react-native/blob/master/Libraries/Core/setUpGlobals.js

The current implementation will always return Browser = true if thats the case.

I can confirm that dimonchik-com  snippet removed the warning on my project.

Hope some of this helps

dimonchik-com:
> @manueliglesias
> I ran into the same warning on a bare bones (i.e. react-native init) project running v0.59.10.
> 
[...]

PR #4803, opened 2020-01-30
fix(@aws-amplify/core): Revert environment checks

manueliglesias:
This fix reverts our environment checks to a previous version that worked fine, but failed under certain curcustances now fixed (see #4678)

The solution using `new Function('return this')()` caused problems when using CSP

_Issue #, if available:_
Fixes: #4723 

_Description of changes:_

### DIFF
--- a/packages/core/__tests__/JS-node-runtime-test.ts
+++ b/packages/core/__tests__/JS-node-runtime-test.ts
@@ -6,7 +6,7 @@
  * Since this is allowed per test file and not per test or describe, we have
  * two tests, one for node and other for browser
  */
-import * as core from '../dist/aws-amplify-core.js';
+import * as core from '../lib';
 
 describe('JS build test', () => {
 	test('when its node ', () => {
--- a/packages/core/src/JS.ts
+++ b/packages/core/src/JS.ts
@@ -165,26 +165,17 @@ export default class JS {
 		return result;
 	}
 
-	/**
-	 * Webpack adds node shim to the bundle overriding the `process` variable
-	 * causing issues detecting nodejs environment. Creating a new function will
-	 * help to avoid incorrect environment detection as new function will have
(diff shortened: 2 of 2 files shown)


ericclemmons:
@manueliglesias Can you clarify how the original PR #4611's reasoning is addressed in this PR?
> Webpack adds node shim to the bundle overriding the process variable
causing issues detecting nodejs environment. Creating a new function will
help to avoid incorrect environment detection as new function will have
it's this binded to global scope. Credit: stackoverflow.com/a/31090240

Was it the switch to `./lib`, or is there more?

manueliglesias:
> @manueliglesias Can you clarify how the original PR #4611's reasoning is addressed in this PR?
> 
> > Webpack adds node shim to the bundle overriding the process variable
> > causing issues detecting nodejs environment. Creating a new function will
> > help to avoid incorrect environment detection as new function will have
> > it's this binded to global scope. Credit: stackoverflow.com/a/31090240
> 
> Was it the switch to `./lib`, or is there more?

Sure @ericclemmons,

Previously, our environment checks were failing when the UMD files (from the `dist` folders) were used in Node environments. This is because webpack defines a `process.browser` field which returns true along some other values that messed up env detection in node.

#4611 tried to fix this by making use of the `new Function('return this')();` trick to detect globals. While this works fine in *all* environments, it triggers errors when used with Content Security Policy (unless you use `unsafe-eval`)

#4678 changed the entry points for Node or SSR envs to use CJS instead of webpack's UMD bundles.

By reverting the environment check, and having the right entry point defined for node, we no longer need the unsafe *eval-like* way of detecting the global scope.

Interesting reads:
- https://github.com/tc39/proposal-global
- https://developers.google.com/web/fundamentals/security/csp

ericclemmons:
Customer confirmed this PR fixes their issue as well:
> https://github.com/aws-amplify/amplify-js/issues/4708#issuecomment-582882616

PR #4804, opened 2020-01-30
feat(@aws-amplify/ui-components): User agent tracking for UI component packages

jordanranz:
_Description of changes:_
- Export function to append to Cognito user agent
- Append to Cognito user agent inside of Auth and UI (amplify-authenticator)


packages/amplify-ui-components/__tests__/__snapshots__/bundle-size.spec.js.snap
```diff
@@ -25,12 +25,12 @@ Object {
   "amplify-ui-components/p-c0fae6c3.js": "2 KB",
   "amplify-ui-components/p-d1csj4vs.entry.js": "1 KB",
   "amplify-ui-components/p-d447982c.js": "12 KB",
+  "amplify-ui-components/p-ddpnluaq.entry.js": "115 KB",
   "amplify-ui-components/p-ea249ef6.system.js": "21 KB",
-  "amplify-ui-components/p-emwljr8n.entry.js": "115 KB",
```
ericclemmons: I wish there were a way to see these at the same, but unfortunately the hashes are necessary to differentiate the different `p-` files :(

(Notice how some _don't_ change, but their contents are the same)

packages/auth/src/Auth.ts
```diff
@@ -416,6 +420,8 @@ export class AuthClass {
 		if (!this.userPool) {
 			return this.rejectNoUserPool();
 		}
+		appendToCognitoUserAgent(Platform.userAgent);
```
Amplifiyer: Shouldn't we do this in configure rather than executing this logic for every signin call?

packages/amazon-cognito-identity-js/src/UserAgent.js
```diff
@@ -3,4 +3,20 @@ export default UserAgent;
 // constructor
 function UserAgent() {}
 // public
-UserAgent.prototype.userAgent = 'aws-amplify/0.1.x js';
+UserAgent.prototype.userAgent = 'amazon-cognito-identity-js';
+
+export const appendToCognitoUserAgent = content => {
+	if (content || !UserAgent.prototype.userAgent.includes(content)) {
```
Amplifiyer: Shouldn't this condition be &&? What if content is null or worse undefined. We should sanitize the incoming args.
We should also have unit tests for this.

packages/amplify-ui-components/__tests__/__snapshots__/bundle-size.spec.js.snap
jordanranz: I think this is due to this [Config value](https://stenciljs.com/docs/config#hashfilenames) in Stencil.  It may be possible to not have these change. Out of scope of this PR but definitely something to check out.

packages/auth/src/Auth.ts
jordanranz: This should go in `configure`, good callout.

packages/amazon-cognito-identity-js/src/UserAgent.js
jordanranz: This should be `&&` I will make this change. I will be adding unit tests for this today :D

jordanranz:
@ericclemmons basically the Platform.userAgent is appended to the Cognito request (via UserAgent) during `Auth.configure`.

I will definitely add unit tests. Thank you for the feedback.

jordanranz:
I have pushed unit tests for this change.

ericclemmons:
Not sure why the build is failing though :(

jordanranz:
@ericclemmons `error Command failed with exit code 137` means that the container is running out of memory. This currently happens in personal circleCI builds as we have limited memory constraints. When this is built on the `aws-amplify` repo it should succeed. That being said we should investigate this a little further and fix the personal CircleCI builds as a follow up.

PR #4842, opened 2020-02-04
feat(@aws-amplify/ui-components): Inject css variables for root theme

jordanranz:
_Description of changes:_
- Inject css variables for root theme when component lib loads


packages/amplify-ui-components/src/global/theme.ts
```diff
@@ -0,0 +1,90 @@
...
+injectGlobal`
+  :root {
+    /* Colors */
+    --amazon-orange: #ff9900;
+    --light-amazon-orange: #ffac31;
+    --dark-amazon-orange: #e88b01;
+    --squid-ink: #232f3e;
+    --light-squid-ink: #31465f;
+    --deep-squid-ink: #152939;
+    --grey: #828282;
```
ericclemmons: There are several collisions with injecting in the root I've come across.  `button-color`, `white`, `red`, `grey`, etc. and _especially_ `font-family` are all common and will impact customers.



1. Prefix all CSS variables with `amplify-`, `amz-`, or something similar.
1. Rather than relying on `:root`, add these to a shared wrapper component's `style` property:

	```jsx
	<AmplifyRoot style={{ '--white': '#ffffff' }}>
	```
1. Do both! 🎉

ashika01: I vote prefix I think that most common pattern for UI lib.

P.S: I was just working on this comment. Eric beat me to it. 😝

ericclemmons: What's interesting about the `style` suggestion is _this is what users can use to override theme colors for free_.  (Bad news, it ends up being a contract 😬)

jordanranz: I am definitely on board with the prefix, I actually remembered this last night before falling asleep haha. Let's leave the other options up for discussion and do a theme design discussion.

packages/amplify-ui-components/src/global/variables.css
```diff
@@ -1,86 +0,0 @@
...
-  --input-background-color: var(--white);
-  --input-border-color: var(--light-grey);
-  --primary-font-size: 1em;
-  --secondary-font-size: 0.875em;
-  --third-font-size: 0.813em;
-
-  --font-family: 'Amazon Ember', 'Helvetica Neue Light', 'Helvetica Neue', 'Helvetica', 'Arial', 'sans-serif';
-  --body-background: #f8f4f4;
-
-  /** Angular Theme **/
```
ashika01: So whats happening to this Angular theming?

jordanranz: These were artifacts from our CSS Modules in `amplify-ui`

packages/amplify-ui-components/stencil.config.ts
```diff
@@ -41,5 +41,5 @@ export const config: Config = {
       serviceWorker: null, // disable service workers
     },
   ],
-  globalStyle: 'src/global/variables.css',
+  globalScript: 'src/global/index.ts'
```
sammartinez: So this changed :D does this work properly with production builds ?

packages/amplify-ui-components/src/global/theme.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
+ * the License. A copy of the License is located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+import { injectGlobal } from 'emotion'
```
sammartinez: NIT / Non-Blocking: can you add “;”

sammartinez: Or do we have prettier setup on this branch ?

Issue #4847, opened 2020-02-05
S3Image component aws-amplify-react not able to distinguish a successful upload 


dudenamedjune:
**Describe the bug**
I'm not sure this is a bug or if this is the expected behavior. If it is the latter I can open a feature request. 

**To Reproduce**
Refresh the page and you can see that onLoad has been called due to the image being loaded. 

1. Consume <S3Image />
2. pass the prop onLoad 
3. notice on component mount onLoad is called

**Expected behavior**
I would expect the behavior to differentiate between a successful load of S3Images vs the transparent initial src loaded in the image tag.

jordanranz:
Closing this issue as the PR for this was merged :) Please open a new issue if there are any related questions or concerns.

PR #4848, opened 2020-02-05
fix(aws-amplify-react): added prop onS3ImageSuccess to the S3Image component to have a clear separation of concerns

dudenamedjune:
fix(aws-amplify-react): added prop onS3ImageSuccess to the S3Image component to have a clear separation of concerns

_Issue #, if available:_
https://github.com/aws-amplify/amplify-js/issues/4847
_Description of changes:_


dudenamedjune:
Seems like there were some outdated snapshots in the auth component

packages/aws-amplify-react/src/Storage/S3Image.tsx
```diff
@@ -36,6 +36,7 @@ export interface IS3ImageProps {
 	onClick?: any;
 	onError?: any;
 	onLoad?: any;
+	onS3ImageSuccess?: any;
```
jordanranz: The `S3Image` component does both uploading and retrieval from S3. Can this variable be renamed to better reflect where the callback is executing. Something like `onUploadSuccess` or `onS3PutSuccess`. Also is there a better type we can apply to the function?

jordanranz:
Thanks for contributing @dudenamedjune! Had a small comment above but otherwise looks good to me

dudenamedjune:
@jordanranz Anytime! I changed the name, but as far as typing goes the other props in the interface have any for function only props. I'm not aware of a function type, but you can do something like this 

```js
export interface IS3ImageProps {
	onUploadSuccess?: () => {};
```

PR #4858, opened 2020-02-06
feat(@aws-amplify/ui-components): amplify-toast component

ashika01:
_Description of changes:_
 - amplify-toast component
 - This PR introduces Hub from the core. The Hub Dispatch for throwing the error and the Hub Listen catches them from authenticator component

_Validation steps:_

- From amplify-js, cd packages/amplify-ui-components
- Run yarn then yarn start
- Try login in without filling the form fields / without proper credientials


packages/amplify-ui-components/src/components/amplify-sign-in/amplify-sign-in.tsx
```diff
@@ -145,6 +146,8 @@ export class AmplifySignIn {
         this.checkContact(user);
       }
     } catch (error) {
+      console.log('error', error);
```
jordanranz: Can we replace this `console.log` with a `logger.error`

packages/amplify-ui-components/src/common/constants.ts
```diff
@@ -133,3 +133,7 @@ export const VERIFY_CONTACT_PHONE_LABEL = 'Phone Number';
 // Common events
 export const SUCCESS = 'SUCCESS';
 export const FAILURE = 'FAILURE';
+
+//Hub Events and Channels
```
jordanranz: NIT: Can you add a space after `//`

packages/amplify-ui-components/src/common/constants.ts
```diff
@@ -133,3 +133,7 @@ export const VERIFY_CONTACT_PHONE_LABEL = 'Phone Number';
...
+export const TOAST_AUTH_ERROR_CHANNEL = 'UI Comp - AuthError';
```
jordanranz: Let's stay consistent with the `Hub` documentation and use title case for the channel. Since we will also be moving to listening to `Hub` for `authState` changes we can probably share a channel here and use diverse events. We can chat offline about specific naming here.

packages/amplify-ui-components/src/components/amplify-toast/amplify-toast.tsx
```diff
@@ -19,6 +20,7 @@ export class AmplifyToast {
     return (
       <div class={toast}>
         <amplify-icon class={toastIcon} name="warning" />
+        {this.message !== '' && <span>{this.message}</span>}
```
jordanranz: It seems the `message` prop above is undefined by default but checking for empty string here. For now let's make this a check for undefined `{this.message && <span>{this.message}</span>}`.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -97,9 +113,23 @@ export class AmplifyAuthenticator {
...
+    Hub.remove(UI_AUTH_CHANNEL, () => {
+      logger.info('Toast auth listener removed');
+    });
+  }
+
   render() {
     return (
       <Host>
+        {this.toastMessage && this.toastMessage !== '' ? (
+          <amplify-toast
```
ericclemmons: If the feedback above is integrated, then this will end up looking like:

```js
{this.payload && <amplify toast onClock={this.handleToastClose} payload={this.payload} />}
```

One consideration is _what would we pass to customers if they customize the toast component_?

packages/amplify-ui-components/src/components/amplify-confirm-sign-in/amplify-confirm-sign-in.tsx
```diff
@@ -110,6 +111,7 @@ export class AmplifyConfirmSignIn {
       await Auth.confirmSignIn(this.user, this.code, mfaType);
       this.checkContact(this.user);
     } catch (error) {
+      dispatchToastHubEvent(error);
       logger.error(error);
```
ericclemmons: Do you know what situations would have `logger.error(error)` vs. not?

(It's not consistent, and I'm personally not sure why)

packages/amplify-ui-components/src/common/helpers.ts
```diff
@@ -1,6 +1,21 @@
...
-export const styleNuker = (override: boolean, cn: string, ecn: string): string => override ? cn : cx(ecn, cn);
+export const styleNuker = (override: boolean, cn: string, ecn: string): string => (override ? cn : cx(ecn, cn));
 
-export const styleBranch = (branch: boolean, baseClass: string, branchClassA: string, branchClassB: string): string => branch ? cx(baseClass, branchClassA) : cx(baseClass, branchClassB);
\ No newline at end of file
+export const styleBranch = (branch: boolean, baseClass: string, branchClassA: string, branchClassB: string): string =>
+  branch ? cx(baseClass, branchClassA) : cx(baseClass, branchClassB);
+
+export const dispatchToastHubEvent = (error: ToastError) => {
+  Hub.dispatch(UI_AUTH_CHANNEL, {
```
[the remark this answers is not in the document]
ashika01: Do you see any benefits to having error passed down in the payload for toast vs just message. My reasoning was to keep Toast component dumb. All toast needs is a message to be put in its component right?

packages/amplify-ui-components/src/common/constants.ts
```diff
@@ -133,3 +133,7 @@ export const VERIFY_CONTACT_PHONE_LABEL = 'Phone Number';
 // Common events
 export const SUCCESS = 'SUCCESS';
 export const FAILURE = 'FAILURE';
+
+// Hub Events and Channels
+export const UI_AUTH_CHANNEL = 'UI Auth';
+export const TOAST_AUTH_ERROR_EVENT = 'ToastAuthError';
```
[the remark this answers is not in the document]
ashika01: I think it's about asking if we need toast to display all auth errors in a project. I would say the distinction there is not well defined to go for a generic naming.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -97,9 +113,23 @@ export class AmplifyAuthenticator {
     }
   }
 
+  async componentDidUnload() {
+    Hub.remove(UI_AUTH_CHANNEL, () => {
```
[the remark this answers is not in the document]
ashika01: Got it.. I just looked into the implementation too, its filtering of callback if its a string channel name. I will make the change. Our docs isn't obvious of that. Thanks

packages/amplify-ui-components/src/components/amplify-confirm-sign-in/amplify-confirm-sign-in.tsx
ashika01: It should be the equivalent as browser console is my understanding.

packages/amplify-ui-components/src/common/helpers.ts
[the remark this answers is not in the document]
ericclemmons: That's a good point.  I was thinking non-error states (e.g. "Success!  Signed in as Ashika"), but those would need to be different events entirely (not errors).

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
ericclemmons: Ashika mentioned keeping these components dumb and `message`-only. 👍

packages/amplify-ui-components/src/components/amplify-confirm-sign-in/amplify-confirm-sign-in.tsx
ericclemmons: Oh, I mean do you know why some `catch`es have `logger.error` while others do not?

If they should be consistent, then we can have `dispatchToastHubEvent` also run `logger.error` for all events to improve visibility.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
ericclemmons: In this case, a falsey check for `this.toastMessage` would be sufficient, correct?

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -17,10 +24,19 @@ export class AmplifyAuthenticator {
   @State() authState: AuthState = AuthState.Loading;
 
   @State() authData: CognitoUserInterface;
+  @State() toastMessage: string = '';
```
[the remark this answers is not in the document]
ericclemmons: @ashika01 wanted to keep the toast component dumb, so tracking message alone is fine. 👍

Is there a compelling reason now to default to `null` or `undefined` and use a falsey check below?

packages/amplify-ui-components/src/common/constants.ts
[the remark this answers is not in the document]
sammartinez: I believe and correct me if Im wrong @jordanranz, but our toast is only currently for Auth forms. So we may want to tailor as such.

packages/amplify-ui-components/src/common/helpers.ts
```diff
@@ -1,6 +1,21 @@
 import { cx } from 'emotion';
+import { Hub } from '@aws-amplify/core';
+import { UI_AUTH_CHANNEL, TOAST_AUTH_ERROR_EVENT } from './constants';
 
+interface ToastError {
```
sammartinez: Could this live under [auth-types](https://github.com/aws-amplify/amplify-js/blob/ui-components/master/packages/amplify-ui-components/src/common/types/auth-types.ts)?

packages/amplify-ui-components/src/components/amplify-toast/amplify-toast.tsx
```diff
@@ -19,6 +20,7 @@ export class AmplifyToast {
     return (
       <div class={toast}>
         <amplify-icon class={toastIcon} name="warning" />
```
sammartinez: Future thinking is all, I believe we need something to at least update the 'name' field based on a success, failure, etc.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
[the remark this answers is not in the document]
ashika01: I went for this approach because, if the message passed to a toast component is an empty string then it’s gonna render the toast with no text on it. So I am making a check for that too. 

What do you think? Could we handle that better? On top of my head we have had a util to check these, but it might be overkill here

packages/amplify-ui-components/src/components/amplify-confirm-sign-in/amplify-confirm-sign-in.tsx
ashika01: Ha ha! Good question. Let me talk it through offline and I would like to keep it consistent too. But might not do it under dispatch tho.

packages/amplify-ui-components/src/components/amplify-toast/amplify-toast.tsx
ashika01: Yeah we want to do that. But we decided on not doing this but right now. I can add it in if needed.

packages/amplify-ui-components/src/common/helpers.ts
ashika01: I wouldn’t put the types in a separate file if it doesn’t need to be exported to other files / if it’s just specific to one file. That’s the best practice too. We do so in DataStore too. but if the norm in UI is different I can take put it in auth types. 

The reasoning for keeping it in the same file for these specific types is the readability , you don’t want to navigate thru bunch of files to just understand a type that exists only for this file.

packages/amplify-ui-components/src/common/constants.ts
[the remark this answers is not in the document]
ashika01: Could be brief me on tailoring? What are your thoughts there?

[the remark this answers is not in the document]
ashika01: Also, call out, toast component might be used only for auth by us. But we might wanna keep toast simple and generic for anyone to use it for other scenarios

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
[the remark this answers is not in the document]
ericclemmons: Your `{this.toastMessage ? ... : ...}` turnary would work fine with an empty `""` string:

```
"" ? console.log("truthy") : console.log("falsey")
// falsey
```

Was there another scenario I'm missing?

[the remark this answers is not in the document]
ashika01: Yup you are right.. I did something else first. Incorporating this in.

jordanranz:
Just leaving a note to add `data-test` attributes and run Cypress tests with this change.

PR #4865, opened 2020-02-07
Modularization aws sdk v3 alpha client

Amplifiyer:
_Issue #, if available:_

_Description of changes:_

* Update @aws-sdk clients with alpha versions
* Update imports of aws-sdk to use barrel files instead of specific commands and clients
* Update the way amplify user agent gets appended
* Update Firehose provider to use new aws-sdk clients
* Update sideeffects to include barrel files to avoid tree shaking for them in angular builds
* Angular: Add default providers for amplifyService
* Removing cognito client hack
* Misc changes based on new aws-sdk client updates


packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -166,6 +169,7 @@ export class AWSS3Provider implements StorageProvider {
 		try {
 			const signer = new S3RequestPresigner({ ...s3.config });
 			const request = await createRequest(s3, new GetObjectCommand(params));
+			console.log('from createRequest', request);
```
jordanranz: Remove this `console.log`

packages/analytics/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
jordanranz: I don't think this file is needed

packages/storage/src/providers/AWSS3Provider.ts
Amplifiyer: Removed

packages/analytics/LICENSE
Amplifiyer: Removed

packages/storage/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
jordanranz: This license file might not be needed.

packages/core/src/Credentials.ts
```diff
@@ -220,11 +221,34 @@ export class CredentialsClass {
...
 			credentials = fromCognitoIdentity(cognitoIdentityParams)();
 		} else {
-			const cognitoIdentityParams: FromCognitoIdentityPoolParameters = {
-				identityPoolId,
-				client: cognitoClient,
+			/*
+			Retreiving identityId with GetIdCommand to mimic the behavior in the following code in aws-sdk-v3:
+			https://git.io/JeDxU
+
+			Note: Retreive identityId from CredentialsProvider once aws-sdk-js v3 supports this.
```
jordanranz: Curious if you spoke to the sdk team about this ask?

packages/aws-amplify/src/index.ts
```diff
@@ -44,9 +43,10 @@ export {
 } from '@aws-amplify/core';
 
 import { Amplify, ServiceWorker } from '@aws-amplify/core';
-
+import { Auth } from '@aws-amplify/auth';
 import Cache from '@aws-amplify/cache';
 
+Amplify.Auth = Auth;
```
jordanranz: Why is `Auth` the only category here?

packages/aws-amplify-angular/src/providers/amplify.service.ts
```diff
@@ -20,6 +20,16 @@ import Amplify, { Logger, I18n } from '@aws-amplify/core';
 import { AuthState } from './auth.state';
 import { authDecorator } from './auth.decorator';
 
+import '@aws-amplify/analytics';
```
jordanranz: This can be removed.

packages/analytics/src/Providers/AmazonPersonalizeHelper/DataType.ts
```diff
@@ -49,7 +49,7 @@ export interface RecordEventPayload {
 	eventId: string;
 	eventType: string;
 	sentAt: number;
-	properties?: Properties;
+	properties?: string;
```
jordanranz: Why was `Properties` changed to `string` here?

packages/aws-amplify-angular/src/providers/amplify.service.ts
Amplifiyer: Removing.

packages/aws-amplify/src/index.ts
Amplifiyer: This is because for cases when unauthorized access is available for categories, clients don't have to import Auth, however that means webpack tree shakes out Auth category and this https://github.com/aws-amplify/amplify-js/blob/a047ce73abe98c3bf82e888c3afb4d2f911805f3/packages/core/src/Credentials.ts#L94-L95 would fail and hence making Auth category available.

Have to test when modular imports happen in the client app, how will Auth be available?

packages/storage/LICENSE
Amplifiyer: Removing

packages/analytics/src/Providers/AmazonPersonalizeHelper/DataType.ts
Amplifiyer: Talked offline, this change is because of aws-sdk new clients. V2 client was doing the serialization for the clients but V3 is not so we are doing it ourselves and passing a stringified object to them

jordanranz: Spoke offline about this.

packages/aws-amplify/src/index.ts
Amplifiyer: As per the current docs, users importing categories individually have to import Auth category and configure it explicitly. Trying to avoid this is going to be a big refactor since Core/Credentials.ts uses Auth but cannot depend on it to avoid cyclic dependency. To avoid users importing Auth category credentials.ts have to depend on and import auth which would be only possible if Credentials has been moved out of core.

PR #4872, opened 2020-02-09
feat(@aws-amplify/ui-components): Use Shadow DOM, use Sass css, authenticator slots

jordanranz:
_Description of changes:_
- Use Shadow DOM for components
- Use Sass for styling instead of Emotion
- Slots for `amplify-authenticator` sub components


packages/amplify-ui-components/src/global/theme.css
```diff
@@ -0,0 +1,51 @@
...
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
+ * the License. A copy of the License is located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+:root {
```
jordanranz: We could still inject these using `emotion` and read the brand variable from json.

packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx
```diff
@@ -50,7 +51,9 @@ export class AmplifyConfirmSignUp {
   /** Passed from the Authenticator component in order to change Authentication states
    * e.g. SignIn -> 'Create Account' link -> SignUp
    */
-  @Prop() handleAuthStateChange: AuthStateHandler;
+  @Prop() handleAuthStateChange: AuthStateHandler = (nextAuthState: AuthState, data?: object) => {
+    Hub.dispatch('AuthenticatorState', { event: nextAuthState, data });
```
jordanranz: This should probably move to a utility function.

packages/amplify-ui-components/src/components/amplify-checkbox/amplify-checkbox.tsx
```diff
@@ -31,7 +27,7 @@ export class AmplifyCheckbox {
 
   render() {
     return (
-      <span class={styleNuker(this.overrideStyle, STATIC_CHECKBOX_CLASS_NAME, checkbox)}>
```
sammartinez: Goodbye StyleNuker 😢

packages/amplify-ui-components/src/common/helpers.ts
```diff
@@ -1,6 +1,7 @@
 import { cx } from 'emotion';
```
sammartinez: Do we need this anymore? I saw that `styleNuker` was being removed so I believe we should remove anything related to emotion since we are no longer going to be using this. Im also curious, does Sass work with React Native? Cause that was the reason we went to use emotion in the first place was for that

jordanranz: Yea, just got rid of this line. I'll look to get rid of the dependency if we decide to not use emotion to inject css variables.

After more investigation, emotion does 'support' React Native but we wouldn't be able to directly share styles because of the differences between css and styling in React Native. What we can share is a common JSON object for brand which would be consumed in web as css variables and in React Native is a JS object. We may still be able to use emotion for this as mentioned in this comment above.

packages/amplify-ui-components/src/components/amplify-checkbox/amplify-checkbox.tsx
jordanranz: RIP

packages/amplify-ui-components/src/common/helpers.ts
sammartinez: Sounds good to me and that would make sense because I believe that can apply to other platforms as well in the future

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -34,14 +34,14 @@ export class AmplifyTOTPSetup {
...
-  @Prop() handleAuthStateChange: AuthStateHandler;
+  @Prop() handleAuthStateChange: AuthStateHandler = dispatchAuthStateChangeEvent;
 
   @State() code: string | null = null;
   @State() setupMessage: string | null = null;
   @State() qrCodeImageSource: string;
   @State() qrCodeInput: string | null = null;
 
-  componentWillLoad() {
+  componentWillRender() {
```
ashika01: What is the reason having the setup before every render vs just on load?

jordanranz: This was an artifact for testing pre-loading components. I'll change it back and investigate in another branch.

packages/amplify-ui-components/src/common/helpers.ts
```diff
@@ -1,21 +1,28 @@
...
+};
 
 export const dispatchToastHubEvent = (error: ToastError) => {
   Hub.dispatch(UI_AUTH_CHANNEL, {
     event: TOAST_AUTH_ERROR_EVENT,
     message: error.message,
   });
 };
+
+export const dispatchAuthStateChangeEvent = (nextAuthState: AuthState, data?: object) => {
```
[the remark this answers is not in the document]
jordanranz: Good idea.

jordanranz:
@ericclemmons I made a change to fix that (see the `tsconfig` change), you may just need to clear `node_modules`.

PR #4883, opened 2020-02-11
fix(#3620): Error when building with AOT enabled when using Angular

jmhummel:
Issue #, if available:
fixes(#3620)

Description of changes:
This will allow users to use the strict template check option for Angular AOT compilation.


isaacdre:
My ts compiler called... it wants it's fullTemplateTypeCheck to flip back to true.

I tried to explain that patience is a virtue, but it replied, "Patience may be a virtue, but forgoing type checking before prod pushing is a moral sin".

Having this patched would go a long way towards fixing my relationship with my compiler. It doesn't like turning a blind eye to my code.

kernowjoe:
I second this, can this be approved + merged?

PR #4895, opened 2020-02-12
chore(@aws-amplify/ui-components): CommonJS named exports from sdk-v3

jordanranz:
_Description of changes:_
- Clean up new commonjs named exports from aws-sdk-v3


jordanranz:
Yea with the merge from modularization we received a bunch of new aws-sdk-v3 modules under the hood that require commonjs named exports for rollup to load. I think we could get rid of these if we work with the sdk team to clean up how they export modules in these files.

PR #4903, opened 2020-02-12
Allow passing custom user attributes during sign up

artemkv:
_Description of changes:_

I want to be able to pass custom user attributes (set programmatically, not visible to user) to Auth.signUp function.
As an example, I would like to pass user locale as a value of an attribute, so later I could use it in the lambda to localize email messages in a language of the user.

```
    const signUpConfig = {
        customeAttributes: [
            {
                name: "user_locale",
                value: "ru"
            }
        ]
    };

    return (
        <Authenticator signUpConfig={signUpConfig} />
    );
```

If there is a better way to achieve this, please let me know.

### DIFF
--- a/packages/aws-amplify-react/__tests__/Auth/SignUp-test.tsx
+++ b/packages/aws-amplify-react/__tests__/Auth/SignUp-test.tsx
@@ -870,3 +870,301 @@ describe('signUp with signUpConfig', () => {
 		expect(spyon).toBeCalled();
 	});
 });
+
+describe('signUp with customAttributes', () => {
+	let wrapper;
+	beforeEach(() => {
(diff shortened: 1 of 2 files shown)


amhinson:
Thanks for opening a PR @artemkv 🙏  

However, it does look like our new UI Components (`@aws-amplify/ui-react`) [supports this use-case](https://docs.amplify.aws/ui/auth/authenticator/q/framework/react#hiding-form-fieldshttps://docs.amplify.aws/ui/auth/authenticator/q/framework/react#hiding-form-fields) you are looking for here. Here is a [migration guide](https://docs.amplify.aws/ui/auth/authenticator/q/framework/react#migration) for upgrading to the new library, if you need it. Thanks!

PR #4929, opened 2020-02-17
feat(@aws-amplify/ui-components): amplify-loading-spinner, loading state, minor form input tweaks

jordanranz:
_Description of changes:_
- amplify-loading-spinner component
- authenticator form loading state
- minor form input tweaks for cypress tests

Kapture 2020-02-17 at 18 26 02


packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.tsx
```diff
@@ -43,14 +43,17 @@ export class AmplifyFormSection {
...
 
   @Prop() primaryFooterContent: string | FunctionalComponent = (
     <amplify-button
       type="submit"
       overrideStyle={this.overrideStyle}
       data-test={this.testDataPrefix + '-' + this.testDataPrefix + '-button'}
     >
-      {this.submitButtonText}
+      <amplify-loading-spinner style={{ display: this.loading ? 'initial' : 'none' }} />
+      <span style={{ display: this.loading ? 'none' : 'initial' }}>{this.submitButtonText}</span>
```
ericclemmons: I see the following pattern more often:

```jsx
{this.loading ? <amplify-loading-spinner /> : this.submitButtonText}
```

Would this work here?

packages/amplify-ui-components/src/components/amplify-icon/amplify-icon.tsx
```diff
@@ -4,7 +4,7 @@ import { icons, IconNameType } from './icons';
 @Component({
   tag: 'amplify-icon',
   styleUrl: 'amplify-icon.scss',
-  shadow: true,
+  scoped: true,
```
ericclemmons: I'd be interested in learning what happened here to warrant `scoped` vs. `shadow?

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -34,7 +34,7 @@ export class AmplifyInput {
         id={this.fieldId}
         aria-describedby={this.fieldId && this.description ? `${this.fieldId}-description` : null}
         type={this.type}
-        onInput={this.handleInputChange}
+        onChange={this.handleInputChange}
```
ericclemmons: I know with React these events have subtle differences.  When does `onChange` happen here vs. `onInput`?

ashika01: I am going to guess it would same. And considering this is input form element I think it should remain `onInput` bcoz we want to capture each keystroke.

packages/amplify-ui-components/src/components/amplify-sign-in/amplify-sign-in.tsx
```diff
@@ -187,8 +187,14 @@ export class AmplifySignIn {
...
             </span>
-            <amplify-button type="submit" overrideStyle={this.overrideStyle} data-test="sign-in-sign-in-button">
-              {this.submitButtonText}
+            <amplify-button
+              type="submit"
+              disabled={this.loading}
+              overrideStyle={this.overrideStyle}
+              data-test="sign-in-sign-in-button"
+            >
+              <amplify-loading-spinner style={{ display: this.loading ? 'initial' : 'none' }} />
```
ashika01: Same callout as Eric's above. Probably we might wanna do `{this.loading ? ... : ... } `

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
ericclemmons: The DOM API means `onChange` fires when focus is lost, but React (incorrectly IMO) binds `onChange` to `onInput`.

It looks like Stencil's JSX does the same, but I'm more curious what prompted the change here: semantics, or difference in behavior?

packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.tsx
jordanranz: This is what I was using initially. We unfortunately run into the flickering render problem that is caused by conditionally loading content in a shadow dom.

Do we go with better UX or better code practice here?
@ericclemmons @ashika01

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
jordanranz: `shadowType` in the cypress shadow dom utility seems to add content to the input without triggering `onInput`. This means that in any of the forms for authenticator input field values were not properly set. `onChange` did trigger the change and worked properly. This may warrant a deeper dive though.

packages/amplify-ui-components/src/components/amplify-icon/amplify-icon.tsx
jordanranz: The `loading-spinner` would not pick up the style (rotation animation) from the parent when shadow was `true`. I can reinvestigate this.

packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.tsx
ericclemmons: We definitely don't want a flicker.  Is that happening the first time the spinner shows, or every time?

It could be several things:

- The bundle doesn't include `amplify-loading-spinner`, but should, so the flicker happens due to a network request.
- Stencil is running some async transition between state changes causing a "flash".  This is pretty unlikely, but if re-rendering anything causes a flicker (even when all components have been loaded), that's a big problem.

It's good that you identified this, because we _do_ want to ensure Authenticator has everything it needs to render for customers.  I can't say for certain what will happen in customer apps with split bundles being loaded asynchronously).

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
ericclemmons: Interesting!  Thanks for the clarification.  👍

packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.tsx
ericclemmons: @jordanranz & I synced up on this: it's the latter. Stencil is doing async DOM reconciliation, so there's a risk of flash whenever any DOM nodes (or Components) are added/removed.

By not changing the DOM structure (but modifying the `style`), the flicker doesn't happen.

👍

jordanranz:
Going to remove the unused type and add loading state to other forms.

PR #4944, opened 2020-02-18
fix(@aws-amplify/ui-components): bundle externals so components can be loaded via <script> tag

ericclemmons:
This reverts parts of #4767 (`rollup-plugin-node-externals`) so that `<script src="...">` includes `@aws-amplify/auth` (+ others) so it can work out-of-the-box.

There is still the question of how `Amplify.configure` will happen, but this also corrects `yarn storybook`, which was broken due to `@aws-amplify/auth` not existing in the bundle.

For reference, **this may be a regression with running `Amplify.configure`** from #4767, since there can be duplicate `Amplify` references now.

My perspective is that we should treat `<script>` usage as a "nice-to-have" in https://github.com/aws-amplify/amplify-js/milestone/16, but cater to the framework users for preview.

This would also mean updating stories, docs, and testing appropriately for those cases which would be running something like `yarn add aws-amplify @aws-amplify/ui-react`.

---


packages/amplify-ui-components/stencil.config.ts
```diff
@@ -1,21 +1,13 @@
 import { Config } from '@stencil/core';
 import { angularOutputTarget } from '@stencil/angular-output-target';
 import { reactOutputTarget } from '@stencil/react-output-target';
-import externals from 'rollup-plugin-node-externals';
```
sammartinez: Are we positive this is not needed?

ericclemmons: No, not yet. This was added in #4767, but see the PR body for risks & why we may want to not cater to the `<script>` usage.

(This PR gets stories and `<script>` tags working again, but we may not be ready for web component usage)

jordanranz: If you get rid of externals, won't we run into the same issue as before where a customer's Amplify instance and the bundled amplify version are disjointed?

ericclemmons: Yes, this is a problem.  This PR shows what it takes to resolve the `<script>` issue, but my argument is that maybe we should focus on framework usage first?

ericclemmons:
Closing now that this has gotten the necessary visibility.

PR #4945, opened 2020-02-18
fix(@aws-amplify/ui-components): federated button styling

ericclemmons:
When testing #4944, I noticed that the federated styles were likely broken by #4872:

> Screen Shot 2020-02-18 at 14 26 37

Switching to `scoped: true` fixes it:

> Screen Shot 2020-02-18 at 14 19 34

---

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
@@ -4,7 +4,7 @@ import { icons, IconNameType } from '../amplify-icon/icons';
 @Component({
   tag: 'amplify-sign-in-button',
   styleUrl: 'amplify-sign-in-button.scss',
-  shadow: true,
+  scoped: true,
 })
 export class AmplifySignInButton {
   /** (Optional) Override default styling */
--- a/packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
@@ -15,11 +15,7 @@ export class AmplifySignInButton {
     return (
       <Host class={`sign-in-button ${this.provider}`}>
         <button>
-          {this.provider in icons && (
-            <span class="icon">
-              <amplify-icon name={this.provider as IconNameType} />
-            </span>
-          )}
+          {this.provider in icons && <amplify-icon name={this.provider as IconNameType} />}
 
           <span class="content">
             <slot />
--- a/packages/amplify-ui-components/src/components/amplify-strike/__snapshots__/amplify-strike.spec.ts.snap
+++ b/packages/amplify-ui-components/src/components/amplify-strike/__snapshots__/amplify-strike.spec.ts.snap
@@ -2,22 +2,18 @@
 
 exports[`amplify-strike stories withOverrideStyle 1`] = `
 <amplify-strike class="strike">
-  <mock:shadow-root>
-    <span class="strike-content">
-      <slot></slot>
-    </span>
-  </mock:shadow-root>
-  Unstyled
+  <!---->
+  <span class="strike-content">
+    Unstyled
+  </span>
 </amplify-strike>
 `;
 
 exports[`amplify-strike stories withText 1`] = `
 <amplify-strike class="strike">
-  <mock:shadow-root>
-    <span class="strike-content">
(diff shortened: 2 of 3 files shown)


packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
```diff
@@ -4,7 +4,7 @@ import { icons, IconNameType } from '../amplify-icon/icons';
 @Component({
   tag: 'amplify-sign-in-button',
   styleUrl: 'amplify-sign-in-button.scss',
-  shadow: true,
+  scoped: true,
```
sammartinez: Whats the difference between shadow vs scoped?

packages/amplify-ui-components/package.json
```diff
@@ -65,7 +65,6 @@
     "prettier": "^1.17.1",
     "puppeteer": "1.19.0",
     "rimraf": "^3.0.0",
-    "rollup-plugin-node-externals": "^2.1.3",
```
sammartinez: I believe we can get this in prior to your other PR, do you mind removing this for now? can also have @jordanranz weigh in on this.

ericclemmons: Whoops, yes, this PR was dependent on #4944 to run, but the commit made it in by mistake.

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
ericclemmons: This came up in #4929 as well:
> https://github.com/aws-amplify/amplify-js/pull/4929/files#r380387938

Styles weren't picked up by the child elements without it.  `scoped` adds a prefix to classes to avoid conflicts.

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
```diff
@@ -15,11 +15,7 @@ export class AmplifySignInButton {
     return (
       <Host class={`sign-in-button ${this.provider}`}>
         <button>
-          {this.provider in icons && (
-            <span class="icon">
```
jordanranz: Does the `icon` css defined in this components scss file get applied to the `amplify-icon` `icon` class now?

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
sammartinez: Gotcha ok

jordanranz:
Would it be possible to fix the minor border inconsistency for the google button in this PR?

ericclemmons:
@jordanranz Regarding styling for the border, it's matching the `aws-amplify-react`'s styles now:

> Screen Shot 2020-02-21 at 10 33 58 AM

Adjusting it such that the border is removed makes the white Google icon blend into the background:

> Screen Shot 2020-02-21 at 10 36 53 AM


Instead, we should correct _all_ styles in future PRs to fit brand guidelines and be uniform.  Currently, federated sign in has to use custom buttons anyway since user pools aren't supported:
> Screen Shot 2020-02-21 at 10 32 04 AM

packages/amplify-ui-components/src/components/amplify-sign-in-button/amplify-sign-in-button.tsx
ericclemmons: There were a few issues that came up:

1. It currently looks like this:

	> Screen Shot 2020-02-21 at 10 45 56 AM

	With the following markup:

	> Screen Shot 2020-02-21 at 10 43 09 AM

1. Adjusting `scoped: true` gets us closer, but with excess padding:

	> Screen Shot 2020-02-21 at 10 43 55 AM

	Markup, with duplicate `icon` classes applied:

	> Screen Shot 2020-02-21 at 10 44 31 AM


1. With wrappers removed, it looks correct:

	> Screen Shot 2020-02-21 at 10 44 44 AM

	Markup:

	> Screen Shot 2020-02-21 at 10 45 03 AM


Honestly, this PR is a small tweak for the off-chance that customers are using federated identity pools, as user pools don't work.

My hope was to have this PR out to prevent breaks, but correct federated buttons (both API & UI) before GA.

ashika01:
@ericclemmons I see that you moved shadow to scoped in these. Is it just for CSS not applying properly? or is there another reason?

PR #4947, opened 2020-02-19
feat(@aws-amplify/ui-components): UI Components theming

ashika01:
_Description of changes:_
 - Updating CSS Variables
 - Bringing in `:host { ... }`

Theming for the components are done using [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) for the components. To use CSS variables in your app, add the following code inside css files, 

```
:root{

  /* You can override any css variable from the list like this */
  --amplify-primary-color: tomato !important;

  }
  
```

#### For Typography -
```
    --amplify-font-family: 'Amazon Ember', 'Helvetica Neue Light', 'Helvetica Neue', 'Helvetica', 'Arial', 'sans-serif';
    --amplify-text-xxs: 0.75rem;
    --amplify-text-xs: 0.81rem;
    --amplify-text-sm: 0.875rem;
    --amplify-text-md: 1rem;
    --amplify-text-lg: 1.5rem;
    --amplify-text-xl: 2rem;
    --amplify-text-xxl: 2.5rem;
```
#### For Colors -

```
    --amplify-primary-color: #ff9900;
    --amplify-primary-contrast: var(--amplify-white);
    --amplify-primary-tint: #ffac31;
    --amplify-primary-shade: #e88b01; 
    
    --amplify-secondary-color: #152939;
    --amplify-secondary-contrast: var(--amplify-white);
    --amplify-secondary-tint: #31465f;
    --amplify-secondary-shade: #1F2A37; 
    
    --amplify-tertiary-color: #5d8aff;
    --amplify-tertiary-contrast: var(--amplify-white);
    --amplify-tertiary-tint: #7da1ff;
    --amplify-tertiary-shade: #537BE5;

    /* Neutral */
    --amplify-grey: #828282;
    --amplify-light-grey: #c4c4c4;
    --amplify-white: #ffffff;
    
    --amplify-red: #dd3f5b;
```


packages/amplify-ui-components/src/global/theme.ts
```diff
@@ -14,40 +14,47 @@ import { injectGlobal } from 'emotion';
...
-    --amplify-amazon-orange: #ff9900;
-    --amplify-light-amazon-orange: #ffac31;
-    --amplify-dark-amazon-orange: #e88b01;
-    --amplify-squid-ink: #232f3e;
-    --amplify-light-squid-ink: #31465f;
-    --amplify-deep-squid-ink: #152939;
+    --amplify-primary-color: #ff9900;
+    --amplify-primary-contrast: var(--amplify-white);
+    --amplify-primary-tint: #ffac31;
+    --amplify-primary-shade: #e88b01;
```
ericclemmons: Love the `contrast`, `tint` and `shade` naming!  Here are the visuals for others to see:

- `--amplify-primary-contrast:` `#ffffff`
- `--amplify-primary-tint:` `#ffac31`
- `--amplify-primary-color:` `#ff9900`
- `--amplify-primary-shade:` `#e88b01`

packages/amplify-ui-components/src/global/theme.ts
```diff
@@ -14,40 +14,47 @@ import { injectGlobal } from 'emotion';
...
+    
+    --amplify-secondary-color: #152939;
+    --amplify-secondary-contrast: var(--amplify-white);
+    --amplify-secondary-tint: #31465f;
+    --amplify-secondary-shade: #1F2A37; 
+    
+    --amplify-tertiary-color: #5d8aff;
+    --amplify-tertiary-contrast: var(--amplify-white);
+    --amplify-tertiary-tint: #7da1ff;
+    --amplify-tertiary-shade: #537BE5;
```
ericclemmons: For visual reference secondary is `#152939` and tertiary is `#5d8aff`.

I've always struggled to find more semantic names for things like primary/secondary/tertiary:

- Brand
- Accent
- (Call to) Action

Then again, I think we've learned that pure semantics (whether HTML or CSS variables) aren't possible (hence Tailwind opting for `bg-${color}-${scale}`).

Nothing to change, but I was interested in your thoughts on this with regards to customizeable design systems.  ([Starbucks'](https://www.starbucks.com/developer/pattern-library/guidelines/color/) is a bit different, since the colors aren't intended to be customized.)

packages/amplify-ui-components/src/global/theme.ts
```diff
@@ -14,40 +14,47 @@ import { injectGlobal } from 'emotion';
...
-    --amplify-font-family: 'Amazon Ember', 'Helvetica Neue Light', 'Helvetica Neue', 'Helvetica', 'Arial', 'sans-serif';
+    
+    --amplify-red: #dd3f5b;
+
+    /* Breakpoints */
+    --breakpointSm: 320px;
+    --breakpointMd: 672px;
+    --breakpointLg: 1056px;
+    --breakpointXl: 1321px;
+    --breakpointMax: 1584px;
```
ericclemmons: Why are these `camelCase` vs. `kebab-case` like the other CSS vars?

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -4,7 +4,7 @@ import { TextFieldTypes } from '../../common/types/ui-types';
 @Component({
   tag: 'amplify-input',
   styleUrl: 'amplify-input.scss',
-  shadow: false,
+  shadow: true,
```
ericclemmons: Since I'm still figuring out the value of Shadow DOM, I assume that an alternative to this would be `scoped: true` and converting `:host` values in `amplify-input.scss` to `.label` values (e.g. `background-color: var(--amplify-secondary-contrast);`)?

(I believe there were some caveats with component customization you mentioned yesterday, so I wanted to better understand)

packages/amplify-ui-components/src/global/theme.ts
ashika01: Ha ha ha.. Thanks for doing this @ericclemmons

packages/amplify-ui-components/src/global/theme.ts
ashika01: Good suggestion I will make it kebab-case.

packages/amplify-ui-components/src/components/amplify-icon/amplify-icon.scss
```diff
@@ -1,3 +1,7 @@
+:host {
+  --icon-fill-color: var(--amplify-white);
+}
+
 .icon {
-  fill: var(--amplify-white);
+  fill: var(--icon-fill-color);
```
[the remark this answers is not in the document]
ashika01: I think with our current implementation. Fill with a color value would work fine. But I can think a little bit on this and we can come back and change this at a later point if needed.

packages/amplify-ui-components/src/global/theme.ts
ashika01: I like primary, secondary, tertiary or equivalents when it comes to global theming or brands. Versus for a full on pattern library, I would say should go for the something like tailwind'd or material-ui or any other equivalent where you give them a complete color palette, step function and Hues etc.

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
ashika01: Ok with CSS, I might not have to change anything here if we go for scoped. I could still use :host from my understanding. The major brownie point for shadow is that, shadow is going to avoid any css leaks.
Eg: [Stencil styling](https://stenciljs.com/docs/styling#shadow-dom-in-stencil)
Also, using ShadowDOM in web components seems like [common (maybe best?) practice](https://developers.google.com/web/fundamentals/web-components/shadowdom#styling).

ashika01:
@jordanranz @ericclemmons I am removing breakpoint from the code. CSS variables might not be the best way to do media queries. Also, we are not using it anywhere now. Removing will be less confusing for people who are looking at the code.

packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
```diff
@@ -18,8 +18,6 @@ export class AmplifyAmazonButton {
    * e.g. SignIn -> 'Create Account' link -> SignUp
    */
   @Prop() handleAuthStateChange: AuthStateHandler = dispatchAuthStateChangeEvent;
-  /** (Optional) Override default styling */
-  @Prop() overrideStyle: boolean = false;
```
ericclemmons: So happy to see this go! 👏

packages/amplify-ui-components/src/global/theme.ts
```diff
@@ -10,44 +10,48 @@
  * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
  * and limitations under the License.
  */
-import { injectGlobal } from 'emotion';
 
-injectGlobal`
-   :root {
+let customStyles = document.createElement('style');
```
ericclemmons: I forgot to check on this: will this be a problem for SSR?

PR #4979, opened 2020-02-25
feat(@aws-amplify/ui-components): I18n Support

ericclemmons:
## What This Does

- [x] `@aws-amplify/ui-components` injects `I18n` with `en` translations set.
- [x] Move UI strings from `constants.ts` to `AuthMessages` enum for typechecking and usage within components.
- [x] All components have been retrofitted to use `I18n.get(AuthMessages.*)`.
- [x] Inlined single-use strings from `constants.ts`.
- [x] Supports custom messages via:

	```js
	I18n.putVocabulariesForLanguage('en-US', {
	  [AuthMessages.SIGN_OUT]: 'Howdy',
	});
	```

#### Why only English?

Originally, I attempted to re-use translations from https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react/src/AmplifyI18n.tsx.

The problem is that those languages have ~41 translated strings, while **the new UI components have ~65**, many new & others being different from what existed before.

Because the new components weren't created to truly be 1-to-1, getting the two translations at parity would require a costly deep-dive into the two implementations to reconcile the differences (if even possible).

Supporting other languages would require effort similar to what it took for React to translate their docs:
> https://www.youtube.com/watch?v=V9PW4HeZyBw

##### Punting on I18n was a costly mistake

By delaying I18n, this PR required a significant level of visual & automated (TypeScript, Jest) validation touching every component created.  Had we opted for translation strings early via `I18n.get`, each component would've supported I18n out of the box with the same level of effort as it took to abstract strings into `constants.ts`.

##### `constants.ts` was an incorrect abstraction

Much has been written on this top that's worth reading, as engineers _really love abstractions_:

- https://twitter.com/CHERdotdev/status/1112819136147742720
- https://kentcdodds.com/blog/aha-programming
- https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction

In retrospect, there are seemingly clear abstraction boundaries to abide by:

- `constant.ts` for sharing **true constants**, such as response codes/strings from services like Cognito.
- `AuthMessages` for enumerating translation constants based on placement & intent (e.g. a header or submit button), rather than English phrases.
- Inlining strings that are used once or a few times meant for developers (and not end-users).  These are typically `logger.debug` messages, or an exception like `throw new Error(...)`.  Co-locating messages with the context in which they happen is a feature, not a bug to abstract away.
- There are **several** calls to `throw new Error(NO_AUTH_MODULE_FOUND)` throughout the code.  There are a couple ways to address this:

	- Preferably, an invariant utility to encapsulate the `if` logic as well:

	```
	checkForAuthModule();
	```

	- Alternatively, a custom `Error`:
	
	```
	throw new MissingAuthModuleError(additionalContext);
	```

[Delivers #171445892]

---


ericclemmons:
@ashika01 A lot of our tests are really just validating the default `@Prop()` values, which doesn't seem as valuable as (1) type assertions or (2) snapshots that we already have.

You recently removed emotion & `overrideStyle` tests in #4947, and I feel confident that I can do the same here:

- Testing rendered values is best left to snapshots & `newSpecPage`:

	> https://stenciljs.com/docs/unit-testing

	The value in snapshots is validating the rendered output (HTML) for a given input (props).  Changing I18n definitions, default language, structural changes are potentially breaking changes, which is why these are so valuable! 🤗



- Testing interactivity should stay:

	> https://stenciljs.com/docs/unit-testing#testing-component-class-logic

**tldr; Remove `.toEqual` and `.toBe` assertions in favor of snapshots. Keep interactivity tests.**

ericclemmons:
In retrospective, I believe my points are valid but can see emotional responses to "losing" tests.

The problem with our "component logic" tests is that they're not testing logic.  But it's a relatively small lift to get them passing without removing them.

packages/amplify-ui-components/src/global/index.ts
```diff
@@ -10,6 +10,11 @@
  * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
  * and limitations under the License.
  */
+import { I18n } from '@aws-amplify/core';
+import * as languages from '../common/translations';
+
+I18n.putVocabularies(languages);
+I18n.setLanguage(window?.navigator?.language || 'en-US');
```
ericclemmons: Stencil is fine with this, but our linter isn't.  I may update `I18n` to support a default locale or a `getLanguage` so we can default it.

ashika01: I like the idea of util.

ashika01: Why do we set this as `en-US`? Could we make it something generic so anyone can inject another language too?

packages/amplify-ui-components/stencil.config.ts
```diff
@@ -51,5 +51,8 @@ export const config: Config = {
       serviceWorker: null, // disable service workers
     },
   ],
-  globalScript: 'src/global/theme.ts',
+  globalScript: './src/global/index.ts',
+  testing: {
```
ashika01: Whats up with this? Just want some knowledge share :)

ericclemmons:
@ashika01 Great questions!  I'd like to take the time to answer this, since my opinions have changed over the years on this:

> I like having the strings used in constant.ts so in future it makes changing the string value easier than going to all the place inside the code.

You're right, as this is where things get subjective.  We agree that it's about abstractions **that make maintaining code easier & clearer**.  (My litmus test for when to or when not to abstract directly correlates with the friction/frustration to maintain code in it's present state.)

For example, IME bulk changes typically happen in dictionaries (e.g. I18n) or in per-env config files (e.g. [node-config](https://github.com/lorenwest/node-config)).  

The majority of the time, changes were within the context of the UI (e.g. label text, button aesthetics, # of retries, etc.). Co-locating/inlining values served to:

- Maintain context between _what_ the code does with _how_ it does it.
- Eliminate dead code (i.e. unused constants & config values are difficult to keep up-to-date).
- Reduce cross-referencing (i.e. fewer files open to know what something does).

Ultimately, my opinions stem from frustration with _this PR taking more effort than it should have_, because one-time use values were removed from the context of the UI and coupled with real constants.

> What are your thoughts there? Especially in similar cases like NO_AUTH_MODULE_FOUND make it as util might be unwanted abstraction vs keeping it in string constant. (Maybe I am missing clarity here :))

In this case, we have **24** occurrences of code like:

```js
    if (!Auth || typeof Auth.federatedSignIn !== 'function' || typeof Auth.currentAuthenticatedUser !== 'function') {
      throw new Error(NO_AUTH_MODULE_FOUND);
    }

```

I think we agree that 24 occurrences of copy/pasta is a bit extra. I can't even say that this check even needs to exist, since we rely on `import { Auth } from '@aws-amplify/auth';` (which would fail when bundling).

More importantly, the code isn't clear on _why_ this check needs to happen and _when_ it should happen in the UX.

Abtracting to an invariant check would be the first step to answering these questions.

packages/amplify-ui-components/stencil.config.ts
ericclemmons: Tests reference the component directly, so `I18n.get(AuthMessages.SIGN_IN)` returns `SIGN_IN`.

This bootstraps with the global script the same way that users will have the global bootstrap ran.

That way, styles & I18n are guaranteed to be set.

packages/amplify-ui-components/src/global/index.ts
ericclemmons: This default is largely here for one reason: `newSpecPage()` doesn't set `window.navigator.language` by default, like browsers do.


This code needs to change, as there's a bigger problem: **we only have an English translation**, so `I18n` defaulting to `fr-ca` will return enum values (e.g. `SIGN_OUT`) instead of the English "defaults".

Based on how `I18n.get` works (https://aws-amplify.github.io/docs/js/i18n#get), we can address this a couple of ways:

1. Use a custom `I18n` wrapper for UI components that falls back to English when a translated value doesn't exist.
1. Document and encourage customization via the `AuthMessages` enum, but move the "default" English phrases to enum values:

	Customers will do this:

	```js
	I18n.putVocabulariesForLanguage('en-US', {
	  [AuthMessages.SIGN_OUT]: 'Howdy',
	});
	```

	Behind the scenes, we have:

	```ts
	enum AuthMessages {
		SIGN_OUT = 'Sign out'
	}
	```

1. Update our `I18n.get(...)` calls to provide the English phrase as a default value:

	> https://aws-amplify.github.io/amplify-js/api/classes/i18n.html#get

The biggest takeaway is that **UI components should render default phrases even if `I18n.setLanguage` isn't English**.

ashika01: > UI components should render default phrases even if I18n.setLanguage isn't English.

Agreed. I like the concept of wrapper to fallback to english.

ashika01:
Cool. Thanks for explaining this @ericclemmons . I am good with this right now. Lets think more on the Auth-Module one and maybe revisit in later PR if needed.

packages/amplify-ui-components/src/global/index.ts
ericclemmons: @ashika01 I ended doing something "simpler", which can be described as _migrating `constants.ts` to `Translations.ts`:

- Customers can customize translations:

	```js
	import { Translations } from "@aws-amplify/ui-components";

	I18n.putVocabulariesForLanguage('en-US', {
	  [Translations.SIGN_OUT]: 'Howdy',
	});
	```

- No more need to default a language via `I18n.setLanguage`

- The values of the Translation enums _are_ the default English values.

- No need for a wrapper.

- Still adds `I18n.get` to the components, so we're ready for customer-provided translations.

- No need to instrument tests.

ericclemmons: Confirmed:

```shell
$ node
> require('@aws-amplify/ui-components').Translations
{ BACK_TO_SIGN_IN: 'Back to Sign In',
  CHANGE_PASSWORD_ACTION: 'Change',
  CHANGE_PASSWORD: 'Change Password',
  CODE_LABEL: 'Verification code',
  CODE_PLACEHOLDER: 'Enter code',
...
```

PR #4986, opened 2020-02-27
chore(@aws-amplify/ui-components): use globalStyle instead of globalScript

ericclemmons:
With `emotion` removed, this PR tests using `globalStyle` instead of `globalScript` for global theme variables:
> https://stenciljs.com/docs/config#globalstyle

I haven't been able to run our example or stories (both broken):
> Screen Shot 2020-02-26 at 4 14 20 PM

---


ericclemmons:
Double-submit bug. #4897 is correct.

PR #4989, opened 2020-02-27
feat (@aws-amplify/datastore): can be specified id in ModelInit.

sonodar:
* `DynamoDB` can specify unique constraints only by primary keys.
* `GraphQL Transform` can create a primary key with a composite key, but DataStore can join only by `id` currently.

Therefore, I want to set an arbitrary values for the primary key and apply a unique constraint.
However, in DataStore, `id` is automatically generated by `uuid` in the model generation process.
For these reasons, I changed it to be able to specify ID in `ModelInit`.

---

### DIFF
--- a/packages/datastore/src/datastore/datastore.ts
+++ b/packages/datastore/src/datastore/datastore.ts
@@ -234,8 +234,9 @@ const createModelClass = <T extends PersistentModel>(
 			} = modelInstanceMetadata;
 
 			const id =
-				// instancesIds is set by modelInstanceCreator, it is accessible only internally
-				_id !== null && _id !== undefined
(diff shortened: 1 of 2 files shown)


manueliglesias:
Hi @sonodar 

Having DataStore manage the ids was a very intentional design decision to help us track objects in non-ambiguous ways both locally and remote.

I think at the moment the *modelgen* step enforces an id for each model type; a change to allow primary key with a composite key would have to consider the surrounding tooling (the amplify CLI, the `npx amplify-app` package and its npm scripts)

sonodar:
umm, I understand the situation.
I try to other ways, thanks.

Issue #4981, opened 2020-02-26
additionalHeaders param is never passed to underlying function -- How do I access the current request headers or set them per request?


jensbodal:
**Which Category is your question related to?**

`API`

**What AWS Services are you utilizing?**

`graphql` via AWS AppSync

**Provide additional details e.g. code snippets**

Foremost, I'd like to ask if this was intentional to include the `additionalHeaders` param in the private `_graphql` function but never use it in any calling function (https://github.com/aws-amplify/amplify-js/blob/master/packages/api/src/API.ts#L380) .

It would seem the signature of the `graphql` function should be 

```
graphql({ query: paramQuery, variables = {}, authMode }, additionalHeaders = {})
```

Would a PR be accepted which passes the `additionalHeaders` param to the function? 

Otherwise, how do I access the request headers or set them on a per request basis? Currently we add a `traceid` header with each request by passing an async function at config time.

```
    Amplify.default.configure({
      API: {
        graphql_headers: async () => ({
          traceid: `${prefix}-${uuidv4()}`,
        }),
      },
    });
```

When graphql comes back with an error, we log the error and we would also like to log the `traceid`. The problem is we don't know what the `traceid` is set to and there is not a reliable way to know this as two simultaneous requests could change the `traceid` prior to logging.

The only option I can think of which would rely on a tight coupling to the implementation details of amplify-js would be to use axios request interceptors and handle errors there.  

Is there a proper way to do this?

jensbodal:
For now I am patching the original method until I know how to better proceed.

```
  let traceid;

  if (!apiPatched) {
    apiPatched = true;

    const privateApi = (API as any)._api.post.bind((API as any)._api);
    (API as any)._api.post = (endpoint: string, init: any) => {
      ({ traceid } = init.headers || {});
      return privateApi(endpoint, init);
    };
  }
```

manueliglesias:
> Would a PR be accepted which passes the `additionalHeaders` param to the function?

That would be great! 😄 

The axios interceptor option is also a good one since we are not planning on replacing it with another library, and if we ever do, it is going to be considered a breaking change and a new major version would be released.

jensbodal:
[...]

PR #5001, opened 2020-02-28
feat(@aws-amplify/api): pass additionalHeaders to graphql function

jensbodal:
Additional headers are merged **after** merging in headers set at config time.
Additional headers will overwrite any existing values and merge the rest.

_Issue #, if available:_ https://github.com/aws-amplify/amplify-js/issues/4981

_Description of changes:_ Existing private `_graphql` function allowed passing a param called `additionalHeaders` but was never used. The public `graphql` function now accepts an optional `additionalHeaders` param and is passed to private `_graphql` method. `_graphql` method was modified to merge in `additionalHeaders` **after** merging in previously set headers. This is how I would expect the parameter to function.

### DIFF
--- a/packages/api/__tests__/API-test.ts
+++ b/packages/api/__tests__/API-test.ts
@@ -954,6 +954,100 @@ describe('API test', () => {
 
 			expect(spyon).toBeCalledWith(url, init);
 		});
+
+		test('happy case query with additionalHeaders', async () => {
+			const spyonAuth = jest
+				.spyOn(Credentials, 'get')
+				.mockImplementationOnce(() => {
+					return new Promise((res, rej) => {
+						res('cred');
+					});
+				});
+
+			const spyon = jest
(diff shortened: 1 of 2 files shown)


bensie:
@manueliglesias @jordanranz It looks like the merge of #5215 accidentally stomped on this change so `additionalHeaders` are no longer working. Can we get the change added back?

bensie:
After more digging I see that `API.graphql(...)` is deprecated in favor of `GraphQLAPI.graphql(...)` and the latter supports `additionalHeaders` (though not for subscriptions - see #5576). This still appears to be an accidental merge error, but switching to the new GraphQLAPI is pretty straight forward so it's not critical to get this fixed. That said, the inline documentation still indicates that `additionalHeaders` are supported on `API.graphql(...)`, so at the very least a docs cleanup is probably warranted.

PR #5004, opened 2020-02-28
fix(core): revert dep addition. Move to peerDeps

iartemiev:


iartemiev:
> Why pinned version?

Good question. I'll change it to `^`

PR #5005, opened 2020-02-28
fix: update declaration for CognitoUserPool.signup and add the clientMetadata parameter.

keningharryscom:
_Issue #, if available:_

_Description of changes:_

`CognitoUserPool.signup` takes a "clientMetadata" parameter, adding this in the type declaration.


ashika01:
@keningharryscom The `clientMetadata` change has been updated a while ago. However looks like we still miss the jsdocs definition so will get that in. Thanks for your contribution. Appreciate it 🎉

PR #5007, opened 2020-02-28
ci: Add React Native integration testing with Detox

iartemiev:
Here's what it looks like when it's running: https://app.circleci.com/jobs/github/aws-amplify/amplify-js/11280

If someone wants to take it for a spin, lmk and we can copy this branch and then temporarily disable the other jobs in the pipeline to have it it only run the detox tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -297,7 +297,7 @@ jobs:
       - save_pods
       - run:
           background: true
-          command: xcrun simctl boot "iPhone 11"
+          command: xcrun simctl boot "iPhone 11" || true
           name: Start iOS simulator (background)
       - run:
           background: true
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -280,6 +280,9 @@ jobs:
       - run:
           name: Yarn Install
           command: yarn install --non-interactive --no-lockfile
+      - run:
+          name: Install netinfo
+          command: yarn add @react-native-community/netinfo && npx react-native link @react-native-community/netinfo
       # TODO: Metro Bundler doesn't work with linked packages. Consider copying the build output instead
       # - run:
       #     name: 'Link aws-amplify'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -254,10 +254,10 @@ jobs:
   integ_rn_ios_setup:
     macos:
       xcode: '11.3.1'
-    working_directory: ~/amplify-js-samples-staging
+    working_directory: ~/
     steps:
       - attach_workspace:
-          at: .
+          at: ./amplify-js-samples-staging
       - run: pwd && ls
       - run:
           name: Configure Detox
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@
 #
 # Check https://circleci.com/docs/2.0/language-javascript/ for more details
 #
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -250,6 +270,108 @@ jobs:
...
+      - run:
+          background: true
+          command: yarn start
+          name: Start Metro Packager (background)
+      - run:
+          name: Configure Detox
+          environment:
+            HOMEBREW_NO_AUTO_UPDATE: 1
+          command: |
+            brew tap wix/brew
```
sammartinez: Why `wix/brew`? Is there a certain homebrew that we need from wix? Or is this what they reccomend for adding detox?

.circleci/config.yml
```diff
@@ -250,6 +270,108 @@ jobs:
...
+            cp -rf ~/amplify-js/packages/storage/dist ./node_modules/@aws-amplify/storage/dist
+            cp -rf ~/amplify-js/packages/storage/lib ./node_modules/@aws-amplify/storage/lib
+            cp -rf ~/amplify-js/packages/storage/lib-esm ./node_modules/@aws-amplify/storage/lib-esm
+
+            cp -rf ~/amplify-js/packages/amplify-ui/dist ./node_modules/@aws-amplify/ui/dist
+            cp -rf ~/amplify-js/packages/amplify-ui/lib ./node_modules/@aws-amplify/ui/lib
+
+            cp -rf ~/amplify-js/packages/xr/dist ./node_modules/@aws-amplify/xr/dist
+            cp -rf ~/amplify-js/packages/xr/lib ./node_modules/@aws-amplify/xr/lib
+            cp -rf ~/amplify-js/packages/xr/lib-esm ./node_modules/@aws-amplify/xr/lib-esm
```
sammartinez: This looks painful, thank you for getting through this! My only question, NON BLOCKING, is Verdaccio will solve this? I know you have it as the callout but just want to double check if itll be something similar we would have to do with Verdaccio

.circleci/config.yml
iartemiev: Yep, that's just what their getting started says: https://github.com/wix/Detox/blob/master/docs/Introduction.GettingStarted.md#3-install-applesimutils

.circleci/config.yml
iartemiev: Lol yea it was a little painful, but fun too. I think we technically only need the `/lib` dirs, since that's what RN consumes, but I copied `dist` and `lib-esm` as well just in case.

Verdaccio will def solve this though. The reason we're doing that ^ is because we can't `yarn link` like we do for the other tests. So instead of linking or copying, we'll just `yarn add aws-amplify` from the Verdaccio registry running in CircleCI

sammartinez: Gotcha makes sense to me on this.

Ashish-Nanda: Yeah we only need `/lib`. This is a problem when doing local dev for amplify-js and React Native too.
But I think we can do it a better way in the future, using Verdaccio ideally or even a script file that runs a build and changes the output directory to the corresponding path in the app's node_modules. And then here we just run that script in the config rather than writing all the steps to copy it over in the config (a bit verbose)?
But yeah its not a blocker for now, and great stuff getting this to work! 👍

PR #4973, opened 2020-02-24
feat(@aws-amplify/ui-components): Storybook

ericclemmons:
_Fixes #3327_

This starts off the `@aws-amplify/ui-storybook` package, but there will be subsequent PRs to incrementally improve upon this:

- Mock the UI flow so that service calls are skipped.
- Add more customer use-cases.
- Convert code preview to show Angular/Vue/React/WebComponent equivalents.

> Screen Shot 2020-02-20 at 6 07 08 PM

> Screen Shot 2020-02-20 at 6 07 27 PM

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5023, opened 2020-03-02
chore(@aws-amplify/ui-components): output dist/docs.json

ericclemmons:
Stencil doesn't output `custom-elements.json`, which is used by other tools to reflect on what custom tags, properties, events, & styles a library provides.

This PR simply uses Stencil's existing `docs-json` plugin to output our components as JSON so that other tools can leverage them more effectively:

- Our own documentation
- Our own Storybook (#4973)

This format isn't the same schema as `custom-elements.json`, but we map `docs.json` to `custom-elements.json` in a future PR, once we identify the need.

Based on my research, `docs.json` is more descriptive than `custom-elements.json` and **closely matches what Stencil's `docs` plugin outputs, but as JSON.**

---

### DIFF
--- a/packages/amplify-ui-components/stencil.config.ts
+++ b/packages/amplify-ui-components/stencil.config.ts
@@ -45,6 +45,7 @@ export const config: Config = {
       proxiesFile: '../amplify-ui-react/src/components.ts',
     }),
     { type: 'dist' },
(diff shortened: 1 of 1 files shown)


ericclemmons:
Closing, as #4973 contained this commit.

PR #5041, opened 2020-03-04
Adding initstate property to aws-amplify-vue Authenticator component

dbhagen:
_Description of changes:_
Adding a initstate property to the Authenticator Vue component to allow the initial render to be a specific sub-component of the parent component.

The reasoning for it is this:
Right now I have an Auth component at `/auth`. This component basically just hosts the Amplify-Authenticator component. By default, this will display with a "sign-in" page. The AmplifyEventBus will not take an initial call during Create or Mount of the parent component, requiring a delay. That delay allows the "wrong" sub-component, the default "sign-in", to be rendered briefly, before the "correct", wanted sub-component is rendered.

The goal is to allow a single component landing page for all auth, with URL segments allowing the specification of the desired Authenticator view. `/auth/forgotPassword`, for example. With this passed via Vue-Router as `/auth/:state`, Authenticator can be populated in the template as follows:
```vue
      <authenticator
        :auth-config="authConfig"
        :initstate="$route.params.state"
      />
```

It's a simple solution for a somewhat simple problem.

### DIFF
--- a/packages/aws-amplify-vue/src/components/authenticator/Authenticator.vue
+++ b/packages/aws-amplify-vue/src/components/authenticator/Authenticator.vue
@@ -1,26 +1,47 @@
-/*
- * Copyright 2017-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
- *
- * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
- * the License. A copy of the License is located at
- *
- *     http://aws.amazon.com/apache2.0/
- *
- * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
- * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
- * and limitations under the License.
- */
+/* * Copyright 2017-2017 Amazon.com, Inc. or its affiliates. All Rights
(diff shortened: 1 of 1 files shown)


dbhagen:
Ok, wrestled with the formatting throwing everything around. Looks like it's build into the commit-hook, so at this point, I'm assuming this is actually the desired outcome. 🤷‍♂ 

What changed:
```vue
props: ['authConfig'],
```
became
```vue
props: ['authConfig', 'initstate'],
```
and
```vue
if (this.initstate) {
  this.displayMap = this.updateDisplayMap(this.initstate);
}
```
was added right below
```vue
async mounted() {
  this.logger = new this.$Amplify.Logger(this.$options.name);
```

amhinson:
If I'm understanding correctly, it looks like the new `@aws-amplify/ui-vue` library [supports this](https://docs.amplify.aws/ui/auth/authenticator/q/framework/vue#prop-initial-auth-state). Is there a reason why you can't use that library instead of `aws-amplify-vue`?

Here's a [migration guide](https://docs.amplify.aws/ui/auth/authenticator/q/framework/vue#migration) for upgrading.

ericclemmons:
Thanks for your contribution!  As @amhinson mentioned previously, we'll continue our support for Vue through `@aws-amplify/ui-vue`:
> https://docs.amplify.aws/ui/auth/authenticator/q/framework/vue#prop-initial-auth-state

PR #5045, opened 2020-03-05
[Feat] UI components - Radio Group

ashika01:
_Issue #, if available:_ 

The radio button lets you choose all radio button under a form or group.
image

_Description of changes:_

This PR adds `amplify-radio-group` thats take care of the state for radio button management.


ashika01:
Closing this based on a fix eric suggested. 

> You can have as many radio groups on a page as you like, as long as each has its own unique name.


This is the gotcha we were missing. We might need radio group for shadow implementation. But that is not of immediate concern so let's come back and revisit this later for that.

PR #5051, opened 2020-03-06
Fixed type for updateAttributes array

ffocsi:
_Description of changes:_
Bug which doesn't allow you to use an array of CognitoUserAttribute but requires ICognitoUserAttributeData, when the method actually accepts either.


ashika01:
@ffocsi Thanks for the PR. Apologies for getting back so late. We are continuously working on internal process to make the contribution process better.

PR #5059, opened 2020-03-06
feat(@ui-components/ui-storybook): show multi-framework code

ericclemmons:
By default, Storybook's `<Source>` container shows the story code:

```js
() => {
	<AmplifyAuthenticator>
		<h1>You are signed in!</h1>
		<AmplifySignOut />
	</AmplifyAuthenticator>
)
```

Instead, the code should (nearly?) be a copy/pasteable snippet for:

- React
- Angular
- Vue

Example:
> https://ionicframework.com/docs/api/card

For reference, Ionic keeps each framework snippet separate:
> https://github.com/ionic-team/ionic/tree/master/core/src/components/card/usage

Other projects have live playgrounds:
> https://evergreen.segment.com/components/button/

So maybe rather than having per-framework source (that'll happen once Storybook launches multi-framework support in Q2!), this can instead make each story a little more contextual than what's shown above (e.g. adding `import`).

---

### DIFF
--- a/packages/amplify-ui-storybook/.storybook/PreviewStory.js
+++ b/packages/amplify-ui-storybook/.storybook/PreviewStory.js
@@ -1,0 +1,37 @@
+import {
+	DocsContext,
+	getSourceProps,
+	Preview,
+	Story,
+} from '@storybook/addon-docs/blocks';
+import React, { useContext } from 'react';
+
+export const PreviewStory = ({ id }) => {
+	const context = useContext(DocsContext);
+	const { code } = getSourceProps({ id }, context);
+	console.log(context.storyStore.fromId(id));
+
+	const frameworkCode = `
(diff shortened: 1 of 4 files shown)


ericclemmons:
Closing until there's formal support within Storybook.

From what I'm seeing, I'd recommend there being a singular `Authenticator.stories.js` and per-framework equivalents (e.g.`react/Authenticator.stories.js` or `Authenticator.stories.react.js`) that extend the `exports`.

Put another way, each framework's stories would export the same stories as the parent one, since decorators & implementations are similar, but different enough to warrant it.

```js
// Authenticator.stories.js

export const defaults = () => {
throw new Error(`Missing story`)
}

// Authenticator.stories.react.js

import * as stories from "./Authenticator.stories"

export {
  ...stories,
  defaults() {
    return <AmplifyAuthenticator />
  }
}
```

PR #5065, opened 2020-03-09
feat(@ui-components/username-attributes): Add usernameAlias support

ashika01:
**_Description of changes:_** Adds usernameAlias support.

**_How to test:_**
- run `yarn start` on `packages/ui-components`
- make sure to use `aws-exports` with auth setup.

**_Usage:_** `usernameAlias` can be one of these `email | phone_number | username`
`<amplify-authenticator usernameAlias="username" />`


packages/amplify-ui-components/src/components/amplify-forgot-password/amplify-forgot-password.tsx
```diff
@@ -29,31 +34,89 @@ export class AmplifyForgotPassword {
...
+            required: true,
+            handleInputChange: event => this.handlePhoneNumberChange(event),
+            inputProps: {
+              'data-test': 'phone-number-input',
+            },
+          },
+        ];
+        break;
+      case 'username':
+      default:
```
ericclemmons: Since `username` is the default, would having `default` throw an error be a useful guard against invalid types?

```js
default:
	throw new Error(`Unsupported usernameAlias: ${this.usernameAlias}. Use `${Object.values(UsernameAlias)`)
```

_This example would require `UsernameAlias` to be an `enum`._

That would help customers why typo `usrname` or `phone` (instead of `phone_number`).

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -22,6 +22,8 @@ export class AmplifyAuthenticator {
   @Prop() initialAuthState: AuthState.SignIn | AuthState.SignUp = AuthState.SignIn;
   /** Federated credentials & configuration. */
   @Prop() federated: FederatedConfig;
+  /** Username Alias is used to setup authentication with `username`, `email` or `phone_number`  */
+  @Prop() usernameAlias: UsernameAlias = 'username';
```
ericclemmons: Since this component doesn't do anything with the value of `usernameAlias`, can the default value be removed?

- That would translate to `usernameAlias={undefined}`, which would not affect snapshots (where `usernamealias="username"` is now part of the contract).
- The components that pivot based on the value of `usernameAlias` would be responsible for defaulting the value (as they already are).

packages/amplify-ui-components/src/components/amplify-forgot-password/amplify-forgot-password.tsx
ashika01: I like this. I can add it in the updates.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
ashika01: Let me check this out. I started it like that. I just defaulted this to update readme I guess.

packages/amplify-ui-components/src/components/amplify-sign-in/amplify-sign-in.tsx
```diff
@@ -139,15 +169,72 @@ export class AmplifySignIn {
...
+        this.handleAuthStateChange(AuthState.ForgotPassword, { username: this.userInput });
       }
     } finally {
       this.loading = false;
     }
   }
+  async componentWillLoad() {
+    const formFieldInputs = [];
+    switch (this.usernameAlias) {
+      case 'email':
```
sammartinez: NIT: Should we make these cases a constant? Im just thinking similar to redux in my head

ashika01: This is the one I was talking with Eric, there is possibility of enum happening here. Looking into that.

ashika01: I am still keeping it as string but it adheres to an enum check now. Also added a quick error thrown for misspelled alias.

ashika01:
@ericclemmons @sammartinez Its ready for re-review. I am working on checking the build error. But as planned, lets get this ready for merge on back of the `pollyClient` merge

packages/amplify-ui-components/src/components/amplify-forgot-password/amplify-forgot-password.tsx
ashika01: Added enum and a util function to throw new error. Let me know how it looks.

ericclemmons:
Looks like we're awaiting https://github.com/aws-amplify/amplify-js/pull/4879's checks before we can continue?

ashika01:
@ericclemmons Yup.. we should okay to test our locally tho.

packages/amplify-ui-components/src/common/helpers.ts
```diff
@@ -23,6 +29,22 @@ export const dispatchAuthStateChangeEvent: AuthStateHandler = (nextAuthState: Au
...
+  }
+
+  const sanitizedPhoneNumberValue = phoneNumber.phoneNumberValue.replace(/[-()\s]/g, '');
+
+  return `${phoneNumber.countryDialCodeValue}${sanitizedPhoneNumberValue}`;
+};
+
+export const checkUsernameAlias = (usernameAlias: any) => {
+  if (!(usernameAlias in UsernameAlias)) {
+    throw new Error(`Invalid username Alias - ${usernameAlias}. Instead use ${Object.values(UsernameAlias)}`);
```
sammartinez: Is this a similar one to what is pre-existing?

packages/amplify-ui-components/src/common/types/auth-types.ts
```diff
@@ -98,3 +98,11 @@ export enum ChallengeName {
 export enum AuthFormField {
   Password = 'password',
 }
+
+export enum UsernameAlias {
+  username = 'username',
```
sammartinez: I believe these all need to be capitalized correct? If we are following enum structure

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -22,6 +22,8 @@ export class AmplifyAuthenticator {
   @Prop() initialAuthState: AuthState.SignIn | AuthState.SignUp = AuthState.SignIn;
   /** Federated credentials & configuration. */
   @Prop() federated: FederatedConfig;
+  /** Username Alias is used to setup authentication with `username`, `email` or `phone_number`  */
+  @Prop() usernameAlias: UsernameAliasStrings;
```
sammartinez: Does just calling them `UsernameAlias` work instead?

ashika01: Nope because we are checking for string value we need utilize `UsernameAliasStrings`.
We are basically doing something like this https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-compile-time

packages/amplify-ui-components/src/common/helpers.ts
ashika01: Nope there is no error handling for this in react now. But this is valuable check for our customers.

packages/amplify-ui-components/src/common/types/auth-types.ts
ashika01: I put lowercase to match with the string value. I can revisit in following PR if need be.

ashika01:
@ericclemmons Agreed. It is also getting difficult to test this locally. Let's test this in our test plan. Make note of all the issues we are seeing and address it in following PRs.

dobrynin:
@ashika01 what if I want to allow users to use email OR phone (allowing either)?

PR #5085, opened 2020-03-12
Export necessary providers and types from the package root

Amplifiyer:
Export necessary providers and types from the package root to avoid importing from generated paths

_Issue #, if available:_

_Description of changes:_


Amplifiyer:
> LGTM 🌮 Do we have an e2e that will accompany this?

Yes, it will be covered by already existing integration tests. Change in this line https://github.com/aws-amplify/amplify-js-samples-staging/blob/master/samples/angular/pubsub/aws-iot-provider/src/main.ts#L8 will be the verification.

Issue #5029, opened 2020-03-03
Inconsistent CodeDeliveryDetails usage.


mousedownmike:
**Describe the bug**
The return of `CodeDeliveryDetails` is inconsistent across Auth functions.

**To Reproduce**

1.  Call `Auth.signUp()`
2. Observe delivery details in: `response.codeDeliveryDetails` note lower case `code`.
3. Call `Auth.resendSignUp()`
4. Observe delivery details in: `Promise<string>`.  That `string` is actually JSON with `CodeDeliveryDetails`, note upper case `Code`.
5. Call `Auth.forgotPassword`
6. Observe delivery details in `Promise<any>` containing `CodeDeliveryDetails`.

**Expected Behavior**

`CodeDeliveryDetails` should be consistently accessible as `CodeDeliveryDetails` either as part of an `ISignUpResult` or a `Promise<any>`.  This would make it consistent with the documented [Forgot Password API](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ForgotPassword.html).  

**Required Changes**

1. Change [`codeDeliveryDetails` in `ISignUpResult`](https://github.com/aws-amplify/amplify-js/blob/ac34816df326331cfe04474fdf35790c52f4a1d3/packages/amazon-cognito-identity-js/index.d.ts#L287) to be `CodeDeliveryDetails`.  There seems to be some mixed case fields in there already so it doesn't look like it would break any conventions.
2. Modify the [signature of `Auth.resendSignUp`](https://github.com/aws-amplify/amplify-js/blob/cf8738ac5e7961a163ce158a7383bc27f693475f/packages/auth/src/Auth.ts#L393) to return a `Promise<any>`.

mousedownmike:
I have a revision to this.  In practice it appears that `Auth.resendSignUp` returns a `CognitoIdentityServiceProvider.ResendConfirmationCodeResponse`.  This makes having the response typed as `string` even more difficult to deal with.  This should really be typed to `any`.

PR #5112, opened 2020-03-16
fix(@aws-amplify/auth): incorrect return type for Auth.resendSignUp

mousedownmike:
Partial fix for #5029 

Updated `Auth.resendSignUp` signature to return `any` type instead of `string`.  Revised Spy to be reflective of an actual `CodeDeliveryDetails` response.


packages/auth/src/Auth.ts
```diff
@@ -388,12 +388,12 @@ export default class AuthClass {
...
 	 * @param {String} username - The username to be confirmed
 	 * @param {ClientMetadata} clientMetadata - Metadata to be passed to Cognito Lambda triggers
-	 * @return - A promise resolves data if success
+	 * @return - A promise resolves code delivery details if successful
 	 */
 	public resendSignUp(
 		username: string,
 		clientMetadata: ClientMetaData = this._config.clientMetadata
-	): Promise<string> {
+	): Promise<any> {
```
ericclemmons: It possible to re-use `CodeDeliveryDetails` here?

https://github.com/aws-amplify/amplify-js/blob/ac34816df326331cfe04474fdf35790c52f4a1d3/packages/amazon-cognito-identity-js/index.d.ts#L6-L10

mousedownmike: @ericclemmons are we sure it always resolves to `CodeDeliveryDetails`?  I wasn't positive so I chose `any`.  The code builds fine with `CodeDelieryDetails` in there and I can push that if you'd like.

mousedownmike: The branch with `CodeDeliveryDetails` is here: 
https://github.com/mousedownmike/amplify-js/tree/resendsignup_codedeliverydetails

Happy to close this PR and start a new one or merge the change into this one.  Whichever solution works best and whenever it works.  

Cheers,
Mike

ericclemmons:
Everything looks good, but we'll need to manually validate the response payload against the types to be sure whether `any` is the correct type.

The underlying call indicates a `string`, but we have to be certain:
https://github.com/aws-amplify/amplify-js/blob/ac34816df326331cfe04474fdf35790c52f4a1d3/packages/amazon-cognito-identity-js/src/CognitoUser.js#L1280-L1286

f22hd:
Exactly what we added to our project to resolve this mistype of the return type.
Hope this PR approved and closed. Thanks.

packages/auth/src/Auth.ts
ericclemmons: I'm going to get this merged in once tests pass.

We'll do some more tests to confirm the payload shape, but for now this is a great step forward 👍

ericclemmons: (I went ahead & opened https://github.com/aws-amplify/amplify-js/pull/6561 for comparison)

PR #5128, opened 2020-03-17
feat(@aws-amplify/datastore): Support non-@model types in DataStore

manueliglesias:
_Issue #, if available:_
See aws-amplify/amplify-cli#2988

_Description of changes:_
This change allows the usage of types with no `@model` directive.

Field values of a non-\@model type will be stored alongside the @model instance

e.g. with the following schema, the JSON shown is stored

```graphql
type Note @model {
  id: ID!
  content: String!
  metadata: Metadata
}

type Metadata {
  author: String!
  tags: [String]
  nested: Nested
}

type Nested {
  x: Int
}
```

```json
{
  "__typename": "Note",
  "_lastChangedAt": 1584144474971,
  "_version": 1,
  "content": "Content 1584144474557",
  "createdAt": "2020-03-14T00:07:54.911Z",
  "id": "b41f57cc-4375-4bc5-83e2-1da694381f7f",
  "metadata": {
    "author": "Manolo",
    "tags": [
      "amazon",
      "aws",
      "mobile"
    ],
    "nested": {
      "x": 12345
    }
  },
  "updatedAt": "2020-03-14T00:07:54.911Z"
}
```

### DIFF
--- a/packages/datastore/src/datastore/datastore.ts
+++ b/packages/datastore/src/datastore/datastore.ts
@@ -15,12 +15,15 @@ import {
 	GraphQLScalarType,
 	InternalSchema,
 	isGraphQLScalarType,
+	ModelFields,
 	ModelFieldType,
 	ModelInit,
 	ModelInstanceMetadata,
+	ModelOrTypeConstructorMap,
 	ModelPredicate,
 	MutableModel,
 	NamespaceResolver,
+	NonModelTypeConstructor,
 	PaginationInput,
 	PersistentModel,
 	PersistentModelConstructor,
@@ -28,6 +31,7 @@ import {
 	Schema,
 	SchemaModel,
 	SchemaNamespace,
+	SchemaType,
 	SubscriptionMessage,
 	SyncConflict,
 	SyncError,
@@ -62,7 +66,7 @@ const SETTING_SCHEMA_VERSION = 'schemaVersion';
 
 let storage: Storage;
 let schema: InternalSchema;
-const classNamespaceMap = new WeakMap<
+const modelNamespaceMap = new WeakMap<
 	PersistentModelConstructor<any>,
 	string
 >();
@@ -70,7 +74,7 @@ const classNamespaceMap = new WeakMap<
 const getModelDefinition = (
 	modelConstructor: PersistentModelConstructor<any>
 ) => {
-	const namespace = classNamespaceMap.get(modelConstructor);
+	const namespace = modelNamespaceMap.get(modelConstructor);
 
 	return schema.namespaces[namespace].models[modelConstructor.name];
 };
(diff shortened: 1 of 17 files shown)


packages/datastore/src/datastore/datastore.ts
```diff
@@ -326,6 +330,29 @@ const createModelClass = <T extends PersistentModel>(
 	return clazz;
 };
 
+const createTypeClass = <T extends PersistentModel>(
```
Amplifiyer: can be renamed as well according to the new names

packages/datastore/src/types.ts
```diff
@@ -146,18 +164,25 @@ type ModelField = {
...
 
 //#region Model definition
+export type NonModelTypeConstructor<T> = {
+	new (init: T): T;
+};
 export type PersistentModelConstructor<T extends PersistentModel> = {
 	new (init: ModelInit<T>): T;
 	copyOf(src: T, mutator: (draft: MutableModel<T>) => T | void): T;
 };
+export type ModelOrTypeConstructorMap = Record<
```
Amplifiyer: As per the new naming convention, can we call this a generic `TypeConstructorMap` which is either a `PersistentModelConstructor` or `NonModelTypeConstructor`?

packages/datastore/src/datastore/datastore.ts
```diff
@@ -187,22 +182,25 @@ const initSchema = (userSchema: Schema) => {
 	return userClasses;
 };
 
-const createModelClassses: (
+const createModelAndTypeClassses: (
```
Amplifiyer: Is it better to call this a generic `createTypeClasses` which can be either model types or non model types?

packages/datastore/src/datastore/datastore.ts
manueliglesias: Cool, yeah, I'll remove the generic constraint since it is not needed here (`PersistentModel` checks for the presence of an `id` field, and non-models don't require it)

packages/datastore/src/types.ts
manueliglesias: Gotcha 👍

packages/datastore/src/datastore/datastore.ts
manueliglesias: It is, thanks! 👍

packages/datastore/src/datastore/datastore.ts
```diff
@@ -330,9 +330,7 @@ const createModelClass = <T extends PersistentModel>(
 	return clazz;
 };
 
-const createTypeClass = <T extends PersistentModel>(
-	typeDefinition: SchemaNonModel
-) => {
+const createTypeClass = <T>(typeDefinition: SchemaNonModel) => {
```
Ashish-Nanda: Since this is only for non-models, should we change it to `createNonModelClass`?

manueliglesias: Sure! Sounds good

PR #5164, opened 2020-03-22
Ui components/accessibilty

ashika01:
_Description of changes:_

- adding some basic keyboard accessibility support to the component's 
- enables proper tabbing order
- add tabbing to toast component

Reference links:
- https://a11yproject.com/posts/creating-valid-and-accessible-links/
- https://www.w3.org/TR/WCAG21/#keyboard-accessible


packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
```diff
@@ -46,7 +46,9 @@ export class AmplifyButton {
   render() {
     return (
       <button
-        class={{
+        class={
+        this.type === "anchor" ? "anchor":
```
jordanranz: Let's make a clear distinction between `type` and maybe something like `variant`. `type` should just be the html button `type` [here](https://www.w3schools.com/tags/att_button_type.asp). A `variant` would be a different curation of the look of the button. For example, an `anchor` button can still have a submit `type`.

jordanranz: Also, this class can just be keyed off of the `variant`. Like:
```
class={{
  [this.variant]: true
}}      
```

So if `variant="anchor"` it would apply the `.anchor` class

ashika01: This makes sense. Making the change right now. Sorry didnt think of this earlier.

packages/amplify-ui-components/src/components/amplify-toast/__snapshots__/amplify-toast.spec.ts.snap
```diff
@@ -6,7 +6,7 @@ exports[`amplify-toast spec: Render logic -> should render an empty link tag by
     <div class="toast">
       <amplify-icon class="toast-icon" name="warning"></amplify-icon>
       <slot></slot>
-      <a class="toast-close"></a>
+      <button class="toast-close"></button>
```
ericclemmons: This is the tldr; of the PR. 👍

jordanranz:
I do agree with @ericclemmons on the comment above representing `variant` with visual language. This is however not a blocker.

Issue #3092, opened 2019-04-14
[I18n] Use slugs instead of English strings as keys 


ChristopheBougere:
**Is your feature request related to a problem? Please describe.**
The current i18n system is working but:
- Need to search in the code to find missing literals to translate.
- There is no complete list of all the existing literals, so we don't know the % of translation in a specific language.
- English strings are the translation keys. If a typo is fixed in English (or just a space added), then all translations will become obsolete (they won't be used anymore, unless someone updates the key).

**Describe the solution you'd like**
Instead of using English strings as keys, we could use some slugs (no punctuation/spaces). It would also be easier to separate translations by modules (`auth.signin.username` and `auth.signin.password` would be together).
Behaviour could be inspired by [https://www.i18next.com](i18next)
I know it's a breaking change, but maybe for a major release? Or maybe at least for new translations?

**Describe alternatives you've considered**
n/a

**Additional context**
I've translated a few literals in #3091. It's difficult to find the keys, which includes spaces and punctuation. Also, there is no list of all literals which is cumbersome.

haverchuck:
@ChristopheBougere Thanks for this request.  We will look into this as part of a future major release.

ericclemmons:
#4979 addresses this, which adds `Translations` as key/value pairs using typed keys:

```ts
import { I18n } from "aws-amplify"
import { Translations } from "@aws-amplify/ui-components"

 I18n.putVocabulariesForLanguage('en-US', {
   [Translations.SIGN_OUT]: 'All done!',
 });
```

See: https://github.com/aws-amplify/amplify-js/blob/master/packages/amplify-ui-components/src/common/Translations.ts

We haven't ported over prior translations yet, but the foundation is there to support it now, thanks to your input @ChristopheBougere!

PR #5166, opened 2020-03-22
fix(@aws-amplify/aws-amplify-react): Improve french translations

ChristopheBougere:
_Description of changes:_

Just improved or added a bunch of french translations.
I must say that it is quite a cumbersome task, and I probably missed some of them because there is no official list of all translations. There are some dynamic strings, or messages depending on thrown Error that makes it very difficult to anticipate everything. I already filed this as a feature request in #3092


ericclemmons:
@ChristopheBougere Thanks for your PR!

I ran into the same issue in #4979 as part of our UI Refactor (#3279).

You can see an example here:
> https://github.com/aws-amplify/amplify-js/blob/ui-components/master/packages/amplify-ui-components/src/components/amplify-sign-in/amplify-sign-in.tsx#L255

The bad news is that there are more strings available in the refactor than `master`.

I'd love to work with you on getting proper translations into #3279.   Your points in #3092 are spot on.

How could I help?

ChristopheBougere:
Hi @ericclemmons 
Great to see that the UI refactor includes a translations centralization!
I'd be happy to take care of french translations. I think there is only English for now (with strings located in [this file](https://github.com/aws-amplify/amplify-js/blob/ui-components/master/packages/amplify-ui-components/src/common/Translations.ts)). What would be the process to add a new language?

ericclemmons:
@ChristopheBougere That's right.  In #4979, I tried to carry forward existing translations, but there the implementations weren't 1-to-1.

What we'll need to do is revert a bit of dc263817d76da3d42f1a19ffd25037f1367ad8d0 so that translations are automatically bootstrapped.

If I setup a PR & lay out how to continue the implementation, would you be able to bring your work over?

ChristopheBougere:
Sure I could do that. Correct me if I'm wrong, but basically it should consist in a translation of all the strings in [this file](https://github.com/aws-amplify/amplify-js/blob/ui-components/master/packages/amplify-ui-components/src/common/Translations.ts)?

ericclemmons:
@ChristopheBougere I'll need a bit to setup the PR, but you're right: it's effectively just a map of those translations:

```ts
import { Translations } from './common/Translations';

export const fr: Record<Translations, string> = {
   ...
   [Translations.SIGN_OUT]: 'Déconnexion',
   ...
}
```

ChristopheBougere:
@ericclemmons sounds good to me, let me know when I can help

thibaultdalban:
@ericclemmons and @ChristopheBougere I'm not sure if it's the right place to post :confused:
I would like to let you know that we just release the first version of an npm package dedicated to the Amplify translations.
Please have a look at https://www.npmjs.com/package/amplify-i18n I would be very interested in your feedback.
Currently, the translations have been done by machine translation but we are going to do an entire review with our professional translators.

Issue #2120, opened 2018-11-15
List users in user pool implementation


PedroS11:
Cognito identity has a method for listing users in a user pool, however, amplify doesn't have one so I created a function to solve this problem and I would like to know your opinion about my solution because I would like to do a PR with this feature

```
var AWS = require('aws-sdk');

exports.getUsers= () => {
	var params = {
	  UserPoolId: USER_POOL_ID
	  AttributesToGet: [
	    'ATTRIBUTE_NAME',
	  ],
	};
	
	return new Promise((resolve, reject) => {
		AWS.config.update({ region: USER_POOL_REGION, 'accessKeyId': AWS_ACCESS_KEY_ID, 'secretAccessKey': AWS_SECRET_KEY });
		var cognitoidentityserviceprovider = new AWS.CognitoIdentityServiceProvider();
		cognitoidentityserviceprovider.listUsers(params, (err, data) => {
			if (err) {
				console.log(err);
				reject(err)
			}
			else {
				console.log("data", data);
				resolve(data)
			}
		})
	});
}
```

In order to this be implemented, it would require to be a function receiving the attributes to get as parameters. I'm open to suggestions/errors.

elorzafe:
@PedroS11 thanks for your feedback. I think listing users is more oriented to **app admin** rather than **app users** which Amplify is focus on. I can mark this as feature-request so the team can discuss more about it. 

Why do you need to list all the app users? 

Thanks!

PedroS11:
Because I had an admin panel to list the users in a user pool and wanted to do actions on them (delete/update)

mcammaert:
+1, Also add/remove users to groups.

undefobj:
@PedroS11 @mcammaert we're looking at providing such a feature in the future. However, it wouldn't be something that goes into the Amplify library because as pointed out by @elorzafe this is more of an administrative feature rather than a client application feature. However, I can see where you might have a client application that needs to list all members in a group in which case executing a REST or GraphQL call would be a more secure and scalable design.

If we were to automate this process by using the Amplify CLI and allow you to make API calls using your own resources would that meet your needs?

undefobj:
One idea is that we could automate setting up an API endpoint (perhaps with a Lambda function or replicating the user information in a DynamoDB table, which supports LIST operations better) using the CLI. Baking admin features into the client library wouldn't be optimal from a scalability perspective and additionally it would add bloat to the bundle size.

mcammaert:
[...]

PR #5175, opened 2020-03-24
feat(@aws-amplify/auth): add listPoolUsers method

pedrobergamini:
_Issue #, if available:_
#2120
_Description of changes:_
This PR introduces the public method `listPoolUsers` in Auth class. It closes the feature request #2120 by implementing a way of listing users through CognitoIdentityServiceProvider from AWS-SDK. In the first moment I was thinking on implementing this by using amazon-cognito-identity-js lib, however, the client request method only allows endpoints which authorization is not required.

I'm still working with the tests.

### DIFF
--- a/packages/auth/__tests__/auth-unit-test.ts
+++ b/packages/auth/__tests__/auth-unit-test.ts
@@ -500,11 +500,14 @@ describe('auth unit test', () => {
 
 			await auth.confirmSignUp('username', code);
 
-			expect(
-				await CognitoUser.prototype.confirmRegistration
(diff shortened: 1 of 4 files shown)


undefobj:
@pedrobergamini we added this feature in Amplify CLI which you can use in Amplify JS libraries last year. You can read more about it here: https://aws-amplify.github.io/docs/cli-toolchain/quickstart#administrative-actions

Unfortunately the linked issue was never closed. I'll go ahead and respond to it now. Thank you for taking the time to open up this PR though and putting work into being a contributor.

PR #5184, opened 2020-03-25
UI components - Cleanup phone-field style

ashika01:
_Description of changes:_
- removing scoped/shadow from `amplify-input` component.
- Looks like this was an unnecessary change. I introduced this in the first place to make `phone-field` component to function properly.

If this doesn't introduce any concern, I suggest we keep the component clean of any `shadow/scoped` stuff until need arises.


ashika01:
Thoughts/concerns?

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -4,7 +4,6 @@ import { TextFieldTypes } from '../../common/types/ui-types';
 @Component({
   tag: 'amplify-input',
   styleUrl: 'amplify-input.scss',
-  scoped: true,
```
jordanranz: Will removing `scoped` here affect the theming?

ashika01: Nope.. Not right now. Since we dont have any scoped style applied to this component as of now nor do any component level styling using css vars. It doesn't create a problem right this moment. However we need to revisit every single style file when we do component level theming and mobile first styles.

PR #5202, opened 2020-03-26
feat(@aws-amplify/datastore): Make DataStore available in aws-amplify…

manueliglesias:
… package

_Issue #, if available:_
N/A

_Description of changes:_
Make DataStore available in the `Amplify` object when using the `aws-amplify` package

### DIFF
--- a/packages/aws-amplify/src/index.ts
+++ b/packages/aws-amplify/src/index.ts
@@ -28,6 +28,7 @@ import Interactions, { InteractionsClass } from '@aws-amplify/interactions';
 import * as UI from '@aws-amplify/ui';
 import XR, { XRClass } from '@aws-amplify/xr';
 import Predictions from '@aws-amplify/predictions';
+import { DataStore } from '@aws-amplify/datastore';
 
 import Amplify, {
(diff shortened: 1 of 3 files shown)


packages/aws-amplify/src/index.ts
```diff
@@ -54,6 +55,7 @@ Amplify.Interactions = Interactions;
 Amplify.UI = UI;
 Amplify.XR = XR;
 Amplify.Predictions = Predictions;
+Amplify.DataStore = DataStore;
```
Amplifiyer: This is going away in the next release. Probably shouldn't add just now?

Ashish-Nanda: Wouldn't it be better to have DataStore be something you import from `aws-amplify`, given all other categories in V2 support that. It should have been this way since DataStore was released and we should release this in V2 with other bug fixes.

In V3 this will go away for all categories, which is fine.

Amplifiyer: makes sense since we intend to support V2 for a while.

PR #5212, opened 2020-03-27
fix(@aws-amplify/core): retry logging showing [object Object]

polymath-shpetim:
If sync fails for any reason, retry util was not logging error properly, javascript object was being logged as `error on : [object Object]`


packages/core/src/Util/Retry.ts
```diff
@@ -36,7 +36,7 @@ export async function retry(
 	try {
 		return await functionToRetry(...args);
 	} catch (err) {
-		logger.debug(`error on ${functionToRetry.name}: ${err} `);
+		logger.debug(`error on ${functionToRetry.name}: ${JSON.stringify(err)} `);
```
ericclemmons: Thanks for the PR!  Have you been able to validate this?

In my experience, `JSON.stringify(err)` usually returns `{}`:

```js
JSON.stringify(new Error("Message"))
// "{}"
```

Whereas this is usually what I have to do:

```suggestion
		logger.debug(`error on ${functionToRetry.name}: ${err.message} `);
```

polymath-shpetim: Thanks for feedback, what is happening here is not an error object, it's returning graphql query object that failed, if you check this file in line 33 it's already handled this way. err.message in this case will return undefined

jordanranz:
Hey @polymath-shpetim. Thank you for contributing this fix!

Since our `logger` accepts varargs we can just pass `err` as a parameter instead of having to use `JSON.stringify` inline. I [submitted a PR here](https://github.com/aws-amplify/amplify-js/pull/5699) to resolve the core issue.

PR #5216, opened 2020-03-27
fix(@aws-amplify/ui-components): Default federated, since aws-exports#oauth gets merged with it

ericclemmons:
`federated` gets merged with `oauth`, but when `undefined` it throws run-time errors unless a user manually adds `<amplify-authenticator federated={...} />`.

https://github.com/aws-amplify/amplify-js/blob/778a39a11b623ed28585fb222b618ec0e4189f95/packages/amplify-ui-components/src/components/amplify-federated-buttons/amplify-federated-buttons.tsx#L27-L35

---


iartemiev:
image
 This PR fixes that ^ error

PR #5234, opened 2020-03-30
Upgrade aws-sdk beta clients

Amplifiyer:
_Issue #, if available:_

_Description of changes:_ Fixes Rekognition, Textract, Kinesis and Storage aws-sdk bugs.


packages/analytics/package.json
```diff
@@ -39,10 +39,10 @@
   "dependencies": {
     "@aws-amplify/cache": "^2.1.7",
     "@aws-amplify/core": "^2.2.6",
-    "@aws-sdk/client-firehose": "1.0.0-beta.1",
-    "@aws-sdk/client-kinesis": "1.0.0-beta.1",
-    "@aws-sdk/client-personalize-events": "1.0.0-beta.1",
-    "@aws-sdk/client-pinpoint": "1.0.0-beta.1",
+    "@aws-sdk/client-firehose": "1.0.0-beta.3",
```
sammartinez: Future thinking, are we going to look to just have a `^` for newer versions? I understand the need for it now but was seeing what we are thinking on this?

Amplifiyer: yes, going forward we do want loose version numbers to avoid having duplicate package versions in client's app. but currently we can't risk it.

sammartinez: Agreed

PR #5235, opened 2020-03-30
Removing the need to install NetInfo in ReactNative Apps when not using DataStore

Ashish-Nanda:
_Issue #, if available:_

_Description of changes:_
This PR removes the need to install NetInfo in ReactNative Apps when not using DataStore

Since Reachability is exported by default, it also requires NetInfo to be installed whenever we import something from '@aws-amplify/core'. This means even in simple apps that have nothing to do with network state, this dependency needs to be installed and linked by customers, which leads to friction in setup and development.

The Reachability code for React Native now requires the consumer to explicitly pass NetInfo to networkMonitor, and hence no longer needs the NetInfo dependency to be installed when only using core.

NetInfo still needs to be a peer dependency and installed by the user because of current limitations in the workflow for native dependencies for React Native libraries: https://github.com/react-native-community/cli/issues/870


packages/datastore/src/sync/datastoreReachability/index.native.ts
```diff
@@ -0,0 +1,3 @@
+import Reachability from '@aws-amplify/core/lib/Util/ReactNative/Reachability';
```
Amplifiyer: why are we using `lib` built paths?

manueliglesias: I'd like to see @Amplifiyer 's take on this style of import.

Shouldn't be a problem for RN, I think, but would like confirmation

Ashish-Nanda: Updated

packages/datastore/src/sync/datastoreReachability/index.native.ts
```diff
@@ -0,0 +1,5 @@
+import { Reachability } from '@aws-amplify/core';
+import { default as NetInfo } from '@react-native-community/netinfo';
+
+const ReachabilityMonitor = new Reachability().networkMonitor(NetInfo);
+export default ReachabilityMonitor;
```
Amplifiyer, line 5: we should use named exports when possible.

packages/core/src/Util/Reachability.native.ts
```diff
@@ -9,13 +8,18 @@ type NetworkStatus = {
 };
 
 export default class ReachabilityNavigator implements Reachability {
-	networkMonitor(): Observable<NetworkStatus> {
+	networkMonitor(netInfo?: any): Observable<NetworkStatus> {
+		if (!netInfo) {
```
Amplifiyer, line 12: I think it's a good idea to check the presence of netinfo.fetch() as users could pass anything and later fail on cryptic errors as we don't have type safety here.

packages/core/src/Util/Reachability.native.ts
```diff
@@ -45,5 +49,5 @@ export default class ReachabilityNavigator implements Reachability {
 }
 
 interface Reachability {
-	networkMonitor(): Observable<NetworkStatus>;
+	networkMonitor(netInfo?: any): Observable<NetworkStatus>;
```
Amplifiyer, line 52: We should probably add some comments here in the interface as what netInfo is and when it's required (or it's purpose), again since because it's not self documenting due to missing types.

Ashish-Nanda: Since the interface is same as web, where netinfo is optional, I added comments above where the check is done in `networkMonitor`.

packages/core/src/Util/Reachability.native.ts
Ashish-Nanda: Done

packages/datastore/src/sync/datastoreReachability/index.native.ts
Ashish-Nanda: Done

PR #5240, opened 2020-03-31
UI Components into Master

jordanranz:
_Description of changes:_
- UI Components into modularization branch


packages/amplify-ui-angular/tsconfig.legacy.json
```diff
@@ -0,0 +1,8 @@
+{
```
ashika01: Do we need this legacy config going forth?

sammartinez: This is auto generated right?

packages/amplify-ui-components/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
sammartinez: I believe we need to change this to Apache License as this is what we use in Amplify

packages/amplify-ui-components/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "@aws-amplify/auth": "^2.1.5",
+    "@aws-amplify/core": "^2.2.4",
+    "@aws-amplify/xr": "^1.1.5"
+  },
+  "husky": {
+    "hooks": {
+      "pre-push": "npm run test"
+    }
+  },
+  "license": "Apache-2.0"
```
sammartinez: interesting.

ashika01:
@jordanranz I think we should leave out the `amplify-examples` folder out.

packages/amplify-ui-angular/tsconfig.legacy.json
jordanranz: @ericclemmons

packages/amplify-ui-components/LICENSE
jordanranz: Changing

.gitignore
```diff
@@ -17,3 +17,4 @@ packages/**/cjs/
 .npm/
 packages/**/cypress/videos/
 package-lock.json
+yarn.lock
```
ericclemmons: `yarn.lock` is removed in https://github.com/aws-amplify/amplify-js/pull/5243/files#diff-a084b794bc0759e7a6b77810e01874f2L18, so this may be out of sync.

The consensus was that `yarn.lock` shouldn't exist so that we catch regressions of transitive dependencies _sooner_, before customers do.

Having `yarn.lock` ignored may result in contributors having it locally while CI doesn't, with different behaviors.

package.json
```diff
@@ -23,7 +23,8 @@
 		"publish:beta": "lerna publish --canary --force-publish \"*\" --yes --dist-tag=beta --preid=beta --exact",
 		"publish:release": "lerna publish --conventional-commits --yes --message 'chore(release): Publish [ci skip]'",
 		"publish:1.0-stable": "lerna publish --conventional-commits --yes --dist-tag=stable-1.0 --message 'chore(release): Publish [ci skip]'",
-		"publish:modularization": "lerna publish --canary --force-publish \"*\" --yes --dist-tag=preview --preid=preview --exact"
+		"publish:modularization": "lerna publish --canary --force-publish \"*\" --yes --dist-tag=preview --preid=preview --exact",
+		"publish:ui-components/master": "lerna publish --canary --force-publish \"*\" --dist-tag=ui-preview --preid=ui-preview --exact"
```
ericclemmons: This can be removed as part of this merge, correct?

.circleci/config.yml
```diff
@@ -413,6 +413,7 @@ workflows:
               only:
                 - release
                 - master
+                - ui-components/master
```
ericclemmons: This can be removed as part of this merge, correct? (Since this branch is "done")

packages/amazon-cognito-identity-js/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "amazon-cognito-identity-js",
   "description": "Amazon Cognito Identity Provider JavaScript SDK",
-  "version": "4.0.0",
+  "version": "4.1.0",
```
ericclemmons: How did this get to `4.1.0` where modularization is at `4.0.0`?
> https://github.com/aws-amplify/amplify-js/pull/5243/files#diff-d1b10c0f795c7e3e1a1e9bcc8db4b76cR4

packages/amplify-ui-angular/tsconfig.legacy.json
ericclemmons: This came from https://github.com/ionic-team/stencil-ds-plugins-demo/blob/master/packages/component-library-angular/tsconfig.legacy.json and is referenced here:
> https://github.com/ionic-team/stencil-ds-plugins-demo/blob/master/packages/component-library-angular/package.json#L16

I'm unsure to what degree we can remove this, unless we refactor it to use our internal `./scripts/build`.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -0,0 +1,187 @@
...
+  @Prop() federated: FederatedConfig;
+  /** Username Alias is used to setup authentication with `username`, `email` or `phone_number`  */
+  @Prop() usernameAlias: UsernameAliasStrings;
+
+  @State() authState: AuthState = AuthState.Loading;
+  @State() authData: CognitoUserInterface;
+  @State() toastMessage: string = '';
+
+  async componentWillLoad() {
+    Hub.listen(UI_AUTH_CHANNEL, data => {
```
ericclemmons: I'll be submitting a PR to listen to the `auth` channel. It's necessary for hosted UI redirects to complete.

package.json
ashika01: I thought about this too, but if we want to test some feature development specific to UI, you could now do it in `ui-preview` before release?

packages/amazon-cognito-identity-js/package.json
ashika01: could be because of all user-agent stuff that is added?

ashika01: Scratch that, you are right, it should be 4.0.0 and then the release should update it to a patch version or something

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
ericclemmons: Resolved via https://github.com/aws-amplify/amplify-js/pull/5244

package.json
sammartinez: At the moment there is no risk to keep this here, we can circle back as we need to do the same for others such as 1.0-stable command

.gitignore
jordanranz: So should we remove `yarn.lock` from the gitignore?

.circleci/config.yml
jordanranz: yep, I will remove these.

package.json
jordanranz: I can remove this.

packages/amazon-cognito-identity-js/package.json
jordanranz: yea somehow it was at 4.0.0, then the user-agent stuff was technically deployed and maybe updated this?

.circleci/config.yml
ericclemmons: Follow-up PR

packages/auth/src/types/Auth.ts
```diff
@@ -119,6 +119,7 @@ export interface FederatedResponse {
 export interface FederatedUser {
 	name: string;
 	email?: string;
+	picture?: string;
```
Amplifiyer, line 122: Is this change related here?

jordanranz: Yes we use `picture` in the federated UI

PR #5243, opened 2020-03-31
Modularization FTW

Amplifiyer:
_Issue #, if available:_#3365

_Description of changes:_ Merge all modularization changes in master


package.json
```diff
@@ -19,7 +19,7 @@
 		"lint": "lerna run lint",
 		"link-all": "yarn unlink-all && lerna exec --no-bail --parallel yarn link",
 		"unlink-all": "lerna exec --no-bail --parallel -- yarn unlink; exit 0",
-		"publish:master": "lerna publish --canary --force-publish \"*\" --yes --dist-tag=unstable --preid=unstable --exact",
+		"publish:master": "lerna publish --conventional-commits --force-publish \"*\" --yes --dist-tag=unstable --preid=unstable --exact",
```
sammartinez: So we don't need `from-package` now?

ashika01: yup. I posted a quick summary on chime.

Amplifiyer: Correct, conventional commits are enough(we were only testing with canary). The only risk is conflicting github releases from master and release

.circleci/config.yml
```diff
@@ -42,9 +42,9 @@ jobs:
     steps:
       - checkout
       - run: yarn config set workspaces-experimental true
-      - run: yarn
-      - run: yarn bootstrap
-      - run: yarn build
+      - run: yarn --frozen-lockfile
+      - run: yarn run bootstrap
+      - run: yarn run build
```
ashika01: I think this could just be `yarn` instead of `yarn run` as well

.gitignore
```diff
@@ -8,11 +8,12 @@ packages/**/lib/
...
+packages/**/esm/
+packages/**/cjs/
 .DS_Store
 .vscode
 .idea
 *.log
 .npm/
 packages/**/cypress/videos/
 package-lock.json
-yarn.lock
```
ashika01: I think we still need to keep `yarn.lock` to git ignore

packages/api-graphql/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
```
sammartinez: Do we need anything in here?

packages/amazon-cognito-identity-js/android/.project
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jordanranz: Curious where these Android files came from?

Amplifiyer: I believe they are from running react native tests, Let me see if I can revert them quickly

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -657,18 +671,14 @@ export default class AWSPinpointProvider implements AnalyticsProvider {
...
-					req.httpRequest.headers['user-agent'] = Platform.userAgent;
-				}
-
-				// TODO: remove this after we switch to using the current API (non-legacy)
-				if (req.httpRequest.path.includes('/events')) {
-					req.httpRequest.path += '/legacy';
-				}
-			});
-		});
+		// TODO FIXME: Find a middleware to do this with AWS V3 SDK
```
jordanranz: Is this fixable?

packages/amazon-cognito-identity-js/android/.project
Amplifiyer: I'm trying to reset the file using `git checkout origin/master -- packages/amazon-cognito-identity-js/android/.project` but somehow git is complaining that it's not in git :O

packages/analytics/src/Providers/AWSPinpointProvider.ts
Amplifiyer: No with the new aws-sdk. They don't allow changing the user-agent header. Also the whole 'user-agent' field is getting deprecated so not sure if this is something we should even try to do. We should remove this comment though

packages/analytics/src/index.ts
```diff
@@ -11,131 +11,12 @@
...
+/**
+ * @deprecated use named import
+ */
 export default Analytics;
-export { AnalyticsProvider };
-export { AnalyticsClass };
+export { AnalyticsProvider, Analytics };
 export * from './Providers';
-
-const listener = capsule => {
```
jordanranz: Curious why these were moved?

Amplifiyer: To remove the side effects from the index files and move the logic to the class instantiation so it doesn't get tree shaked out.

phacks:
Congrats to everyone involved! I was really looking forward to this to reduce the JS payload for our users, I can only imagine the amount of work you folks poured into this. Bravo!

Artyom-Ganev:
Any doc for this feature?

Issue #5232, opened 2020-03-30
UI Components: Fed sign in with Hosted UI bug


iartemiev:
**Describe the bug**
Federated sign in using Hosted UI does not show user as logged in after getting redirected back to the app. Requires user to refresh browser twice before correct logged in state is shown. Reproduced with React and Angular 9.

**To Reproduce**

1. Configure federated sign in via Amplify CLI (issue occurs with Sign in with Amazon, fb, etc.)

2. Start app with the following code in App.js:
```jsx
<AmplifyAuthenticator federated={{}}>
  <div className="App">Main
  <AmplifySignOut />
  </div>
</AmplifyAuthenticator>
```

3. Use the **Sign in with AWS** button to sign in with Cognito Hosted UI

4. Use social provider button to sign in

5. Hosted UI redirects back to app

6. App still showing log in form. The following is logged to the console:

![image](https://user-images.githubusercontent.com/29709626/77949124-71e85800-7294-11ea-9ec6-8c8913277214.png)

7. Refreshing the page **twice** results in the app showing the user as logged in successfully.

**Expected behavior**

Expected behavior is getting signed in without having to refresh the page.

<details>
  <summary><strong>Environment</strong></summary>

```
npx envinfo --system --binaries --browsers --npmPackages --npmGlobalPackages

  System:
    OS: macOS 10.15.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 37.12 GB / 64.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.16.1 - /usr/local/bin/node
    Yarn: 1.22.4 - ~/.yarn/bin/yarn
    npm: 6.13.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 80.0.3987.149
    Firefox: 74.0
    Safari: 13.1
  npmPackages:
    @aws-amplify/ui-react: ui-preview => 0.0.2-ui-preview.5545+6aae9f44 
    aws-amplify: ui-preview => 1.3.1-ui-preview.40+6aae9f44 
    react: ^16.8.6 => 16.13.1 
    react-dom: ^16.8.6 => 16.13.1 
    react-scripts: 2.1.8 => 2.1.8 
  npmGlobalPackages:
    @aws-amplify/cli: 4.17.2
    cordova: 9.0.0
    detox-cli: 10.0.7
    expo-cli: 3.15.5
[...]

Issue #3841, opened 2019-08-13
Auth: Delay of Hub events using federatedSignIn (part 2)


tinymarsracing:
My original issue #3214 has been closed, but I believe the problem has not been fixed by the linked PR. I cannot re-open my old issue, so I open this new issue. If it is based on a misunderstanding on my side, I would appreciate an explanation in the original issue.

bhill9270:
It looks like version 1.1.36 of aws-amplify. But the fix is only for React it appears. I'm getting the same error situation in Angular.

jordanranz:
Hey @tinymarsracing, sorry that the PR did not fix your issue. Can you post which version of `aws-amplify-react` you are using for a bit more context?

bhill9270:
For anyone else out there having this issue, my workaround is testing the localstorage for the right tokens before initializing the Auth.currentAuthenticatedUser() method.

ErickMwazonga:
> My original issue #3214 has been closed, but I believe the problem has not been fixed by the linked PR. I cannot re-open my old issue, so I open this new issue. If it is based on a misunderstanding on my side, I would appreciate an explanation in the original issue.

++

ericclemmons:
I was able to reproduce this when going through the following flow:

1. Visit <http://localhost:3000/> unauthenticated.
1. Enable debugging: `window.LOG_LEVEL = 'DEBUG';`. (I also added debugging to a custom `useAuth` hook I have).
1. "Sign in", which redirects to the federated hosted UI.
1. Get redirected back to <http://localhost:3000/?code=...>.
1. URL gets replaced with <http://localhost:3000/>

When redirected back to <http://localhost:3000/>, I captured the following logs:

```shell
The development server has disconnected.
Refresh the page if necessary. webpackHotDevClient.js:65
[DEBUG] 36:34.304 Amplify - amplify config 
Object { ...SNIP... }
ConsoleLogger.js:111
[DEBUG] 36:34.320 I18n - configure I18n ConsoleLogger.js:101
[DEBUG] 36:34.321 I18n - create I18n instance ConsoleLogger.js:101
[DEBUG] 36:34.321 I18n - en-US ConsoleLogger.js:101
[DEBUG] 36:34.321 AnalyticsClass - configure Analytics 
Object { ...SNIP... }
ConsoleLogger.js:111
[DEBUG] 36:34.324 Parser - parse config 
Array(3) [ {…}, "to amplifyconfig", {…} ]
[...]

PR #5244, opened 2020-03-31
fix(@aws-amplify/ui-components): Listen to auth channel for managing Hosted UI auth state

ericclemmons:
_Issue #, if available:_ #5232

_Description of changes:_

I've been able to reproduce Ivan's hosted UI bug.  Found several ancillary issues to be resolved: https://github.com/aws-amplify/amplify-js/issues/5232#issuecomment-606345994

This PR brings over logic from `aws-amplify-react` that was missing in `amplify-authenticator`:

https://github.com/aws-amplify/amplify-js/blob/38e55b090e10b71251439344365f6aa9344d049a/packages/aws-amplify-react/src/Auth/Authenticator.tsx#L70-L79

> hosted-ui

It's worth noting that the `Auth` category still returns lots of errors (which isn't specific to the components, but is in `master`):

> Screen Shot 2020-03-31 at 1 27 18 PM

It's worth noting that there have been consistent issues with hosted UI & redirecting:
> https://github.com/aws-amplify/amplify-js/issues/3841#issuecomment-543338595

---

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
@@ -1,6 +1,7 @@
 import { Component, State, Prop, h, Host } from '@stencil/core';
 import { AuthState, CognitoUserInterface, FederatedConfig, UsernameAliasStrings } from '../../common/types/auth-types';
 import {
+  AUTH_CHANNEL,
   NO_AUTH_MODULE_FOUND,
   SIGNING_IN_WITH_HOSTEDUI_KEY,
   AUTHENTICATOR_AUTHSTATE,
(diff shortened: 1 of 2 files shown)


packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -30,6 +31,19 @@ export class AmplifyAuthenticator {
...
   async componentWillLoad() {
+    Hub.listen(AUTH_CHANNEL, ({ payload: { event, data } }) => {
+      switch (event) {
+        case 'cognitoHostedUI':
+          return this.onAuthStateChange(AuthState.SignedIn, data);
+
+        case 'cognitoHostedUI_failure':
+        case 'parsingUrl_failure':
+        case 'signOut':
+        case 'customGreetingSignOut':
```
ashika01: Why do we have these two cases here? Wouldnt we just listen to the one within UI component for that?

ericclemmons: I took these events from:
https://github.com/aws-amplify/amplify-js/blob/38e55b090e10b71251439344365f6aa9344d049a/packages/aws-amplify-react/src/Auth/Authenticator.tsx#L137-L160

Where are you recommending?

ashika01: So I mean, you are `signOut` from `AUTH_CHANNEL` right, is it even needed? bcoz, **dispatchAuthChangeEvent** will dispatch an event for _signout_ and right now we are already listening to that in `UI_AUTH_CHANNEL`. Shouldnt that be enough?

jordanranz: If a user called `Auth.signOut` outside of the context of the `authenticator` (aka not using the sign out component) this change would pick up that context I believe.

ashika01: yeah eric just pinged me too. 👍 makes sense

PR #5265, opened 2020-04-01
chore(@aws-amplify/storage): Test responseType for iOS blob bug

ericclemmons:
_Issue #, if available:_ #5260 

- [x] Tests for `responseType`, which is the difference between Native and Web:

     https://github.com/ericclemmons/amplify-js/blob/5b1d2e9cd56dbb6cf7ed43f26126e24a98bcca5c/packages/storage/src/providers/httpHandlerOptions.native.ts#L1-L3

- [x] Change from `axios` to `axios.request` to enable mocking:

    > https://github.com/axios/axios#request-method-aliases

Delivers: #172112765

---


packages/storage/src/providers/axios-http-handler.ts
```diff
@@ -69,7 +69,7 @@ export class AxiosHttpHandler implements HttpHandler {
 		}
 
 		const raceOfPromises = [
-			axios(axiosRequest).then(response => {
+			axios.request(axiosRequest).then(response => {
```
sammartinez, line 72: Just for clarity, wouldn't axios still resolve to request here?

ericclemmons: I'm not sure I understand. Can you rephrase that?

In axios' docs, `axios` and `axios.request` are equivalent.  The problem I had was that it wasn't possible to mock the `axios` call from `import axios from "axios"`, so I had to mock `axios.[something]`:
> https://github.com/axios/axios#request-method-aliases

bneigher:
perhaps related? https://github.com/aws-amplify/amplify-js/issues/5267

ericclemmons:
#5267 looks specific to React Native Debugger, but if there are any code changes to `blob` support we'll update tests accordingly.

Issue #5293, opened 2020-04-03
Server-side rendering not supported in @aws-amplify/ui-components and @aws-amplify/ui-react


sawyerh:
**Describe the bug**
`@aws-amplify/ui-components` and `@aws-amplify/ui-react` reference web-only interfaces like `window` and `document`. This breaks React apps that utilizes server-side rendering, since `window` and `document` are not defined in a Node.js environment.

**To Reproduce**

1. Clone https://github.com/sawyerh/amplify-nextjs-ssr
1. Run the app locally, following instructions in the `README`
1. Open the app and observe the compilation errors related to undefined `document` and `window` references

**Expected behavior**
Code that calls web-only APIs should be conditional on the API being defined, for example:

```js
if (typeof document !== 'undefined') {
  let customStyles = document.createElement('style');
```

**Screenshots**

```bash
ReferenceError: document is not defined
    at Object.<anonymous> (/Users/Dev/amplify-nextjs-ssr/node_modules/@aws-amplify/ui-components/dist/cjs/app-globals-dbb54dcb.js:15:20)
    at Module._compile (internal/modules/cjs/loader.js:956:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)
    at Module.load (internal/modules/cjs/loader.js:812:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Module.require (internal/modules/cjs/loader.js:849:19)
    at require (internal/modules/cjs/helpers.js:74:18)
```

<details>
  <summary><strong>Environment</strong></summary>

```
    System:
    OS: macOS 10.15.3
    CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
    Memory: 56.16 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.13.0 - ~/.nvm/versions/node/v12.13.0/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.13.7 - ~/.nvm/versions/node/v12.13.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 80.0.3987.163
    Firefox: 72.0.2
    Safari: 13.0.5
  npmPackages:
    @aws-amplify/ui-react: ^0.2.0 => 0.2.0
    aws-amplify: ^3.0.5 => 3.0.5
[...]

PR #5294, opened 2020-04-03
Check if document is undefined before using it, Fix phone field width

jordanranz:
_Description of changes:_
- Check for `document` existence

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
@@ -32,15 +32,9 @@ export class AmplifyButton {
         const formSection = this.el.closest('amplify-form-section');
         form = formSection && formSection.shadowRoot.querySelector('form');
       }
-      if (form && typeof document !== 'undefined') {
+      if (form) {
         ev.preventDefault();
-
-        const fakeButton = document.createElement('button');
-        fakeButton.type = this.type;
-        fakeButton.style.display = 'none';
-        form.appendChild(fakeButton);
-        fakeButton.click();
-        fakeButton.remove();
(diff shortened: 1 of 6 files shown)


packages/amplify-ui-components/src/global/theme.ts
```diff
@@ -11,47 +11,49 @@
...
-
-    --amplify-red: #dd3f5b;
-  }
-`),
-);
-
-let parentElement = document.getElementsByTagName('head')[0];
-const firstChild = parentElement.firstChild;
-parentElement.insertBefore(customStyles, firstChild);
+if (typeof document !== 'undefined') {
```
ericclemmons, line 14: Should we use `isBrowser` here?
https://github.com/aws-amplify/amplify-js/blob/2f27a6e2d002364af5c44b9f38542ce1e2c8c119/packages/core/src/JS.ts#L159-L171

ashika01: Yup we should do that.

jordanranz: done

packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
```diff
@@ -32,7 +32,7 @@ export class AmplifyButton {
         const formSection = this.el.closest('amplify-form-section');
         form = formSection && formSection.shadowRoot.querySelector('form');
       }
-      if (form) {
+      if (form && typeof document !== 'undefined') {
```
jordanranz, line 35: This needs re-thinking

sawyerh:
Related (but doesn't completely fix): #5293

Issue #5322, opened 2020-04-07
@aws-amplify/ui-react doesn't transpile ES6 to ES5


ericclemmons:
**Describe the bug**

_Originally discovered by @sawyerh: https://github.com/aws-amplify/amplify-js/issues/5282#issuecomment-608615158_

When using `@aws-amplify/ui-react` in a Next.js project, I get:

```console
node_modules/@aws-amplify/ui-react/dist/index.js:1
export * from './components'
^^^^^^

SyntaxError: Unexpected token 'export'
```

**To Reproduce**

1. `yarn create next-app`
2. Update `pages/index.tsx` with:

   ```tsx
	import { withAuthenticator } from '@aws-amplify/ui-react'
	```

3. http://localhost:3000/

**Expected behavior**

Work like `aws-amplify-react` does.

**Screenshots**

<img width="858" alt="Screen Shot 2020-04-07 at 3 01 57 PM" src="https://user-images.githubusercontent.com/15182/78723660-bef5ab00-78e0-11ea-9331-ce650190b518.png">

<details>
  <summary><strong>Environment</strong></summary>

```
  System:
    OS: macOS Mojave 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i7-7660U CPU @ 2.50GHz
    Memory: 234.59 MB / 16.00 GB
    Shell: 5.7.1 - /usr/local/bin/zsh
  Binaries:
    Node: 10.16.3 - ~/.nvm/versions/node/v10.16.3/bin/node
    Yarn: 1.22.0 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.3/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 80.0.3987.149
    Firefox: 68.6.0
    Safari: 13.1
  npmPackages:
    @aws-amplify/ui-react: unstable => 0.2.2-unstable.3+bb17bcb83
    @types/node: ^13.9.1 => 13.11.0
    @types/react: ^16.9.23 => 16.9.32
    aws-amplify: unstable => 3.0.7-unstable.3+bb17bcb83
    next: latest => 9.3.4
    react: latest => 16.13.1
    react-dom: latest => 16.13.1
    typescript: ^3.8.3 => 3.8.3
  npmGlobalPackages:
    @aws-amplify/cli: 4.13.3
    create-react-app: 3.1.1
    npm: 6.9.0
[...]

PR #5324, opened 2020-04-08
fix(@aws-amplify/ui-react): Export CJS

ericclemmons:
_Issue #, if available:_ #5322 

Stencil's demo project didn't transpile CJS. This PR re-uses much of `aws-amplify-react`.

```console
node_modules/@aws-amplify/ui-react/dist/index.js:1
export * from './components'
^^^^^^

SyntaxError: Unexpected token 'export'
```

**See commits for history**.

---


scripts/build.js
```diff
@@ -188,9 +193,14 @@ async function buildES5(typeScriptCompiler) {
...
 function buildES6(typeScriptCompiler) {
-	const jsx = packageInfo.name === 'aws-amplify-react' ? 'react' : undefined;
+	const jsx = ['@aws-amplify/ui-react', 'aws-amplify-react'].includes(
+		packageInfo.name
+	)
+		? 'react'
+		: undefined;
 	// tsconfig for ESM generating
 	let compilerOptions = {
+		esModuleInterop: true,
```
ericclemmons, line 203: Without this, `import React from "react"` results in:

> `Cannot read property 'Component' of undefined`

See: https://github.com/kulshekhar/ts-jest/issues/146#issuecomment-442871767

scripts/build.js
```diff
@@ -148,9 +148,14 @@ function reportWatchStatusChanged(diagnostic, newLine, options, errorCount) {
 }
 
 async function buildES5(typeScriptCompiler) {
-	const jsx = packageInfo.name === 'aws-amplify-react' ? 'react' : undefined;
+	const jsx = ['@aws-amplify/ui-react', 'aws-amplify-react'].includes(
```
sammartinez, line 151: Do we have `includes` in our library? I believe we are using find but just want to make sure there are no build errors with using this.

ericclemmons: [`Array.prototype.includes` has been in since Node 6](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes), and it works locally & in CI.

If there's another aspect I'm unaware of, I'm happy to swap it to `.find`.

sammartinez: `.find` is what we use now, I believe it was something on the polyfill piece of it. @manueliglesias Do you have any clarity on this?

sammartinez: AS per our offline, I am fine. Approving

Issue #5325, opened 2020-04-08
TypeError: form.requestSubmit is not a function


ericclemmons:
**Describe the bug**

Introduced in ce74563b427cb2713ad0f1a778441b8f1f3f3ac5, `form.requestSubmit` does not work in Firefox (current version is **68**):

- https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit (Will land in _future_ Firefox **75**)
- https://www.stefanjudis.com/today-i-learned/requestsubmit-offers-a-way-to-validate-a-form-before-submitting-it/#a-new-method-that-does-what-developers-expect

**To Reproduce**

1. Add `<amplify-authenticator />` to an app.
2. Click "Sign In"
3. See error:

    > <img width="1034" alt="Screen Shot 2020-04-07 at 5 11 26 PM" src="https://user-images.githubusercontent.com/15182/78730943-1b61c600-78f3-11ea-90fd-b4fa5ed7b372.png">

**Expected behavior**

Complete sign in.

```
  System:
    OS: macOS Mojave 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i7-7660U CPU @ 2.50GHz
    Memory: 104.03 MB / 16.00 GB
    Shell: 5.7.1 - /usr/local/bin/zsh
  Binaries:
    Node: 10.16.3 - ~/.nvm/versions/node/v10.16.3/bin/node
    Yarn: 1.22.0 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.3/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 81.0.4044.92
    Firefox: 68.6.0
    Safari: 13.1
  npmPackages:
    @aws-amplify/ui-react: unstable => 0.2.1
    @types/node: ^13.9.1 => 13.11.0
    @types/react: ^16.9.23 => 16.9.32
    aws-amplify: unstable => 3.0.7-unstable.3+bb17bcb83
    next: latest => 9.3.4
    react: latest => 16.13.1
    react-dom: latest => 16.13.1
    typescript: ^3.8.3 => 3.8.3
  npmGlobalPackages:
    @aws-amplify/cli: 4.13.3
    create-react-app: 3.1.1
    npm: 6.9.0
    pure-prompt: 1.10.3
```

</details>

**Additional context**

[...]

PR #5333, opened 2020-04-08
Revert formSubmit, add isBrowser check

jordanranz:
_Issue #, if available:_
https://github.com/aws-amplify/amplify-js/issues/5325

_Description of changes:_
- Revert back to simulate form submit with button
- Add browser check for server side rendering

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
@@ -1,6 +1,7 @@
 import { Element, Component, Prop, h } from '@stencil/core';
 import { ButtonTypes, ButtonVariant } from '../../common/types/ui-types';
 import { hasShadowDom } from '../../common/helpers';
+import { browserOrNode } from '@aws-amplify/core';
 
 @Component({
   tag: 'amplify-button',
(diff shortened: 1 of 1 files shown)


packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
```diff
@@ -32,9 +33,15 @@ export class AmplifyButton {
         const formSection = this.el.closest('amplify-form-section');
         form = formSection && formSection.shadowRoot.querySelector('form');
       }
-      if (form) {
+      if (form && browserOrNode().isBrowser) {
```
ericclemmons, line 36: What error happens here?  It’s interesting that an event handler would run during SSR.

packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
```diff
@@ -32,9 +33,15 @@ export class AmplifyButton {
...
         ev.preventDefault();
-        form.requestSubmit();
+
+        const fakeButton = document.createElement('button');
+        fakeButton.type = this.type;
+        fakeButton.style.display = 'none';
+        form.appendChild(fakeButton);
+        fakeButton.click();
```
ericclemmons, line 43: I’m ignorant on Shadow DOM details. If the sign in button was `<button type=submit>`, would an onSubmit handler on form work?

I assume the shadow DOM prevents that from bubbling?

packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
jordanranz: got rid of this :D

packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
jordanranz: Spoke offline about this.

PR #5338, opened 2020-04-08
fix(@aws-amplify/ui-react): Exclude files via .npmignore rather than files

ericclemmons:
`@aws-amplify/ui-react` whitelisted `files` instead of ignoring:

Screen Shot 2020-04-08 at 2 16 30 PM

`aws-amplify-react` uses `.npmignore`:

Screen Shot 2020-04-08 at 2 16 34 PM

---

### DIFF
--- a/packages/amplify-ui-react/package.json
+++ b/packages/amplify-ui-react/package.json
@@ -12,7 +12,7 @@
 	},
 	"scripts": {
 		"test": "jest",
-    "build-with-test": "npm test && npm run build",
+		"build-with-test": "npm test && npm run build",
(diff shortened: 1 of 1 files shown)


packages/amplify-ui-react/package.json
```diff
@@ -12,7 +12,7 @@
 	},
 	"scripts": {
 		"test": "jest",
-    "build-with-test": "npm test && npm run build",
+		"build-with-test": "npm test && npm run build",
```
ericclemmons, line 15: This came from GitHub's conflict resolution UI, as it doesn't run prettier.

Issue #5340, opened 2020-04-08
signUp method of SignUp class from aws-amplify-react return void


NastyZ98:
**Describe the bug**
Hi, I realized by extending the SignUp class that his `signUp()` method was not returning a promise ([source](https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react/src/Auth/SignUp.tsx#L206)), compared to the `SignIn()` method of the SignIn class, which return a promise ([source](https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react/src/Auth/SignIn.tsx#L78)).
In my case I wanted to extends those classes to create a custom components.

Where I encountered difficulties was when I tried to disable the sign up button depending on the request made by the parent class (in this case SignUp class), because the signUp method does not return any promise.
It was made possible to disable the signIn button on my child component, because the `signIn()` method of the parent component (SignIn class) return a promise.

**To Reproduce**
```ts
import { SignIn } from "aws-amplify-react";

class CustomSignIn extends SignIn {
  constructor(props: any) {
    super(props);
    this._validAuthStates = ["signIn", "signedOut", "signedUp"];
  }
  showComponent(theme: any) {
      // We can't know when the request is resolved
       return (<Button onClick={() => super.signUp()}>Sign Up</Button>);
  }
}
```
**Expected behavior**
Make signUp method return a promise

I opened a pull request to make the signUp method return a promise.

<details>
  <summary><strong>Environment</strong></summary>

```
  System:
    OS: macOS 10.15.4
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 296.80 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.16.1 - /usr/local/bin/node
    Yarn: 1.22.4 - ~/Documents/Workspace/aws-boilerplate-cognito/node_modules/.bin/yarn
    npm: 6.13.4 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 80.0.3987.149
    Safari: 13.1
  npmPackages:
    @bahmutov/add-typescript-to-cypress: ^2.1.2 => 2.1.2 
    @testing-library/jest-dom: ^4.2.4 => 4.2.4 
    @testing-library/react: ^9.3.2 => 9.5.0 
[...]

PR #5341, opened 2020-04-08
fix(@aws-amplify/aws-amplify-react): incorrect return value of signUp method from the SignUp class

NastyZ98:
_Issue https://github.com/aws-amplify/amplify-js/issues/5340_

_Description of changes:_

**Fixed the return value of the signUp method inside the SignUp class component**


NastyZ98:
Hi @jordanranz, thank you! 
I also have noticed, when I inherited Amplify components that we can’t pass new props because classes that we inherit are not of generic type.
When I will have time, I will open PRs to change these classes to generic type, what do you thin about it, maybe I have to open an issue about it ?

jordanranz:
@NastyZ98 we will be moving away from supporting the legacy UI components so this work would probably be unnecessary. Can you open up a new issue and provide details about your use case. This could be something we build an experience around in the new components.

PR #5342, opened 2020-04-09
Convert blobs to arrayBuffer in react native before uploading to S3 

### LINKED ISSUE #5311: Expo blob fails to upload with Storage.put
**Describe the bug**
A blob read using RN/expo fetch()/blob() is not supported by Storage.put resulting in a 403 error

**To Reproduce**

Use expo `expo-image-picker` to pick an image use `fetch()/blob()` to fetch the picked uri as a blob.
Pass the blob to be uploded with `Storage.put`

**Expected behavior**
Picked image should be uploaded

<details>
  <summary><strong>Environment</strong></summary>

```
  System:
    OS: Windows 10 10.0.18363
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 16.83 GB / 31.86 GB
  Binaries:
    Node: 12.16.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.16.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.9.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.18362.449.0
    Internet Explorer: 11.0.18362.1
  npmPackages:
    aws-amplify: ^3.0.5 => 3.0.5
    aws-amplify-react-native: ^4.0.2 => 4.0.2
    babel-preset-expo: ~8.1.0 => 8.1.0
    expo: ~37.0.4 => 37.0.4
    expo-image-picker: ~8.1.0 => 8.1.0
    expo-media-library: ~8.1.0 => 8.1.0
    expo-permissions: ~8.1.0 => 8.1.0
    expo-web-browser: ^8.1.0 => 8.1.0
    mime-types: ^2.1.26 => 2.1.26
    path: ^0.12.7 => 0.12.7
    react: ~16.9.0 => 16.9.0
    react-dom: ~16.9.0 => 16.9.0
    react-native: https://github.com/expo/react-native/archive/sdk-37.0.0.tar.gz => 0.61.4
    react-native-screens: ~2.2.0 => 2.2.0
    react-native-web: ~0.11.7 => 0.11.7
    typescript: ~3.8.3 => 3.8.3
```

</details>

- Device: Pixel 3a
- OS: Android 10
- Browser expo
- Version 37.0.4

**Additional context**

**Sample code**
```typescript
const pickerResult = await ImagePicker.launchImageLibraryAsync({
      exif: true
    });

    if (pickerResult.cancelled) {
      console.log("No image picked")
      return;
    }

     const response = await fetch(pickerResult.uri)
     const blob = await response.blob()
     await Storage.put(blob.data.name, blob, {
        level: 'private',
        contentType: blob.type,
        progressCallback: progress => {
          console.log(progress)
        }
      })
```

The issue is with https://github.com/aws-amplify/amplify-js/blob/b8dd7f9c256c464acc4d24f09970626feba7dbdf/packages/storage/src/providers/AWSS3ProviderManagedUpload.ts#L280
https://github.com/aws-amplify/amplify-js/blob/b8dd7f9c256c464acc4d24f09970626feba7dbdf/packages/storage/src/providers/AWSS3ProviderManagedUpload.ts#L296

It does not check for blob and treats it as another JS object and does a `JSON.stringify` and converts it to a string that is then read into a buffer as a UTF-8 string. Instead of a file/blob, it gets read as a string like 

`{"_data":{"lastModified":0,"name":"f360774e-f008-43cf-b6b1-4416641d1fb0.jpg","type":"image/jpeg","size":165090,"offset":0,"blobId":"6caaaf05-5b9a-4d54-883f-bc8da6bda951","__collector":{}}}`

This string is tried to be uploaded and mismatches the content-type (image/jpeg) resulting in a 403. Uploading a simple string works.

No significant logs are shown with LOG_LEVEL='DEBUG'

Amplifiyer:
_Issue #, if available:_ fixes #5311 

_Description of changes:_

There seems to be the issue with axios checking the request body for blob here https://github.com/axios/axios/blob/master/lib/utils.js#L134-L136

In react native environment, the blob object results in `"[object Object]"` rather than what axios expects, i.e. `"[object Blob]"` and that results in JSONification of the blob object before uploading https://github.com/axios/axios/blob/master/lib/defaults.js#L50-L53 (Confirmed in debugger by stepping through axios code)

To confirm my theory, I ran the following simple commands(to mimic exactly what's happening in axios) in a react native environment (debugger console) and simple browser console.

```js
var obj = "{hello: 'world'};"
var blob = new Blob([obj], {type : 'application/json'});
var toString = Object.prototype.toString;
toString.call(blob);
```

and the output

### React Native
Screen Shot 2020-04-07 at 7 31 03 PM

### Browser
Screen Shot 2020-04-07 at 7 32 15 PM

The change in this PR is to convert blobs to arrayBuffer (for RN only) before uploading which is correctly handled by the axios.

Also did some cleanups and error reporting


packages/storage/src/providers/AWSS3ProviderManagedUpload.ts
```diff
@@ -28,10 +28,10 @@ import {
 	AbortMultipartUploadCommand,
 } from '@aws-sdk/client-s3';
 import { AxiosHttpHandler, SEND_PROGRESS_EVENT } from './axios-http-handler';
-import { fromString } from '@aws-sdk/util-buffer-from';
 import * as events from 'events';
 import { parseUrl } from '@aws-sdk/url-parser-node';
 import { httpHandlerOptions } from './httpHandlerOptions.native';
```
Ashish-Nanda, line 33: Not related to your change but you should not have to provide `.native` in the file path. Change to:
```
import { httpHandlerOptions } from './httpHandlerOptions;
```

Issue #5371, opened 2020-04-11
Require cycles in amplify libs


andreialecu:
**Describe the bug**

We've noticed the following warnings being logged:

![image](https://user-images.githubusercontent.com/697707/79049672-b1685a00-7c2d-11ea-8bb0-ca653353052d.png)

It seems that under certain circumstances, this can lead to this problem: https://github.com/aws-amplify/amplify-js/issues/4859

**Expected behavior**

No require cycles should exist.

andreialecu:
This specific line of code can result in a lot of problems:

https://github.com/aws-amplify/amplify-js/blob/c28851a6379f3ba05c4fef529bb5be7f18ebe316/packages/predictions/src/types/Providers/AbstractConvertPredictionsProvider.ts#L12

If the babel config is set up to allow path aliases, or a different `module-resolver` root is configured (see https://github.com/aws-amplify/amplify-js/issues/4859#issuecomment-612462602) the wrong module will be imported.

I assume there are other scenarios where things could blow up because of require cycles, as seen from the reports in #4859 and #4687

lucianomlima:
Same here!

yumsquare:
Same warnings here as well.

Issue #4687, opened 2020-01-09
after update aws-amplify "Object prototype may only be an Object or null: undefined" issue


ulpsilva:
i update my react project aws-amplify to version 2.2.1 from version 0.3.4 and after i am getting this error
![Annotation 2020-01-09 104414](https://user-images.githubusercontent.com/6129644/72039982-0d3ce300-32cd-11ea-93f7-9358c5411b9c.png)

iartemiev:
Hi @ulpsilva, please provide a code sample of how you're using Amplify, as well as your config (aws-exports.js file with any sensitive information removed)

kaybee99:
I'm facing the same issue.

Managed to get it working by only importing the modules I needed. I'm using Vue, but I've basically done [this](https://stackoverflow.com/questions/53925177/aws-amplify-modular-imports-with-vue)

ulpsilva:
> Hi @ulpsilva, please provide a code sample of how you're using Amplify, as well as your config (aws-exports.js file with any sensitive information removed)

this is my aws-export.js
`const awsExports = {
  Auth: {
    // REQUIRED - Amazon Cognito Identity Pool ID
    identityPoolId: config.identityPoolId,
    // REQUIRED - Amazon Cognito Region
    region: config.region,
    // OPTIONAL - Amazon Cognito User Pool ID
    userPoolId: config.userPoolId,
    // OPTIONAL - Amazon Cognito Web Client ID
    userPoolWebClientId: config.userPoolWebClientId,
  },
  API: {
    endpoints: [
      {
        name: 'ConsultantApi',
        endpoint: config.ConsultantApi,

        service: 'lambda',
        region: config.region,
      },
    ],
  },
  Storage: {
    bucket: config.bucket, // REQUIRED -  Amazon S3 bucket
    region: config.region, // OPTIONAL -  Amazon service region
    identityPoolId: 'xxxxxxxxxxxx',
    level: 'public',
  },
};
export default awsExports;`

this is how i am using it
`import React, { Component } from 'react';
import { Col } from 'reactstrap';
import { Redirect } from 'react-router-dom';
import Amplify from '@aws-amplify/core';
import { withAuthenticator, VerifyContact, RequireNewPassword } from 'aws-amplify-react';

// aws cognito configuration
import awsExports from '../../../../aws_exports';

// Custom Sign Up Ui
import CustomSignIn from '../../../containers/AuthenticationFormContainer/CustomSignInContainer';
[...]

PR #5372, opened 2020-04-11
fix: require cycles in various packages

### LINKED ISSUE #4859: Object prototype may only be an Object or null
After upgrading aws-amplify on a react native expo project i got the following error:
"TypeError: Prototype value can only be an object or null"

andreialecu:
Fixes #4859 #5371 #4687

_Description of changes:_

There were various require cycles in the code base resulting in warnings and errors listed in the issues linked above.

Additionally, some exports were imported from `"."` which can cause additional problems with certain babel and packager setups (see https://github.com/aws-amplify/amplify-js/issues/5371#issuecomment-612474990, and https://github.com/aws-amplify/amplify-js/issues/4859#issuecomment-612460339)


andreialecu:
In order to prevent this in the future, it might make sense implementing this eslint rule:

https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-cycle.md

The project seems to be using `tslint` though, which is being deprecated. So a migration to `eslint` would be necessary first.

packages/predictions/src/Providers/AmazonAIPredictionsProvider.ts
```diff
@@ -3,7 +3,7 @@ import {
 	AmazonAIConvertPredictionsProvider,
 	AmazonAIInterpretPredictionsProvider,
 	AmazonAIIdentifyPredictionsProvider,
-} from '.';
+} from './index';
```
Amplifiyer, line 6: Will this not create a cycle as well since `index.ts` imports from `AmazonAIPredictionsProvider.ts` as well

andreialecu: I fixed these too

Issue #5173, opened 2020-03-23
Sync processor does not follow @model subscriptions level


onlybakam:
**Describe the bug**
Give this type definition:

```graphql
type Comment
  @model(subscriptions: { level: public })
  @auth(rules: [{ allow: owner, operations: [create, update, delete] }]) {
  id: ID!
  owner: ID!
  postId: ID!
  createdAt: AWSDateTime
  content: String!
  author: User @connection(fields: ["owner"])
}
```

the graphql transform generates this subscription:

```graphql
type Subscription {
  onCreateComment: Comment @aws_subscribe(mutations: ["createComment"])
  onUpdateComment: Comment @aws_subscribe(mutations: ["updateComment"])
  onDeleteComment: Comment @aws_subscribe(mutations: ["deleteComment"])
}
```

however the JS datastore implementation does not follow the convention and tries to set up a subscription filtered on the owner. This causes the subscription to return an error and breaks datastore.

**To Reproduce**
generate schema as shown above

**Expected behavior**
datastore should respect public configuration for the subscription.
to address, following could be done:

in packages/datastore/src/sync/utils.ts

```typescript
                if (isOperationAuthorized) {
-                       const rule = {
+                       const rule: AuthorizationRule = {
                                identityClaim,
                                ownerField,
                                provider,
                                groupClaim,
                                authStrategy,
                                groups,
+                               areSubscriptionsPublic: false,
                        };

+                       if (authStrategy === 'owner' || authStrategy === 'groups') {
+                               // look for the subscription level override
+                               // only pay attention to the public level
+                               const modelConfig = []
+                                       .concat(modelDefinition.attributes)
+                                       .find(attr => attr && attr.type === 'model');
+
+                               // find the subscriptions level. ON is defeault
+                               const { properties: { subscriptions: { level = 'on' } = {} } = {} } =
+                                       modelConfig || {};
+
+                               rule.areSubscriptionsPublic = level === 'public';
+                       }
+
```

in packages/datastore/src/sync/processors/subscription.ts,

```typescript
                        if (ownerValue) {
                                result = {
                                        authMode: GRAPHQL_AUTH_MODE.AMAZON_COGNITO_USER_POOLS,
-                                       isOwner: true,
+                                       isOwner: ownerAuthRule.areSubscriptionsPublic ? false : true,
                                        ownerField: ownerAuthRule.ownerField,
                                        ownerValue,
                                };
```

and 

```typescript
                        if (ownerValue) {
[...]

PR #5390, opened 2020-04-13
fix(datastore): Fix subscription creation with model subscription level is public

onlybakam:
_Issue #, if available:_ https://github.com/aws-amplify/amplify-js/issues/5173

_Description of changes:_

Create subscriptions with no "owner" auth check when the model subscription level is public and an owner auth rule is used. This aligns the datastore behavior with the subscription definition created by the Amplify cli.


mdoesburg:
@manueliglesias Can you review this please?

packages/datastore/src/sync/utils.ts
```diff
@@ -183,15 +183,30 @@ export function getAuthorizationRules(
...
 				groups,
+				areSubscriptionsPublic: false,
 			};
 
+			if (authStrategy === 'owner') {
+				// look for the subscription level override
+				// only pay attention to the public level
+				const modelConfig = []
+					.concat(modelDefinition.attributes)
+					.find(attr => attr && attr.type === 'model');
```
manueliglesias, line 201:
```suggestion
				const modelConfig = (<typeof modelDefinition.attributes>[])
					.concat(modelDefinition.attributes)
					.find(attr => attr && attr.type === 'model');
```

packages/datastore/src/sync/utils.ts
```diff
@@ -183,15 +183,30 @@ export function getAuthorizationRules(
...
+
+				// find the subscriptions level. ON is defeault
```
manueliglesias, line 203:
```suggestion
				// find the subscriptions level. ON is default
```

packages/datastore/__tests__/subscription.test.ts
```diff
@@ -85,6 +85,91 @@ describe('sync engine subscription module', () => {
 			)
 		).toEqual(authInfo);
 	});
+	test('owner authorization with public subscription', () => {
+		const model = {
```
manueliglesias, line 89:
```suggestion
		const model: SchemaModel = {
```

packages/datastore/__tests__/subscription.test.ts
```diff
@@ -85,6 +85,91 @@ describe('sync engine subscription module', () => {
...
+		};
+		const authInfo = {
+			authMode: 'AMAZON_COGNITO_USER_POOLS',
+			isOwner: false,
+			ownerField: 'owner',
+			ownerValue: 'user1',
+		};
+
+		expect(
+			SubscriptionProcessor.prototype.getAuthorizationInfo(
```
manueliglesias, line 165:
```suggestion
			// @ts-ignore
			SubscriptionProcessor.prototype.getAuthorizationInfo(
```

Issue #5272, opened 2020-04-02
Mobsf Report: Insecure Random Number Generator. 


vikas5914:
Recently we ran a security scanner (MobSF) on our application and got an error regarding insecure random number on the following line:

https://github.com/aws-amplify/amplify-js/blob/b82c5119ce231f44ba97a81d788546c3f50b6b5a/packages/pushnotification/android/src/main/java/com/amazonaws/amplify/pushnotification/RNPushNotificationMessagingService.java#L110

CVSS V2: 7.5 (high)
CWE: CWE-330 - Use of Insufficiently Random Values
OWASP Top 10: M5: Insufficient Cryptography
OWASP MASVS: MSTG-CRYPTO-6 

Can you please advice on how to fix this?

amhinson:
@vikas5914 Seems like we could replace that with [SecureRandom](https://docs.oracle.com/javase/7/docs/api/java/security/SecureRandom.html) and potentially resolve that issue. Would you be able to make the following changes to that file in your node_modules and see if the error goes away?

```diff
// line 34
- import java.util.Random;
+ import java.security.SecureRandom;

// line 110
- Random randomNumberGenerator = new Random(System.currentTimeMillis());
+ Random randomNumberGenerator = new SecureRandom(System.currentTimeMillis());
```

vikas5914:
@amhinson We got error when using `System.currentTimeMillis()` for the seed. 

So we removed it and its worked. 
Final code : 

`SecureRandom randomNumberGenerator = new SecureRandom()`

mauerbac:
HI @vikas5914 - looks like that is working for you, so will close out. Let me know if you need further help.

vikas5914:
@mauerbac will you accept the pull request for this?

mauerbac:
do you have a link to the PR @vikas5914 ?

vikas5914:
@mauerbac https://github.com/aws-amplify/amplify-js/pull/5418

mauerbac:
Yes @vikas5914 - the team is on it!

PR #5418, opened 2020-04-15
feat(aws-amplify-react-native): Use SecureRandom instead of Random for better security.

vikas5914:
_Issue #5272, 

_Description of changes:_

Replace `java.util.Random` with `import java.security.SecureRandom` for better security. 


packages/pushnotification/android/src/main/java/com/amazonaws/amplify/pushnotification/RNPushNotificationMessagingService.java
```diff
@@ -107,7 +107,7 @@ private void handleFCMMessagePush(ReactApplicationContext context, Bundle bundle
     private void sendNotification(ReactApplicationContext context, Bundle bundle, Boolean isForeground) {
         // If notification ID is not provided by the user for push notification, generate one at random
         if (bundle.getString("id") == null) {
-            Random randomNumberGenerator = new Random(System.currentTimeMillis());
+            SecureRandom randomNumberGenerator = new SecureRandom();
```
amhinson, line 110: I'm wondering if we should still seed this, just for consistency. Maybe something like:

```java
SecureRandom randomNumberGenerator = new SecureRandom();
randomNumberGenerator.setSeed(System.currentTimeMillis());
```
Reference: https://docs.oracle.com/javase/8/docs/api/java/security/SecureRandom.html#setSeed-long-

vikas5914: SecureRandom requires a long seed and System.currentTimeMillis() output is not long.

amhinson: Oh interesting. Based on the [docs](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html), `Random` constructor uses a `long seed`, but `SecureRandom` constructor uses a `byte[] seed`. However you can pass a `long seed` to `setSeed`.

Did you try using the `setSeed` method for `SecureRandom`?

vikas5914: with setSeed , it working.

PR #5421, opened 2020-04-15
fix(aws-amplify): add DataStore export

ericclemmons:
With v3 (modularization) merged, there was a miss with #5202 exporting `{ DataStore}`.

This PR corrects this so customers can use:

```js
import { DataStore, Predicates } from "aws-amplify"
```

I've added test coverage for our current exports (e.g. `import *`) which is **shockingly** large.

There are clearly mistakes (e.g. `a`) that shouldn't be exported that we are likely unaware of.

---


packages/aws-amplify/package.json
```diff
@@ -46,5 +46,47 @@
     "@aws-amplify/storage": "^3.1.6",
     "@aws-amplify/ui": "^2.0.2",
     "@aws-amplify/xr": "^2.1.6"
+  },
+  "jest": {
```
ericclemmons, line 50: Borrowed this config from `@aws/amplify/auth`.

packages/aws-amplify/src/__tests__/exports-test.ts
```diff
@@ -0,0 +1,95 @@
...
+			  "AWSKinesisProvider",
+			  "AWSKinesisFirehoseProvider",
+			  "AmazonPersonalizeProvider",
+			  "Auth",
+			  "Storage",
+			  "StorageClass",
+			  "API",
+			  "APIClass",
+			  "graphqlOperation",
+			  "DataStore",
```
ericclemmons, line 20: This is added via a07e0accac043b186891205f2672607283c56b41

sammartinez: ...

jordanranz: `export * from '@aws-amplify/ui';` in index.ts of `aws-amplify`

ericclemmons:
@manueliglesias Let me know if there are any other exports we should add from `@aws-amplify/datastore`.

The only ones remaining would be TypeScript types or enums, but those are directly available via `@aws-amplify/datastore` if needed (for now).

Issue #5337, opened 2020-04-08
Analytics Firehose encodes data improperly


vincekd:
**Describe the bug**
The data files in s3 produced from the firehose are no longer json text after switching to amplify analytics. There is no data transformation functions or lambda conversion set up on the firehose stream.

**To Reproduce**

1. Set up an angular web page using the instructions on the analytics page:
https://aws-amplify.github.io/docs/js/analytics#using-amazon-kinesis-firehose

2. Use amplify to record some data
```
const record = {
            data: "{json: string}\n",
            streamName: "stream-name",
};
await this.amplify.analytics().record(record, "AWSKinesisFirehose");
```

**Expected behavior**
The data is stored in the s3 bucket in json format.

<details>
  <summary><strong>Environment</strong></summary>

```
 > npx envinfo --system --binaries --browsers --npmPackages --npmGlobalPackages
npx: installed 1 in 0.97s

  System:
    OS: Linux 5.5 Arch Linux
    CPU: (8) x64 Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
    Memory: 4.75 GB / 15.53 GB
    Container: Yes
    Shell: 5.0.16 - /bin/bash
  Binaries:
    Node: 13.11.0 - /usr/bin/node
    Yarn: 1.22.4 - /usr/bin/yarn
    npm: 6.14.4 - /usr/bin/npm
  Browsers:
    Firefox: 74.0
  npmPackages:
    @angular-devkit/architect: ^0.801.3 => 0.801.3 
    @angular-devkit/build-angular: ^0.801.3 => 0.801.3 
    @angular-devkit/core: ^8.1.3 => 8.3.26 
    @angular-devkit/schematics: ^8.1.3 => 8.3.26 
    @angular/cli: ^8.1.3 => 8.3.26 
    @angular/common: ~8.1.2 => 8.1.3 
    @angular/compiler: ~8.1.2 => 8.1.3 
    @angular/compiler-cli: ~8.1.2 => 8.1.3 
    @angular/core: ~8.1.2 => 8.1.3 
    @angular/forms: ~8.1.2 => 8.1.3 
    @angular/language-service: ~8.1.2 => 8.1.3 
    @angular/platform-browser: ~8.1.2 => 8.1.3 
    @angular/platform-browser-dynamic: ~8.1.2 => 8.1.3 
    @angular/router: ~8.1.2 => 8.1.3 
[...]

PR #5451, opened 2020-04-17
fix(analytics): fix Firehose data encoding

iartemiev:
https://github.com/aws-amplify/amplify-js/issues/5337

Fixes encoding issue when recording to Kinesis Firehose. 
Remove references to `PartitionKey` in `AWSKinesisFirehoseProvider.ts:_sendEvents` as that param is only used by Kinesis, not Firehose.

### DIFF
--- a/packages/analytics/src/Providers/AWSKinesisFirehoseProvider.ts
+++ b/packages/analytics/src/Providers/AWSKinesisFirehoseProvider.ts
@@ -47,21 +47,15 @@ export class AWSKinesisFirehoseProvider extends AWSKinesisProvider {
 		const records = {};
 
 		group.map(params => {
-			// spit by streamName
+			// split by streamName
 			const evt = params.event;
-			const { streamName } = evt;
+			const { streamName, data } = evt;
 			if (records[streamName] === undefined) {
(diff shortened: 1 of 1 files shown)


packages/analytics/src/Providers/AWSKinesisFirehoseProvider.ts
```diff
@@ -47,18 +47,15 @@ export class AWSKinesisFirehoseProvider extends AWSKinesisProvider {
...
 			if (records[streamName] === undefined) {
 				records[streamName] = [];
 			}
 
-			const PartitionKey =
-				evt.partitionKey || `partition-${credentials.identityId}`;
-
-			Object.assign(evt.data, { PartitionKey });
-			const Data = JSON.stringify(evt.data);
+			const bufferData = typeof data === 'string' ? data : JSON.stringify(data);
```
jordanranz, line 57: Can data be `undefined`?

iartemiev: `data` is whatever the customer sets it to, so yes, it can be. Good call. I'll add a check for that

Issue #5466, opened 2020-04-20
Pressing enter does not trigger Sign In in AmplifySignIn


faroceann:
**Describe the bug**
When signing in using AmplifyAuthenticator, hitting the enter key does not trigger sign in. I am using the React component, but this is likely the case for all of the wrappers. Users must instead click the "Sign in" button. 

It looks like this was previously fixed in #1023, but as far as I can tell this repo has been refactored to have common components since then and a regression may have been introduced when that occurred. 

**To Reproduce**

1. Wrap app in `withAuthenticator` 
2. Enter credentials in the fields
3. Press enter key while in the password field
4. Nothing happens

**Expected behavior**
When pressing enter using the SignIn component, the onKeyDown event captures charcode `13` and triggers Sign In.

jordanranz:
We opted to not implement this from a `document` level due to conflicts with shadow DOM. I will however PR adding this back to listen to the keyboard event from a component level.

faroceann:
@jordanranz  Thanks, this is so much better. But, this very reliably triggers an error when using `AmplifySignIn`, for example. If you enter some credentials and hit enter, the password is not populated in state yet. This triggers the `Custom auth lambda trigger is not configured for the user pool.` error, because a request goes out to the Cognito IDP without a `password`. If instead of hitting enter, you click **Sign In**, this error does not occur.

PR #5467, opened 2020-04-20
Trigger submit on form section Enter key press

jordanranz:
Fixes: https://github.com/aws-amplify/amplify-js/issues/5466

_Description of changes:_
- Trigger submit for form section on "Enter" key press


packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.tsx
```diff
@@ -19,6 +19,13 @@ export class AmplifyFormSection {
   /** Secondary footer component or text */
   @Prop() secondaryFooterContent: string | FunctionalComponent | null = null;
 
+  @Listen('keydown')
+  handleKeyDown(ev: KeyboardEvent) {
+    if (ev.key === 'Enter') {
```
sammartinez, line 24: Could this turn into a helper? So we could utilize this else where if we need to?

jordanranz: Which part would go into a helper. IMO it is simple enough as is. Especially due to the fact that the @Listen is bound to this specific component

Issue #4360, opened 2019-11-07
please update amplify modules to be adapted with new versions of react-native, android and ios because there are issues that is solved by changing in the module itself


AhmedMesawer:
When I want to build signed APK it gives me this error
```
> Task :@aws-amplify_pushnotification:verifyReleaseResources FAILED
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':@aws-amplify_pushnotification:verifyReleaseResources'.
> 1 exception was raised by workers:
  com.android.builder.internal.aapt.v2.Aapt2Exception: Android resource linking failed
```

I found that when I change SDK version in node_modules\@aws-amplify\pushnotification\android\build.gradle from 26 to 28
```
compileSdkVersion 26 -> 28
    buildToolsVersion "26.0.3" -> "28.0.3"

    defaultConfig {
        minSdkVersion 16
        targetSdkVersion 26 -> 28
        versionCode 1
        versionName "1.0"
    }
```

sammartinez:
@AhmedMesawer Can you please fill out the request within the issue itself? I'm seeing the placeholder information within the issue itself. Thanks ahead of time!

bneigher:
I've been having to do this manually for the past 4 months every time I update my node_modules.  Why isn't aws testing against the current RN version?

AhmedMesawer:
This issue is urgent because it is failing our ci build process for android apps

amhinson:
Closing this issue, as it has been addressed in https://github.com/aws-amplify/amplify-js/pull/5475.

PR #5475, opened 2020-04-21
inherit compileSdkVersion from project

anthonyhumphreys:
Relate to: [this issue](https://github.com/aws-amplify/amplify-js/issues/4360) - few more examples from googling.

I've updated the build.gradle for the push notification module to inherit the compileSdkVersion from the root project - not sure if this needs some safety around it like a fallback version.


amhinson:
Hey @anthonyhumphreys Thanks for the PR!

I'm thinking that it might be a bit safer to have some defaults as well, just in case. Maybe we could do something similar to [this example](https://github.com/react-native-community/react-native-netinfo/blob/master/android/build.gradle#L28) where there are some helper methods that check for the root project versions but fallback to versions defined in a `gradle.properties`?

anthonyhumphreys:
Great idea – I’ll update my PR.

I knew I’d seen that somewhere but couldn’t find the repo!

Anthony Humphreys | Full Stack ITPI Developer
ISS | Lancaster University

Pronouns: He/Him/His
Working hours: Mondays, Tuesdays and Wednesdays 9am – 5:30pm; Thursday and Friday 9am-5pm

www.lancaster.ac.uk<https://www.lancaster.ac.uk/itpi>

[Lancaster University]<https://www.lancaster.ac.uk/>

[Facebook]<https://www.facebook.com/lancasteruniversity/> [Instagram] <https://www.instagram.com/lancasteriss>  [Twitter] <https://www.twitter.com/ISSLancaster>  [LinkedIn] <https://www.linkedin.com/in/anthonyphumphreys/>

anthonyhumphreys:
Hey, I've updated my PR - not sure what a sensible fallback value is... is there a wildcard for 'latest'?

amhinson:
I think we can set both the default `compileSdkVersion` & `targetSdkVersion` at 28 if the root project doesn't have it defined. Does that sound good to you?

anthonyhumphreys:
Sounds like a plan - also, looks like I forgot to push 🤦‍♂️

anthonyhumphreys:
There we go 😆 this should work 

https://github.com/aws-amplify/amplify-js/pull/5475/commits/f204bd02b947f8e441fcd71099e486aaeb820d6c

packages/pushnotification/android/build.gradle
```diff
@@ -19,15 +19,19 @@ allprojects {
     }
 }
 
+def getExtOrIntegerDefault(name) {
+  return rootProject.ext.has(name) ? rootProject.ext.get(name) : 28
+}
+
```
amhinson, line 25: Can we make this function a bit more generic to handle setting all of the version values?
```suggestion
def getExtOrDefault(name, default) {
  return rootProject.ext.has(name) ? rootProject.ext.get(name) : default
}
```

and then we can use it below for all of the versions:
```gradle
compileSdkVersion getExtOrDefault('compileSdkVersion', 28) 

buildToolsVersion getExtOrDefault('buildToolsVersion', "28.0.3") 

minSdkVersion getExtOrDefault('minSdkVersion', 16) 

targetSdkVersion getExtOrDefault('targetSdkVersion', 28) 
```

Also (as noted in the code above), we should bump up the `buildToolsVersion` as well.

anthonyhumphreys: Yeah great suggestion, that slipped me by. Will fix.

anthonyhumphreys:
Should be good to go now @amhinson :) - I hope! Thanks for your patience

anthonyhumphreys:
Cool, i've moved those inside the defaultConfig block now :)

packages/pushnotification/android/build.gradle
```diff
@@ -19,15 +19,18 @@ allprojects {
...
-    compileSdkVersion 26
-    buildToolsVersion "26.0.3"
-
     defaultConfig {
-        minSdkVersion 16
-        targetSdkVersion 26
+        compileSdkVersion getExtOrIntegerDefault('compileSdkVersion', 28)
+        buildToolsVersion getExtOrDefault('buildToolsVersion', "28.0.3")
+        minSdkVersion getExtOrDefault('minSdkVersion', 16)
+        targetSdkVersion getExtOrDefault('targetSdkVersion', 28)
```
amhinson, line 33: Sorry, I think we need one more change and then it should be good to go. Only `minSdkVersion` and `targetSdkVersion` should be inside `defaultConfig`. The other 2 can stay at the root level inside `android`.

anthonyhumphreys: Gotcha :)

PR #5487, opened 2020-04-21
fix(@aws-amplify/storage) Add contentType and Metadata to multipart upload

### LINKED ISSUE #5432: S3 multipart upload using storage.put does not propagate metadata
**Describe the bug**
When using Amplify's `storage.put`, i noticed when the file to upload is larger than a few MB, the upload is completed using s3 multipart upload. However, when this happens, the metadata args passed into `storage.put` is not sent to S3. 
When i use the same code to upload a small file, i can see the metadata being successfully sent to S3. 

**To Reproduce**

When using `Storage.put` or `Storage.vault.put`,and upload a file > 5MB, one can observe the metadata is not being sent and received by s3 when multipart upload happens

Link to sample code
[code here](https://github.com/aws-samples/amplify-photo-gallery-workshop/blob/master/photoalbums/src/App.js#L203)

</details>

**Additional context**
Version of amplify used:

```
    "@aws-amplify/api": "^3.1.6",
    "@aws-amplify/storage": "^3.1.6",
```

**Sample code**
Include additional sample code or a sample repository to help us reproduce the issue. (Be sure to remove any sensitive data)
 
My code is similar to the example [code here](https://github.com/aws-samples/amplify-photo-gallery-workshop/blob/master/photoalbums/src/App.js#L203)  
```
     const result = await Storage.vault.put(
      fileName,
      file,
      {
        metadata: {
          albumid: props.albumId,
          owner: user.username,
        }
      }
    );

```

**_You can turn on the debug mode to provide more info for us by setting window.LOG_LEVEL = 'DEBUG'; in your app._**

Amplifiyer:
_Issue #, if available:_ fixes #5432 

_Description of changes:_
* For multipart uploads, any file properties/metadata are supposed to be added in the CreateMultipartRequest as per the S3 documentation
> If you want to provide any metadata describing the object being uploaded, you must provide it in the request to initiate multipart upload. 
* If the POST request body is empty but content-type header is set, axios is forcibly removing it. See https://github.com/axios/axios/issues/1535 and axios refusing to fix it https://github.com/axios/axios/issues/755. This change is a workaround to set the data as null (instead of undefined) to prevent axios from removing the content-type header. Link for the [source code](https://github.com/axios/axios/blob/dc4bc49673943e35280e5df831f5c3d0347a9393/lib/adapters/xhr.js#L121-L123)

* Because of this workaround, forcing the axios version to avoid newer axios versions breaking this.

### DIFF
--- a/packages/storage/package.json
+++ b/packages/storage/package.json
@@ -40,7 +40,7 @@
     "@aws-sdk/s3-request-presigner": "1.0.0-beta.3",
     "@aws-sdk/util-create-request": "1.0.0-beta.3",
     "@aws-sdk/util-format-url": "1.0.0-beta.2",
-    "axios": "^0.19.0",
+    "axios": "0.19.0",
     "events": "^3.1.0",
     "sinon": "^7.5.0"
   },
--- a/packages/api-rest/package.json
+++ b/packages/api-rest/package.json
@@ -34,7 +34,7 @@
   "homepage": "https://aws-amplify.github.io/",
   "dependencies": {
     "@aws-amplify/core": "^3.2.3",
-    "axios": "^0.19.0"
+    "axios": "0.19.0"
   },
   "jest": {
     "globals": {
--- a/packages/storage/__tests__/providers/AWSS3ProviderManagedUpload-unit-test.ts
+++ b/packages/storage/__tests__/providers/AWSS3ProviderManagedUpload-unit-test.ts
@@ -171,10 +171,7 @@ describe('multi part upload tests', () => {
 		expect(s3ServiceCallSpy).toBeCalledTimes(4);
 
 		// Create multipart upload call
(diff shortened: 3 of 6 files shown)


packages/storage/package.json
```diff
@@ -40,7 +40,7 @@
     "@aws-sdk/s3-request-presigner": "1.0.0-beta.3",
     "@aws-sdk/util-create-request": "1.0.0-beta.3",
     "@aws-sdk/util-format-url": "1.0.0-beta.2",
-    "axios": "^0.19.0",
+    "axios": "0.19.0",
```
elorzafe, line 43: If this is needed we may need to pin the version also for `api-rest` package to avoid duplication of axios that will increase the bundle size

Amplifiyer: yes, makes sense. I'll update it.

Issue #5448, opened 2020-04-17
In production datastore doesn't sync


apoorvmote:
**Describe the bug**
I manually deployed 3 different react app with 
- graphql api dynamodb
- cognito auth with email
- everything default with `create-react-app` settings
- using typescript with react

Everything syncs in development (localhost) but doen't sync in deployed app in both cases amplify app domain & custom domain.

### auth works in production but graphql api doesn't sync

I am able to login with email in production but cannot sync.

I deployed manually with 
- directly from console
- using amplify cli to deploy manually 

Both cases no queries are made and no mutations are made. Only offline access is avaialble.

**Expected behavior**
If it can sync on development (localhost) then it should also sync in production (amplifyapp domain or custom domain).

apoorvmote:
In case there is confusion it doesn't sync in react production build on amplify app or custom domain. It does sync on react development build on localhost. I checked the build folder and in `build/static/js/mainxxxxchunk.js.map` has actual graphql endpoint.

luissombra:
the same issue here

dan8f:
Having the same problem with the sync here.

In my case, a React web app created with create-react-app, DataStore does not sync with the cloud when the production version of the app is published but it does sync with development build.

luissombra:
I ran the  command `yarn upgrade --latest` to update the packages but the problem persists. The versions of my dependencies are as follows:
```
 "@aws-amplify/analytics": "^3.1.7",
"@aws-amplify/api": "^3.1.7",
"@aws-amplify/auth": "^3.2.4",
"@aws-amplify/core": "^3.2.4",
"@aws-amplify/datastore": "^2.0.8",
"@aws-amplify/interactions": "^3.1.7",
"@aws-amplify/pubsub": "^3.0.8",
"@aws-amplify/storage": "^3.1.7",
"@aws-amplify/ui": "^2.0.2",
"@aws-amplify/xr": "^2.1.7",
"@material-ui/core": "^4.9.11",
"@material-ui/icons": "^4.9.1",
"@material-ui/lab": "^4.0.0-alpha.50",
"@material-ui/pickers": "^3.2.10",
"@material-ui/styles": "^4.9.10",
"@testing-library/jest-dom": "^5.5.0",
"@testing-library/react": "^10.0.2",
"@testing-library/user-event": "^10.0.2",
"apexcharts": "^3.18.1",
"autosuggest-highlight": "^3.1.1",
"aws-amplify-react": "^4.1.7",
[...]

PR #5488, opened 2020-04-22
Fix DataStore production build

elorzafe:
_Issue #, if available:_
Fixes: #5448 
_Description of changes:_
Adding `Amplify.register` to DataStore class


cnegrisanu:
@elorzafe @manueliglesias 
Hi guys, for me it is not fixed. I am using datastore 2.0.10 and it works fine when on localhost but doesn't work at all when deployed to the amplify hosting.
Going from offline to online, like the above example, doesn't show ANY attempt to reach out to graphql.

Issue #4175, opened 2019-10-12
withAuthenticator with usernameAttributes not working


jasper72:
My cognito pool is set up to use email as my username.

![image](https://user-images.githubusercontent.com/939156/66704426-daa63d00-ed13-11e9-9a04-966e4a619c8d.png)

And I am using withAuthenticator with usernameAttributes:

![image](https://user-images.githubusercontent.com/939156/66704483-2b1d9a80-ed14-11e9-889a-aa9b3fab5274.png)

But I still get validation errors in the browser complaining about the username:

![image](https://user-images.githubusercontent.com/939156/66704500-4f797700-ed14-11e9-9d0f-c1a499b72ac0.png)

I'm using the latest version of aws-amplify. Any tips gratefully received.

elorzafe:
Thanks @jasper72 we will take a look at this

elorzafe:
@jasper72 I just tried it and is working for me, are you sure that username does not exist on the userpool?

jasper72:
username is a confusing word in this context so I shall carefully say that the email is unique (its an empty pool).

Even if that was not true it would still be a bug because the error message is not that it is not unique but that the format is wrong.

I have also tried this using Auth.signup with the same error message although I can't find an example of the syntax so I may have this wrong.

jasper72:
I'll make a general comment about amplify: why am I configuring this at all? It should configure itself on the basis of the pool it is attached to just as the default signup page does for the pool.

jasper72:
Random thought: could the difference between our environments be because I use Amplify.configure() to establish the pool rather than using a pool generated by Amplify?

elorzafe:
@jasper72 the issue is that you have `name` attribute required but that is not part of the signUp form. You can customized your signUp form like this

```javascript
const signUpConfig = {
  header: 'My Customized Sign Up',
  hideAllDefaults: true,
  defaultCountryCode: '1',
  signUpFields: [
    {
      label: 'Name',
      key: 'name',
      required: true,
      displayOrder: 1,
      type: 'string'
    },
    {
      label: 'Email',
      key: 'email',
      required: true,
      displayOrder: 2,
      type: 'string'
    },
    {
      label: 'Password',
      key: 'password',
      required: true,
      displayOrder: 3,
      type: 'password'
    },
    // and other custom attributes
  ]
};

export default withAuthenticator(App, { usernameAttributes: 'email', signUpConfig, includeGreetings: true });
```

[...]

PR #5491, opened 2020-04-22
fixing a problem where auth config level is different than expected f…

codeglider:
Referring to Issue [Issue 4175](https://github.com/aws-amplify/amplify-js/issues/4175)

awsreactnativeattributeissue

_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/packages/aws-amplify-react-native/src/Auth/index.js
+++ b/packages/aws-amplify-react-native/src/Auth/index.js
@@ -65,11 +65,11 @@ export function withAuthenticator(
 			if (typeof includeGreetings === 'object' && includeGreetings !== null) {
 				this.authConfig = Object.assign(this.authConfig, includeGreetings);
 			} else {
-				this.authConfig = {
+				this.authConfig = {...signUpConfig.authConfig, ...{
 					includeGreetings,
 					authenticatorComponents,
-					signUpConfig,
-				};
+					signUpConfig
+				}};
 			}
 		}
 


amhinson:
@codeglider You can just pass the `usernameAttributes` at the root level of the config instead of inside the `authConfig` object and it should work without this change.

```js
const signUpConfig = {
  usernameAttributes: "email",
  signUpFields: [...]
}
```

https://docs.amplify.aws/ui/auth/authenticator?framework=react-native#sign-upin-with-emailphone-number

codeglider:
> ```js
> usernameAttributes: "email",
> ```

That doesn't seem to work.

Screen Shot 2020-04-22 at 5 13 41 PM

amhinson:
Can you check and make sure you're running the latest version of `aws-amplify-react-native` without any custom changes to it? I have a sample app running with a similar `signUpConfig` and the email is being correctly used for email on the sign in screen.

```js
export default withAuthenticator(App, {
  usernameAttributes: 'email',
  signUpConfig: {
    signUpFields: [...],
  },
});
```

codeglider:
@amhinson that does work!  But it's so confusing in the documentation.
I had it as 
```
export default withAuthenticator(
  App,
  true,
  null,
  null,
  MyTheme,
  signUpConfig,
);
```
When I changed to 
```
export default withAuthenticator(App, {
  includeGreetings: true,
  usernameAttributes: 'email',
  signUpConfig: signUpConfig,
});
```
Working just fine.
Thank you for your guidance.

amhinson:
@codeglider I'm glad that worked for you! We will be working on getting better documentation for this sort of thing in the future, so your feedback will definitely be taken into account. 👍

codeglider:
Thanks. Now post login fails something about `Unrecognized font family 'Material Icons'`

amhinson:
I'd recommend opening up another issue so we can track it better.

oxbits:
@codeglider im havin the same issue _  did u open another tix ?  
or did u solve `Unrecognized font family 'Material Icons'`

PR #5495, opened 2020-04-22
Added missing parameter to getUserData()

### LINKED ISSUE #5493: Wrong getUserData() signature in index.d.ts
**Bug description**
https://github.com/aws-amplify/amplify-js/blob/5314076b68bf3f2c8a5e7e958b98945c3ef9054e/packages/amazon-cognito-identity-js/index.d.ts#L220

The signature in the interface has a single attribute, while the one in the [CognitoUser.ts](https://github.com/aws-amplify/amplify-js/blob/5314076b68bf3f2c8a5e7e958b98945c3ef9054e/packages/amazon-cognito-identity-js/src/CognitoUser.js#L1214) has two attributes. This causes build failure if you also want to pass to the call
`{ bypassCache: true }`.

```
error TS2554: Expected 1 arguments, but got 2.
10:38:52 
10:38:52 299         {
10:38:52             ~
10:38:52 300           bypassCache: true
10:38:52     ~~~~~~~~~~~~~~~~~~~~~~~~~~~
10:38:52 301         });
10:38:52     ~~~~~~~~~
```

**Steps to reproduce**
Try code for use case 31 - Retrieve the user data for an authenticated user.

```javascript
// If you want to force to get the user data from backend,
// you can set the bypassCache to true
cognitoUser.getUserData(
	function(err, userData) {
		if (err) {
			alert(err.message || JSON.stringify(err));
			return;
		}
		console.log('User data for user ' + userData);
	},
	{ bypassCache: true }
);
```

CiprianSt:
Fixes #5493 

Added missing parameter in the signature of getUserData() method.

### DIFF
--- a/packages/amazon-cognito-identity-js/index.d.ts
+++ b/packages/amazon-cognito-identity-js/index.d.ts
@@ -217,7 +217,7 @@ declare module 'amazon-cognito-identity-js' {
 		public enableMFA(callback: NodeCallback<Error, string>): void;
 		public disableMFA(callback: NodeCallback<Error, string>): void;
 		public getMFAOptions(callback: NodeCallback<Error, MFAOption[]>): void;
-		public getUserData(callback: NodeCallback<Error, UserData>, params:any): void;
+		public getUserData(callback: NodeCallback<Error, UserData>, params?:any): void;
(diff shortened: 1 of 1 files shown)


CiprianSt:
Not created after guidelines

PR #5497, opened 2020-04-22
Add stencil eslint config

ashika01:
_Description of changes:_
- This PR adds `stencil-eslint` to our workflow.

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
@@ -8,7 +8,7 @@ import { hasShadowDom } from '../../common/helpers';
   shadow: true,
 })
 export class AmplifyButton {
-  @Element() el!: HTMLElement;
+  @Element() el!: HTMLAmplifyButtonElement;
   /** Type of the button: 'button', 'submit' or 'reset' */
   @Prop() type: ButtonTypes = 'button';
   /** Variant of a button: 'button' | 'anchor' */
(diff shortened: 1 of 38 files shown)


packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
```diff
@@ -8,7 +8,7 @@ import { hasShadowDom } from '../../common/helpers';
   shadow: true,
 })
 export class AmplifyButton {
-  @Element() el!: HTMLElement;
+  @Element() el!: HTMLAmplifyButtonElement;
```
ericclemmons, line 11: This is interesting!  Was there an error/warning, or is this narrowing the scope of what `@Element` is?

ashika01: This was an error caught will applying the lint with the fix suggested to [type element properly](https://github.com/ionic-team/stencil-eslint/blob/master/tests/lib/rules/element-type/element-type.good.tsx).

ashika01:
@ericclemmons Yeah we caught couple of missing jsdocs content for props. Mostly prop usage. Major recommendation was [strict-boolean-expressions](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md). But with our cases i felt it costs more than it provides, so I have skipped it. If we feel the need for it at some point we can always bring that in. But nothing major related to stencil.

Issue #5061, opened 2020-03-08
Problems with callback url being on different domain than SPA


Groot90:
** Which Category is your question related to? **
Auth
** What AWS Services are you utilizing? **
Cognito
** Provide additional details e.g. code snippets **
I'm trying to have a single Cognito User Pools auth sign in callback url ( redirectSignIn in Amplify's terms). However, the app I'm building needs to be able to run from multiple (many) domains.

I'm not sure why it doesn't seem to be possible to have Amplify.Auth.oAuth.redirectSignIn url hosted on a different domain than  the React application (SPA)? 

What if the React app needs to be run from multiple domains? I see that customState parameter can be used to identify the actual domain sending the login request, but I'm having issues when trying that approach. Any time I get a redirect back from the central callback url to the React App, React app fails with the following JS error:

EVENT AUTH HUB: index.js:115
Object { event: "customOAuthState", data: "http__colon____forward____forward__localhost__colon__3001", message: "State for user Google_111083168159815163414" }
index.js:116
[DEBUG] 25:10.750 AuthClass - Error in cognito hosted auth response DOMException: "The operation is insecure."

It seems to be generated by replaceState() in this file
https://github.com/aws-amplify/amplify-js/blob/cf8738ac5e7961a163ce158a7383bc27f693475f/packages/auth/src/Auth.ts

(line 1886),  function _handleAuthResponse

Groot90:
To add some more context here:
I'm trying to implement "State" parameters redirection, described here, using Amplify + Cognito User Pools:
https://auth0.com/docs/users/guides/redirect-users-after-login#keep-reading

Groot90:
The code change in PR #5062 seems to be working - callback url on another domain can be used this way. However, I'd like to hear your comments on this issue. I'm wondering why this is needed? Was this intended behavior or a bug? Thank you!

AntonioAngelino:
We have exactly the same use case and issue described by @Groot90 !
We need to redirect the user to a route that is not `oauth.redirectSignIn`.

I'd suggest implementing the solution explained by @hotgazpacho here: https://github.com/aws-amplify/amplify-js/issues/3391 
> I would expect that the location would not be replaced on successful federated sign in, OR, allow consumers to configure (preferably at call time) the URL to redirect to after successful sign in, keeping it separate from the callback URL.

Issue #3391, opened 2019-06-03
Successful federated OAuth2 sign-in replaces location with callback url


hotgazpacho:
**Describe the bug**
In #3005, when a successful federated OAuth2 sign in occurs, the browser's location is replaced with the callback url:

https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/Auth.ts#L1579-L1581

This is, at best, extraneous, as the user should already be on the callback URL after a successful federated sign in. In my case, however, my callback url is distinct from the main application URL, and performing a `history.replaceState` actively subverts my application's use of a client-side router (`react-router`). That is, I have the component that handles the callback perform a redirect _(possibly to a previously saved location)_ after authentication occurs. However, the timing of events causes this to occur before the call to `history.replaceState` completes. The end result is that my application renders the content for one url, while the browser displays a different URL.

Basically, the idea of the OAuth/OIDC callback url is conflated with the idea of the url that I want to redirect a user to on successful sign in.

**Expected behavior**
I would expect that the location would not be replaced on successful federated sign in, OR, allow consumers to configure _(preferably at call time)_ the URL to redirect to after successful sign in, keeping it separate from the callback URL.

**Sample code**
Include additional sample code or a sample repository to help us reproduce the issue. (Be sure to remove any sensitive data)

My Amplify configuration:
```js
const config = {
  Auth: {
    region: process.env.REACT_APP_AWS_REGION,
    identityPoolId: process.env.REACT_APP_IDENTITY_POOL_ID,
    userPoolId: process.env.REACT_APP_USER_POOL_ID,
    userPoolWebClientId: process.env.REACT_APP_USER_POOL_CLIENT_ID,
    oauth: {
      domain: process.env.REACT_APP_AUTH_DOMAIN,
      scope: ['openid', 'profile', 'phone', 'email', 'aws.cognito.signin.user.admin'],
      redirectSignIn: window.location.origin + '/implicit/callback',
      redirectSignOut: window.location.origin,
      responseType: 'code',
      identityProvider: process.env.REACT_APP_IDP,
      options: {
        // Indicates if the data collection is enabled to support Cognito advanced security features. By default, this flag is set to true.
        AdvancedSecurityDataCollectionFlag: true
      }
    }
  }
};

Amplify.configure(config);
```

maxim-xu:
Exact same problem. Is this fixed?

AntonioAngelino:
> Exact same problem. Is this fixed?
> 
[...]

PR #5513, opened 2020-04-23
Make redirectUrl after oAuth/Federated Sign-in configurable

AntonioAngelino:
_Issue #, if available:_

#3391 , #5061 

_Description of changes:_

Added 2 new AwsCognitoOAuthOpts optional parameters:
- disableRedirectAfterSignIn
- redirectAfterSignIn

By setting `disableRedirectAfterSignIn` to true, [replaceState()](https://github.com/pulsedotso/amplify-js/blob/feature/make-redirect-on-sign-in-more-flexible/packages/auth/src/Auth.ts#L1883) method won't be called.

By setting the `redirectAfterSignIn` URL, replaceState() will redirect the user to the custom redirectAfterSignIn URL.

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -70,7 +70,7 @@ const logger = new Logger('AuthClass');
 const USER_ADMIN_SCOPE = 'aws.cognito.signin.user.admin';
 
 const AMPLIFY_SYMBOL = (typeof Symbol !== 'undefined' &&
-typeof Symbol.for === 'function'
+	typeof Symbol.for === 'function'
 	? Symbol.for('amplify_default')
 	: '@@amplify_default') as Symbol;
 
@@ -1164,7 +1164,7 @@ export class AuthClass {
 						} else {
 							logger.debug(
 								`Unable to get the user data because the ${USER_ADMIN_SCOPE} ` +
-									`is not in the scopes of the access token`
+								`is not in the scopes of the access token`
 							);
 							return res(user);
 						}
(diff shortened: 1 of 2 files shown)


amhinson:
@AntonioAngelino Thanks for the PR! Could you update the conflicts here? We'll get some eyes on it soon 👀

AntonioAngelino:
@amhinson done!

AntonioAngelino:
@amhinson @elorzafe We are still using a forked version of Amplify because this PR was never merged.

Do you have any plans to merge it if I fix the conflicts?

AntonioAngelino:
@amhinson @elorzafe  Any news?

Please tell me if I've implemented the feature in the wrong way and how you think the issue should be addressed.

We cannot use Amplify w/o this PR merged and we're using a forked version of Amplify for several months.

elorzafe:
@AntonioAngelino 

I am looking on this PR and the issues #3391 , #5061.

I am not completely sure that this is the right approach to solve the problem but I am open and more than happy to help with the problems you are facing and find the best long term solution for this.

The intention for `replaceState` after the app is redirected back from hosted UI is to remove the `code` from the url. I understand the problem that the app user might be on a different path of the Web app (e.g. looking for a specific product and then after doing sign in it should return to the same product location). 

For that purpose you can use `customState` when doing `federatedSignIn` and then listen for the `Hub` event for `customOAuthState`. Like I mentioned [on this issue](https://github.com/aws-amplify/amplify-js/issues/3391#issuecomment-756473970) that is the safest way to wait that the code from Hosted UI was processed by Amplify.

Let me know if you have any concerns doing this. I will be looking this issue.

PR #4502, opened 2019-12-03
Update AWSS3Provider.ts

purefan:
Honor the acl config param.
Attempts to fix https://github.com/aws-amplify/amplify-js/issues/960

_Issue #, if available:_
960

_Description of changes:_
The put function does nothing if a config value for acl or level is passed, the comment in the function suggests that "level" should be valid but the code does nothing about it. Chose to implement acl for compatibility with [the sdk](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#upload-property)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5520, opened 2020-04-23
feat(@aws-amplify/storage): AWSS3Provider.ts supports ACL

### LINKED ISSUE #960: Storage.put should set ACL public-read for public files
public files should be put with ACL of public-read

Inside the Storage.put
```
    const params: any = {
      Bucket: bucket,
      Key: final_key,
      Body: file,
      ContentType: type
    };

if (opt.level === "public") {
       params["Acl"] = "public-read";
}
```

ericclemmons:
Resolving merge conflicts for https://github.com/aws-amplify/amplify-js/pull/4502.

Supports `acl` to [match the SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#upload-property):

```js
await storage.put('key', 'object', { acl: 'public-read' }))
```

Fixes #960
Closes #4502

After release, https://docs.amplify.aws/lib/storage/upload/q/platform/js#encrypted-uploads needs to be updated to include.

---


GeorgeBellTMH:
Would be nice if it added the PutObjectAcl's into parameters.json and s3-cloudformation-template.json as well...maybe via amplify storage add/update

PR #5535, opened 2020-04-24
CognitoUser.ListDevices Typescript fix

RidhwanDev:
_Issue #, if available:_
Currently, you are unable to pass null and undefined values to listDevices for paginationToken.
If you send an empty string to listDevices for pagination token it will error and return

`2 validation errors detected: Value '' at 'paginationToken' failed to satisfy constraint: Member must have length greater than or equal to 1; Value '' at 'paginationToken' failed to satisfy constraint: Member must satisfy regular expression pattern: [\S]+`

_Description of changes:_
This is an update for the typescript file to allow you to pass in null values also. I've tested this and it works as it should.


ashika01:
@RidhwanDev : The Cognito provider excepts the pagination token to be string and hence we wouldn't change the type here. For reference - https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListDevices.html

RidhwanDev:
@ashika01  Yes it expects it to be a string, but when passing in an empty string it throws an exception and the call fails. As per the documentation on [amazon-cognito-identity-js github](https://github.com/aws-amplify/amplify-js/tree/master/packages/amazon-cognito-identity-js) it states that the FIRST call should pass in a null, which makes perfect sense as you would not have a pagination token to start with. Just test it out with an empty string, you will se it fails.

Screenshot 2020-04-28 at 00 43 04

RidhwanDev:
Please could you review this issue again.

Thank you

ashika01:
@RidhwanDev Sorry for the confusion. I will get back on this shortly. Thanks for clarifying :)

ashika01:
@RidhwanDev : This PR is gonna need couple of things for doing this the right way. Your changes for paginationToken makes sense. However from cognito's docs I pointed above, the pagination token ought to be string and is optional. So this is what needs to be done.

1. make paginationToken optional in types file.
2. the params below needs to be constructed based on availability of paginationToken. Like if its null, construct params without paginationToken and if present use it.
https://github.com/aws-amplify/amplify-js/blob/master/packages/amazon-cognito-identity-js/src/CognitoUser.js#L1851-L1855

RidhwanDev:
This will cause me to change the order of the parameters as optional parameters cannot be followed by required parameters. 

Should I proceed with this?

RidhwanDev:
Hey, will be going into production in the coming weeks. Could we have further discussion about this solution? 
Thanks

ashika01:
@RidhwanDev Sorry for getting on this late. I did not think about order change of parameters, my bad. That will definitely impact us huge in a bad way. let me take a stab at this.

ashika01:
@RidhwanDev Yup right, if not null then set the pagination token. I think that would be clearer.

RidhwanDev:
But `if (!paginationToken) {` means if NOT paginationToken, which means it's either null or false. Sorry if I'm missing something here..

RidhwanDev:
Sorry for being a nuisance, are we just waiting for approval by somebody else now?

ashika01:
@RidhwanDev Yeah we need another person to review before merge. @elorzafe is working on the review.

RidhwanDev:
Brilliant, thanks. When are we expecting the next release to be due?

RidhwanDev:
Thank you <3

PR #5545, opened 2020-04-24
fix(ui): amplify-input component input change

iartemiev:
Per @sammartinez's suggestion, adding both event handlers. 

The problem with just keeping `onInput` seems to be that `shadowType` (https://github.com/aws-amplify/amplify-js-samples-staging/blob/ui-components/master/cypress/test-utils/amplify-authenticator-tasks.js#L92) doesn't trigger that event. 

The integ tests passed for me when ran locally after this change.


iartemiev:
Closing because the PR to the samples repo fixes this

Issue #5551, opened 2020-04-24
[react] Greetings message does not display


hauntingEcho:
**Describe the bug**
The greetings message (Hello, $USERNAME) does not display.

**To Reproduce**
Go to the greetings documentation example, and see no message:
https://docs.amplify.aws/ui/auth/authenticator/q/framework/react#greetings

If you right click and inspect the greetings element in Firefox, you'll be able to see a span element containing the text `Hello, Test Username` in the DOM.  It is not showing in the page, though.

This issue also affects any attempt I've made to use the greeter.

**Expected behavior**
Message `Hello, Test Username` is visible

**Screenshots**
![Capture](https://user-images.githubusercontent.com/1661988/80242290-c203d180-862a-11ea-9bfd-4d254d904b4c.PNG)

**Additional context**

This issue is visible for me in both Firefox 75.0 (64-bit), and Chrome Version 81.0.4044.122 (Official Build) (64-bit).  I'm using Windows Server 2019.

jordanranz:
The `Greetings` component has a username prop that you can pass in to display the username.

hauntingEcho:
right - the example on the documentation page passes "Test Username" to that property, and is not displaying anything

jordanranz:
This will be released to `@aws-amplify/ui-react@unstable` in the next few minutes.

mtliendo:
Is there a release cycle for when items move from unstable?

mtliendo:
https://github.com/aws-amplify/amplify-js/pull/5555#issuecomment-619483949

jordanranz:
@mtliendo the color and styling should be done using CSS rather than adding to the prop api. Could you try adding the following to your `App.css` or similar file:
```
amplify-greetings {
	--background-color: black;
	--border-color: black;
	color: white;
}
```
If this doesn't work for you please feel free to open a new feature request with your asks

amirfarrahi:
what was the fix?

amirfarrahi:
@jordanranz what was the fix to see the user name?

PR #5555, opened 2020-04-24
Fix improper greetings slot

jordanranz:
_Issue #, if available:_
fixes: https://github.com/aws-amplify/amplify-js/issues/5551

_Description of changes:_
- Remove slot from greetings that was causing improper rendering

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-greetings/amplify-greetings.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-greetings/amplify-greetings.tsx
@@ -20,7 +20,7 @@ export class AmplifyGreetings {
         {// TODO: user <amplify-logo> instead of <span>
         this.logo && <span>{this.logo}</span>}
         <amplify-nav>
-          {this.username && <span slot="greetings-message">Hello, {this.username}</span>}
+          {this.username && <span class="message">Hello, {this.username}</span>}
           <amplify-sign-out handleAuthStateChange={this.handleAuthStateChange} />
         </amplify-nav>
       </header>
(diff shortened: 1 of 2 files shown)


mtliendo:
I'm still not able to view the greeting given the following settings.

Running `@aws-amplify/cli` version 4.18.1
`"@aws-amplify/ui-react": "^0.2.5-unstable.2"`

image
image

packages/amplify-ui-components/src/components/amplify-greetings/amplify-greetings.tsx
```diff
@@ -20,7 +20,7 @@ export class AmplifyGreetings {
         {// TODO: user <amplify-logo> instead of <span>
         this.logo && <span>{this.logo}</span>}
         <amplify-nav>
-          {this.username && <span slot="greetings-message">Hello, {this.username}</span>}
+          {this.username && <span class="message">Hello, {this.username}</span>}
```
mtliendo: Is this supposed to be `this.props.username`?

mtliendo: oh nevermind this. See below for the issue

mtliendo:
It appears to be inheriting the color of it's parent. Given that it's in a white box, would it be ok to default this to black?
image

Black makes for a reasonable default, though another route is to wrap the component like so:

```js
<div style={{ color: "blue" }}>
    <AmplifyGreetings username="Test Username" />
</div>
```

Seems a bit hacky though.

PR #5586, opened 2020-04-28
fix(@aws-amplify/ui-components):  keypress on form

ashika01:
_**Problem:**_
- Right now, when keyboard is used to navigate between button like `create account` or `reset password`  it invokes handle submit.

image

**_Solution:_**
We need to invoke handleSubmit method only if its from form's input field as of now. This PR addresses that by finding the element where the event is .


packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.tsx
```diff
@@ -21,8 +21,8 @@ export class AmplifyFormSection {
 
   // eslint-disable-next-line
   @Listen('keydown')
-  handleKeyDown(ev: KeyboardEvent) {
-    if (ev.key === 'Enter') {
+  handleKeyDown(ev) {
+    if (ev.key === 'Enter' && ev.composedPath()[0].tagName === "INPUT") {
```
Amplifiyer, line 25: it doesn't seem to be supported in IE https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath#Browser_compatibility

ashika01: Yeah... I will try to investigate maybe for polyfill there.

jordanranz, line 25: Maybe we can take advantage of event bubbling here.

ashika01: Yup.. I am already looking at that as solution 👍

ashika01: moved to a different solution using stencil.

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -13,7 +13,7 @@ export class AmplifyInput {
   /** The input type.  Can be any HTML input type. */
   @Prop() type?: TextFieldTypes = 'text';
   /** The callback, called when the input is modified by the user. */
-  @Prop() handleInputChange?: (inputEvent) => void;
+  @Prop() handleInputChange?: (inputEvent: InputEvent) => void;
```
ashika01, line 16: There was a clash between handleInput's argument Event type and import of stencil core 'Event'. I have created `InputEvent` to mitigate that. There maybe better way to clean up this TS which I will revisit later.

packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.tsx
```diff
@@ -20,11 +20,9 @@ export class AmplifyFormSection {
   @Prop() secondaryFooterContent: string | FunctionalComponent | null = null;
 
   // eslint-disable-next-line
-  @Listen('keydown')
-  handleKeyDown(ev: KeyboardEvent) {
```
sammartinez, line 24: I feel like if we remove this, this may lead to updating the integ test we have for shadowTypes right ?

packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.tsx
```diff
@@ -20,11 +20,9 @@ export class AmplifyFormSection {
...
-    if (ev.key === 'Enter') {
-      this.handleSubmit(ev);
-    }
+  @Listen('formSubmit')
```
sammartinez, line 23: Does this break the above? Would it make sense to have two events for this? Since we need one to have if the keydown is Enter so the customer can push this through and a separate one for what the focus is?

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -24,6 +24,21 @@ export class AmplifyInput {
...
+  @Event({
+    eventName: 'formSubmit',
+    composed: true,
+    cancelable: true,
+    bubbles: true,
+  }) formSubmit: EventEmitter;
+
+   // eslint-disable-next-line
+   @Listen('keydown')
+   handleKeyDown(ev) {
```
sammartinez, line 37: awe I see

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
ericclemmons: Can we do `import { Event as StencilEvent } from '@stencil/core'` so we can use the native `Event` interface?

ashika01: the problem there is we loose auto documentation for events and unit test errors. That was my initial approach with all that I decided going this route as thats how some on ionic and other web components handle this.

packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.tsx
ashika01: Sorry I am getting what you mean by break above?

Issue #5423, opened 2020-04-16
Pinpoint: Exceeded maximum endpoint per user count: 10


joebernard:
**Describe the bug**
If I re-install my app enough times it hits the 10 endpoint limit and does not clear unused ones. I've only installed this app on a single test device so there should be plenty of unused endpoints to delete.

It seems this line isn't recognizing the error message correctly and not attempting to clear out the old endpoints:

https://github.com/aws-amplify/amplify-js/blob/master/packages/analytics/src/Providers/AWSPinpointProvider.ts#L460

My IAM policy:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": [
                "mobiletargeting:GetUserEndpoints",
                "mobiletargeting:PutEvents",
                "mobiletargeting:UpdateEndpoint"
            ],
            "Resource": "arn:aws:mobiletargeting:*:*:apps/*"
        }
    ]
}
```

**To Reproduce**
Install the app on a device more than 10 times.

**Expected behavior**
Amplify should clear old endpoints as mentioned in the docs.

<details>
  <summary><strong>Environment</strong></summary>

```
System:
    OS: macOS 10.15.4
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 28.12 MB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.15.3 - ~/.nvm/versions/node/v10.15.3/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.11.3 - ~/.nvm/versions/node/v10.15.3/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 81.0.4044.113
    Firefox: 75.0
    Safari: 13.1
  npmPackages:
    @aws-amplify/analytics: ^3.1.7 => 3.1.7 
    @aws-amplify/auth: ^3.2.4 => 3.2.4 
    @aws-amplify/cache: ^3.1.7 => 3.1.7 
    @aws-amplify/core: ^3.2.4 => 3.2.4 
    @aws-amplify/pushnotification: ^3.0.8 => 3.0.8 
    @aws-amplify/storage: ^3.1.7 => 3.1.7 
    @babel/core: ^7.6.2 => 7.9.0 
[...]

PR #5618, opened 2020-04-29
fix(analytics): correctly extract statusCode from failed request

iartemiev:
_Issue #, if available:_
Fixes aws-amplify/amplify-js#5423


packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -431,8 +431,9 @@ export class AWSPinpointProvider implements AnalyticsProvider {
 
 	private async _handleEndpointUpdateFailure(failureData: EndpointFailureData) {
 		const { err, endpointObject } = failureData;
-		const { statusCode } = err;
-		logger.debug('updateEndpoint failed', err);
+		const statusCode = err.$metadata && err.$metadata.httpStatusCode;
+
+		logger.error('updateEndpoint failed', err);
```
alexandersandberg, line 436: The endpoints seem to update successfully now, but this message is still in the log and is now throwing an error:

error message

iartemiev: You're right, that's a regression introduced in this PR. Will restore it to `logger.debug`

alexandersandberg: There is a similar problem with the 'Missing address' warning now turned into an error, but I noticed that your PR reverts that as well.

IMG_9535F2591BC3-1

PR #5658, opened 2020-05-02
feat: Add support for Pinpoint Journey in PushNotifications

Ashish-Nanda:
### PushNotifications should support the Pinpoint Journey Feature

_Description of changes:_

This PR adds support for Pinpoint Journey in PushNotifications
Originally PushNotifications were designed to only handle `Campaigns` but now the code has been updated to handle any other expected pinpoint `eventSource` which currently includes both `Campaign` and `Journey` and can be extended further later.


packages/pushnotification/src/PushNotification.ts
```diff
@@ -186,39 +197,70 @@ export default class PushNotification {
...
+			rawMessage
+		);
+
+		if (!eventSource) {
+			logger.debug('message received is not from a pinpoint eventSource');
 			return;
 		}
 
+		const isAppInForeground =
+			Platform.OS === 'ios'
```
zhinang-amazon, line 252: Why do we only use `_currentState` for iOS? Based on [doc](https://reactnative.dev/docs/appstate), AppState is cross platform.

Ashish-Nanda: In the case of Android we dont use `this._currentState` to determine foreground information because we get the `foreground` attribute as part of the notification data from the native code that tells us whether the app was in background or foreground state when the notification was received. 
Since we dont get this information in iOS we need to determine it from `AppState`.

ericclemmons:
Awaiting approval from pinpoint developers prior to merging (per @Ashish-Nanda).

packages/pushnotification/src/PushNotification.ts
```diff
@@ -186,87 +193,106 @@ export default class PushNotification {
...
+			const message = this.parseMessageFromIOS(rawMessage);
+			const pinpointData =
+				message && message.data ? message.data.pinpoint : null;
+			if (pinpointData.campaign) {
+				eventSource = 'campaign';
+				eventSourceAttributes = pinpointData.campaign;
+			} else if (pinpointData.journey) {
+				eventSource = 'journey';
+				eventSourceAttributes = pinpointData.journey;
+			}
```
amhinson, line 210: Adding some additional checks to make sure `pinpointData` object exists.

```suggestion
			const pinpointData =
				message && message.data ? message.data.pinpoint : null;
			if (pinpointData && pinpointData.campaign) {
				eventSource = 'campaign';
				eventSourceAttributes = pinpointData.campaign;
			} else if (pinpointData && pinpointData.journey) {
				eventSource = 'journey';
				eventSourceAttributes = pinpointData.journey;
			}
```

Ashish-Nanda: Good point. Updated.

amhinson:
LGTM 🎉

packages/analytics/src/Providers/EventBuffer.ts
```diff
@@ -87,7 +88,11 @@ export default class EventsBuffer {
 			clearInterval(this._interval);
 		}
 
-		if (this._pause || !bufferLength) {
+		// Do not send the batch of events if
+		// the Buffer is paused or is empty or the App is not in the foreground
+		// Apps should be in the foreground since
+		// the OS may restrict access to the network in the background
+		if (this._pause || !bufferLength || !isAppInForeground()) {
```
ericclemmons, line 95: I <3 comments like these!

Ashish-Nanda: Me too 😄

Issue #5638, opened 2020-04-30
Vague warning when DataStore sync fails


mdoesburg:
**Describe the bug**
Take the following schema:
```graphql
type Contact @model
{
  id: ID!
  name: String
  tags: [String]
}
```
As we know there is an issue with optional lists which hasn't been resolved yet. See here: https://github.com/aws-amplify/amplify-js/issues/5292, https://github.com/aws-amplify/amplify-js/issues/5040 In short, the problem is that if you define an optional list, it is not allowed to be empty or your app will crash.

The bug I am running into now however, is that if you observe a model which contains an empty list, the app doesn't crash, but gives a very vague warning after 1 minute.

In addition to that, if you attempt to run the syncContacts query from the AppSync Queries page you get the same vague error:
```json
{
  "data": {
    "syncContacts": null
  },
  "errors": [
    {
      "path": [
        "syncContacts"
      ],
      "data": null,
      "errorType": "InternalFailure",
      "errorInfo": null,
      "locations": [
        {
          "line": 17,
          "column": 3,
          "sourceName": null
        }
      ],
      "message": "An internal error occurred."
    }
  ]
}
```

The data in DynamoDB looks like this:
```json
{
  "__typename": "Contact",
  "_deleted": false,
  "_lastChangedAt": 1588265702639,
  "_version": 1,
  "createdAt": "2020-04-30T16:55:02.639Z",
  "updatedAt": "2020-04-30T16:55:02.639Z",
  "id": "123",
  "name": "John Doe",
  "tags": []
}
```
When removing tags from the data or adding 1 string to the tags array, the app works again and properly syncs and observes data.

**To Reproduce**
1. Create an app with the above schema
2. Make sure that the DynamoDB table is seeded with the data above
2. Start observing the Contact model:
```javascript
const subscription = DataStore.observe(Contact).subscribe((msg) => {
  console.log(msg);
});
```
3. Witness vague warning after 1 minute

**Expected behavior**
I expect to get a less vague error/warning to make debugging easier. Is there a way to see more error/warning details that I am not aware of?

 

<details>
  <summary><strong>Environment</strong></summary>

```
  System:
    OS: macOS 10.15.2
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 83.70 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 13.5.0 - /usr/local/bin/node
[...]

PR #5666, opened 2020-05-03
Fix : vague error message when sync-retry failed

moweex:
_Issue #, if available:_ #5638

_Description of changes:_

Changing the logging message to convert the error into a string using JSON.stringify, the logger was printing [object object] for retry error messages.


jordanranz:
Hey @moweex, thanks for contributing this fix.

Since our `logger` accepts varargs we can just pass `err` as a parameter instead of having to use `JSON.stringify` inline. I [submitted a PR here](https://github.com/aws-amplify/amplify-js/pull/5699) to resolve the core issue.

PR #5668, opened 2020-05-03
fix(@aws-amplify/api-graphql) Respect additionalHeaders in GraphQLAPI subscription requests

bensie:
_Issue #5576_

_Description of changes:_

An `additionalHeaders` object can be supplied to `GraphQLAPI.graphql({query, variables}, additionalHeaders)`, but prior to this PR the option is only respected for query and mutation requests.

In the case of AppSync, these headers are passed to the resolver as input in `context.request.headers`, allowing for queries, mutations, and now subscriptions to be authorized (or execute any custom logic) based on these headers.

### DIFF
--- a/packages/api-graphql/__tests__/GraphQLAPI-test.ts
+++ b/packages/api-graphql/__tests__/GraphQLAPI-test.ts
@@ -858,6 +858,62 @@ describe('API test', () => {
 			expect(observable).not.toBe(undefined);
 		});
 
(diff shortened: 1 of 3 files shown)


jordanranz:
This looks good to me. I'll wait for some of the code owners to review as well.

bensie:
Thank you all for reviewing and merging!

PR #5676, opened 2020-05-04
fix(@aws-amplify/auth): Only handle ?code when redirected from Hosted UI

ericclemmons:
Fixes #3157 (https://github.com/aws-amplify/amplify-js/issues/3157#issuecomment-487839872), fixes #5602 

This PR improves upon the existing `amplify-redirected-from-hosted-ui` value in storage to properly track the redirect flow:

1. Before redirecting, set `amplify-redirected-from-hosted-ui` in storage.
1. When redirected back to the app (or the `urlListener` fires), continuing checking for `code` or `error` **but only parse if `amplify-redirected-from-hosted-ui` is set**. (This is the difference from current behavior).
1. Finally, `amplify-redirected-from-hosted-ui` & continue existing flow.

This also removes code introduced in #4388 via #4798, since the double-handling bug is addressed by only handling requests that were initiated via this flow.

Why not [`customState`](https://docs.amplify.aws/lib/auth/social/q/platform/js#setup-frontend)? The current format is a `string`.  If provided by the customer, our internal custom state would need to be merged, serialized, and deserialized.  Rather than changing this logic, improving the existing "handshake" was a lower lift with less risk.

## TODO

- [x] Test to validate that `Auth.federatedSignIn` sets `amplify-redirected-from-hosted-ui`.
- [x] Test to validate that `Auth._handleAuthResponse` only calls `Auth._oAuthHandler.handleAuthResponse` when `amplify-redirected-from-hosted-ui` is set, and removes it.

## Reproducing

1. Load up any app with Amplify with an invalide `?code`:

  > https://www.amplifyauth.dev/?code=TOTES_FAKE

1. `Error: invalid_grant`

### Before

> Screen Shot 2020-05-04 at 11 16 13 AM

### After

> Screen Shot 2020-05-04 at 11 16 23 AM

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5678, opened 2020-05-04
fix(@aws-amplify/ui-components): Check for 'amplify-redirected-from-hosted-ui' instead of 'amplify-signin-with-hostedUI'

ericclemmons:
**EDIT: Fixed by mirroring #4258 by @Amplifiyer**.

While working on https://github.com/aws-amplify/amplify-js/pull/5676, I discovered that refreshing the page in Firefox after being redirected back from Cognito Hosted UI gets stuck on _Loading..._.

The UI was coupled to an internal storage token `amplify-signin-with-hostedUI`) to bypass checking for the current authenticated user. (This only happens once, because the UI removes this token immediately afterwards)

This PR removes this check and instead, always checks for the current user when mounted.

The code in question originated in #4258. @Amplifiyer may have input on what edge-case I'm missing.

The crux of the issue is really the race-condition that exists: https://github.com/aws-amplify/amplify-js/milestone/23.  If the hosted UI flow completed before `Auth.currentAuthenticatedUser()` resolved, then leaking `amplify-signin-with-hostedUI`to the UI components wouldn't be necessary.  (Note that customers with fully-custom UIs don't have logic for `amplify-signin-with-hostedUI`!)

### Before

> bug-ui-react

### After

> fix-ui-react

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
@@ -3,7 +3,7 @@ import { AuthState, CognitoUserInterface, FederatedConfig, UsernameAliasStrings 
 import {
   AUTH_CHANNEL,
   NO_AUTH_MODULE_FOUND,
-  SIGNING_IN_WITH_HOSTEDUI_KEY,
+  REDIRECTED_FROM_HOSTED_UI,
   AUTHENTICATOR_AUTHSTATE,
   UI_AUTH_CHANNEL,
   TOAST_AUTH_ERROR_EVENT,
@@ -62,8 +62,8 @@ export class AmplifyAuthenticator {
     });
 
     appendToCognitoUserAgent('amplify-authenticator');
-    const byHostedUI = localStorage.getItem(SIGNING_IN_WITH_HOSTEDUI_KEY);
-    localStorage.removeItem(SIGNING_IN_WITH_HOSTEDUI_KEY);
+    const byHostedUI = localStorage.getItem(REDIRECTED_FROM_HOSTED_UI);
+    localStorage.removeItem(REDIRECTED_FROM_HOSTED_UI);
     if (byHostedUI !== 'true') await this.checkUser();
   }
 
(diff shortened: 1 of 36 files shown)


packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -62,9 +61,7 @@ export class AmplifyAuthenticator {
     });
 
     appendToCognitoUserAgent('amplify-authenticator');
-    const byHostedUI = localStorage.getItem(SIGNING_IN_WITH_HOSTEDUI_KEY);
-    localStorage.removeItem(SIGNING_IN_WITH_HOSTEDUI_KEY);
```
ashika01, line 66: Doesn't removing this impact us? If I remember from react code, one of the reasons it was added was to check if the redirect is actually from the hostedUI?!

Amplifiyer:
>  If the hosted UI flow completed before `Auth.currentAuthenticatedUser()` resolved, then leaking `amplify-signin-with-hostedUI`to the UI components wouldn't be necessary.

This is the exact reason why this flag exist in the `aws-amplify-react` package. Do we know why it's different in `ui-components`?

ericclemmons:
@Amplifiyer The UI code was copy/pasted from `aws-amplify-react`, so this behavior may exist there. I'll pop open a sample & check it out.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
ericclemmons: How do you mean impact? My understanding is the same as yours, but with my research (and videos), the refresh behavior blocked the app from rendering at all.

Here's where it originates:
https://github.com/aws-amplify/amplify-js/blob/8cf409559754aa7dd41d42b7a024845a6226709e/packages/auth/src/Auth.ts#L1816-L1818

My take is that there's a leaky abstraction (or incorrect separation of concerns), where `Auth.ts` is setting a value that **only our UI components** are responsible for removing.

I discovered this while working on #5676, but we may address this differently (especially if the race-condition is fixed!) in the near future.

ericclemmons:
Going to update to match #4258 (using the `amplify-redirected-from-hosted-ui`)

Issue #5671, opened 2020-05-04
AuthError - not been configured correctly logged when Auth module is not imported


veproza:
**Describe the bug**
When Auth module is not imported anywhere in project, an AuthError is logged into console, confusing newcomers
_(note that I encountered it on React Experimental brach, but I don't expect it to be a factor)_

**To Reproduce**
Create a tiny demo where `Auth` is not used, e.g.

```javascript
import React from "react";
import Amplify from "@aws-amplify/core";
import {AmplifyAuthenticator} from "@aws-amplify/ui-react";
import awsconfig from './aws-exports';

Amplify.configure(awsconfig);

const Authenticated = (props) => {
    return (
        <AmplifyAuthenticator usernameAlias="email">
            {props.children}
        </AmplifyAuthenticator>
    );
};
export default Authenticated; // or render
```

**Expected behavior**
Console does not contain any error logs

**Code Snippet**
See above for not-working code. All it takes to get rid of the issue is to import `Auth` anywhere in the code while making sure it does not get tree-shaked for never using it, eg:
```javascript
import React, {FunctionComponent} from "react";
import Amplify from "@aws-amplify/core";
import {Auth} from 'aws-amplify';
import {AmplifyAuthenticator} from "@aws-amplify/ui-react";
import awsconfig from './aws-exports';

Amplify.configure(awsconfig);


const Authenticated: FunctionComponent = (props) => {
    Auth.currentAuthenticatedUser(); // noop because of bug in Amplify requiring Auth is imported.
                                     // otherwise a AuthError - Amplify has not been configured correctly will be thrown
    return (
        <AmplifyAuthenticator usernameAlias="email">
            {props.children}
        </AmplifyAuthenticator>
    );
};
export default Authenticated;
```
**Screenshots**
![image](https://user-images.githubusercontent.com/1926602/80993155-e33e9c00-8e3a-11ea-8f1f-9fc134316f04.png)

**What is Configured?**
AWS Exports
   ```
const awsmobile = {
    "aws_project_region": "eu-central-1",
    "aws_appsync_graphqlEndpoint": "https://xxx.appsync-api.eu-central-1.amazonaws.com/graphql",
    "aws_appsync_region": "eu-central-1",
    "aws_appsync_authenticationType": "AMAZON_COGNITO_USER_POOLS",
    "aws_cognito_identity_pool_id": "eu-central-1:xxx",
    "aws_cognito_region": "eu-central-1",
[...]

PR #5681, opened 2020-05-05
Configure all new components or categories when imported

Amplifiyer:
_Issue #, if available:_ #5671, 

_Description of changes:_ 

## What
This change is to call configure() and apply current _"Amplify"_ configuration to all the new categories imported or initialized (especially when imported in a UI component). For instance the following App code will work with the change which right now throws exception "Auth module is not configured"
```
...
import { withAuthenticator } from "@aws-amplify/ui-react";

import Amplify from '@aws-amplify/core';
Amplify.configure(awsconfig);

function App() {
  return (
    <div className="App">
            ...
      </header>
    </div>
  );
}

export default withAuthenticator(App);

```

## Why
After the modularization updates, when an `Amplify.configure()` is called, not all the modules/categories are imported/initialized and hence `Amplify.configure()` misses these categories that are later imported in the App and initialized. 

For instance if users are using UI-components which import Auth (or other categories) and users don't import Auth category directly in the file where `Amplify.configure()` is called, it results in Auth category to not be configured with the aws_exports.

## Assumptions and side effects
This operation `comp.configure(this._config);` is safe with the assumption that all categories' configure() method is idempotent and non-destructive

This also means that categories will always start out configured with default values and hence a lot of redundant code and tests are removed that expect configure is not yet called on a new category imported/initialized.

## Tested
I've tested this PR with the above code and the existing apps with various types of import styles such as

* No root level imports and using Auth UI components
```
import Amplify from '@aws-amplify/core';
import Storage from '@aws-amplify/storage`;
import { withAuthenticator } from "@aws-amplify/ui-react";

Amplify.configure(aws_exports);

// Works now after this change (before Auth was not configured but storage was)
```

* No root level imports and importing category in a different file
```
// App.js
import Amplify from '@aws-amplify/core';
Amplify.configure(aws_exports);

// upload.js
import Storage from '@aws-amplify/storage`;
Storage.put();

// Works now after this change (before this change Storage was not configured)
```

* No root level imports and not using any Auth UI components
```
import Amplify from '@aws-amplify/core';
import Storage from '@aws-amplify/storage`;

Amplify.configure(aws_exports);

// Results in Auth module not present, same as V2 as Auth is needed for guest access still
```

### DIFF
--- a/packages/core/src/Amplify.ts
+++ b/packages/core/src/Amplify.ts
@@ -33,6 +33,14 @@ export class Amplify {
 		} else {
 			logger.debug('no getModuleName method for component', comp);
 		}
+
+		// Finally configure this new component(category) loaded
+		// With the new modularization changes in Amplify V3, all the Amplify
+		// component are not loaded/registered right away but when they are
+		// imported (and hence instantiated) in the client's app. This ensures
+		// that all new components imported get correctly configured with the
+		// configuration that Amplify.configure() was called with.
+		comp.configure(this._config);
 	}
 
 	static configure(config) {
--- a/packages/analytics/__tests__/Analytics-unit-test.ts
+++ b/packages/analytics/__tests__/Analytics-unit-test.ts
@@ -85,16 +85,6 @@ describe('Analytics test', () => {
 			await analytics.startSession();
 			expect(record_spyon).toBeCalled();
 		});
-		test('analytics not configured', async () => {
-			const analytics = new Analytics();
-
-			try {
-				await analytics.startSession();
-			} catch (e) {
-				expect(e.message).toBe('Analytics has not been configured');
-			}
-			expect.assertions(1);
-		});
 	});
 
 	describe('stopSession test', () => {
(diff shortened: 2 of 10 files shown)


packages/core/src/Amplify.ts
```diff
@@ -33,6 +33,14 @@ export class Amplify {
...
 			logger.debug('no getModuleName method for component', comp);
 		}
+
+		// Finally configure this new component(category) loaded
+		// With the new modularization changes in Amplify V3, all the Amplify
+		// component are not loaded/registered right away but when they are
+		// imported (and hence instantiated) in the client's app. This ensures
+		// that all new components imported get correctly configured with the
+		// configuration that Amplify.configure() was called with.
+		comp.configure(this._config);
```
ericclemmons, line 43: Ahh, this makes much more sense!  Especially since categories have the contract of registering themselves with `Amplify`, I wonder why this wasn't the implementation all along?

Amplifiyer: That's why I was wondering given that we had modular imports for a while. Users importing Amplify from `@aws-amplify/core` would have been having this issue even before V3.

Issue #5298, opened 2020-04-04
UI Components - AmplifyAuthenticator - How to use formFields for Sign Up? [React]


budevelop:
** Which Category is your question related to? **
Auth

** What AWS Services are you utilizing? **
Amplify, Cognito, API GW, Lambda

** Provide additional details e.g. code snippets **
I want to change the fields shown in the Sign Up form. Currently Username, Password, Email and Phone Number is displayed. I only need to get Email and Password. How can I change SignUp form fields?

I was able to do it by providing signUpConfig parameter while using withAuthenticator; however with AmplifyAuthenticator I could not make it. I tried to use formFields property as below; however it didn't work.
```
import React from "react";
import Amplify from "aws-amplify";
import { AmplifyAuthenticator, AmplifySignIn, AmplifySignUp, AmplifySignOut } from "@aws-amplify/ui-react";
import awsconfig from "./aws-exports";

Amplify.configure(awsconfig);

const App = () => (
  <AmplifyAuthenticator usernameAlias="email">
    <AmplifySignIn
      headerText="Sign In Header"
      slot="sign-in" />
    <AmplifySignUp headerText="Sign Up Header"
      formFields={[{ type: 'email', label: 'Email', placeholder: 'Email', hint: 'Enter Your Email', required: true }, { type: 'password', label: 'Password', placeholder: 'Enter Your Password', hint: 'Password', required: true }]}
      slot="sign-up" />

    <div>
      Internal App
      <AmplifySignOut />
    </div>
  </AmplifyAuthenticator>
);

export default App;
```

NastyZ98:
Hello I am facing the same issue.

Here is my code below:
```ts
import { AmplifySignUp } from "@aws-amplify/ui-react";

const signUpFields = [
  {
    type: "email",
    label: "custom_label",
    placeholder: "Custom placeholder",
    hint: null,
    required: true,
  },
];

const Auth = () => {
  return (
    <div>
      <AmplifySignUp headerText="Custom form" formFields={signUpFields} />
    </div>
  );
};
```
![Capture d’écran 2020-04-06 à 12 43 19](https://user-images.githubusercontent.com/7148684/78550531-841e4500-7804-11ea-9e5f-49de603b1fb8.png)

Thank you.

ashika01:
@budevelop : If you are trying to achieve sign in or sign up with `email` or `phone_number` instead of `username` , I would suggest you use the property `usernameAlias`. You just add 
```
[...]

PR #5682, opened 2020-05-05
fix(@aws-amplify/ui-components): formFields hint handle

ashika01:
_Issue #, if available:_ #5298

_Description of changes:_
Handle provide hint with default value under certain conditions.

This PR handles hint for password and code and  in the following ways,
- `hint: null` is passed to `formFields`
image

- string is to `hint` in `formFields`
image

- if nothing is passed in or a function component, it would resolve to fallback
image

### DIFF
--- a/packages/amplify-ui-components/src/common/helpers.ts
+++ b/packages/amplify-ui-components/src/common/helpers.ts
@@ -59,16 +59,11 @@ export const onAuthUIStateChange = (authStateHandler: AuthStateHandler) => {
         }
         break;
     }
-  }
+  };
   Hub.listen(UI_AUTH_CHANNEL, authUIStateHandler);
   return () => Hub.remove(UI_AUTH_CHANNEL, authUIStateHandler);
-}
+};
 
-export const handleAuthFormFieldHint = (field) => {
-  if(field["hint"] === null || typeof field["hint"] === "string"){
-    return false;
-  }
-  else{
-    return true;
-  }
-}
+export const isHintValid = field => {
+  return !(field['hint'] === null || typeof field['hint'] === 'string');
+};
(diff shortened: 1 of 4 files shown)


ericclemmons:
Can you share a screenshot/gif of how this looks when used? As well as how, if at all, customers would customize hints?  

I couldn't make it out from the code alone 😅

jordanranz:
Can we put this on `amplify-auth-fields` instead of individual components

jordanranz:
Looking at it further, I am not sure my previous comment would work since only certain subcomponents have hints. Maybe we should allow for a hint on every form field used.

ashika01:
> Looking at it further, I am not sure my previous comment would work since only certain subcomponents have hints. Maybe we should allow for a hint on every form field used.

@jordanranz That might be the way to go and have it `null` by default unless something is mentioned.

I am probably gonna redo hint after we merge this PR https://github.com/aws-amplify/amplify-js/pull/5691

ashika01:
@jordanranz After looking into this a little bit, I think this is probably the best approach for unblocking our customers for now. And like we discussed, we should do api design around FormFields and think about this?!

ashika01:
@ericclemmons @jordanranz Updated the description with whats happening right now. Let me know your thoughts.?

packages/amplify-ui-components/src/common/helpers.ts
```diff
@@ -63,3 +63,12 @@ export const onAuthUIStateChange = (authStateHandler: AuthStateHandler) => {
   Hub.listen(UI_AUTH_CHANNEL, authUIStateHandler);
   return () => Hub.remove(UI_AUTH_CHANNEL, authUIStateHandler);
 }
+
+export const handleAuthFormFieldHint = (field) => {
```
jordanranz, line 67: can we change this to `isHintValid` and just have the content be:
```
return !(field["hint"] === null || typeof field["hint"] === "string")
```

Issue #5435, opened 2020-04-16
Server-Side Rendering (SSR) with Next.js


ericclemmons:
Given Next.js’ popularity, customer impact, and being a hybrid framework supporting both SSR & SSG, solving for Next.js first unblocks other frameworks (e.g. Gatsby, Nuxt).

There is additional work to be done for CLI & Console support (particularly around deployments), but this Epic focuses on JS support.

**This feature is live! See the [Getting Started](https://docs.amplify.aws/start/q/integration/next) tutorial and [SSR docs](https://docs.amplify.aws/lib/ssr/q/platform/js) for the latest!**

## Public Preview

We'd love your help testing out Amplify's `API`, `Auth`, and `DataStore` categories in your Next.js app!

1. First, reinstall Amplify using the `@preview` tag on NPM:

	```shell
	yarn add aws-amplify@preview
	# or
    npm install aws-amplify@preview --save
	```

1. If you're only accessing public data (e.g. `API.graphql(listBlogs)`), you're done! 😌 

1. For secure access the current user on the server, you'll need to scope Amplify to the current request (and _only_ that request) using our new `withSSRContext` helper:

	```diff
	diff --git a/my-next-app/pages/index.tsx b/my-next-app/pages/index.tsx
	index 22b0c471..d7ce454b 100644
	--- a/my-next-app/pages/index.tsx
	+++ b/my-next-app/pages/index.tsx
	@@ -1,6 +1,6 @@
	 import { CognitoUser } from '@aws-amplify/auth'
	 import { GRAPHQL_AUTH_MODE } from '@aws-amplify/api-graphql'
	-import { Amplify, API, Auth } from 'aws-amplify'
	+import { Amplify, withSSRContext } from 'aws-amplify'
	 import { Authenticator } from 'aws-amplify-react'
	 import { GetServerSideProps } from 'next'
	 import Head from 'next/head'
	@@ -13,8 +13,6 @@ import { CreateTodoMutationVariables } from '../src/API'
	
	 Amplify.configure(awsconfig)
	
 	 // 👇 Your frontend code is now "SSR-aware", giving the backend access to the current session
	+const { Auth, API } = withSSRContext()
	+
	 export default function Home(props) {
	@@ -392,8 +390,6 @@ export default function Home(props: Props) {
	 }
	
	 export const getServerSideProps: GetServerSideProps = async (context) => {
       // 👇 Auth & API are now scoped to the _current_ user's session
[...]

PR #5710, opened 2020-05-07
feat(@aws-amplify/auth): UniversalStorage for SSR

ericclemmons:
## Why

`window.localStorage` isn't available to server-side processes, unless credentials are explicitly sent via headers, query params, or POST body.

However, cookies _are_, which means that loading `https://example.com/profile` can render a user-specific page on the server without making a round-trip on the client as you would in a SPA.

## What

Add support for Next.js (#5435) via a new `UniversalStorage` adapter that persists the minimum subset of credentials in cookies, so that they're available on the server for authenticated requests.

The following real, working examples come from my private sample: https://github.com/aws-amplify/amplify-js-samples-staging/pull/90

```tsx
// /pages/index.tsx

import { CognitoUser } from '@aws-amplify/auth'
import { GRAPHQL_AUTH_MODE } from '@aws-amplify/api-graphql'
import { Amplify, API, Auth, UniversalStorage } from 'aws-amplify'
import awsconfig from '../src/aws-exports'

const storage = new UniversalStorage()
Amplify.configure({ ...awsconfig, storage })

type Props = { user?: CognitoUser }

export default function Home(props: Props) {
	...
}

export const getServerSideProps: GetServerSideProps = async (context) => {
  storage.setServerContext(context)

  let user = null

  try {
    user = await Auth.currentAuthenticatedUser({ bypassCache: false })
  } catch (error) {
    console.error(error)
  }

  return {
    props: {
      user: JSON.parse(JSON.stringify(user)),
    },
  }
}
```

```tsx
// /pages/api/profile.tsx
import { Amplify, Auth, UniversalStorage } from 'aws-amplify'
import { NextApiRequest, NextApiResponse } from 'next'

import awsconfig from '../../src/aws-exports'

const storage = new UniversalStorage()
Amplify.configure({ ...awsconfig, storage })

export default async (req: NextApiRequest, res: NextApiResponse) => {
  storage.setServerContext({ req })

  try {
    const user = await Auth.currentAuthenticatedUser()

    return res.status(200).json({ user })
  } catch (error) {
    console.error(error)
    return res.status(500).json({ error })
  }
}
```

## How 

- [x] Use `isomorphic-unfetch` to support SSR fetch for Cognito
- [x] Add `UniversalStorage` from [my sample](https://github.com/aws-amplify/amplify-js-samples-staging/pull/90#issue-406965972) to be exported from both `aws-amplify` and `@aws-amplify/core`.
  - [x] Only `authToken` and `idToken` are persisted for the server.
  - [x] All tokens persist in `localStorage`, as usual.
- [ ] Determine if `UniversalStorage` is the default in the browser or not.
  - Opting _not_ to do this, because `UniversalStorage.setContext(ctx)` is still required per-request.
- [x] Works with User Pools
- [x] Works with https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html (Thanks you @elorzafe!)

---


packages/aws-amplify/src/index.ts
```diff
@@ -48,6 +48,7 @@ export {
 	Signer,
 	I18n,
 	ServiceWorker,
+	UniversalStorage,
```
sammartinez, line 51: Is this a breaking change that we would need to do a Major publish for?

ericclemmons: It would be a minor.

packages/core/src/UniversalStorage/index.ts
```diff
@@ -0,0 +1,109 @@
...
+		return Object.prototype.hasOwnProperty.call(this.store, key)
+			? this.store[key]
+			: null;
+	}
+
+	protected getUniversalItem(key: keyof Store) {
+		const { [key]: value } = isBrowser
+			? nookies.get()
+			: // @ts-ignore Argument of type 'Record<string, string>' is not assignable to parameter of type 'Pick<any, "req"> | { req: any; }'.
+			  // Property 'req' is missing in type 'Record<string, string>' but required in type '{ req: any; }'.ts(2345)
```
ericclemmons, line 36: There are a few `// @ts-ignore` comments that exist because of `nookies`' forked behavior between server & client.  On the server, `context` is `{ req, res }`, while on the browser, it comes from `document.cookies`.

ericclemmons:
Fixed via https://github.com/aws-amplify/amplify-js/pull/6146

Issue #5397, opened 2020-04-13
Refresh token request is going right after Auth.SignIn with two active JWT tokens at the same time  


ters52:
**Describe the bug**
After using Auth.SignIn I see 4 requests to Cognito. The 4th request is a refresh token request and it happens right after sign in. The problem is that old JWT token which Amplify got on the sign-in process
is still active and in this case, I have 2 active Bearer tokens at the same time.

**To Reproduce**
To reproduce it just check requests after Auth.SignIn with 
`AuthFlow: "USER_SRP_AUTH" `

**Expected behavior**
Only one Bearer token has to be generated on sign-in.

**Additional context**

POST Requests list right after SignIn:

1. 
payload:
```
{
  AuthFlow: "USER_SRP_AUTH"
  ClientId: "***"
  AuthParameters: {USERNAME: "***",…}
  ClientMetadata: {}
}
```

response:
```
{
  "success": true
}
```

2. 
payload:

```
{
  ChallengeName: "PASSWORD_VERIFIER"
  ClientId: "***"
  ChallengeResponses: {USERNAME: "***",…}
  ClientMetadata: {}
}
```

response:

```
{"AuthenticationResult":{"AccessToken"...}
```

3. 
payload:
```
{  AccessToken: *** }
```
response:
```
{"UserAttributes":...}
```

4.
payload:
```
{
  ClientId: "***"
  AuthFlow: "REFRESH_TOKEN_AUTH"
  AuthParameters: {,…}
}
```

**Sample code**
Amplify configuration: 
```
Amplify.configure({
  Auth: {
    region: '***',
    userPoolId: '***',
    userPoolWebClientId: '***',
  },
});
```

```
Auth.signIn(email, password)
      .then(() => {
        Auth.currentAuthenticatedUser().then((user) => {
          console.log(user);
        });
      })
      .catch((err) => {
        console.log(err);
      });
```

elorzafe:
@ters52 I reproduced this issue, I will be working on a fix

TimNZ:
@elorzafe 
Can you please look at this post - I'm getting 8 Cognito calls triggered by Auth.signIn()

https://github.com/aws-amplify/amplify-js/discussions/5755

uclaeamsavino:
Is there any version of aws-amplify 2.x.x that doesn't do this? I tried 2.3.0, 2.2.0 and 2.1.0. All immediately make the REFRESH_TOKEN_AUTH call when calling Auth.signIn(). 

[...]

PR #5747, opened 2020-05-11
fix(amazon-cognito-identity-js): Not refresh token immediately after sign in

elorzafe:
_Issue #, if available:_
Fixes: #5397

_Description of changes:_
Refactoring some of the `amazon-cognito-identity-js` package and update logic to refresh the token only for `bypassCache` when calling `getUserData`


packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -1206,9 +1206,36 @@ export default class CognitoUser {
 		return undefined;
 	}
 
+	/**
+	 * PRIVATE ONLY: This is an internal only method and should not
+	 * be directly called by the consumers.
+	 */
+	createGetUserRequest() {
+		return this.client.promisableRequest('GetUser', {
```
ashika01, line 1214:
```suggestion
		return this.client.promisifyRequest('GetUser', {
```
I think this needs to be same name as method in client. Typo?

packages/amazon-cognito-identity-js/src/Client.js
```diff
@@ -10,6 +20,27 @@ export default class Client {
...
 	}
 
+	/**
+	 * Makes an unauthenticated request on AWS Cognito Identity Provider API
+	 * using fetch
+	 * @param {string} operation API operation
+	 * @param {object} params Input parameters
+	 * @returns Promise<object>
+	 */
+	promisifyRequest(operation, params) {
```
ashika01, line 30: Just curious why create a new function here? Can't we just wrap this logic into `createGetUserRequest` method below considering this is only being used there?

Also, what are you thoughts on doing something like this?
```
this.request(operation, params, (err, data) => {
 if(err) {Promise.reject(...)}
else {Promise.resolve(...)}
}
```

packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -1217,44 +1244,42 @@ export default class CognitoUser {
...
-			}
+		// Get Data from Cognito
+		if (!userData) {
+			this.createGetUserRequest
+				.then(data => {
+					this.cacheUserData(data);
+					callback(null, data);
+				})
+				.catch(callback);
+			return;
```
ashika01, line 1260: why return? I think we could handle this by else if blocks? I might be missing something here.

packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -1217,44 +1244,42 @@ export default class CognitoUser {
...
+				.then(data => {
+					this.cacheUserData(data);
+					return data;
+				})
+				.then(data => {
+					return this.refreshSessionIfPossible().then(() => data);
+				})
+				.then(data => callback(null, data))
+				.catch(callback);
+			return;
```
ashika01, line 1274: same as above

packages/amazon-cognito-identity-js/src/Client.js
elorzafe: > Just curious why create a new function here? Can't we just wrap this logic into createGetUserRequest method below considering this is only being used there?

I did that because when refactoring other client calls it would be same code. 

> Also, what are you thoughts on doing something like this?

> this.request(operation, params, (err, data) => {
 if(err) {Promise.reject(...)}
else {Promise.resolve(...)}
}

Is a good point, I tried to not make changes on API but `Client` is not exported from the package.

packages/amazon-cognito-identity-js/src/CognitoUser.js
elorzafe: I don't like so many if-else blocks, but I could clean up a little bit more.

packages/amazon-cognito-identity-js/src/CognitoUser.js
elorzafe: I will clean it a little more.

Thanks for your comments!

PR #3769, opened 2019-07-31
feat(api): Cancel in flight api request

kbuechl:
*Issue #, if available:*
Fixes #989 

*Description of changes:*
Added cancel token options to rest api as well as graphql. Aliased axios' `isCancel` as well as `CancelToken` to keep the interface familiar.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #3098, opened 2019-04-15
How to cancel Storage file upload


pippo111:
I cannot find it anywhere in Amplify Storage Documantation, but is there a way to cancel file upload (put) ? It will be useful to give user possibility to cancel upload of files.

I'm using Storage.put like so:
```
return Storage.vault.put(filename, buffer, {
    metadata,
    progressCallback (e) {
      // log progress
    }
  })
```
but all I get is Promise, which I cannot cancel simply by js.

haverchuck:
@pippo111 -  I do not believe this is supported at the moment;  I will mark this as a feature request.

pippo111:
@haverchuck Any news on that ? It is getting really problematic, when internet connection breaks and user is unable to cancel his upload. Is there any way to do it maybe via some lower level API ? Any hints on that ?

videoAbyss:
Any idea when this might be supported ?

nitinkaveriappa:
Will a cancel feature be added???

juanludlf:
there is a PR in progress
https://github.com/aws-amplify/amplify-js/pull/3769

sammartinez:
Resolving as per the above implementation has been complete

sorokinvj:
I will open this one, because API.cancel() as mentioned [here](https://github.com/aws-amplify/amplify-js/pull/5753#issuecomment-654481080) has not solved this issue. 

But feature is very needed! In my app (but I suspect in any app), user may start upload huge file by mistake and there is no workaround to cancel it

boggsey:
Agreed that this is very needed. I'm uploading a lot of very large files and need the ability to easily cancel in case an incorrect file is being uploaded.

sammartinez:
All, 

Wanted to provide an update on this feature. We are planning this work out for next quarter. While we don't have a timeline on when this feature will complete, know that we have heard you and to look out for more updates on this feature in the future. 

Thanks! 

Amplify JavaScript

sammartinez:
#7984

enriqueavg:
@sammartinez Thanks for the update. Cancel feature for put and get are very needed.

agileurbanite:
Our latest PR ☝️ introduces the ability to cancel put and get requests for storage.  Please let us know if this addresses your needs which will go out in the next release.

PR #5753, opened 2020-05-12
Add feature to cancel axios requests used by API rest and graphql category

### LINKED ISSUE #989: Add the ability to cancel any API call
Do you want to request a feature or report a bug?
feature
What is the current behavior?
an api call can't be canceled once made.

What is the expected behavior?
https://github.com/axios/axios#cancellation

Amplifiyer:
Co-authored-by: kbuechl <karlrbuechler@gmail.com>

_Issue #, if available:_ fixes #989, closes #3769

_Description of changes:_ This is a replacement PR for #3769 since that PR has a lot of merge conflicts which requires a pretty much rewrite since API category has been refactored into api-rest and api-graphql categories.

This PR builds on the same concept as https://github.com/aws-amplify/amplify-js/pull/3769#issuecomment-552660025 but instead of keeping keeping the promises from axios in the weakmap, we are utilizing the promises that library return to the clients. This is because using async/await wraps any promises being returned and renders the whole idea of using promises as key useless.

### **Gist of the implementation:**
We keep a weakmap<promise, canceltoken> for every request made in the RestClient. If the client wants to cancel a request, they can use the `promise` received after making the request and call `API.cancel(promise)`

1. For every axios request, a unique cancel token is generated and added in the request.
2. Promise for fulfilling the request is then mapped to that unique cancel token.
3. The promise is returned to the client.
4. Clients can either wait for the promise to fulfill or call `API.cancel(promise)` to cancel the request.
5. If `API.cancel(promise)` is called, then the corresponding cancel token is retrieved from the map.
6. Promise returned to the client will get rejected state with the error message provided during cancel.
7. Clients can check if the error is because of cancelling by calling `API.isCancel(error)`.

### **Testing**
Other than unit tests, I have manually tested api-rest category by creating a server API that returns success after 5 secs. I then made multiple calls to the API and cancelled half of them randomly after 1 secs of initiation. Validated that the calls cancelled are the ones that were supposed to be cancelled, rest completed successfully.

### **Callouts**
This PR required a slight change in the API category, All the APIs are now marked non-async. They are still returning promises and can be awaited upon (as shown in the unit tests). This is required so that the returned promise doesn't get changed. I don't consider this to be a breaking change in my opinion, want to know what others think.

### DIFF
--- a/packages/api-graphql/__tests__/GraphQLAPI-test.ts
+++ b/packages/api-graphql/__tests__/GraphQLAPI-test.ts
@@ -14,7 +14,20 @@ import {
 import PubSub from '@aws-amplify/pubsub';
 import Cache from '@aws-amplify/cache';
 import * as Observable from 'zen-observable';
+import axios, { CancelTokenStatic } from 'axios';
 
+axios.CancelToken = <CancelTokenStatic>{
+	source: () => ({ token: null, cancel: null }),
+};
+axios.isCancel = (value: any): boolean => {
+	return false;
+};
+
+let isCancelSpy = null;
+let cancelTokenSpy = null;
+let cancelMock = null;
+let tokenMock = null;
+let mockCancellableToken = null;
 jest.mock('axios');
 
 const config = {
@@ -29,6 +42,17 @@ afterEach(() => {
 });
 
 describe('API test', () => {
+	beforeEach(() => {
+		cancelMock = jest.fn();
+		tokenMock = jest.fn();
+		mockCancellableToken = { token: tokenMock, cancel: cancelMock };
+		isCancelSpy = jest.spyOn(axios, 'isCancel').mockReturnValue(true);
+		cancelTokenSpy = jest
+			.spyOn(axios.CancelToken, 'source')
+			.mockImplementation(() => {
+				return mockCancellableToken;
+			});
+	});
 	describe('graphql test', () => {
 		test('happy-case-query', async () => {
 			const spyonAuth = jest
@@ -96,10 +120,98 @@ describe('API test', () => {
 					service: 'appsync',
 					region,
 				},
+				cancellableToken: mockCancellableToken,
 			};
 
 			await api.graphql(graphqlOperation(GetEvent, variables));
 
+			expect(spyon).toBeCalledWith(url, init);
+		});
+
+		test('cancel-graphql-query', async () => {
+			const spyonAuth = jest
+				.spyOn(Credentials, 'get')
+				.mockImplementationOnce(() => {
+					return new Promise((res, rej) => {
+						res('cred');
+					});
+				});
+
+			const spyon = jest
+				.spyOn(RestClient.prototype, 'post')
+				.mockImplementationOnce((url, init) => {
+					return new Promise((res, rej) => {
+						rej('error cancelled');
+					});
+				});
+
+			const api = new API(config);
+			const url = 'https://appsync.amazonaws.com',
+				region = 'us-east-2',
+				apiKey = 'secret_api_key',
+				variables = { id: '809392da-ec91-4ef0-b219-5238a8f942b2' };
+			api.configure({
+				aws_appsync_graphqlEndpoint: url,
+				aws_appsync_region: region,
+				aws_appsync_authenticationType: 'API_KEY',
+				aws_appsync_apiKey: apiKey,
+			});
+			const GetEvent = `query GetEvent($id: ID! $nextToken: String) {
+				getEvent(id: $id) {
+					id
+					name
+					where
+					when
+					description
+					comments(nextToken: $nextToken) {
+						items {
+						commentId
+						content
+						createdAt
+						}
+					}
+				}
+			}`;
+
+			const doc = parse(GetEvent);
+			const query = print(doc);
+
+			const headers = {
+				Authorization: null,
+				'X-Api-Key': apiKey,
+				'x-amz-user-agent': Constants.userAgent,
+			};
+
+			const body = {
+				query,
+				variables,
+			};
+
+			const init = {
+				headers,
+				body,
+				signerServiceInfo: {
+					service: 'appsync',
+					region,
+				},
+				cancellableToken: mockCancellableToken,
+			};
+
+			const promiseResponse = api.graphql(
+				graphqlOperation(GetEvent, variables)
+			);
+			api.cancel(promiseResponse as Promise<any>, 'testmessage');
+
+			expect.assertions(5);
+
+			expect(cancelTokenSpy).toBeCalledTimes(1);
+			expect(cancelMock).toBeCalledWith('testmessage');
+			try {
+				await promiseResponse;
+			} catch (err) {
+				expect(err).toEqual('error cancelled');
+				expect(api.isCancel(err)).toBeTruthy();
+			}
 			expect(spyon).toBeCalledWith(url, init);
 		});
 
@@ -169,6 +281,7 @@ describe('API test', () => {
 					service: 'appsync',
 					region,
 				},
+				cancellableToken: mockCancellableToken,
 			};
 
 			await api.graphql(graphqlOperation(doc, variables));
@@ -258,6 +371,7 @@ describe('API test', () => {
 					service: 'appsync',
 					region,
 				},
+				cancellableToken: mockCancellableToken,
 			};
 
 			await api.graphql(graphqlOperation(GetEvent, variables));
@@ -334,6 +448,7 @@ describe('API test', () => {
 					service: 'appsync',
 					region,
 				},
+				cancellableToken: mockCancellableToken,
 			};
 
 			await api.graphql({
@@ -344,6 +459,7 @@ describe('API test', () => {
 
 			expect(spyon).toBeCalledWith(url, init);
 		});
+
 		test('multi-auth default case api-key, using AWS_IAM as auth mode', async () => {
 			expect.assertions(1);
 			jest.spyOn(Credentials, 'get').mockReturnValue(Promise.resolve('cred'));
@@ -397,6 +513,7 @@ describe('API test', () => {
 					service: 'appsync',
 					region,
 				},
+				cancellableToken: mockCancellableToken,
 			};
 
 			await api.graphql({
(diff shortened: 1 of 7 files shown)


ericclemmons:
@Amplifiyer Based on your description, is my assumption of real-world usage correct?

```js
import { API } from "aws-amplify"

...

export function TodoList() {
	const [todos, setTodos] = useState([])

	useEffect(() => {
		const promise = API.graphql(listTodos).then((response) => setTodos(resposne.data.listTodos));

		return () => API.cancel(promise)
	}, []);

	return (
		<ul>
			{todos.map((todo) => <li>{todo.title}</li>}
		</ul>
	)
}
```

Amplifiyer:
@ericclemmons it's not necessary just for cleanups, but providing an opportunity to let customers have a `cancel` button somewhere on the UI to cancel the ongoing (a huge download/upload) request.

packages/api-graphql/src/GraphQLAPI.ts
```diff
@@ -280,6 +301,24 @@ export class GraphQLAPIClass {
 		return response;
 	}
 
+	/**
+	 * Checks to see if an error thrown is from an api request cancellation
+	 * @param {any} error - Any error
+	 * @return {boolean} - A boolean indicating if the error was from an api request cancellation
+	 */
+	isCancel(error) {
+		return this._api.isCancel(error);
```
ericclemmons, line 310: Just noting that this is consistent with `axios.isCancel`.

packages/api-rest/src/RestClient.ts
```diff
@@ -212,6 +236,48 @@ export class RestClient {
...
+	}
+
+	/**
+	 * Checks to see if an error thrown is from an api request cancellation
+	 * @param {any} error - Any error
+	 * @return {boolean} - A boolean indicating if the error was from an api request cancellation
+	 */
+	isCancel(error): boolean {
+		return axios.isCancel(error);
+	}
```
ericclemmons, line 259: Is there an opportunity to make `API.isCancel` less ambiguous (e.g. `isCancellableToken`), given that we have `getCancellableToken()`?

There may be an opportunity for `API.cancel(promise)` having `API.isCancelled(promise)`.

packages/api-graphql/__tests__/GraphQLAPI-test.ts
```diff
@@ -29,6 +42,17 @@ afterEach(() => {
...
+	beforeEach(() => {
+		cancelMock = jest.fn();
+		tokenMock = jest.fn();
+		mockCancellableToken = { token: tokenMock, cancel: cancelMock };
+		isCancelSpy = jest.spyOn(axios, 'isCancel').mockReturnValue(true);
+		cancelTokenSpy = jest
+			.spyOn(axios.CancelToken, 'source')
+			.mockImplementation(() => {
+				return mockCancellableToken;
+			});
```
ericclemmons, line 54: Can you tell me more about why `axios.CancelToken` & `axios.isCancel` both replaced at the top of the test & mocked in `beforEach`?

ericclemmons:
@Amplifiyer That makes sense.  Real-world snippets help me understand PRs better :)  From what I understand, cancellation is limited to `API` uploads/downloads (likely via REST), while `Storage` is unaffected?

Amplifiyer:
@ericclemmons That is correct. This feature is only applicable for API category, both rest and graphql (Both use REST Client anyway)

packages/api-graphql/__tests__/GraphQLAPI-test.ts
Amplifiyer: That is somehow the only way to mock the axios static members. Axios in general is very hard to mock  🤯

packages/api-rest/src/RestClient.ts
Amplifiyer: I'm not sure I understand the question correctly. Do you mean renaming the methods or introducing new methods/functionality to figure out isCancel using cancellable tokens?

ericclemmons: Sorry, I'll be clearer.

In the interest of reducing ambiguity, I was proposing:

- `isCancellableToken(error): boolean`, for symmetry with `getCancellableToken()`.
- `isCanceled(promise): boolean` for checking if a `promise` has been cancelled via `API.cancel(promise)`.

Amplifiyer: I see. `getCancellableToken()` is not exposed to the clients of Amplify. We are only proposing two APIs `cancel(promise)` and `isCancel(error)`

To support `isCanceled(promise)`, we would need to keep additional state of that promise in the library whether it has been canceled or not. For `isCancel(error)`, we are using the axios's out of the box functionality.

ericclemmons: Right, in this case I was proposing `isCancellableToken(error)`, since Axio's `axios.isCancel(error)` is ambiguous to me (even though `getCancellableToken()` isn't public).

I misunderstood how the WeakMap was being utilized: it appeared that it could be referenced via `isCanceled(promise)`.

I'm good to move forward.  We're not locked into anything at this point, and these suggestions are all minor IMO.

Let's move forward!

packages/api-graphql/src/GraphQLAPI.ts
```diff
@@ -241,14 +253,17 @@ export class GraphQLAPIClass {
 			variables,
 		};
 
-		const init = {
-			headers,
-			body,
-			signerServiceInfo: {
-				service: !customGraphqlEndpoint ? 'appsync' : 'execute-api',
-				region: !customGraphqlEndpoint ? region : customEndpointRegion,
+		const init = Object.assign(
```
manueliglesias, line 256: Non blocking at all, just a question. I noticed the usage of `Object.assign`, while I don't feel strongly about spread operator vs `Object.assign`, I'd like to hear your opinion and why you went with `Object.assign` while other parts of the code base tend to use the spread operator.

packages/api-rest/src/RestClient.ts
```diff
@@ -19,7 +19,7 @@ import {
 } from '@aws-amplify/core';
 
 import { apiOptions } from './types';
-import axios from 'axios';
+import axios, { CancelTokenSource, CancelToken } from 'axios';
```
manueliglesias, line 22: `CancelToken` is an unused import

packages/api-rest/src/RestClient.ts
```diff
@@ -42,12 +42,32 @@ export class RestClient {
...
+	 * 1. For every axios request, a unique cancel token is generated and added in the request.
+	 * 2. Promise for fulfilling the request is then mapped to that unique cancel token.
+	 * 3. The promise is returned to the client.
+	 * 4. Clients can either wait for the promise to fulfill or call `API.cancel(promise)` to cancel the request.
+	 * 5. If `API.cancel(promise)` is called, then the corresponding cancel token is retrieved from the map below.
+	 * 6. Promise returned to the client will be in rejected state with the error provided during cancel.
+	 * 7. Clients can check if the error is because of cancelling by calling `API.isCancel(error)`.
+	 *
+	 * For more details, see https://github.com/aws-amplify/amplify-js/pull/3769#issuecomment-552660025
+	 */
```
manueliglesias, line 59: I love the comment! 😄

packages/api-rest/src/RestClient.ts
```diff
@@ -42,12 +42,32 @@ export class RestClient {
...
+	private _cancelTokenMap: WeakMap<any, CancelTokenSource> = null;
+
 	/**
 	 * @param {RestClientOptions} [options] - Instance options
 	 */
 	constructor(options: apiOptions) {
 		this._options = options;
 		logger.debug('API Options', this._options);
+		if (this._cancelTokenMap == null) {
+			this._cancelTokenMap = new WeakMap();
```
manueliglesias, line 69: WeakMap FTW!!! 💯

packages/api-rest/src/RestClient.ts
Amplifiyer: Removing it.

packages/api-graphql/src/GraphQLAPI.ts
Amplifiyer: Didn't really give it a thought. Was deep into reading the `configure()` code for an earlier PR and was fresh in my mind since `Object.assign` is used a lot there so just naturally went with it. If spread is better,  I can change it.

berzavlu:
Some example using API.cancel?
in debug i cannot see cancel method in __proto__
`javascript
APIClass {_api: RestClient, _options: {…}}
_api: RestClient {_region: "us-east-2", _service: "execute-api", _custom_header: undefined, _options: {…}}
_options: {Auth: {…}, Storage: {…}, endpoints: Array(2)}
__proto__:
configure: ƒ (options)
createInstance: ƒ ()
del: ƒ (apiName, path, init)
endpoint: ƒ (apiName)
get: ƒ (apiName, path, init)
getGraphqlOperationType: ƒ (operation)
getModuleName: ƒ ()
graphql: ƒ (_a, additionalHeaders)
head: ƒ (apiName, path, init)
patch: ƒ (apiName, path, init)
post: ƒ (apiName, path, init)
put: ƒ (apiName, path, init)
_ensureCredentials: ƒ ()
_graphql: ƒ (_a, additionalHeaders)
_graphqlSubscribe: ƒ (_a)
_headerBasedAuth: ƒ (defaultAuthenticationType)
constructor: ƒ APIClass(options)
__proto__: Object
`

some help please i am using aws-amplify@3.0.12

berzavlu:
SOLVED, i was using import { API } from "aws-amplify";
only need to change: import { RestAPI as API } from "@aws-amplify/api-rest";
👍

sorokinvj:
Guys, can I cancel Storage.put with that? I suspect not, right?

Amplifiyer:
> 
> 
> Guys, can I cancel Storage.put with that? I suspect not, right? @Amplifiyer

No, Storage category doesn't use API category (or axios) for making network requests. There is another issue https://github.com/aws-amplify/amplify-js/issues/3098 that is opened for this feature request but hasn't been prioritized.

agileurbanite:
Folks we recently introduced work to allow users to cancel put and get ☝️ , please let us know if this meets your needs and it will be made available in the next release.  Appreciate everyones feedback in helping us get better!

PR #5772, opened 2020-05-13
Disable autocomplete on all Input components

### LINKED ISSUE #5048: Disable autocomplete for username input
Currently autocomplete is only disabled for password inputs, but it is also considered a security threat
for username inputs
disable autocomplete for username input or provide option to disable it

matteofigus:
Fixes #5048

_Description of changes:_
When an app using react components is shared between multiple customers or a machine is left unlocked, username/email/phone autocomplete may be persisted in the browser and accessed by other users. 

By making the `<Input>` component to automatically set autocomplete=OFF to all scenarios, we can include not only passwords but also all the other fields.

If you disagree we can also make this configurable (but perhaps the default should still be OFF).

Thoughts?

Thanks for your help.

### DIFF
--- a/packages/aws-amplify-react/__tests__/Amplify-UI/__snapshots__/Amplify-UI-Components-React-test.tsx.snap
+++ b/packages/aws-amplify-react/__tests__/Amplify-UI/__snapshots__/Amplify-UI-Components-React-test.tsx.snap
@@ -22,6 +22,7 @@ exports[`AmplifyUi test render FormRow correctly 1`] = `
 
 exports[`AmplifyUi test render Input correctly 1`] = `
 <input
+  autoComplete="off"
   className="Input__input___3e_bf"
   style={Object {}}
 />
--- a/packages/aws-amplify-react/__tests__/Auth/__snapshots__/ConfirmSignIn-test.tsx.snap
+++ b/packages/aws-amplify-react/__tests__/Auth/__snapshots__/ConfirmSignIn-test.tsx.snap
@@ -1235,7 +1235,6 @@ exports[`ConfirmSignIn render test render correctly with Props confirmSignIn 1`]
            *
         </Component>
         <Component
-          autoComplete="off"
           autoFocus={true}
           data-test="confirm-sign-in-code-input"
           key="code"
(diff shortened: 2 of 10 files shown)


ericclemmons:
Defaulting to `off` can be risky, since it changes the default browser behavior and, at least in my past experience, has negatively impacted UX with password managers (whether in-browser or external, such as 1Password).

With it defaulted, we would also need to support configuration so other customers can opt-_in_.

@jordanranz Are there existing mechanisms to customize `autoComplete` today?

matteofigus:
Thanks for your comment @ericclemmons.

In general I was thinking that while breaking the default browser behaviour, this change could be acceptable because I've seen previous PRs doing the same for password related fields, so I found strange seeing the autocomplete off only for passwords, and not for other fields (I guess that already impacts password managers).
Perhaps as part of this discussion, we should clarify first if we think password inputs should behave differently than other inputs. I can't think of valid reasons but I may be missing something.

Thanks for your help

jordanranz:
@ericclemmons yes, it would however require passing in custom forms which would be a lot of friction. Basically you would use: https://docs.amplify.aws/ui/auth/authenticator/q/framework/react#custom-form-fields, then pass the configuration to inputProps for each form field.

I would however like to dive deeper and ask which form fields should and should not have auto complete by default. I don't think globally disabling them is the best approach, but maybe we can intelligently provide defaults and a more simple way to override.

This work will also need to be built into the new UI components.

matteofigus:
> @ericclemmons yes, it would however require passing in custom forms which would be a lot of friction. Basically you would use: https://docs.amplify.aws/ui/auth/authenticator/q/framework/react#custom-form-fields, then pass the configuration to inputProps for each form field.

Ok I tried with this approach without success. The fields you can pass are outlined here and autocomplete is not whitelisted: https://github.com/aws-amplify/amplify-js/blob/master/packages/amplify-ui-components/src/components/amplify-auth-fields/component-field-mapping.tsx
 
> I would however like to dive deeper and ask which form fields should and should not have auto complete by default. I don't think globally disabling them is the best approach [...]

Can you explain why? I think the goal here is protecting PII on a shared/unlocked machine and it seems like any possible field we handle in amplify auth flows is relevant.

> This work will also need to be built into the new UI components.

Sure, happy to update the PR after we find an agreement.

matteofigus:
Do you want me to close this one? I am unblocked because I'm setting autocomplete=off in each component where I need it.

jordanranz:
I believe this can be closed. For the new UI library, we are going to look into implementing a form-wide autocomplete disable prop. At the moment you should be able to use the `inputProps` via the `formFields` prop to disable `autocomplete` on each individual form field's input.

PR #5778, opened 2020-05-13
feat(@aws-amplify/datastore): enable keyName relations

### LINKED ISSUE #5088: Datastore cannot delete one to many relationship
**Describe the bug**
I have simple one to many relationship. I haven't even gotten to many part yet. I am just testing adding and deleting to one part of relationship. I can add but I cannot delete. I get this error
` Failed to execute 'index' on 'IDBObjectStore': The specified index was not found`

**To Reproduce**
`npx create-react-app blog --use-npm --template typescript`
`amplify init`
`amplify add api` not adding auth just trying to keep it simple.
add following schema
```
type Post @model {
  id: ID!
  title: String!
  postComments: [Comment] @connection(keyName: "byPost", fields: ["id"])
}

type Comment @model
  @key(name: "byPost", fields: ["postID"]) {
  id: ID!
  postID: ID!
  content: String!
}
```

then I create post this part works.
But when I try to delete post then it errors out.
I haven't tried adding comments yet.
```
async function deletePost(id: string) {
    
    try {

      const original = await DataStore.query(Post, id)

      console.log(original)

      await DataStore.delete(original)

    } catch (err) {

      console.error(err)
    }
  }
```
I do get the original post but I just cannot delete it. I get following error

` Failed to execute 'index' on 'IDBObjectStore': The specified index was not found`

iartemiev:
resolves #5088 

Enable relational modeling using the `@key` directive

### DIFF
--- a/packages/datastore/src/util.ts
+++ b/packages/datastore/src/util.ts
@@ -134,7 +134,9 @@ export const establishRelation = (
 					modelName: fieldAttribute.type.model,
 					relationType: connectionType,
 					targetName: fieldAttribute.association['targetName'],
+					associatedWith: fieldAttribute.association['associatedWith'],
 				});
+
 				if (connectionType === 'BELONGS_TO') {
 					relationship[mKey].indexes.push(
 						fieldAttribute.association['targetName']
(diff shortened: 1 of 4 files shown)


packages/datastore/src/util.ts
```diff
@@ -134,14 +134,32 @@ export const establishRelation = (
...
+		// create indexes from key fields
+		if (model.attributes) {
+			model.attributes.forEach(attribute => {
+				if (attribute.type === 'key') {
+					const { fields } = attribute.properties;
+					if (fields) {
+						fields.forEach(field => {
+							// only add index if it hasn't already been added
+							const exists = relationship[mKey].indexes.includes(field);
+							!exists && relationship[mKey].indexes.push(field);
```
manueliglesias, line 157: what are your thoughts on changing this to an if statement?

I kind of agree with [this opinion](https://www.codereadability.com/dont-use-short-circuiting-when-you-want-an-if-statement/):

> However, unlike an if statement, the AND operator doesn't communicate a conditional operation to the developer who's reading the code.

iartemiev: That's fair. I agree

iartemiev: Updated

PR #5850, opened 2020-05-20
fix(@aws-amplify/auth): Fix OAuth multiple scopes

elorzafe:
_Issue #, if available:_
Multiple scopes were creating a wrong query string (separated by comma not space)
_Description of changes:_
Replacing `','` with `' '`


packages/auth/src/OAuth/OAuth.ts
```diff
@@ -55,6 +55,8 @@ export default class OAuth {
 		this._urlOpener = config.urlOpener || launchUri;
 		this._config = config;
 		this._cognitoClientId = cognitoClientId;
+
+		if (!Array.isArray(scopes)) throw Error('scopes must be a String Array');
```
undefobj, line 59: This is a check that `scopes` is an array but do we have to ensure that every element is actually a String, per your error message? If so I think a bit more is needed. Maybe  something like:

```
function elementsAreStrings(x) {
    return x.every(function(i){ return typeof i === "string" });
}
```

elorzafe: Totally makes sense, fixed!

Issue #5708, opened 2020-05-06
textToSpeech fails with special characters


bdpaul99:
**Describe the bug**

When using the Predictions library to 1. translate text to a new language, and then 2. use polly to synthesize that text to speech, the api call to polly fails when there are characters with accent marks in the string.

For example, if i want to translate the string "The number" to spanish, it successfully translates to "el número". Then the function call

Predictions.convert({
textToSpeech: {
source: {
text: "el número",
language: "es-MX"
},
voiceId: "Mia"
}
})

would fail, resulting in this error:
{ "name": "InvalidSignatureException", "$fault": "client", "$metadata": { "httpStatusCode": 403, "httpHeaders": { "content-type": "application/json", "date": "Wed, 06 May 2020 22:08:54 GMT", "x-amzn-errortype": "InvalidSignatureException:http://internal.amazon.com/coral/com.amazon.coral.service/", "x-amzn-requestid": "872289c6-e704-41c1-a2fe-5ce93aa954a6" }, "requestId": "872289c6-e704-41c1-a2fe-5ce93aa954a6", "retries": 3, "totalRetryDelay": 430 } }

When the accent mark is removed, it succeeds.

**What is Configured?**
If applicable, please provide what is configured for Amplify CLI:
* Which steps did you follow via Amplify CLI when configuring your resources.
$ npm install -g @aws-amplify/cli
$ amplify configure
$ amplify init
$ amplify add predictions
* Which resources do you have configured?
Predictions- Convert, translateText and textToSpeech

```
npx envinfo --system --binaries --browsers --npmPackages --npmGlobalPackages
```
 System:
    OS: macOS Mojave 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
    Memory: 311.01 MB / 8.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.8.0 - /usr/local/bin/node
    npm: 6.10.3 - /usr/local/bin/npm
  Browsers:
    Firefox: 68.6.1
    Safari: 13.1
  npmPackages:
    @testing-library/jest-dom: ^4.2.4 => 4.2.4 
    @testing-library/react: ^9.5.0 => 9.5.0 
    @testing-library/user-event: ^7.2.1 => 7.2.1 
    aws-amplify: ^3.0.10 => 3.0.10 
    aws-amplify-react: ^4.1.9 => 4.1.9 
[...]

Issue #5582, opened 2020-04-27
ES6 arrow function syntax breaking Internet Explorer 11 (IE11)


sawyerh:
**Describe the bug**

Amplify does not work in Internet Explorer 11. It's possible this was broken by a dependency like `@aws-sdk`. It's unclear how to restore IE11 support to an application implementing Amplify.

**To Reproduce**

Repro available here: https://github.com/sawyerh/amplify-nextjs-ssr

**Expected behavior**
It's unclear what the official browser support is for Amplify, however the Amplify SDK previously worked in IE 11, and according to [this comment](https://github.com/aws-amplify/amplify-js/issues/1195#issuecomment-502847380) AWS is still supporting IE11. 

**Screenshots**
![image](https://user-images.githubusercontent.com/371943/80402763-6769b000-888c-11ea-9e0b-160bca41e905.png)

**What is Configured?**

```
"dependencies": {
    "aws-amplify": "3.0.9",
    "aws-amplify-react": "4.1.8",
    "next": "9.3.5",
    "react": "16.13.1",
    "react-dom": "16.13.1"
  }
```

Amplifiyer:
Same as https://github.com/aws-amplify/amplify-js/issues/5405
The issue is with an upgrade of underlying aws-sdk that is exporting the package with ES2016 syntax https://github.com/aws/aws-sdk-js-v3/blob/v1.0.0-beta.5/packages/middleware-serde/tsconfig.json#L3 and there is an open issue related to this https://github.com/aws/aws-sdk-js-v3/issues/1099

ericclemmons:
Awaiting a version bump for https://www.npmjs.com/package/@aws-sdk/eventstream-serde-browser

slavco86:
Having same issue.
@ericclemmons 
How long till the version bump?

jonasduever:
Still an issue, I think. :/

```
  "dependencies": {
    ...
    "aws-amplify": "~3.0.12",
    "aws-amplify-angular": "~5.0.12",
    ...
  }
```

![image](https://user-images.githubusercontent.com/5775325/82896209-d999e700-9f55-11ea-9e15-1ec6f1c84e2d.png)

Amplifiyer:
it seems there is one more package that got missed which targets ES2016 output https://github.com/aws/aws-sdk-js-v3/blob/v1.0.0-gamma.2/packages/middleware-serde/tsconfig.json#L3
[...]

Issue #5534, opened 2020-04-24
Authentication with hosted UI fails on Edge 18 with TypeError: Invalid Argument


Stereobit:
**Describe the bug**
Authentication with any Identity prover fails when redirecting back to WebApp

**What AWS Services are you utilizing?**
Amplify, Auth

**Environment**
Web

**To Reproduce**

1. Initialise with `Auth.configure`
2. Redirect to Hosted UI
3. Authenticate
4. Redirect back
5. See error

**Expected behavior**
Authentication and removal of code query param, like in other browsers

**Code Snippet**
https://github.com/Stereobit/amplify-test

**Screenshots**
![image](https://user-images.githubusercontent.com/150493/80206115-b2cd5580-8623-11ea-8ff1-5799f6de38d4.png)

**What is Configured?**
```
Auth.configure({
  identityPoolId: "XX-XXXX-X:XXXXXXXX-XXXX-1234-abcd-1234567890ab",
  region: "eu-west-1",
  userPoolId: "XX-XXXX-X",
  userPoolWebClientId: "a1b2c3d4e5f6g7h8i9j0k1l2m3",
  oauth: {
    domain: "my_cognito_domain",
    scope: ["email"],
    redirectSignIn: "my_cognito_domain",
    responseType: "code",
  },
});
```
See repo for webpack & babel configs

<details>
  <summary><strong>Environment</strong></summary>

```
  System:
    OS: macOS 10.15.4
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
    Memory: 68.20 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.0.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 81.0.4044.122
    Safari: 13.1
  npmPackages:
    @aws-amplify/auth: ^3.2.4 => 3.2.4
    @babel/core: ^7.9.0 => 7.9.0
    @babel/polyfill: ^7.8.7 => 7.8.7
    @babel/preset-env: ^7.9.5 => 7.9.5
    babel-loader: ^8.1.0 => 8.1.0
    html-webpack-plugin: ^4.2.0 => 4.2.0
    webpack: ^4.43.0 => 4.43.0
    webpack-cli: ^3.3.11 => 3.3.11
[...]

Issue #5472, opened 2020-04-21
Predication API Error for comprehend and translate - Unhandled Rejection (InvalidSignatureException): The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.


wongcyrus:
**Describe the bug**
After upgrade to  "aws-amplify": "^3.0.8" and  "aws-amplify-react": "^4.1.6",
Prediction API for comprehend and translate with error.

Unhandled Rejection (InvalidSignatureException): The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.
(anonymous function)
/home/protocols/Aws_json1_1.ts:1090
  1087 | case "ExternalServiceException": return [3 /*break*/, 2];
  1088 | case "com.amazonaws.cognito.identity.model#ExternalServiceException": return [3 /*break*/, 2];
  1089 | case "InternalErrorException": return [3 /*break*/, 4];
> 1090 | case "com.amazonaws.cognito.identity.model#InternalErrorException": return [3 /*break*/, 4];
       | ^  1091 | case "InvalidParameterException": return [3 /*break*/, 6];
  1092 | case "com.amazonaws.cognito.identity.model#InvalidParameterException": return [3 /*break*/, 6];
  1093 | case "NotAuthorizedException": return [3 /*break*/, 8];
View compiled
▼ 3 stack frames were expanded.
step
node_modules/tslib/tslib.es6.js:100
next
node_modules/tslib/tslib.es6.js:81
fulfilled
node_modules/tslib/tslib.es6.js:71

      let language = await Predictions.interpret({
        text: {
          source: {
            text: this.state.question,
          },
          type: "LANGUAGE"
        }
      });
      sourceLanguage = language.textInterpretation.language;
      if (!sourceLanguage.startsWith("en")) {
        const translatedMessage = await Predictions.convert({
          translateText: {
            source: {
              text: this.state.question,
              language: sourceLanguage
            },
            targetLanguage: "en"
          }
        });
        message = translatedMessage.text;
        this.setState({ englishQuestion: message });
        console.log(this.state.question, message);
      }

I tried to refresh the new token but it does not work too.

Amplifiyer:
@wongcyrus, can you post your request headers and request body/payload here (remove any sensitive information such as credentials)

wongcyrus:
[...]

PR #5854, opened 2020-05-21
Upgrade aws-sdk to gamma version

Amplifiyer:
_Issue #, if available:_ 
* #5472
* #5708 
* #5387 
* #5582 
* #5405
* #5320 
* #5332
* #5534
* #5393 

_Description of changes:_

**Note: The build won't succeed until aws-sdk release their gamma version. I've been testing with a local copy of cached verdaccio release.**

This PR is to upgrade aws-sdk version which fixes the above mentioned amplify issues. There are some compatibility changes since this release of aws-sdk is not 100% backwards compatible.

1. Browser environment now requires the http response to be either a blob or a stream. Updated axios http handler to now always request a blob and return it. This also makes it consistent for react-native and browsers environment to always return blobs to amplify customers. Need to discuss since this would be a breaking change. See table below for reference

Release | Browser | React Native
------------ | ------------ | -------------
Amplify V2 | UInt8Array | UInt8Array
Amplify V3 | String (#5393) | Blob
This PR | Blob | Blob

2. Local mock storage now requires another parameter `tls` to enforce http endpoints.


packages/storage/src/providers/detectReactNative.native.ts
```diff
@@ -0,0 +1 @@
+export const isReactNative: boolean = true;
```
jordanranz, line 1: Is this something we can benefit from in other categories?

Amplifiyer: Realized that `Platform.isReactNative` already exists. Replacing this.

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -508,8 +509,9 @@ export class AWSS3Provider implements StorageProvider {
 		if (dangerouslyConnectToHttpEndpointForTesting) {
 			localTestingConfig = {
 				endpoint: localTestingStorageEndpoint,
-				s3BucketEndpoint: true,
-				s3ForcePathStyle: true,
+				tls: false,
+				bucketEndpoint: false,
```
Ashish-Nanda, line 513: Why was bucketEndpoint changed to false?

Amplifiyer:
> is to change the return type of `Storage.get` when `download: true` 

Thanks @elorzafe. The type that's changing is the result.Body. Is there any trick shortcut to specify a type definition that explicitly has one member of a specific type but otherwise generic?

packages/storage/src/providers/AWSS3Provider.ts
Amplifiyer: it's a change in the aws-sdk interpretation of `bucketEndpoint`

V3: https://github.com/aws/aws-sdk-js-v3/blob/c96f35f972c44706a391bb07e0a897e73b8d6dfe/packages/middleware-bucket-endpoint/src/bucketEndpointMiddleware.ts#L27

V2: https://github.com/aws/aws-sdk-js/blob/dfb49100f7db707a2479726f911ca31443f5a861/lib/services/s3.js#L404-L407

Ashish-Nanda: Ok, makes sense!

sammartinez: should we call this out in the readme?

Amplifiyer: This is a 100% non-breaking change for Amplify customers

masbaehr:
Just for reference: This update to gamma cause this critical issue, as all S3 uploads stopped working in Chrome:

https://github.com/aws-amplify/amplify-js/issues/5873

PR #5867, opened 2020-05-22
Handle launching an app from killed state through notifications in iOS

Ashish-Nanda:
_Description of changes:_
Handle launching an app from killed state through notifications in iOS

### DIFF
--- a/packages/pushnotification/src/PushNotification.ts
+++ b/packages/pushnotification/src/PushNotification.ts
@@ -178,14 +178,19 @@ export default class PushNotification {
 	/**
 	 * This function handles the React Native AppState change event
 	 * And checks if the app was launched by a Push Notification
+	 * Note: Current AppState will be null or 'unknown' if the app is coming from killed state to active
 	 * @param nextAppState The next state the app is changing to as part of the event
 	 */
 	_checkIfOpenedByNotification(nextAppState, handler) {
-		// the app is turned from background to foreground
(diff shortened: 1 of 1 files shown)


packages/pushnotification/src/PushNotification.ts
```diff
@@ -178,14 +178,19 @@ export default class PushNotification {
...
 	 * And checks if the app was launched by a Push Notification
+	 * Note: Current AppState will be null or 'unknown' if the app is coming from killed state to active
 	 * @param nextAppState The next state the app is changing to as part of the event
 	 */
 	_checkIfOpenedByNotification(nextAppState, handler) {
-		// the app is turned from background to foreground
+		// the App state changes from killed state to active
 		if (
-			this._currentState.match(/inactive|background/) &&
+			(!this._currentState || this._currentState === 'unknown') &&
```
sammartinez, line 187: Is 'unknown' true ? The currentState will hit 'unknown' and match this string?

sammartinez: as our offline convo, thanks for the clarity on this!

Ashish-Nanda: Thanks! Context for others: currentState will have the value `unknown` when the app is launched from killed state.

This is missing in the React Native docs but here is an issue that mentions it: https://github.com/facebook/react-native-website/issues/273

PR #5869, opened 2020-05-22
feat: Automatically handle clock drift/skew for API calls

### LINKED ISSUE #2014: correctClockSkew for API

### LINKED ISSUE #3719: Cognito Signature expired

ericclemmons:
Fixes #2014, Fixes #3719
Milestone: https://github.com/aws-amplify/amplify-js/milestone/17

- [x] Solves clock drift for `@aws-amplify/api-graphql` via the underlying `@aws-amplify/api-rest` category.

- [x] Improves testing for `DateUtils`

### Reproduction

1. With the correct time set, load example that makes unauthenticated IAM calls via `API.graphql`. (Notice API calls succeed)
	
	Screen Shot 2020-05-22 at 11 34 39 AM

1. Move the clock forward or back by 1+ hour(s).

	Screen Shot 2020-05-22 at 1 34 26 PM

1. Refresh the page and observe `API.graphql` returns a `400`.

1. Automatically, a 2nd request is made that succeeds, thanks to `DateUtils.getDateWithClockOffset()` using the difference between request/response dates.

1. Subsequent calls succeed, as new requests reference `DateUtils.getDateWithClockOffset()`.

I'm using my SSR example because it uses IAM for fetching guest credentials.
> skew

Pivotal: https://www.pivotaltracker.com/n/projects/2174488

---


packages/api-graphql/package.json
```diff
@@ -17,6 +17,8 @@
     "build-with-test": "npm test && npm run build",
     "build:cjs": "node ./build es5 && webpack && webpack --config ./webpack.config.dev.js",
     "build:esm": "node ./build es6",
+    "build:cjs:watch": "node ./build es5 --watch",
+    "build:esm:watch": "node ./build es6 --watch",
```
ericclemmons, line 21: I can break the addition of watch-mode for these packages out into another PR, if this one gets held up.

packages/api-rest/src/RestClient.ts
```diff
@@ -168,7 +169,27 @@ export class RestClient {
 
 		// Signing the request in case there credentials are available
 		return Credentials.get().then(
-			credentials => this._signed({ ...params }, credentials, isAllResponse),
+			credentials => {
+				return this._signed({ ...params }, credentials, isAllResponse).catch(
+					error => {
+						if (
+							error.response.headers['x-amzn-errortype'] ===
+								'BadRequestException' &&
```
ericclemmons, line 177: What's the best heuristic to determine when clock drift should be applied?

For `API.graphql` calls, there's an consistent "shape" that we could use (e.g. matching for `Signature not yet current`), but this can't be guaranteed for all calls:

```json
{
  "errors" : [ {
    "errorType" : "BadRequestException",
    "message" : "Signature not yet current: 20200522T004530Z is still later than 20200521T235032Z (20200521T234532Z + 5 min.)"
  } ]
}
```

packages/api-rest/src/RestClient.ts
```diff
@@ -168,7 +169,27 @@ export class RestClient {
...
+							DateUtils.getClockOffset() === 0
```
ericclemmons, line 178: For now, short-circuiting an infinite loop by only adjusting clock-drift _once_.  (A refresh would be needed to reset it).

Depending on how many APIs are affected, this will get moved & changed along with the heuristic above.

packages/api-rest/src/RestClient.ts
```diff
@@ -168,7 +169,27 @@ export class RestClient {
...
+						) {
+							const responseDate = new Date(error.response.headers.date);
+							const requestDate = DateUtils.getDateFromHeaderString(
+								params.headers['x-amz-date']
+							);
+							const offset = responseDate.getTime() - requestDate.getTime();
+
+							DateUtils.setClockOffset(offset);
```
ericclemmons, line 186: Reminder that `DateUtils` is yet another singleton, so the stored `offset` is globally available and used for subsequent requests.

packages/api-graphql/package.json
elorzafe: I think is fine

packages/api-rest/src/RestClient.ts
elorzafe: Why not passing ajax a param called that is not going to retry BadRequest, because the code will be coupled with DateUtils instead of handling that case internally

packages/api-rest/src/RestClient.ts
sammartinez: may be a good question for @Amplifiyer on this.

Amplifiyer: I think an added protection to be compare `existingClockDrift` !== `currentCalculatedClockDrift` i.e. 
```
DateUtils.getClockOffset() !== responseDate.getTime() - requestDate.getTime()
```
This kind of confirms that we are out of sync when making this particular request (even if clockOffset was previously set for an ever slowing clock)

packages/api-graphql/package.json
ericclemmons: This was a direct copy/paste from another `package.json`. Somehow these didn't get the `*` watch commands 🤷

ericclemmons:
Sorry, to be clear, I really need help with the **TODOS** section at the top of this PR, namely:
> Which APIs may are/aren't susceptible to clock drift? (Each of these will need to be tested with/without this PR)

The code **will** change, but only served to prove that clockdrift was solveable via `DateUtils` that was put in place waaay back in November.

packages/api-rest/src/RestClient.ts
ericclemmons: I can investigate this.  It'll likely require some rounding, because milliseconds are thrown away for `x-amz-date`.  

Back to the original callout: I can't say that `BadRequestException` _only_ exists for clockdrift scenarios: I recall it appearing when the signature was incorrect for whatever reason.

Because of this, there's risk in recursively retrying requests when drift was one of perhaps many issues.

The other question I ask in my TODOS is what degree this value persists/expires/recalculates.

ericclemmons:
Just tested `Auth.federatedSignIn()` with an incorrect date, and Cognito fails with:
> Screen Shot 2020-05-27 at 12 25 03 PM

It **strongly** appears that Cognito Hosted UI doesn't allow for > 5 minutes of click drift.

ericclemmons:
In testing `Analytics.send` (which fails when the clock is skewed with a `403` with the same "Signature not yet current" error), I noticed that the AWS SDK has a helpful method for checking if the clock is skewed:

https://github.com/aws/aws-sdk-js-v3/blob/4895ac7070cf6d845846e0c3b40aa4ee6ad94f82/packages/middleware-signing/src/middleware.ts#L13-L15

So the fix for #3699 would be different.  What's interesting is that clock skew is supposed to already be taken care of within the AWS SDK, but it _does_ use a different signer:

https://github.com/aws/aws-sdk-js-v3/blob/4895ac7070cf6d845846e0c3b40aa4ee6ad94f82/clients/client-pinpoint/PinpointClient.ts#L830-L831

More experimentation can be done via:

```diff
diff --git a/packages/analytics/src/Providers/AWSPinpointProvider.ts b/packages/analytics/src/Providers/AWSPinpointProvider.ts
index 11e250784..1d16620bb 100644
--- a/packages/analytics/src/Providers/AWSPinpointProvider.ts
+++ b/packages/analytics/src/Providers/AWSPinpointProvider.ts
@@ -579,6 +579,8 @@ export class AWSPinpointProvider implements AnalyticsProvider {
                                );
                                return updateEndPointData;
                        } catch (err) {
+                               console.error('error', err, err.response);
+
                                logger.debug('Failed to update the endpoint', err);
                                throw err;
                        }
@@ -673,6 +675,26 @@ export class AWSPinpointProvider implements AnalyticsProvider {
        }

        private _customizePinpointClientReq() {
+               (this.pinpointClient as PinpointClient).middlewareStack.add(
+                       (next, context) => {
+                               return async args => {
+                                       console.log({ next, context, args });
+                                       try {
+                                               return await next(args);
+                                       } catch (error) {
+                                               console.error(error, JSON.stringify(error));
+                                               debugger;
+                                               throw error;
+                                       }
+                               };
+                       },
+                       {
+                               name: 'clickSkewMiddlware',
+                               step: 'finalizeRequest',
+                               tags: ['RETRY'],
+                       }
+               );
+
                // TODO FIXME: Find a middleware to do this with AWS V3 SDK
                // if (Platform.isReactNative) {
                //      this.pinpointClient.customizeRequests(request => {

```

The problem I ran into is the thrown `error` doesn't have access to the response body anymore, so I can't manually adjust clock drift. However, this should already be taken care of:

https://github.com/aws/aws-sdk-js-v3/blob/4895ac7070cf6d845846e0c3b40aa4ee6ad94f82/packages/middleware-signing/src/middleware.ts#L45-L48

ericclemmons:
This PR is scoped & titled to fixing clock skew for API calls, so I'm going to leave this scope as-is (rather than solving for `Auth` or `Analytics` bugs for now) in the interest of improving iteratively, vs. blocking for completeness.

`aws-sdk-js` v3 solves for clock skew for most cases, but API uses it's own signer + request implementation, which is why it appears here.

I'll update this PR and get it ready for a proper review...

packages/api-rest/src/RestClient.ts
ericclemmons: The AWS SDK does a good job with the middleware pattern of supporting this, so I'll borrow some abstractions from there to leverage this.


It'll look something like:

```js
if (isClockSkewError(error) && DateUtils.isClockSkewed())
```

https://github.com/aws/aws-sdk-js-v3/blob/4bc605be4eda41b47ecf44bdea31d289f88aa699/packages/middleware-signing/src/middleware.ts#L13-L15

https://github.com/aws/aws-sdk-js-v3/blob/4bc605be4eda41b47ecf44bdea31d289f88aa699/packages/middleware-signing/src/middleware.ts#L13-L15

ericclemmons:
Update this PR with additional tests.

packages/api-rest/src/RestClient.ts
```diff
@@ -168,7 +169,30 @@ export class RestClient {
...
+						}
+
+						const { headers } = error.response;
+						const dateHeader = headers && (headers.date || headers.Date);
+						const responseDate = new Date(dateHeader);
+						const requestDate = DateUtils.getDateFromHeaderString(
+							params.headers['x-amz-date']
+						);
+
+						if (DateUtils.isClockSkewed(requestDate, responseDate)) {
```
Amplifiyer, line 186: Should we throw error again in the else case?

packages/core/src/Util/DateUtils.ts
```diff
@@ -26,6 +28,48 @@ export const DateUtils = {
...
+				Number(second)
+			)
+		);
+	},
+
+	isClockSkewed(serverDate: Date) {
+		return (
+			Math.abs(
+				serverDate.getTime() - DateUtils.getDateWithClockOffset().getTime()
+			) >= FIVE_MINUTES_IN_MS
```
Amplifiyer, line 56: I think a comment here would be useful in future to understand the reasoning to use 5 minutes

packages/api-rest/src/RestClient.ts
ericclemmons: Yes. I'm going to restructure this to return `this.ajax` only when both conditions pass.

packages/core/src/Util/DateUtils.ts
ericclemmons: Resolved via 8353bdab50bd71f0c1c4e15fe5c7a3cd359ade11

MedElba:
Hi everyone!

I'm using amplify on a React Native application and iOS gives sometimes this 403 errors, this cannot even be resolved by a Signout and Signin! I've seen you wrote about the clockdrift problem and that you merged the fix! I'm using the latest version but it is still present! Any idea on how to solve?
Really urgent for a in-production iPhone Application. 

@ericclemmons  Is it still the same issue?

Thank you very much!

PR #5877, opened 2020-05-23
fix(amazon-cognito-identity-js): allow storage to return missing items

duarten:
_Issue #, if available:_

_Description of changes:_

Change the typings for ICognitoStorage so getItem can return missing,
and thus undefined, values.


duarten:
ping

packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -49,7 +49,7 @@ declare module 'amazon-cognito-identity-js' {
 
 	export interface ICognitoStorage {
 		setItem(key: string, value: string): void;
-		getItem(key: string): string;
+		getItem(key: string): string | null;
```
amhinson, line 52: Do you think `undefined` might be better here? Here's the reference to what the code is doing: https://github.com/aws-amplify/amplify-js/blob/4f916432c94e814c720ce3bf763d2908c906a799/packages/amazon-cognito-identity-js/src/StorageHelper.js#L42

duarten: Seems like a bug in `MemoryStorage`. Here's the `Storage` interface definition that `ICognitoStorage` is wrapping: https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L15286

duarten: And the documentation: https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem

amhinson: Ah yes, you're right! We are looking to improve the TypeScript definitions in the future, so we will take that into account. Seems like there are a few different implementations for the different Storage plugins.

PR #5994, opened 2020-06-03
Improve types for GraphQLResult errors

neilime:
_Issue #, if available:_

```ts
import { GraphQLResult } from '@aws-amplify/api-graphql';

const result = { errors: [{ message: 'test' }] } as GraphQLResult;

if (result.errors) {
  console.log(result.errors[0].message); 
  // Typescript error `Property 'message' does not exist on type 'object'`
}
```

_Description of changes:_


packages/api-graphql/src/types/index.ts
```diff
@@ -28,7 +28,7 @@ export enum GRAPHQL_AUTH_MODE {
 
 export interface GraphQLResult<T = object> {
 	data?: T;
-	errors?: [object];
+	errors?: Array<{ [key: string]: any; }>;
```
amhinson, line 31: Thanks for the update @neilime! How do you feel about using `GraphQLError` from `graphql` so we can be more specific about what `errors` is:
https://github.com/aws-amplify/amplify-js/blob/40c4e21b8cab18dd203e942083d8b6e16ed06997/packages/api-graphql/src/GraphQLAPI.ts#L291

Maybe something like this:
```suggestion
	errors?: GraphQLError[];
```

and you might need to import it like this:
```js
import { GraphQLError } from 'graphql/error/GraphQLError';
```

I haven't tested this locally, but let me know what you find!

neilime: @amhinson using GraphQLError is better and it works !

PR #6030, opened 2020-06-06
Fix - PubSub realtime provider not reset when configuration endpoint changes

ume-daniel:
When PubSub is initialized, it does a lazy initial instantiation but on subsequent updates of the configuration such as a client changing which AppsyncAPI endpoint they are using, the realtime provider does not update. Subsequent subscriptions will be added to the initial provider, not the new provider.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
 		"typedoc": "^0.11.0",
 		"typescript": "~3.6.4",
 		"uglifyjs-webpack-plugin": "^0.4.6",
+		"uuid-validate": "^0.0.3",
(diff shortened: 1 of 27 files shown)


amhinson:
Is this PR still valid? I only see one commit from 12/3/19.

ume-daniel:
> 
> 
> Is this PR still valid? I only see one commit from 12/3/19.

I do believe the issue still exists but the workaround is straightforward

amhinson:
Would you be able to update this PR with the fix? It seems like right now it is only pulling in an old commit.

PR #6042, opened 2020-06-09
feat(core): expose observable for testing

iartemiev:
Exposing observable as a private static property in order to allow testing offline scenarios in integration tests.

Allows us to simulate offline mode (as far as DataStore Sync is concerned) by adding the following to our sample apps
```js
import { Reachability } from '@aws-amplify/core';
// go offline
Reachability._observer.next({online: false});
// come back online
Reachability._observer.next({online: true});
```


packages/core/src/Util/Reachability.ts
```diff
@@ -15,12 +18,28 @@ export default class ReachabilityNavigator implements Reachability {
...
 			};
 		});
 	}
+
+	// expose observers to simulate offline mode for integration testing
+	private static _observerOverride(status: NetworkStatus): void {
+		for (const _observer of ReachabilityNavigator._observers) {
+			if (!(_observer.next && typeof _observer.next === 'function')) {
+				continue;
+			}
```
manueliglesias, line 39: Since we re the only ones adding observers to `ReachabilityNavigator._observers`, I think the only check we need to do is that the observable is still open:

```suggestion
			if (!_observer.closed)) {
				// This would be a good place and time to remove it from ReachabilityNavigator._observers
				continue;
			}
```

packages/core/src/Util/Reachability.ts
```diff
@@ -1,10 +1,13 @@
-import Observable from 'zen-observable-ts';
+import Observable, { Observer } from 'zen-observable-ts';
 
 type NetworkStatus = {
 	online: boolean;
 };
 
 export default class ReachabilityNavigator implements Reachability {
+	private static _observers: Array<Observer<NetworkStatus>> = [];
+	private _observer: Observer<NetworkStatus> = null;
```
manueliglesias, line 9: We don't need to keep this one as an instance variable, on the unsubscribe function we will have access to the `observer` since a closure is made.

See next comment

packages/core/src/Util/Reachability.ts
```diff
@@ -15,12 +18,28 @@ export default class ReachabilityNavigator implements Reachability {
...
 
+			this._observer = observer;
+			ReachabilityNavigator._observers.push(observer);
+
 			return () => {
 				window.removeEventListener('online', notifyOnline);
 				window.removeEventListener('offline', notifyOffline);
+				ReachabilityNavigator._observers = ReachabilityNavigator._observers.filter(
+					_observer => _observer !== this._observer
+				);
```
manueliglesias, line 29:
```suggestion
				ReachabilityNavigator._observers = ReachabilityNavigator._observers.filter(
					_observer => _observer !== observer
				);
```

packages/core/src/Util/Reachability.ts
manueliglesias: BTW, my `if` check is wrong, we should continue if it is closed

PR #6044, opened 2020-06-09
add missing optional parameter clientMetadata to CognitoUser refreshSession 

pcasari:
_Issue #, if available:_

_Description of changes: add missing optional parameter clientMetadata to CognitoUser refreshSession 


pcasari:
> Thanks for the addition @pcasari! Could you update this to use tabs instead of spaces on line 81?

Thansk for the reply @amhinson!
I updated the pr with the correct spacing.

PR #6055, opened 2020-06-11
CircleCI Android Detox Support

amhinson:
_Description of changes:_
This PR adds Android Detox support for CircleCI and runs the React Native Storage integration tests.

Successful pipeline with Android integration tests running: https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/5291/workflows/f71c4685-e6d8-4a6c-95d5-545d0806d8aa

---


.circleci/android-rn.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+case $1 in
+  export-env)
+    echo 'export PATH="$PATH:/usr/local/opt/node@10/bin:~/.yarn/bin:~/project/node_modules/.bin:~/project/example/node_modules/.bin"' >> $BASH_ENV
```
sammartinez, line 5: Just for clarity on the echo statements, is there anything that would expose the workspace that we do not want exposed?

amhinson: Not that I'm aware of. Nothing shows in the [Configure Environment Variables step](https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/5291/workflows/f71c4685-e6d8-4a6c-95d5-545d0806d8aa/jobs/17054).

.circleci/config.yml
```diff
@@ -17,6 +17,10 @@ executors:
       - image: verdaccio/verdaccio
     resource_class: large
 
+  android-executor:
+    macos:
```
Ashish-Nanda, line 21: I remember you had the docker images from react-native-community before and I've seen other images for android in circleci configs as well. Any reason to pick macos over those?
I would assume those images are better set up for react-native/android.

.circleci/config.yml
```diff
@@ -209,6 +213,103 @@ commands:
...
+            avdmanager create avd --force --name TestingAVD
+            --package "system-images;android-29;default;x86_64"
+            --tag default --device pixel
+          name: Create Android Emulator
+      - run:
+          background: true
+          command: |
+            cd ~/amplify-js
+            ./.circleci/android-rn.sh start-emulator
+          name: Start Android Emulator (background)
```
Ashish-Nanda, line 288: Do we need to start the emulator manually?
What if we use npx react-native run-android, wont that start the packager and AVD for us?

Ashish-Nanda:
Thanks for getting this set up. Just had a few comments about the config and whether we can reduce some of the steps.

.circleci/config.yml
amhinson: I got it really close to working with the `reactnativecommunity/react-native-android` image, but there just were a number of issues working with the Android SDK trying to get the emulator to work correctly. There was also a limitation with not being able to use an `x86` or `x86_64` emulator, so it would have had to run with SDK 24 instead of the latest (29).

Overall, I think those images work great for building Android apps, but it just wasn't quite setup to run Detox tests easily.

.circleci/config.yml
amhinson: There were some additional parameters when running `emulator` that I came across that were recommended when running the emulator in a CI environment (see `start-emulator` in `android-rn.sh`), I assume to cut down on startup time. 

Also, it takes much longer for the emulator to load in CI, so this step is doing it manually and then the next `wait-for-avd` step is needed so the rest of the pipeline doesn't timeout.

Ashish-Nanda: Makes sense. Would be great to cut down on the time if possible in the future.

amhinson: I'd say there still room for optimizing this in the future once we start working with it a bit 👍

PR #6131, opened 2020-06-21
Adds lodash as an explicit dependency.

CharlesStover:
Using Yarn2, I get the following (valid) error:

```
Module not found: @aws-amplify/analytics tried to access lodash, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

You can see imports from `lodash` in both of these files:
* [AmazonPersonalizeProvider.ts](https://github.com/aws-amplify/amplify-js/blob/e85640a4e18fa1d55411038fee58919ad73121fa/packages/analytics/src/Providers/AmazonPersonalizeProvider.ts)
* [SessionInfoManager.ts](https://github.com/aws-amplify/amplify-js/blob/e85640a4e18fa1d55411038fee58919ad73121fa/packages/analytics/src/Providers/AmazonPersonalizeHelper/SessionInfoManager.ts)


amhinson:
@CharlesStover thanks for the PR! After talking with @iartemiev offline, what do you think about adding these 3 functions that are being used directly as dependencies instead of the entire library? 

For example: 
https://github.com/aws-amplify/amplify-js/blob/d6035dd576c9d62790e68bad616b6dcdfac4b2c7/packages/aws-amplify-vue/package.json#L16

https://github.com/aws-amplify/amplify-js/blob/e85640a4e18fa1d55411038fee58919ad73121fa/packages/aws-amplify-vue/src/components/authenticator/SignUp.vue#L63

packages/analytics/src/Providers/AmazonPersonalizeHelper/SessionInfoManager.ts
```diff
@@ -10,10 +10,11 @@
  * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
  * and limitations under the License.
  */
-import { RequestParams, SessionInfo } from './DataType';
-import { isEmpty, isEqual } from 'lodash';
+import { SessionInfo } from './DataType';
+import isEmpty from 'lodash/isEmpty';
+import isEqual from 'lodash/isEqual';
 import { v1 as uuid } from 'uuid';
-import { StorageHelper, ConsoleLogger as Logger, JS } from '@aws-amplify/core';
```
sammartinez, line 16: Do we want to remove the `StorageHelper` here?

amhinson: It wasn't being used (along with `RequestParams`) so I just removed them.

PR #5450, opened 2020-04-17
feat(@aws-amplify/datastore): Support SSR

ericclemmons:
## What

Add Next.js support (#5435) for `DataStore`:

```tsx
import { Amplify, DataStore } from 'aws-amplify';
import { useState, useEffect } from 'react';

import awsConfig from '../src/aws-exports';
import { Post } from '../src/models';

Amplify.configure(awsConfig);

type Props = {
	posts: Post[];
};

export default function Home(initialData: Props) {
	// @ts-ignore Property 'fromJSON' does not exist on type 'typeof Post'.ts(2339)
	const [posts, setPosts] = useState(Post.fromJSON(initialData.posts));

	useEffect(() => {
		const subscription = DataStore.observe<Post>(Post).subscribe(
			({ element, opType }) => {
				const opTypeUpdaters = {
					DELETE: (posts) => posts.filter((post) => post.id !== element.id),
					INSERT: (posts) => [element, ...posts],
					UPDATE: (posts) =>
						posts.map((post) => (post.id === element.id ? element : post)),
				};

				setPosts(opTypeUpdaters[opType]);
			}
		);

		return () => subscription.unsubscribe();
	}, []);

	return ...
}

export async function getServerSideProps(context) {
	const posts = await DataStore.query<Post>(Post);

	return {
		props: {
			posts: DataStore.toJSON(posts),
		},
	};
}
```

## Why

https://www.pivotaltracker.com/n/projects/2174488

## How

- [x] Add `DataStore` for `import { DataStore, Predicates } from "aws-amplify"` #5421 
- [x] Support `WebSocket`s on the server.
- [x] Add `DataStore.toJSON()` to automatically serialize models on the server.

	I would've preferred `model.toJSON()` so that `JSON.stringify(posts)` worked implicitly & automatically, but that only serializes _properties_ of `Post`, rather than the object itself.  The result is: `[Post {}, Post {}, Post {}]` 

	To solve for this, we would have to provide an `Array`-like wrapper with `.toJSON()` to cast `Post` as a plain `Object`.

	Alternatives for users would be to `JSON.parse(JSON.stringify(posts))`, but it's likely that we'll need to support serialization/deserialization based on the `schema`.

- [x] Add `Model.fromJSON()` to cast JSON as a new model instance (including `id`)

- [ ] Generate types for `Model.fromJSON()`:

[...]

PR #6146, opened 2020-06-23
feat(SSR): withSSRContext

ericclemmons:
_Issue #, if available:_ #5435 

This is an ☂️ PR that combines https://github.com/aws-amplify/amplify-js/pull/5710 & https://github.com/aws-amplify/amplify-js/pull/5450 to validate a `withServerContext` utility that scopes categories (e.g. `Auth`, and `DataStore`) to a single request.

- [x] #5710
- [x] #5450
- [x] Add `withServerContext(context)` for new `Auth` & `DataStore` instances.
- [x] Should DataStore's `sync` be scoped to `this` @manueliglesias?
- [x] Set `UniversalStorage` adapter for `Auth`.
- [x] `Amplify.*` properties are registered once.
- [x] Rename `withServerContext` to `withSSRContext`
- [x] **Revert 123e90c86bb54db335e6141578efbd217169bb9a**
- [x] Publish to a tag (e.g. `preview`) for testing.
- [x] Notify customers in how to install & use.
- [x] Fix references to `await Auth.*` that should use `this.amplify.Auth` or `this.Credentials` in DataStore's `subscription.ts` and `GraphqlAPI.ts`.
- [x] ~~Change `Post.fromJSON` to `DataStore.fromJSON(Post, initialData.posts)`~~
- [x] Wait 2-4 weeks for incorporating customer feedback.
- [x] Fix bug where federated sign in doesn't immediately set cookies.
- [ ] Passing test suite (makes more sense to solve this _after_ changes)
- [ ] Ship to `latest`
- [ ] Update docs with Next.js “framework support”.

    * Note that import { Auth } is still entirely valid usage for SSR, so existing docs may not require huge changes.
    * However, authenticated requests will require withSSRContext usage, so this may be more of a callout/alert, similar to React Native special instructions.

### Auth Example:

```tsx
import { Amplify } from "aws-amplify"
import awsconfig from '../src/aws-exports'

Amplify.configure(awsconfig)

// On the client, `Auth` is configured to be "SSR-aware"
const { Auth } = withServerContext()

export default function Home({ user }) {
  if (!user) {
	return <Authenticator />
  }

  return <h1>Howdy, {user.attributes.email}!</h1>	
}

export const getServerSideProps: GetServerSideProps = async (context) => {
  // Server gets an instance of Auth scoped to this request (including cookies)
  const { Auth } = withServerContext(context)

  let user = null

  try {
    user = await Auth.currentAuthenticatedUser()
  } catch (error) {
	// Usually "not authenticated
    console.error(error)
  }

  return {
    props: {
      user: JSON.parse(JSON.stringify(user)),
    },
  }
}
```

### DataStore Example:

```tsx
import { Amplify, withServerContext } from 'aws-amplify';
import awsConfig from '../src/aws-exports';
import { Post, PostStatus } from '../src/models';

Amplify.configure(awsConfig);

export async function getServerSideProps(context) {
  const { DataStore, serializeModel } = withSSRContext(context);
  cosnt posts = await DataStore.query(Post)
  
  return {
    props: {
      posts: serializeModel(posts)
    }
  }
}

const { DataStore, deserializeModel } = withSSRContext();

function Page(initialData) {
  const [posts] = useState(deserializeModel(Post, initialData.posts));
  
```

I put logging into each `DataStore` and storage class to confirm new instantiations took place:

```console
[ event ] build page: /
[ wait ]  compiling ...
[ info ]  bundled successfully, waiting for typecheck results...
[ ready ] compiled successfully - ready on http://localhost:3000
NEW DataStore
NEW DataStore
NEW AsyncStorage
NEW StorageClass
NEW ExclusiveStorage
NEW AsyncStorage
NEW AsyncStorageDatabase
[ event ] build page: /next/dist/pages/_error
[ wait ]  compiling ...
[ info ]  bundled successfully, waiting for typecheck results...
The schema has already been initialized


NEW PAGE ...

The schema has already been initialized
NEW DataStore
NEW StorageClass
NEW ExclusiveStorage
NEW AsyncStorage
NEW AsyncStorageDatabase


ANOTHER REFRESH

NEW DataStore
NEW StorageClass
NEW ExclusiveStorage
NEW AsyncStorage
NEW AsyncStorageDatabase
```

Between refreshes, the client will see (because it uses IndexedDB):

```console
NEW DataStore
NEW StorageClass
NEW ExclusiveStorage
```

---

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license._


manueliglesias:
>     * Should DataStore's `sync` be scoped to `this` @manueliglesias?

Yeah, probably a good chunk of these should be moved to instance variables.

https://github.com/aws-amplify/amplify-js/blob/af53902a660c1f65a442c878502c9feb67a3bb9f/packages/datastore/src/datastore/datastore.ts#L372-L378

ericclemmons:
@manueliglesias No problem. I did that in 952bedf, but that moved _everything_ around (not just those variables).

I'll move the minimum this time...

packages/core/src/UniversalStorage/index.ts
```diff
@@ -0,0 +1,111 @@
...
+
+	protected setUniversalItem(key: keyof Store, value: string) {
+		// @ts-ignore Argument of type 'Record<string, string>' is not assignable to parameter of type 'Pick<any, "res"> | { res: any; }'.
+		// Property 'res' is missing in type 'Record<string, string>' but required in type '{ res: any; }'.ts(2345)
+		nookies.set(this.store, key, value, {
+			// `httpOnly` cannot be set via JavaScript: https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#JavaScript_access_using_Document.cookie
+			sameSite: true,
+			// Allow unsecure requests to http://localhost:3000/ when in development.
+			secure: process.env.NODE_ENV === 'development' ? false : true,
+		});
```
ericclemmons, line 109: Because the browser acts as primary and the cookies are replicas, JavaScript **cannot** set `httpOnly`.  JavaScript can't _read_ `document.cookie`, nor write/create. **Only the server can send an `httpOnly` cookie.**

The current implementation yields this result:
> Screen Shot 2020-07-01 at 12 45 40 PM

Given that `accessToken` is readible via `document.cookie`, this may be warrant encryption. **However, `window.localStorage` has these exact same values readable via JavaScript _already_, so this isn't technically new**, with the exception that these values are sent to the server as well.

packages/core/src/Amplify.ts
```diff
@@ -2,34 +2,34 @@ import { ConsoleLogger as LoggerClass } from './Logger';
...
-	static I18n = null;
-	static Cache = null;
-	static PubSub = null;
-	static Interactions = null;
-	static Pushnotification = null;
-	static UI = null;
-	static XR = null;
-	static Predictions = null;
-	static DataStore = null;
+	Auth = null;
```
sammartinez, line 11: Changing from `static` does this introduce the issue we previously had around have two instances of Amplify ?

ericclemmons: Can you share some background on that issue?

My experience with two copies of Amplify was with `yarn link` and there being 2 different `aws-amplify` references.

sammartinez: When customers configure auth and go about their day with only needing this use case, everything is happy. When the customer goes to add Analytics and configures, it leads to a separate instance than the Auth instance present and can cause conflicts or error messages as "No Credentials, auth configure".

ericclemmons: `Amplify` is still exported as a singleton via `new AmplifyClass` (whereas before it was `export { Amplify }`.

On the client, the exported singletons (e.g. `Amplify`, `Auth`, `Storage`) are all associated together, but new instances _are not_.  (Otherwise, multiple requests would leak to the `Amplify.*` object)

If you have a code snippet or issue for reference, I can test it out.

The goal here was keeping clients working as-is, but keeping new instances separate.

sammartinez: Sounds good, let me find the issue from the past and I will paste it here

packages/aws-amplify/__tests__/withServerContext-test.ts
```diff
@@ -0,0 +1,83 @@
...
+			const amplify = withServerContext();
+			const config = amplify.configure();
+
+			expect(Amplify.Auth.Credentials).not.toBe(config.Credentials);
+			expect(amplify.Auth.Credentials).toBe(config.Credentials);
+		});
+	});
+
+	describe('DataStore', () => {
+		it('should be a different instance than Amplify.Auth', () => {
```
sammartinez, line 73:
```suggestion
		it('should be a different instance than Amplify.DataStore', () => {
```

packages/api-graphql/src/GraphQLAPI.ts
```diff
@@ -46,13 +46,14 @@ export class GraphQLAPIClass {
 	private _options;
 	private _api = null;
 
+	Credentials = Credentials;
+
```
ericclemmons, line 50: This should move from being `Credentials` to `amplify = Amplify`.

packages/datastore/src/datastore/datastore.ts
```diff
@@ -334,6 +336,14 @@ const createModelClass = <T extends PersistentModel>(
...
 			});
 		}
+
+		static fromJSON(json: T | T[]) {
+			if (Array.isArray(json)) {
+				return json.map(init => this.fromJSON(init));
+			}
+
+			return modelInstanceCreator(clazz, json);
+		}
```
ericclemmons, line 346: From my discussion with @manueliglesias on July 10th, this is going to get converted to support this example:

```js
export async function getServerSideProps(context) {
  const { DataStore, serializeModel } = withSSRContext(context);
  cosnt posts = await DataStore.query(Post)
  
  return {
    props: {
      posts: serializeModel(posts)
    }
  }
}

const { DataStore, deserializeModel } = withSSRContext();

function Page(initialData) {
  const [posts] = useState(deserializeModel(Post, initialData.posts));
  
```

**The only problem is that there needs to be access to `modelInstanceCreator` for this to work...**.

packages/amazon-cognito-identity-js/package.json
```diff
@@ -64,6 +64,7 @@
   "dependencies": {
     "buffer": "4.9.1",
     "crypto-js": "^3.3.0",
+    "isomorphic-unfetch": "^3.0.0",
```
ericclemmons, line 67: Necessary on the server due to usage of `fetch`:

https://github.com/aws-amplify/amplify-js/blob/38350d7e581d4341ca569ecea57988ff38779faf/packages/amazon-cognito-identity-js/src/Client.js#L55-L73

package.json
```diff
@@ -12,6 +12,7 @@
 		"coverage": "codecov || exit 0",
 		"docs": "typedoc packages/**/src --name amplify-js --hideGenerator --excludePrivate --ignoreCompilerErrors --mode file --out docs/api --theme docs/amplify-theme/typedoc/ --readme README.md",
 		"build": "lerna run build --stream",
+		"build:watch": "concurrently 'lerna run build:cjs:watch --parallel' 'lerna run build:esm:watch --parallel' --raw",
```
ericclemmons, line 15: It's a crapshoot whether CRA/Next/whatever wants the ESM or CJS build when `yarn link`ing.  This ensures we build both 🤷

packages/amplify-ui-angular/tsconfig.json
```diff
@@ -8,6 +8,7 @@
 		"fullTemplateTypeCheck": false
 	},
 	"compilerOptions": {
+		"skipLibCheck": true,
```
ericclemmons, line 11: Angular's was failing due to `nookies`' reference to `@types/express` and `@types/cookies`, even though those are devDependencies.

GitHub issues & Stackoverflow point to this being a common problem with Angular's strictness.

packages/analytics/src/Analytics.ts
```diff
@@ -75,7 +75,6 @@ export class AnalyticsClass {
 		Hub.listen('auth', listener);
 		Hub.listen('storage', listener);
 		Hub.listen('analytics', listener);
-		Amplify.register(this);
```
ericclemmons, line 78: This is moved outside of the constructor because we want there to be a single instance of `Analytics` registered with the global `Amplify`.

New instances of `Analytics` on the server shouldn't override the first one.

packages/amazon-cognito-identity-js/package.json
sammartinez: Im good with this @elorzafe what do you think?

packages/auth/src/index.ts
```diff
@@ -26,6 +26,7 @@ import { AuthErrorStrings } from './common/AuthErrorStrings';
 export default Auth;
 export {
 	Auth,
+	AuthClass,
```
sammartinez, line 29: Por que?

packages/datastore/src/datastore/datastore.ts
sammartinez: @iartemiev can you weigh in on this a little more since Manuel is out?

packages/amazon-cognito-identity-js/package.json
ericclemmons: We'll be doing build comparisons to determine if there are any bundle size concerns.

By design, this package is a polyfill that falls back to `window.fetch` on the browser.

packages/api-graphql/src/GraphQLAPI.ts
```diff
@@ -46,13 +44,14 @@ export class GraphQLAPIClass {
...
 
+	amplify = Amplify;
+
 	/**
 	 * Initialize GraphQL API with AWS configuration
 	 * @param {Object} options - Configuration object for API
 	 */
 	constructor(options) {
 		this._options = options;
-		Amplify.register(this);
```
ericclemmons, line 55: Again, new instances should be registered to the global `Amplify`, only the exported singleton.

packages/api-rest/src/RestAPI.ts
```diff
@@ -26,13 +26,14 @@ export class RestAPIClass {
...
 
+	amplify = Amplify;
+
 	/**
 	 * Initialize Rest API with AWS configuration
 	 * @param {Object} options - Configuration object for API
 	 */
 	constructor(options) {
 		this._options = options;
-		Amplify.register(this);
```
ericclemmons, line 35: Again, new instances should be registered to the global `Amplify`, only the exported singleton.

packages/api/src/API.ts
```diff
@@ -42,7 +42,6 @@ export class APIClass {
 		this._options = options;
 		this._restApi = new RestAPIClass(options);
 		this._graphqlApi = new GraphQLAPIClass(options);
-		Amplify.register(this);
```
ericclemmons, line 45: Again, new instances should be registered to the global `Amplify`, only the exported singleton.

packages/auth/src/Auth.ts
```diff
@@ -116,7 +118,6 @@ export class AuthClass {
 					break;
 			}
 		});
-		Amplify.register(this);
```
ericclemmons, line 119: Again, new instances should be registered to the global `Amplify`, only the exported singleton.

packages/aws-amplify/src/withSSRContext.ts
```diff
@@ -0,0 +1,88 @@
...
+	model: PersistentModelConstructor<T>,
+	init: T | T[]
+) {
+	if (Array.isArray(init)) {
+		return init.map(init => deserializeModel(model, init));
+	}
+
+	// `fromJSON` is intentionally hidden from types as a "private" method (though it exists on the instance)
+	// @ts-ignore Property 'fromJSON' does not exist on type 'PersistentModelConstructor<T>'.ts(2339)
+	return model.fromJSON(init);
```
ericclemmons, line 24: TODO: Add screenshot of what intellisense sees.

Danm72:
>const { Auth } = withServerContext()
The example up top uses the old syntax right? Its now withSSRContext

packages/auth/src/Auth.ts
```diff
@@ -94,6 +93,8 @@ export class AuthClass {
 	private _storageSync;
 	private oAuthFlowInProgress: boolean = false;
 
+	amplify = Amplify;
```
ericclemmons, line 96: We can't rely on `amplify` to be a service registry / container because it's not aware that `Credentials` is a required dependency of `Auth`. (i.e. Instantiating `Auth` needs to also create `Credentials`)

```suggestion
	credentials = Credentials;
```

packages/datastore/src/datastore/datastore.ts
```diff
@@ -973,21 +509,511 @@ function getNamespace(): SchemaNamespace {
...
+			const fullSyncIntervalInMilliseconds = this.fullSyncInterval * 1000 * 60; // fullSyncInterval from param is in minutes
+			syncSubscription = this.sync
+				.start({ fullSyncInterval: fullSyncIntervalInMilliseconds })
+				.subscribe({
+					next: ({ type, data }) => {
+						// In Node, we need to wait for queries to be synced to prevent returning empty arrays.
+						// In the Browser, we can begin returning data once subscriptions are in place.
+						const readyType = isNode
+							? ControlMessage.SYNC_ENGINE_SYNC_QUERIES_READY
+							: ControlMessage.SYNC_ENGINE_STORAGE_SUBSCRIBED;
```
ericclemmons, line 579: On the server, `DataStore.query(Post)` will return an empty array `[]` when the cache isn't primed.

packages/aws-amplify/src/withSSRContext.ts
```diff
@@ -0,0 +1,49 @@
...
+	// Auth cannot function without Credentials
+	Credentials,
+];
+
+// These modules have been tested with SSR
+const defaultModules = [API, Auth, DataStore];
+
+type Context = Pick<NextPageContext, 'req'> & {
+	modules?: any[];
+};
```
ericclemmons, line 24: `modules` is optional and context optionally accepts `req`, which is either (1) a Next.js `Request` type, or an `express`-compatible type.

(Note: for this PR , we're scoped to Next.js usage)

packages/auth/src/Auth.ts
```diff
@@ -150,7 +152,9 @@ export class AuthClass {
 			// backward compatability
 			if (cookieStorage) this._storage = new CookieStorage(cookieStorage);
 			else {
-				this._storage = new StorageHelper().getStorage();
+				this._storage = config.ssr
+					? new UniversalStorage()
+					: new StorageHelper().getStorage();
```
ericclemmons, line 157: @amhinson @Ashish-Nanda We'll have to test this to see if this is an issue for `react-native`.

It _shouldn't_ be, but it may mean instead passing `new StorageHelper({ ssr }).getStorage()` to avoid the split.

PR #6180, opened 2020-06-26
fix(@aws-amplify/auth): User Attributes coerced to boolean

elorzafe:
_Issue #, if available:_
Fixes: https://github.com/aws-amplify/docs/issues/1877
_Description of changes:_
Only coerce `email_verified` and `phone_number_verified` to boolean 

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -1987,7 +1987,8 @@ export class AuthClass {
 					attribute.Name === 'email_verified' ||
 					attribute.Name === 'phone_number_verified'
 				) {
-					obj[attribute.Name] = attribute.Value === 'true';
+					obj[attribute.Name] =
+						attribute.Value === 'true' || attribute.Value === true;
 				} else {
 					obj[attribute.Name] = attribute.Value;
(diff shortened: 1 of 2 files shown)


packages/auth/src/Auth.ts
```diff
@@ -1983,10 +1983,11 @@ export class AuthClass {
...
 			attributes.map(attribute => {
-				if (attribute.Value === 'true') {
-					obj[attribute.Name] = true;
-				} else if (attribute.Value === 'false') {
-					obj[attribute.Name] = false;
+				if (
+					attribute.Name === 'email_verified' ||
+					attribute.Name === 'phone_number_verified'
+				) {
+					obj[attribute.Name] = attribute.Value === 'true';
```
sammartinez, line 1990: Would we need the false use case here still?

elorzafe: In the case of: `attribute.Value = 'false'`

`attribute.Value === 'false'` =>  `attribute.Value = false` // old statement
`attribute.Value === 'true'` => `attribute.Value = false` // new statement

Issue #5858, opened 2020-05-21
InvalidSignatureException received when using multiple API endpoints with mixed service types


brandold:
**Describe the bug**

I have 3 API endpoints declared in my awsconfig. 2 of them are API Gateway REST APIs with the service parameter defined as 'execute-api'. The remaining endpoint is an Elasticsearch REST API with the service parameter defined as 'es'.

The behavior I'm seeing is that there seems to be a top level service value in the API client object that is overriding the service value defined in a specific endpoint. From what I've observed, that top level value is set by whatever endpoint first invokes an API call. After that, the top level service key (regardless of what endpoint is called thereafter) persists until the DOM is reloaded. 

This behavior results in the sigv4 signature to get incorrectly created, e.g. instead of signing a request to API Gateway for service 'execute-api', the signature is created for 'es', leading to an Invalid Signature Exception and failed request.

**To Reproduce**

Create multiple API endpoints in your awsconfig with different service values, then try using the different endpoints.

**Expected behavior**

I expect the API client to pull in the service parameter value from the endpoint definition itself (why even have it otherwise?) rather than a top level key that is subject to manipulation from other endpoints. 

**Code Snippet**
Please provide a code snippet or a link to sample code of the issue you are experiencing to help us reproduce the issue. (Be sure to remove any sensitive data)

**Screenshots**

Valid signature for API gateway that I expect:

![Screen Shot 2020-05-21 at 2 38 02 PM](https://user-images.githubusercontent.com/46355297/82611352-299f3500-9b75-11ea-871d-c750dc7d5864.png)

Same endpoint, but incorrect signature due to the issue described:

![Screen Shot 2020-05-21 at 2 37 46 PM](https://user-images.githubusercontent.com/46355297/82611421-5ce1c400-9b75-11ea-9449-b455dcd0e495.png)

Here you can see the API client object and top level key I described:

![Screen Shot 2020-05-21 at 2 37 11 PM](https://user-images.githubusercontent.com/46355297/82611496-7edb4680-9b75-11ea-805c-a6f057c6bf15.png)

It's clear that the top level key is 'es' in this case and the endpoints are 'execute-api', yet the client uses the 'es' value and not whats defined in its endpoint configuration. 

Invalid Signature Exception:

![Screen Shot 2020-05-21 at 2 37 41 PM](https://user-images.githubusercontent.com/46355297/82611590-b813b680-9b75-11ea-8646-fa5042de5abf.png)

**What is Configured?**
[...]

PR #6268, opened 2020-07-09
fix(@aws-amplify/api) Side effect endpoint method

elorzafe:
_Issue #, if available:_
Fixes: #5858
_Description of changes:_
Ignore original `endpoint` method (keeped for backwards compatibility) that has a side effect. Add `getEndpointInfo` method which searches the api configuration and returns the new type `ApiInfo`.

Previously was failing with InvalidSignatureException on an App like this.

```jsx
// App.tsx
import React, { useEffect, useState } from "react";
import logo from "./logo.svg";
import "./App.css";
import Amplify, { API } from "aws-amplify";

Amplify.configure(awsconfig);

function App() {
  const [lambdaResult, setLambdaResult] = useState({});
  const [apiGetResult, setApiGetResult] = useState({});
  const [apiPostResult, setApiPostResult] = useState({});
  
  useEffect(() => {
    API.post("api1", "/items", { body: {} }).then((data) => setApiPostResult(data)).catch(err => console.warn({err}));
    API.post("lambda", "", { body: {} }).then((data) => setLambdaResult(data)).catch(err => console.warn({err}));
    API.get("api1", "/items", {}).then((data) => setApiGetResult(data)).catch(err => console.warn({err}));
    API.post("lambda", "", { body: {} }).then((data) => setLambdaResult(data)).catch(err => console.warn({err}));
  }, []);
  return (
    <div className="App">
        <img src={logo} className="App-logo" alt="logo" />
        <pre data-test="api-get">{JSON.stringify(apiGetResult, null, 2)}</pre>
        <pre data-test="api-post">{JSON.stringify(apiPostResult, null, 2)}</pre>
        <pre data-test="lambda">{JSON.stringify(lambdaResult, null, 2)}</pre>
      </header>
    </div>
  );
}

export default App;
```

With a config like this
```javascript
/* eslint-disable */
// WARNING: DO NOT EDIT. This file is automatically generated by AWS Amplify. It will be overwritten.

const awsmobile = {
    "aws_project_region": "us-west-2",
    "aws_cognito_identity_pool_id": "us-west-2:xxxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
    "aws_cognito_region": "us-west-2",
    "aws_user_pools_id": "us-west-2_xxxxxxxx",
    "aws_user_pools_web_client_id": "xxxxxxxxxxx",
    "oauth": {},
    "aws_cloud_logic_custom": [
        {
            "name": "api1",
            "endpoint": "https://xxxxxx.execute-api.us-west-2.amazonaws.com/dev",
            "region": "us-west-2"
        },
        {
            name: "lambda",
            endpoint: "https://lambda.us-west-2.amazonaws.com/2015-03-31/functions/xxxxxxxxxx-dev/invocations",
            service: "lambda",
            region: "us-west-2"
          },
    ],
};


export default awsmobile;

```

With this cypress spec
```javascript
describe('API tests', () => {
    beforeEach(() => {
        cy.visit('/');
    });

    it('calling api and lambda', () => {
        cy.get('[data-test="api-get"]').contains('get call succeed!');
        cy.get('[data-test="api-post"]').contains('post call succeed!');
        cy.get('[data-test="lambda"]').contains('Hello');
    });
});
```


packages/api-rest/src/RestAPI.ts
```diff
@@ -242,6 +281,50 @@ export class RestAPIClass {
...
+		};
+
+		if (typeof apiConfig.region === 'string') {
+			response.region = apiConfig.region;
+		} else if (typeof this._options.region === 'string') {
+			response.region = this._options.region;
+		}
+
+		if (typeof apiConfig.service === 'string') {
+			response.service = apiConfig.service || 'execute-api';
```
ericclemmons, line 315: Is `apiConfig.service` often `""`?  I'm wondering iin what situation would `execute-api` be a default here.

packages/api-rest/src/RestClient.ts
```diff
@@ -81,15 +81,29 @@ export class RestClient {
...
+		let service: string = 'execute-api';
+		let custom_header: () => {
+			[key: string]: string;
+		} = undefined;
+
+		if (typeof urlOrApiInfo === 'string') {
+			parsed_url = this._parseUrl(urlOrApiInfo);
+			url = urlOrApiInfo;
+		} else {
+			({ endpoint: url, custom_header, region, service } = urlOrApiInfo);
```
ericclemmons, line 104: This is the first time I've seen destructuring over `let` variables!

packages/api-rest/src/RestAPI.ts
```diff
@@ -204,16 +237,22 @@ export class RestAPIClass {
...
-			return Promise.reject('API ' + apiName + ' does not exist');
+		try {
+			const apiConfig = this.getEndpointInfo(apiName, path);
+
+			const cancellableToken = this._api.getCancellableToken();
+
+			const initParams = Object.assign({}, init);
+			initParams.cancellableToken = cancellableToken;
+
+			const responsePromise = this._api.head(apiConfig, initParams);
```
ericclemmons, line 248: The other methods call this `apiInfo`.  Not a bug, but may be inconsistent?

```suggestion
			const apiInfo = this.getEndpointInfo(apiName, path);

			const cancellableToken = this._api.getCancellableToken();

			const initParams = Object.assign({}, init);
			initParams.cancellableToken = cancellableToken;

			const responsePromise = this._api.head(apiInfo, initParams);
```

elorzafe:
> This looks good @elorzafe! Nothing to block over, but I did want to confirm my understanding:
> 
> After reading @manueliglesias's comment ([#5858 (comment)](https://github.com/aws-amplify/amplify-js/issues/5858#issuecomment-632383146)), I understand the "why" better: _the methods customers interact with (e.g. `API.post('myApi', 'path', ...)`) "lose" their `region`/`service` values in the underlying `_api.*` calls.
> 
> This PR makes those calls accept the previous signature (a string) as well as the new `ApiInfo` struct so `region` and `service` make their way through the entire callstack.
> 
> Did I get that right? 😅

Exactly! internal classes are exported also so I tried to not brake those. But the category will be using the `getEndpointInfo` that return `ApiInfo`.

Thanks @ericclemmons !

packages/api-rest/src/types/index.ts
```diff
@@ -64,3 +64,10 @@ export interface apiOptions {
 	endpoints: object;
 	credentials?: object;
 }
+
+export type ApiInfo = {
+	endpoint: string;
+	region?: string;
+	service?: string;
+	custom_header?: () => { [key: string]: string };
```
manueliglesias, line 72: Shouldn't this be plural?

```suggestion
	custom_headers?: () => { [key: string]: string };
```

manueliglesias: If this name bubbles up all the way to the public API, we might have to live with it in its singular form as to not break consumers.

packages/api-rest/src/RestAPI.ts
elorzafe: Good question, 'execute-api' is the one that is used by api gateway. Is the default and most common option.

packages/api-rest/src/RestClient.ts
elorzafe: :)

packages/api-rest/src/RestAPI.ts
elorzafe: Thanks, I will rename that

packages/api-rest/src/RestClient.ts
```diff
@@ -81,15 +81,29 @@ export class RestClient {
...
 	 * @param {string} method - Request HTTP method
 	 * @param {json} [init] - Request extra params
 	 * @return {Promise} - A promise that resolves to an object with response status and JSON data, if successful.
 	 */
-	async ajax(url: string, method: string, init) {
-		logger.debug(method + ' ' + url);
-
-		const parsed_url = this._parseUrl(url);
+	async ajax(urlOrApiInfo: string | ApiInfo, method: string, init) {
+		logger.debug(method + ' ' + urlOrApiInfo);
```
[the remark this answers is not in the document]
elorzafe: Thanks @manueliglesias I will fix that

packages/api-rest/src/types/index.ts
elorzafe: is all the way up :(

packages/api-rest/src/RestAPI.ts
ericclemmons: Put another way, what values of `apiConfig.service` will both be a `string` and default to `execute-api`?

`""` is the only one I'm aware of, but if that won't be the case, then `|| 'execute-api'` wouldn't be necessary here.

packages/api-rest/src/RestAPI.ts
```diff
@@ -62,23 +63,25 @@ export class RestAPIClass {
 			});
 		}
 
-		if (!Array.isArray(opt.endpoints)) {
+		if (Array.isArray(opt.endpoints)) {
```
sammartinez, line 66: Do we need the `!` here?

sammartinez: As per our offline this is fine

Issue #6139, opened 2020-06-22
Authenticator authState is not updated on successful Auth.SignIn request


mchilds04:
**Describe the bug**
I'm using the aws-amplify-react-native Authenticator's authState prop to manage showing and hiding my signin/signup screen s and my main screen. On a successful sign in request the authState is not updated and as a result the app just sits on the sign in screen until I reload the app. Once the app is reloaded this triggers the authState to transition from "loading" to "signedIn" as expected.

**To Reproduce**
1. Download the sample project from [here](https://github.com/mchilds04/sample-expo-app) and run it.
2. On the sign in screen (the username and password should be prepopulated) click the sign in button. 
3. Observe the black banner at the bottom of the app that prints out the current authState

**Expected behavior**
On a successful sign in request I expect the authState to be updated to "signedIn" and  the signin screen to be hidden and the main app screen to be displayed. 

**Code Snippet**
```
const AppContainer = ({ authData, authState }) => {

  const AuthNavigationWithProps = () =>
    <AuthNavigation authState={ authState } />

  return (
    <React.Fragment>
      <Stack.Navigator>
        {
          authState === "signedIn" ?
            <Stack.Screen name="Root" component={ BottomTabNavigator } /> :
            <Stack.Screen
              name="Auth"
              component={ AuthNavigationWithProps }
              options={ { headerShown: false } }
            />
        }
      </Stack.Navigator>
      <View style={ { backgroundColor: "#000", padding: 10 } }>
        <Text style={ { color: "white", fontSize: 16, textAlign: "center" } }>
          AUTH STATE: { authState }
        </Text>
      </View>
    </React.Fragment>
  );
};

export default function App(props) {
  return (
    <View style={ styles.container }>
      <NavigationContainer linking={ LinkingConfiguration }>
        <Authenticator hideDefault={ true } container={ false }>
          <AppContainer />
        </Authenticator>
      </NavigationContainer>
    </View>
  );
}
```

**Video**
View video [here](https://www.dropbox.com/s/uhp302uju4t51zi/aws-amplify-bug.mp4?dl=0)

**What is Configured?**
   
<details>
  <summary><strong>Environment</strong></summary>
  <strong>System: </strong></br>
    OS: Linux 5.4 Ubuntu 20.04 LTS (Focal Fossa) </br>
    CPU: (8) x64 Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz </br>
[...]

PR #6276, opened 2020-07-09
fix(aws-amplify-react-native): Update authState after a successful Auth.signIn call

cedricgrothues:
_Issue:_ #6139

_Description of changes:_ This PR adds a `signIn` case to aws-amplify-react-native Authenticator's `onHubCapsule` method to fix an issue where the authState is not updated after a successful `Auth.signIn` call.

The `signIn` event is omitted if the login attempt is successful and was not previously handled by the `onHubCapsule` method. 


cedricgrothues:
You're right, I missed that. It should probably also be added to the legacy `aws-amplify-react` Authenticator.
https://github.com/aws-amplify/amplify-js/blob/main/packages/aws-amplify-react/src/Auth/Authenticator.tsx#L137-L160

amhinson:
> You're right, I missed that. It should probably also be added to the legacy `aws-amplify-react` Authenticator.
> https://github.com/aws-amplify/amplify-js/blob/main/packages/aws-amplify-react/src/Auth/Authenticator.tsx#L137-L160

Yea good catch! I'm good after those get added 👍

Issue #6059, opened 2020-06-11
pasring error in oauth callback handling


DeltaByte:
**Describe the bug**
Error handling auth response. TypeError: Cannot read property 'substr' of null

**To Reproduce**
1. set up Cognito User Pool with facebook as identity provider
2. call `Auth.federatedSignIn({ provider: 'Facebook' })`
3. go through oauth flow until you get redirected back to your application

**Expected behavior**
SDK makes call out to Cognito to complete the auth flow, using the `code` query parameter from the callback

**Screenshots**
![](https://i.imgur.com/Zuh56r1.png)

**What is Configured?**
```js
Auth.configure({
  // REQUIRED - Amazon Cognito Region
  region: process.env.AMPLIFY_AUTH_REGION,

  // OPTIONAL - Amazon Cognito User Pool ID
  userPoolId: process.env.AMPLIFY_AUTH_POOL_ID,

  // OPTIONAL - Amazon Cognito Web Client ID (26-char alphanumeric string)
  userPoolWebClientId: process.env.AMPLIFY_AUTH_CLIENT_ID,

  oauth: {
    domain: process.env.AMPLIFY_OAUTH_DOMAIN,
    // scope: ['phone', 'email', 'profile', 'openid', 'aws.cognito.signin.user.admin'],
    redirectSignIn: 'http://localhost:3000/oauth2',
    redirectSignOut: 'http://localhost:3000/oauth2'
    // responseType: 'code' // or 'token', note that REFRESH token will only be generated when the responseType is code
  }
})
```
 

<details>
  <summary><strong>Environment</strong></summary>

```
  System:
    OS: Linux 5.6 Arch Linux
    CPU: (8) x64 Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz
    Memory: 1.24 GB / 3.84 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 12.18.0 - ~/.nvm/versions/node/v12.18.0/bin/node
    npm: 6.14.4 - ~/.nvm/versions/node/v12.18.0/bin/npm
  npmGlobalPackages:
    @cloudflare/wrangler: 1.9.2
    npm: 6.14.4
```

</details>

**Additional context**

```
[ERROR] 42:16.970 OAuth - Error handling auth response. TypeError: Cannot read property 'substr' of null
    at OAuth.eval (OAuth.js?1700:190)
    at step (OAuth.js?1700:55)
[...]

PR #6279, opened 2020-07-09
fix(@aws-amplify/auth): Default hash to `#` when null

ericclemmons:
_Issue #, if available:_ Fixes https://github.com/aws-amplify/amplify-js/issues/6059

_Description of changes:_

When a URL doesn't contain `#...`, `parse(url)` returns `{ hash: null }`, which creates a run-time error for `hash.substr(1)`.

---


davewthompson:
Can we get this fix in, please? It's breaking my app at the moment.

PR #6290, opened 2020-07-10
feat(vscode-extension): Remove unnecessary files

cedricgrothues:
_Issue #, if available:_ N/A

_Description of changes:_ This PR removes unnecessary files from the Amplify vscode extension.
The `src/` directory is not necessary to publish the extension and since all actual TypeScript files have been commented out, they and all related files (`tsconfig.json`, `tslint.json`) can be removed.


ashika01:
Do we need the package-lock here? could we ignore that.?

Issue #6274, opened 2020-07-09
We are facing problems with Amplify related to social network login


abudakhmad:
Here is the description of the problem:
I opened totaly new browser and swithced on incognito mode, than I entered our platform via facebook login. after that I did log-out from the system and opened facebook page and did log-out from the facebook itself, after that I opened our platform one more time and pressed sign in via facebook and it landed me directly into the system without asking any credentials. I was trying to clear all local storage and cookies and it also didn't help. I'm attaching prototype project with Amplify and it also has the same behaviour like I described above. Can you please assist us with this issue

elorzafe:
@abudakhmad 

```javascript
<button onClick={() => Auth.signOut({ global: true }).then(() => window.location.href = "/")}>Logout</button>
```

For some reason the promise ` Auth.signOut({ global: true })` is resolved before the logout request is sent from the browser. I am looking at right now, `window.location.href="/"` is cancelling the request.

elorzafe:
@abudakhmad 

Why do you need to do `window.location.href = "/"`, even if OAuth flow has sign out url?

abudakhmad:
> @abudakhmad
> 
> Why do you need to do `window.location.href = "/"`, even if OAuth flow has sign out url?

We need it in order to redirect user to the main page. So if I understand right, we need to wait until there will be fix from the Amplify side?

PR #6294, opened 2020-07-10
fix(@aws-amplify/auth): signout promise on OAuth flow

elorzafe:
_Issue #, if available:_
Fixes: #6274
_Description of changes:_
Resolving the promise during signout for OAuth flow, create a race condition where logout request is cancelled. Now promise is only rejected after a timeout.


packages/auth/__tests__/hosted-ui.test.ts
```diff
@@ -245,4 +245,113 @@ describe('Hosted UI tests', () => {
...
+		expect.assertions(1);
+
+		try {
+			await auth.signOut({ global: false });
+
+			expect(true).toBe(false);
+		} catch (err) {
+			expect(err).toEqual('Signout timeout fail');
+		}
+
```
ericclemmons, line 353: Which `expect` is being tested?

The test name doesn't make it clear:
> SignOut for signed with hosted ui

What do you think about naming the test to indicate expected behavior?

packages/auth/src/Auth.ts
```diff
@@ -1480,7 +1480,8 @@ export class AuthClass {
 						onSuccess: data => {
 							logger.debug('global sign out success');
 							if (isSignedInHostedUI) {
-								return res(this._oAuthHandler.signOut());
+								this._oAuthHandler.signOut();
+								setTimeout(() => rej('Signout timeout fail'), 3000);
```
ericclemmons, line 1484: Should this have an `await` or a `res`? It looks like this will always reject, since there's not an explicit `res` called in this callstack.

```js
await new Promise((res, rej) => {
  setTimeout(() => rej('Failed'), 1000)
  res('Succeeded')
})
  
// "Succeeded"

await new Promise((res, rej) => {
  setTimeout(() => rej('Failed'), 1000)
})
  
// uncaught exception: Failed

File: packages/auth/__tests__/hosted-ui.test.ts (hunk 1 above)
Comment (elorzafe) [reply packages/auth/__tests__/hosted-ui.test.ts:353]: You are correct, the name is not clear at all.

When you signOut using OAuth flows, the app should be redirected to the signout url which means if logout is succesfull it should never reach that code.

Is testing the rejection case which is called only when the app is not redirected.

File: packages/auth/src/Auth.ts (hunk 2 above)
Comment (elorzafe) [reply packages/auth/src/Auth.ts:1484]: Yes, that is the intended behaviour. I should add a comment or creating a function that explains that signout OAuth flows always redirect the app.

File: packages/auth/__tests__/hosted-ui.test.ts [hunk 3]
```diff
@@ -245,4 +246,220 @@ describe('Hosted UI tests', () => {
...
+					);
+
+					done();
+				});
+			}
+		);
+		const options = {
+			...authOptionsWithOAuth,
+		};
+		options.oauth.urlOpener = urlOpener;
```
ericclemmons, line 437: This is smart!

PR #6311, opened 2020-07-13
Suggestion: [WIP] Begin Migration to TS Project References, Stricten TS Configs

harrysolovay:
As of right now, the TS compilation setup is non-standard, and emits builds regardless of TS errors. Useful compiler flags are disabled. Let's phase this out, package by package, starting with the Auth category.

Thanks to TS project references, we have a clear "best practices" solution to managing our packages' incremental compilation and their interdependencies. This solution will enable us to...

- more easily reason about package-specifics (is JSX enabled? Do we need to enable down-level iteration? Etc.)
- reduce friction for new contributors
- cut down on boilerplate from every package's manifest
- eliminate the need for custom build scripts 
- **gradually stricten our use of TypeScript, per-package**

This is a first pass. Once we get this in good shape for Auth (I've temporarily broken the dual TS+Webpack watch for Auth, I've also ts-ignored some 53 type errors), we can begin to do the same for other categories.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,6 @@
 	"name": "aws-amplify-monorepo",
 	"private": true,
 	"version": "0.1.30",
-	"description": "",
 	"scripts": {
 		"setup-dev": "yarn && yarn bootstrap && yarn link-all && yarn build",
 		"bootstrap": "lerna bootstrap",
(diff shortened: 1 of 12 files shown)


harrysolovay:
Also, I've updated the version of TypeScript. The lowest necessary version is `3.7` (for references). Is there a reason that we're currently listed at `3.6.4`?

harrysolovay:
@sammartinez @ericclemmons has this been the case for you two? The TS language server is crashing (hence no intellisense, hover-to-view-type, etc.).

Screen Shot 2020-07-13 at 3 33 48 PM

sammartinez:
> Also, I've updated the version of TypeScript. The lowest necessary version is `3.7` (for references). Is there a reason that we're currently listed at `3.6.4`?

There was a breaking change that was occurring where we were having issues with upgrade to that version `3.7.x` hence the reason for the lock.

PR #6334, opened 2020-07-15
Update any and 'SUCCESS' response types of resendConfirmationCode and forgotPassword

ricosandyca:
- change any type at onSuccess argument as CodeDeliveryDetails instead
- change 'SUCCESS' type at resolve argument of resendConfirmationCode's callback as CodeDeliveryDetails instead

Description of changes:
Changed some types and make it more strict


amhinson:
Looks like there are some conflicts here @ricosandyca. Would you be able to address those?

Issue #6115, opened 2020-06-18
Inconsistent Translation Behaviour for Amplify UI Components


liamaws:
**Describe the bug**
When providing translations for amplify UI components only some fields update with the provided translation.

**To Reproduce**

1. Clone https://github.com/xavierraffin/fizz-demo
2. Checkout branch noformfields
3. npm install
4. npm start
5. amplify init 
6. toggle the lang using the top button

**Expected behavior**
The language of all fields that have a provided translation should update.

**Code Snippet**
Using https://github.com/aws-amplify/amplify-js/blob/master/packages/amplify-ui-components/src/common/Translations.ts as reference for labels

String file https://github.com/xavierraffin/fizz-demo/blob/noformfields/src/strings.js

```
export const strings = {
    en: {
        signInHeader: "Sign in your Fizz account",
        signUpHeader: "Create your fizz account",
        forgotPasswordHeader: "Custom Header Reset Password",
    },
    fr: {
        signInHeader: "Sign in your Fizz account FR",
        signUpHeader: "Create your fizz account FR",
        forgotPasswordHeader: "Custom Header Reset Password FR",
        [Translations.NO_ACCOUNT_TEXT]: 'No account? FR',
        [Translations.CREATE_ACCOUNT_TEXT]: 'Create account FR',
        [Translations.SIGN_IN_ACTION]: 'Sign In FR',
        [Translations.SIGN_IN_HEADER_TEXT]: 'Sign in to your account FR',
        [Translations.SIGN_IN_TEXT]: 'Sign in FR',
        [Translations.SIGN_UP_EMAIL_PLACEHOLDER]: 'Email FR',
        [Translations.SIGN_UP_HAVE_ACCOUNT_TEXT]: 'Have an account? FR',
        [Translations.SIGN_UP_HEADER_TEXT]: 'Create a new account FR',
        [Translations.SIGN_UP_PASSWORD_PLACEHOLDER]: 'Password FR',
        [Translations.SIGN_UP_SUBMIT_BUTTON_TEXT]: 'Create Account FR',
        [Translations.SIGN_UP_USERNAME_PLACEHOLDER]: 'Username FR',
        [Translations.SEND_CODE]: 'Send Code FR',
        [Translations.BACK_TO_SIGN_IN]: 'Back to Sign In FR',
        [Translations.EMAIL_LABEL]: 'Email Address * FR',
        [Translations.EMAIL_PLACEHOLDER]: 'Enter your email address FR',
[...]

PR #6346, opened 2020-07-16
(@aws-amplify/ui-components): Move I18n.get to render instead of on mount

jordanranz:
_Issue #, if available:_
https://github.com/aws-amplify/amplify-js/issues/6115

_Description of changes:_
- Move `I18n.get` to render method for translation updates on render instead of mount


wei:
@jordanranz @ericclemmons 

We are using `<AmplifySignIn />` without props. The language strings are displaying correctly on initial render but calling `I18n.setLanguage('fr')` does not rerender the component. Doing something like `<AmplifySignIn key={lang} />` works because it forces recreation of the component.

**Quick question:** How do we trigger a rerender on `I18n.setLanguage()` without recreating component using key? 

Thanks a bunch!

ericclemmons:
You're right, `I18n` doesn't trigger anything for components to know to re-render.  Can you open a new discussion with this request & the scenario for updating language post-render?

If `I18n.setLanguage` fired a `Hub` event, we _could_ re-render the component 🤔 
https://github.com/aws-amplify/amplify-js/blob/377acd0507e6d4be34bf36ea9c0694fa98ac04b1/packages/core/src/I18n/I18n.ts#L64-L66

wei:
Thanks @ericclemmons , issue opened #6986

PR #6355, opened 2020-07-17
(@aws-amplify/ui-components): User agent appending for UI components

jordanranz:
_Description of changes:_
- appendToAmplifyUserAgent in core
- User agent content append S3 UI component load

### DIFF
--- a/packages/core/src/Platform/index.ts
+++ b/packages/core/src/Platform/index.ts
@@ -39,6 +39,18 @@ export const getAmplifyUserAgent = () => {
 	return Platform.userAgent;
 };
 
+export const appendToAmplifyUserAgent = content => {
+	if (!content) {
+		return;
+	}
+	if (Platform.userAgent && !Platform.userAgent.includes(content)) {
+		Platform.userAgent = Platform.userAgent.concat(' ', content);
+	}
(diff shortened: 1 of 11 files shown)


packages/core/src/Platform/index.ts
```diff
@@ -39,6 +39,18 @@ export const getAmplifyUserAgent = () => {
 	return Platform.userAgent;
 };
 
+export const appendToAmplifyUserAgent = content => {
+	if (!content) {
+		return;
+	}
+	if (Platform.userAgent && !Platform.userAgent.includes(content)) {
+		Platform.userAgent = Platform.userAgent.concat(' ', content);
+	}
```
Amplifiyer, line 48: Does this change the userAgent for that particular request or any subsequent calls will use this user agent?

jordanranz: It does at the moment. Open to suggestions. This gets back into the modification of the API to support 1 time user agent appended strings we had before.

Amplifiyer: it can get into weird state for concurrent calls and not reporting the right metrics. Generally we should avoid changing a static/global value in scope of a request. We may need some refactor in our library, fortunately aws-sdk does support per-request user agent modifications.

Issue #6358, opened 2020-07-18
DataStore breaks when record is constantly queried after updating


nubpro:
**Describe the bug**
This is the extreme way of reproducing the problem, however we are hitting this problem quite often and we have our own workaround for this issue.

When a user spams the update button on the same record, like I am talking about tapping the update button 10 times in a row. DataStore would throw this error: `TypeError: undefined is not an object (evaluating 'Object.getPrototypeOf(model)')`.

You can refer to the screenshot below:
![image](https://user-images.githubusercontent.com/762914/87845845-cbd16400-c8fd-11ea-95b4-743c50506119.png)
I did some additionally work and find out the error is thrown from the delete function within the asyncstorage.js file due to the `model` being undefined.

The consequence of this error is fairly large, it causes any outgoing mutations not be sent out to the backend.
This error is especially hard to catch on production mode because the error is not visible on the frontend.

**To Reproduce**
1. Using the [example React Native app](https://docs.amplify.aws/lib/datastore/examples/q/platform/js#example-react-application), add a new function to allow for Post update and an update button to the UI.

Update function
```
onUpdate = (postId) => {
    const original = await DataStore.query(Post, "123");
    
    await DataStore.save(
    Post.copyOf(original, updated => {
        updated.title = `title ${Date.now()}`;
    })
    );
}
```

Update button
```
<Text style={styles.text} onPress={() => this.onUpdate('<ID here>')}>
    Update this post
</Text>
```

2. Make sure to keep the `DataStore.observe(Post)` subscription, otherwise you won't be able to reproduce the error.
3. Tap the update button 10 times in a row as quickly as possible.
4. You are very likely to hit this error at this point.

**Expected behavior**
DataStore should somehow recover or prevent this error from happening.

**What is Configured?**
<details>
  <summary><strong>Environment</strong></summary>

```

  System:
    OS: macOS 10.15.5
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Memory: 77.89 MB / 8.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.3.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 83.0.4103.116
    Safari: 13.1.1
[...]

PR #6361, opened 2020-07-18
fix(@aws-amplify/datastore): Fix dequeue bug caused by undefined head

nubpro:
_Issue #, if available:_
#6358

_Description of changes:_
Attempting to delete an `undefined` head when dequeueing in outbox would cause DataStore to break and throw error (`TypeError: undefined is not an object (evaluating 'Object.getPrototypeOf(model)')`).


packages/datastore/src/sync/outbox.ts
```diff
@@ -80,7 +80,9 @@ class MutationEventOutbox {
 	public async dequeue(storage: StorageFacade): Promise<MutationEvent> {
 		const head = await this.peek(storage);
 
-		await storage.delete(head);
+		if (head !== undefined) {
+			await storage.delete(head);
+		}
```
ericclemmons, line 85: @manueliglesias Your suggestion here sounds correct to me:
https://github.com/aws-amplify/amplify-js/issues/6358#issuecomment-662684129

Namely, `head` becoming `undefined` sounds like a symptom of an underlying queue problem.  (We can discuss this in Discord/Slack if you'd like)

Granted, this PR short-circuits the bug, so I'm happy to accept this if the "correct" solution differs & is a bigger lift.

ericclemmons:
Thanks for your contribution!  

Based on https://github.com/aws-amplify/amplify-js/issues/6358#issuecomment-662684129's workaround, this PR address the symptom, but the root cause still needs to be addressed within the team.

We have ongoing work that should address this (if it hasn't been already), since DataStore is in active development. 🙏

mrsimply:
@ericclemmons  Is there already a solution in sight? I still experience the issue in v.4.1.2. It's hard to throttle the updates everywhere, since in react or react-native my DataStore updates often are invoked through useEffect (which may fire a lot)... Thanks for your update!

Issue #6324, opened 2020-07-14
Audiocontext was not allowed to start


shresthapradip:
**Motivation**
While trying to use lex, voice conversation doesn't work with warning: audiocontext was not allowed to start.

**To Reproduce**

1. Create new Vue app.
2. Setup amplify and lex correctly. Use amplify-chatbot for chatbot ui.
3. Open app on chrome (safari works as expected)
3. Initiate voice conversation. See it keeps showing listening. 
4. If you go to browser console, you will see the warning Audiocontext was not allowed to start.

**Expected behavior**
No audiocontext error.

**Code Snippet**
````ts
        exports.audioRecorder = function () {
        /**
         * Creates an audio context and calls getUserMedia to request the mic (audio).
         */
        var requestDevice = function requestDevice() {
          if (typeof audio_context === 'undefined') {
            window.AudioContext = window.AudioContext || window.webkitAudioContext;
            audio_context = new AudioContext(); // Error here
          }

          return navigator.mediaDevices.getUserMedia({
            audio: true
          }).then(function (stream) {
            audio_stream = stream;
          });
        };
````

**Additional context**
Only happening on chrome browser.

wlee221:
Hello @shresthapradip, thanks for letting us know! This is currently [being worked on](https://github.com/wlee221/amplify-js/pull/1) along with other bugs we found with the Interactions category. We'll let you know when this makes into the repo. Feel free to reach out if you have any questions :)

wlee221:
Note that you can unblock yourself for now by setting the Chrome flag "Autoplay Policy" to "No user gesture is required", as [explained here](https://developers.google.com/web/updates/2017/09/autoplay-policy-changes). Alternatively you can require a user gesture before loading the chatbot (e.g. adding a button to start the Chatbot instance). We'll keep you updated with the fix 😃

wlee221:
Hi @shresthapradip, this should be fixed on our latest release. Please feel free to open a new issue if problem still persists.

Issue #5024, opened 2020-03-03
aws-amplify-angular cannot be used with strict Content Security Policy


matulef:
**Describe the bug**
When using the `AmplifyService`  from `aws-amplify-angular`, a reasonable Content Security Policy cannot be used without causing an error. This is because the `aws-lex-audio.js` file requires `unsafe-eval` (and `worker-src blob:`) permissions. 

**To Reproduce**

1. Create a new angular app:
```
ng new csp-test
cd csp-test
```

2. Add ` AmplifyAngularModule, AmplifyService, AmplifyModules` as described in 
https://aws-amplify.github.io/docs/js/angular. I am only using Auth, so in my `app.module.ts` file I have 
```
{
      provide: AmplifyService,
      useFactory: () => {
        return AmplifyModules({
          Auth,
        })
      },
    },
```
2. In `index.html`, add a CSP:
```
 <meta http-equiv="Content-Security-Policy" content="script-src 'self'; object-src 'self'"/>
```
3. Start the app with `ng serve`
4. See the following error in the console:

![Screen Shot 2020-03-02 at 4 13 11 PM](https://user-images.githubusercontent.com/974305/75730210-12048e80-5ca1-11ea-8c8a-b8f6efe843dc.png)

**Expected behavior**
This should not require `unsafe-eval` permissions, especially since the Lex audio stuff isn't even being used. The error can be avoided with a loose content security policy, e.g. 
`worker-src blob:; script-src 'self' 'unsafe-eval'; object-src 'self'`
but `unsafe-eval` is strongly discouraged for security reasons. 

Note that this issue also applies when building the app to run as a Chrome extension, where the content security policy is defined inside of the `manifest.json` file. 

<details>
  <summary><strong>Environment</strong></summary>

```
   System:
    OS: macOS 10.15.3
    CPU: (8) x64 Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
    Memory: 132.70 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 10.14.2 - ~/.nvm/versions/node/v10.14.2/bin/node
    Yarn: 1.12.3 - ~/.nvm/versions/node/v10.14.2/bin/yarn
    npm: 6.5.0 - ~/.nvm/versions/node/v10.14.2/bin/npm
  Browsers:
    Chrome: 80.0.3987.122
    Firefox: 71.0
    Safari: 13.0.5
  npmPackages:
[...]

PR #6381, opened 2020-07-21
fix(@aws-amplify/interactions): fix interactions v3 bugs and refactor type

wlee221:
_Issue #, if available:_ #6324

_Description of changes:_ This PR fixes current bugs with the Interactions category and strongly types the `Interactions.send()` request.

#### Bug fixes:
- Convert `aws-sdk-js-v3` response from `Lex.PostContent` to `Uint8Array`.
- Fix ReferenceError in react native chatbot ui. 
- Resume `AudioContext` when user presses the voice chat button on browser ui components.

#### Unit Test:
- Fix a bug where `jest.runAllTimers()` were called prematurely. This prevented `onComplete` callbacks from running and jest passed the tests vacuously.
- Fix a bug where jest was falsely expecting `onComplete` callback from text messages for voice messages.
- Test `Uint8Array` conversion.

#### Category Refactor:
- Strongly type `Interactions.send` API request type.
- Overload `Interactions.send`.


packages/interactions/src/types/Interactions.ts
```diff
@@ -14,9 +14,20 @@ export interface InteractionsOptions {
...
+	content: string;
 	options: {
-		[key: string]: string;
+		messageType: 'text';
 	};
-}
+};
+
+export type InteractionsVoiceMessage = {
+	content: string | object;
```
elorzafe, line 25: Should this be only object?

packages/interactions/src/Providers/AWSLexProvider.ts
```diff
@@ -23,6 +27,7 @@ import {
 	Credentials,
 	getAmplifyUserAgent,
 } from '@aws-amplify/core';
+import { convert } from './AWSLexHelper/convert';
```
elorzafe, line 30: Maybe I will call it `AWSLexProviderHelper`

packages/interactions/src/Providers/AWSLexHelper/convert.ts
```diff
@@ -0,0 +1,42 @@
...
+	}
+};
+
+const readOnBrowser = (stream: ReadableStream | Blob) => {
+	return new Response(stream)
+		.arrayBuffer()
+		.then(buffer => new Uint8Array(buffer));
+};
+
+// Node.js support
```
elorzafe, line 18: I will remove that line, Amplify doesn't officially support Node.js. Method name is clear enough. In which scenarios do you think that will be used?

packages/interactions/__tests__/Interactions-unit-test.ts
```diff
@@ -556,46 +605,77 @@ describe('Interactions', () => {
...
+						m1: 'hi',
+						m2: 'done',
 					},
+					audioStream: typedStream,
 				});
+				jest.runAllTimers();
+			});
 
-				const textResponse = await interactions.send(
+			test('Interactions configuration and send voice message to existing bot and call onComplete from configure onComplete', async () => {
```
elorzafe, line 631: I would rename the test to something more concise (what is testing, rather what is doing). Like  `onComplete callback from configure being called`

packages/interactions/__tests__/Interactions-unit-test.ts
```diff
@@ -475,57 +495,88 @@ describe('Interactions', () => {
...
+						m1: 'voice:hi',
+						m2: 'voice:done',
 					},
+					audioStream: typedStream,
 				});
+				jest.runAllTimers();
 			});
 
-			test('Interactions configuration and send message to existing bot and call onComplete from configure onComplete', async () => {
+			test('Interactions configuration and send text message to existing bot and call onComplete from configure onComplete', async () => {
```
elorzafe, line 572: I would rename it as my other comment

packages/interactions/__tests__/Interactions-unit-test.ts
```diff
@@ -433,11 +452,13 @@ describe('Interactions', () => {
...
 				dialogState: 'ElicitSlot',
 				message: 'echo:hi',
+				audioStream: typedStream,
 			});
 		});
 
 		describe('Sending messages to bot', () => {
-			test('Interactions configuration and send message to existing bot and call onComplete from Interaction.onComplete', async () => {
+			jest.useFakeTimers();
+			test('Interactions configuration and send text message to existing bot and call onComplete from Interaction.onComplete', async () => {
```
elorzafe, line 461: I would rename it to something shorter (check my other comment)

packages/interactions/src/types/Interactions.ts
wlee221: I thought so, but content (to Lex.PostContent) can be `string | object` as specified [here](https://github.com/aws/aws-sdk-js-v3/blob/master/clients/client-lex-runtime-service/commands/PostContentCommand.ts#L25).

packages/interactions/src/Providers/AWSLexProvider.ts
wlee221: Sounds good!

packages/interactions/__tests__/Interactions-unit-test.ts
wlee221: Sure!

packages/interactions/src/Providers/AWSLexHelper/convert.ts
wlee221: That makes sense; I just added this for basic Node.js coverage (also helps jest) but not something to be actively used / maintained. I'll remove the line as suggested.

packages/interactions/src/types/Interactions.ts
wlee221: After looking at the [API](https://docs.aws.amazon.com/lex/latest/dg/API_runtime_PostContent.html), it seems that it is safe to say content would be only `object`. The only problem is that our unit tests send `string` content when sending voice messages. Let me see if I can fix that.

wlee221: Unit tests are running fine, so I'll change as requested :D

wlee221:
@jordanranz, @elorzafe thanks for the review!

- I will write a shorter typescript version of `aws-lex-audio` and consolidate into Interactions. That way, there will be only one `aws-lex-audio` in our libraries. There's also [content security concern](https://github.com/aws-amplify/amplify-js/issues/5024) with the file, so I'll work on that too.
- This has not been tested with Next.js, but I can test it (good time to learn Next.js also) after this merge.

Issue #1715, opened 2018-09-21
user.completeNewPasswordChallenge is not a function


goatandsheep:
**Describe the bug**

**To Reproduce**
1. Use flat aws-exports.js structure
2. Set Cognito up such that Only Admin creates accounts
3. Use Amplify but not mobile hub
3. Create an account
5. try to change the temporary password using `Auth.completeNewPassword(userFromLoginAttempt, newPassword, requiredAttributes)`
4. Responds with `user.completeNewPasswordChallenge is not a function`

**Expected behavior**
Password is changed and the user can login

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - Version: 1.0.0

***You can turn on the debug mode to provide more info for us by setting window.LOG_LEVEL = 'DEBUG'; in your app.***

powerful23:
@goatandsheep hi can you confirm that the `userFromLoginAttempt` object should be a `CognitoUser` object?

goatandsheep:
## More details

upon `Auth.signIn()`, I receive a cognito user object. Unfortunately both `Auth.currentSession()` and `Auth.currentAuthenticatedUser()` came up as error instead of returning the user, so I couldn't access the `CognitoUser` object. As a workaround, I stored the user temporarily in localStorage. However, Amplify did not recognize the object after the JSON had been parsed. In addition, my router was auth-protected and clears the store when the above two functions return errors.

## Fix

I whitelisted the password change page to not be auth protected and put the value into my store

## Suggestion

Provide a way to obtain the user object when the user is not yet confirmed other than signIn. Even if it's located in the error message. I don't even know if it's cached.

powerful23:
@goatandsheep When you get resolved from `Auth.signIn()` with that user object, you are not actually signed in because you need to complete the new password. So at that moment `Auth.currentSession` and `Auth.currentAuthenticatedUser` won't work because you are not signed in. You need to directly use that user object returned from `Auth.signIn()` to help you finish the login process.

vaddanki:
Same error..
[...]

PR #6387, opened 2020-07-22
enhanced typings

donaldpipowitch:
_Issue #, if available:_
Partially applies to https://github.com/aws-amplify/amplify-js/issues/1715. (You try to run `completeNewPasswordChallenge` without being able to do it.)

_Description of changes:_

Some quick type changes to make the usage more DX friendly.

### DIFF
--- a/packages/amazon-cognito-identity-js/index.d.ts
+++ b/packages/amazon-cognito-identity-js/index.d.ts
@@ -165,7 +165,7 @@ declare module 'amazon-cognito-identity-js' {
 				onFailure: (err: Error) => void;
 			}
 		): void;
-		public completeNewPasswordChallenge(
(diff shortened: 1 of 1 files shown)


amhinson:
Thanks for the PR @donaldpipowitch! Would you be able to elaborate a bit on the issues/errors that this is fixing?

donaldpipowitch:
CognitoUser is stateful. The complete New Password Challenge is not always set. People are confused by this (eg https://github.com/aws-amplify/amplify-js/issues/1715). The updated interface makes this clearer.

donaldpipowitch:
@amhinson Who will review this? ☺️

goatandsheep (approved):

PR #6403, opened 2020-07-23
fix(@aws-amplify/interactions): remove `Readable` dependencies

wlee221:
_Issue #, if available:_

_Description of changes:_

This removes `Readable` dependency from the AWS Lex provider and unit testing.


packages/interactions/src/Providers/AWSLexProviderHelper/convert.ts
```diff
@@ -1,41 +1,9 @@
-import { Readable } from 'stream';
-export const convert = async (
-	stream: Readable | ReadableStream | Blob
-): Promise<Uint8Array> => {
-	if (stream instanceof Readable) {
-		return readOnNode(stream);
+export const convert = (stream: object): Promise<Uint8Array> => {
```
ashika01, line 1: Could we still keep the `stream` type to be `Readable | ReadableStream | Blob` ??

ashika01:
Just nit! Looking good overall?!

packages/interactions/src/Providers/AWSLexProviderHelper/convert.ts
wlee221: That should be fine as long as I don't do `instanceof Readable`; I'll add that in.

wlee221:
Thanks @jordanranz @ashika01 @elorzafe 👍

PR #6418, opened 2020-07-27
fix parameters in sendMFASelectionAnswer

donaldpipowitch:
The wrong parameters were passed to the callbacks of `sendMFASelectionAnswer`.


ericclemmons:
@donaldpipowitch Can you share some sample code on how you're using `CognitoUser` directly or indirectly so we can validate this PR?

Thanks!

donaldpipowitch:
> @donaldpipowitch Can you share some sample code on how you're using `CognitoUser` directly or indirectly so we can validate this PR?

```
cognitoUser.sendMFASelectionAnswer('SMS_MFA', {
  mfaRequired: (challengeName, challengeParameters) => {
    // those params are always undefined without the fix
  },
});
```

If you search for those params in the file you'll see that the API returns `ChallengeParameters`/`challengeName` and not `challengeParameters`/`ChallengeName`.

image

PR #6425, opened 2020-07-27
fix (@aws-amplify/storage) - track property

ashika01:
_Issue #, if available:_
The storage category provides a way to track events in pinpoint provided the track event is set to `true`. This was not working.

Initially, the tracking wasn't working since there was a [pinpoint issue around CORS](https://github.com/aws-amplify/amplify-js/pull/6392). 

Once that was resolved the storage track `putEvents` where failing with an error
Screen Shot 2020-07-27 at 11 24 15 AM

This was happening due to `null` value was being sent in place of metrics by default. This was blocking storage UI component's `track` functionality too

_Description of changes:_

This PR address the issue and append metrics to `Analytics.record` call only if present.

After the PR events are being sent (screenshot of network tab):
image

### DIFF
--- a/packages/storage/src/providers/AWSS3Provider.ts
+++ b/packages/storage/src/providers/AWSS3Provider.ts
@@ -46,7 +46,7 @@ const dispatchStorageEvent = (
 	message: string
 ) => {
 	if (track) {
-		let data = { attrs };
+		const data = { attrs };
 		if (metrics) {
 			data['metrics'] = metrics;
 		}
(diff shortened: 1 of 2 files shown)


packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -46,11 +46,15 @@ const dispatchStorageEvent = (
 	message: string
 ) => {
 	if (track) {
+		let data = { attrs };
```
jordanranz, line 49: Could we write a unit test for this? Seems there is an `AWSS3Provider-unit-test.ts` we can add to.

ashika01: There is already a test to this, which is what failed now. Yay. But i guess it is tricky to test the actual case itself e2e but we could possibly add a test to our integration ones with tracking enabled to verify this e2e.

PR #6436, opened 2020-07-28
fix(@aws-amplify/pubsub): Fix undefined observer

nubpro:
_Issue #, if available:_
none

_Description of changes:_
This would result to a fatal exception that leads to a crash because `observer` is undefined as it is already closed and cleaned.


nubpro:
@ashika01 Thanks for the approval!
I wouldn't want bug anybody, but would this get merged?

Issue #5403, opened 2020-04-14
No notification when websocket closes


Stephan221713:
**Describe the bug**
When AppSync-WebSockets close, the observer is not notified.

**To Reproduce**
Observed on Android devices running Google Chrome. When the device is in standby mode (+ maybe power save mode), webosckets often get closed. Our observer, which uses Amplify to subscribe to AppSync, is not notified of this, so we cannot tell if we need to reconnect after returning to the device.

**Expected behavior**
When the websocket closes, the observer should receive an error.

**Details**
We are using aws-amplify 3.0.6
The problematic code is in packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts lines 674-686:
```
if (type === MESSAGE_TYPES.GQL_CONNECTION_ACK) {
	ackOk = true;
	this.keepAliveTimeout = connectionTimeoutMs;
	this.awsRealTimeSocket.onmessage = this._handleIncomingSubscriptionMessage.bind(
		this
	);
	this.awsRealTimeSocket.onerror = err => {
		logger.debug(err);
		this._errorDisconnect(CONTROL_MSG.CONNECTION_CLOSED);
	};
	res('Cool, connected to AWS AppSyncRealTime');
	return;
}
```

As you can see, once the connection is established, messages and errors from the socket are passed to the observer, but a socket close is not. In our case, the socket only fires an onclose event, but no onerror when the websocket closes, so our observer is not notified. We patch our local version by treating onclose exactly like onerror and calling this._errorDisconnect(CONTROL_MSG.CONNECTION_CLOSED), which solves the issue for us. We decided not to create a Pull Request on this, because we are not sure if our solution has problematic side effects. However, it is very important for observers to be notified about closed websockets in some way.

adamup928:
Can confirm that applying the local patch mentioned by @Stephan221713 allows our subscriptions to reconnect gracefully when returning from standby on MacOS. Without this fix, the subscriptions do not connect.

greimela:
I just opened a pull request fixing this issue, since it seems to impair multiple people.

ericclemmons:
I'm reviewing #6448 now.

@adamup928 can you describe your scenario so I can test that as well?  It sounds like you have an app with `GraphQL.subscribe(...)`, close the lid, then get socket errors in the devtools' Network tab?

adamup928:
[...]

PR #6448, opened 2020-07-29
fix(@aws-amplify/pubsub): Handle onclose events in AwsRealTimeSocket

greimela:
_Issue #, if available:_
#5403 

_Description of changes:_
This change allows a user of the socket to restart the subscription after it terminates without an error.
This can sometimes happen when devices enter or leave standby.


bensie:
We're seeing the same behavior as described in #5403 and would love to see this merged!

ericclemmons:
@greimela @bensie I was able to validate this (including @adamup928's example from #5403).

Since a socket reconnect doesn't happen automatically, how are/would you handling reconnecting?

greimela:
@ericclemmons Nice, thanks for the confirmation!

We are using the error callback to restart our subscriptions. So if the close events arrive there it's perfect for us 👌

bensie:
@greimela Would you mind sharing how you trigger the websocket reconnect from within the subscribe `error` function? It doesn't seem like triggering resubscribes causes the websocket connection to reconnect if it's no longer active. Thanks!

greimela:
@bensie We are actually opening a completely new subscription in the case of an error.

bensie:
@greimela Guessing I'm missing a step here and I think we might be mixing up terminology - with 1 websocket connection and many subscriptions on that connection, if the websocket connection is closed and the `onclose` event is fired, how do you reconnect the websocket ahead of re-initiating those subscriptions? Do you have a code snippet you could share?

greimela:
@bensie We are using this lib only through the GraphQL subscriptions, so we might speak of different `subscriptions` here.

We just call `API.graphql(...).subscribe(...)` again after we get the error.

bensie:
@greimela Interesting, thanks! I'll keep digging, when we do that the websocket connection doesn't seem to reconnect automatically and the resubscribe fails.

Issue #5782, opened 2020-05-14
(React) UI Components don't support password managers


Purii:
**Describe the bug**
Browsers don't recognize the input elements as Auth related.

-> No "password suggestion" on during account creation
-> No autofill during sign in
-> No prompt "save password?" during sign in

I'm not sure what details are necessary for this to work, but I tried setting the `name` attribute and others through `formFields`. But that didn't help solve this issue. (https://www.chromium.org/developers/design-documents/form-styles-that-chromium-understands)

-> Form is not part of a "form" element
-> The "sign in/up" buttons aren't labelled as type "submit"

Purii:
Seems that the usage of `shadow DOM` is the root cause: https://bugs.chromium.org/p/chromium/issues/detail?id=770175

anantoghosh:
[Vaadin login component](https://vaadin.com/components/vaadin-login/java-examples)  handles this by not using the shadow dom for the form
https://github.com/vaadin/vaadin-login/blob/9e76baf5e8a7dccf5e65116d0fc91b49ea3bdf2a/src/vaadin-login-form.html#L50

PeterBurner:
does not work with `@aws-amplify/ui-vue` either

richiewebgate:
same issue with aws-amplify-angular

kjellski:
This looks like a bug to every end-user - any hints on how to help fix this here?

BrnoPCmaniak:
So I tried to fix this. Following upon @anantoghosh I googled and found out that the most likely issue is that the form fields are scattered across different shadow DOM roots.

I also noticed this commit https://github.com/aws-amplify/amplify-js/commit/538ca667296568fc58d77ce5c45f201b59e085ea

So I tried to undo some of it works. And instead of using shadow dom for the whole thing I tried to force it to use fallback (`scoped: true` https://stenciljs.com/docs/component#component-decorator).

When I applied this to authenticator and everything underneath this have happend:

![Pokus-1](https://user-images.githubusercontent.com/1249759/94354258-d3d61e80-0079-11eb-91d2-acd8280068f5.png)

```
<amplify-authenticator>
  <amplify-sign-in slot="sign-in" hideSignUp="true" [formFields]="this.signInFormFields"></amplify-sign-in>
</amplify-authenticator>
```

[...]

PR #6454, opened 2020-07-30
Add inputs name for SignIn form

daaru00:
**Description**

Without an input name some browsers (Firefox for example) cannot understand that is a login form and does not trigger the "remind credentials" message. With these default values it should fix this annoy behaviour and let users to save the password and remember it on next login.


packages/amplify-ui-components/src/components/amplify-sign-in/amplify-sign-in.tsx
```diff
@@ -206,6 +207,7 @@ export class AmplifySignIn {
       case 'phone_number':
         formFieldInputs.push({
           type: 'phone_number',
+          name: 'tel',
```
ashika01, line 210: Could we have this as `phone_number` too. Also, looks like you are only making the change to `sign_in` page. I think it needs to be for all different formfield input in all of `ui-components`

daaru00: I've used the names that browser use to identify the input type and autocompletion https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete. I don't have a project with a Cognito configuration with phone/password login to check what happens. No problem to rename to "phone_number".

> Also, looks like you are only making the change to sign_in page. I think it needs to be for all different formfield input in all of ui-components
yeah, just because the main issue is during login.. it's really annoying to have to write emails and passwords every time without any help from the browser.

ashika01: Ideally I would suggest you look up for `formField` in the `ui-components` package and make the same change everywhere. And I can test the change by pulling your branch then.

andidev:
@ashika01 
This is really needed for us. Is there any workaround currently?

andidev:
@ashika01 @daaru00 
I just found out that input props can be set on formFields. So I did set it this way:
```jsx
<AmplifyAuthenticator>
    <AmplifySignIn
        slot="sign-in"
        formFields={[
            { type: "username", inputProps: {name: 'username', autocomplete: 'username'} },
            { type: "password", hint: '', inputProps: {name: 'password', autocomplete: 'current-password'} }
        ]}
        hideSignUp
    />
</AmplifyAuthenticator>
```
which lead to the following html
image

however the autofill for credentials still does not work.

wlee221:
Hi @daaru00, we'll be closing this because this improvement was made as part of #8181. As mentioned, autofill doesn't work due to shadow DOM. We're working to remove it completely -- please see #5782.

daaru00:
Hi @wlee221, happy to hear that! Close it without problems, I will take a look to what you reported, thanks!

PR #6458, opened 2020-07-30
fix(@aws-amplify/datastore): Allow subscribing and querying with partial auth

seang96:
In prior versions, DataStore would go through each subscription and query instead of failing all if one fails. If the app has roles or guest users that need access to only certain pieces of the data it now fails. This update allows it to try other requests like it previously used to.

Though this is a quick fix and how it worked in the past I assume in the future codegen models from the cli should be updated to store authentication information for each request and skip the ones they aren't authorized for.


ericclemmons:
Can you share a `schema.graphql` that I can test this behavior against?

From your description, it sounds like this change will allow guest queries to succeed even if auth queries fail? (Whereas a single failure fails all today?)

seang96:
Yes this will allow any single query that succeeds to be stored rather than none as it is today. Not only does this include support for unauthenticated vs authenticated users but also users that are authenticated that lack full authentication.

The example below would prevent guest and authenticated non-admin users from reading.

In my project's schema I also have limitations on reading attributes to a model which is what causes the queries to fail similarly, but those queries are not needed for guest / regular users (in my case it is a list of users that are a part of a custom membership, but authenticated users can know what custom memberships they have). If you want to see my project's schema I am willing to share privately.

```
type System
@model
@auth(rules: [
  {allow: groups, groups: ["admin"]},
  {allow: private, operations: [read]},
  {allow: public, provider: iam, operations: [read]}
])
{
  id: ID!
  name: String!
  overview: SystemOverview @connection
}

type SystemOverview
@model(queries: null)
@auth(rules: [
  {allow: groups, groups: ["admin"]},
  {allow: private, operations: [read]},
  {allow: public, provider: iam, operations: [read]}
])
{
  id: ID!
  level: String
  identityID: String
}

type Test
@auth(rules: [
  {allow: groups, groups: ["admin"]}
])
{
  id: ID!
  example: String
}
```

seang96:
I think I found it why this has been open for so long with no responses. I made an accidental error when I first tested this by having old data already synced from DataStore in my browser which made my original patch 'work'.

I dug into the code a little more and found that when the observers for subscriptions and queries (retrievePages) when emitted an error would bounce up to the DataStore start and cause a fail. I ended up checking in both cases if the error is Unauthorized to log the error as a warning to let developers be aware that it may be an issue, but could not be, and do not emit an error on the observer when the errorType is Unauthorized.

I also noticed that my IDE was performing some linting of its own on my commits and prevented that from occurring in my new commits.

This PR now reverts back from previous breaking changes that were made in an update. This by itself does not handle the authentication passed through Graphql, it just allows for unauthorized error from stopping DataStore from working.

In react I built three Hooks that detects when a user state changes, one that keeps the amplify state using react-tracked, and one that sets graphql authentication. The result is anything that requires useAmplify hook uses it and then gets amplifyState from useAmplifyState hook to check if it is loading or has finished. Default authentication is Cognito, when logged out switch to AWS_IAM. If you guys think this may be a good idea to pull into a react package, I also have hooks for useQuery, useSubscription, and useMutation to simplify DataStore coding with react not sure which package to put it into though.

User State Hook
```
export const useUser = (): CognitoUser | 'guest' | undefined => {
    const [user, setUser] = useState<CognitoUser | 'guest'>();

    // Check auth first time
    useEffect(() => {
        let unsubscribed = false;
        Auth.currentAuthenticatedUser()
            .then(data => {
                if (unsubscribed) { return; }
                setUser(data);
            })
            .catch(() => console.debug('User Not signed in'));
        return () => { unsubscribed = true; };
    }, []);

    // Subscribe to auth changes
    useEffect(() => {
        let unsubscribed = false;
        const updateUser: HubCallback = ({ payload: { event, data } }) => {
            if (unsubscribed) { return; }
            switch (event) {
            case 'signIn':
                setUser(data);
                break;
            case 'configured':
                Auth.currentAuthenticatedUser()
                    .then(setUser)
                    .catch(() => console.debug('User Not signed in'));
                break;
            default:
                setUser('guest');
                break;
            }
        }
        
        Hub.listen('auth', updateUser);
        return (): void => {
            unsubscribed = true;
            Hub.remove('auth', updateUser);
        }
    }, []);

    return user;
};
```

Amplify State
```
export interface AmplifyState {
    amplifyLoading: boolean;
    previousUser?: CognitoUser | 'guest';
}

const _useAmplifyState = () =>
    useReducer(
        (state: AmplifyState, newValue: Partial<AmplifyState>) => ({
            ...state,
            ...newValue,
        }),
        {
            amplifyLoading: true,
        }
    );

export const {
    Provider: AmplifyStateProvider,
    useTracked: useAmplifyState,
} = createContainer(_useAmplifyState);
```

Handling Graphql authentication
```
//DataStore needs schema loaded
import '../../models';

export const useAmplify = (): void => {
    const user = useUser();
    const [{ previousUser }, setAmplifyState] = useAmplifyState();

    const dataStoreReady = useCallback(
        ({ payload: { event } }) => {
            if (event === 'ready') {
                setAmplifyState({ amplifyLoading: false });
            }
        },
        [setAmplifyState]
    );

    useEffect(() => {
        // If user is undefined, useUser is still getting user information
        if (!user) {
            return;
        }
        const userString =
            user instanceof CognitoUser ? user.getUsername() : user;
        const previousUserString =
            previousUser instanceof CognitoUser
                ? previousUser.getUsername()
                : previousUser;
        Hub.listen('datastore', dataStoreReady);

        console.log('auth state', userString, previousUserString);

        // User is guest and previousUser is defined
        if (user === 'guest' && !!previousUser) {
            console.log('user logged out, clearing...');
            DataStore.clear()
                // Required to set amplifyLoading to false while partial authentication is unsupported in DataStore
                .then(() => setAmplifyState({ amplifyLoading: false }));
        }

        // Auth state changed
        if (userString !== previousUserString) {
            console.log('auth state changed, reloading amplify...');

            setAmplifyState({ amplifyLoading: true, previousUser: user });

            if (user) {
                console.log('auth');
                API.configure(awsconfig);
            } else {
                console.log('unauth');
                // eslint-disable-next-line camelcase
                API.configure({
                    ...awsconfig,
                    aws_appsync_authenticationType: 'AWS_IAM',
                });
            }

            DataStore.configure();
            DataStore.start();
        }

        return () => Hub.remove('datastore', dataStoreReady);
    }, [dataStoreReady, previousUser, setAmplifyState, user]);
};
```

iartemiev:
@seang96 thank you for the update and for opening this PR.
 
We're currently building up our internal integration test coverage around various combinations of `@auth` rules in DataStore models. We're expecting to have a comprehensive test baseline in the very near future. 

Once this happens, I will test your changes against those tests, and if no issues are detected, I will promptly merge the PR.

packages/datastore/src/sync/processors/sync.ts
```diff
@@ -85,10 +88,22 @@ class SyncProcessor {
...
+						query,
+						variables,
+					});
+				} catch (error) {
+					const unauthorized = (error.errors as [any]).some(err => err.errorType === 'Unauthorized');
+					if (unauthorized) {
+						const result = error;
+						result.data[opName].items = result.data[opName].items.filter(item => item !== null);
+						logger.warn('queryError', 'User is unauthorized, some items could not be returned.');
+						return result;
```
iartemiev, line 102: @seang96 we've validated the changes in `subscription.ts` and it's working as expected but just wanted to ask if you were getting these unauthorized errors in the base/delta sync queries as well, as this code suggests. If so, what are the steps to reproduce?

seang96: > 
> 
> @seang96 we've validated the changes in `subscription.ts` and it's working as expected but just wanted to ask if you were getting these unauthorized errors in the base/delta sync queries as well, as this code suggests. If so, what are the steps to reproduce (so that we can test this part of the PR)?

This is due to group the @auth rules declaration in the schema regarding specific access to groups with a group variable attached to each item. The user can query some items depending on the groups that they are in.

amhinson: @seang96 I'm still unable to see the unauthorized errors in `sync.js` when testing. Would you be able to share a sample schema that replicates the behavior?

seang96: > 
> 
> @seang96 I'm still unable to see the unauthorized errors in `sync.js` when testing. Would you be able to share a sample schema that replicates the behavior?

Try this sample:
```
type Project
@model
@auth(
    rules: [
        { allow: groups, groups: ["admin"] }
        { allow: private, operations: [read] }
        { allow: public, provider: iam, operations: [read] }
    ]
) {
    id: ID!
    name: String!
    apiIntegration: APIIntegration @connection
}

type APIIntegration
@model(queries: null)
@auth(
    rules: [
        { allow: groups, groupsField: "group", operations: [read] }
        { allow: groups, groups: ["admin"] }
    ]
) {
    id: ID!
    url: String!
    group: String!
}
```

Create a Project, then create an APIIntegration and connect it to project, login as a user that has access to Project, but does not belong into the group required for APIIntegration. when syncProject request goes through it will have an error "Not Authorized to access apiIntegration on type Project".

Another example that results in some values directly returning as null which causes error due to id being null on sync/index.ts at counts.deleted++ from the snippet below; note this is avoided by filtering null values from the items from my commit:
https://github.com/aws-amplify/amplify-js/blob/b0668ddb187ce756264cecf114768fac52f9af1b/packages/datastore/src/sync/index.ts#L535-L537

To test this one, two users should be created and a project with a role. Create a project membership that links with one user. On the other user, when it does syncProjectMemberships it should return that item as null or with filter from my commit, an empty list of memberships. Create a membership for the user your logged in as and you should only get that user's membership rather than an empty list. The use case for this schema is to hook with PreTokenGeneration to add users to custom groups for a specific project to get them access to certain parts of the application.

Sample:
```
type Project
@model
@auth(
    rules: [
        { allow: groups, groups: ["admin"] }
        { allow: private, operations: [read] }
        { allow: public, provider: iam, operations: [read] }
    ]
) {
    id: ID!
    name: String!
    roles: [ProjectRole] @connection(keyName: "rolesByProject", fields: ["id"])
}

type ProjectRole
@model
@key(name: "rolesByProject", fields: ["ProjectID", "name"])
@auth(
    rules: [
        { allow: private, provider: iam }
        { allow: groups, groups: ["admin"] }
        { allow: private, operations: [read] }
        { allow: public, provider: iam, operations: [read] }
    ]
) {
    id: ID!
    name: String!
    ProjectID: ID!
    Project: Project! @connection(fields: ["ProjectID"])
    users: [ProjectMembership]
    @connection(keyName: "byRole", fields: ["id"])
    @auth(
        rules: [
            { allow: private, provider: iam }
            { allow: groups, groups: ["admin"] }
        ]
    )
}

type ProjectMembership
@model(queries: null)
@key(name: "byUser", fields: ["userID", "roleID"])
@key(name: "byRole", fields: ["roleID", "userID"])
@auth(
    rules: [
        { allow: private, provider: iam }
        { allow: private, operations: [read] }
        { allow: groups, groups: ["admin"] }
    ]
) {
    id: ID!
    roleID: ID!
    userID: ID!
    role: ProjectRole! @connection(fields: ["roleID"])
    user: User! @connection(fields: ["userID"])
}

type User
@model
@auth(
    rules: [
        { allow: private, provider: iam }
        { allow: groups, groups: ["admin"] }
        {
            allow: owner
            ownerField: "id"
            identityClaim: "sub"
            operations: [read]
        }
    ]
) {
    id: ID!
    email: String!
    projects: [ProjectMembership] @connection(keyName: "byUser", fields: ["id"])
}
```

amhinson: Excellent! Thanks for the great sample. I was able to validate the changes you've made 👍

packages/datastore/src/sync/processors/sync.ts
```diff
@@ -85,10 +88,22 @@ class SyncProcessor {
...
+				try {
+					return await API.graphql({
+						query,
+						variables,
+					});
+				} catch (error) {
+					const unauthorized = (error.errors as [any]).some(err => err.errorType === 'Unauthorized');
+					if (unauthorized) {
+						const result = error;
+						result.data[opName].items = result.data[opName].items.filter(item => item !== null);
```
amhinson, line 100: Could you also add a comment here as to why this is needed?

seang96: > 
> 
> Could you also add a comment here as to why this is needed?

I did a new commit that has a comment to explain the unauthorize check.

amhinson (approved):
Thanks again @seang96! 🙏

Issue #5860, opened 2020-05-22
FederatedSignIn custom provider & custom state


djerraballi:
**Describe the bug**
If a custom provider and a custom state are used in conjunction, amplify incorrectly removes the custom provider and sets the provider to 'COGNITO' in the federated sign in flow. 

https://github.com/aws-amplify/amplify-js/blob/b4b40aa62644ea3688068e4e64f4573c568677a0/packages/auth/src/Auth.ts#L1733-L1736 

**To Reproduce**
Use a custom provider and a custom state together, see that your custom provider will be swapped out in the authorize url to the COGNITO. 

**Expected behavior**

the expectation is that if i am using custom provider, and customState is conditionally present, i shouldnt have to change the input parameter. 

**Code Snippet**

```
Auth.federatedSignIn({
  customProvider: 'some-provider',
  customState: 'something'
})
```

djerraballi:
My work around is in the cases where i need to use customProvider and customState, i change the input param of customProvider to provider. 

if customState is however present, i need to change customProvider back to provider. 

workaround does work, but seems unintentional?

manueliglesias:
Hi @djerraballi 

This is a bug, I am labeling the issue as such, thanks for the report!

And also a big thank you for sharing a workaround for other people that might encounter it.

div5yesh:
This issue is resolved in our latest release 3.0.23.

PR #3588, opened 2019-07-03
Allow customers to add a custom state to federatedSignIn method

sammartinez:
*Issue #, if available:*
Fixes #3188
Fixes #3194
Fixes #3177
Fixes #3125
Fixes #3569

*Description of changes:*
This will allow customers to create a custom state and pass it through the federated Sign In method. Docs PR #[770](https://github.com/aws-amplify/docs/pull/770)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6480, opened 2020-08-03
bugfix #5860: using custom state forces congnito UI for custom provider

div5yesh:
_Issue #, if available:_ #5860 

_Description of changes:_ 

- use provider/customProvider keys to determine the type of Provider instead of also relying on customState
- add check for customState key if no provider is passed

### DIFF
--- a/packages/auth/src/types/Auth.ts
+++ b/packages/auth/src/types/Auth.ts
@@ -91,8 +91,11 @@ export function isFederatedSignInOptionsCustom(
 	return obj && !!keys.find(k => obj.hasOwnProperty(k));
 }
 
-export function hasCustomState(obj: any): obj is FederatedSignInOptions {
-	const keys: (keyof FederatedSignInOptions)[] = ['customState'];
+export function hasCustomState(obj: any): boolean {
+	const keys: (
+		| keyof FederatedSignInOptions
+		| keyof FederatedSignInOptionsCustom
+	)[] = ['customState'];
 	return obj && !!keys.find(k => obj.hasOwnProperty(k));
 }
 
--- a/packages/auth/src/types/Auth.ts
+++ b/packages/auth/src/types/Auth.ts
@@ -92,10 +92,10 @@ export function isFederatedSignInOptionsCustom(
 }
 
 export function hasCustomState(obj: any): boolean {
-	const keys: (
-		| keyof FederatedSignInOptions
-		| keyof FederatedSignInOptionsCustom
-	)[] = ['customState'];
+	const keys: (keyof (
+		| FederatedSignInOptions
+		| FederatedSignInOptionsCustom
+	))[] = ['customState'];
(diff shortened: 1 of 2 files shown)


packages/auth/src/types/Auth.ts
```diff
@@ -80,17 +80,19 @@ export type FederatedSignInOptionsCustom = {
 export function isFederatedSignInOptions(
 	obj: any
 ): obj is FederatedSignInOptions {
-	const keys: (keyof FederatedSignInOptions)[] = ['provider', 'customState'];
+	const keys: (keyof FederatedSignInOptions)[] = ['provider'];
```
sammartinez, line 83: there was a reason I wrote this before... @manueliglesias can you remember as to why?

div5yesh: Yes, I saw this in a year old commit.
https://github.com/aws-amplify/amplify-js/pull/3588/files

packages/auth/src/types/Auth.ts
```diff
@@ -80,17 +80,19 @@ export type FederatedSignInOptionsCustom = {
...
 ): obj is FederatedSignInOptionsCustom {
-	const keys: (keyof FederatedSignInOptionsCustom)[] = [
-		'customProvider',
-		'customState',
-	];
+	const keys: (keyof FederatedSignInOptionsCustom)[] = ['customProvider'];
+	return obj && !!keys.find(k => obj.hasOwnProperty(k));
+}
+
+export function hasCustomState(obj: any): obj is FederatedSignInOptions {
```
elorzafe, line 94: `obj is FederatedSignInOptions` it seems odd to me because it can also be `FederatedSignInOptionsCustom` right?

div5yesh: Yes, I think just `bool` return type will do.

packages/auth/src/types/Auth.ts
```diff
@@ -91,8 +91,11 @@ export function isFederatedSignInOptionsCustom(
 	return obj && !!keys.find(k => obj.hasOwnProperty(k));
 }
 
-export function hasCustomState(obj: any): obj is FederatedSignInOptions {
-	const keys: (keyof FederatedSignInOptions)[] = ['customState'];
+export function hasCustomState(obj: any): boolean {
+	const keys: (
+		| keyof FederatedSignInOptions
+		| keyof FederatedSignInOptionsCustom
+	)[] = ['customState'];
```
manueliglesias, line 98: Your change works fine, I'd suggest narrowing down the type further like this: (so the resulting type is `const keys: "customState"[]` instead of `const keys: ("provider" | "customState" | "customProvider")[]`)

```suggestion
	const keys: (keyof (
		| FederatedSignInOptions
		| FederatedSignInOptionsCustom
	))[] = ['customState'];
```

PR #6491, opened 2020-08-04
Add draft chatbot ui

wlee221:
_Issue #, if available:_ 

_Description of changes:_ This adds a draft implementation of the chatbot. It has working visual design with no functionalities yet, which will be added in later PRs.

- Add icon variant to `amplify-button`
- Expose additional css variables in `amplify-button`, `amplify-icon`, and `amplify-input`.
- Add microphone and send icon SVG to `amplify-icon`.
- Remove `span` inside `amplify-icon`.

Screen Shot 2020-08-11 at 3 53 32 PM


packages/amplify-ui-components/src/components/amplify-chatbot/amplify-chatbot.spec.tsx
```diff
@@ -0,0 +1,21 @@
+// import { newSpecPage } from '@stencil/core/testing';
+// import { AmplifyChatbot } from './amplify-chatbot';
+
+describe('amplify-chatbot', () => {
```
jordanranz, line 4: Let's clean up these tests :D

wlee221: Sorry about the late reply. I added a snapshot test now!

packages/amplify-ui-components/src/components/amplify-chatbot/amplify-chatbot.scss
```diff
@@ -0,0 +1,71 @@
...
+  font-weight: bold;
+  text-align: center;
+  padding-bottom: 24px;
+  border-bottom: 1px solid rgba(0, 0, 0, 0.05);
+}
+.body {
+  padding-top: 2rem;
+  display: flex;
+  flex-direction: column;
+  height: 500px;
```
ashika01, line 28: Same as above. Lets move everything to `rem`

packages/amplify-ui-components/src/components/amplify-icon/amplify-icon.scss
```diff
@@ -1,7 +1,12 @@
 :host {
+  display: inline-block;
   --icon-fill-color: var(--amplify-white);
+  --width: auto;
+  height: var(--height);
 }
 
-.icon {
+svg {
```
ashika01, line 8: Have you tested to see if this messes up any icon in existing usage?!

packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
```diff
@@ -55,6 +58,7 @@ export class AmplifyButton {
         disabled={this.disabled}
         onClick={this.handleClick}
       >
+        {this.variant === 'icon' && <amplify-icon name={this.icon}></amplify-icon>}
```
ashika01, line 61: Could you provide a snapshot on how this change looks like?! Example of an icon as button I guess?!

ashika01:
Had minor overall feedback. Looking sharp in general 👍

packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
wlee221: Yeah I should have put a screenshot! I'll update the post.

wlee221: they are the microphone and the send button bottom right. They are clickable icons, behaving like https://polaris.a2z.com/components/awsui-button/?example=icon-button.

packages/amplify-ui-components/src/components/amplify-button/amplify-button.scss
```diff
@@ -10,6 +10,9 @@
   --link-hover: var(--amplify-primary-tint);
   --link-active: var(--amplify-primary-shade);
   --text-transform: uppercase;
+  --icon-color: var(--amplify-primary-color);
+  --icon-height: 1.25rem;
```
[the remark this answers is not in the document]
wlee221: Nice catch!

packages/amplify-ui-components/src/components/amplify-chatbot/amplify-chatbot.scss
wlee221: Gotcha

packages/amplify-ui-components/src/components/amplify-icon/amplify-icon.scss
wlee221: Yes, there seemed to be no problem because the span with .icon class was just an empty wrapper.

packages/amplify-ui-components/src/components/amplify-chatbot/amplify-chatbot.scss
```diff
@@ -0,0 +1,71 @@
...
+  --header-size: var(--amplify-text-lg);
+}
+.amplify-chatbot {
+  display: inline-block;
+  background-color: var(--background-color);
+  border-radius: 6px;
+  box-shadow: 1px 0px 4px 0 rgba(0, 0, 0, 0.15);
+  box-sizing: border-box;
+  font-family: var(--amplify-font-family);
+  width: 460px;
```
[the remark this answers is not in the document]
wlee221: Right! We do need a min-width, especially when I make the width customizable. I'll add that in later PRs.

packages/amplify-ui-components/src/components/amplify-icon/amplify-icon.scss
ashika01: yeah it might not affect this one, but any svg I think. This style would apply to any svg on a page. So you might have to bring this style under `amplify-icon`

wlee221: I see, because `amplify-icon` is not shadowed so that is probably true. I'll make that change, thanks!

wlee221:
Thanks @ashika01! Making the changes now.

packages/amplify-ui-components/src/components/amplify-icon/amplify-icon.scss
wlee221: I tested this out, and it looks like this was only scoped to `svg`s inside `amplify-icon`. I will still test that these PRs do not affect existing components before merging :D

packages/amplify-ui-components/src/components/amplify-button/amplify-button.scss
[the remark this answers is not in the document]
wlee221: Upon second look, I think I need to have this name different from `--icon-fill-color` because I use it to set `--icon-fill-color` in line 72. I tried just setting `--icon-fill-color` just at `:host` but then it got overwritten later in `amplify-icon`. I'll try to find cleaner alternative. Do you know a better way to do this?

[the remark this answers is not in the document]
ashika01: you could just change `--icon-color` to `--icon-fill-color` and then send it down to `amplify-icon`. Because for customer styling, they will only be exposed to this from `amplify-button` level. 

This,
`--icon-color: var(--amplify-primary-color);`

will become
`--icon-fill-color: var(--amplify-primary-color);`

packages/amplify-ui-components/src/components/amplify-button/amplify-button.scss
```diff
@@ -59,6 +62,18 @@
...
 }
 
+.icon {
+  background-color: inherit;
+  border: none;
+  font: inherit;
+  cursor: pointer;
+  padding: var(--padding);
+  amplify-icon {
+    --icon-fill-color: var(--icon-color);
```
ashika01, line 72:
```suggestion
    --icon-fill-color: var(--icon-fill-color);
```

packages/amplify-ui-components/src/components/amplify-button/amplify-button.scss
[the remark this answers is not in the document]
wlee221: I tried that (with suggested code commit below), but it did not set the fill color properly as opposed to using a different name.

When using `icon-fill-color`: 

Screen Shot 2020-08-12 at 12 58 48 AM

When using `icon-color`:

Screen Shot 2020-08-12 at 1 08 05 AM

This is weird; maybe css has trouble passing down css variable with itself, in scoped context?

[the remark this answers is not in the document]
ashika01: Ah interesting.. Yeah Lets leave this for now then 👍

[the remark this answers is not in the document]
ashika01: maybe just `icon-fill` rather than `icon-color`

So, in css, [fill](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill) and [color](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/color) on svg have different meaning. Wanna make we represent it right.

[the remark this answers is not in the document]
wlee221: Sounds good! Will make that change.

PR #6499, opened 2020-08-05
Update api pubsub peer dependencies to 3.0.22

joeltio:
_Issue #, if available: #6061_

_Description of changes:_
- Update peer dependencies of `api` and `api-graphql` to 3.0.22
- Remove unnecessary pubsub dependency when already listed in peer dependencies.

It seems that the dependencies listed on `api` and `api-graphql` are incorrect. I believe this is an error since in the [git blame](https://github.com/aws-amplify/amplify-js/blame/572e746b83e6ae3b3e2f8858163941e081f3a4e5/packages/api-graphql/package.json#L40-L52), the `dependencies` was updated automatically very recently while the `peerDependencies` seems untouched since 8 months ago.


joeltio:
It looks like the dependency on `pubsub` in `api-graphql` was added in #5352 to prevent race conditions. I have added them back but now in `devDependencies`. Checking the lerna topological sort, it has the same effect and I think it makes more sense since its sole purpose of being there is for lerna to build the packages in the right order and not for users to install.

packages/api-graphql/package.json
```diff
@@ -35,20 +35,20 @@
...
     "@aws-amplify/cache": "^3.1.21",
     "@aws-amplify/core": "^3.4.4",
-    "@aws-amplify/pubsub": "^3.0.22",
     "graphql": "14.0.0",
     "uuid": "^3.2.1",
     "zen-observable-ts": "0.8.19"
   },
   "peerDependencies": {
-    "@aws-amplify/pubsub": "^2.1.1"
+    "@aws-amplify/pubsub": "^3.0.22"
```
ericclemmons, line 51: Thanks for this 👍

packages/api/package.json
```diff
@@ -42,7 +42,7 @@
     "@aws-amplify/api-rest": "^1.1.4"
   },
   "peerDependencies": {
-    "@aws-amplify/pubsub": "^2.1.1"
+    "@aws-amplify/pubsub": "^3.0.22"
   },
```
ericclemmons, line 46: I was just checking and I think this can be safely removed: `@aws-amplify/api-graphql` has the dependency (and peerDependency) on `@aws-amplify/pubsub`, so there's no explicit usage within `@aws-amplify/api`.

packages/api-graphql/package.json
```diff
@@ -35,20 +35,20 @@
...
-    "@types/zen-observable": "^0.8.0"
+    "@types/zen-observable": "^0.8.0",
+    "@aws-amplify/pubsub": "^3.0.22"
   },
   "dependencies": {
     "@aws-amplify/api-rest": "^1.1.4",
     "@aws-amplify/auth": "^3.3.3",
     "@aws-amplify/cache": "^3.1.21",
     "@aws-amplify/core": "^3.4.4",
-    "@aws-amplify/pubsub": "^3.0.22",
```
ericclemmons, line 45: This should say so that customers can use `@aws-amplify/api-graphql` without having to explicitly install `@aws-amplify/pubsub`.

`@aws-amplify/api-graphql` imports `@aws-amplify/pubsub` alongside `@aws-amplify/core`, so it makes sense to keep this for safety alone.

Issue #5040, opened 2020-03-04
Models not saving due to incorrect data model, again.


raelmiu:
**Describe the bug**
This morning all my saves had stopped working. Again.

**To Reproduce**
1. Add any model, with an undefined "createdAt".
2. DataStore sync comes back with an error. This error also seems to be impossible to solve.

**Expected behavior**
Sync should work. Or the error should allow me to debug.
It is also surprising that my app changes behavior even though _I'm not making any changes_. How is that happening? No packages were updated?

**Screenshots**
Original error:
<img width="670" alt="Screenshot 2020-03-04 at 17 00 40" src="https://user-images.githubusercontent.com/186083/75898528-760e8a80-5e3a-11ea-8580-560fe5b16e02.png">

Empty string added:
<img width="672" alt="Screenshot 2020-03-04 at 17 00 09" src="https://user-images.githubusercontent.com/186083/75898554-81fa4c80-5e3a-11ea-90a1-7347c2ea6e03.png">

Date string added:
<img width="674" alt="Screenshot 2020-03-04 at 17 00 16" src="https://user-images.githubusercontent.com/186083/75898572-8aeb1e00-5e3a-11ea-852a-542522a37b63.png">

Date ISO string added:
<img width="658" alt="Screenshot 2020-03-04 at 17 03 39" src="https://user-images.githubusercontent.com/186083/75898607-94748600-5e3a-11ea-96a9-c40e980c38ae.png">

Seems like AppSync has changed how it handles models, but updating my packages does nothing?

raelmiu:
Running the mutation from the AppSync console works fine, with the same data. So this seems to be a syncing error in datastore. If I discover anything else I'll update this thread.

This is the graphql
```graphql
type Ting
  @model
  @auth(rules: [{ allow: owner }, { allow: groups, groups: ["admins"] }]) {
  id: ID!
  createdAt: AWSDateTime
  references: [String]
  referencedBy: [String]
  content: String!
  access: [UserTing] @connection(keyName: "byUser", fields: ["id"])
}
```

raelmiu:
If I run DataStore locally from the amplify-js repo, it works fine. So this is only an issue in NPM version `1.0.8` of DataStore

raelmiu:
Still having this issue, if an array is empty it doesn't save.
If an array only contains an empty string, it doesn't save.

raelmiu:
Can't save any models now, getting an undecipherable error:

[...]

PR #6500, opened 2020-08-05
Nullable arrays in models

ondojo:
Fix array notation bug related to #5040

_Issue #, if available:_ #5040 

_Description of changes:_ Validation of model ignored the isRequired attribute in case the attribute value was of type array. Added this consideration to the if clause to allow for graphlQL annotations such as `[foobar]` or `[foobar!]`. Now the attributes can be set to null or empty arrays and the sync still works.


ondojo:
Hello,

the circle ci error message says 

```bash
Creating workspace archive...

Uploading workspace archive...

Error uploading workspace archive: AccessDenied: Access Denied
status code: 403, request id: 444FF68393D2C9DC, host id: 5giQT2G09PD0Ig+4Ei+SZLZXJo1sYyleelJIH9wZB15H9ZAAcjO5WKcN0Q6Yeb456o/6lO2INlg=
```

Is there any Chance we can fix this?

mauerbac:
hi @ondojo - sorry for the delay. I just looked at the CircleCi tests on my end and it looks like one of the datastore unit tests failed. Are you able to run these locally and update the test? Thanks for the contribution

packages/datastore/src/datastore/datastore.ts
```diff
@@ -241,13 +241,13 @@ const validateModelFields = (modelDefinition: SchemaModel | SchemaNonModel) => (
...
 			if (isArray) {
-				if (!Array.isArray(v)) {
+				if (!Array.isArray(v) && isRequired) {
 					throw new Error(
 						`Field ${name} should be of type ${jsType}[], ${typeof v} received. ${v}`
 					);
 				}
 
-				if ((<[]>v).some(e => typeof e !== jsType)) {
+				if (v !== null && (<[]>v).some(e => typeof e !== jsType)) {
```
seang96, line 250: I believe this should check for v being undefined as well. I have two reasons for this:

1. The require check at line 236 checks for undefined as well
2. Typescript allows copyOf records to be undefined but not null

Another note, 2 opens up a new issue. Setting the value to undefined will result in it being ignored as an input in the graphql request resulting in the field not being updated. If set to null it works, but Typescript checks says null is not valid in the copyOf mutation. I am unsure if this is due to this PR or it is an existing issue with DataStore that I discovered upon testing this PR.

sammartinez:
Resolving as the above PR looks to have been taken care of in a separate pull request which has be related to this pull request.

Issue #6408, opened 2020-07-24
ExpiredToken Error after 1 hr during uploading large file


patelvarun:
We use AWS amplify JS library in our React - UI with Cognito auth role to upload files. we are storing session details (like token) in react state to access in the upload page. 

when we upload a bigger file (in GBs like above 5 GB), it’s failing with error  - "AWSS3Provider - error uploading ExpiredToken: The provided token has expired while uploading large file".

we did some research on it. we found that we have default accessToken time 1 hr in Cognito.  the aws-amplify library will automatically handle refreshing access token (using refreshToken) when we do some action on the page & it will refresh the page. it is working fine. But in case of bigger file upload, the token is getting expired after 1 hr since we are not doing any action on page while uploading so it won't get refresh token & There is no option to increase expire time in Cognito. 
If I manually get refresh token and set to react state by calling refreshSession method before refreshToken expire (before 1 hr end) then the page is refresh because of status change. so I lost my uploaded files. 

Below are my questions
1) How we can increase refresh token time? 
2) Is there any option to increase it then, How we can get refresh token & pass it to uploading file once token expire?
3) What is the best way to handle this?

patelvarun:
Hi, Can anyone help me out on this? Let me know if you require more information on this. 
Thanks in advance.

patelvarun:
Is this issue being considered or is it feature-request?

sammartinez:
@patelvarun Thanks for this feedback, there is a similar feature request open as I have related. We will take this into consideration for our project.

patelvarun:
Thanks for considering this as a feature request.

iartemiev:
This feature has been released as part of aws-amplify@3.0.23 / @aws-amplify/storage@3.2.12

patelvarun:
@iartemiev - I appreciate you! Thanks for the support.

Issue #4179, opened 2019-10-13
Refreshing Auth Token for Unauthenticated Cognito Users during storage.put


DoctahPopp871:
** Which Category is your question related to? **
Auth/Storage
** What AWS Services are you utilizing? **
Cognito, s3
** Provide additional details e.g. code snippets **
We are able to successfully authenticate and initiate uploads to s3. We have added transfer acceleration to the storage module and it works very well. Our current issue is that the auth token expires after 1 hour, and our uploads stop, and fail due to the expired token. Ive seen many examples and "solutions" in other issues but they just dont fit our use-case. 
https://github.com/aws-amplify/amplify-js/issues/446#issuecomment-373904697
https://github.com/aws-amplify/amplify-js/issues/3310
#6408 

Does amplify support storage puts that last for longer than an hour, when using Cognito Identity pools ? 

If so, how ? Ive attached a snippit of code from our file upload attempt, where we have tried to refresh the token prior to the hour mark. We seem to be able to do so, however, that token is not used during the s3 upload. help!

[s3client.txt](https://github.com/aws-amplify/amplify-js/files/3722577/s3client.txt)

elorzafe:
@DoctahPopp871 how large are your files? or is because is a slow connection? We may need to consider this use case @pravgupt

DoctahPopp871:
Hello @elorzafe,

These are large genomic files, 100GiB. Our use-case, is unique, but we have high speed low latent connections we are utilizing when moving these large files. We can move 60-70GiB files under the hour mark easily. ~100GiB seems to be when we reach the hour token timeout threshold. 
As mentioned, we ensured the fastest transmission by enabling transfer acceleration on our buckets, and adding that functionality to our forked version of the storage module. 

This feature would be a boon for us, and assist in helping solve a tricky health problem. Thanks!

DoctahPopp871:
I started looking into where this change might be implemented, and notice there already seems to be a function handling the refresh of tokens in the storage provider https://github.com/aws-amplify/amplify-js/blob/master/packages/storage/src/Providers/AWSS3Provider.ts#L195 

Which leverages the core credentials file, and this function `_pickupCredentials` https://github.com/aws-amplify/amplify-js/blob/master/packages/core/src/Credentials.ts#L59

[...]

PR #6509, opened 2020-08-06
fix(@aws-amplify/storage): fix s3 multipart upload for very large files

iartemiev:
Addresses:
#4179 and #6408 

Check/refresh credentials after each multipart upload chunk to allow for uploads that take > 1hr to complete.


packages/storage/src/providers/AWSS3ProviderManagedUpload.ts
```diff
@@ -76,7 +77,7 @@ export class AWSS3ProviderManagedUpload {
 			// Multipart upload is not required. Upload the sanitized body as is
 			this.params.Body = this.body;
 			const putObjectCommand = new PutObjectCommand(this.params);
-			const s3 = this._createNewS3Client(this.opts, this.emitter);
+			const s3 = await this._createNewS3Client(this.opts, this.emitter);
```
elorzafe, line 80: I am not sure If `this.opts` is passed by reference or value?

If is value
I would refactor a little bit `_createNewS3Client` function to use the updated credentials instead of using `config` argument (that contains the updated credentials)

If is passed by reference nvm :)

iartemiev: It should be by ref, but I refactored it either way :)
https://github.com/aws-amplify/amplify-js/pull/6509/files#diff-f24d9b3f7d58ec285e3a34b45991cc58R334-R335

packages/storage/__tests__/providers/AWSS3ProviderManagedUpload-unit-test.ts
```diff
@@ -136,7 +137,7 @@ describe('multi part upload tests', () => {
 			protected minPartSize = testMinPartSize;
 			protected async uploadParts(uploadId: string, parts: BodyPart[]) {
 				// Make service calls and set the event listeners first from the base impl
-				super.uploadParts(uploadId, parts);
+				await super.uploadParts(uploadId, parts);
```
elorzafe, line 140: I would save the promise and await it after the emitter code. What do you think?

Issue #6076, opened 2020-06-12
Overriding the .section class within the Vue Authenticator component


sammartinez:
**Is your feature request related to a problem? Please describe.**
Customer within the discussion issue is looking to have the ability to override the `.section` class within the Vue Authenticator component.

**Describe the solution you'd like**
Provide a solution in order to override the `.section` styling within the Authenticator component.

**Additional context**
#6047

sammartinez:
Resolving as above PR is merged and will be going out in the next release

PR #6531, opened 2020-08-07
Authenticator style overrides

dozyio:
Issue #6076 fix

Description of changes:
Add a few more styling overrides to the Authenticator box - box-shadow, border-radius, padding & margin bottom

Additional context
#6047 

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.scss
+++ b/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.scss
@@ -4,4 +4,8 @@
   --min-width: 20rem;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  --box-shadow: 1px 1px 4px 0 rgba(0, 0, 0, 0.15);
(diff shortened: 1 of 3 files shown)


jordanranz:
Hey @dozyio, this is a great update. To keep the global theme cleaner, we can actually add css variable to the `amplify-authenticator.scss` file and cascade them down to the section.

Could you move the code in `theme.ts` to [here](https://github.com/aws-amplify/amplify-js/blob/main/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.scss#L7) and remove the `amplify` prefix like:
```
--box-shadow: 1px 1px 4px 0 rgba(0, 0, 0, 0.15);
--border-radius: 6px;
--padding: 35px 40px;
--margin-bottom: 20px;
```

Then just change the variables used in `section`

dozyio:
Thanks @jordanranz for the feedback - yes that does make more sense. New to the codebase so wasn't sure where to put the vars. Just pushed the commit and it seems to work fine on chrome / firefox / safari in a Vue app. Also I had to use amplify-authenticator { ... } as the css selector to style it, opposed to :root

PR #6542, opened 2020-08-10
fix(@aws-amplify/datastore): Retry mutation if GraphQL timeout due to bad network condition.

nubpro:
_Issue #, if available:_
I didn't open a new issue, but you can reproduce the problem as follow:
1. Open a react native app on iOS.
2. Go to device's Settings > Developer > Network Link Conditioner > Enable and select 100% Loss.
3. Immediately switch back to the app, update a record.
4. Approximately 15 seconds later, you'll see a warning appearing on your app. (you can refer to the attachment below)
image

The consequences of not handling this error is that DataStore will dequeue this mutation and the record will never be updated to the server which is a big nono.

_Description of changes:_
This PR will make sure mutation processor to perform retry if the mutation GraphQL returns `timeout of 0ms exceeded`.

### DIFF
--- a/packages/datastore/src/sync/processors/mutation.ts
+++ b/packages/datastore/src/sync/processors/mutation.ts
@@ -228,10 +228,11 @@ class MutationProcessor {
 					} catch (err) {
 						if (err.errors && err.errors.length > 0) {
 							const [error] = err.errors;
+							const { originalError: { code = null } = {} } = error;
 
 							if (
 								error.message === 'Network Error' ||
-								error.message === 'timeout of 0ms exceeded'
+								code === 'ECONNABORTED'
 							) {
 								if (!this.processing) {
 									throw new NonRetryableError('Offline');
--- a/packages/datastore/src/sync/processors/mutation.ts
+++ b/packages/datastore/src/sync/processors/mutation.ts
@@ -232,7 +232,7 @@ class MutationProcessor {
 
 							if (
 								error.message === 'Network Error' ||
-								code === 'ECONNABORTED'
+								code === 'ECONNABORTED' // refers to axios timeout error caused by device's bad network condition
 							) {
 								if (!this.processing) {
 									throw new NonRetryableError('Offline');
--- a/packages/datastore/src/sync/processors/mutation.ts
+++ b/packages/datastore/src/sync/processors/mutation.ts
@@ -228,7 +228,11 @@ class MutationProcessor {
 					} catch (err) {
 						if (err.errors && err.errors.length > 0) {
 							const [error] = err.errors;
-							if (error.message === 'Network Error') {
+
+							if (
+								error.message === 'Network Error' ||
+								error.message === 'timeout of 0ms exceeded'
+							) {
 								if (!this.processing) {
 									throw new NonRetryableError('Offline');
 								}
--- a/.github/actions/bundle-size-action/webpack/package.json
+++ b/.github/actions/bundle-size-action/webpack/package.json
@@ -43,7 +43,7 @@
       },
       {
         "path": "dist/withSSRContext.js.min.js",
-        "maxSize": "145kB"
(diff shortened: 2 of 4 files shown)


packages/datastore/src/sync/processors/mutation.ts
```diff
@@ -221,7 +221,11 @@ class MutationProcessor {
 					} catch (err) {
 						if (err.errors && err.errors.length > 0) {
 							const [error] = err.errors;
-							if (error.message === 'Network Error') {
+
+							if (
+								error.message === 'Network Error' ||
+								error.message === 'timeout of 0ms exceeded'
```
ericclemmons, line 227: This looks related to our underlying `axios` implementation:
> https://github.com/axios/axios/issues/2103#issuecomment-530516464

@manueliglesias Is this a resilient enough message to key off of?

manueliglesias: Good catch @ericclemmons, I don't think it would be resilient enough, I think it would be better to include additional keys in the GraphQL error if needed on `GraphQLAPI.ts` (e.g. an error code with values like the `ECONNABORTED` mentioned in the linked issue, or similar)

https://github.com/aws-amplify/amplify-js/blob/c1352845fcec30057b257d945ea5d3f14d339aeb/packages/api-graphql/src/GraphQLAPI.ts#L291

nubpro: Hey thanks for the review! Would you guys do the honor to make the changes so they would work gracefully.

I'll be happy to test it out on my end!

nubpro:
Hey guys, I've also noticed I would get 'Execution timeout' as well.

```
{
  "localModel": {
      "x": "xxxx"
  },
  "message": "Execution timed out.",
  "operation": "Update",
  "errorType": "ExecutionTimeout",
  "errorInfo": null,
  "remoteModel": null
}
```

Can I get some heads up as to why this is happening? Is it from axios too?
I was able to reproduce this issue by going offline, and doing a mutation, then going online afterwards.

packages/datastore/src/sync/processors/mutation.ts
nubpro: May I know how can I pass the errorCode into `GraphQLError` interface?

I was thinking of passing the original error, is that ideal?
```
			response = {
				data: {},
				errors: [new GraphQLError(err.message, null, null, null, null, err)],
			};
```

iartemiev:
I was able to successfully test the PR in iOS simulator

1. Reproduced the issue by following the steps in the description with the current latest version of aws-amplify
2. Built and published the changes in the PR to my locally-running Verdaccio registry
3. Installed the aws-amplify dependencies into the sample app from Verdaccio
4. Confirmed that the changes in the PR resolve the issues, namely:
  a. Network Error with code ECONNABORTED gets handled
  b. Reachability is set to `false`
  c. After Network Link Conditioner is disabled/network condition is stable, Reachability is set to `true` and the mutation is retried and the request succeeds
  d. Subsequent mutations succeed as well

Since it's not possible to simulate 100% Packet Loss (or other network conditions) in Detox (the e2e testing framework we use for RN apps) we won't be able to add an e2e test for this fix but we should be able to at least add a unit test that captures this. I will work on adding the test today and will merge the PR afterward.

Thank you so much for this contribution, @nubpro!

iartemiev:
Unit test has been added: https://github.com/aws-amplify/amplify-js/pull/6542/files#diff-b21a0f0d7db920e7ce7ad487d762f40fc3aca6b288e621a8db17347f50a480db

Going to ask @amhinson to give the test a once-over and then we'll merge this  PR today.

iartemiev (approved):
Approving. See previous comment for more info

amhinson (approved):
LGTM 👍

Issue #3938, opened 2019-08-29
 Force refresh of credentials with updated role assignments


Cayce:
** Which Category is your question related to? **
Auth, Credentials

** What AWS Services are you utilizing? **
Auth, Storage, PubSub

** Provide additional details e.g. code snippets **
I'm using federatedSignIn, with user pool, identity pool, and LWA.
After successful login user gets credentials attached to an authenticated role, when user attributes change, he should get a different role by "Authenticated role selection" rules in identity pool, but his permissions don't refresh until I run "react-native run-android" again. How can I make Amplify get new credentials by code?

Cayce:
My mistake: re-login doesn't update credentials. Only after I re-run the app with "react-native run-android" it updates its credentials, which means there is a way to update credentials for sure.

I found this https://github.com/aws-amplify/amplify-js/issues/446#issuecomment-481704572 suggested solution, but it doesn't work for me. The issue there is open and dealing with updating token, which is apparently not enough in this case.

JakubRakus:
I have the same issue, in my app roles and permissions of a user are changing when user make specific actions. Tried various solution form #446 and other related bugs/issues but they doesn't work. The problem is that Amplify lacks an ability/function/method to manually reload the session and get new tokens. I've checked session tokens before and after invoking various solutions from #446, but tokens remains the same.
Currently I'm using very crude solution, but looks like it works: after user actions that should change permissions I just sign out and sign in again:
```
async function refreshCredentials(usr, psw) {
    console.log('creds before', await Auth.currentUserCredentials());
    await Auth.signOut();
    await Auth.signIn(usr, psw);
    console.log('creds after', await Auth.currentUserCredentials());
}
```

Cayce:
This will refresh credentials(access key, secret key and session token):
>  Auth.currentCredentials().then((cred) => {
>        cred.refresh();
>     });

But the user will still have the same permissions.

Lollipop001:
We are dealing with the same issue!
We have policy in role for "Authenticated role" which allows invoking API gateway, and looks something like this(asterisks here only for simplicity):
>       {
>             "Effect": "Allow",
>             "Action": [
>                 "execute-api:Invoke"
>             ],
>             "Resource": [
>                 "*"
>             ]
>         }

Then, I have a rule for "Authenticated role selection" that "says" if some user custom attribute equals **true**, then it should assign a different role to the user. The different role policy looks something like this:
>         {
>             "Effect": "Allow",
[...]

PR #6581, opened 2020-08-14
Force clear credentials when bypassCache: true

elorzafe:
_Issue #, if available:_
Fixes: #3938
_Description of changes:_
Clearing credentials when `bypassCache: true` for `Auth.currentAuthenticatedUser`, that will lazy allow to lazy load AWS Credentials with using the refreshed token.

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -1150,7 +1150,7 @@ export class AuthClass {
 					}
 
 					// refresh the session if the session expired.
-					user.getSession((err, session) => {
+					user.getSession(async (err, session) => {
 						if (err) {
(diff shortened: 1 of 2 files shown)


packages/auth/src/Auth.ts
```diff
@@ -1150,7 +1150,7 @@ export class AuthClass {
 					}
 
 					// refresh the session if the session expired.
-					user.getSession((err, session) => {
+					user.getSession(async (err, session) => {
```
harrysolovay, line 1153: Why does this callback fn need to be async?

harrysolovay: Ah nevermind, I see the `await` now

PR #6594, opened 2020-08-17
feat(aws-amplify-react-native): Add S3Text

cedricgrothues:
_Issue #, if available:_ N/A

_Description of changes:_ This PR is part of the **React Native Storage Refactor** and adds an `S3Text` component to `aws-amplify-react-native`.

S3Text Demo


packages/aws-amplify-react-native/src/Storage/S3Text.tsx
```diff
@@ -0,0 +1,162 @@
...
+import AmplifyTheme, { AmplifyThemeType } from '../AmplifyTheme';
+import { AccessLevel } from './common/types';
+
+const logger = new Logger('Storage.S3Text');
+
+interface IS3TextProps {
+	textKey?: string;
+	path?: string;
+	body?: string;
+	contentType?: string;
```
cedricgrothues, line 14: @ashika01 This, similar to S3Album's contentType, should probably be removed.

Also on lines 27, 42, 57, 67, and 86.

ashika01: I dont think you need to here. Either case I will go through and let you know.

ashika01:
@cedricgrothues Looks like this is ready for review, could you change the status then?

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+				setText(source);
+			} catch (error) {
+				logger.error(error);
+				setError(error);
+			}
+		})();
+	}, [textKey, path, body]);
+
+	return [text, error];
```
amhinson, line 116: What are your thoughts on returning an object here instead of an array? That could give us more flexibility in the future if we want to change or add anything to this API. As an array, the only option would be too add another index, which could make it more obscure as the number grows.

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
```diff
@@ -0,0 +1,162 @@
...
+}: IUseS3Text): [string?, Error?] => {
+	if (!textKey && !path) {
+		logger.debug('empty textKey and path');
+		return [undefined, new Error('empty textKey and path')];
+	}
+
+	const key = (textKey ?? path) as string;
+
+	const [text, setText] = useState<string | undefined>();
+	const [error, setError] = useState<Error | undefined>();
```
amhinson, line 77: These could be written as `string?` and `Error?`, right?

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
```diff
@@ -0,0 +1,162 @@
...
+					level,
+					track,
+					identityId,
+					logger
+				);
+
+				setText(source);
+			} catch (error) {
+				logger.error(error);
+				setError(error);
```
amhinson, line 111: When setting the error or text, it would be good to also clear the other. It might make sense to just combine the two `useState` values into one object so you can update them together, i.e.:
`setState({ text: source, error: undefined })`

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+export const getTextSource = async (
+	key: string,
+	level: AccessLevel,
+	track: boolean,
+	identityId: string,
+	logger: Logger
+) => {
+	if (!Storage || typeof Storage.get !== 'function') {
+		throw new Error();
```
amhinson, line 146: Let's add a message to this error, similar to what is being done in web UI Components:
https://github.com/aws-amplify/amplify-js/blob/main/packages/amplify-ui-components/src/components/amplify-s3-album/amplify-s3-album.tsx#L51

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
cedricgrothues: TypeScript doesn't allow `?` in type variables.

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
```diff
@@ -0,0 +1,162 @@
...
+	});
+
+	const textStyle = Object.assign({}, StyleSheet.flatten(theme.text), style);
+
+	return <Text style={textStyle}>{error ? I18n.get(fallbackText) : text}</Text>;
+};
+
+interface IUseS3Text {
+	textKey?: string;
+	path?: string;
```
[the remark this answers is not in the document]
cedricgrothues: S3Image only has an `imgKey` and not both because I doesn't support uploading to S3. I only used `path` as a fallback when loading text from S3, if `textKey` is not defined.

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
```diff
@@ -0,0 +1,162 @@
...
+		textKey,
+		path,
+		level,
+		track,
+		identityId,
+		body,
+		contentType,
+	});
+
+	const textStyle = Object.assign({}, StyleSheet.flatten(theme.text), style);
```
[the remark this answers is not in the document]
cedricgrothues: I'd say the `style` prop might be important to stay consistent with React Native's "standard library" components.

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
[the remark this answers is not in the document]
amhinson: Ah ok, thanks for the clarification!

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
amhinson: Understandable! Although I think with these custom hooks, it isn't necessarily expected that they use the same structure as the base hooks since their usage can vary widely. For example, I could see a scenario where a `loading` value could be helpful here in the future. This is somewhat similar to other custom hooks that already exist in the wild, such as Apollo's [useQuery](https://www.apollographql.com/docs/react/api/react/hooks/#usequery) for instance.

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
amhinson: Ah ok good to know!

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
cedricgrothues: Good point, I'll update the code.
> For example, I could see a scenario where a loading value could be helpful here in the future.

Agreed, I could add this to this PR as well, if that's alright with you.

amhinson: Yea sure! I think that could be helpful.

packages/aws-amplify-react-native/src/AmplifyTheme.ts
```diff
@@ -158,4 +158,8 @@ export default StyleSheet.create({
 		textAlign: 'center',
 		padding: 20,
 	},
+	storageText: {
+		color: '#152939',
```
ashika01, line 162: How are we deciding on this color?!

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
```diff
@@ -0,0 +1,180 @@
...
+	);
+
+	return <Text style={textStyle}>{error ? I18n.get(fallbackText) : text}</Text>;
+};
+
+interface IUseS3TextProps {
+	textKey?: string;
+	path?: string;
+	level: AccessLevel;
+	track: boolean;
```
ashika01, line 58: I think track should be optional too

packages/aws-amplify-react-native/src/Storage/S3Text.tsx
```diff
@@ -0,0 +1,180 @@
...
+	text?: string;
+	loading: boolean;
+	error?: Error;
+}
+
+export const useS3Text = ({
+	textKey,
+	path,
+	level = AccessLevel.Public,
+	track = false,
```
ashika01, line 74: If track is made optional I dont think you need this default value

packages/aws-amplify-react-native/src/AmplifyTheme.ts
cedricgrothues: That's the color used by [`ui-components S3Text`](https://github.com/aws-amplify/amplify-js/blob/main/packages/amplify-ui-components/src/global/theme.ts#L40)

ashika01: Awesome. I was gonna suggest that.

cedricgrothues:
> Also, since this is would be the first implementation for aws-amplify-react-native using hooks, would you be able to verify any potential limitations of using this component or the library as a whole in previous version of React Native that don't support hooks (< 0.59)?

@amhinson @ashika01 – As far as I can tell, `react-native` < 0.59 does not support any iOS <= 13.

Last version that doesn't support hooks: [0.58.6](https://github.com/facebook/react-native/releases/tag/v0.58.6) (published Feb 28, 2019)
The issue tracking support for iOS 13:  https://github.com/facebook/react-native/issues/25181 (published Jun 6, 2019)

amhinson:
@cedricgrothues I have come across some apps that are still using 0.58 (albeit a small percentage). In general, we're just looking to know what someone using < 0.59 might encounter, i.e.:
- Will they see an error using `withAuthenticator` even if they don't use `S3Text`?
- Will an error only show if using `S3Text` directly?

cedricgrothues:
@amhinson – I just tested the `S3Text` component on Android with `react-native` < 0.59, and it turns out that importing the library, even without importing `S3Text`, causes an `undefined is not a function` error.

abdallahshaban557:
Hi @cedricgrothues - these React Native components for Storage are on their path to deprecation. I am closing this PR. Apologies for not getting to this sooner!

PR #6602, opened 2020-08-18
feat(aws-amplify-react-native): Update S3Album

kendevops:
_Description of changes:_
Updating the React Native S3Album UI Component to include more features 


packages/aws-amplify-react-native/src/Storage/S3Album.tsx
```diff
@@ -33,25 +47,76 @@ interface IS3AlbumState {
...
+			.catch(err => {
+				logger.debug('handle pick error', err);
+				if (onError) {
+					onError(err);
+				}
+			});
 	}
 
 	render() {
+		const { picker, level, identityId } = this.props;
```
eddiekeller, line 118: nit - looks like this `identityId` variable isn't being used anywhere in the render function. We can ignore destructuring it here.

packages/aws-amplify-react-native/src/Storage/S3Album.tsx
```diff
@@ -33,25 +47,76 @@ interface IS3AlbumState {
 export default class S3Album extends Component<IS3AlbumProps, IS3AlbumState> {
 	constructor(props: IS3AlbumProps) {
 		super(props);
-
 		this.state = { images: [] };
 	}
 
+	getKey(file) {
```
eddiekeller, line 53: Since we are expecting certain properties to be present on this `file` argument, let's add a typing here so we have some explicit type safety.

packages/aws-amplify-react-native/src/Storage/S3Album.tsx
```diff
@@ -33,25 +47,76 @@ interface IS3AlbumState {
...
+			} else {
+				key = encodeURI(JSON.stringify(fileToKey));
+			}
+			if (!key) {
+				logger.debug('key is empty');
+				key = 'empty_key';
+			}
+		}
+
+		return key.replace(/\s/g, '_');
```
eddiekeller, line 73: For my own knowledge - why is replacing the whitespace with an underscore the correct choice here?

packages/aws-amplify-react-native/src/Storage/S3Album.tsx
```diff
@@ -33,25 +47,76 @@ interface IS3AlbumState {
...
+				key = fileToKey;
+			} else if (callback_type === 'function') {
+				key = fileToKey({ name, size, type });
+			} else {
+				key = encodeURI(JSON.stringify(fileToKey));
+			}
+			if (!key) {
+				logger.debug('key is empty');
+				key = 'empty_key';
+			}
```
eddiekeller, line 70: Let's pull this check outside of the `if (fileToKey)` block. I don't know if `name` is a required property on the `file` argument, but for safety, we can catch that case too if we move it outside of the current scope it is in.

packages/aws-amplify-react-native/src/Storage/S3Album.tsx
```diff
@@ -33,25 +47,76 @@ interface IS3AlbumState {
...
+		let key = encodeURI(name);
+		if (fileToKey) {
+			const callback_type = typeof fileToKey;
+			if (callback_type === 'string') {
+				key = fileToKey;
+			} else if (callback_type === 'function') {
+				key = fileToKey({ name, size, type });
+			} else {
+				key = encodeURI(JSON.stringify(fileToKey));
+			}
```
eddiekeller, line 66: It feels to me like we can split this out into its own method. Then, on line 57, we could simply do something like:

```
const key = fileToKey ? this.getKeyFromFile(fileToKey) : encodeURI(name);
```

packages/aws-amplify-react-native/src/Storage/S3Album.tsx
```diff
@@ -33,25 +47,76 @@ interface IS3AlbumState {
...
+			level: level ? level : 'public',
+			track,
+			identityId,
+		})
 			.then(data => {
 				logger.debug(data);
 				if (filter) {
 					data = filter(data);
+				} else {
+					logger.debug('update an image');
```
eddiekeller, line 102: nit - I'm not totally sure what this message is trying to say. Can we make it more clear?

packages/aws-amplify-react-native/src/Storage/S3Album.tsx
```diff
@@ -33,25 +47,76 @@ interface IS3AlbumState {
...
+				}
+				if (onLoad) {
+					onLoad(data);
 				}
 				this.setState({ images: data });
 			})
-			.catch(err => logger.warn(err));
+			.catch(err => {
+				logger.debug('handle pick error', err);
```
eddiekeller, line 110: nit - let's keep this `logger.warn()` since we are logging a caught error

packages/aws-amplify-react-native/src/Storage/S3Album.tsx
```diff
@@ -33,25 +47,76 @@ interface IS3AlbumState {
...
+				if (onError) {
+					onError(err);
+				}
+			});
 	}
 
 	render() {
+		const { picker, level, identityId } = this.props;
+		const pickerTitle = this.props.pickerTitle || 'Pick';
```
eddiekeller, line 119: nit - syntax

```
const { pickerTitle = 'Pick' } = this.props;
```

Actually, we could also just add this to the destructuring on line 118

eddiekeller (changes requested):
Thanks for the submission! I have a few comments of things that we can update here.

abdallahshaban557:
Hi @kendevops - this React Native component is on its path to deprecation. Apologies for not getting to this sooner!

PR #6672, opened 2020-08-28
@feat(@aws-amplify/ui-components): Error/audio handling and css improvements

wlee221:
_Issue #, if available:_

_Description of changes:_  This PR makes improvements to chatbot that is on `ui-preview` right now.

Improvements:
- Errors are now divided into recoverable and unrecoverable errors and are handled accordingly. 
- Users no longer need to wait for bot to finish speaking. They can interrupt the audio by pressing mic or send button.
- Loading dots are now customizable. 
- Use different input placeholder when only voice is enabled.

Fix:
- Disable text sending through `enter` keystroke when only voice is enabled.

### DIFF
--- a/packages/amplify-ui-components/src/common/audio-control/recorder.ts
+++ b/packages/amplify-ui-components/src/common/audio-control/recorder.ts
@@ -23,6 +23,7 @@ export class AudioRecorder {
   private audioSupported: boolean;
 
   private analyserNode: AnalyserNode;
+  private playbackSource: AudioBufferSourceNode;
   private onSilence: SilenceHandler;
   private visualizer: Visualizer;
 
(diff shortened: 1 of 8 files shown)


packages/amplify-ui-components/src/common/audio-control/recorder.ts
```diff
@@ -144,17 +145,19 @@ export class AudioRecorder {
...
+          this.playbackSource.buffer = buf;
+          this.playbackSource.connect(this.audioContext.destination);
+          this.playbackSource.onended = () => {
             return res();
           };
-          playbackSource.start(0);
+          this.playbackSource.start(0);
         };
         const errorCallback = err => {
+          console.error('test');
```
ashika01, line 160: I think this needs to be removed?

wlee221: oh no i did it again

ashika01: Happens to all. No worries

PR #6708, opened 2020-09-02
fix: type for completeNewPasswordChallenge

donaldpipowitch:
_Issue #, if available:_ None, I'm aware of.

_Description of changes:_

I fixed the type for `completeNewPasswordChallenge`. It currently uses a custom type for the `callbacks` param. But we run into an issue where calling `completeNewPasswordChallenge` would result in an `Can't call selectMFAType. It's undefined.` error. I had a look into the source code and as you can  see `completeNewPasswordChallenge` might call `authenticateUserInternal` and just passes `callbacks` to that function. That one basically expects `callbacks` to be in shape of `IAuthenticationCallback`. Now I can properly specify a `selectMFAType` callback when I call `completeNewPasswordChallenge`.

See https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/src/CognitoUser.js#L632 and https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/src/CognitoUser.js#L448.


donaldpipowitch:
@elorzafe can you please review this? We hit the 2nd bug in production, because of this. Not just because we missed `selectMFAType`, but now also `totpRequired`. 😞

PR #6712, opened 2020-09-02
fix(@aws-amplify/datastore): DataStore regression with AsyncStorage

ericclemmons:
See commits for further details.

Introduced by https://github.com/aws-amplify/amplify-js/pull/6146

### DIFF
--- a/packages/aws-amplify/src/withSSRContext.ts
+++ b/packages/aws-amplify/src/withSSRContext.ts
@@ -3,8 +3,6 @@ import { Auth } from '@aws-amplify/auth';
 import { Credentials } from '@aws-amplify/core';
 import { AmplifyClass, UniversalStorage } from '@aws-amplify/core';
 
-import { DataStore } from '@aws-amplify/datastore';
-
 // ! We have to use this exact reference, since it gets mutated with Amplify.Auth
 import { Amplify } from './index';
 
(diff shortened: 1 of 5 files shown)


packages/aws-amplify/src/withSSRContext.ts
```diff
@@ -3,8 +3,6 @@ import { Auth } from '@aws-amplify/auth';
 import { Credentials } from '@aws-amplify/core';
 import { AmplifyClass, UniversalStorage } from '@aws-amplify/core';
 
-import { DataStore } from '@aws-amplify/datastore';
-
```
ericclemmons, line 7: Removed until DataStore subscription limit is addressed.

packages/datastore/src/storage/adapter/AsyncStorageDatabase.ts
```diff
@@ -21,7 +21,7 @@ class AsyncStorageDatabase {
 	 */
 	private _collectionInMemoryIndex = new Map<string, Map<string, string>>();
 
-	private storage = new InMemoryStore();
+	private storage = createInMemoryStore();
```
ericclemmons, line 24: Fix so that `AsyncStorage` is returned as a singleton while JS gets new instances.

elorzafe (approved):
Thanks @ericclemmons !! 👍

amhinson (approved):

PR #6726, opened 2020-09-04
feat(@aws-amplify/datastore): Add SSR support for DataStore

ericclemmons:
_Issue #, if available:_ Based on #6146, this solves the security concerns with DataStore & `PubSub.subscribe` in Node environments.

- [x] `PubSub.subscribe` throws in Node.  This behavior wasn't explicitly supported and shouldn't be for security reasons. (Multi-tenant apps may share sensitive information across a single, accessible socket)

- [x] `DataStore` doesn't start the sync engine in Node.

- [x] `withSSRContext` includes `DataStore` as a default module (along with `API` and `Auth`).


sammartinez (approved):
Soft approval as I want weigh in from @manueliglesias and @elorzafe on this before we merge

elorzafe (approved):
Nice work @ericclemmons 🌮 🎉 🥇 !

sammartinez (changes requested):
Lets add the unit tests needed to resolve the unit test failure and Ill be happy to do a full approval

packages/datastore/src/sync/index.ts
```diff
@@ -162,31 +163,39 @@ export class SyncEngine {
 								},
 							});
 
-							//#region GraphQL Subscriptions
-							const [
-								ctlSubsObservable,
-								dataSubsObservable,
-							] = this.subscriptionsProcessor.start();
+							let ctlSubsObservable, dataSubsObservable;
```
manueliglesias, line 166: Can we type-annotate these?

manueliglesias (approved):
Minor comment, LGTM! 🎉

packages/datastore/src/sync/index.ts
ericclemmons: @manueliglesias Sure thing.  `this.subscriptionsProcessor.start()` returns these types:

```
const ctlObservable: Observable<CONTROL_MSG>
const dataObservable: Observable<[TransformerMutationType, SchemaModel, Readonly<{
    id: string;
} & Record<string, any>>]>
```

I'll add those, or use a `ReturnType` utility of `start()` to cast those unless you have a better solution!

packages/pubsub/src/PubSub.ts
```diff
@@ -151,6 +153,12 @@ export class PubSubClass {
 		topics: string[] | string,
 		options?: ProvidertOptions
 	): Observable<any> {
+		if (isNode) {
+			throw new Error(
+				'Subscriptions are not supported in Node to prevent memory leaks.'
```
ericclemmons, line 158: I'm leaning towards the following @sammartinez :

```suggestion
				'Subscriptions are not supported in Node.'
```

packages/datastore/src/sync/index.ts
manueliglesias: `Readonly<{id: string;} & Record<string, any>>` should be the same as `PersistentModel`, so:

```
let ctlSubsObservable: Observable<CONTROL_MSG>;
let dataSubsObservable: Observable<[
	TransformerMutationType,
	SchemaModel,
	PersistentModel
]>;
```

ericclemmons: I'll update. See my latest commits for what I re-used.

sammartinez (approved):
LGTM 🌮

packages/datastore/src/sync/index.ts
ericclemmons: Fixed! Pushing now.  Funny how TypeScript's annotations showed me the definition rather than `PersistentModel`.

adamup928:
I see this change now prevents subscriptions from being established in a Node environment, and I understand the justification above regarding multi-tenant apps sharing sockets.  In our implementation, however, we are using the PubSub module in a client application running in a desktop environment and built with Electron.  This pull request breaks the functionality of the application entirely, and we can not use the latest update.

Is there a recommendation for this scenario?  Could we introduce a mode that allows subscriptions to be made when explicitly set?

ericclemmons:
@adamup928 Oh no! Can you open a [new issue as a feature request](https://github.com/aws-amplify/amplify-js/issues/new?assignees=&labels=feature-request&template=feature_request.md&title=)?

You're absolutely right: the restriction in Node **does not apply** to Electron apps, so we can do better detection to unblock you.

Let's discuss more in that issue, but here are 2 solutions I'm aware of:

- Remove the `isNode` check from this PR in favor of a `if (config.ssr)`. In this scenario, we're blocking those connections in SSR scenarios specifically.  If Amplify is being used in an Electron app, SSR wouldn't be applicable. (Unless you say otherwise :D)
- Expand our `browserOrNode` utility to have other checks like `isElectron`, `isReactNative` so we're less generic.

adamup928:
Thanks, @ericclemmons.  Done!  Here it is: https://github.com/aws-amplify/amplify-js/issues/6873

Issue #1313, opened 2018-07-26
API client init arguments not optional in TypeScript typings


LosD:
> Do you want to request a feature or report a bug?

Bug

> What is the current behavior?

When using `API.get`/`API.put`/`API.del`/`API.head` with TypeScript, the current typings does not specify 
that the `init` argument it optional, even though the documentation says otherwise.

> If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than AWS Amplify.

1. make an app in TypeScript.
2. Try to use the API client
😄 

Joking aside, just check API.d.ts (in the packaged aws-amplify):
```typescript
    /**
     * Make a GET request
     * @param {string} apiName  - The api name of the request
     * @param {string} path - The path of the request
     * @param {json} [init] - Request extra params
     * @return {Promise} - A promise that resolves to an object with response status and JSON data, if successful.
     */
    get(apiName: any, path: any, init: any): Promise<any>;
```
`init: any` should be `init?: any`.

(furthermore, arguments should really be typed instead of any. At least according to the documentation, `apiName` and `path` are always strings and the possible init fields are fixed... But I guess that's more of a feature request than a bug. It's terrible use of TypeScript, but doesn't really stop anyone)

> What is the expected behavior?

init argument is optional.

> Which versions of Amplify, and which browser / OS are affected by this issue? Did this work in previous versions?

Amplify 1.0.2 (browser/OS irrelevant). It has been like this at least since 1.0.0.

nadetastic:
Related to #9429

PR #6732, opened 2020-09-05
Add optionality to init type for api get op

harrysolovay:
_Issue #, if available:_ #1313

_Description of changes:_ add optionality to `init` param for `get` operation in API category.

### DIFF
--- a/packages/api/src/API.ts
+++ b/packages/api/src/API.ts
@@ -87,7 +87,7 @@ export class APIClass {
 	 * @param {json} [init] - Request extra params
 	 * @return {Promise} - A promise that resolves to an object with response status and JSON data, if successful.
 	 */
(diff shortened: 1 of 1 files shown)


packages/api/src/API.ts
```diff
@@ -87,7 +87,7 @@ export class APIClass {
 	 * @param {json} [init] - Request extra params
 	 * @return {Promise} - A promise that resolves to an object with response status and JSON data, if successful.
 	 */
-	get(apiName, path, init): Promise<any> {
+	get(apiName, path, init?: any): Promise<any> {
```
sammartinez, line 90: Do we know the shape? Or can we have it where the optional doesn't have a type? I'm not sure that `any` is right but I could be wrong here. Also wouldn't we want to do this for the remainder CRUD operations?

PR #6757, opened 2020-09-09
chore: pin @types/babel__traverse to allow ui-angular to build

iartemiev:
Upstream dependency @types/babel__traverse was updated to require TypeScript@3.4.0. Since ui-angular is still on 3.2, the build script for that package is failing.

This change forces packages to resolve @types/babel__traverse: 7.0.8, which supports TS@3.2

More of a band-aid fix to unblock us. We should look into upgrading ui-angular to more recent versions of TS, Angular, and Angular Compiler

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
 		"winston": "^3.2.1"
 	},
 	"resolutions": {
+		"@types/babel__traverse": "7.0.8",
 		"terser": "4.6.7",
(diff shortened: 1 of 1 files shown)


wlee221 (approved):
Thanks @iartemiev, lgtm 👍  We have inconsistent typescript version throught out packages... we should keep them consistent and updated :D

Issue #6327, opened 2020-07-14
Custom auth lambda trigger is not configured for the user pool -> on Sign Up Confirmation


dbtorrance:
**Describe the bug**
Using the @aws-amplify/ui-react AmplifyAuthenticator and AmplifySignUp components to generate a custom Sign-Up page works as expected. However, when a user attempts to confirm sign-up by supplying the correct code, we receive the "Custom auth lambda trigger is not configured for the user pool" error, and the Application does not automatically sign-in the user. The user's confirmation is processed correctly, and can sign-in as expected, but the error is confusing.

**To Reproduce**

1. Generate a simple react app (create-react-app ...) for Amplify
2. Add Auth to the Amplify App (amplify add auth)
3. Create a conditional app rendering setup for the application
4. Sign Up a New user
5. Once the new user attempts to confirm the sign-up, you should receive the error in the Title.

**Expected behavior**
The error should not occur.

**Code Snippet**
Please provide a code snippet or a link to sample code of the issue you are experiencing to help us reproduce the issue. (Be sure to remove any sensitive data)

AppWithAuth.js
```
import React from "react";
import App from "./App";
import Amplify from 'aws-amplify';
import { AmplifyAuthenticator, AmplifySignUp, AmplifySignIn } from "@aws-amplify/ui-react";
import { AuthState, onAuthUIStateChange } from '@aws-amplify/ui-components';
import awsconfig from './aws-exports';

Amplify.configure(awsconfig);

const AppWithAuth = () => {
    const [authState, setAuthState] = React.useState();
    const [user, setUser] = React.useState();

    React.useEffect(() => {
        return onAuthUIStateChange((nextAuthState, authData) => {
            setAuthState(nextAuthState);
            setUser(authData)
        });
    }, []);

  return authState === AuthState.SignedIn && user ? (
      <App />
    ) : (
        <AmplifyAuthenticator usernameAlias="email">
            <AmplifySignUp
                slot="sign-up"
                usernameAlias="email"
                formFields={[
                {
                    type: "email",
                    label: "Email Address",
                    required: true,
                },
                {
                    type: "password",
                    label: "Password",
                    placeholder: "********************",
                    required: true,
                },
                {
                    type: "given_name",
                    label: "First Name",
                    required: false,
                },
                {
                    type: "family_name",
                    label: "Last Name",
                    required: false,
                },
                {
                    type: "custom:team",
                    label: "Team",
                    required: true,
                },
                ]} 
            />
[...]

PR #6783, opened 2020-09-14
fix(@aws-amplify/ui-components): Confirm sign up fix

jordanranz:
_Issue #, if available:_
https://github.com/aws-amplify/amplify-js/issues/6327

_Description of changes:_
- Confirm sign up was not receiving the proper sign up attributes from a successful sign up action. This PR ensures that a valid copy of sign up attributes is passed to confirm sign up.
- If no password is sent or the password is empty, fallback to the sign in screen instead of auto signing in.


packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
```diff
@@ -140,7 +140,8 @@ export class AmplifySignUp {
       if (data.userConfirmed) {
         await handleSignIn(this.signUpAttributes.username, this.signUpAttributes.password, this.handleAuthStateChange);
       } else {
-        this.handleAuthStateChange(AuthState.ConfirmSignUp, { ...data.user, signUpAttrs: this.signUpAttributes });
+        const signUpAttrs = JSON.parse(JSON.stringify(this.signUpAttributes));
```
ashika01, line 143: Just curious on why we need to stringify then parse JSON here.

jordanranz: This is essentially what fixes the clone that was improperly happening before. I am going to quickly investigate using spread for the `signUpAttrs` as this would be a better solution.

jordanranz: Updated to use spread :D

ashika01 (approved):
LGTM 👍

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
ashika01: Yeah.. spread is ok for shallow copy but `JSON.parse` for deep copy.

ericclemmons (approved):
Nice, easy fix for this!

I agree with @ashika01  that `JSON.parse(JSON.stringify(this.signUpAttibutes))` may be desirable for a deep-clone, but I don't recall this object being more than 1 level deep (so `...` would cover our use-case).

Thanks for this @jordanranz !

sammartinez (approved):
LGTM 🌮

wlee221:
LGTM, thanks @jordanranz

PR #6785, opened 2020-09-14
feat(@aws-amplify/datastore): add query sorting

iartemiev:
* Enables passing a sort predicate to `DataStore.query`

E.g.,
```typescript
import { DataStore, Predicates, SortDirection } from '@aws-amplify/datastore';

await DataStore.query(User, Predicates.ALL, {
  page: 0,
  limit: 50,
  sort: (s) => s.lastName(SortDirection.ASCENDING),
});
```
* The sort predicate is chainable, allowing for multi-field sort: `s.lastName('ASCENDING').age('DESCENDING')`
* Full TypeScript Intellisense support, like with the query predicate

### DIFF
--- a/packages/datastore/__tests__/AsyncStorage.ts
+++ b/packages/datastore/__tests__/AsyncStorage.ts
@@ -17,6 +17,7 @@ import {
 	Person as PersonType,
 } from './model';
 import { newSchema } from './schema';
+import { SortDirection } from '../src/types';
 
 let AsyncStorageAdapter: typeof AsyncStorageAdapterType;
 
@@ -312,7 +313,7 @@ describe('AsyncStorage tests', () => {
 		const sortedPersons = await DataStore.query(Person, null, {
 			page: 0,
 			limit: 20,
-			sort: s => s.firstName('DESCENDING'),
+			sort: s => s.firstName(SortDirection.DESCENDING),
 		});
 
 		expect(sortedPersons[0].firstName).toEqual('Meow Meow');
@@ -351,9 +352,9 @@ describe('AsyncStorage tests', () => {
 				limit: 20,
 				sort: s =>
 					s
-						.firstName('ASCENDING')
-						.lastName('ASCENDING')
-						.username('ASCENDING'),
+						.firstName(SortDirection.ASCENDING)
+						.lastName(SortDirection.ASCENDING)
+						.username(SortDirection.ASCENDING),
 			}
 		);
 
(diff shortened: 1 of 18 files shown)


renebrandel:
image

amhinson (approved):
Looks great! 👍

packages/datastore/__tests__/AsyncStorage.ts
```diff
@@ -279,6 +283,85 @@ describe('AsyncStorage tests', () => {
...
+
+		await DataStore.save(p1);
+		await DataStore.save(p2);
+		await DataStore.save(p3);
+		await DataStore.save(p4);
+
+		const sortedPersons = await DataStore.query(Person, null, {
+			page: 0,
+			limit: 20,
+			sort: s => s.firstName('DESCENDING'),
```
manueliglesias, line 315: Let's use the exported enum for SortDirection. If not here, at least in the docs

packages/datastore/package.json
```diff
@@ -18,8 +18,8 @@
 		"build:cjs": "node ./build es5 && webpack && webpack --config ./webpack.config.dev.js",
 		"build:esm": "node ./build es6",
 		"build:cjs:watch": "node ./build es5 --watch",
-		"build:esm:watch": "node ./build es6 --watch",
-		"build": "npm run clean && npm run build:esm && npm run build:cjs",
+		"build:esm:watch": "rimraf lib-esm && node ./build es6 --watch",
```
manueliglesias, line 21: Question, is this to match what other packages in the monorepo do?

ericclemmons, line 21: What behavior differences did you see with this vs. `yearn clean && ...`?

manueliglesias (approved):
LGTM 🎉 

Just left a very minor non-blocking comment and a question

packages/datastore/__tests__/AsyncStorage.ts
iartemiev: Sounds good! I've updated the PR description. Will use that in the docs as well

packages/datastore/package.json
iartemiev: No, I threw that in because `build:esm:watch` doesn't work anymore without that. It tries to re-transpile already transpiled files in lib-esm and throws an error. This only affects running :watch, so only for team members and contributors.

iartemiev: For this line specifically, since `build:esm:watch` only builds the es6 esm modules, I didn't want to `yarn clean ` to blow away `lib` and `dist` just limiting the effects to what the script interacts with

ericclemmons: Makes perfect sense.  Great call on this & the `lib-esm` errors: I've experienced that as well!

packages/datastore/__tests__/AsyncStorage.ts
iartemiev: Updated the unit tests as well

sammartinez (approved):
LGTM 🌮

PR #6790, opened 2020-09-15
fix(aws-amplify-react-native): auto-fill the user name field on ConfirmSignUp page

### LINKED ISSUE #6789: [aws-amplify-react-native] email field on ConfirmSignUp is not auto-filled after redirecting from SignUp page
**Describe the bug**
aws-amplify-react-native v4.2.5

After signed up, the auth flow goes to ConfirmSignUp correctly, however, the email field is not been auto filled

**To Reproduce**

1. Create an app with <Authenticator> (`usernameAttributes="email"`)
1. Go to 'SignUp'
1. fill out email and password and sign up
1. network request returns 200 and page redirected to ConfirmSignUp
1. email field shows up, but it is empty

**Expected behavior**
The email field should auto filled

**Code Snippet**
The problem is here:
https://github.com/aws-amplify/amplify-js/blob/6abbf5053978420ef008fc45968a54d0762943de/packages/aws-amplify-react-native/src/Auth/ConfirmSignUp.tsx#L75

It should consider the `usernameAttributes` cases.

Because the value been filled is coming from this logic:
https://github.com/aws-amplify/amplify-js/blob/6abbf5053978420ef008fc45968a54d0762943de/packages/aws-amplify-react-native/src/Auth/AuthPiece.tsx#L69

It is checking 3 cases, but the `ConfirmSignUp.getDerivedStateFromProps` always returns `username` which breaks it,

**Screenshots**
After changing that line, the bug is fixed
image

image

Albert-Gao:
close #6789

### DIFF
--- a/packages/aws-amplify-react-native/src/Auth/ConfirmSignUp.tsx
+++ b/packages/aws-amplify-react-native/src/Auth/ConfirmSignUp.tsx
@@ -76,8 +76,8 @@ export default class ConfirmSignUp extends AuthPiece<
 			if (props.usernameAttributes === 'email') {
 				return { email: username };	
 			} else if (props.usernameAttributes === 'phone_number') {
-			   	return { phone_number: username };
-		   	}
+				return { phone_number: username };
+			}
 			
 			return { username };
 		}
--- a/packages/aws-amplify-react-native/src/Auth/ConfirmSignUp.tsx
+++ b/packages/aws-amplify-react-native/src/Auth/ConfirmSignUp.tsx
@@ -73,6 +73,12 @@ export default class ConfirmSignUp extends AuthPiece<
 		const username = props.authData;
 
 		if (username && !state.username) {
+			if (props.usernameAttributes === 'email') {
+				return { email: username };	
+			} else if (props.usernameAttributes === 'phone_number') {
+			   	return { phone_number: username };
+		   	}
+			
 			return { username };
 		}
 


Albert-Gao:
The CI error is irrelevant to this PR, right?

image

Albert-Gao:
So, this logic:

```javascript
if (username && !state.username) {
			if (props.usernameAttributes === 'email') {
				return { email: username };	
			} else if (props.usernameAttributes === 'phone_number') {
				return { phone_number: username };
			}
         
         return { username };
}
```

can be simplified to just this:

```javascript
if (username && !state.username) {
		  return { [props.usernameAttributes]: username }
}
```

But just wonder if it introduces some implicitly and potential for some future bugs where we change the convention in `AuthPiece`. So I went with the vanilla nasty nested `if else`, seems more clearer

sammartinez:
> The CI error is irrelevant to this PR, right?
> 
> image

You are correct that this is related to that. Can you look to resolve this?

sammartinez:
> So, this logic:
> 
> ```js
> 			if (props.usernameAttributes === 'email') {
> 				return { email: username };	
> 			} else if (props.usernameAttributes === 'phone_number') {
> 				return { phone_number: username };
> 			}
>          
>          return { username };
> ```
> 
> can be simplified to just this:
> 
> ```js
> 		  return { [props.usernameAttributes]: username }
> ```
> 
> But just wonder if it introduces some simplicity and potential for some future bugs where we change the convention in `AuthPiece`. So I went with the vanilla nasty nested `if else`, seems more clearer

I do like the second option and seems to be more clearer to myself and others on the team. Can you look to update this?

Albert-Gao:
Thanks, changes been made, and new PR here: https://github.com/aws-amplify/amplify-js/pull/7047

PR #6795, opened 2020-09-15
Upgrade Cypress to 5.2.0

amhinson:
_Description of changes:_
- Upgrade Cypress to 5.2.0
- Remove usage of `ui-components/master` from samples along with the `integ_setup_ui` job and other unneeded commands since all integ tests can now be run from `master`.

Successful workflow: https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/6143/workflows/1951b1ad-9f06-4267-abb7-80f8b8cf4e85

---


ericclemmons (approved):

amhinson:
Code coverage report seems inaccurate, FYI. Not sure why it is going down since this is only change the CircleCI config file (and removing more lines than it adds)

iartemiev (approved):
LGTM!

.circleci/config.yml
```diff
@@ -13,7 +13,7 @@ executors:
 
   js-test-executor:
     docker:
-      - image: cypress/included:4.12.1
+      - image: cypress/included:5.2.0
```
jordanranz, line 16: 👍

.circleci/config.yml
```diff
@@ -567,87 +528,69 @@ jobs:
...
           sample_name: amplify-authenticator
           spec: custom-authenticator
-      - integ_test_ui:
+      - integ_test_js:
           test_name: 'React Custom Authenticator'
           framework: react
           category: auth
           sample_name: with-authenticator
-          spec: amplify-authenticator
+          spec: ui-amplify-authenticator
```
jordanranz, line 560: We can migrate these to [`integ_react_auth`](https://github.com/aws-amplify/amplify-js/blob/main/.circleci/config.yml#L366) and do the same for the other frameworks. This will also allow us to pass the browser parameter [like so](https://github.com/aws-amplify/amplify-js/blob/main/.circleci/config.yml#L381)

amhinson: It actually looks like those tests are already being run in `integ_react_auth`, right? If so, we could just remove this section entirely.

amhinson: For React, that is. We can still add them for Angular and Vue.

.circleci/config.yml
```diff
@@ -567,87 +528,69 @@ jobs:
...
           category: auth
           sample_name: amplify-authenticator
-      - integ_test_ui:
+          spec: ui-amplify-authenticator
+      - integ_test_js:
           test_name: 'Angular Custom Authenticator'
           framework: angular
           category: auth
-          sample_name: amplify-authenticator
+          sample_name: ui-amplify-authenticator
```
jordanranz, line 578: This sample name is incorrect and can be reverted. Curious how this succeeded local testing.

jordanranz (approved):
LGTM

PR #6819, opened 2020-09-17
chore(@aws-amplify/analytics): Remove lodash from types

### LINKED ISSUE #5735: Problem with dependencies in analytics package.
**Describe the bug**
Problem with dependencies in analytics package. 

**To Reproduce**
File [AmazonPersonalizeProvider.ts](https://github.com/aws-amplify/amplify-js/blob/master/packages/analytics/src/Providers/AmazonPersonalizeProvider.ts) has got this code: 

```import { isEmpty, isEqual, get } from 'lodash';```

There are 2 problems with it: 

1. lodash is not specified as a dependency in [package.json](https://github.com/aws-amplify/amplify-js/blob/master/packages/analytics/package.json) of analytics project which makes our lambda functions break unless we add lodash as dependency.  
2. This code imports the whole lodash module instead of importing individual functions which significantly increases bundle size in the frontend. 

**Expected behavior**
```
import isEmpty from 'lodash/isEmpty';
...
```

I would check similar imports in other files as well. Please let me know if you want me to create a PR for this. 
Cheers

### PR DESCRIPTION
**Pull request by ericclemmons**
_Issue #, if available:_ Resolves #5735

#6131 fixes #5735, but this PR removes a previous `lodash` with tslint:
> https://github.com/aws-amplify/amplify-js/issues/5735#issuecomment-626315384

### CODE REVIEW

File: packages/analytics/tsconfig.json [hunk 1]
```diff
@@ -1,35 +1,27 @@
...
-        "types": ["node", "lodash"],
-        "esModuleInterop": true
-    },
-    "include": [
-        "src/**/*"
-    ],
-    "exclude": [
-        "src/setupTests.ts"
-    ]
-}
```
ericclemmons, line 35: I would prefer that our `.prettierrc` correctly formatted `.json` to be consistent with `JSON.stringify(obj, null, 2)` (which is what NPM & yarn use), but I didn't want to bother with that here.

iartemiev (approved):
👍🏻

PR #6837, opened 2020-09-20
Adding default and break to switch

nicholasgriffintn:
Description of changes:

Fixing ESLint errors that occur from the new UniversalStorage file due to the case not having a break clause and their being no default, quite simple.


ashika01:
@nicholasgriffintn I dont think we have any lint rules that complains about this. It might be your private settings thats resulting in this error.

nicholasgriffintn:
Well the linter is not the issue, it's just how I noticed it lol

The issue is that without breaks the switch just continues to the next case as it isn't returning and default provides a return for if no case is matched, not only best practice but how switch should be used tbh.

ashika01:
@nicholasgriffintn Yup.. My worries are on the same point too. I know thats the best practice, but if we add the break, it might break something in the auth flow. I am not sure about. We are working on some major auth stuff. We can make note of this to clean up and test this properly then.

nicholasgriffintn:
Looking forward to the major Auth stuff!

sammartinez:
@nicholasgriffintn Thank you for opening the pull request. As @ashika01 has called out, this has potential of breaking existing customers so we are unable to except this pull request without more test coverage around this. Also, we are unable to see any linting errors around this. Again, thank you for your time on this. Resolving

PR #6844, opened 2020-09-21
fix(aws-amplify): Amplify.register is called per-instance, not per-import

### LINKED ISSUE #6761: AmplifyClass not getting configured on production build
**Describe the bug**
I'm building an angular app and when I run the production build, amplify modules doesn't work.

image

**Expected behavior**
It's suppose to configure itself and run like it works on "ng serve"

### LINKED ISSUE #6836: No Auth module registered in Amplify after Fix #6811
**Describe the bug**
Fix #6811 has not fixed "No Auth module registered in Amplify" error 

**To Reproduce**

1. Deleted `node_modules` folder and `package-lock.json` file
2. Ran `npm install`
3. Ran `npm start`
4. No data received from GraphQL endpoints and error in console

**Expected behavior**
Data should be received from endpoints 

**Code Snippet**
App.js
```
import React, { Component, lazy, Suspense } from 'react';
import './App.css';

import Amplify from 'aws-amplify';

Amplify.configure(awsconfig);
...
```

.aws-exports.js
```
/* eslint-disable */
// WARNING: DO NOT EDIT. This file is automatically generated by AWS Amplify. It will be overwritten.

const awsmobile = {
    "aws_project_region": "eu-west-1",
    "aws_cognito_identity_pool_id": "eu-west-1:xxxx",
    "aws_cognito_region": "eu-west-1",
    "aws_user_pools_id": "eu-west-1_xxxx",
    "aws_user_pools_web_client_id": "xxxx",
    "oauth": {},
    "aws_appsync_graphqlEndpoint": "https://xxxx.appsync-api.eu-west-1.amazonaws.com/graphql",
    "aws_appsync_region": "eu-west-1",
    "aws_appsync_authenticationType": "AMAZON_COGNITO_USER_POOLS"
};

export default awsmobile;
```

SignIn Function
```
import { Auth } from 'aws-amplify';

export default function (payload) {
  return Auth.signIn(payload.inputUsername.toLowerCase(), payload.inputPassword);
}
```

getProjects.js (API call once logged in to populate home page)
```
import { API, graphqlOperation } from 'aws-amplify';
import * as Queries from '../../graphql/custom-queries';

export default function(username) {
  const query = {
    cognitoID: username
  }
  return API.graphql(graphqlOperation(Queries.getProjectsForHomePage, query));
}
```

custom-queries.js
```
export const getProjectsForHomePage = /* GraphQL */ `
  query GetProjectsForHomePage($cognitoID: String!) {
    userByCognitoID(cognitoID: $cognitoID) {
      items {
        projects {
          items {
            project {
              id
              projectName
              projectDescription
              projectStatus
              projectAvatar
              createdAt
              teamMembers {
                items {
                  isAdministrator
                  attendee {
                    id
                    accreditations {
                      items {
                        expiryDate
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
`;
```

package.json
```
  "dependencies": {
    "@aws-amplify/auth": "^3.4.4",
    "@aws-amplify/core": "^3.5.4",
    "@aws-amplify/pubsub": "^2.1.9",
    "aws-amplify": "^3.3.1",
    "aws-amplify-react": "^4.2.5",
    ...
  },
```

**Screenshots**

Console before user logs in
[Image](https://imgur.com/xWYa19p)

Console after user logs in and API calls are made
[Image](https://imgur.com/bmrpV3q)

**What is Configured?**
Configured Amplify Parts

- Auth
- API
- Functions x 4

All configured through Amplify CLI

**Additional context**
The problem occurred after reinstalling `node_modules` folder.  I have another Mac that has the same project installed and that is still working after pulling the latest commit from Git.

```
@aws-amplify/api: v3.2.4
@aws-amplify/api-graphql: v1.2.4
@aws-amplify/auth: v3.4.4
@aws-amplify/core: v3.5.4
```

ericclemmons:
_Issue #, if available:_ Fixes #6761, Fixes #6836 

Introduced by #6146, `Amplify.register` was moved to per-import (so that `Amplify.Auth` was always the same singleton) rather than per-instance (so that each `req` creating `new Auth` on the server wouldn't create memory leaks for the lifetime of the server!).

`withSSRContext` explicitly registers modules to a _new_ `amplify` instance:

https://github.com/aws-amplify/amplify-js/blob/26ce5dfb0270009fc10f003f5627046ddaf5ae4e/packages/aws-amplify/src/withSSRContext.ts#L34-L40

This PR reverts:

1. 167c787990727093bc1cfb53efa80fb57f9f0467
2. (Partially) fd830eb2598476b0d54b97990533472c0da110ca

For validation, I ran **production** builds of Next.js, CRA, & Angular (https://github.com/aws-amplify/amplify-js-samples-staging/pull/166).

---

### DIFF
--- a/packages/pushnotification/__tests__/PushNotification-test.ts
+++ b/packages/pushnotification/__tests__/PushNotification-test.ts
@@ -73,7 +73,7 @@ describe('PushNotification:', () => {
 			// Spy should be at 0 (it was already called on import)
 			expect(registerSpy).toHaveBeenCalledTimes(0);
 
-			// Global Amplify will always reference last registered intance
+			// Global Amplify will always reference last registered instance
 			const NewPushNotification = new PushNotification(null);
 			expect(Amplify.Pushnotification).toEqual(NewPushNotification);
 
--- a/packages/pushnotification/__tests__/PushNotification-test.ts
+++ b/packages/pushnotification/__tests__/PushNotification-test.ts
@@ -1,3 +1,14 @@
+jest.mock('@aws-amplify/core', () => ({
+	__esModule: true,
+	...jest.requireActual('@aws-amplify/core'),
+	browserOrNode() {
+		return {
+			isBrowser: true,
+			isNode: false,
+		};
+	},
+}));
+
 import { DeviceEventEmitter, Platform, NativeModules } from 'react-native';
 import Amplify from 'aws-amplify';
 import PushNotificationIOS from '@react-native-community/push-notification-ios';
--- a/packages/pushnotification/__tests__/PushNotification-test.ts
+++ b/packages/pushnotification/__tests__/PushNotification-test.ts
@@ -73,12 +73,12 @@ describe('PushNotification:', () => {
 			// Spy should be at 0 (it was already called on import)
 			expect(registerSpy).toHaveBeenCalledTimes(0);
 
-			// Global Amplify should keep original instance, not new instances
+			// Global Amplify will always reference last registered intance
 			const NewPushNotification = new PushNotification(null);
-			expect(Amplify.Pushnotification).not.toEqual(NewPushNotification);
+			expect(Amplify.Pushnotification).toEqual(NewPushNotification);
 
-			// Amplify.register should not have been called for the new instance
-			expect(registerSpy).toHaveBeenCalledTimes(0);
+			// Amplify.register should be called for the new instance
+			expect(registerSpy).toHaveBeenCalledTimes(1);
 			registerSpy.mockClear();
 		});
 	});
--- a/packages/pushnotification/src/PushNotification.ts
+++ b/packages/pushnotification/src/PushNotification.ts
@@ -19,7 +19,12 @@ import {
 	AppState,
 } from 'react-native';
 import PushNotificationIOS from '@react-native-community/push-notification-ios';
-import { Amplify, ConsoleLogger as Logger, JS } from '@aws-amplify/core';
+import {
+	Amplify,
+	browserOrNode,
+	ConsoleLogger as Logger,
+	isEmpty,
+} from '@aws-amplify/core';
 
 const logger = new Logger('Notification');
 
@@ -57,7 +62,10 @@ export default class PushNotification {
 
 		this._notificationOpenedHandlers = [];
 
-		Amplify.register(this);
+		// Register module each time on the client, but not on the server to prevent memory leaks
+		if (browserOrNode().isBrowser) {
+			Amplify.register(this);
+		}
 	}
 
 	getModuleName() {
@@ -65,7 +73,7 @@ export default class PushNotification {
 	}
 
 	configure(config) {
-		if (JS.isEmpty(config)) return this._config;
+		if (isEmpty(config)) return this._config;
 		let conf = config ? config.PushNotification || config : {};
 
 		if (config['aws_mobile_analytics_app_id']) {
--- a/packages/pushnotification/src/PushNotification.ts
+++ b/packages/pushnotification/src/PushNotification.ts
@@ -62,7 +62,10 @@ export default class PushNotification {
 
 		this._notificationOpenedHandlers = [];
 
-		Amplify.register(this);
+		// Register module each time on the client, but not on the server to prevent memory leaks
+		if (!browserOrNode().isNode) {
+			Amplify.register(this);
+		}
 	}
 
 	getModuleName() {
--- a/packages/pushnotification/src/PushNotification.ts
+++ b/packages/pushnotification/src/PushNotification.ts
@@ -62,10 +62,7 @@ export default class PushNotification {
 
 		this._notificationOpenedHandlers = [];
 
-		// Register module each time on the client, but not on the server to prevent memory leaks
-		if (browserOrNode().isBrowser) {
-			Amplify.register(this);
-		}
+		Amplify.register(this);
 	}
 
 	getModuleName() {
--- a/packages/pushnotification/src/PushNotification.ts
+++ b/packages/pushnotification/src/PushNotification.ts
@@ -56,6 +56,8 @@ export default class PushNotification {
 		this._iosInitialized = false;
 
 		this._notificationOpenedHandlers = [];
+
+		Amplify.register(this);
 	}
 
 	getModuleName() {
--- a/packages/pushnotification/src/index.ts
+++ b/packages/pushnotification/src/index.ts
@@ -10,11 +10,9 @@
  * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
  * and limitations under the License.
  */
-import { Amplify } from '@aws-amplify/core';
 import NotificationClass from './PushNotification';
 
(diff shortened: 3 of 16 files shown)


packages/pushnotification/src/PushNotification.ts
```diff
@@ -56,14 +61,19 @@ export default class PushNotification {
 		this._iosInitialized = false;
 
 		this._notificationOpenedHandlers = [];
+
+		// Register module each time on the client, but not on the server to prevent memory leaks
+		if (browserOrNode().isBrowser) {
+			Amplify.register(this);
+		}
```
ericclemmons, line 68: @amhinson Since this class is iOS-specific, I think I should remove this.  Do you know what the values of `browserOrNode()` are off-hand in this environment?

packages/pushnotification/src/PushNotification.ts
```diff
@@ -56,14 +61,19 @@ export default class PushNotification {
...
 	}
 
 	getModuleName() {
 		return 'Pushnotification';
 	}
 
 	configure(config) {
-		if (JS.isEmpty(config)) return this._config;
+		if (isEmpty(config)) return this._config;
```
ericclemmons, line 76: `JS` is `@deprecated`, so I fixed it while I was here.

packages/pushnotification/src/PushNotification.ts
amhinson: You're correct. I don't think there is any value in adding that here.

packages/pushnotification/src/index.ts
```diff
@@ -10,11 +10,9 @@
  * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
  * and limitations under the License.
  */
-import { Amplify } from '@aws-amplify/core';
 import NotificationClass from './PushNotification';
 
 const _instance = new NotificationClass(null);
 const PushNotification = _instance;
-Amplify.register(PushNotification);
```
amhinson, line 18: This is still needed though, right?

packages/pushnotification/__tests__/PushNotification-test.ts
```diff
@@ -73,12 +73,12 @@ describe('PushNotification:', () => {
 			// Spy should be at 0 (it was already called on import)
 			expect(registerSpy).toHaveBeenCalledTimes(0);
 
-			// Global Amplify should keep original instance, not new instances
+			// Global Amplify will always reference last registered intance
```
amhinson, line 76:
```suggestion
			// Global Amplify will always reference last registered instance
```

packages/pushnotification/src/index.ts
ericclemmons: It is, but this was hoisted into `Pushnotification.ts`' `constructor`.

ericclemmons:
Hmmm, I've published to Verdaccio (https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#verdaccio), but now `ng build --prod` fails with this:

```
❯ yarn build
yarn run v1.22.4
$ ng build

ERROR in node_modules/@aws-amplify/ui-angular/dist/amplify-module.d.ts:1:22 - error NG6002: Appears in the NgModule.imports of AppModule, but could not be resolved to an NgModule class.

This likely means that the library (@aws-amplify/ui-angular) which declares AmplifyUIAngularModule has not been processed correctly by ngcc, or is not compatible with Angular Ivy. Check if a newer version of the library is available, and update if so. Also consider checking with the library's authors to see if the library is expected to be compatible with Ivy.

1 export declare class AmplifyUIAngularModule {
                       ~~~~~~~~~~~~~~~~~~~~~~

error Command failed with exit code 1.
```

I may test with https://docs.amplify.aws/start/q/integration/angular instead of https://github.com/aws-amplify/amplify-js-samples-staging/pull/166...

Amplifiyer:
Changes look good to me, just for my understanding, how does `Amplify.register()` gets called in server/node context?

ericclemmons:
@Amplifiyer I ninja-edited the description above:

`withSSRContext` explicitly registers modules to a _new_ `amplify` instance:

https://github.com/aws-amplify/amplify-js/blob/26ce5dfb0270009fc10f003f5627046ddaf5ae4e/packages/aws-amplify/src/withSSRContext.ts#L34-L40

wlee221 (approved):
This LGTM, thanks @ericclemmons. This is also for my understanding -- now that we’re registering modules if `browserOrNode().isBrowser` is true, how do modules get registered for react native? Trying to wrap my head around RN flows with Amplify :p

ericclemmons:
@wlee221 I'll test RN & Gatsby tomorrow.

I'm wondering the same thing 🤔 

I forgot our integration tests run on `main`, not in PRs, so I'm not comfortable merging and potentially getting surprised!

ericclemmons:
@amhinson I'll need to change my logic to `if (!browserOrNode().isNode)`. In React Native, `browserOrNode` returns:
> Screen Shot 2020-09-22 at 10 06 37 AM

Otherwise, we would need `if (isBrowser || isReactNative)` where `isReactNative` comes from:

```js
typeof navigator != 'undefined' && navigator.product == 'ReactNative')
```

ericclemmons:
Going through tests now.  Turns out our Jest environment is returning `{ isBrowser: true, isNode: true }` 🤦

ericclemmons:
Next steps:

- [ ] Remove all `if (...)` checks around `Amplify.register`.
- [ ] Audit what references are added to `Amplify._components`.  **In my testing, I saw this being a combination of categories, `[Function]`, and `undefined`!** This may be because some categories lack `getModuleName`, but I couldn't track it down.
- [ ] Ensure that `Amplify.register(category)` overrides `Amplify[category]` so that there is only a **single reference**.
- [ ] **On the client (browser & React Native)**, `console.warn` if `Amplify.register` is called again for a pre-registered category.
	
	We don't want this behavior on the server, because it's expected that new instances of `Auth` will be created and registered repeatedly.
- [ ] Ideally, `_modules` is used in favor of `_components` because pushing to `_components` on the server creates memory leaks! [96fc69b12ccf2428f65242eff045a36da8b612a9 ](https://github.com/aws-amplify/amplify-js/pull/6146/commits/96fc69b12ccf2428f65242eff045a36da8b612a9)


The goal of this is to accomplish the following:

- Prevent every call of `withSSRContext` on each request from pushing to `Amplify._components`, creating memory leaks.
- Warn customers on the client when they register multiple versions of a category (e.g. Auth).
- Simplify the `Amplify.register` usage to match pre-SSR implementation.
- Move `Amplify.register` outside of module scope so that we honor `sideEffects: false`.

ericclemmons:
Going to go with the simpler `sideEffects` implementation with @manueliglesias .

mordka:
@ericclemmons could you provide which PR to track and the rough bugfix release timeline? We are blocked by #6836 and don't know any workaround fix.

sammartinez:
@mordka We will provide a link to the PR once we have one, we are working on it as we speak to unblock yourself.

Issue #6802, opened 2020-09-16
@aws-amplify/ui-components requires @aws-amplify/interactions


seang96:
**Describe the bug**
Installing the a version of @ws-amplify/ui-react after 0.2.15 with only packaging requires @aws-amplify modules causes an error that module @aws-amplify/interactions cannot be resolved. I believe this is due to #6684. Can we get this updated to be a peer dependency or something since I do not need to use a chat bot component?

**To Reproduce**

1. Install latest @aws-amplify/ui-components library or the respective frameworks without installing @aws-amplify as a whole.

**Expected behavior**
@aws-amplify/interactions to not be a required module to use the components library and compile.

ashika01:
@seang96: Looking into this. Just curious, Is there a reason for you to not use `aws-amplify` but import the modularized imports? Because with modularization changes we rolled out, it should matter if you import `aws-amplify` or modular ones. Ultimately when you go production, it should enable tree-shaking for you so you wouldnt have any perf impact right?

seang96:
@ashika01,
Yeah, that may be true now. Earlier though it did require installing modules individually though. Which my software is doing already.

PR #6811 shows I am not the only one with software that is still importing them as such. From what I saw with that issue Amplify Core registers Credentials as soon as Amplify.Configure is called when then instantiates Auth, where it may not be needed yet in the case the application is serving public pages that doesn't necessarily require backend services until the user reaches a login screen. Anyways, in the case of that issue though if Auth wasn't registered before Credentials registered any library that used Credentials would fail to authenticate properly, AWSS3Provider being one of them.

Another use case is I am using local builds for some packages, more often than not for DataStore, due to PR's that have not been merged yet that are required for my application to run. For development purposes I'd rather build only one module rather than all of them.

ashika01:
@seang96 Thanks for the info. The issue solved by #6811 is different from our current issue. 

I looked more into this. Looks like one of our entry point in the module generated does not have a clear distinction in the modules needed. Moving it to peer dependency doesn't solve the problem. I am analyzing the bundles generated to find the root cause. We will keep you posted here.

ashika01:
Closing this issue since the PR for this is merged in and released.

Issue #6515, opened 2020-08-06
Can't resolve '@aws-amplify/storage' in '@aws-amplify/ui-components/dist/esm'


gsans:
**Describe the bug**
When building an angular app with `ng serve` these errors show:

```
ERROR in ./node_modules/@aws-amplify/ui-components/dist/esm/amplify-s3-text.entry.js
Module not found: Error: Can't resolve '@aws-amplify/storage' in '/Users/gsans/Desktop/Projects/tutorials/amplify-app-demo/node_modules/@aws-amplify/ui-components/dist/esm'

ERROR in ./node_modules/@aws-amplify/ui-components/dist/esm/amplify-s3-text-picker.entry.js
Module not found: Error: Can't resolve '@aws-amplify/storage' in '/Users/gsans/Desktop/Projects/tutorials/amplify-app-demo/node_modules/@aws-amplify/ui-components/dist/esm'

ERROR in ./node_modules/@aws-amplify/ui-components/dist/esm/amplify-s3-image-picker.entry.js
Module not found: Error: Can't resolve '@aws-amplify/storage' in '/Users/gsans/Desktop/Projects/tutorials/amplify-app-demo/node_modules/@aws-amplify/ui-components/dist/esm'

ERROR in ./node_modules/@aws-amplify/ui-components/dist/esm/amplify-s3-image.entry.js
Module not found: Error: Can't resolve '@aws-amplify/storage' in '/Users/gsans/Desktop/Projects/tutorials/amplify-app-demo/node_modules/@aws-amplify/ui-components/dist/esm'

ERROR in ./node_modules/@aws-amplify/ui-components/dist/esm/amplify-s3-album.entry.js
Module not found: Error: Can't resolve '@aws-amplify/storage' in '/Users/gsans/Desktop/Projects/tutorials/amplify-app-demo/node_modules/@aws-amplify/ui-components/dist/esm'

ERROR in ./node_modules/@aws-amplify/ui-components/dist/esm/storage-helper-21c6d15f.js
Module not found: Error: Can't resolve '@aws-amplify/storage' in '/Users/gsans/Desktop/Projects/tutorials/amplify-app-demo/node_modules/@aws-amplify/ui-components/dist/esm'
```

In order to resolve I had to run `npm i @aws-amplify/storage` which shouldn't be necessary.

**To Reproduce**

1. Follow the Getting started guide for Angular [here](https://docs.amplify.aws/start/q/integration/angular)
[...]

Issue #6239, opened 2020-07-06
[aws-amplify-angular 3.X.X] Can't resolve '@aws-amplify/*' while using just the auth module


urfnaino:
**Describe the bug**
I'm currently on version 2.X.X and here are my aws-amplify dependencies in package.json:
```
"@aws-amplify/auth": "^2.1.8",
"@aws-amplify/core": "^2.3.0",
"aws-amplify-angular": "^4.1.9"
```
This configuration works without any issue.

But when I try to upgrade to version 3.X.X, the exact set of dependencies doesn't work. Here're the ng serve logs:
```
chunk {main} main.js, main.js.map (main) 480 kB [initial] [rendered]
chunk {polyfills} polyfills.js, polyfills.js.map (polyfills) 414 kB [initial] [rendered]
chunk {runtime} runtime.js, runtime.js.map (runtime) 6.15 kB [entry] [rendered]
chunk {styles} styles.js, styles.js.map (styles) 3.4 MB [initial] [rendered]
chunk {vendor} vendor.js, vendor.js.map (vendor) 25.3 MB [initial] [rendered]
Date: 2020-07-02T14:36:15.680Z - Hash: f254b388275b37a4e3ac - Time: 64792ms

ERROR in ./node_modules/aws-amplify-angular/dist/src/providers/amplify.service.js
Module not found: Error: Can't resolve '@aws-amplify/analytics' in 'C:\Source\Git\cloudanalytics\visualization\client\node_modules\aws-amplify-angular\dist\src\providers'
ERROR in ./node_modules/aws-amplify-angular/dist/src/providers/amplify.service.js
Module not found: Error: Can't resolve '@aws-amplify/api' in 'C:\Source\Git\cloudanalytics\visualization\client\node_modules\aws-amplify-angular\dist\src\providers'
ERROR in ./node_modules/aws-amplify-angular/dist/src/providers/amplify.service.js
Module not found: Error: Can't resolve '@aws-amplify/interactions' in 'C:\Source\Git\cloudanalytics\visualization\client\node_modules\aws-amplify-angular\dist\src\providers'
ERROR in ./node_modules/aws-amplify-angular/dist/src/providers/amplify.service.js
Module not found: Error: Can't resolve '@aws-amplify/pubsub' in 'C:\Source\Git\cloudanalytics\visualization\client\node_modules\aws-amplify-angular\dist\src\providers'
[...]

PR #6847, opened 2020-09-22
[@aws-amplify/ui-components] Module import resolution

ashika01:
_Issue #, if available:_
#6802 
#6239 
#6515 

_Description of changes:_
- Cleaning up the dependencies. 
- The modularization should take care of tree shaking. 
- And the change provides for better DX.
- Tested locally in all framework and against our integration tests.


packages/amplify-ui-components/package.json
```diff
@@ -36,11 +36,14 @@
   },
   "dependencies": {
     "qrcode": "^1.4.4",
-    "uuid": "^8.2.0"
-  },
-  "devDependencies": {
+    "uuid": "^8.2.0",
```
ericclemmons, line 39: Looks like there's a trailing ` ` here. If you ran `yarn install uuid`, would the formatting of this file change?

ericclemmons (approved):
This looks this should resolve customer issues!

It appears that the root cause is that we (mostly) document installing `aws-amplify`, and that's what we test for.

Customers who've only installed `@aws-amplify/auth` and `core` won't have the other dependencies, which this addresses.

I'm in favor of shipping this to unblock those customers (since tree-shaking should remove dupes, and these versions are in sync with `aws-amplify`).

However, it points to the root issue that our docs (1) need updating to point to a singular `aws-amplify` import and (2) being able to depend on it for UI components and tree-shaking.

Until the docs are updated and we can confidently inform customers that tree-shaking has been solved for ~6+ months, we'll likely see more edge-cases with modular imports.

Thanks for getting this tested & resolved @ashika01 !

Amplifiyer (approved):

ashika01:
@ericclemmons With this change all of them should work. Using Named import of `Amplify` from core or usage of `Amplify` from `aws-amplify` or configuring other modular import like `Auth`, `Storage` etc.

I will follow this up with a doc PR to unify to `aws-amplify` import.

Issue #5218, opened 2020-03-27
Using custom state with Cognito Hosted UI leads to OAuth Error while validating state


jomuhm:
**Description**
I am attempting to implement redirect after signin following this pattern https://reacttraining.com/react-router/web/example/auth-workflow. I start the auth flow from a page that is auth protected and want to end at the same location after users successfully authenticate. In order to do so, I am passing the pathname of the page as customState to  ```Auth.federatedSignIn({customState: location.pathname}); // i.e. location.pathname = '/protected'```.

When Cognito redirects back to my sign in page. The auth flow fails with the following error.

```
OAuth - Error handling auth response. Error: Invalid state in OAuth flow
    at OAuth.../node_modules/@aws-amplify/auth/lib-esm/OAuth/OAuth.js.OAuth._validateState (OAuth.js:244)
    at OAuth.<anonymous> (OAuth.js:216)
    at step (OAuth.js:55)
    at Object.next (OAuth.js:36)
    at OAuth.js:30
    at new Promise (<anonymous>)
    at ../node_modules/@aws-amplify/auth/lib-esm/OAuth/OAuth.js.__awaiter (OAuth.js:26)
    at OAuth.../node_modules/@aws-amplify/auth/lib-esm/OAuth/OAuth.js.OAuth.handleAuthResponse (OAuth.js:190)
    at AuthClass.<anonymous> (Auth.js:1683)
    at step (Auth.js:44)
```

The problem appears to be related to how Amplify and the Cognito Auth service handle url encoding of the state parameter. When I perform the above, Amplify url encodes the customState value and stores that url encoded value in it's local storage. The redirect URL to Cognito contains the URL encoded state parameter:

```state=cfjcnuSJgihzcAJP5uFZVuggAVcGDU34-%2Fprotected.```

After authenticating, Cognito redirects back to by redirect_uri passing the state parameter unencoded.

```state=cfjcnuSJgihzcAJP5uFZVuggAVcGDU34-/protected```

The[ state validation logic](https://github.com/aws-amplify/amplify-js/blob/master/packages/auth/src/OAuth/OAuth.ts#L240) then fails and stops the auth flow.

**To Reproduce**
1. Using ```"aws-amplify": "^2.2.1"```
1. Configure Amplify to use a Cognito Hosted UI for federated signin.
1. Pass a path to redirect as customState to ```Auth.federatedSignIn({customState:'/path/to/redirect/after/auth'});```
[...]

PR #6851, opened 2020-09-23
bugfix #5218: de/encode custom state using base16

div5yesh:
_Issue #, if available:_ #5218

_Description of changes:_

 - encode custom state using base16 to preserve state from Cognito's URI encoding.
 - add base16 encoding/decoding util functions to **aws-amplify/core**.


packages/core/src/Util/StringUtils.ts
```diff
@@ -0,0 +1,18 @@
+export function str2hex(str: string) {
```
ericclemmons, line 1: Can you add tests for these?

- Capture the type of `customState` that `encodeURIComponent` ran into issues with (URLs)
- Capture the encode/decode behavior
- Test that an encoded string isn't affected by `decodeURI*`.

packages/auth/src/OAuth/OAuth.ts
```diff
@@ -79,10 +79,10 @@ export default class OAuth {
 	) {
 		const generatedState = this._generateState(32);
 		const state = customState
-			? `${generatedState}-${customState}`
```
ericclemmons, line 82: Can you add a comment on why `customState` is getting encoded?  With the Auth refactoring, I don't want us to miss how we would see double-encoding/decoding through federated sign-in flows, but single-encoding/decoding when already signed in.

ericclemmons: Now that I say that, maybe we won't need as much of a comment with a different name that states the intent?  Like `urlSafeEncode`/`urlSafeDecode`?

I'm not loving the suggestion, but this was tricky enough that I want to make it clear to future contributors that `encodeURIComponent` is **not** safe.

ericclemmons (commented):
This is looking great @div5yesh! I'm ready to approve, but would want to see a couple tests validating this since we ran into interesting behavior.

packages/auth/src/OAuth/OAuth.ts
div5yesh: Yes, I agree. A Note and descriptive function name will add more clarity and context to this issue.

packages/core/src/Util/StringUtils.ts
```diff
@@ -0,0 +1,9 @@
+import { toHex, fromHex } from '@aws-sdk/util-hex-encoding';
+
+export function urlSafeEncode(str: string) {
+	return toHex(new TextEncoder().encode(str));
```
ericclemmons, line 4: TIL about `TextEncoder`!
> https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder

Since IE (including 11) doesn't support it, will this be a risk for people who've gotten Amplify to work with their IE 11 projects with existing polyfills?

(Like, are we using `TextEncoder` in other parts of our code, so this isn't "new" behavior?)

ericclemmons (approved):
This looks good!

The only concern I have is if this patch release will break existing IE 11 apps due to the use of `TextEncoder`.

IE 11 usage requires a good deal of polyfills by customers today, so needing a polyfill won't be _new_, but I'd like to also avoid breaking their apps the next time they `npm install`.

If we're already using `TextEncoder` elsewhere, I think we're good!  Otherwise, are we converting strings to streams differently? (This implementing looks like what we needed to re-use `toHex`/`fromHex`)

@elorzafe What are your thoughts?

packages/core/src/Util/StringUtils.ts
div5yesh: Agreed. Removed `TextEncoder`.

ericclemmons: Thanks @div5yesh !

I feel much better about this now.

Sorry for the back-and-forth, but this is approved and ready to go once tests pass.

@elorzafe Anything else you'd like to see before approving?

elorzafe (approved):
Thanks @div5yesh ! 👍 🌮 🎉

Issue #3682, opened 2019-07-18
null is not an object (evaluating RNAWSCognito.computeModPow)


kirkryan:
**Describe the bug**
I hope this saves someone some time (and frustration)

A user can sign up or view the login page, but upon login, the following error is observed:

**null is not an object (evaluating RNAWSCognito.computeModPow)**

**To Reproduce**
Create new >=RN0.60.0 project
cd ios
pod install
build

**Expected behavior**
Amplify podspec should correctly list RNAWSCognito.

**Smartphone (please complete the following information):**
 - All

**Additional context**
remember  **not** to use react-native link when working in >=0.60.0

**Fix**
simply run the following in your project root

```
yarn add amazon-cognito-identity-js
cd ios
pod install
```

haverchuck:
@kirkryan - Thanks for posting this.  Can you post a little bit of detail on how you've configured your app?  Are you using the aws-amplify-react-native authenticator component?

kirkryan:
@haverchuck - that’s correct. This is when using the withAuthenticator HOC on iOS builds.

haverchuck:
@kirkryan  - Thanks.  I think this is an expected step, but we need to update our docs.

BradyShober:
I have also run into this issue but when trying to use the recommended fix, I get a duplicate symbols Linker Error. In my app we are not using aws-amplify-react-native component. The specific error is 

> duplicate symbol _OBJC_CLASS_$_JKBigInteger in:
    /Users/brady/Library/Developer/Xcode/DerivedData/traduki-ekuctotjpfcwpphgyaufmdogypjm/Build/Products/Debug-iphonesimulator/JKBigInteger2/libJKBigInteger2.a(JKBigInteger.o)
    /Users/brady/Library/Developer/Xcode/DerivedData/traduki-ekuctotjpfcwpphgyaufmdogypjm/Build/Products/Debug-iphonesimulator/libRNAWSCognito.a(JKBigInteger.o)
duplicate symbol _OBJC_METACLASS_$_JKBigInteger in:
    /Users/brady/Library/Developer/Xcode/DerivedData/traduki-ekuctotjpfcwpphgyaufmdogypjm/Build/Products/Debug-iphonesimulator/JKBigInteger2/libJKBigInteger2.a(JKBigInteger.o)
[...]

PR #6872, opened 2020-09-25
Support Microsoft's react-native-windows (macOS)

mandrigin:
_Issue #, if available:_ N/A

_Description of changes:_

The current version of authenticator in amplify-js not working on react-native on macOS (https://github.com/Microsoft/react-native-windows), it fails with this issue: https://github.com/aws-amplify/amplify-js/issues/3682.

No workaround works because the podfile doesn't allow this code to run on macOS. At the same time, all the code I see there is perfectly valid for macOS, compiles and runs there.

After this change in the Podfile, I am able to compile and run the authenticator on react-native app on macOS.


mandrigin:
Hmm, I'm not really sure why `yarn install` failed on circleci; there are no JS or dependency changes here...

ericclemmons:
@mandrigin I restarted the build in CircleCI. Also no idea why it failed 🤷

mandrigin:
Thank you! Now it’s a different failure that has nothing to do with iOS 😢. 
Maybe I used a broken commit as the base to my branch?

mandrigin:
@ericclemmons it was super outdated branch for some reason... I've rebased it and let's see how it goes now.

sammartinez:
@mandrigin, Thanks for opening this pull request and contributing. At this time, we are not looking to add this support as we would need to do a full design review and add supporting documentation for this. However, we will take this into consideration for the future as well look further for feature request specific to React Native. Thanks again!

mandrigin:
This was a single config change in a podspec but sure...

PR #6895, opened 2020-10-01
ci: minor improvements

iartemiev:
Just some minor improvements:
* Generate and store yarn.lock as an artifact (then delete it from source)
  * This will allow us to more easily debug dependency-related build issues by diff'ing against the lockfile from the last known working build. We've spent entire days simply trying to pinpoint which upstream dependency got updated with breaking changes. This should expedite that process if it happens again.
* Remove save_cache/restore_cache steps for CocoaPods in RN integ test jobs. More trouble than it's worth, causes issues when RN itself and RN-deps are upgraded in the sample apps. 
* Upgrade the macOS executor to XCode 11.5.0 (this is the latest we can use until we upgrade our Android emulator)
* Test run for pipeline changes: https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/6296/workflows/7ba7fbe0-22f7-460a-a962-7f61906ee7c5

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,5 @@
 # Javascript Node CircleCI 2.1 configuration file
 #
-# Check https://circleci.com/docs/2.0/language-javascript/ for more details
-#
 version: 2.1
 
 executors:
@@ -19,7 +17,7 @@ executors:
 
   macos-executor:
     macos:
-      xcode: 11.4.1
+      xcode: 11.5.0
 
 test_env_vars: &test_env_vars
   environment:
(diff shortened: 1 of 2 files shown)


sammartinez (approved):
LGTM 🌮

.circleci/config.yml
```diff
@@ -279,10 +261,22 @@ jobs:
...
+          name: yarn install --non-interactive
+          # temporarily rename .yarnrc before installing so that we can generate a yarn.lock artifact
+          command: |
+            mv .yarnrc ._yarnrc
+            yarn
+      - run: yarn build
+      # storing yarn.lock as an artifact, so that we can quickly get a dependency diff
+      # with the last working build in the event that some upstream deps break the build in the future
+      - store_artifacts:
+          path: yarn.lock
```
amhinson, line 274: This is great!

PR #6900, opened 2020-10-01
add static props to CognitoUserAttribute

k-yle:
Take a look at https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/src/CognitoUserAttribute.js#L26-L27, note how the `CognitoUserAttribute` class has properties `Name` and `Value` which are strings. 

These properties are missing from the type definitions, this PR adds them.

Usage example:

```ts
const attributes = await Auth.userAttributes(user);

attributes[0].Name // this works but the typedefs didn't reflect this until now
attributes[0].Value
```

This is the proper fix to #5051, which makes that PR redundant. I also got confused by this and opened #6766 which also wasn't the proper solution

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license. ✅ Yes


sammartinez (approved):
LGTM 🌮

Issue #6521, opened 2020-08-06
Create new account page does not detect filled out fields after translation - Amplify Authenticator UI Component


liamaws:
**Describe the bug**
After filling out the create account page of the Amplify Authenticator UI Component and then changing the translation, the fields remain filled out on the UI but are not detected when clicking the create account button.

**To Reproduce**
1. Go to https://d2zxvlukxu4lhp.cloudfront.net/
2. Click on Create Account
3. Fill out email address, password and phone number
4. Click top button to switch language to french
5. Click Creer Compte (Create account in french)

I would like to point out that this problem is not specific to using a translation.
It can also be reproduced by doing the following

1. Go to https://d2zxvlukxu4lhp.cloudfront.net/
2. Click on Create Account
3. Fill out email address, password and phone number
4. Don't click on create account but instead click sign in to go back to the sign in page
5. From the sign in page click create account again
6. Now on the create account page the information you entered before will still be there. Click create account

**Expected behavior**
An account will be created with the information that has been filled out and is present on the UI.

**Code Snippet**
```
          <AmplifyAuthenticator usernameAlias="email" style={{ textAlign: 'center' }}>
            <AmplifyForgotPassword
              usernameAlias="email"
              slot="forgot-password"
              formFields={[
                {
                  type: "email",
                  required: true,
                },
              ]}></AmplifyForgotPassword>
            <AmplifySignIn
              usernameAlias="email"
              slot="sign-in"
              formFields={[
                {
                  type: "email",
                  required: true,
                },
                {
                  type: "password",
                  required: true,
                }
              ]}></AmplifySignIn>
            <AmplifySignUp
              usernameAlias="email"
              slot="sign-up"
              formFields={[
                {
                  type: "email",
                  required: true,
                },
                {
                  type: "password",
                  required: true,
                },
                {
                  type: "phone_number",
                  required: false,
                },
                {
                  type: "locale",
                  value: this.state.lang,
                  inputProps: {
                    type: 'hidden',
                  }
                }
              ]}></AmplifySignUp>
            <div>
              {I18n.get("WAIT_REDIRECTION")}
              <AmplifySignOut />
            </div>
          </AmplifyAuthenticator>
```

**Screenshots**
![image](https://user-images.githubusercontent.com/67014943/89562640-45b38980-d7e8-11ea-9c5d-47a3b9da9677.png)

xavierraffin:
[...]

PR #6901, opened 2020-10-02
Bugfix: Translation and field saving

xavierraffin:
_Issue #, if available:_
fixes: #6521

_Description of changes:_
Now the translation update and the value of the field is taken appropriately in:
* signUp
* signIn
* forgotPassword
* confirmSignUp


sammartinez:
@xavierraffin I'm trying to understand, how will this solve for the rerender case for Translations? Or is this looking to solve something else that I may be missing. Thanks ahead of time!

xavierraffin:
OK, the problem come from the fact that `buildFormFields` is called in two cases:

```
  componentWillLoad() {
    checkUsernameAlias(this.usernameAlias);	    
    this.buildFormFields();	    
 }
  @Watch('formFields')	 
  formFieldsHandler() {	 
    this.buildFormFields();	  
 }
```

When the component is build first time + when the formFields are updated (this happen when you change language).

Before PR, if you start entering value in field then change language, the buildFormFields execute and erase the value to the one provided in formFields (empty for example). When you submit the form , because 

Because the behavior (before this PR) is to submit the value in newFormFields  variable the submit fail, complaining for empty value where the field are not empty.

Also it fixes some copy paste error that has been introduce in the initial PR from Jordan (in amplify-forgot-password.tsx the array was not flushed).

xavierraffin:
Excuse me but the bug is still there and this is the fix.
Please let me know if you want anything more on this PR?

sammartinez:
> @sammartinez me but the bug is still there and this is the fix.
> Please let me know if you want anything more on this PR?

Hey @xavierraffin, followed up with you offline on next steps here. Will provide an update from the team once I am able to run this implementation by them.

PR #6937, opened 2020-10-08
fix(@aws-amplify/pubsub): Block subscriptions for SSR, not Node

### LINKED ISSUE #6873: PubSub: Unblock subscriptions in desktop Node environments
A recent update to the `PubSub` component in this [pull request](https://github.com/aws-amplify/amplify-js/pull/6726), checks for a Node environment, and if it exists, prevents subscriptions from being established.

I understand the justification for this regarding multi-tenant apps sharing sockets. In our implementation, however, we are using the PubSub module in a client application running in a desktop environment and built with Electron. This pull request breaks the functionality of the application entirely, and we can not use the latest update.

Per my conversation with @ericclemmons in the PR, I'd like to request a change to the `isNode` check.  The two solutions proposed are both workable; however, I like the simplicity of checking for `config.ssr` over checking for frameworks.

ericclemmons:
_Issue #, if available:_ Fixes #6873 

#6146 & #6726 blocked `PubSub.subscribe` from being used in Node environments for security reasons (multi-tenant requests sharing the same socket), but `isNode` isn't a reliable indicator of that usage: `ssr: true` is.

`isNode` is a valid runtime for single-tenant environments, such as Electron.

This PR improves the specificity of the `PubSub.subscribe` check to match the use-case it was designed for (SSR) vs. blocking legitimate usages of sockets.

---


packages/pubsub/__tests__/PubSub-unit-test.node.ts
```diff
@@ -1,24 +0,0 @@
-jest.mock('@aws-amplify/core', () => ({
```
ericclemmons, line 1: This was actually renamed to `.ssr.ts`, but showing as a deletion 🤷

manueliglesias (approved):
LGTM Thanks @ericclemmons !

sammartinez (approved):
LGTM 🌮

Issue #6080, opened 2020-06-12
Auth.verifyUserAttribute() does not handle clientMetadata


damien-monni:
I want to resend the email verification code on user request so I try to use Auth.verifyUserAttribute:

```javascript
const user = await Auth.currentAuthenticatedUser();
await Auth.verifyUserAttribute(user, 'email');

// I also tried:
await Auth.verifyUserAttribute(user, 'email', { foo: 'bar' });
```

I need to send clientMetadata that are set on Auth.configure but it seems that no metadata are sent.

I don't understand why because it seems to be implemented:
https://github.com/aws-amplify/amplify-js/blob/a047ce73abe98c3bf82e888c3afb4d2f911805f3/packages/auth/src/Auth.ts#L1352

Any help would be very appreciated.

To reproduce:

```javascript
 Auth.configure({
    ...
    clientMetadata: {
      foo: 'bar',
    },
  });

...

Auth.verifyUserAttribute(user, 'email');
```

Then log metadata on the customMessage lambda trigger. They won't be set.

sammartinez:
@damien-monni Can you provide us with your `package.json`? Thanks ahead of time!

damien-monni:
@sammartinez sure:

```json
{
  "name": "keep-blogging-front",
  "version": "0.0.0",
  "author": "",
  "license": "ISC",
  "scripts": {
    "dev": "next -p 3030",
    "build": "next build",
    "start": "next start -p 3030"
  },
  "dependencies": {
    "@anzerr/base.util": "^1.0.10",
    "@apollo/react-hooks": "^3.1.5",
    "@apollo/react-ssr": "^3.1.5",
    "@aws-amplify/auth": "^3.2.13",
    "@aws-amplify/core": "^3.3.3",
    "@aws-amplify/storage": "^3.2.6",
    "@material-ui/core": "^4.10.2",
    "@material-ui/icons": "^4.9.1",
    "@material-ui/lab": "^4.0.0-alpha.56",
    "@material-ui/pickers": "^4.0.0-alpha.8",
    "apollo-cache-inmemory": "^1.6.6",
    "apollo-client": "^2.6.10",
    "apollo-link-context": "^1.0.20",
    "apollo-link-http": "^1.5.17",
    "chroma-js": "^2.1.0",
    "clsx": "^1.1.1",
    "dotenv": "^8.2.0",
    "get-video-id": "^3.1.4",
    "graphql": "^15.1.0",
[...]

Issue #5486, opened 2020-04-21
verifyUserAttribute does not appear to send clientMetadata


rifflock:
**Describe the bug**
While implementing MFA I'm attempting to add a phone number attribute to the user. UpdateUser works fine (though it doesn't trigger the CustomMessage_UpdateUserAttribute lambda, but that's not an amplify issue).
When I call `Auth.verifyUserAttribute(user, 'phone_number', { product: 'myProductName' }` the cloudwatch logs show no `clientMetadata` property in the event.

Error from my backend code:
```
{code: "UserLambdaValidationException", name: "UserLambdaValidationException", message: "CustomMessage failed with error Unsupport product undefined."}
```

**To Reproduce**

1. Create a pool with a user that has an unverified phone number, and a CustomMessage trigger
2. Use the Amplify package to call `verifyUserAttribute` for that user and pass the third parameter some object;
3. Check the logs of the CustomMessage trigger and see something like
```
    "triggerSource": "CustomMessage_VerifyUserAttribute",
    "request": {
        "userAttributes": {
            "sub": "<sub>",
            "email_verified": "true",
            "cognito:user_status": "CONFIRMED",
            "cognito:email_alias": "<email>",
            "phone_number_verified": "false",
            "phone_number": "<phone>",
            "given_name": "Michael",
            "family_name": "Riffle",
            "email": "<email>"
        },
        "codeParameter": "{####}",
        "linkParameter": "{##Click Here##}",
        "usernameParameter": null
    },
```

**Expected behavior**
The `request` property should have a `clientMetadata` property that was passed into the function.

**Code Snippet**
Please provide a code snippet or a link to sample code of the issue you are experiencing to help us reproduce the issue. (Be sure to remove any sensitive data)

```
  async verifyUserAttribute() {
    try {
      const user = await Auth.currentAuthenticatedUser();
      if (!user) {
        throw new Error('User must be initialized to verify attributes');
      }

      const res = await Auth.verifyUserAttribute(user, 'phone_number', {
        product: process.env.REACT_APP_PRODUCT,
      });
    } catch (err) {
      throw err;
    }
  }
```

**Additional context**
We're using lambda triggers from cognito for a bunch of other use-cases and everything is working fine. Our app supports
      CustomMessage_SignUp
      CustomMessage_ResendCode
      CustomMessage_ForgotPassword
      PostConfirmation_ConfirmSignUp
      PostAuthentication_Authentication
And they all work great.
[...]

PR #6941, opened 2020-10-09
fix(@aws-amplify/auth) fix the way clientMetadata is used in Auth.verifyUserAttribute

xanagi:
_Issue #, if available:_

#5486, #6080

_Description of changes:_

When clientMetadata was set in Auth.verifyUserAttribute, it was set as a part of callback object.
It should be the third argument of user.getAttributeVerificationCode, not a part of the second.
This is the reason why clientMetadata was not handled.

This PR fixes  fix the way clientMetadata is passed in Auth.verifyUserAttribute.


amhinson (approved):
Great catch @xanagi 🙏  LGTM

elorzafe (approved):
Thanks @xanagi 🥇

PR #6957, opened 2020-10-13
ci: speed up post_release job

iartemiev:
* only build @aws-amplify/core to generate *version.ts* instead of re-building the entire lib
* will speed up post_release by about 4x


amhinson (approved):
Nice!

PR #6979, opened 2020-10-15
Add case insensitive operators for strings

RandyLarzabal:
Add case insensitive version of string operators


amhinson:
Thanks for the PR @RandyLarzabal! Do you have a reference as to why this is needed and the problem that it solves?

djbuch:
> Thanks for the PR @RandyLarzabal! Do you have a reference as to why this is needed and the problem that it solves?

Hello I’m working with @RandyLarzabal on a projet where we need to filter data from the DataStore. In certain case it is much more UX not to worry about case, when searching for a name for example, not needing to know if person entered his name in full capital or only with the first letter as a capital. But there would be a lot of other use cases.

Regards

svidgen:
I can definitely understand the benefit of a feature like this. Unfortunately, I don't think this can be done so simply across the board. The SQLite adapter, for example, does NOT create columns or query with `nocase` collation (AFAIK). And, some in-flight IndexedDB optimization work leverages indexes, which I believe are still strictly case-sensitive to-date.

As a workaround, you could create additional columns in your schema reserved explicitly for searching. You would then just populate those columns with canonicalized versions of your search fields &mdash; I'd posit that this is often beneficial for normalizing in other ways too (spacing, character and abbreviation, replacements, etc.).

That said, if we want to pursue this feature, let's see a feature request to gauge demand/impact, flesh out the intended use-cases, and make sure we actually meet the needs across the board on this one.

Issue #6352, opened 2020-07-17
Publish Hub event on token refreshes


jpignata:
**Is your feature request related to a problem? Please describe.**
Provide a Hub notification when tokens refresh.

**Describe the solution you'd like**
When a token refresh event occurs, publish this to Hub, so an application can take some action. (More details on specific use cases would be helpful here).
```
Hub.listen('auth', (data) => {
 switch (data.payload.event) {
   case 'token_refresh': 
   logger.log("New token has been process");
  ...
  }
}
```

**Describe alternatives you've considered**
None. These events aren't generated, so there isn't any way to hook onto this.

**Additional context**
we would need to update our [docs page](https://docs.amplify.aws/lib/utilities/hub/q/platform/js#authentication-events) for this.

PR #6983, opened 2020-10-15
feat: enable users to listen for token refresh event

harrysolovay:
_Issue:_ #6352 

_Description of changes:_

Enables users to do the following.

```ts
Hub.listen('auth', (data) => {
  switch (data.payload.event) {
    case 'token_refresh': {
      logger.log("New token has been process");
    }
  }
});
```


packages/auth/src/Auth.ts
```diff
@@ -84,6 +84,19 @@ const dispatchAuthEvent = (event: string, data: any, message: string) => {
...
 };
 
+const eventEmitters = {
+	onTokenRefresh() {
+		dispatchAuthEvent('token_refresh', undefined, `New token retrieved`);
+	},
+	onTokenRefreshFailure(error: Error) {
+		dispatchAuthEvent(
+			'token_refresh_failure',
+			error,
```
sammartinez, line 94: Can you provide an example of this `error`?

harrysolovay: Could be any of [these](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/CommonErrors.html), for instance.

packages/auth/src/Auth.ts
```diff
@@ -84,6 +84,19 @@ const dispatchAuthEvent = (event: string, data: any, message: string) => {
 	Hub.dispatch('auth', { event, data, message }, 'Auth', AMPLIFY_SYMBOL);
 };
 
+const eventEmitters = {
+	onTokenRefresh() {
+		dispatchAuthEvent('token_refresh', undefined, `New token retrieved`);
```
elorzafe, line 89: What will happen when destructuring `undefined`?

elorzafe: nvm

PR #7008, opened 2020-10-20
fix: bring #6921 commit to main

wlee221:
_Issue #, if available:_ #6921 

_Description of changes:_ #6921 was targeted to master and not main. Pulling this commit back to main.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6921, opened 2020-10-06
Update SignIn.vue

sksabircn:
Forget your password? the text has a typo

_Issue #, if available:_

_Description of changes:_

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7009, opened 2020-10-20
fix: rewrite "forget password" to "forgot password"

wlee221:
_Issue #, if available:_ #6921, #7008.

_Description of changes:_ rewrite "forget password" to "forgot password". This is a follow up to #7008 and #6921.


Ashish-Nanda (approved):
LGTM 🚀

sammartinez (approved):
LGTM 🌮

amhinson:
The only other thing I'd consider here is customers with their own language customizations (https://docs.amplify.aws/ui/customization/translations/q/framework/react), as this wouldn't show the correct translation without updating their code.

wlee221:
@amhinson, that is a good point. Because this is a deprecating component, we may want to just leave it as-is. What do you think @sammartinez?

sammartinez:
> @amhinson, that is a good point. Because this is a deprecating component, we may want to just leave it as-is. What do you think @sammartinez?

I believe we can do this fix as this would be a patch for customers utilizing this component. I am OK with this change going in but to @amhinson, customers will need to upgrade their code.

PR #7023, opened 2020-10-21
Add ts interface for the federation options so that its easier to understand the option to pass

Anenth:
Add ts interface for the federation options so that its easier to understand the option to pass

_Issue #, if available:_

_Description of changes:_

Add ts interface for the federation options so that it's easier to understand the option to pass. 


amhinson:
Hey @Anenth thanks for the PR! However, it does look like this addition you're adding is already [implemented](https://github.com/aws-amplify/amplify-js/blob/main/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx#L45) in our new UI components (`@aws-amplify/ui-react`). Here's the [migration guide](https://docs.amplify.aws/ui/auth/authenticator/q/framework/react#migration) going from the old library to the new one.

PR #7040, opened 2020-10-22
deps: pin amplify versions

wlee221:
_Issue #, if available:_

_Description of changes:_ This pins amplify versions across our monorepo. See #6972 for discussions. 🥕 


wlee221:
Not familiar with how lerna handles version bumps, will lerna no longer add 🥕  based on this commit?

sammartinez:
> Not familiar with how lerna handles version bumps, will lerna no longer add 🥕 based on this commit?

@iartemiev would know more so on this.

packages/amplify-ui-components/package.json
```diff
@@ -35,11 +35,11 @@
...
-    "@aws-amplify/auth": "^3.4.7",
-    "@aws-amplify/core": "^3.7.0",
-    "@aws-amplify/interactions": "^3.3.7",
-    "@aws-amplify/storage": "^3.3.7",
-    "@aws-amplify/xr": "^2.2.7",
+    "@aws-amplify/auth": "3.4.7",
+    "@aws-amplify/core": "3.7.0",
+    "@aws-amplify/interactions": "3.3.7",
+    "@aws-amplify/storage": "3.3.7",
+    "@aws-amplify/xr": "2.2.7",
```
ericclemmons, line 42: This part seems riskiest IMO.

Ideally, we'd want `@aws-amplify/ui-components` to depend on `aws-amplify` as a single peer-dependency.

However, since we had customer issues where they installed `@aws-amplify/core` and `@aws-amplify/auth` separately, they were missing `@aws-amplify/storage`.

This may be better served as `3.x.x` or `^3.0.0` (also `2.x.x` or `^2.0.0`).

**Put another way, we want to mitigate any risk that `@aws-amplify/ui-react` and `@aws-amplify/auth` have two different versions in customers' projects**.

sammartinez: Valid point @ericclemmons, I'm also just curious, why would we not want to just have `aws-amplify` be a dependency? do we really need each module import?

ericclemmons:
> Not familiar with how lerna handles version bumps, will lerna no longer add 🥕 based on this commit?
> 
> @iartemiev would know more so on this.

`lerna` _should_ respect the presence/absence of a caret (this can be tested locally with verdaccio), though `lerna --init`'s behavior defaults to `^`.

We can force the version using `--exact`, IIRC:
> https://github.com/lerna/lerna/blob/master/commands/version/README.md#--exact

packages/amplify-ui-components/package.json
ericclemmons: @sammartinez Ideally with the next major release, we would have a peer dependency on `aws-amplify` as you described.

From what I remember, this change came from https://github.com/aws-amplify/amplify-js/pull/6847#pullrequestreview-493742316 due to **customers using modular imports and not having some of the dependencies for UI components**. (i.e. they tried to use a storage component, but only had Auth/Core installed)

I suppose we could change to use `aws-amplify` and have a single circular dependency instead of multiple packages, but would still need to make it a range (e.g. `3.x.x`):

**whatever we do, we need to ensure there is only _1_ copy of each `@aws-amplify/*` package in customer projects**.

wlee221: Sorry about the delay. I have applied comments from @ericclemmons. I'll test with verdaccio to check that we do not have multiple versions given modular import scenarios.

wlee221:
Running integration on https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/6509/workflows/781214eb-0d4a-4781-857b-3a1151ec661d. I will have to test for duplicate packages in modular import scenarios. eta. this Fri.

wlee221:
To do based on today's discussion: we should test that there are no duplicated packages when customers install divergent `ui-components` and `aws-amplify`. Also need to test this behavior with modular imports, although this is the discouraged behavior.

lerna.json
```diff
@@ -1,8 +1,7 @@
...
-  "npmClient": "yarn",
-  "useWorkspaces": true,
-  "packages": [
-    "packages/*"
-  ],
-  "version": "independent"
+	"npmClient": "yarn",
+	"useWorkspaces": true,
+	"packages": ["packages/*"],
+	"exact": true,
```
wlee221, line 5: This is equivalent to `--exact` tag.

wlee221:
Based on our offline discussion today, this is ready to be reviewed.

sammartinez (approved):
LGTM 🌮  Thanks for getting this change in @wlee221. Lets get @iartemiev and @ericclemmons to approve and we will be good to merge

iartemiev (approved):
This looks good to me. 
I appreciate all the thought you put into solutions to this problem 😸  great job!

packages/aws-amplify-angular/package.json
```diff
@@ -68,10 +68,10 @@
 		"zone.js": "^0.8.26"
 	},
 	"dependencies": {
-		"@aws-amplify/ui": "^2.0.2",
+		"@aws-amplify/ui": "2.0.2",
 		"rxjs-compat": "^6.2.1"
 	},
 	"peerDependencies": {
-		"aws-amplify": "^3.0.0"
+		"aws-amplify": "3.0.0"
```
ericclemmons, line 75: This can be `3.x.x`, right?  Otherwise every non `3.0.0` version will have a warning?

(I understand it's been this way since `3.0.0`)

packages/aws-amplify-react-native/package.json
```diff
@@ -31,7 +31,7 @@
 		"buffer": "^5.2.1"
 	},
 	"peerDependencies": {
-		"aws-amplify": "^3.0.0",
+		"aws-amplify": "3.0.0",
```
ericclemmons, line 34: Same Q about `3.x.x`

packages/aws-amplify-react/package.json
```diff
@@ -108,13 +108,13 @@
...
-		"@aws-amplify/ui": "^2.0.0",
-		"@aws-amplify/xr": "^2.0.0"
+		"@aws-amplify/analytics": "3.0.0",
+		"@aws-amplify/api": "3.0.0",
+		"@aws-amplify/auth": "3.0.0",
+		"@aws-amplify/core": "3.0.0",
+		"@aws-amplify/interactions": "3.0.0",
+		"@aws-amplify/storage": "3.0.0",
+		"@aws-amplify/ui": "2.0.0",
+		"@aws-amplify/xr": "2.0.0"
```
ericclemmons, line 118: These will always warn, won't they? Unless they're `N.x.x`?

ericclemmons (approved):
Great work @wlee221!

This will allow customers to "roll back" to a previous version if they have any issues. Plus, it'll allow us to reproduce issues more easily using their exact version!

**I added a few notes regarding `peerDependencies` for legacy packages (e.g. `aws-amplify-react`) to avoid warnings since they're unpinned.**

IIRC, `lerna` doesn't manage `peerDependencies`, so those can remain as `^3.0.0`? Otherwise can they be `3.x.x`?

I don't consider this a blocker, but possible noise when customers install...

packages/aws-amplify-angular/package.json
wlee221: Thanks for the catch! I'll change the peer deps back to range.

wlee221: Thanks for the catch! You're right, I'll change them to range again.

wlee221:
Thanks everyone for the help! Merging it in now.

PR #7044, opened 2020-10-23
fix(datastore): validate model fields to allow undefined

### LINKED ISSUE #5206: Deleting optional string doesn't work with Datastore
**Describe the bug**
I save image on S3 and the url for the image on Dynamodb with optional String. When I delete image I tried to set it to undefined and I get following error
`Field imageKey should be of type string, undefined received. undefined`

**Amplify CLI Version**
4.17.1

**To Reproduce**
1. `npx create-react-app test --use-npm --template typescript`
1. `amplify init`
1. `amplify add auth` default everything
1. `amplify add api` Then add following schema
```
type Product @model 
  @auth(rules:[ 
    {allow: owner}
  ]) {
  id: ID!
  owner: String!
  name: String!
  imageKey: String
}
```

Then I had image url saved in `imageKey` Now I deleted the image so I tried to set `imageKey` to undefined
```
const original = await DataStore.query(Product, id)

const results = await DataStore.save(
            Farmer.copyOf(original, updated => {updated.imageKey = undefined})
        )
```
I got following error
`Field imageKey should be of type string, undefined received. undefined`

**Expected behavior**
It should delete the key and set it as undefined 

 - OS: Ubuntu 18.04 with Windows WSL2
 - Node Version. 12.16.1

**Additional context**
I also tried with setting imageKey to '' string with nothing I also got another error. I will post it soon.

### LINKED ISSUE #6453: DataStore - Omitting an optional field from a model object results in error on save
**Describe the bug**
Given a GraphQL schema which defines a model type with an optional (string) value, if a model instance is saved without defining that value (it's optional after all), it will fail; definging a `null` value for the optional value succeeds.

**To Reproduce**
This was recently demonstrated to us when we had a graphql schema like the following:

```graphql
type Movie @model {
  id: ID!
  name: String!
  description: String
}
```

When calling something like:

```ts
await DataStore.save(
  new Movie({
    name: "Dumb and Dumber",
  })
);
```

The local DataStore appears to be updated but I see the following _error_ logged as a _warning_:

```
[WARN] 06:03.591 DataStore - failed to execute errorHandler, [Error: Field description should be of type string, undefined received. undefined]
```

**Expected behavior**
All optional fields should be able to be omitted when saved and it should succeed without populating that value. Alternatively, it can also be explicitly specified as `undefined` or `null` which is particularly useful on updates to clear out a field.

**Investigation**
A quick search of the codebase seems to show the culprit of this error [here](https://github.com/aws-amplify/amplify-js/blob/%40aws-amplify/datastore%402.2.8/packages/datastore/src/datastore/datastore.ts#L257).

Perhaps the following line can simply be changed from:

```diff
- } else if (typeof v !== jsType && v !== null) {
+ } else if (typeof v !== jsType && v != null) {
```

(e.g. it avoid strict equality on the `null` check to also include `undefined` and avoid erroring in that case?)

Earlier up in `validateModelFields()` it is already checking if the field is required:

```ts
if (isRequired && (v === null || v === undefined)) {
  throw new Error(`Field ${name} is required`);
}
```

So my suggested change should be sufficient? I don't know if there are other side-effects of this change that are undesired, or if the original exclusion of `undefined` was intentional, but hopefully not?

**Amplify Version**
2.2.8

### LINKED ISSUE #6744: DataStore failed to handle optional attributes
```gql
type A @model {
  id: ID!
  name: String!
  organisationID: ID # optional attribute
}
```

```
[WARN] 49:06.236 DataStore - failed to execute errorHandler Error: "Field organisationID should be of type string, undefined received. undefined"
    validateModelFields datastore.ts:260
    initializeInstance datastore.ts:275
    initializeInstance datastore.ts:274
    instance datastore.ts:288
    produce immerClass.ts:97
    Model datastore.ts:285
    modelInstanceCreator datastore.ts:226
    jitteredRetry mutation.ts:300
    step 46.chunk.js:97542
    verb 46.chunk.js:97472
    rejected 46.chunk.js:97434
```

wei:
_Issue #, if available:_
Fix #5206 
Fix #6453
Fix #6744 

_Description of changes:_
https://github.com/aws-amplify/amplify-js/issues/5206#issuecomment-714200927
`validateModelFields` currently throws an error when setting `undefined` to an optional field. This PR fixes this wrong behavior.

cc/ @iartemiev 

### DIFF
--- a/packages/datastore/__tests__/DataStore.ts
+++ b/packages/datastore/__tests__/DataStore.ts
@@ -195,6 +195,73 @@ describe('DataStore tests', () => {
 			expect(model1.id).toBe(model2.id);
 		});
 
+		test('Optional field can be initialized with undefined', () => {
+			const { Model } = initSchema(testSchema()) as {
+				Model: PersistentModelConstructor<Model>;
+			};
+
+			const model1 = new Model({
+				field1: 'something',
+				optionalField1: undefined,
+			});
+
+			expect(model1.optionalField1).toBeUndefined();
+		});
+
+		test('Optional field can be initialized with null', () => {
+			const { Model } = initSchema(testSchema()) as {
+				Model: PersistentModelConstructor<Model>;
+			};
+
+			const model1 = new Model({
+				field1: 'something',
+				optionalField1: null,
+			});
+
+			expect(model1.optionalField1).toBeNull();
+		});
+
+		test('Optional field can be changed to undefined inside copyOf', () => {
+			const { Model } = initSchema(testSchema()) as {
+				Model: PersistentModelConstructor<Model>;
+			};
+
+			const model1 = new Model({
+				field1: 'something',
+				optionalField1: 'something-else',
+			});
+
+			const model2 = Model.copyOf(model1, draft => {
+				(<any>draft).optionalField1 = undefined;
+			});
+
+			// ID should be kept the same
+			expect(model1.id).toBe(model2.id);
+
+			expect(model1.optionalField1).toBe('something-else');
+			expect(model2.optionalField1).toBeUndefined();
+		});
+
+		test('Optional field can be set to null inside copyOf', () => {
+			const { Model } = initSchema(testSchema()) as {
+				Model: PersistentModelConstructor<Model>;
+			};
+
+			const model1 = new Model({
+				field1: 'something',
+			});
+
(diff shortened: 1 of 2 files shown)


iartemiev (approved):
LGTM! Thank you, @wei!

PR #7047, opened 2020-10-23
fix(aws-amplify-react-native): auto-fill the user name field on ConfirmSignUp page

### LINKED ISSUE #6789: [aws-amplify-react-native] email field on ConfirmSignUp is not auto-filled after redirecting from SignUp page

Albert-Gao:
close #6789

_Issue #6789, if available:_

_Description of changes:_
auto-fill the user name field on ConfirmSignUp page


Albert-Gao:
close https://github.com/aws-amplify/amplify-js/pull/6790 and focus on this one to catch up with the main branch

eddiekeller:
Hey @Albert-Gao , thanks for the submission. It looks good to me, but I don't really know the context around this change. Can you explain it for me so that I better understand?

Albert-Gao:
@eddiekeller Sure, when using the Auth module, after signed up, the module will lead the user to the Confirm-Sign-Up page, the problem is the "email" / "username" field will be empty, so the user needs to type twice, which does not aligned with the other products out there.

with this PR being merged, 
- the user will see the "email"/"username" field being filled up with the registered value, unless the user refreshes the UI so we lose the data in the memory (but should not happen here, since it is react-native)
- also, this updates is scaleable, it is not hard-coded "email"/"username" field, it will aligned with whatever set as the main auth field from the `props.usernameAttributes`, so if the `props.usernameAttributes` will be something like `custom-ID`, this PR still works.

Just checked the current code, it still applies, since no changes to this file during this period.

eddiekeller:
Hey @Albert-Gao , thanks for that explanation! Honestly, this is a great idea and a welcome change.

I just kicked off a "merge from main". I'll be watching it and as long as no tests fail, I'll approve!

eddiekeller (approved):

eddiekeller:
I will merge soon - we have a deployment going out this morning so I am waiting until that is finished.

Albert-Gao:
Hi, thanks 😊 this is great!

PR #7050, opened 2020-10-23
Ui components/bug/fix button loading

cb372:
Fix a few minor bugs related to the behavior of buttons on forms while in the `loading` state:

* amplify-form-section: disable the submit button when `loading` is true
* amplify-confirm-sign-up: propagate the `loading` flag to amplify-form-section
* amplify-sign-up: set the loading flag before making the HTTP request, and clear it afterwards
* amplify-sign-up: disable the submit button when `loading` is true

Also removed a duplicate test case I found.

I tried extending the e2e test for amplify-form-section to add some assertions about the button's `disabled`, but I couldn't find a way to make the e2e tests run. It looks like they are somehow excluded but I couldn't find where that is configured. They don't appear to run on CircleCI either.

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.spec.ts
+++ b/packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.spec.ts
@@ -32,14 +32,5 @@ describe('amplify-form-section spec:', () => {
 
       expect(page.root).toMatchSnapshot();
     });
-
-    it('should render form section', async () => {
-      const page = await newSpecPage({
-        components: [AmplifyFormSection],
-        html: `<amplify-form-section></amplify-form-section>`,
-      });
-
-      expect(page.root).toMatchSnapshot();
-    });
   });
 });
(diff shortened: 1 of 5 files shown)


packages/amplify-ui-components/src/components/amplify-form-section/amplify-form-section.spec.ts
```diff
@@ -32,14 +32,5 @@ describe('amplify-form-section spec:', () => {
...
     });
-
-    it('should render form section', async () => {
-      const page = await newSpecPage({
-        components: [AmplifyFormSection],
-        html: `<amplify-form-section></amplify-form-section>`,
-      });
-
-      expect(page.root).toMatchSnapshot();
-    });
```
eddiekeller, line 43: Instead of just deleting the test, can you modify the tests to reflect the new behavior that you have implemented?

eddiekeller (changes requested):
Thanks for the submission! I like the idea. Obviously, since this PR is rather old now (sorry that nobody has reviewed it yet!!!), there are some conflicts. If you can address those conflicts and then address my comment regarding the test file, that would be awesome!

ericclemmons:
It may be best to resolve this PR given we have _some_ of this functionality supported:

https://github.com/aws-amplify/amplify-js/blob/dbe82d8296f36541b11abd3006f9ff25c98f2270/packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx#L217-L219

https://github.com/aws-amplify/amplify-js/blob/dbe82d8296f36541b11abd3006f9ff25c98f2270/packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx#L422-L432


What do you think @eddiekeller ?

Issue #7055, opened 2020-10-24
withAuthenticator does not pass on props to wrapped component


simonxabris:
**Describe the bug**

In NextJS when you're exporting a page component with the `withAuthenticator` HOC, it does not receive the props from `getServerSideProps`, because `withAuthenticator` does not pass them along. This prevents the usage of the `withAuthenticator` component in server rendered pages.

The problem is on [this line](https://github.com/aws-amplify/amplify-js/blob/958f61ef2918728cc46b9b210d60e868edd87f39/packages/amplify-ui-react/src/withAuthenticator.tsx#L55), if it was:

```jsx
return <Component {...props} />
``` 

it would work just fine. I'll submit a PR shortly.

I did not provide any additional info from the issue template because I could identify where the problem comes from, so I believe there's no debugging needed from the maintainers. However If you need the info I'm happy to edit the issue and provide them.

UPDATE:

I've done some digging and I found out that `useEffect` does not run in server side environments, so even if the props are passed along, the markup that will be server rendered is the authenticator markup. This leads me to think that `withAuthenticator` is not intended to be used in server rendered pages. If this is the case, could this be included in the docs somewhere? Maybe [this dev.to post](https://dev.to/dabit3/the-complete-guide-to-next-js-authentication-2aco) from @dabit3 could be turned into a page in the official docs.

harrysolovay:
You can use the produced component alongside a context:

```tsx
import React from "react";

const propsCtx = React.createContext();

const AuthenticatedPage = withAuthenticator(MyPageComponent);

export default function MyPage(props) {
  return (
    <propsCtx.Provider value={props}>
      <AuthenticatedPage />
    </propsCtx.Provider>
  );
}
```

Then, within `MyPageComponent`, you can make use of the `propsCtx`.

```tsx
function MyPageComponent() {
  const props = React.useContext(propsCtx);

  return (
    <div>{JSON.stringify(props)}</div>
  );
}
```

PR #7056, opened 2020-10-24
pass on props from withAutenticator to wrapped component

simonxabris:
Issue #7055

Description of changes:

With this change, the `withAuthenticator` now passes along the props that were intended for the wrapped component. This makes it possible to use it for example in nextjs page routes which are server rended using `getServerSideProps`

Note: I've tried to add a unit test that checks this behavior, but I could not get the tests working. What I've found is that this effect https://github.com/aws-amplify/amplify-js/blob/958f61ef2918728cc46b9b210d60e868edd87f39/packages/amplify-ui-react/src/withAuthenticator.tsx#L20-L25 never runs in tests and since the initial `signedIn` value is `false`, it will always render the markup for the authenticator.

### DIFF
--- a/packages/amplify-ui-react/src/withAuthenticator.tsx
+++ b/packages/amplify-ui-react/src/withAuthenticator.tsx
@@ -10,23 +10,23 @@ import { Logger } from '@aws-amplify/core';
 
 const logger = new Logger('withAuthenticator');
 
-export function withAuthenticator(
-	Component: ComponentType,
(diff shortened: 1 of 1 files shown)


harrysolovay:
I agree with you––it would be nice to provide a generic way for users to specify additional props for the `withAuthenticator`-produced component. And thank you so much for this contribution. Unfortunately, this would be breaking to current users. Closing for now, although we'll discuss and consider for the next major release.

PR #7111, opened 2020-11-03
feat: enable hub listening of token refresh & refresh failure events

harrysolovay:
_Issue #, if available:_ #6352 

A cleaner implementation to that of #6983 

_Description of changes:_

```ts
Hub.listen('auth', (data) => {
  switch (data.payload.event) {
    case 'tokenRefresh': {
      logger.log("New token has been process");
      break;
    }
    case 'tokenRefresh_failure': {
      logger.log("Failed to retrieve new token");
      break;
    }
  }
});
```

### DIFF
--- a/packages/amazon-cognito-identity-js/src/CognitoUserPool.js
+++ b/packages/amazon-cognito-identity-js/src/CognitoUserPool.js
@@ -35,7 +35,7 @@ export default class CognitoUserPool {
 	 *        to support cognito advanced security features. By default, this
 	 *        flag is set to true.
 	 */
-	constructor(data) {
+	constructor(data, wrapRefreshSessionCallback) {
 		const {
 			UserPoolId,
 			ClientId,
@@ -64,6 +64,10 @@ export default class CognitoUserPool {
 			AdvancedSecurityDataCollectionFlag !== false;
 
 		this.storage = data.Storage || new StorageHelper().getStorage();
+
+		if (wrapRefreshSessionCallback) {
+			this.wrapRefreshSessionCallback = wrapRefreshSessionCallback;
+		}
 	}
 
 	/**
@@ -143,9 +147,7 @@ export default class CognitoUserPool {
 	 * @returns {CognitoUser} the user retrieved from storage
 	 */
 	getCurrentUser() {
-		const lastUserKey = `CognitoIdentityServiceProvider.${
-			this.clientId
-		}.LastAuthUser`;
+		const lastUserKey = `CognitoIdentityServiceProvider.${this.clientId}.LastAuthUser`;
 
 		const lastAuthUser = this.storage.getItem(lastUserKey);
 		if (lastAuthUser) {
--- a/packages/auth/__tests__/auth-unit-test.ts
+++ b/packages/auth/__tests__/auth-unit-test.ts
@@ -1,5 +1,6 @@
 import OAuth from '../src/OAuth/OAuth';
 import * as oauthStorage from '../src/OAuth/oauthStorage';
+import { NodeCallback } from 'amazon-cognito-identity-js';
 
 jest.mock('crypto-js/sha256', () => {
 	return {
@@ -238,6 +239,7 @@ import {
 } from '@aws-amplify/core';
 import { AuthError, NoUserPoolError } from '../src/Errors';
 import { AuthErrorTypes } from '../src/types/Auth';
+import { reject } from 'q';
 
 const authOptions: AuthOptions = {
 	userPoolId: 'awsUserPoolsId',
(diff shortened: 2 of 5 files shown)


packages/auth/src/Auth.ts
```diff
@@ -245,6 +249,22 @@ export class AuthClass {
...
 	}
 
+	private wrapRefreshSessionCallback = (callback: NodeCallback.Any) => {
+		const wrapped: NodeCallback.Any = (error, data) => {
+			if (data) {
+				dispatchAuthEvent('token_refresh', undefined, `New token retrieved`);
+			} else {
+				dispatchAuthEvent(
+					'token_refresh_failure',
+					error,
```
elorzafe, line 259: What is the data on that error?

harrysolovay: The error that was thrown within the refresh session call (passed to the callback)

harrysolovay:
@elorzafe I added tests

packages/amazon-cognito-identity-js/src/CognitoUserPool.js
```diff
@@ -35,7 +35,7 @@ export default class CognitoUserPool {
 	 *        to support cognito advanced security features. By default, this
 	 *        flag is set to true.
 	 */
-	constructor(data) {
+	constructor(data, wrapRefreshSessionCallback) {
```
ericclemmons, line 38: IIRC, the design review had the constructor accepting a more config-like structure.

I see above that this argument is being used to wrap a callback:

```js
const wrappedCallback = this.pool.wrapRefreshSessionCallback
			? this.pool.wrapRefreshSessionCallback(callback)
			: callback
```

Can you help me understand the need to "reach in" to the pool instance to wrap the callback?
Ignoring the callback-hell problem, I'm trying to understand what events are happening that can't be handled other ways (since this `constructor` signature would be hard to change later).

ericclemmons (commented):
Thanks for adding tests! This looks like it should work!

I just wanted to understand the underlying control flow difficulties.

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -696,6 +698,57 @@ describe('auth unit test', () => {
...
+			// saving a reference to this fn's return before triggering `wrappedCallback` lets us capture the payload
+			const captureEvent = () => {
+				return new Promise(resolve => {
+					Hub.listen('auth', capsule => {
+						switch (capsule.payload.event) {
+							case 'token_refresh': {
+								return resolve(true);
+							}
+
+							case 'token_refresh_failure': {
```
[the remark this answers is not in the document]
harrysolovay: Good point. No strong preference on naming. I’ll leave that to you & @elorzafe’s discretion.

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -238,6 +239,7 @@ import {
 } from '@aws-amplify/core';
 import { AuthError, NoUserPoolError } from '../src/Errors';
 import { AuthErrorTypes } from '../src/types/Auth';
+import { reject } from 'q';
```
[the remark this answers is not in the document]
harrysolovay: Must have auto-imported. Good catch!

packages/amazon-cognito-identity-js/src/CognitoUserPool.js
harrysolovay: Sure. We want token refresh request responses to be followed by an event emission (success or failure). If there’s success, the callback is called with a second argument supplied (the token data, which is NOT included in the event payload). Otherwise, it’s called with only the first (an error, which is included in the payload). Wrapping the callback lets us weave the effect into the preexisting callback flow. To your other point about updating constructor params: we have to make that callback-wrapping fn accessible to amazon-cognito-identity-js, and adding the second constructor param (optional) is non-breaking. I’m not concerned about its evolvability. It needs to be rewritten. & soon. In fact, I’d argue it’s a crucial next step for the longevity of the auth category. I’ve implemented several approaches and I believe this one to be the safest.

ericclemmons: That makes sense. I agree with all of that.

So it sounds like there's not really any alternative architecture to support this flow _without_ reaching into the pool with `this.pool.wrapRefreshSessionCallback`?

harrysolovay: That is correct, as the refresh logic is happening internally in the `amazon-cognito-identity-js` package.

manueliglesias (approved):
LGTM 👍

ericclemmons (approved):
@harrysolovay Thanks for getting this taken care of for customers!

Thanks for your thoughtful feedback: I understand that "reaching in" to `CognitoUserPool` isn't desirable, but the current architecture necessitates it without breaking encapsulation further.

In the *very* unlikely event the `constructor` will need to change, we'll be able to use duck-typing to get around it.

But, as you've rightfully called out: these auth implementations need refactoring to account for customers' needs today.

Thanks again @harrysolovay! Great work!

packages/auth/__tests__/auth-unit-test.ts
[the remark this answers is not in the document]
elorzafe: I like Eric suggestion

elorzafe (approved):
Thanks @harrysolovay !!! Nice work 🎉 🌮 🥇 

I would only change the name of the error to camelCase_error

`tokenRefresh_failure`

packages/auth/src/Auth.ts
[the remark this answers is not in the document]
harrysolovay: Will do

ericclemmons:
@harrysolovay Only a few things left and we should be all set!

- [ ] The `tokenRefresh` casing, based on mine & @elorzafe's suggestion.
- [ ] Update the title to match conventional commits.  This would constitute a minor bump, since it adds a feature. (@iartemiev do you agree or disagree?) @iartemiev What should it be to bump both `auth` and `amazon-cognito-identity-js`?

ericclemmons (approved):
Great work @harrysolovay ! Thanks for the insights into your implementation, the limitations our current architecture has, and refining this with the input of the team.

🚢 !

PR #7121, opened 2020-11-05
fix(@aws-amplify/ui-angular) catch unhandled promise from `currentAuthenticatedUser`

### LINKED ISSUE #6530: Unhandled Promise rejection when adding in <amplify-authenticator></amplify-authenticator>
**Describe the bug**
When adding the <amplify-authenticator></amplify-authenticator> component into the HTML for an Angular App, on the page load it results in:
`zone-evergreen.js:659 Unhandled Promise rejection: not authenticated ; Zone: <root> ; Task: Promise.then ; Value: not authenticated undefined`
Chrome, Firefox, Safari (OSX) and Edge (new version) can handle this and carry on so you are still able to log in, but on Edge (legacy version) and Safari on mobile it prevents the login UI loading making the application useless.

If the HTML component is not there this error does not occur

**To Reproduce**
Simplest reproduction, follow this tutorial - [https://docs.amplify.aws/ui/auth/authenticator/q/framework/angular](url)
Add the following into polyfill.ts
`(window as any).global = window;
(window as any).process = {
  env: { DEBUG: undefined },
};`
Note you will need to create the aws-exports.js file

Then do ng serve and open the console and there is the error. In Chrome you would still be able to login but IE / Legacy Edge the UI does not even load.

See Stackblitz for example app with minimal setup.
[https://stackblitz.com/edit/angular-ivy-mkqpwz](url)

**Expected behavior**
There should not be a promise rejection, or it should be handled correctly so it does not stop the login components from working.

**Code Snippet**
https://stackblitz.com/edit/angular-ivy-mkqpwz

**What is Configured?**
Angular project using Angular 10, following package.json
`  "dependencies": {
    "@angular/animations": "~9.0.6",
    "@angular/common": "~9.0.6",
    "@angular/compiler": "~9.0.6",
    "@angular/core": "~9.0.6",
    "@angular/forms": "~9.0.6",
    "@angular/platform-browser": "~9.0.6",
    "@angular/platform-browser-dynamic": "~9.0.6",
    "@angular/router": "~9.0.6",
    "@aws-amplify/ui-angular": "^0.2.14",
    "aws-amplify": "^3.0.23",
    "rxjs": "~6.5.4",
    "tslib": "^1.10.0",
    "zone.js": "~0.10.2"
  },
  "devDependencies": {
    "@angular-devkit/build-angular": "~0.900.6",
    "@angular/cli": "~9.0.6",
    "@angular/compiler-cli": "~9.0.6",
    "@angular/language-service": "~9.0.6",
    "@types/node": "^12.11.1",
    "@types/jasmine": "~3.5.0",
    "@types/jasminewd2": "~2.0.3",
    "codelyzer": "^5.1.2",
    "jasmine-core": "~3.5.0",
    "jasmine-spec-reporter": "~4.2.1",
    "karma": "~4.3.0",
    "karma-chrome-launcher": "~3.1.0",
    "karma-coverage-istanbul-reporter": "~2.1.0",
    "karma-jasmine": "~2.0.1",
    "karma-jasmine-html-reporter": "^1.4.2",
    "protractor": "~5.4.3",
    "ts-node": "~8.3.0",
    "tslint": "~5.18.0",
    "typescript": "~3.7.5"
  }`
Everything was configured by the Amplify CLI, for a minimal setup these steps can be skipped as the issue is loading the HTML component that triggers the error.  
 
**Desktop:**
- OS: Windows 10, OSX
- Browser: Latest versions of Chrome, Firefox, Safari, Edge and legacy Edge

Any help would be greatly appreciated 👍

wlee221:
_Issue #, if available:_ Fixes #6530, Related: [angular#35873](https://github.com/angular/angular/issues/35873)

_Description of changes:_ This catches unhandled promise thrown by `Auth.currentAuthenticatedUser`.
On `latest` tag:

Screen Shot 2020-11-05 at 1 02 15 AM (2)

on `pr-6530` tag:

Screen Shot 2020-11-05 at 12 59 31 AM (2)


packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -88,30 +88,31 @@ export class AmplifyAuthenticator {
...
-        logger.debug('Failed to get the auth state from local storage', error);
-      }
-      try {
-        if (cachedAuthState === AuthState.SignedIn) {
-          await Auth.signOut();
+    return Auth.currentAuthenticatedUser()
+      .then(user => {
+        dispatchAuthStateChangeEvent(AuthState.SignedIn, user);
+      })
+      .catch(async () => {
```
wlee221, line 100: I am not sure why `.then(...).catch(...)` pattern catches the unhandled promise opposed to using `try catch` with `async await` pattern, but I have verified this fix to work. I suspect that this is related to how zone.js (used by angular cli) handles unhandled promises. I could not find other workaround though, so we should revisit this after we upgrade our @ui-angular library and see if this behavior persists.

packages/auth/src/Auth.ts
```diff
@@ -1281,7 +1281,7 @@ export class AuthClass {
 					);
 				}
 				logger.debug('The user is not authenticated by the error', e);
-				throw 'not authenticated';
+				return Promise.reject('The user is not authenticated');
```
wlee221, line 1284: Unrelated to this fix, but wanted to make this message more user-friendly.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -88,30 +88,31 @@ export class AmplifyAuthenticator {
...
+          logger.debug('Failed to get the auth state from local storage', error);
         }
-        dispatchAuthStateChangeEvent(this.initialAuthState);
-      } catch (error) {
-        logger.debug('Failed to sign out', error);
-      }
-    }
+        try {
+          if (cachedAuthState === AuthState.SignedIn) {
+            await Auth.signOut();
```
ericclemmons, line 109: Does zone.js throw a similar error if this fails?

(You can force it with `throw new Error('test')` to see)

I'm wondering if the entire `checkUser` function shouldn't be `async` anymore.

wlee221: Yes, I have confirmed the same behavior with `Auth.signOut`, or any errors thrown in general. For example,

```javascript
export class AmplifyAuthenticator {
  throwError() {
    return Promise.reject('test');
  };
  async componentWillLoad() {
    try {
      this.throwError();
    } catch (err) {
      console.log('error caught');
    }
  }
}
```
logs `unhandled promise error` although it successfully enters the catch block. 

So we really need to change `async/await` pattern across all `ui-components` to fully address behavior.  Instead, I think we should just catch  `currentAuthenticatorUser` because it throws by default and creates noise in customer apps. As a long-term fix, I can investigate further when we update our ui libraries for latest frameworks -- and create an issue in angular repo if this persists. How does that sound @ericclemmons?

ericclemmons: @wlee221 I agree.

Let's get this merged in to solve the noise problem. But you're correct: **this is buggy behavior in Angular that they should resolve.**

Can you cross-link the zone.js issue in this thread for future Googlers, too?

ericclemmons (approved):
Thanks for getting this resolved @wlee221!

There may be _some_ risk to changing the error message (users often treat error messages as contracts), but I agree having a clearer message is worth it.

wlee221:
Thanks @ericclemmons!

Issue #6479, opened 2020-08-03
Add slot name to `amplify-toast`


nkint:
**Is your feature request related to a problem? Please describe.**
I'd like to add custom style to toast notification. I've seen other issue about this, like #5604 #6167 #2834

**Describe the solution you'd like**
Fully customize toast notification via `slot` as other ui components do.

I think it's matter of few lines in this file: https://github.com/aws-amplify/amplify-js/blob/3642e6af455ee1848c1cc743a7a503e96baefa41/packages/amplify-ui-components/src/components/amplify-toast/amplify-toast.tsx

```
render() {
    return (
      <slot name="toast-notification">
        <div class="toast">
          <amplify-icon class="toast-icon" name="warning" />
          {this.message ? <span>{this.message}</span> : null}
          <slot />
          <button class="toast-close" onClick={this.handleClose} />
        </div>
      </slot name="toast-notification">
    );
  }
```

Could you consider a PR?

ashika01:
@nkint If you are open to getting a PR out I can work with you to get it merged in.

ashika01:
@nkint In your case, I think adding `slot` wouldn't ideally fix the problem. Since all the error message for toast comes from within amplify, having a slot would just override that. Rather if styling is what you are after, could you explain your style use case deeper so we could see if we can incorporate that into `css variables`.

nkint:
Toast notification are really important part of the ui: feedback against user actions. Due to its importance is something really hard to just incorporate in a css variables and override it. Moreover, notifications can go throughout all my application (not only where amplify is involved).

What about using some other library (in #2834 they told about [react-notification-alert](https://www.npmjs.com/package/react-notification-alert))?

I have access to the error anyway, let's take React as example:
```js
	React.useEffect(() => {
		Auth.currentAuthenticatedUser()
			.then((user) => updateUser(user))
			.catch(() => console.log('No signed in user.'));
		Hub.listen('auth', (data) => {

			console.log({ data });  // <------------ here I have error payload anyway

			switch (data.payload.event) {
				case 'signIn':
					return updateUser(data.payload.data);
				case 'signOut':
					return updateUser(null);
				default:
					return null;
			}
		});
	}, []);
```

And I can override `amplify-toast` hiding it and using my own feedback notification library as I prefer.

ashika01:
[...]

PR #6098, opened 2020-06-16
Add hideSignUp prop to sign in

jordanranz:
_Issue #, if available:_
fixes: https://github.com/aws-amplify/amplify-js/issues/6045, https://github.com/aws-amplify/amplify-js/issues/6089, https://github.com/aws-amplify/amplify-js/issues/6091, https://github.com/aws-amplify/amplify-js/issues/5790, https://github.com/aws-amplify/amplify-js/issues/5594

_Description of changes:_
- Add `hideSignUp` prop to `amplify-sign-in` component

Usage:
React:
```jsx
<AmplifyAuthenticator>
      <AmplifySignIn slot="sign-in" hideSignUp />
</AmplifyAuthenticator>
```
Other frameworks:
```
<amplify-authenticator>
    <amplify-sign-in slot="sign-in" hide-sign-up></amplify-sign-in>
</amplify-authenticator>
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7129, opened 2020-11-06
add hideToast prop to authenticator

Luke-Davies:
_Issue #, if available:_
None

_Description of changes:_
Adds a property to the `amplify-authenticator` called ~showToasts~`hideToast` which is defaulted to `false`.

Setting this to `true` will prevent `amplify-toast` from rendering when `TOAST_AUTH_ERROR_EVENT` events are raised on the `UI_AUTH_CHANNEL`.

This is useful if you want to roll your own alerting components for auth error messages.

React Example:

```
import React, { useState, useEffect } from 'react';
import { Hub, HubCallback } from '@aws-amplify/core';
import {
  AmplifyAuthenticator,
} from '@aws-amplify/ui-react';
import {
 UI_AUTH_CHANNEL, TOAST_AUTH_ERROR_EVENT
} from '@aws-amplify/ui-components';

const MyAuth: React.FC = ({ children }) => {

  const [alertMessage, setAlertMessage] = useState('');

  const handleToastErrors: HubCallback = ({ payload }) => {
    if (payload.event === TOAST_AUTH_ERROR_EVENT && payload.message) {
      setAlertMessage(payload.message);
    }
  };

  useEffect(() => {
    Hub.listen(UI_AUTH_CHANNEL, handleToastErrors);
    return () => Hub.remove(UI_AUTH_CHANNEL, handleToastErrors);
  });

  return (
    <>
      {alertMessage && (<div>{alertMessage}</div>)}
      <AmplifyAuthenticator hideToast>
        // ...........
      </AmplifyAuthenticator>
    </>
  );
}
```


ThrowingSpoon:
Could this get reviewed please? Would be awesome to have 💪

Luke-Davies:
Not sure if I'm supposed to randomly bug people for reviews sorry! @elorzafe @sammartinez @manueliglesias 

This is a pretty simple change.

Could argue that it should be `hideToasts` instead of `showToasts` to match [this prop in amplify-sign-in](https://github.com/aws-amplify/amplify-js/blob/main/packages/amplify-ui-components/src/components/amplify-sign-in/amplify-sign-in.tsx#L67). Opinions welcome.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -47,6 +47,8 @@ export class AmplifyAuthenticator {
   @Prop() usernameAlias: UsernameAliasStrings;
   /** Callback for Authenticator state machine changes */
   @Prop() handleAuthStateChange: AuthStateHandler = () => {};
+  /** Show amplify-toast for auth errors */
+  @Prop() showToasts: boolean = true;
```
wlee221, line 51: As you mentioned, I like `hideToasts = false` better because users would only use this prop when they want to hide them. And you can do `<amplify-authenticator hide-toasts /> too!

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -207,7 +209,7 @@ export class AmplifyAuthenticator {
   render() {
     return (
       <Host>
-        {this.toastMessage ? (
+        {this.showToasts && this.toastMessage ? (
```
wlee221, line 212: Instead of mixing `&&` and `?`, can we use `&&` for both?
```ts
this.showToasts && this.toastMessage && (
  // ...
)
```

wlee221 (changes requested):
Sorry about the late review! Just got some nits but looks good to me otherwise 👍

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
wlee221: On second thought, I think we should use `hideToast` because there is only one toast at a time.

Luke-Davies:
Thanks @wlee221! 
Both changes made and tested locally.
I've updated the PR name & description to reflect the prop change from `showToasts` --> `hideToast`

ericclemmons:
Thanks for taking this need on @Luke-Davies.

I support @wlee221's feedback, and wanted to offer some context into past discussions around this need.

The UI components are, as we know, powered by Web Components and often rely on "slots" for customizing individual components.

What we _didn't know_ at the time is if `hideToast` would be equivalent to overriding an `amplify-toast` slot (if it existed):

```js
<AmplifyAuthenticator>
  <!-- Default components here -->
  <div slot="ampify-toast">
    <!-- Nothing to see here! -->
  </div>
</AmplifyAuthenticator>
```

Do either of y'all know if this is better, worse, or possible?  IIRC, the rationale for `hideSignUp` was because the sign-up language was "too deep" in the component to replace with an empty `slot`:
https://github.com/aws-amplify/amplify-js/pull/6098/files

Again, I'm supportive of this PR to unblock customers! I just wanted to share _where we left off_ in conversations in the past.

Luke-Davies:
Thanks for the context @ericclemmons. I had considered trying to implement this feature by introducing a slot and would absolutely support that solution if it is more in-keeping with the design choices made for the project so far.

If a slot is introduced then I would suggest something other than `amplify-toast` for the name.

For example; our app displays a modal for auth errors because we found that some customers were not seeing the toast on certain devices unless they happened to scroll to the top of the page. Currently, we're showing this modal as well as the toast (until we can hide the toast).

Perhaps `amplify-auth-alert` or just `amplify-alert` would be a better name for the slot if that approach is preferred?

ericclemmons:
I agree, `amplify-toast` isn't great. It matches the original desire of a "toast" notification, but we also considered `amplify-notification` among others.

I just found https://github.com/aws-amplify/amplify-js/issues/6479 as well, so we're not the only ones who've talked about this :D

So, in terms of preference having _fewer props_ would be preferential. Especially since a `slot` would allow _customization_ compared to hiding entirely.

@wlee221 What are your thoughts on the matter? Do you prefer props vs. slots, and what name would you use? (Assuming it would work!)

wlee221:
I can't believe I missed the slot approach! I just tested that using slots works as intended, so this is really up to what we want to do.

I'm kind of divided on slots vs props though. If it were just for hiding toast, I would have argued that `hideToast` is more intuitive for customers. I find it unintuitive to pass an empty element to slot to hide a section:
```
<amplify-authenticator>
  <div slot="amplify-toast"></div>
</amplify-authenticator>
```
And it's not clear what the code is trying to do on first glance. I think the purpose of hiding an element deviates from customization, which slot specializes in. 

That being said, customers are asking for customizations on amplify-toast as well on #6479. Slots do well for covering both use cases, so I think slot is the best way to go. And I agree with @ericclemmons' comment that fewer props are preferential. 

So tldr; we should go with slots. For the slot name, I like `amplify-alert` or just `alert` because it conveys that there's an error better than other suggestions imo.

Luke-Davies:
@wlee221 @ericclemmons 
I've raised a new PR for the slotted approach. Please see PR #7601 

I believe we can close one.

Luke-Davies:
Coming back to this following a discussion on #7601 (`slot` approach). To understand why we've come back to `props`, see [this thread](https://github.com/aws-amplify/amplify-js/pull/7601#discussion_r588767972).

Luke-Davies:
@wlee221 One final change I wasn't sure of on this was whether to wrap `Hub.listen(UI_AUTH_CHANNEL, this.handleToastEvent);` and `Hub.remove(UI_AUTH_CHANNEL, this.handleToastEvent);` with `if (!this.hideToast)`. The `prop` could technically be updated though (I'm not sure why anyone would) which could result in the callback not being removed.

Let me know if you think it's worth doing.

wlee221:
I think this is a pretty simple optimization worth doing. As mentioned, I don't see a good reason for changing the `hideToast` value dynamically.

wlee221 (approved):
This LGTM 👍  Thanks a lot for your contribution throughout this!

Issue #7102, opened 2020-11-01
Hub 'auth' channel 'signIn' event is not invoked if account has temporary password.


ArthurAmes:
**Describe the bug**
The 'signIn' event is not invoked on the 'auth' channel when using Hub.listen('auth', () => { }) if the user account being used to sign in has the 'FORCE_CHANGE_PASSWORD' status

**To Reproduce**
1. Register a new listener: `Hub.listen('auth', data => {console.log(data.payload.event})`
2. Create a new user using the Amazon Cognito Admin panel; notice this new user's account status is 'FORCE_CHANGE_PASSWORD'
3. Auth.signIn() will not invoke the log if the admin-created user is used

**Expected behavior**
The event 'signIn' should be invoked on the 'auth' listener channel

**Versions**
Amplify version 4.32.0
Node version 14.4.0

wei:
@ArthurAmes You didn't specify which framework so I tested on React. Could not reproduce, I am able to receive a signIn event after password change. 

Please try starting fresh with the latest packages.
```
"aws-amplify": "^3.3.7",
"@aws-amplify/ui-react": "^0.2.27",
```

![image](https://user-images.githubusercontent.com/5880908/98485553-468ff900-21e5-11eb-9be6-20b0576392fc.png)

![image](https://user-images.githubusercontent.com/5880908/98485186-81446200-21e2-11eb-87a5-344fce35b646.png)

The hub event is being triggered correctly after new password is confirmed:
https://github.com/aws-amplify/amplify-js/blob/5173a9911096627ce1b45067808af249668b260b/packages/auth/src/Auth.ts#L959

This is the hub event for normal users with `CONFIRMED` status signing in:
https://github.com/aws-amplify/amplify-js/blob/5173a9911096627ce1b45067808af249668b260b/packages/auth/src/Auth.ts#L496-L501

@mauerbac @amhinson 
I did notice one thing though:
`${user} has signed in` is being rendered as `[object Object]` should probably change to match the message from the second one `A user ${user.getUsername()} has been signed in`  as `user` is of type `CognitoUser`

amhinson:
> I did notice one thing though:
> `${user} has signed in` is being rendered as `[object Object]` should probably change to match the message from the second snippet `A user ${user.getUsername()} has been signed in` as `user` is of type `CognitoUser`

Yep, good catch @wei! That seems like an appropriate update to make 👍

wei:
[...]

PR #7151, opened 2020-11-09
fix(@aws-amplify/auth): fix signIn event message

wei:
_Issue #, if available:_
Relates to #7102 

_Description of changes:_
`${user} has signed in` is being rendered as `[object Object] has signed in` because user is of type `CognitoUser`

It should say `A user ${user.getUsername()} has been signed in`, as seen in:
https://github.com/aws-amplify/amplify-js/blob/5173a9911096627ce1b45067808af249668b260b/packages/auth/src/Auth.ts#L496-L500

Screenshot:

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -935,7 +935,11 @@ export class AuthClass {
 						} finally {
 							that.user = user;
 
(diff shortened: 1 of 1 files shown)


amhinson (approved):
Thanks @wei 👍

Issue #6369, opened 2020-07-20
[Feedback] `API.graphql` return signature is hard to use in TypeScript


swyxio:
**Is your feature request related to a problem? Please describe.**

When I try to `await` an `API.graphql` call in TS, I get this:

![image](https://user-images.githubusercontent.com/6764957/87945385-3b4c7c80-cad3-11ea-8320-9accacbb42ee.png)

This is because you have defined the `graphql` method to also return an `Observable` in case of a subscription:

https://github.com/aws-amplify/amplify-js/blob/b5d14124dfd5f1000f102ac8c84aa964c9d916c5/packages/api/src/API.ts#L173-L178

This is understandable but makes it very annoying to use, particularly because if i need to assert types, the `GraphQLResult` type is not also exported for me to use.

**Describe the solution you'd like**

use overloading and/or a different function for subscriptions to make this experience easier.

ultimately i want to be able to write this:

```ts
const query = await API.graphql<TodoType[]>({query: listTodos})
```

and it should not complain. (a *REALLY* good experience, ofc, would give me that return type for free based on `listTodos`, but that would involve a much deeper redesign)

sw-yx:
Secondly, specifying the return signature like this: 

https://github.com/aws-amplify/amplify-js/blob/b5d14124dfd5f1000f102ac8c84aa964c9d916c5/packages/api/src/API.ts#L173-L178

always assigns the generic type `T` to be `object` because that is how the `GraphQLResult` interface is designed. This causes nasty, nasty type errors when people try to assign the results of the graphql query to a properly typed variable:

![image](https://user-images.githubusercontent.com/6764957/87951748-8d919b80-cadb-11ea-9434-da83aa987d26.png)

i would recommend making `graphql` a generic as well so I can pass along the type i want:

```ts
// pseudocode, modified from source slightly
	graphql<T>(
		options: GraphQLOptions,
		additionalHeaders?: { [key: string]: string }
	): Promise<GraphQLResult<T>> | Observable<object> {
		return this._graphqlApi.graphql(options, additionalHeaders);
	}
```

sw-yx:
for those finding this.. heres how to write a vaguely ergonomic wrapper around this.

```ts
import { GraphQLOptions} from '@aws-amplify/api-graphql'
async function gql<T extends object>(options: GraphQLOptions, additionalHeaders?: {
  [key: string]: string;
}): Promise<GraphQLResult<T>> {
[...]

PR #7155, opened 2020-11-10
use conditional type for `graphql`

yaquawa:
I had a hard time to cast the result of `graphql` in ts.

This PR make it use the conditional type instead of union type, so we can:
```ts
interface MyData {
    foo: string
    bar: number
}

async getFoo() {
    return (
        await API.graphql<'query', MyData>(
            graphqlOperation(getUserMeta, { id: userMetaId })
        )
    ).data.foo
}
```


sw-yx:
clever! would be a solution for #6369

yaquawa:
Takes too long time to be merged,  just turned to use https://github.com/awslabs/aws-mobile-appsync-sdk-js#readme for querying instead.

Issue #7084, opened 2020-10-29
How can you get the next 1000 records in the S3 bucket?


adsk-alexand:
https://github.com/aws-amplify/amplify-js/blob/958f61ef2918728cc46b9b210d60e868edd87f39/packages/storage/src/providers/AWSS3Provider.ts#L395

PR #7183, opened 2020-11-13
feat(@aws-amplify/storage): S3 listObject enhancements

### LINKED ISSUE #7133: Storage Enhancement: provide pagination for returning records
**Problem**
A customer (https://github.com/aws-amplify/amplify-js/discussions/7084) found that S3 will return only 1000 records via the list API. While this is true there are options that we can add to our S3 Provider in order to get the NextKeyMarker from S3 when the data IsTruncated. Reference to the S3 SDK (https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#listObjects-property)

**Solution**

We can look to add additional parameters to our params object within our AWSS3Provider.ts (https://github.com/aws-amplify/amplify-js/blob/main/packages/storage/src/providers/AWSS3Provider.ts#L409) file. Along with adding a while() loop around our s3.send(...) call (https://github.com/aws-amplify/amplify-js/blob/main/packages/storage/src/providers/AWSS3Provider.ts#L418).

**Output**

* Look to add additional parameters from S3 SDK
* Create a while loop when the list is truncated that allows to keep ammending to the array.
* Unit tests
* Integ tests

**References**
* https://github.com/aws-amplify/amplify-js/blob/main/packages/storage/src/providers/AWSS3Provider.ts#L395-L419
* https://github.com/aws-amplify/amplify-js/discussions/7084
* https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#listObjects-property

wei:
_Issue #, if available:_
Closes #7133
Addresses #7084

_Description of changes:_
This PR upgrades `Storage.list` from `ListObjects` to `ListObjectsV2` and provides a way for users to page through S3 object listings through the use of `ContinuationToken` & `NextContinuationToken` provided by AWS.S3 JavaScript SDK. This change is backwards compatible as none of data returned are changed.

Sample code on how to use it to page through all records, note `track: true` is required in order for Hub events to fire in `storage` channel:

```javascript
let fullList = [];

let loadMore = async (continuationToken) => {
  fullList = fullList.concat(await Storage.list('', { continuationToken, track: true }));
}

Hub.listen('storage', ({ payload }) => {
  const { event, data } = payload;
  if (event === 'list') {
    const { method, result, isTruncated, nextContinuationToken } = data.attrs;
    if (method === 'list' && result === 'success') {
      if (isTruncated) {
        loadMore(nextContinuationToken);
      }
    }
  }
});

loadMore();
```

`continuationToken` can be used in conjunction with the already existing `maxKeys` option to control the number of records to return per page (up to 1000). 

[Example repo](https://github.com/wei/aws-amplify-react/tree/7133) branch `7133` [App.js](https://github.com/wei/aws-amplify-react/blob/7133/src/App.js) is a demo of first inserting 100 objects to S3, then using `maxKeys: 10` and continuation token to retrieve all the listings 10 at a time (10 requests total):
Kapture 2020-11-13 at 00 38 36

Co-authored-by: @CryogenicPlanet


elorzafe:
@wei Thank you for this PR.

I have a question, why did you choose using `Hub` to notify the `continuationToken`? 

I think that can end up in race conditions. For example if you are doing `Storage.list` on different parts of your app concurrently? This complexity will have to be handled by customers, which is something Amplify prefer to avoid.

I would prefer to have pagination handled inside of the provider. What do you think?

wei:
> @wei Thank you for this PR.
> 
> I have a question, why did you choose using `Hub` to notify the `continuationToken`?
> 
> I think that can end up in race conditions. For example if you are doing `Storage.list` on different parts of your app concurrently? This complexity will have to be handled by customers, which is something Amplify prefer to avoid.
> 
> I would prefer to have pagination handled inside of the provider. What do you think?

Hi @elorzafe Thanks for the reply.

The main reason for letting app developers handle the paging is to provide maximum flexibility. With the approach in the PR, take example in a react native app where resources are constrained, a common use-case for an app is to lazy load 10-20 items at a time as the user scrolls down (As seen in the gif demo) instead of always loading a potentially huge list of objects in memory. 

> I think that can end up in race conditions. For example if you are doing `Storage.list` on different parts of your app concurrently?

We should also pass back `path` in attrs. But yes, app developer will have to write logic to send subsequent requests.

> I would prefer to have pagination handled inside of the provider. 

This is something I recommended in my original proposal (#7133) as well, but more as an alternative convenience helper instead of the solution itself. 

Let me know what you think and next steps. Thanks!!

Issue #7201, opened 2020-11-16
Phone number input - type tel


adampblack:
**Describe the bug**
The phone number input does not have the correct type set.

The problem without this is the mobile phone devices do not automatically suggest a phone number from contacts.

**To Reproduce**

<AmplifyAuthenticator
        style={{
          alignItems: 'flex-start',
          display: 'flex',
          flexWrap: 'wrap',
          justifyContent: 'center',
        }}
        usernameAlias="phone_number"
      />

It should have **<input type="tel"**
https://www.w3schools.com/tags/att_input_type_tel.asp

uncodable:
Type `tel` is on the `amplify-phone-field` component https://github.com/aws-amplify/amplify-js/blob/dbd46299af9c558251b8c652c3e50208982655c8/packages/amplify-ui-components/src/components/amplify-phone-field/amplify-phone-field.tsx#L36

sammartinez:
@adampblack, @uncodable is correct. Are you using these [UI Components](https://docs.amplify.aws/ui/auth/authenticator/q/framework/react) within Amplify? Please let us know.

adampblack:
I have run:
yarn upgrade --latest

info All dependencies
├─ @aws-amplify/analytics@3.3.11
├─ @aws-amplify/api-graphql@1.2.11
├─ @aws-amplify/datastore@2.7.3
├─ @aws-amplify/predictions@3.2.11
├─ @aws-amplify/ui-components@0.9.3
├─ @aws-amplify/ui-react@0.2.28
├─ @aws-amplify/ui@2.0.2

To answer your question I am using:
import { AmplifyAuthenticator, AmplifyConfirmSignIn, AmplifyConfirmSignUp, AmplifyForgotPassword, AmplifySignIn, AmplifySignUp } from '**@aws-amplify/ui-react**';

I had been using ui-components previously but the form field customization stopped working when I upgraded the modules for some reason so I upgraded to ui-react.

In terms of the field type it definitely says type="text" - but this might be a bug specific to ui-react not in ui-components.
![Screenshot 2020-11-17 at 15 29 52](https://user-images.githubusercontent.com/11552373/99360587-c7d54300-28eb-11eb-8fff-aefda0e936fc.png)

This is in the AmplifySignIn component but also occurs in the other components under ui-react as well.

tjleing:
This should be resolved in the next release now that #7211 is merged in!

adampblack:
Great job, thanks @tjleing

PR #7211, opened 2020-11-17
Fix phone number input type

tjleing:
Fixes: #7201 

type="tel" now on correct component

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-phone-field/__snapshots__/amplify-phone-field.spec.ts.snap
+++ b/packages/amplify-ui-components/src/components/amplify-phone-field/__snapshots__/amplify-phone-field.spec.ts.snap
@@ -4,10 +4,10 @@ exports[`amplify-phone-field spec: Render logic -> should render \`placeholder\`
 <amplify-phone-field label="New phone *" placeholder="Enter new phone">
   <mock:shadow-root>
     <div>
-      <amplify-form-field label="New phone *" type="tel">
+      <amplify-form-field label="New phone *">
         <div class="phone-field" slot="input">
           <amplify-country-dial-code></amplify-country-dial-code>
-          <amplify-input fieldid="phone" name="phone" placeholder="Enter new phone"></amplify-input>
+          <amplify-input fieldid="phone" name="phone" placeholder="Enter new phone" type="tel"></amplify-input>
         </div>
       </amplify-form-field>
     </div>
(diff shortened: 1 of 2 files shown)


packages/amplify-ui-components/src/components/amplify-phone-field/__snapshots__/amplify-phone-field.spec.ts.snap
```diff
@@ -4,10 +4,10 @@ exports[`amplify-phone-field spec: Render logic -> should render \`placeholder\`
 <amplify-phone-field label="New phone *" placeholder="Enter new phone">
   <mock:shadow-root>
     <div>
-      <amplify-form-field label="New phone *" type="tel">
+      <amplify-form-field label="New phone *">
         <div class="phone-field" slot="input">
           <amplify-country-dial-code></amplify-country-dial-code>
-          <amplify-input fieldid="phone" name="phone" placeholder="Enter new phone"></amplify-input>
+          <amplify-input fieldid="phone" name="phone" placeholder="Enter new phone" type="tel"></amplify-input>
```
harrysolovay, line 10: Do we know that `type="tel"` flows through into the resulting input component?

tjleing: It's absolutely supposed to...  been having trouble testing it on a sample though

tjleing: https://github.com/aws-amplify/amplify-js/blob/main/packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx

here you can see that it accepts a type variable that gets put on the resulting input component, so it _should_ work...

sammartinez: @tjleing is correct @harrysolovay, this will allow it to mark the input as type of tel which is what the snapshot is calling out here

tjleing: still would rather be able to verify it locally before merging in though..

wlee221 (approved):
This fix lgtm 👍

sammartinez (approved):
LGTM 🌮

PR #7212, opened 2020-11-17
feat(@aws-amplify/datastore): add local validations to AppSync scalars

### LINKED ISSUE #5963: [Datastore] Improve error handling when data is syncing
**Describe the bug**
The console does not produce error (actually it does, read along) when one of the AWSDateTime field stored in DynamoDB is not in the correct format. Took us a long time to figure out why our query isn't returning anything.

**To Reproduce**
1. Create schema.graphql:
```
type Order @model @auth(rules: [{ allow: owner }]) {
  id: ID!
  orderDate: AWSDateTime
  owner: String
}
```

2. Populate the table with orderDate as `2020-05-19T11:5:29.000Z`. It's missing a zero before the 5.
3. Query the Order table.
4. Huala, the query doesn't return anything.
5. Promptly check the console, no error found.
6. But...if you wait long enough, this errors shows up:
```
Possible Unhandled Promise Rejection (id: 408):
Object {
  "data": Object {
    "syncKDSOrders": Object {
      "items": Array [
        Object {
          "_deleted": null,
          "_lastChangedAt": 1590025175868,
          "_version": 2,
          "cancelBy": null,
          "customerName": "XXXX",
          "customerNumber": "XXXXXXXXXXXXXXX",
          "deliveryAddress": null,
          "deliveryNumber": null,
.
.
.

"errors": Array [
    Object {
      "locations": null,
      "message": "Can't serialize value (/syncOrders/items[32]/orderDate) : Unable to serialize `2020-05-23T06:4:37.000Z` as a valid DateTime Object.",
      "path": Array [
        "syncOrders",
        "items",
        32,
        "orderDate",
      ],
    },
  ],
}
```
Interesting bits is contained in the error.

**Expected behavior**
1. I tried catching the error in the query and that didn't work.
2. I tried catching with
```
DataStore.configure({
  errorHandler: (error) => {
    console.warn('Unrecoverable error', { error });
  },
});
```
That didn't work too.
My guess is that it's an error deep inside the source, perhaps somebody can enlighten me.

**What is Configured?**
```
yarn list --depth=0 --pattern amplify
yarn list v1.22.4
├─ @aws-amplify/analytics@3.1.9
├─ @aws-amplify/api-graphql@1.0.11
├─ @aws-amplify/api-rest@1.0.11
├─ @aws-amplify/api@3.1.9
├─ @aws-amplify/auth@3.2.6
├─ @aws-amplify/cache@3.1.9
├─ @aws-amplify/core@3.2.6
├─ @aws-amplify/datastore@2.0.10
├─ @aws-amplify/interactions@3.1.9
├─ @aws-amplify/predictions@3.1.9
├─ @aws-amplify/pubsub@3.0.10
├─ @aws-amplify/storage@3.1.9
├─ @aws-amplify/ui@2.0.2
├─ @aws-amplify/xr@2.1.9
├─ aws-amplify-react-native@4.1.1
└─ aws-amplify@3.0.10
```

Would be great if amplify team members can sort this out with us.

wei:
_Issue #, if available:_
Relates to #5963

_Description of changes:_
Perform some basic local validations for all [AppSync Defined Scalars](https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html). Currently there are no validations on these fields so one can store anything in e.g. `AWSDateTime` field locally. During datastore sync, AppSync does validations and graphql requests will fail. 

This PR uses basic regex to validate all the fields listed. However, some of the fields could use better validations. e.g. `2020-99-99T00:00:00Z` will pass validation with the current regex, We cannot use the built-in date function because AppSync supports non-standard ISO 8601 standard:

> The seconds field in the timezone offset will be considered valid even though it is not part of the ISO 8601 standard. 

Ideally, we should reuse AppSync's validation logic here for parity but having basic validations is much better than no validations : )

Screenshot of example validation error:
image

### DIFF
--- a/packages/datastore/__tests__/DataStore.ts
+++ b/packages/datastore/__tests__/DataStore.ts
@@ -68,6 +68,7 @@ describe('DataStore tests', () => {
 
 			const model = new Model({
 				field1: 'something',
+				dateCreated: new Date().toISOString(),
 			});
 
 			expect(model).toBeInstanceOf(Model);
@@ -86,6 +87,7 @@ describe('DataStore tests', () => {
 			const now = Date.now();
 			const model = new LocalModel({
 				field1: 'something',
+				dateCreated: new Date().toISOString(),
 			});
 
 			expect(model).toBeInstanceOf(LocalModel);
@@ -149,6 +151,7 @@ describe('DataStore tests', () => {
 
 			const model = new Model({
 				field1: 'something',
+				dateCreated: new Date().toISOString(),
 			});
 
 			expect(() => {
@@ -163,6 +166,7 @@ describe('DataStore tests', () => {
 
 			const model1 = new Model({
 				field1: 'something',
+				dateCreated: new Date().toISOString(),
 			});
 
 			const model2 = Model.copyOf(model1, draft => {
@@ -185,6 +189,7 @@ describe('DataStore tests', () => {
 
 			const model1 = new Model({
 				field1: 'something',
+				dateCreated: new Date().toISOString(),
 			});
 
 			const model2 = Model.copyOf(model1, draft => {
@@ -202,6 +207,7 @@ describe('DataStore tests', () => {
 
 			const model1 = new Model({
 				field1: 'something',
+				dateCreated: new Date().toISOString(),
 				optionalField1: undefined,
 			});
 
@@ -215,6 +221,7 @@ describe('DataStore tests', () => {
 
 			const model1 = new Model({
 				field1: 'something',
+				dateCreated: new Date().toISOString(),
 				optionalField1: null,
 			});
 
@@ -228,6 +235,7 @@ describe('DataStore tests', () => {
 
 			const model1 = new Model({
 				field1: 'something',
+				dateCreated: new Date().toISOString(),
 				optionalField1: 'something-else',
 			});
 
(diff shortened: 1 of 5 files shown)


packages/datastore/__tests__/DataStore.ts
```diff
@@ -415,6 +445,81 @@ describe('DataStore tests', () => {
...
+				'All elements in the ips array should be of type string | null | undefined, [object] received. '
+			);
+
+			expect(() => {
+				new Model({
+					field1: 'someField',
+					dateCreated: new Date().toISOString(),
+					ips: ['1.1.1.1'],
+				});
+			}).not.toThrow();
```
iartemiev, line 480: Let's also add a mix of valid and invalid items in a list and expect it to throw. 

E.g., 
```js
ips: ['1.1.1.1', 'not.an.ip'],
```

iartemiev (changes requested):
Looks great, @wei! Thank you! 🚀

I added a minor ask for an additional test, otherwise, PR is ready to be merged.

packages/datastore/__tests__/DataStore.ts
wei: Added in latest commit https://github.com/aws-amplify/amplify-js/pull/7212/commits/816635410fa8f140b8ef9906b211d309ddbd06a7

wei:
This PR should not be merged until rebasing after https://github.com/aws-amplify/amplify-js/pull/7216 is reviewed and merged.

iartemiev (approved):
Thanks again @wei this is a very helpful improvement!

PR #7216, opened 2020-11-17
fix(@aws-amplify/datastore): fix validation for array with optional element

wei:
_Issue #, if available:_
N/A

_Description of changes:_
1. Update error message to show `[null]` instead of `[object]` for `[null]`s.
2. The current test suite does not handle scenarios for something like
    ```
    misc: [String]
    ``` 
    `amplify codegen models` generates the following type:
    ```
    readonly misc?: (string | null)[];
    ```
    and schema:
    ```
    "misc": {
        "name": "misc",
        "isArray": true,
        "type": "String",
        "isRequired": false,
        "attributes": [],
        "isArrayNullable": true
    },
    ```
    With the current validation function, setting `misc: [null]` throws an exception stating 
    ```
    All elements in the rewards array should be of type string | null | undefined , [null] received. 
    ```
    This is clearly a mistake because `[null]` should pass validation. 

**To summarize**, this PR (1) fixes the datastore validation logic for arrays with optional elements and (2) updates error message to show `[null]` instead of `[object]` where applicable.

### DIFF
--- a/packages/datastore/src/datastore/datastore.ts
+++ b/packages/datastore/src/datastore/datastore.ts
@@ -271,7 +271,7 @@ const validateModelFields = (modelDefinition: SchemaModel | SchemaNonModel) => (
 				if (
 					!isNullOrUndefined(v) &&
 					(<[]>v).some(
-						e => typeof e !== jsType || (isNullOrUndefined(e) && isRequired)
+						e => isNullOrUndefined(e) ? isRequired : typeof e !== jsType
 					)
 				) {
 					const elemTypes = (<[]>v).map(e => typeof e).join(',');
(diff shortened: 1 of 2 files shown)


packages/datastore/src/datastore/datastore.ts
```diff
@@ -271,10 +271,10 @@ const validateModelFields = (modelDefinition: SchemaModel | SchemaNonModel) => (
 				if (
 					!isNullOrUndefined(v) &&
 					(<[]>v).some(
-						e => typeof e !== jsType || (isNullOrUndefined(e) && isRequired)
+						e => isNullOrUndefined(e) ? isRequired : typeof e !== jsType
```
wei, line 274: Please review this logic. 

e.g. for schema
```
"misc": {
    "name": "misc",
    "isArray": true,
    "type": "String",
    "isRequired": false,
    "attributes": [],
    "isArrayNullable": true
},
```

e.g. input is `[null]`

With old (incorrect) logic:
```
typeof e !== jsType || (isNullOrUndefined(e) && isRequired)
=>
'object' !== 'string' || (true && false)
=>
true || (true && false)
=>
true
```

With new logic:
```
isNullOrUndefined(e) ? isRequired : typeof e !== jsType
=>
true ? false : false
=>
false
```

`false` is the right return value.

iartemiev (approved):

PR #7225, opened 2020-11-18
fix(@aws-amplify/pubsub): Fix unhandled promised rejection and crash

nubpro:
_Issue #, if available:_
These issues surfaced when I making some changes to [AWSAppSyncRealTimeProvider.ts#L634-L653](https://github.com/aws-amplify/amplify-js/blob/main/packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts#L634-L653).
I was basically issuing a reject to the promise.

What I was doing:
image

_Description of changes:_
1. The app would crash if `this.awsRealTimeSocket.send()` is called despite awsRealTimeSocket being null.
2. Throwing an error inside a Promise is causing unhandled promised rejection error to surface in the console. To prove my point, here's a [demo](https://jsfiddle.net/6L0w7mvj/1/).


delighteckv:
Will this pull request solve below problem for datastore sync with "aws_appsync_authenticationType": "AMAZON_COGNITO_USER_POOLS"?


"type":"error","payload":{"errors":[{"errorType":"UnauthorizedException","message":"Permission denied"}]}}

nubpro:
> Will this pull request solve below problem for datastore sync with "aws_appsync_authenticationType": "AMAZON_COGNITO_USER_POOLS"?
> 
> "type":"error","payload":{"errors":[{"errorType":"UnauthorizedException","message":"Permission denied"}]}}

No it does not. Perhaps you can open a ticket with some code. From the look of it, it looks like the user is not signed in. Or the user is trying to modify a record that he/she does not have permission to.

iartemiev (approved):
Thanks for the PR, @nubpro! It looks good to me. Going to wait for @manueliglesias or @elorzafe to also take a quick look.

Here's how I was able to reproduce this and test _without_ modifying the library code.
1. Navigate to AppSync Console > settings
2. Change API auth mode to be something other than the default being used by the client
3. Start app
4. Unhandled promise rejection occurs

Apply change in this PR (remove `throw err` from line 619)
1. Start app
2. com.amazonaws.deepdish.graphql.auth#BadRequestException outputted to console but app does not enter unrecoverable state.
3.  Change API auth mode back to default
4. `await DataStore.stop() && await DataStore.start()` in the app
5. Syncing gets established successfully and DS reaches a 'ready' state

manueliglesias (approved):
Thanks @nubpro !

LGTM 👍 

thanks for testing it, @iartemiev

PR #7235, opened 2020-11-19
fix: remove unneeded orphan pkg zen-observable-ts

### LINKED ISSUE #7229: Orphan (and maybe unneccessary) dependency zen-observable-ts has an invalid sourcemap
**Describe the bug**
One of the aws-amplify dependencies, [zen-obsevable-ts](https://github.com/apollographql/apollo-link/tree/master/packages/zen-observable-ts), has an invalid sourcemap. Here's the first line of its sourcemap:
```
{"version":3,"file":"bundle.esm.js","sources":["../src/zenObservable.ts"]
```

This package didn't publish its source code to NPM, so the link to `../src/zenObservable.ts` will cause build-time warnings for some tools like `source-map-loader`.

Normally I'd just file a bug against the zen-obsevable-ts repo, but it's archived. Normally for archived repos, I'd file a PR to Amplify to switch to the new, non-archived repo. But zen-obsevable-ts doesn't seem to have any successor repo.

So I'm filing a bug report here because this repo is probably the only place to fix the problem. 

Looking at https://unpkg.com/zen-observable-ts, there's no executable code; the zen-observable-ts package seems to be solely a TS wrapper to add types around zen-observable. Here's the entire ESM [implementation](https://unpkg.com/zen-observable-ts@0.8.21/lib/bundle.esm.js): 

```js
import zenObservable from 'zen-observable';

var Observable = zenObservable;

export default Observable;
export { Observable };
//# sourceMappingURL=bundle.esm.js.map
```

So it might be possible to remove this package completely and just rely on `zen-observable` and `@types/zen-observable`.

I'm removing the rest of the bug report template since it doesn't seem to apply to this case, but let me know if I should put it back. 

I'll take a stab at PR-ing a fix to this by removing this package and relying on @types/zen-observable instead.

justingrant:
_Issue #, if available:_
Fixes #7229

_Description of changes:_
The zen-observable-ts package is no longer maintained, its [repo](https://github.com/apollographql/apollo-link/tree/master/packages/zen-observable-ts) is archived, and there's no clear successor package.  Also, it causes #7229 because its sourcemap points to sources that don't exist.

zen-observable-ts is also not needed; it's a thin wrapper around zen-observable that only seems to exist to make TS named imports a little cleaner. The wrapper doesn't seem to contain any runtime code; only TS declarations (which are mirrored by @types/zen-observable) and an import/re-export of zen-observable.

Therefore, this commit fixes #7229 because it:
* Replaces zen-observable-ts with zen-observable + @types/zen-observable in four amplify packages
* Updates imports to match to zen-observable's use of global variables
* Works around https://github.com/microsoft/TypeScript/issues/5711 in two files by adding a TS type import statement in those files.
* Removes unused @types/zen-observable dep from aws-amplify-angular


justingrant:
Apologies for the force-push spam. It took a few tries to find the right workaround to https://github.com/microsoft/TypeScript/issues/5711 that LGTM wouldn't complain about .

Also, tests won't run for me in CI because I don't have a paid plan for CircleCI. Let me know if there's a better way to get tests run for this PR.

justingrant:
Hmm, now that LGTM is passing I'm now seeing test failures, which might be related to the same problems that prompted #5155 to add zen-observable-ts in the first place. Sigh. I'll continue to investigate this, but in the meantime I switched this PR to draft status.

Issue #7244, opened 2020-11-20
Race condition in DataStore when reachability is switching back and forth


nubpro:
**Describe the bug**
I actually notice this bug on production where I would see **duplicate subscriptions** of the same model but had no idea why causes it. I believe it happens because of unstable network connection which triggered Reachability on and off repeatedly.

I poked around with some testing and find it pretty easy to reproduce.

**To Reproduce**

1. On Client A, spam reachability on and off.
2. On Client B, add new item.
3. On Client A, you would get 2 `DataStore.observe()`, one being an INSERT and UPDATE. This is because there's two active subscriptions on `onCreatePost`.

Here's a video: https://streamable.com/iankoh

**Expected behavior**
At any time, there should be only 1 active subscription per subscription (`onCreatePost`).

**Code Snippet**
schema.graphql
```
enum PostStatus {
    ACTIVE
    INACTIVE
}

type Post @model {
    id: ID!
    title: String!
    rating: Int!
    status: PostStatus!
}
```

App.js
```
import React, { useEffect } from 'react';
import './App.css';

import { Hub, Reachability } from '@aws-amplify/core';
import { DataStore, Predicates } from '@aws-amplify/datastore';
import { Post, PostStatus } from './models';
import { withAuthenticator, AmplifySignOut } from '@aws-amplify/ui-react';

window.LOG_LEVEL = 'DEBUG';

function onCreate() {
    DataStore.save(
        new Post({
            title: `New title ${Date.now()}`,
            rating: (function getRandomInt(min, max) {
                min = Math.ceil(min);
                max = Math.floor(max);
                return Math.floor(Math.random() * (max - min)) + min;
            })(1, 7),
            status: PostStatus.ACTIVE,
        })
    );
}

function onDeleteAll() {
    DataStore.delete(Post, Predicates.ALL);
}

async function onQuery() {
    const posts = await DataStore.query(Post, (c) => c.rating('gt', 4));

    console.log(posts);
}

function App() {
    useEffect(() => {
        const subscription = DataStore.observe(Post).subscribe((msg) => {
            console.warn('Post subscription:', msg.opType, msg.element);
        });
        return () => subscription.unsubscribe();
    }, []);

    useEffect(() => {
        // Create listener
        const listener = Hub.listen('datastore', async (hubData) => {
            const { event, data } = hubData.payload;

            console.log(event, data);
        });

        // Remove listener
        return () => {
            listener();
        };
    }, []);

    return (
        <div className="App">
[...]

PR #7250, opened 2020-11-21
fix(@aws-amplify/datastore): Fix ctlSubsSubscription not getting unsubscribed when device goes offline

nubpro:
_Issue #, if available:_
#7244

_Description of changes:_
The use `await` causes the subscription of `ctlSubsObservable` not to be immediately returned and pushed into the `subscriptions` array. Hence, if the promise is not resolved by the time the device goes offline, the subscription is not unsubscibed as it has yet to be pushed into the `subscriptions` array.

https://github.com/aws-amplify/amplify-js/blob/b7eef9dc2a03e7fa4d9a6bb7420a44b0731ecc35/packages/datastore/src/sync/index.ts#L197-L202


iartemiev (approved):
Was able to reproduce the linked issue and confirmed that this PR fixes it.

Thank you so much, @nubpro!

Issue #7103, opened 2020-11-02
Ability to remove an AWSIoTProvider pluggable or change the endpoint after calling addPluggable?


matwerber1:
**Is your feature request related to a problem? Please describe.**

I'm creating an AWS IoT message tester, similar to what you get in the AWS IoT web console.

I am configuring with an IoT pluggable, per below:

```js
Amplify.addPluggable(new AWSIoTProvider({
      aws_pubsub_region: region,
      aws_pubsub_endpoint: endpoint,
    }));  
  }
```

I would like the ability to populate `region` and `endpoint` based on form input. This is easy enough... however, I also want the user to be able to change the endpoint and region. That's where it gets tricky.

I can call `addPluggable` again, which seems to work... but it still leaves the old provider there.. and subsequent `Pubsub.publish()` commands publish to both the old and new provider. 

It doesn't seem like there's a way to change the existing (old) provider endpoint, nor is there a way to remove the old provider when I add the new one.

**Describe the solution you'd like**

I'd like to either (a) change the IoT endpoint of a previously configured AWSIoTProvider pluggable, or (b) remove a pluggable that was previously added via `Amplify.addPlugable()` (so that I can re-add a new one).

**Describe alternatives you've considered**

**Additional context**
Add any other context, an idea of the implementation or screenshots about the feature request here.

zacharyrogers1:
@matwerber1 I took a shot at this and created a PubSub.removePluggable() function to be able to remove and add providers as needed. This should work for being able to change region, endpoint, etc as needed.

david-mcafee:
[The PR](https://github.com/aws-amplify/amplify-js/pull/7256) for this has been merged, closing this ticket. Thanks again, @zacharyrogers1!

PR #7256, opened 2020-11-22
Added removePluggable function to PubSub

zacharyrogers1:
Added the ability to remove providers from PubSub after creation. This should allow for users to be able to reconfigure their PubSub providers when changing authentication states or wanting to reconfigure regions. 

Issue #, if available: #7103

Description of changes:

Previously the pubsub providers list could not be modified once created. This meant that when a user changed authentication states or wanted to change their connection region for IoT they would be unable to do so. New function will allow websocket providers to be removed so that new addpluggables may be added. Calling PubSub.removePluggable() will not delete the old provider, simply remove it from the internal list of providers so that a new one can be added.


AraiHaruki:
Will this be merged?
I need this feature.

kirkryan:
@renebrandel - Hey! Any chance we could have AWS reviewer allocated to this please amigo? This is a REALLY useful feature - basically it's a key functionality that is missing from the PubSub client in that there is currently no way of removing a provider - which means you can end up with multiple duplicate providers and in the worst-case sending AND receiving multiple times duplicate data which quickly has performance and cost implications for any client using the current AWS Pubsub client.

Kirk - AWS Community Builder

sammartinez:
> @renebrandel - Hey! Any chance we could have AWS reviewer allocated to this please amigo? This is a REALLY useful feature - basically it's a key functionality that is missing from the PubSub client in that there is currently no way of removing a provider - which means you can end up with multiple duplicate providers and in the worst-case sending AND receiving multiple times duplicate data which quickly has performance and cost implications for any client using the current AWS Pubsub client.
> 
> Kirk - AWS Community Builder

Hey Kirk! Thanks for raising this to us, we are starting the process internally and will provide an update next week on progress of getting this pull request merged in for our customers.

kirkryan:
> > @renebrandel - Hey! Any chance we could have AWS reviewer allocated to this please amigo? This is a REALLY useful feature - basically it's a key functionality that is missing from the PubSub client in that there is currently no way of removing a provider - which means you can end up with multiple duplicate providers and in the worst-case sending AND receiving multiple times duplicate data which quickly has performance and cost implications for any client using the current AWS Pubsub client.
> > Kirk - AWS Community Builder
> 
> Hey Kirk! Thanks for raising this to us, we are starting the process internally and will provide an update next week on progress of getting this pull request merged in for our customers.

AWSome - thanks @sammartinez - if you need any further technical/business justification @renebrandel know's my details and I'm happy to help out.

Many thanks! Once implemented I'm happy to update the documentation as a separate PR.

packages/pubsub/src/PubSub.ts
```diff
@@ -110,6 +110,14 @@ export class PubSubClass {
 		}
 	}
 
+	removePluggable(options: ProvidertOptions) {
```
elorzafe, line 113: We used on the library in other categories `providerName` as the argument

Can you add the JS doc for this. For example
```
/**
 * remove plugin from PubSub category
 * @param providerName - the name of the plugin
 */

File: packages/pubsub/src/PubSub.ts [hunk 2]
```diff
@@ -110,6 +110,14 @@ export class PubSubClass {
...
+		const { provider: providerName } = options;
+
+		const indexOfPluggableToRemove = this._pluggables.indexOf(providerName);
+
+		return this._pluggables.splice(indexOfPluggableToRemove, 1)[0];
```
elorzafe, line 118: I think is better to do something like this, `indexOf` for `providerName` would not work because is an array of objects not an array of strings.

```
this._pluggables = this._pluggables.filter(
			pluggable => pluggable.getProviderName() !== providerName
		);
```


Why do you need to return the plugin that was removed, just to know if it was removed?

elorzafe (changes requested):
I would check the function argument (plugin name instead of passing the actual plugin and the other will be the way is filtered

packages/pubsub/src/PubSub.ts
david-mcafee: +1

packages/pubsub/src/PubSub.ts
david-mcafee: +1 on this approach.

I'd also be curious to know why we're returning the plugin that was removed. Perhaps we could include a return type on the function to make that explicit, if there's a valid reason to do so?

packages/pubsub/__tests__/PubSub-unit-test.ts
```diff
@@ -196,6 +196,37 @@ describe('PubSub', () => {
...
+					aws_pubsub_region: 'region',
+					aws_pubsub_endpoint: 'wss://iot.mymockendpoint.org:443/notrealmqtt',
+				},
+			};
+			const pubsub = new PubSub({});
+			pubsub.configure(config);
+			const awsIotProvider = new AWSIoTProvider();
+			pubsub.addPluggable(awsIotProvider);
+
+			const removedPluggable = pubsub.removePluggable({provider: 'AWSIoTProvider'});
```
david-mcafee, line 212: Could you update this unit test to do the following?

1. Add a provider
2. Remove the provider
3. Add a new provider (different from the first)
4. Verify that the correct provider is being used
3. This way, instead of validating the return value of `removePluggable`, validate that the correct provider exists (and that the removed one does not). Otherwise, we are only testing the return value of `removeProvider`

packages/pubsub/__tests__/PubSub-unit-test.ts
```diff
@@ -238,6 +269,20 @@ describe('PubSub', () => {
 
 			awsIotProvider.onDisconnect({ errorCode: 1, clientId: '123' });
 		});
+
+		test('should remove MqttOverWSProvider', () => {
```
david-mcafee, line 273: The above comment could also be applied here, as well

david-mcafee (changes requested):
Left a few comments. Thanks again for creating this PR!!

packages/pubsub/src/PubSub.ts
zacharyrogers1: Changed signature of function to accept providerName. Thanks for the suggestion.

packages/pubsub/src/PubSub.ts
zacharyrogers1: Changed implementation of removePluggable to be identical to what you suggested. Thanks for the feedback!

packages/pubsub/__tests__/PubSub-unit-test.ts
zacharyrogers1: Updated test to incorporate test flow you suggested:

1. addProvider
2. removeProvider
3. add new Provider
4. Call publish on pubsub class and verify that the new provider publish function was called and the old provider was not called.

packages/pubsub/__tests__/PubSub-unit-test.ts
zacharyrogers1: Added same test logic to this function.

zacharyrogers1:
Thank you everyone for the review. It was very helpful for my first contribution to this project. I believe I have taken all of your feedback into account and updated the tests/code. 

Please let me know if there is anything else to add/remove.

david-mcafee:
> Thank you everyone for the review. It was very helpful for my first contribution to this project. I believe I have taken all of your feedback into account and updated the tests/code.
> 
> Please let me know if there is anything else to add/remove.

Thanks, @zacharyrogers1! Taking a look at your changes now.

packages/pubsub/src/PubSub.ts
```diff
@@ -110,6 +110,16 @@ export class PubSubClass {
 		}
 	}
 
+	/**
+	 * remove plugin from PubSub category
+	 * @param providerName - the name of the plugin
+	 */
```
david-mcafee, line 116: I added this minor JSDoc comment

packages/pubsub/__tests__/PubSub-unit-test.ts
```diff
@@ -196,6 +196,41 @@ describe('PubSub', () => {
...
+			const newProvider = new AWSIoTProvider({
+				aws_pubsub_region: 'new-region',
+				aws_pubsub_endpoint: 'wss://iot.newEndPoint.org:443/newEndPoint',
+			});
+			jest.spyOn(newProvider, 'publish');
+
+			pubsub.addPluggable(originalProvider);
+			pubsub.removePluggable('AWSIoTProvider');
+			pubsub.addPluggable(newProvider);
+			pubsub.publish('someTopic', { msg: 'published Message' });
```
david-mcafee, line 216: I made a minor update here to remove the Provider on publish. It's necessary to specify the Provider if there are multiple that are registered. If we specify it in this test, we aren't testing that the second Provider was successfully removed. By not specifying the Provider, we ensure that the second (removed) Provider did not publish a message.

packages/pubsub/__tests__/PubSub-unit-test.ts
```diff
@@ -238,6 +273,28 @@ describe('PubSub', () => {
...
+			const newProvider = new MqttOverWSProvider({
+				aws_pubsub_region: 'region',
+				aws_appsync_dangerously_connect_to_http_endpoint_for_testing: true,
+			});
+			jest.spyOn(newProvider, 'publish');
+
+			pubsub.addPluggable(originalProvider);
+			pubsub.removePluggable('MqttOverWSProvider');
+			pubsub.addPluggable(newProvider);
+			pubsub.publish('someTopic', { msg: 'published Message' });
```
david-mcafee, line 293: I made the same update here

david-mcafee (approved):
Everything looks good! Thanks so much for this PR! I added a few updates to this PR as well (JS Doc, and updates to the unit tests).

packages/pubsub/src/PubSub.ts
david-mcafee: Ofcourse! Thanks for addressing this feedback!

elorzafe (approved):
LGTM! thanks @zacharyrogers1 🌮 🥇

PR #7275, opened 2020-11-24
deps(@aws-amplify/core): remove unused aws-sdk dependencies

wlee221:
_Issue #, if available:_

_Description of changes:_ This removes two unused `@aws-sdk` dependencies within `@aws-amplify/core`.


harrysolovay:
@wlee221 are you sure these aren't used by the rollup process for node compatibility in bundling UI components? / did you make sure UI builds and works correctly locally? If so, I'll approve :) lmk

tjleing:
To make sure our earlier conversation doesn't get lost -- possibly remove `url` and `@aws-sdk/types` as well -- looks like they're unused too!

wlee221:
Hi all, I'll double check with @manueliglesias or @Amplifiyer to double check why these packages were added in the first place.

manueliglesias (approved):
The last usage of `@aws-sdk/util-user-agent-browser` was removed in https://github.com/aws-amplify/amplify-js/commit/c127e77c551520950013f0e4d113ef0e368a6f60 but the dependency wasn't removed from `package.json`.


`@aws-sdk/node-http-handler` was also added in that commit but looks like it wasn't used (unless it used to have a side effect I am not aware of, I couldn't find any at https://github.com/aws/aws-sdk-js-v3/tree/%40aws-sdk/node-http-handler%401.0.0-gamma.7/packages/node-http-handler)

this LGTM, let's see if the pipeline catches any issue

wlee221:
Thanks! I removed `url` as per @tjleing's comment. I found that `@aws-sdk/types` are used [here](https://github.com/aws-amplify/amplify-js/blob/74b895c45e629edc5ecf333e2769f222ae500834/packages/core/src/Credentials.ts#L19, so I left it in the package.json.

tjleing (approved):
LGTM!

wlee221:
Thanks for the reviews guys 👍

PR #7288, opened 2020-11-26
fix(@aws-amplify/core): refresh manually after TTL has passed

### LINKED ISSUE #6343: 50 Minute Token Refresh is Not Working
**Describe the bug**
https://github.com/aws-amplify/amplify-js/pull/4205 is not working - tokens should be automatically refreshed once they have 10 min or less to expire, but this is not happening.

**To Reproduce**
Open an `amplify-js` application (with cognito authentication), wait for 55 min, then call `const session = await Auth.currentSession()` and see that `session.accessToken.payload.exp` is less than 300 - with each repeated call you will see this value decrease, and the token will not be refreshed.

**Expected behavior**
I expect tokens to refresh when they have less than 10 min to expire.

**Code Snippet**
```
    const session = await Auth.currentSession()
    const remainingTime = session.accessToken.payload.exp - (new Date().getTime() / 1000)
    console.log(remainingTime)
```

I should note that the following snippet works just fine (and we're writing custom code now to check `session.accessToken.payload.exp` and run this snippet when there is little time left):
```
        user.refreshSession(session?.refreshToken, (error) => {
          if (error) {
            console.log('failed to refresh token')
          }
        })
```

<details>
  <summary><strong>Environment</strong></summary>

```
  System:
    OS: macOS 10.15.4
    CPU: (8) x64 Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
    Memory: 281.47 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.18.0 - ~/.nvm/versions/node/v12.18.0/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v12.18.0/bin/npm
  Browsers:
    Chrome: 83.0.4103.116
    Firefox: 78.0.1
    Safari: 13.1
  npmPackages:
    @amaabca/react-components: ^14.9.0 => 14.9.0
    @amaabca/sensitive-param-filter: ^1.0.6 => 1.0.6
    @babel/cli: ^7.2.3 => 7.10.1
    @babel/plugin-transform-runtime: ^7.5.5 => 7.10.1
    @datadog/browser-rum: ^1.5.0 => 1.12.1
    @testing-library/cypress: ^6.0.0 => 6.0.0
    @testing-library/jest-dom: ^5.1.1 => 5.9.0
    @testing-library/react: ^9.4.0 => 9.5.0
    @testing-library/user-event: ^10.3.1 => 10.4.1
    @types/lodash.orderby: ^4.6.6 => 4.6.6
    @welldone-software/why-did-you-render: ^4.0.5 => 4.2.5
    apollo-link: ^1.2.11 => 1.2.14
    aws-amplify: ^3.0.18 => 3.0.18
    aws-amplify-react: ^3.1.2 => 3.1.9
    aws-appsync: ^3.0.2 => 3.0.3
    aws-sdk: ^2.374.0 => 2.690.0
    axios: ^0.18.0 => 0.18.1
    axios-mock-adapter: ^1.18.1 => 1.18.1
    babel-preset-react: ^6.24.1 => 6.24.1
    commander: ^4.1.0 => 4.1.1
    currency.js: ^1.2.1 => 1.2.2
    cypress: 4.10.0 => 4.10.0
    danger: ^7.0.1 => 7.1.4
    deep-equal: ^2.0.1 => 2.0.3
    deepcopy: ^2.0.0 => 2.0.0
    enzyme: ^3.9.0 => 3.11.0
    enzyme-adapter-react-16: ^1.11.2 => 1.15.2
    enzyme-to-json: ^3.3.5 => 3.5.0
    esm: ^3.2.25 => 3.2.25
    fuse.js: ^3.4.5 => 3.6.1
    glob: ^7.1.6 => 7.1.6
    graphql-tag: ^2.10.1 => 2.10.3
    husky: ^1.3.1 => 1.3.1
    idle-js: ^1.2.0 => 1.2.0
    inquirer: ^6.2.1 => 6.5.2
    isomorphic-fetch: ^2.2.1 => 2.2.1
    jest-styled-components: ^7.0.2 => 7.0.2
    jest-when: ^2.3.1 => 2.7.2
    lodash.orderby: ^4.6.0 => 4.6.0
    mock-fs: ^4.10.1 => 4.12.0
    mock-spawn: ^0.2.6 => 0.2.6
    mockdate: ^2.0.5 => 2.0.5
    moment: 2.25.3 => 2.25.3
    moment-timezone: 0.5.28 => 0.5.28
    node-sass: ^4.12.0 => 4.14.1
    nvmrc-check: ^1.0.2 => 1.0.2
    prop-types: ^15.7.2 => 15.7.2
    query-string: ^6.4.0 => 6.12.1
    random-name: ^0.1.2 => 0.1.2
    react: ^16.12.0 => 16.13.1
    react-dom: ^16.12.0 => 16.13.1
    react-hotkeys: ^2.0.0 => 2.0.0
    react-joyride: ^2.2.1 => 2.2.1
    react-router-dom: ^5.0.0 => 5.2.0
    react-scripts: 3.3.0 => 3.3.0
    react-select: ^2.1.2 => 2.4.4
    react-testing-library: 6.1.2 => 6.1.2
    regenerator-runtime: ^0.13.2 => 0.13.5
    request-promise-native: ^1.0.8 => 1.0.8
    rollbar: ^2.11.0 => 2.16.2
    snazzy: ^8.0.0 => 8.0.0
    standard: ^12.0.1 => 12.0.1
    striptags: ^3.1.1 => 3.1.1
    styled-components: ^5.0.0 => 5.1.1
    use-deep-compare-effect: ^1.3.1 => 1.3.1
    uuid: ^3.3.2 => 3.4.0
    whatwg-fetch: ^3.0.0 => 3.0.0
  npmGlobalPackages:
    npm: 6.14.4
```

</details>

wlee221:
_Issue #, if available:_ Fixes #6343

_Description of changes:_ This is a follow-up fix to #4205, which refreshes the user session after 50 minute has passed.

**Problem Description**
#4205 introduces 50 minute TTL to [`Credentials.ts`](https://github.com/aws-amplify/amplify-js/pull/4205/files#diff-529fe0eb9dda8c67eb9f7cf7c63a52b642bc7c4919a1e5195580c747073e885dR11). If TTL has passed, `Credentials._keepAlive` function calls [`user.getSession`](https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/src/CognitoUser.js#L1366) in `amazon-cognito-identity-js` package. `user.getSession` checks if the [tokens are valid](https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/src/CognitoUser.js#L1406-L1409), then it calls `user.refreshSession` that *should* refresh the session.

The problem is that `amazon-cognito-identity-js` is unaware of 50 minute TTL logic because the logic resides in `Credentials.ts`. So if the session's TTL has passed but session has not expired yet (ie. between 50 min and 60 min), `cognito-identity-js` still thinks that the session is still valid and skips the session refresh.

**Solution**
This PR manually refreshes the session if it has passed TTL but not yet expired, to "fill" the hole between 50 and 60 minutes. I have confirmed this fix locally. This has some concerns that I'll address in the comments -- any review on that is much appreciated!


packages/core/src/Credentials.ts
```diff
@@ -94,10 +94,10 @@ export class CredentialsClass {
 		return this._gettingCredPromise;
 	}
 
-	private _keepAlive() {
+	private async _keepAlive() {
```
wlee221, line 97: `_keepAlive` returns a promise anyway, so I changed this function to async to use await.

packages/core/src/Credentials.ts
```diff
@@ -108,11 +108,26 @@ export class CredentialsClass {
...
-		} else {
+		if (!Auth || typeof Auth.currentUserCredentials !== 'function') {
 			return Promise.reject('No Auth module registered in Amplify');
 		}
+
+		if (!this._isExpired(cred) && this._isPastTTL()) {
+			logger.debug('ttl has passed but token is not yet expired');
+			try {
+				const user = await Auth.currentUserPoolUser();
+				const session = await Auth.currentSession();
```
wlee221, line 119: Are there any concerns in using `Auth` APIs here? We were already using `Auth.currentUserCredentials()` here, so I assumed this is fine.

packages/core/src/Credentials.ts
```diff
@@ -108,11 +108,26 @@ export class CredentialsClass {
...
+				const refreshToken = session.refreshToken;
+				user.refreshSession(refreshToken, (err, data) => {
```
wlee221, line 121: Is there an easier way to refresh the session? I see that there is [`user.refreshSessionIfPossible`](https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/src/CognitoUser.js#L1223-L1233) API in `CognitoUser.js`, but this is marked as private.

manueliglesias: I am not aware of an easier way. While you are not doing something meaningful with the error case, and the private method you link seems to be doing the same, I'd prefer if we stay on the public apis. We have type cheking on those and (as you  did here) we can log a meaningful message for troubleshooting purposes.

packages/core/src/Credentials.ts
```diff
@@ -200,13 +215,11 @@ export class CredentialsClass {
...
-			ts < this._nextCredentialsRefresh
-		) {
-			return false;
-		}
-		return true;
+		return expiration.getTime() <= ts + delta;
+	}
+
+	private _isPastTTL(): boolean {
+		return this._nextCredentialsRefresh <= Date.now();
```
manueliglesias, line 222: I am not sure of the implications for this in the work done as part of clock-skew issue. Please check this with either @ericclemmons or @elorzafe

manueliglesias (approved):
Overall LGTM 👍 I just would like you to verify that the TTL plays nicely with the clock skew work

amhinson (approved):

packages/core/src/Credentials.ts
wlee221: Sure! I see, maybe we can have a tighter check for `_isExpired` with possible clock skew. We can use `DateUtils.getDateWithClockOffset` instead of `Date.now()` in case Amplify is aware of clock skew. AFAIK `credentials` by itself cannot learn clock skews until an actual api call is made. But I don't see harm in using it because it [defaults](https://github.com/aws-amplify/amplify-js/blob/main/packages/core/src/Util/DateUtils.ts#L19-L21) to `Date.now()` if no clock skew hasn't been detected yet.

wlee221: I see however that our entire `Credentials.ts` file uses `Date.now()` across all time stamps, so I think this is a separate improvements pr that we can make later. So for this PR, I think what I have is consistent with what we have right now. But I'm happy to run it by @ericclemmons as I don't have experience in this topic! @ericclemmons, can you confirm this?

ericclemmons: This is a good question. I agree with @wlee221's assessment:
> AFAIK `credentials` by itself cannot learn clock skews until an actual api call is made...

The clock skew logic is resolving a mismatch of timestamps between client & server (often called "clock drift").

I believe it's safe for expiration logic to go off of `Date.now()`, since we're computing a relative 50-60 minute expiration window.

`DateUtils.getDateWithClockOffset` is best used when we're signing HTTP requests and need to account for "clock drift".

packages/core/src/Credentials.ts
ericclemmons, line 121: I see that `refreshSession` isn't being `await`'d and is being used as an optimistic update: whether it succeeds or fails doesn't affect `Auth.currentUserCredentials()` from being returned (or delayed!).

I agree with this approach, but wanted to comment on it in the event there's a reason we're missing on why we _should_ `await` it (with a Promisify wrapper, since it's a callback).

ericclemmons (approved):
@wlee221 This PR looks good!  First, I want to commend you on your research in the PR body: **your narrative gives the context necessary to understand (1) why this behavior exists in the first place and (2) what's necessary to resolve it.

The PR is an artifact of that discovery, and requires almost no reverse-engineering of what the code is _trying_ to accomplish.

Second, you took the opportunity to clean up code & "return early" for edge-cases and de-indent the common (happy) path.

As always, great work @wlee221!

If there were one opportunity here to prevent regressions, it would be adjusting one of the `Credentials.ts` tests to assert that, when the token isn't expired but past TTL, `refreshSession` is called:

```js
expect(refreshSessionSpy).toHaveBeenCalled() // or similar
```

packages/core/src/Credentials.ts
wlee221: Good point! I actually think promisifying this is. a better idea because customers would expect the updated session information when they call `Credentials.get`. I'll promisify this!

wlee221:
> If there were one opportunity here to prevent regressions, it would be adjusting one of the `Credentials.ts` tests to assert that, when the token isn't expired but past TTL, `refreshSession` is called:

Thanks @ericclemmons! I actually caught one bug while writing the unit test, where we were mixing in TTL logic within `_isExpired` function ([this commit](https://github.com/aws-amplify/amplify-js/pull/7288/commits/d00e414e32ed927d0bed5ec6bd0292375329b00e)).

I'll merge this in monday after I confirm the changes one last time. Thanks everyone for reviewing!

packages/core/src/Credentials.ts
```diff
@@ -209,13 +209,12 @@ export class CredentialsClass {
 		}
 		logger.debug('are these credentials expired?', credentials);
 		const ts = Date.now();
-		const delta = 10 * 60 * 1000; // 10 minutes in milli seconds
```
wlee221, line 197: 10 minute check isn't needed anymore because it is checked by `_isPastTTL`

PR #7295, opened 2020-11-27
Datastore/feat/save with id from client

crolly:
**TLDR: This pull request allows model ids to be set from the client.**

According to the amplify-cli docs, it is possible to set the primary key field `id` from the client through the graphql mutation.
The resolver looks like this:
`"id":   $util.dynamodb.toDynamoDBJson($util.defaultIfNullOrBlank($ctx.args.input.id, $util.autoId()))`

Datastore uses the same resolver, however, redefines the id, that has been set with its own newly generated uuid ignoring the one being set in the input of `DataStore.save(Model, { id: ...}`.
To me, it is unclear which advantages this would bring and only see the disadvantage of not having full control over the assignment of ids.

Hence this PR tries to changes this.

### DIFF
--- a/packages/datastore/__tests__/DataStore.ts
+++ b/packages/datastore/__tests__/DataStore.ts
@@ -1,6 +1,7 @@
 import 'fake-indexeddb/auto';
 import { decodeTime } from 'ulid';
 import uuidValidate from 'uuid-validate';
+import { v4 as uuidv4 } from 'uuid';
 import Observable from 'zen-observable-ts';
 import {
 	DataStore as DataStoreType,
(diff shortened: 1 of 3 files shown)


crolly:
PR is the result of discussion #5976

crolly:
hey, are there any uncertainties or considerations before merging this?

stheobald:
Is this merge in the pipeline anytime soon? 

I'm really fighting this current limitation, it's causing me a LOT of extra work to manage things which are in my local recoil state before saving them to dataStore, and which then get new ids. This means I have to update my recoil relationship indexes for the new id, causing a selector cache rebuild, these then get re-calculated again as the child nodes get saved and in turn get new ids, causing another rebuild.

This is a horrible, inefficient, costly and entirely unnecessary exercise which causes ui jank. This single limitation is rapidly turning what should be a smooth exercise into something entirely unpleasant.

Can I assist in getting this merge progressed in any way?

kzhang-dsg:
Hi @manueliglesias its been a year. Any update on the new feature that allows user to set their own IDs when creating new records?

crispiestsquid:
@manueliglesias I would also like to bump this along, as it seems to be a pretty annoying issue to solve for...

I really would like to create models that extend my cognito users so that I don't have to add a ton of custom attributes to them. It would make it so much cleaner to be able to create these models with the assumption that I can set their ids equal to the cognito user's id.

I hope these changes are still in the works!

Issue #6363, opened 2020-07-18
DataStore latter updates should take precedent on a slow network


nubpro:
**Describe the bug**
Say you have a button to update a Post title.
With that you change the post title to 'Hairy Potter'. This mutation is queued in the outbox.
However before this mutation is processed, you decides to update the post title again to 'Marry in the Soberland'.
As a user, you would expect your post title to be permanently change to the latter title as the local updates reflects that.
Once the both mutations are processed, the post title would return to 'Hairy Potter' which isn't correct.

**To Reproduce**
1. On iOS, enable Network Link Conditioner, choose Edge(or even worse) as the profile.
2. Update a post title to 'Title A'.
3. Before the mutation is processed, quickly update the post title to 'Title B'.
4. After both mutations are done, you would notice the post title is back to 'Title A'.

**Expected behavior**
The expected post title should be 'Title B'.

**What is Configured?**
We are using 'Auto merge' as the Conflict Handler, however it shouldn't matter for this case. As I would expect DataStore to give precedent to the second mutation because the first mutation is not processed yet. Personally, I think the solution would be to increment the version number by 1 to the second mutation after knowing that the first mutation is processed.

<details>
  <summary><strong>Environment</strong></summary>

```
  System:
    OS: macOS 10.15.5
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Memory: 59.48 MB / 8.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.3.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 83.0.4103.116
    Safari: 13.1.1
  npmPackages:
    @babel/core: ^7.9.0 => 7.10.3 
    @babel/runtime: ^7.9.2 => 7.10.3 
    @react-native-community/async-storage: ^1.11.0 => 1.11.0 
    @react-native-community/eslint-config: ^1.0.0 => 1.1.0 
    @react-native-community/masked-view: ^0.1.9 => 0.1.10 
    @react-native-community/netinfo: ^5.7.1 => 5.9.4 
    @react-navigation/drawer: ^5.4.1 => 5.8.4 
    @react-navigation/native: ^5.1.5 => 5.6.1 
[...]

PR #7354, opened 2020-12-07
fix(@aws-amplify/datastore): fix consecutive updates

iartemiev:
_Issue #, if available:_ https://github.com/aws-amplify/amplify-js/issues/6363

```graphql
type Todo @model {
  id: ID!
  name: String!
}
```

Currently, if we run the following code in DataStore:

```js
async function changeTodoName() {
  const [todo] = await DataStore.query(Todo); // { name: A,  _version: 1 }

  for (let letter of ['B', 'C', 'D']) {
    await DataStore.save(
      Todo.copyOf(todo, (updated) => {
        updated.name = letter;
      })
    );
  }
}
```

...the name of our Todo will be `'B'` after all AppSync requests are resolved, i.e., latter updates get ignored.

The expected behavior is for `name` to be `'D'`.

Here's a sequence diagram detailing the current behavior (abstracted/simplified):
consecutive writes seq diagram

It happens on both fast and slow networks.

**Fast (~500 Mbps):**
fast-network

**Slow (10 Kbps):**
slow-network

### Proposal
Reconcile the version of update mutations in the mutation queue when an AppSync mutation for the same model ID resolves.

Diagram of updated behavior:
consecutive writes seq diagram postpr

Here's the same example running with the code in this PR:
**Fast (~500 Mbps):**
fast-network-pr

**Slow (10 Kbps):**
slow-network-pr

**Note: the `_version` in the response will only be synced to related items in the queue when the AppSync mutation response contains the same item that was sent in the request, i.e., if the original mutation was sent by the same client.
In the event of a handled conflict, where the response payload is different from the request payload, we do not sync the `_version` and subsequent writes will effectively be discarded.**

Sequence diagram for multi-client behavior:
DataStore Consecutive Updates - Multi #2


nubpro:
Been wondering about this, how does iOS and Android library counterpart deal with such issue?
Do they share the same implementation?

tagging @jamesonwilliams perhaps he can share some insides from the Android side.

iartemiev:
@nubpro they most likely have the same issue on iOS and Android. We're going to sync up on this across teams after the holidays.

kylanry:
Hi guys! Does anybody know when this fix will be added to the new Amplify release? I am using the iOS Amplify SDK and also got the same issue, it is blocker for us. Will this fix also applied for iOs version?

iartemiev:
> Hi guys! Does anybody know when this fix will be added to the new Amplify release? I am using the iOS Amplify SDK and also got the same issue. It is blocker for us. Will this fix also applied for iOS version?

@kylanry, yes, this fix will be applied across the Amplify libraries (JS, iOS, Android, Flutter). We're making some final modifications to the solution and are planning on releasing it soon.

packages/datastore/src/sync/processors/mutation.ts
```diff
@@ -151,14 +151,22 @@ class MutationProcessor {
...
 			const record = result.data[opName];
-			await this.outbox.dequeue(this.storage);
-
-			const hasMore = (await this.outbox.peek(this.storage)) !== undefined;
+			let hasMore = false;
+
+			await this.storage.runExclusive(async storage => {
+				// using runExclusive to prevent possible race condition
+				// when another record gets enqueued between dequeue and peek
+				console.log('dequeue record', record);
```
nubpro, line 166: Might want to remove this `console.log`?

iartemiev:
@manueliglesias @amhinson - this is ready for review now 😸 

`outbox.test.ts` is very hairy because I had to do a ton of manual dependency injection to test the outbox functionality without the Sync Engine running.

amhinson (approved):
Looks good to me 👍  Great job on the tests!

scripts/build.js
```diff
@@ -157,7 +157,13 @@ async function buildES5(typeScriptCompiler) {
 	let compilerOptions = {
 		esModuleInterop: true,
 		noImplicitAny: false,
-		lib: ['dom', 'es2017', 'esnext.asynciterable', 'es2018.asyncgenerator'],
+		lib: [
+			'dom',
+			'es2017',
+			'esnext.asynciterable',
+			'es2018.asyncgenerator',
+			'es2019.object',
```
amhinson, line 165: Just curious, why was this needed?

iartemiev: I added `es2019.object` in order to use `Object.fromEntries` [here](https://github.com/aws-amplify/amplify-js/pull/7354/files#diff-89f8afda2fb41bbdaea63170a71ab23374df7b81bef86a6f7cf1eb9af0baea78R456-R457). I have actually found myself needing it before, but had avoided it since we didn't have that in our build target. Figured now's as good of a time as any to add it.

nubpro:
Just want to drop by to say that the included diagrams are supppper duper helpful, I wish there's more......could you guys spill more beans on the rest of datastore ins and outs?

I was gonna try to understand the solution u guys came out with and honestly, I wasn't able to totally comprehend them.
Mad props to @iartemiev and the team for working on this!!

PR #7375, opened 2020-12-09
docs(CONTRIBUTING.md): update verdaccio instructions

wei:
_Issue #, if available:_
N/A

_Description of changes:_
The current verdaccio publish instructions do not work. This is because lerna is using yarn not npm. This PR updates docs so contributors can get verdaccio set up without surprises.


iartemiev (approved):
Thanks @wei!

PR #7400, opened 2020-12-13
Docs updated: define the variable before using it.

ChihweiLHBird:
I believe there should be a `var` because the variable was not defined before in the document.

### DIFF
--- a/packages/amazon-cognito-identity-js/README.md
+++ b/packages/amazon-cognito-identity-js/README.md
@@ -843,7 +843,7 @@ cognitoUser.authenticateUser(authenticationDetails, {
 **Use case 28.** Enabling and setting SMS MFA as the preferred MFA method for the user.
 
 ```js
-smsMfaSettings = {
+var smsMfaSettings = {
 	PreferredMfa: true,
 	Enabled: true,
 };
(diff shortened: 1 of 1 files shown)


wlee221 (changes requested):
Thanks for the catch @ChihweiLHBird! 

Can you rebase your fork to latest `main`? This contains a fix that would resolve the ci build failure. While you are on this, I see that `Use Case 28` and `Use Case 29` are also missing `var` keyword on variables `smsMfaSettings` and `totpMfaSettings` respectively. Can you add `var` to them too? I'll be happy to merge this after so 👍

ChihweiLHBird:
> Thanks for the catch @ChihweiLHBird!
> 
> Can you rebase your fork to latest `main`? This contains a fix that would resolve the ci build failure. While you are on this, I see that `Use Case 28` and `Use Case 29` are also missing `var` keyword on variables `smsMfaSettings` and `totpMfaSettings` respectively. Can you add `var` to them too? I'll be happy to merge this after so 👍

Thank you, and I have added `var` on them. My fork is on the latest branch, not sure why the CI wasn't passes. Maybe rerun it?

Issue #7427, opened 2020-12-18
Missing loader on "Create Account"


ignaciolarranaga:
**Describe the bug**
When you hit create account there is no "loader" (as with Login for example), so the user might attempt to hit the button again or get confused because nothing happens while really the call is already in progress and just taking more time than expected.

Same happens for the verification (in my case through SMS message)

**To Reproduce**

Following any tutorial can be reproduced (e.g. https://docs.amplify.aws/lib/auth/getting-started/q/platform/js#enable-sign-up-sign-in-and-sign-out)
Just use the create account functionality and pay attention when you hit the button.

**Expected behavior**
A loading/loader is displayed.

**Code Snippet**
N/A

**Screenshots**
<img width="529" alt="Missing Loader" src="https://user-images.githubusercontent.com/15932965/102628348-4a8a3180-4128-11eb-8868-0d6d842625cc.png">

**What is Configured?**
N/A

**Smartphone (please complete the following information):**
- Device: Desktop
- OS: MacOS
- Browser Chrome
- Version  87.0.4280.88

package.json:
- "@aws-amplify/ui-react": "^0.2.32",
- "aws-amplify": "^3.3.12",

**Additional context**
N/A

sammartinez:
@ignaciolarranaga Thanks for providing this feedback. Are you looking for the button to be disabled while the request is being made? Just wanted some clarity prior to placing this into our backlog and getting it groomed.

ignaciolarranaga:
I think it has to be disabled yes (just to prevent resubmission I believe), but my mine concern is the user experience.

sammartinez:
Thanks for the feedback here @ignaciolarranaga, we will get this groomed and look where this will fit into the upcoming sprints

SijmenHuizenga:
Hi there!  Today I felt inspired to contribute to this project. This issue seemed like something I could solve so I fixed the bug and created pr #7436. I hope this helps :rocket:

wlee221:
Closing as @SijmenHuizenga's PR has been merged. Thanks!

PR #7436, opened 2020-12-22
fix(@aws-amplify/ui-components): display spinner during submission of sign-up and sign-up-confirm form

SijmenHuizenga:
Solves issue #7427

_Description of changes:_

When submitting the sign-up form, a spinner should be shown. Most of the code is already there, this PR just finishes it. This just updates the `loading` property according to the submission status. 

In the sign-up-confirm form, the `loading` property is set correctly but never used in UI. Fixed this by feeding it into `amplify-form-section`.


wlee221 (approved):
This LGTM, thanks!

PR #7442, opened 2020-12-23
fix: watch for user prop update

### LINKED ISSUE #7406: Cannot read property 'completeNewPasswordChallenge' of undefined
The intent of my app is to log in to a AWS Cognito User Pool where new users are created by the admin and sent a temporary password.

The user pool was created with the Amplify CLI.

This code is pretty much copied and pasted from the AWS Amplify documentation:

```
import React from "react";
import { AmplifyAuthenticator, AmplifyRequireNewPassword } from "@aws-amplify/ui-react";
import './App.css';

Amplify.configure(awsconfig);

function App() {
  return (
      <AmplifyAuthenticator>
          <AmplifyRequireNewPassword
              headerText="My Custom Require New Password Text"
              slot="require-new-password"
          ></AmplifyRequireNewPassword>
          Logged In
      </AmplifyAuthenticator>
  );
}

export default App;
```

After creating a new user in the Cognito console, on a test run of the app I successfully signed in and was prompted to change my password. However, when I submitted the new password, I was presented with the following error:

Cannot read property 'completeNewPasswordChallenge' of undefined
The browser was Microsoft Edge 86.0

wlee221:
_Issue #, if available:_ Fixes #7406. 

_Description of changes:_ When customer uses `amplify-require-new-password` with `slots`, `user` prop is updated after `componentWillLoad` lifecycle fires. Therefore, `currentUser` doesn't get defined and ends up in the #7406 error. This PR watches for `user` prop change and initializes `currentUser` on change.


packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -47,31 +47,38 @@ export class AmplifyRequireNewPassword {
...
+        console.log(user);
         if (user) this.currentUser = user;
       } catch (error) {
         dispatchToastHubEvent(error);
       }
+    } else {
+      this.currentUser = this.user;
     }
-    if (this.currentUser && this.currentUser.challengeParam.requiredAttributes) {
+    if (this.currentUser && this.currentUser.challengeParam && this.currentUser.challengeParam.requiredAttributes) {
```
wlee221, line 79: This can be simplified to `this.currentUser?.challengeParam?.requiredAttributes`, but `eslint` is throwing syntax error because it is referencing monorepo `typescript@3.6.4`. I'll follow up with another pr that bumps monorepo typescript version.

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -47,31 +47,38 @@ export class AmplifyRequireNewPassword {
...
-  private requiredAttributes: object = {};
 
+  @Watch('user')
+  watchHandler() {
+    this.setCurrentUser();
+  }
+
+  private requiredAttributes: object = {};
   private newFormFields: FormFieldTypes = this.formFields;
+  private currentUser: CognitoUserInterface = null;
```
wlee221, line 60: `currentUser` has no reason to be a state, so I changed it to a private variable.

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -47,18 +47,23 @@ export class AmplifyRequireNewPassword {
     },
   ];
 
-  @State() currentUser: CognitoUserInterface = this.user;
   @State() password: string;
   @State() loading: boolean = false;
-  private requiredAttributes: object = {};
```
harrysolovay, line 53: Use a `Record<PropertyKey, any>` (or preferably a narrower type) instead of `object`. `object` will allow seemingly-incompatible types. For instance.

```ts
const x: object = [];
``

File: packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx [hunk 4]
```diff
@@ -47,18 +47,23 @@ export class AmplifyRequireNewPassword {
...
 
+  @Watch('user')
+  watchHandler() {
+    this.setCurrentUser();
+  }
+
+  private requiredAttributes: object = {};
   private newFormFields: FormFieldTypes = this.formFields;
+  private currentUser: CognitoUserInterface = null;
```
harrysolovay, line 60: Why assign `null`? If we had a stricter config, that would be a type error. Why not just leave uninitialized? It's not preferable, but we've disabled strict property initialization. So this'll do the trick:

```ts
private currentUser: CognitoUserInterface;
```

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
harrysolovay: I believe there's a reason we haven't already done this. You should first follow up with @sammartinez about TS version.

harrysolovay (changes requested):
Looking pretty good. Just a few minor comments :)

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
wlee221: Good point, I will!

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
wlee221: Agreed, will do!

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
wlee221: you're right, fixing it now.

harrysolovay (approved):
Sweet! LGTM

wlee221:
Thanks @harrysolovay for the comments!

PR #7466, opened 2020-12-30
fix(@aws-amplify/datastore): only include changed fields in update mutation input

iartemiev:
_Description of changes:_
The current implementation of DataStore will include all of the fields for a model in the mutation input for update mutations. This prevents the AutoMerge conflict resolution strategy from working as expected.

This PR relies on Immer patches to discern which fields were modified by the user and only sends the changed fields in the update mutation query input variables.

(Also, bumps Immer to latest, which fixes: #7598)

If we have the following schema
```gql
type Todo @model {
  id: ID!
  name: String!
  completed: Boolean
}
```

And we perform an update
```js
const current = await DataStore.query(Todo, '0a121e7d-188c-4773-81b6-d3bd30a31c05');

DataStore.save(Todo.copyOf(current, updated => {
  updated.name = 'updated';
}));
```

The resulting GraphQL mutation will be:

**Currently**
```json
{
    "query": "mutation operation($input: UpdateTodoInput!, $condition: ModelTodoConditionInput) {\n  updateTodo(input: $input, condition: $condition) {\n    id\n    name\n    completed\n    owner\n    _version\n    _lastChangedAt\n    _deleted\n  }\n}\n",
    "variables": {
        "input": {
            "id": "0a121e7d-188c-4773-81b6-d3bd30a31c05",
            "name": "updated",
            **"completed": null,**
            "_version": 1
        },
        "condition": null
    }
}
```
**With PR**
```json
{
    "query": "mutation operation($input: UpdateTodoInput!, $condition: ModelTodoConditionInput) {\n  updateTodo(input: $input, condition: $condition) {\n    id\n    name\n    completed\n    owner\n    _version\n    _lastChangedAt\n    _deleted\n  }\n}\n",
    "variables": {
        "input": {
            "id": "0a121e7d-188c-4773-81b6-d3bd30a31c05",
            "name": "updated",
            "_version": 1
        },
        "condition": null
    }
}
```


packages/datastore/src/util.ts
```diff
@@ -439,19 +439,25 @@ export const isAWSDate = (val: string): boolean => {
...
+	return !!/^\d{2}:\d{2}(:\d{2}(.\d{3})?)?(Z|[+-]\d{2}:\d{2}($|:\d{2}))?$/.exec(
+		val
+	);
 };
 
 export const isAWSDateTime = (val: string): boolean => {
-	return !!/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}(:\d{2}(.\d{3})?)?(Z|[+-]\d{2}:\d{2}($|:\d{2}))?$/.exec(val);
+	return !!/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}(:\d{2}(.\d{3})?)?(Z|[+-]\d{2}:\d{2}($|:\d{2}))?$/.exec(
+		val
+	);
```
nubpro, line 450: is something off with your formatter?

iartemiev: No. Mine is working as expected. 
The contributor who originally added that code must have not had theirs configured correctly.

nubpro: Cool, I believe it was @wei

amhinson (approved):
LGTM 👍

packages/datastore/src/storage/storage.ts
```diff
@@ -98,17 +99,33 @@ class StorageClass implements StorageFacade {
...
 		result.forEach(r => {
-			const [element, opType] = r;
+			const [originalElement, opType] = r;
+
+			let updatedElement;
+			if (opType === OpType.UPDATE && patches && patches.length) {
+				const { id, _version, _lastChangedAt, _deleted } = originalElement;
+
+				// For update mutations we only want to send fields with changes
+				// As well as the internal fields
```
manueliglesias, line 117: What about fields that weren't updated but are marked as `required: true` in the schema?

manueliglesias (changes requested):
It is looking good!

I would like to see some unit tests, specially around handling of fields of type `array` and the scenario where required fields weren't updated but they should still be sent to the mutation.

packages/datastore/src/storage/storage.ts
iartemiev: Those fields are only required when performing create mutations. For update mutations, AppSync only requires `id` and `_version`

iartemiev:
@manueliglesias agreed re: tests. I don't think it'll be possible to validate the functionality all the way through via unit tests (will be adding an integration test for that) but should be able to at least test the relevant changes in storage.ts and mutation.ts.

Regarding required fields, AppSync does not expect them on an update. They're only required when creating a new record. WRT arrays/lists, this PR treats them the same way as we are now, i.e., sending the entire updated list as the value. Will be adding tests around these scenarios (:

packages/datastore/__tests__/DataStore.ts
```diff
@@ -1,4 +1,5 @@
 import 'fake-indexeddb/auto';
+import FDBCursor from 'fake-indexeddb/build/FDBCursor';
```
manueliglesias, line 2: I didn't see where this is being used

packages/datastore/__tests__/DataStore.ts
```diff
@@ -378,7 +389,144 @@ describe('DataStore tests', () => {
 
 			const result = await DataStore.save(model);
 
+			const [settingsSave, modelCall] = <any>save.mock.calls;
+			const [_model, _condition, _mutator, patches] = modelCall;
+
+			expect(result).toMatchObject(model);
+			expect(patches).toBeUndefined();
```
manueliglesias, line 396: Why would we want `undefined` vs `[]`?

packages/datastore/src/datastore/datastore.ts
```diff
@@ -506,7 +525,10 @@ async function checkSchemaVersion(
 			{ page: 0, limit: 1 }
 		);
 
-		if (schemaVersionSetting !== undefined) {
+		if (
+			schemaVersionSetting !== undefined &&
+			schemaVersionSetting.value !== undefined
+		) {
```
manueliglesias, line 531: Is this block fixing a bug? (the new condition).

packages/datastore/src/storage/storage.ts
```diff
@@ -98,17 +99,45 @@ class StorageClass implements StorageFacade {
...
 
 		result.forEach(r => {
-			const [element, opType] = r;
+			const [originalElement, opType] = r;
+
+			let updatedElement;
+			if (opType === OpType.UPDATE && patches && patches.length) {
+				updatedElement = {};
+				// extract array of updated fields from patches
+				const updatedFields = patches.map(patch => patch.path && patch.path[0]);
```
manueliglesias, line 116: I see in [immer docs](https://immerjs.github.io/immer/docs/patches) that `path` is an array, why do we take just `[0]`? I am assuming that it is because we handle non-model fields separately, is this assumption correct?

Do we need something special here to handle setting non-model fields? (e.g. `myModel.someField.someOtherField`)

> The generated patches are similar (but not the same) to the RFC-6902 JSON patch standard, except that the path property is an array, rather than a string. This makes processing patches easier. If you want to normalize to the official specification, patch.path = patch.path.join("/") should do the trick

manueliglesias (approved):
LGTM 👍 🎉 

Thanks @iartemiev , I just left a couple of questions

packages/datastore/__tests__/DataStore.ts
iartemiev: Good catch. I think VSC auto-imported that for me. Will delete

packages/datastore/__tests__/DataStore.ts
iartemiev: Patches are set inside of `copyOf` so we should only expect a value for updates. For saves, it will be `undefined`.

packages/datastore/src/datastore/datastore.ts
iartemiev: It's a bug that happened when running some of the tests I had added. It was breaking on the next line (532) when attempting to JSON parse an undefined value.

packages/datastore/src/storage/storage.ts
iartemiev: Manuel and I synced on this offline. Basically, the path can contain either the field name `["someField"]` - for replace operations. Or for add/delete, it will be the field name and index, e.g., `["someField", 2]`. All we care about is the field name because we’re not actually applying the patches, but just using them to infer which fields were changed. Hence why we're grabbing the 0th element.

Issue #7477, opened 2021-01-03
Typescript Auth SDK doesn't support federated sign in with user pools


jgiunta1:
**Describe the bug**
I'm able to sign in with `Auth.federatedSignIn({provider: 'Facebook'});` in js, but I can't use that in typescript. `Auth.federatedSignIn({provider: CognitoHostedUIIdentityProvider.Facebook})` fails because it calls cognito-identity instead of cognito-idp. My local storage isn't updated with the cognito tokens and I get a 404 from `cognito-identity.us-east-1.amazonaws.com`. js does not call this endpoint ever, not sure why the ts SDK is calling it, or why there is no user-pool support for this. my export file is the same for both js and ts. The only difference is the federated sign in call above.

**To Reproduce**

1. setup amplify with user pools and facebook identity provider
2. create a new next ts project
3. import amplify newest version: `import { Auth } from 'aws-amplify';`
4. Call `Auth.federatedSignIn({provider: CognitoHostedUIIdentityProvider.Facebook});`
5. Open your network tab in chrome and see that cognito-identity is being called. even though the aws-export file should have `"federationTarget": "COGNITO_USER_POOLS"`

**Expected behavior**
The vanilla js version of this works. `Auth.federatedSignIn({provider: 'Facebook'});` works as expected. You can go to the application tab in chrome and see local storage updated. CognitoIdentityServiceProvider object will be under the local storage tab.

**Code Snippet**
Please provide a code snippet or a link to sample code of the issue you are experiencing to help us reproduce the issue. (Be sure to remove any sensitive data)

Working: `Auth.federatedSignIn({provider: 'Facebook'});` in js 
Not working: `Auth.federatedSignIn({provider: CognitoHostedUIIdentityProvider.Facebook})` in tsx

**What is Configured?**
If applicable, please provide what is configured for Amplify CLI:
* Which steps did you follow via Amplify CLI when configuring your resources.
* Which resources do you have configured?
  * If applicable, please provide your `aws-exports` file:
   ```
const awsmobile = {
    "aws_project_region": "us-east-1",
    "aws_cognito_identity_pool_id": "us-east-1:xxx",
    "aws_cognito_region": "us-east-1",
    "aws_user_pools_id": "us-east-1_xxx",
    "aws_user_pools_web_client_id": "xxx",
    "oauth": {
        "domain": "homulator-auth-dev.auth.us-east-1.amazoncognito.com",
        "scope": [
            "phone",
            "email",
            "openid",
[...]

Issue #3611, opened 2019-07-08
Auth.federatedIdentitySignIn({provider: 'Google'}) not working


colehendo:
**Describe the bug**
When I try to use `Auth.federatedIdentitySignIn({provider: 'Google'});` in my code it throws the error: 
`"Argument` of type '{ provider: string; }' is not assignable to parameter of type 'FederatedSignInOptionsCustom'. Object literal may only specify known properties, and 'provider' does not exist in type 'FederatedSignInOptionsCustom'.`

When I use `Auth.federatedIdentitySignIn();` by itself it works fine, but I would like to skip the step of clicking the button.

**To Reproduce**
run `npm i aws-angular`
run following code in a typescript/angular file:
`import Amplify, { Auth } from 'aws-amplify';`
`const oauth = {
    domain: 'XXXXXX.auth.us-east-1.amazoncognito.com/',
    scope: ['email', 'profile', 'openid'],
    redirectSignIn: 'http://localhost:4200/',
    redirectSignOut: 'http://localhost:4200/',
    responseType: 'token'
};`

`Amplify.configure({
    Auth: {
        oauth: oauth,
        region: environment.AWS_REGION,
        userPoolId: environment.AWS_USER_POOL_ID,
        userPoolWebClientId: environment.CLIENT_ID,
    }
});`

`ngOnInit() {
    Auth.federatedSignIn({provider: 'Google'});
}`

**Expected behavior**
It should redirect to an option to sign in with a google email listing all the user's google accounts.

**Desktop (please complete the following information):**
 - OS:  iOS
 - Browser: chrome

Running in visual studio code if that helps at all

aws-amplify version: 1.1.29
@aws-amplify/auth version: 1.2.25
@aws-amplify/core version: 1.0.28

manueliglesias:
Hi @colehendo 

This seems to me as a TypeScript compiler thrown error, is that correct?

Which version of typescript are you using?

The `federatedSignIn` function has [Overloads](https://www.typescriptlang.org/docs/handbook/functions.html#overloads) like so:

https://github.com/aws-amplify/amplify-js/blob/d384e5dd01b2719111c5121538259571730b1218/packages/auth/src/Auth.ts#L1438-L1451

where

https://github.com/aws-amplify/amplify-js/blob/d384e5dd01b2719111c5121538259571730b1218/packages/auth/src/types/Auth.ts#L58-L64

colehendo:
[...]

PR #7478, opened 2021-01-03
fix(aws-amplify) : federated sign provider undefined error.

Jeontaeyun:
_Issue #, if available:_
#7477 
_Description of changes:_
The way I see it, amplify library seem to be unified with `amplify-aws`.  
But when I use `Auth.federatedSignIn({provider: "Google"})` with typescript. 
There is some error like,
```
Type "Google" is not assignable to type "CognitoHostedUIIdentityProvider".
```
So, I have researched some reference, and I know I can use `@aws-amplify/auth` with #3611  like,

```typescript
import {CognitoHostedUIIdentityProvider} from "@aws-amplify/Auth";
```
But What I am thinking is that It is useful to export `CognitoHostedUIIdentityProvider` from `aws-amplify`. 
And when I check `aws-amplify` package, there is many provider enum. So I fix it.
[packages/aws-amplify/src/index.ts | line 33]
### Before
```typescript
export { Auth } from '@aws-amplify/auth';
```

### After
```typescript
export { Auth , CognitoHostedUIIdentityProvider} from '@aws-amplify/auth';
```

### DIFF
--- a/packages/aws-amplify/src/index.ts
+++ b/packages/aws-amplify/src/index.ts
@@ -30,7 +30,7 @@ export {
 	AmazonPersonalizeProvider,
 } from '@aws-amplify/analytics';
 
-export { Auth } from '@aws-amplify/auth';
+export { Auth, CognitoHostedUIIdentityProvider } from '@aws-amplify/auth';
 export { Storage, StorageClass } from '@aws-amplify/storage';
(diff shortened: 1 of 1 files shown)


Jeontaeyun:
Oh can I ask how can I fix it?

PR #7489, opened 2021-01-04
deps: bump axios to 0.21.1

wlee221:
_Issue #, if available:_

_Description of changes:_ This bump is in response to a medium secuirty vulnerability with axios described in https://github.com/axios/axios/pull/3410. This PR bumps axios to 0.21.1 in which the security patch is released. 

Looking at the [changelog](https://github.com/axios/axios/blob/hotfix/0.21.1/CHANGELOG.md) from 0.19.0 to 0.21.1, I don't see any commits that would cause regression, but it would be great if anyone else can confirm this.

### DIFF
--- a/packages/api-rest/package.json
+++ b/packages/api-rest/package.json
@@ -42,7 +42,7 @@
   "homepage": "https://aws-amplify.github.io/",
   "dependencies": {
     "@aws-amplify/core": "3.8.8",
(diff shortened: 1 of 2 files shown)


manueliglesias (approved):
Thanks @wlee221 

I took a look too at the change logs, I couldn't see anything risky

LGTM 👍

PR #7512, opened 2021-01-06
fix(@aws-amplify/ui-components): handle custom totp-setup slots properly

### LINKED ISSUE #7241: AmplifyTotpSetup component renders broken image for QR code
**Describe the bug**
The QR Code image for TOTP Setup fails to render if `AmplifyTotpSetup` is used in `AmplifyAuthenticator`

**To Reproduce**

1. Create auth flow with `<AmplifyAuthenticator >...</AmplifyAuthenticator>`
2. Add `<AmplifyTotpSetup slot="totp-setup" />` child
3. Register a new user or sign in with one that still needs TOTP Setup
4. View broken qr code...

**Expected behavior**
QR Code renders correctly with no overrides, with issuer override, with header text override, etc

**Code Snippet**
```
import React from "react";
import "./App.css";
import {
  AmplifyAuthenticator,
  AmplifySignOut,
  AmplifyTotpSetup,
} from "@aws-amplify/ui-react";
import { AuthState, onAuthUIStateChange } from "@aws-amplify/ui-components";

Amplify.configure(awsconfig);

const AuthStateApp: React.FunctionComponent = () => {
  const [authState, setAuthState] = React.useState<AuthState>();
  const [user, setUser] = React.useState<object | undefined>();

  React.useEffect(() => {
    return onAuthUIStateChange((nextAuthState, authData) => {
      setAuthState(nextAuthState);
      setUser(authData);
    });
  }, []);

  return authState === AuthState.SignedIn && user ? (
    <div className="App">
      <div>Hello, {(user as any).username}</div>
      <AmplifySignOut />
    </div>
  ) : (
    <AmplifyAuthenticator >
      <AmplifyTotpSetup headerText="My Custom TOTP Setup Text" slot="totp-setup" />
    </AmplifyAuthenticator>
  );
};

export default AuthStateApp;
```

**Screenshots**

Working if you don't use the `AmplifyTotpSetup` component:
image

Broken with no overrides `<AmplifyTotpSetup slot="totp-setup" />`
image

Broken with overrides `<AmplifyTotpSetup headerText="My Custom TOTP Setup Text" slot="totp-setup" />`
image

wlee221:
_Issue #, if available:_ Fixes #7241

_Description of changes:_ Handle custom `amplify-totp-setup` elements properly

**Problem**
Custom `amplify-totp-setup` components connected with slots weren't functioning for two reasons. 

First, slotted component gets connected to DOM as soon as `amplify-authenticator` loads. Thus `componentWillLoad` lifecycle gets called prematurely. Because `user` prop isn't defined at that point, `Auth.setupTOTP` fails. This doesn't happen with default `amplify-authenticator` because its subcomponents are connected **after** proper `authState` is set. 

Second, when we use slots to override the default `amplify-totp-setup`, the default one is still in DOM! It's just not visible without any content. The problem is that the default one still has its lifecycles run. In this case, both default and overridden `amplify-totp-setup` call `Auth.setupTOTP` and makes a race condition. Because `Auth.setupTOTP` cannot be called twice in a row, this results in broken QR code image.

These oversights with slots are across the library -- for example, #7442 is caused by this as well.

**Solution**
This PR fixes the two problems by:
- use `@Watch` handler to wait until a valid `user` prop is passed
- Do not render / connect the default `amplify-totp-setup` if it has been overridden with slots.


wlee221:
Closing in favor of #7522.

PR #7500, opened 2021-01-05
Bundle size tracking

ericclemmons:
_Issue #, if available:_  https://www.pivotaltracker.com/story/show/174995888

This PR introduces the foundation to check real-world application bundle sizes within PRs. 

See https://ja2r7.app.goo.gl/6PJJdRZMA6mz3ZWH7 for an example of the results & the GitHub status checks below 👇 

Future improvements would be:

- [ ] Set thresholds on the net ± impact
- [ ] Update the status check with results (https://github.com/mitchheddles/github-action-status-check)

## Naive Implementation

We _could_ simply report on the fully-bundled cost of our own code (i.e. the size of a package if all source was concatenated, excluding dependencies):

```
PASS packages/amazon-cognito-identity-js/dist/amazon-cognito-identity.min.js: 24.3KB < 400KB (gzip)
PASS packages/amplify-ui-react/dist/aws-amplify-react.min.js: 55.31KB < 400KB (gzip)
PASS packages/amplify-ui/dist/aws-amplify-ui.min.js: 1.53KB < 400KB (gzip)
PASS packages/analytics/dist/aws-amplify-analytics.min.js: 55.34KB < 400KB (gzip)
PASS packages/api-graphql/dist/aws-amplify-api-graphql.min.js: 37.89KB < 400KB (gzip)
PASS packages/api-rest/dist/aws-amplify-api-rest.min.js: 12.57KB < 400KB (gzip)
PASS packages/api/dist/aws-amplify-api.min.js: 38.91KB < 400KB (gzip)
PASS packages/auth/dist/aws-amplify-auth.min.js: 164.98KB < 400KB (gzip)
PASS packages/aws-amplify-react/dist/aws-amplify-react.min.js: 50.56KB < 400KB (gzip)
PASS packages/aws-amplify-vue/dist/aws-amplify-vue.umd.min.js: 48.02KB < 400KB (gzip)
PASS packages/aws-amplify/dist/aws-amplify.min.js: 346.76KB < 400KB (gzip)
PASS packages/cache/dist/aws-amplify-cache.min.js: 3.97KB < 400KB (gzip)
PASS packages/core/dist/aws-amplify-core.min.js: 42.63KB < 400KB (gzip)
PASS packages/datastore/dist/aws-amplify-datastore.min.js: 255.19KB < 400KB (gzip)
PASS packages/interactions/dist/aws-amplify-interactions.min.js: 23.86KB < 400KB (gzip)
[...]

Issue #7313, opened 2020-12-01
Considerable bundle size increase when upgrading from @aws-amplify/auth@3.2.4 to @aws-amplify/auth@3.4.13


huoxito:
Hi!

We're trying to upgrade to latest `@aws-amplify/auth` and we noticed a considerable build size increase. Our vendor bundle went from 213,58KB to 304,46KB.

Looking into npm it seems the unpacked size went from [2.16 MB](https://www.npmjs.com/package/@aws-amplify/auth/v/3.2.4) to [6.47 MB](https://www.npmjs.com/package/@aws-amplify/auth/v/3.4.14). Any chance there's a way to decrease this bundle size? Maybe opt out some features not needed for some apps. Is there any on going project / task to try decrease that bundle size?

We haven't found any bugs with 3.4.13, so far all good. But since this build increase affects our TTI I was hoping to understand better what happened or how we could make it better moving forward with amplify/js.

kushmangal:
Any update on this?

quinnturner:
More specifically, it's between `3.4.0` and `3.4.2`, or `aws-amplify` `3.1.0` and `3.2.0`.

On `aws-amplify` `3.1.0`, my bundle size is `216kb`. On `aws-amplify` `3.2.0`, my bundle size is `301kb`.

PaulTime:
still having problems with this issue 
![image](https://user-images.githubusercontent.com/26630246/103099983-3dce7780-4619-11eb-80be-a2bef8ea04d9.png)

bundle size increased from 200kb gzip to 440kb gzip after I installed @aws-amplify/auth and called configure() method

PaulTime:
![image](https://user-images.githubusercontent.com/26630246/103100429-24c6c600-461b-11eb-82e0-898a9eb7405a.png)

picardplaisimond:
This is a very important problem in my opinion. I have to support this.
In my case, I only use:
`import Amplify from '@aws-amplify/core'`
`import Auth from '@aws-amplify/auth'`

It increases the size of the application by ***244kb*** ! It's way too much to just handle authentication !
Chrome DevTools tells us that most of the javascript isn't even executed. So it's an optimization problem.

From what I've seen, if we use the `Auth` module, all the functions of this class are imported. There must be a way to optimize that. Maybe with more targeted *dynamic imports* in that class? I'm talking about a big refactoring.

I found nothing else to fix it. If anyone has a better solution, that would be great.
[...]

PR #7521, opened 2021-01-08
fix(amazon-cognito-identity-js): avoid unnecessary crypto polyfills in webpack builds

nnyath:
## _Issue #, if available:_
https://github.com/aws-amplify/amplify-js/issues/7313

## _Description of changes:_
Webpack bundle sizes are unnecessary large when using anything that has a dependency/sub-dependency on `amazon-cognito-identity-js@^4.5.2` or `@aws-amplify/auth@^3.4.9`

This is due to Webpack using polyfills for crypto node libs (https://stackoverflow.com/a/45862270). In this case it was added as a dynamic require in this PR https://github.com/aws-amplify/amplify-js/pull/7136.

Even though it's a conditional `require`, Webpack will still evaluate the line and include it into the final build raising bundle sizes by _at least_ `~125kb (gzipped)`

## _Recreate:_
You can recreate this with a fresh CRA install and adding any one of these deps (but there's probably more affected by this)

```
import { API, Auth, PubSub } from 'aws-amplify'

import API from '@aws-amplify/api'
import Auth from '@aws-amplify/auth'
import PubSub from '@aws-amplify/pubsub'

import * as CognitoJS from 'amazon-cognito-identity-js'
```
## _Analysis:_
**NOTE: Sizes are gzipped**

A simple Hello World bundle size gives me
`41.21 KB`

Adding `@aws-amplify/auth` gives me an increase of
`235.4 KB (+194.19 KB)`

Here's what the `source-map-explorer` output is (notice a lot of `browserify-sign`, `bn.js`, `elliptic`, etc)
Screen Shot 2021-01-07 at 6 31 22 PM
which corresponds to the SO answer referenced above.

After this fix, the bundle drops a respectable amount
`108.95 KB (-126.44 KB)`

Here's the `source-map-explorer` after the fix which looks much better.
Screen Shot 2021-01-07 at 6 38 05 PM
Though, I think there still might be a few other polyfills being added due to seeing `node-lib-browser`

## _Caveats:_
This will only work for Webpack configs that specifically look for and apply `.web.` files, which fortunately CRA has supported for a while now. But other custom/dated configs or tooling that reference the ESM output could still run into this issue. ***Edit** - Not familiar with other tooling options (rollup), but I suppose it's up to how they generally handle references to node libs and whether or not to error out or provide a polyfill

Here's some specific Webpack things that might work
- https://webpack.js.org/guides/code-splitting/#dynamic-imports
- https://webpack.js.org/api/module-methods/#requireensure
- Conditional based on`process.env` would omit the require, but not enforceable enough I think

but again, these seem specific to Webpack, and ~~probably don't belong in the source code~~. Not sure if there's a universal fix here, but maybe someone more knowledgeable in this area can chime in.

***Edit** - Might be fine since consumers shouldn't need to worry about the building the source code using configs/tooling outside of this repo, as long as the build outputs correctly we should be good

***Edit** - The other alternative would be to only use the crypto functions needed and providing those via a cross platform compatible lib/implementation as suggested by the SO answer.

However, I think the current proposed approach of adding in `.web.` variant in is a step in the right direction and not likely to introduce additional regressions.

I also see there's some work on monitoring bundle sizes in https://github.com/aws-amplify/amplify-js/pull/7500 which is great, ~~but I think an issue like this would have flown under the radar.~~
***Edit**: Overlooked the real world bundle size section, I would think this would be caught in that


amhinson (approved):
Thanks for this @nnyath! 🙏 

I was able to validate the bundle size improvements with a CRA app. As you mentioned, there is still room for improvement regarding this `crypto` module as it relates to other custom configurations and framework defaults, but this seems like a quick win right now for CRA developers. Because of the various other platforms that this library supports, implementing any of the other alternative solutions you provided would be rather difficult or impossible at this time. 

I'd also like to get @ericclemmons's thoughts on this before merging.

ericclemmons (approved):
I'll add this test to #7500 as well!

I'm 👍  for including this.  Unfortunately, it doesn't fix "default" behavior with `webpack@4`, but it will benefit CRA users, who make up a sizeable % of installs.

PR #7522, opened 2021-01-08
fix(@aws-amplify/ui-components): handle slotted elements properly

### LINKED ISSUE #7241: AmplifyTotpSetup component renders broken image for QR code

wlee221:
_Issue #, if available:_ Fixes #7241, #7406

_Description of changes:_ Handles custom implementation of slottable elements properly.

**Problem**

`amplify-authenticator` provides auth subcomponents that can be overridden custom implementation. This hasn't really been working for two reasons:

1) Custom component gets connected to DOM *as soon as* amplify-authenticator loads.

Normally, auth subcomponents render *after* a proper `authStateChange` occurs. But slotted element connected immediately. Although invisible, it still calls its lifecycle hooks where it runs setups. This breaks auth subcomponents that depends on `user` prop, because `user` is only defined after proper `authStateChange` is run.

2) Even when an element is slotted, the default one is still in the DOM along with the custom one!

Although the overridden element is hidden, it still remains in DOM and calls all the lifecycle methods as usual, on chrome. Hence there exist duplicate components, and results in race condition when they the same resource / API. (e.g. duplicate `amplify-setup-totp` calling `Auth.setupTOTP fails with concurrency exception).

This is a big problem across all auth subcomponents. Customer reports:
- `amplify-require-new-password`: #7406
- `amplify-setup-totp`: #7241 

**Solution**
1. To resolve premature DOM connection, we use the `watch` decorator to run setup when prop it needs is changed or available.
2. To resolve duplicate components, we do not render the auth subcomponent if it has been slotted. Instead, we just render a named slot without content.


packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -135,69 +139,45 @@ export class AmplifyAuthenticator {
...
-      case AuthState.SignedIn:
-        return [<slot name="greetings"></slot>, <slot></slot>];
+        return <div>Loading...</div>;
       default:
         throw new Error(`Unhandled auth state: ${authState}`);
     }
   }
 
+  // Returns a slot containing the Auth component corresponding to the given authState
+  private getSlotWithAuthCompoent(authState: AuthState): JSXBase.IntrinsicElements {
```
sammartinez, line 169: Can you fix the typo please :)

wlee221: Done 👍

wlee221:
If we decide to remove shadow dom (see [discussion](https://github.com/aws-amplify/amplify-js/discussions/7486)), problem 2 will be resolved. Problem 1 will likely be resolved too, but I'd have to confirm that since stencil's slots seem to behave differently without shadow dom.

After we drop shadow dom support, we would remove `slotIsEmpty` logic on https://github.com/aws-amplify/amplify-js/blob/a4290f0c525ce20afaec59d8a25349e2c4a0c92d/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx#L178
and just render `<slot name={slotName}>{authComponent}</slot>`.

I think `@Watch` decorators can remain here, because user prop changes on every auth UI event and it's a good idea to run setup again because setup depends on `user` prop.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -217,9 +197,9 @@ export class AmplifyAuthenticator {
           />
         ) : null}
         {this.authState === AuthState.SignedIn ? (
-          this.renderAuthComponent(this.authState)
+          [<slot name="greetings"></slot>, <slot></slot>]
```
tjleing, line 200: What's the reasoning behind doing this array as opposed to two JSX lines?  I'm guessing there's a difference between this and 
```
? (
<slot name="greetings"></slot>
<slot></slot>
)
```
 that I just don't know of.

packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx
```diff
@@ -60,25 +66,36 @@ export class AmplifyConfirmSignUp {
...
   }
 
+  @Watch('user')
+  userHandler() {
+    this.setup();
+  }
+
+  private setup() {
+    this.userInput = this.user ? this.user.username : null;
+    this._signUpAttrs = this.user && this.user.signUpAttrs ? this.user.signUpAttrs : null;
```
tjleing, line 94: Looks like you _can_ do
```
this.userInput = this.user && this.user.username;
this._signUpAttrs = this.user && this.user.signUpAttrs;
```
 -- the one downside is that if the property doesn't exist, it'll be `undefined` instead of `null`, which is still falsey and would have the same behavior in an `if(!this._signUpAttrs)` statement.  The other downside is that it might be a little harder to read.

Either way, the two statements should match (i.e. both `this.user && ... ? ...` or `this.user ? ...`)!

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up-interface.ts
```diff
@@ -1,14 +0,0 @@
-export interface AmplifySignUpFormFooterProps {
```
tjleing, line 1: Was this just totally unused?

packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx
```diff
@@ -8,7 +8,13 @@ import {
 } from '../amplify-auth-fields/amplify-auth-fields-interface';
 import { NO_AUTH_MODULE_FOUND, COUNTRY_DIAL_CODE_DEFAULT } from '../../common/constants';
 import { Translations } from '../../common/Translations';
-import { AuthState, CognitoUserInterface, AuthStateHandler, UsernameAliasStrings } from '../../common/types/auth-types';
+import {
+  AuthState,
+  CognitoUserInterface,
+  AuthStateHandler,
+  UsernameAliasStrings,
+  AmplifySignUpAttributes,
```
tjleing, line 16: `AmplifySignUpAttributes` looks slightly out of place here compared to all of the other imported names -- to my eyes, anyways.  `CognitoUserInterface` is prefixed by `Cognito` because it's not Amplify, but everything Amplify-related should just be the feature itself imo.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -135,69 +139,45 @@ export class AmplifyAuthenticator {
...
+        return <div>Loading...</div>;
       default:
         throw new Error(`Unhandled auth state: ${authState}`);
     }
   }
 
+  // Returns a slot containing the Auth component corresponding to the given authState
+  private getSlotWithAuthComponent(authState: AuthState): JSXBase.IntrinsicElements {
+    const authComponent = this.getAuthComponent(authState);
+    const slotName = authSlotNames[authState];
```
tjleing, line 171: This makes me wish that the slot names were directly equal to the auth states, instead of redirected through this map...  Although it does seem like we can't change it now due to backwards compatibility.

packages/amplify-ui-components/src/common/types/auth-types.ts
```diff
@@ -74,6 +74,14 @@ export interface CognitoUserInterface {
...
+  username: string;
+  password: string;
+  attributes?: {
+    [userAttributes: string]: string;
+  };
+}
+
 export type AuthStateHandler = (nextAuthState: AuthState, data?: object) => void;
 
 export enum MfaOption {
```
tjleing, line 87: Unrelated, but why is this one `MfaOption` when the next one is `MFATypesInterface`...

tjleing (commented):
Looks good!  I don't have much on the actual logic, but I sure do have a few tangentially related comments for you 😄

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
wlee221: I tried that too! But typescript was complaining with `JSX expressions must have one parent element`. Also fragments `<> ... </>`  aren't available if `jsxFragmentFactory` option is set to true. So I wrapped them in an array 😛

packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx
wlee221: Good point! I'll use `?` pattern in favor of `&&` for readability.

wlee221: It would be great if we can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), but we would have to bump the monorepo typescript version to `>=3.7.0`  for `eslint` to parse it (eslint references the monorepo ts instead of the one in ui-components for some reason).

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up-interface.ts
wlee221: Yup 🙃

packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx
wlee221: You're right, good catch for code quality! Will adjust.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
wlee221: Right. https://github.com/aws-amplify/amplify-js/blob/760374ee94e740d8772db9913eebb5019d0d7c97/packages/amplify-ui-components/src/common/types/auth-types.ts#L3-L7
These values can be set to slot names instead, but this would still be a breaking change.

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up-interface.ts
tjleing: 🙃 indeed...

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
tjleing: Oh that's so weird... glad you fixed it!

packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx
tjleing: Sounds good!

packages/amplify-ui-components/src/common/types/auth-types.ts
tjleing: Probably doesn't fit in this PR, but maybe another one?

packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx
wlee221: On a second thought, I'm going with your code snippet because ternary operators were getting out of hand: 
`this.user ? (this.user.username ? this.user.username : null) : null`

packages/amplify-ui-components/src/common/types/auth-types.ts
wlee221: Yeah, they could be better named. Let's make a new PR for that.

wlee221:
Thanks for the review @tjleing! Committed the suggested changes.

tjleing (approved):

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
```diff
@@ -135,69 +139,45 @@ export class AmplifyAuthenticator {
...
-        );
+        return <amplify-totp-setup user={this.authData} />;
       case AuthState.Loading:
-        return (
-          <slot name="loading">
-            <div>Loading...</div>
-          </slot>
-        );
-      case AuthState.SignedIn:
-        return [<slot name="greetings"></slot>, <slot></slot>];
```
elorzafe, line 195: What happen with `greetings`?

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
elorzafe, line 200: Why you had to move `greetings` here instead of `getAuthComponent` maybe I am missing something

packages/amplify-ui-components/src/components/amplify-totp-setup/__snapshots__/amplify-totp-setup.spec.ts.snap
```diff
@@ -5,7 +5,6 @@ exports[`amplify-totp spec: Render logic -> should render a \`less than 2 mfa ty
   <mock:shadow-root>
     <amplify-form-section headertext="Scan then enter verification code" submitbuttontext="Verify Security Token">
       <div class="totp-setup">
-        <img alt="qrcode">
```
elorzafe, line 8: What happened with this snap?

elorzafe (commented):
Looks good @wlee221, just a couple of questions before approve

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
wlee221: I'll answer in the comment below.

packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
wlee221: `greetings` is an edge case where it has two empty slots without any contents like the other ones. This is really for the user app content after sign-in is finished, so I left it out of `getAuthComponent`.

packages/amplify-ui-components/src/components/amplify-totp-setup/__snapshots__/amplify-totp-setup.spec.ts.snap
wlee221: It's from this diff: https://github.com/aws-amplify/amplify-js/pull/7522/files#diff-af4e2f59620b5baf02e9ffff96af7f49ae65121c51335003ffff49f6a8ec4bfeL133-R141. We no longer render an image if the qr code src has been resolved yet. This has been causing broken image to show briefly before correct image appears. 

On this snap, setupTOTP doesn't run so the image no longer exists.

elorzafe (approved):
@wlee221 thanks!!! 🎉 🌮 🏅

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -85,13 +83,17 @@ export class AmplifyTOTPSetup {
...
+      this.code = this.buildOtpAuthPath(this.user, encodedIssuer, secretKey);
       this.generateQRCode(this.code);
     } catch (error) {
-      dispatchToastHubEvent(error);
+      /**
+       * When user has slotted totp-setup and signs out, this component unfortunately still remains in DOM and calls setup
+       * whenever `user` prop updates. The problem is that `this.user` can lose its value during the execution and throw
+       * an error here. We only display toast message if this is not the case, ie. `this.user` is still present.
+       */
+      if (this.user) dispatchToastHubEvent(error);
```
wlee221, line 96: Just found one more bug today... Slotted `amplify-totp-setup` still remains in DOM after it has been unrendered. The problem is that `this.user` prop can update *during* the execution of setup and will dispatch an error toast message. This code detects such cases and hides this message (bc the error is from an unrendered component and hence negligible).

packages/amplify-ui-components/src/components/amplify-totp-setup/amplify-totp-setup.tsx
```diff
@@ -36,14 +35,22 @@ export class AmplifyTOTPSetup {
...
+
+  async componentWillLoad() {
+    /**
+     * We didn't use `@Watch` here because it doesn't fire when we go from require-new-password to totp-setup.
+     * That is because `Auth.completeNewPassword` only changes `user` in place and Watch doesn't detect changes
+     * unless we make a clone.
+     */
+    this.removeHubListener = onAuthUIStateChange(authState => {
+      if (authState === AuthState.TOTPSetup) this.setup();
+    });
```
elorzafe, line 48: Great idea!

elorzafe (approved):
Thanks @wlee221 !! 🌮 🎉 🥇

wlee221:
Thanks for all the reviews!

PR #7408, opened 2020-12-16
feat(@aws-amplify/ui-components): Trim inputs before submit

wlee221:
_Issue #, if available:_ Fixes #6737

_Description of changes:_ When user inputs have trailing or leading whitespace, Cognito returns 
```
Value at 'username' failed to satisfy constraint: Member must satisfy regular expression pattern: [\p{L}\p{M}\p{S}\p{N}\p{P}]+
```
which is not readable to end users. This pr trims inputs when forms are submitted to avoid these errors. We do not trim password and instead throw an error. 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7543, opened 2021-01-11
fix(@aws-amplify/ui-components): throw an error for invalid inputs

wlee221:
_Issue #, if available:_

_Description of changes:_ This is a follow-up fix to #7408. We should throw on whitespace errors instead of just dispatching the error to toast. Otherwise the code keeps executing and calls Cognito API, which throws an regex error and overwrites the friendly error message.


elorzafe (approved):
Looks good to me

Thanks @wlee221 ! 👍

Issue #7554, opened 2021-01-13
e.getSession is not a function after authenticating using withAuthenticator with federated identity


pnewhook:
**Describe the bug**
After setting up a new application and authenticating using `withAuthenticator(App, { federated: { facebookAppId: "XYZ" },});` the following error is printed to the browser console after login. This is happening in both my real application and a test app I just created.

```
Auth.ts:1389 Uncaught (in promise) TypeError: e.getSession is not a function
    at Auth.ts:1389
    at new Promise (<anonymous>)
    at e.userSession (Auth.ts:1387)
    at Auth.ts:1119
    at new Promise (<anonymous>)
    at e.userAttributes (Auth.ts:1118)
    at e.verifiedContact (Auth.ts:1133)
    at auth-helpers-442d4bea.js:56
    at auth-helpers-442d4bea.js:32
    at Object.next (auth-helpers-442d4bea.js:13)
```

**To Reproduce**
1. Create new react typescript app using cra
2. `amplify init`
3. Setup hosting and authentication using a social provider (facebook in my case)
4. Wrap app in HOC `withAuthenticator`
5. `amplify publish`
6. View app and signin with valid user account
7. Look at chrome console

**Expected behavior**
User should be logged in without errors

**What is Configured?**
* Followed steps on https://docs.amplify.aws/lib/auth/social/q/platform/js
* Configured hosting and auth

```
/* eslint-disable */
// WARNING: DO NOT EDIT. This file is automatically generated by AWS Amplify. It will be overwritten.

const awsmobile = {
    "aws_project_region": "us-east-1",
    "aws_cognito_identity_pool_id": "us-east-1:*****",
    "aws_cognito_region": "us-east-1",
    "aws_user_pools_id": "us-east-1_*****",
    "aws_user_pools_web_client_id": "*****",
    "oauth": {
        "domain": "amplifytest*****-dev.auth.us-east-1.amazoncognito.com",
        "scope": [
            "phone",
            "email",
            "openid",
            "profile",
            "aws.cognito.signin.user.admin"
        ],
        "redirectSignIn": "https://dev.*****.amplifyapp.com/",
        "redirectSignOut": "https://dev.*****.amplifyapp.com/",
        "responseType": "code"
    },
    "federationTarget": "COGNITO_USER_POOLS"
};


export default awsmobile;

   ```

<details>
  <summary><strong>Environment</strong></summary>

```
npx: installed 1 in 1.485s

  System:
[...]

PR #7562, opened 2021-01-15
fix(@aws-amplify/ui-components): handle federated user in checkContact

amhinson:
_Issue #, if available:_
#7554 

_Description of changes:_
When using the Amplify UI Components, `Auth.verifiedContact(...)` is being called after signing in to check whether the user should be routed to the app or the "Verify Contact" screen. However, this causes a problem when using Identity Pool federation because the `user` is [not actually a `CognitoUser`](https://docs.amplify.aws/lib/auth/advanced/q/platform/js#identity-pool-federation) and `verifiedContact` calls `user.getSession` (via [`userSession`](https://github.com/aws-amplify/amplify-js/blob/843183b094281b42ec829b010dd41d214899c23d/packages/auth/src/Auth.ts#L1389)) which doesn't exist.

This PR checks to make sure `user` is an instance of `CognitoUser` before calling `Auth.verifiedContact` and changes the auth state to "signed in" if it is not an instance.

I tested with a federated user as well as regular user and validated the behavior.

---

### DIFF
--- a/packages/amplify-ui-components/src/common/auth-helpers.ts
+++ b/packages/amplify-ui-components/src/common/auth-helpers.ts
@@ -1,4 +1,4 @@
-import { Auth } from '@aws-amplify/auth';
+import { Auth, CognitoUser } from '@aws-amplify/auth';
 import { Logger, isEmpty } from '@aws-amplify/core';
 import { AuthState, ChallengeName, CognitoUserInterface, AuthStateHandler } from './types/auth-types';
 import { dispatchToastHubEvent } from './helpers';
@@ -11,6 +11,14 @@ export async function checkContact(user: CognitoUserInterface, handleAuthStateCh
   if (!Auth || typeof Auth.verifiedContact !== 'function') {
     throw new Error(NO_AUTH_MODULE_FOUND);
   }
+
+  // If `user` is a federated user, we shouldn't call `verifiedContact`
+  // since `user` isn't `CognitoUser`
+  if (!(user instanceof CognitoUser)) {
+    handleAuthStateChange(AuthState.SignedIn, user);
+    return;
+  }
+
   try {
     const data = await Auth.verifiedContact(user);
     if (!isEmpty(data.verified) || isEmpty(data.unverified)) {
--- a/packages/amplify-ui-components/src/common/auth-helpers.ts
+++ b/packages/amplify-ui-components/src/common/auth-helpers.ts
@@ -14,7 +14,7 @@ export async function checkContact(user: CognitoUserInterface, handleAuthStateCh
 
   // If `user` is a federated user, we shouldn't call `verifiedContact`
   // since `user` isn't `CognitoUser`
-  if (!(user instanceof CognitoUser)) {
+  if (!isCognitoUser(user)) {
     handleAuthStateChange(AuthState.SignedIn, user);
     return;
   }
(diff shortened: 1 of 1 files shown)


packages/amplify-ui-components/src/common/auth-helpers.ts
```diff
@@ -11,6 +11,14 @@ export async function checkContact(user: CognitoUserInterface, handleAuthStateCh
...
     throw new Error(NO_AUTH_MODULE_FOUND);
   }
+
+  // If `user` is a federated user, we shouldn't call `verifiedContact`
+  // since `user` isn't `CognitoUser`
+  if (!(user instanceof CognitoUser)) {
+    handleAuthStateChange(AuthState.SignedIn, user);
+    return;
+  }
+
```
elorzafe, line 21: It seems there has to be some attribute inside`CognitoUserInterface` that must validate the kind of the user is not federated. How is that user hydrated to the library?

amhinson: Aside from checking if an arbitrary attribute like `getSession` exists, there doesn't appear to be anything specific that indicates that `user` is not a regular `CognitoUser`. It is is being hydrated to the library here:

https://github.com/aws-amplify/amplify-js/blob/843183b094281b42ec829b010dd41d214899c23d/packages/amplify-ui-components/src/common/helpers.ts#L67-L68

elorzafe (approved):
Thanks @amhinson 🌮 

we have to think over a way to handle state better between Auth category and components

packages/amplify-ui-components/src/common/auth-helpers.ts
wlee221: Yeah, I couldn't find a simpler way to differentiate `CognitoUser`. Checking `user.getSession` would suffice too, but what you have is more explicit 👍

packages/amplify-ui-components/src/common/auth-helpers.ts
```diff
@@ -11,6 +11,14 @@ export async function checkContact(user: CognitoUserInterface, handleAuthStateCh
   if (!Auth || typeof Auth.verifiedContact !== 'function') {
     throw new Error(NO_AUTH_MODULE_FOUND);
   }
+
+  // If `user` is a federated user, we shouldn't call `verifiedContact`
+  // since `user` isn't `CognitoUser`
+  if (!(user instanceof CognitoUser)) {
```
wlee221, line 17: I think we might benefit from separating this out to another helper function:
```ts
export const isCognitoUser = (user: CognitoUserInterface) => {
  return user instanceof CognitoUser;
}
```
so that we can reuse it if similar issues occur with federated users.

wlee221 (approved):
This LGTM 🚀  Just got a minor suggestion.

packages/amplify-ui-components/src/common/auth-helpers.ts
wlee221: Unrelated, but I just found `CognitoUserInterface` is a misnomer because it may not be `CognitoUser` 🙃

packages/amplify-ui-components/src/common/auth-helpers.ts
amhinson: Yea, good call here. Even though it'd be ideal to update the Auth category to better differentiate Cognito users and federated users, it's still a good idea to use a helper function just in case this needs to be used again before that change to Auth is made. I've updated the PR to reflect this 👍

PR #7592, opened 2021-01-20
feat(@aws-amplify/auth): Add confirmSignUp or confirmSignUp_failure events

ericclemmons:
There are two items to be addressed:

- [x] A customer has reported the concern related to the list of auth events available in the documentation through the Amplify Hub is not complete: 

    > https://docs.amplify.aws/lib/auth/auth-events/q/platform/js

    Namely, `signUp_failure` is not listed.

- [x] That same customer could not find a trigger for when the user enters a confirmation code.

    This PR adds `confirmSignUp` and `confirmSignUp_failure` that's consistent with the `signIn` & `signUp` events. For customers using the `<AmplifyAuthenticator>`, they can hook into these events now.

- [x] Docs can be merged afterwards

    > https://github.com/aws-amplify/docs/pull/2872

### Prior Art

- #7111 

---


packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -491,7 +491,21 @@ describe('auth unit test', () => {
...
+			const unsubscribe = Hub.listen('auth', capsule => {
+				switch (capsule.payload.event) {
+					case 'confirmSignUp': {
+						unsubscribe();
+						return expect(capsule.payload.data).toBe('Success');
+					}
+
+					case 'confirmSignUp_failure': {
+						unsubscribe();
+						return expect(capsule.payload.data).toBe('false');
```
sammartinez, line 503: What am I missing here? Should it be as below: 
```suggestion
						return expect(capsule.payload.data).toBe(false);
```

ericclemmons: This (failure) block never fires, but is here for completeness in the event a mock or something breaks around it. (Otherwise we'll only get a "Expected 2 assertions, got 1" error from the test.)

We could either make it similar to the test below:

```suggestion
						return expect(capsule.payload.data).toBe('Should not be called');
```

Or **remove it entirely**.  Honestly, I'm totally fine to only have the needed `expect` in place with `expect.assertions(2)`!

sammartinez: That makes sense to me. I believe we should keep it in case we miss this writing additional test in the future

cwomack:
Closing this PR out due to it no longer being applicable to the most recent major version of Amplify (v6).  Feel free to open a new PR tied to v6 if you'd like and we appreciate your contribution.

PR #7596, opened 2021-01-21
Added Accessibility Labels to UI components

### LINKED ISSUE #7595: Automating UI flow with Appium
Found the same problem as the issue described here: #730
There was an early attempt to fix this with this PR #739 but was never merged.

Would be good to have the `accessibilityLabel` so that test can be automated for Android as well.

mirgj:
Fixes #7595

_Description of changes:_

Adding Accessibility Labels to the React Native components rendered by AmplifyUI.
This will allow to run automation test with Appium on Android


packages/aws-amplify-react-native/src/AmplifyUI.tsx
```diff
@@ -150,7 +151,7 @@ export class PhoneField extends Component<IPhoneProps, IPhoneState> {
 interface ILinkCellProps {
 	disabled?: boolean;
 	onPress: () => void;
-	testID?: string;
+	id?: string;
```
amhinson, line 154: Since this is solely used for testing via the `testID` & `accessibilityLabel`, I'd prefer to leave this as `testID` just so there isn't any question as to what it is for.

packages/aws-amplify-react-native/src/AmplifyUI.tsx
```diff
@@ -178,15 +179,15 @@ export const LinkCell: FC<ILinkCellProps> = props => {
 };
 
 interface IHeaderProps {
-	testID?: string;
+	id?: string;
```
amhinson, line 182: Same comment as above.

packages/aws-amplify-react-native/src/Utils/index.ts
```diff
@@ -0,0 +1,7 @@
+import { Platform } from 'react-native';
+
+export const setTestId = (id: string) => {
+	return Platform.OS === 'android'
+		? { accessible: true, accessibilityLabel: id }
```
amhinson, line 5: Is `accessible: true` needed? I haven't tested this on my own, but I don't see any references in the [documentation](https://reactnative.dev/docs/view#accessibilitylabel) about `accessibilityLabel` requiring `accessible` to be `true`. Another thing to note from the [documentation](https://reactnative.dev/docs/view#accessible) is that all touchable elements have `accessible` enabled by default.

packages/aws-amplify-react-native/src/Utils/index.ts
```diff
@@ -0,0 +1,7 @@
+import { Platform } from 'react-native';
+
+export const setTestId = (id: string) => {
+	return Platform.OS === 'android'
```
amhinson, line 4: From my understanding, I don't think there would be anything wrong with adding both `accessibilityLabel` and `testID` to all platforms. For simplicity, can we just remove this additional logic here and use both properties?

amhinson (changes requested):

packages/aws-amplify-react-native/src/Utils/index.ts
mirgj: `accessible: true` should be redundant I can remove that

packages/aws-amplify-react-native/src/Utils/index.ts
mirgj: As I know there are issues on putting both properties together: https://stackoverflow.com/questions/64706072/how-to-set-the-testid-and-the-accessibilitylabel-together-with-react-native

Most of the reference I've read use that syntax to overcome this limitations

amhinson: Ah ok that makes sense. Thanks for the link!

amhinson (approved):
LGTM 👍

PR #7618, opened 2021-01-25
fix: AuthenticationHelper - Handle negative BigIntegers

manueliglesias:
Using two's complement for negative numbers

_Issue #, if available:_

_Description of changes:_
Use two's complement representation


packages/amazon-cognito-identity-js/__tests__/AuthenticationHelper-test.js
```diff
@@ -0,0 +1,555 @@
...
+		[248, 'F8', '00F8'],
+		[249, 'F9', '00F9'],
+		[250, 'FA', '00FA'],
+		[251, 'FB', '00FB'],
+		[252, 'FC', '00FC'],
+		[253, 'FD', '00FD'],
+		[254, 'FE', '00FE'],
+		[255, 'FF', '00FF'],
+		[256, '00', '0100'],
+	])('%s padHex(%s) === %s', (i, _hex, expected) => {
```
crockeea, line 547: What is the purpose of the middle value in the triple?

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -152,8 +162,7 @@ export default class AuthenticationHelper {
 		const combinedString = `${deviceGroupKey}${username}:${this.randomPassword}`;
 		const hashedString = this.hash(combinedString);
 
-		const hexRandom = randomBytes(16).toString('hex');
-		this.SaltToHashDevices = this.padHex(new BigInteger(hexRandom, 16));
+		this.SaltToHashDevices = randomBytes(16).toString('hex');
```
crockeea, line 165: Add a comment here. What is the difference between `toString` and `padHex`, and why is `toString` the right one? Does `toString` drop leading 0-bytes?

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -337,17 +346,52 @@ export default class AuthenticationHelper {
 	}
 
 	/**
-	 * Converts a BigInteger (or hex string) to hex format padded with zeroes for hashing
-	 * @param {BigInteger|String} bigInt Number or string to pad.
+	 * Converts a BigInteger to hex format, left padded with zeroes to an even number of digits for hashing
```
crockeea, line 349: This would be improved with some examples for the four cases covered by the tests (e.g., -200, -20, 20, and 200). When is a `FF` prepended vs `00` vs neither? What functionality is this function trying to emulate?

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -337,17 +346,52 @@ export default class AuthenticationHelper {
...
+	 * 
+	 * It never returns a negative sign, for negatives it returns hex representation of two's complement
+	 * 
+	 * @param {BigInteger} bigInt Number or string to pad.
 	 * @returns {String} Padded hex string.
```
crockeea, line 354: I don't think "padding" is the right way to look at what this function does. I would describe it in terms of its equivalent Java functionality, or say that it gives an unambiguous, even-length hex string of the two's complement encoding of an integer.

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -22,12 +22,22 @@ import SHA256 from 'crypto-js/sha256';
...
 
+/**
+ * Tests if a string is a hex value with an optional negative sign in the front
+ */
+const HEX_REGEX = /^(-)?([\da-f]+)$/i;
+
+/**
+ * Tests if a hex string has it most significant bit set
+ */
+const HEX_MSB_REGEX = /^[89a-f]/i;
```
crockeea, line 39: Is this a case-insensitive regex? If not, add `A-F`. Similarly with `HEX_REGEX`.

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -337,17 +346,52 @@ export default class AuthenticationHelper {
...
-		if (hashStr.length % 2 === 1) {
-			hashStr = `0${hashStr}`;
-		} else if ('89ABCDEFabcdef'.indexOf(hashStr[0]) !== -1) {
-			hashStr = `00${hashStr}`;
+		if (!(bigInt instanceof BigInteger)) {
+			throw new Error('Not a BigInteger');
 		}
-		return hashStr;
+
+		const [, negativeSign, hexNumber] = bigInt.toString(16).match(HEX_REGEX);
```
crockeea, line 361: What is the un-matched first component of this regex?

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -337,17 +346,52 @@ export default class AuthenticationHelper {
...
+
+		let hex = hexNumber.length % 2 === 1 ? `0${hexNumber}` : hexNumber;
+
+		if (negativeSign) {
```
crockeea, line 365: Nit: perhaps clearer if you compare the `BigInteger` input to 0

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -337,17 +346,52 @@ export default class AuthenticationHelper {
...
+			hex = this.twosComplement(bigInt).toString(16);
+
+			hex = hex.length % 2 !== 0 ? `0${hex}` : hex;
```
crockeea, line 368: Is this a possibility? I think not. It should be documented at `twosComplement`. (If it is a possibility, this is wrong: after taking the two's complement, you should prepend an `F`.

packages/amazon-cognito-identity-js/__tests__/AuthenticationHelper-test.js
manueliglesias: No real purpose beyond pretty printing the test case description in the unit test tool

e.g.

```
 PASS  __tests__/AuthenticationHelper-test.js
  AuthenticatorHelper
    ✓ -256 padHex(00) === FF00 (1ms)
    ✓ -255 padHex(01) === FF01
    ✓ -254 padHex(02) === FF02
    ✓ -253 padHex(03) === FF03
    ✓ -252 padHex(04) === FF04 (1ms)
```

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -337,17 +346,52 @@ export default class AuthenticationHelper {
...
+		} else {
+			hex = hex.length % 2 === 1 ? `0${hex}` : hex;
+
+			hex = HEX_MSB_REGEX.test(hex) ? `00${hex}` : hex;
+		}
+
+		return hex;
+	}
+
+	twosComplement(bigInt) {
```
crockeea, line 380: Let's carefully document this.

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
manueliglesias: It is a case-insensitive regex, the `/i` at the end is javascript's flag for it

See [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#advanced_searching_with_flags_2](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#advanced_searching_with_flags_2)

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -63,7 +73,7 @@ export default class AuthenticationHelper {
 		);
```
crockeea, line 73: Update computation of `this.k` to use `padHex` rather than ad-hoc padding.

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
manueliglesias: The first component is the first complete match, in this case it would be the whole value returned by `bigInt.toString(16)` (including the `-` sign)

[Reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#return_value)

packages/amazon-cognito-identity-js/__tests__/AuthenticationHelper-test.js
crockeea: It's just that, e..g, '00' appears three times: for -256, for 0, and 256. And I'm not sure what, e.g., `-256 padHex(00) === FF00` means. How does '00' relate to -256?

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
crockeea: Perhaps a comment, then.

packages/amazon-cognito-identity-js/__tests__/AuthenticationHelper-test.js
manueliglesias: I'll remove it, it seems to introduce unnecessary noise, and as mentioned before, its only purpose was to use it in the test case description.

It was just the `bytesToHex(x)` output from the Java code used to generate the test cases.

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
manueliglesias: This can happen because BigInteger's toString method doesn't return an even-length string, e.g.:

```
> new BigInteger('5', 16).toString(16)
'5'
```

manueliglesias: This can happen because `BigInteger'`s `toString` method doesn't return an even-length string, e.g.:

```
> new BigInteger('5', 16).toString(16)
'5'
```

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -337,17 +349,51 @@ export default class AuthenticationHelper {
...
-		} else if ('89ABCDEFabcdef'.indexOf(hashStr[0]) !== -1) {
-			hashStr = `00${hashStr}`;
+		if (!(bigInt instanceof BigInteger)) {
+			throw new Error('Not a BigInteger');
+		}
+
+		const isNegative = bigInt.compareTo(BigInteger.ZERO) < 0;
+		const [, magnitude] = bigInt.toString(16).match(HEX_REGEX);
+
+		let hex = magnitude.length % 2 === 1 ? `0${magnitude}` : magnitude;
```
crockeea, line 365: Nit: move this line and line 363 into the body of the `else`, since they are not used in the `if`.

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -337,17 +349,51 @@ export default class AuthenticationHelper {
...
+
+	twosComplement(bigInt) {
+		if (!(bigInt instanceof BigInteger)) {
+			throw new Error('Not a BigInteger');
 		}
-		return hashStr;
+
+		const hashStr = bigInt.abs().toString(16);
+
+		hashStr = hashStr.length % 2 !== 0 ? `0${hashStr}` : hashStr;
```
crockeea, line 389: I don't think there's any point in padding here, is there? This function returns a BigInteger, and the `invertedNibbles` is always interpreted as positive (since there is no leading `-` sign).

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -337,17 +349,51 @@ export default class AuthenticationHelper {
...
+		} else {
+			hex = hex.length % 2 === 1 ? `0${hex}` : hex;
+
+			hex = HEX_MSB_REGEX.test(hex) ? `00${hex}` : hex;
+		}
+
+		return hex;
+	}
+
+	twosComplement(bigInt) {
```
crockeea, line 382: Initially, I suggested pulling out `twosComplement` as its own function. However, upon further thought, this isn't cohesive as written. Consider what happens when the input is `BI(-255)`: the output is `BI(1)`. This is really bizarre, since these two numbers are essentially unrelated.

Let's ditch this function and inline it into `padHex`. 

```
padHex(bigInt) {
	if (!(bigInt instanceof BigInteger)) {
		throw new Error('Not a BigInteger');
	}
	
	const isNegative = bigInt.compareTo(BigInteger.ZERO) < 0;

	if (isNegative) {
		const hashStr = bigInt.abs().toString(16);
		/* Flip the bits of the representation */
		let invertedNibbles = hashStr.split('').map(x => {
			const invertedNibble = ~parseInt(x, 16) & 0xf;
			return '0123456789ABCDEF'.charAt(invertedNibble);
		}).join('');

		/* After flipping the bits, add one to get the 2's complement representation */
                flippedBitsBI = new BigInteger(invertedNibbles, 16).add(BigInteger.ONE);
                
		/** Pad on the left with `F` nibbles:
		 *   - If the representation of `flippedBitsBI` is odd-length, prepend one `F`
		 *   - If the representation has even length, but the top bit is a 0, prepend `FF`
		 */
		if (hex.length % 2 !== 0) {
 			hex = `F${hex}`;
		else if(!HEX_MSB_REGEX.test(hex)) {
		 	hex = `FF${hex}`;
		}
	}
	else {
                # ENOTE: I don't think we need to apply the regex here since we know the number is positive.
		const [, magnitude] = bigInt.toString(16).match(HEX_REGEX);

		/** For positive numbers, pad on the left with `0` nibbles:
		 *   - If the representation of `magnitude` is odd-length, prepend one `0`
		 *   - If the representation has even length, but the top bit is a 1, prepend `00`
		 */
		if (hex.length % 2 !== 0) {
 			hex = `0${hex}`;
		else if(HEX_MSB_REGEX.test(hex)) {
		 	hex = `00${hex}`;
		}
	}
	return hex;
```

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
crockeea: Update: see suggested code for `padHex`.

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
manueliglesias: I am expanding the comments on this function to address this

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
manueliglesias: See updated code

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -337,17 +342,80 @@ export default class AuthenticationHelper {
...
+ * Returns an even-length hex string representation of a BigInteger
+ * 
+ * @param {BigInteger} A Big Integer
+ * @returns {string} The even-length hex representation of the Big Integer
+	 */
+function bigIntToEvenLengthHex(bigInt) {
+		if (!(bigInt instanceof BigInteger)) {
+			throw new Error('Not a BigInteger');
+		}
+
```
crockeea, line 417:
```suggestion
/**
 * Returns an even-length hex string representation of a non-negative BigInteger
 * 
 * @param {BigInteger} A Big Integer
 * @returns {string} The even-length hex representation of the Big Integer
 */
function bigIntToEvenLengthHex(bigInt) {
	if (!(bigInt instanceof BigInteger)) {
		throw new Error('Not a BigInteger');
	}
	if (bigInt < 0) {
		throw new Error('Input to bigIntToEvenLengthHex must be non-negative');
	}

```

crockeea: Fixed formatting, improved comment, and added a validation check.

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -337,17 +342,80 @@ export default class AuthenticationHelper {
...
+	const hexStr = bigInt.toString(16);
+
+	return hexStr.length % 2 !== 0 ? `0${hexStr}` : hexStr;
+	}
```
crockeea, line 421:
```suggestion
}
```

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -152,8 +158,7 @@ export default class AuthenticationHelper {
 		const combinedString = `${deviceGroupKey}${username}:${this.randomPassword}`;
 		const hashedString = this.hash(combinedString);
 
-		const hexRandom = randomBytes(16).toString('hex');
-		this.SaltToHashDevices = this.padHex(new BigInteger(hexRandom, 16));
+		this.SaltToHashDevices = randomBytes(16).toString('hex');
```
crockeea, line 161:
```suggestion
		// This will be interpreted as a postive 128-bit integer
		this.SaltToHashDevices = randomBytes(16).toString('hex');
```

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -105,9 +110,10 @@ export default class AuthenticationHelper {
 		const hexRandom = randomBytes(128).toString('hex');
```
crockeea, line 110:
```suggestion
		// This will be interpreted as a postive 128-bit integer
		const hexRandom = randomBytes(128).toString('hex');
```

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
manueliglesias: This function is no longer needed / included

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
manueliglesias: Formatting has been fixed on the latest commit

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
manueliglesias: This function was discussed offline, we agreed on an implementation

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -337,17 +344,72 @@ export default class AuthenticationHelper {
...
+		For hex strings starting with 'FF8', 'FF' can be dropped, e.g. 0xFFFF80=0xFF80=0x80=-128
+
+		Any sequence of '1' bits on the left can always be substituted with a single '1' bit
+		without changing the represented value.
+
+		This only happens in the case when the input is 80...00
+		*/
+		if(hexStr.toUpperCase().startsWith('FF8')) {
+			hexStr = hexStr.substring(2);
 		}
```
crockeea, line 411:
```suggestion

			/*
			For hex strings starting with 'FF8', 'FF' can be dropped, e.g. 0xFFFF80=0xFF80=0x80=-128

			Any sequence of '1' bits on the left can always be substituted with a single '1' bit
			without changing the represented value.

			This only happens in the case when the input is -0x80...00
			*/
			if(hexStr.toUpperCase().startsWith('FF8')) {
				hexStr = hexStr.substring(2);
			}
		}
```

crockeea (approved):

Jeff17Robbins:
I apologize for jumping in here, but I'm confused about the very idea of a "negative" hex string.  My understanding is that Cognito was erroneously returning a "salt" value as a hex string preceded by a hyphen.  I don't think the right fix is to ingest that salt value.  The existing JavaScript code was already ignoring the hyphen "by luck", but I didn't think that the hyphen prefix was there intentionally, but, rather, it was a bug.

The reason I think the JavaScript code should assert an error if confronted with a broken hex string is that then someone would have noticed the (arguably bad) Cognito bug sooner.  

Again, if I am misinformed about the nature of this PR, I apologize in advance!

manueliglesias:
Hi @Jeff17Robbins 

To give some additional context, the idea of a negative hex string is built into the [BigInteger implementation used in this client](https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/src/BigInteger.js#L220-L222) and Java's [BigInteger class](https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html#toString-int-). Cognito was returning a value compatible with the `BigInteger.toString()` of a negative number, which in and of itself is not invalid, it's just how BigInteger encodes negative numbers. However, this client was not properly parsing this format. This PR updates the client to properly parse this format. Although Cognito doesn't return negative salt values anymore, other places in the client use this format, and we need to ensure compatibility between the JS client and Cognito.

Jeff17Robbins:
Hi @manueliglesias,

I think maybe there are two topics here.  

1. 
I am under the impression that the SRP salt cannot properly be a negative hex string. (Although to be honest, I can't prove my impression is correct!)  But presumably, there's some "fact of the matter" when it comes to that security protocol.  If the salt cannot be a negative hex string, then the Cognito client shouldn't silently ignore an invalid SRP parameter.  

My point here is about the problem of silently accepting erroneous protocol from Cognito.  In this instance, Cognito offered an unfortunate "tell" whereby I could check the validity of the username merely from examining the salt value, without needing to execute the relatively slow SRP math which normally would deter a hacker from a dictionary attack.  Had the receiving code recognized the "negative" hex string as invalid, it could have thrown an error, thereby alerting the community that there was a real problem in Cognito's SRP implementation.

So I think we should determine if the salt value in SRP is allowed to be a "negative" hex string.  If it is, then I agree with you on this.  If it isn't, then I don't think it is correct to parse it with code that accepts an invalid value. 

Perhaps the devs who work on the Cognito SRP implementation, who fixed their end of this bug, could be approached as to what the protocol demands?

2.
As to other uses for negative hex strings: this is new to me (as a somewhat old timer who came up from machine code/assembly language where there really was no such thing. ) Negative numbers when represented in hex simply had the high bit set of the top byte, in those days.  Of course I believe you regarding the support for this feature in the two libraries you cite!  I am curious what the use-case is for that notation.  If you have any pointers for further reading, I'd like to learn more!


> Cognito was returning a value compatible with the BigInteger.toString() of a negative number, which in and of itself is not invalid

crockeea:
> I am under the impression that the SRP salt cannot properly be a negative hex string.
A salt value is just a string of bytes, it doesn't really have a sign, and there is nothing wrong with a negative salt value from the protocol's perspective. It is not incorrect for Cognito to send such a value to the client, as long as it is properly handled in the client.

> I am curious what the use-case is for that notation.
I couldn't say, other than that that is how `BigInteger` chooses to serialize numbers in hex (as opposed to two's complement).

Jeff17Robbins:
@crockeea,

re 

> A salt value is just a string of bytes, it doesn't really have a sign...It is not incorrect for Cognito to send such a value to the client

I wanted to make sure my reasoning is clear, so please be patient with the somewhat lengthy exposition presented here:

The original SRP paper [RFC 2945](http://srp.stanford.edu/rfc2945.txt) uses a non-negative number for its salt.  I base this on the fact that the salt value is the output of the `random()` function in the paper's pseudo-code.  This same `random()` function is used for `b`, and since `g` is raised to the `b`th power, we can see that the `random()` function doesn't return a negative number. 

Since the encoding of integers into hex strings is specified also, there's no way that SRP's salt could correctly be a hex string preceded by a hyphen, or a "negative hex string", when represented in the protocol.  

Therefore, it **is** incorrect for Cognito to send such a value, in this instance.

I understand that the `BigInteger` software construct supports such "negative hex strings", but, again, since the SRP salt cannot properly ever be such a thing, a client SRP algorithm shouldn't silently accept it.

Again, I think it would be better to check the above reasoning with the Cognito team.  After all, they recently fixed a bug in Cognito's SRP code, that produced the negative sign in front of the salt hex string in the case of an unknown user name.  The erroneous behavior of this bug should not be taken as evidence that the salt can be "negative".  Cognito no longer returns erroneous negative salt hex strings.  If it does, it will be due to another bug, and the JavaScript client should sound the alarm.

Given the sensitive nature of cybersecurity, I think silently accepting invalid security protocol parameters (if my reasoning above is correct) is of high concern, and shouldn't be influenced by the behavior of the `BigInteger` library.  Based on my reading of the SRP specification, no hex string presented by the server (Cognito) will ever have a negative sign in front of it.

paulie4:
> A salt value is just a string of bytes, it doesn't really have a sign, and there is nothing wrong with a negative salt value from the protocol's perspective.

Since bytes cannot be negative, it is therefore NOT valid for a hexadecimal string that represents those bytes to have a negative symbol.

crockeea (approved):

amhinson (approved):
Look good to me! I'd also suggest testing React Native iOS & Android as well as Expo, since they all have different underlying implementations of `BigInteger`.

Jeff17Robbins:
If Cognito erroneously returns a hyphen in front of its salt hex string (a "negative hex string"), will this code process it as if it were correct, or will it alert us to an error in Cognito?

Issue #1125, opened 2018-06-27
Possibility to specify role ARN for IAM credentials


pjarts:
Feature request.

Currently, is it not possible to specify which role arn to use when creating `CognitoIdentityCredentials`. According to [the documentation](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CognitoIdentityCredentials.html) you can specify a `RoleArn` to use when creating `CognitoIdentityCredentials`.

I have multiple applications with a shared user pool. Each application has created a set of groups in the user pool with IAM roles attached. Each user can be a member of multiple groups, e.g. _User1_ is a member of groups _app1-admins_ and _app2-admins_. When _User1_ logs into _App1_ he should assume the role attached to the _app1-admins_ group, but the only way to achieve this currently, is to give _app1-admins_ a lower precedence than _app2-admins_ so that the corresponding role's arn gets assigned to the Id token's `cognito:preferred_role` attribute. However, this will cause problems when _User1_ logs into _App2_..

My suggestion is to add another option to the `Auth` configuration that works as a filter for the user's `cognito:groups` (or `cognito:roles`) list:

```js
Amplify.configure({
  Auth: {
    // ... some configuration
    groups: [
      'app1-admins',
      'app1-users'
    ]
  }
})
```

I would prefer to list groups, but I am not sure whether the index of the elements in the `cognito:roles` and `cognito:groups` arrays will match. If they are guaranteed to be listed in the same order you could use the index of the group name to retrieve the corresponding role. To determine which role to use, you would filter the user's `cognito:groups` list for names listed in `Auth._config.groups` and then pick the role connected to the group with the lowest precedence.

victorzw895:
I need this feature too, right now its pointless for my users to belong to multiple groups with multiple roles, when they can only access one and not dynamically either.
I am trying to restrict access to API gateway endpoints based on the roles attached to the groups the user is in.
There should be a way for all role policies to be merged or at least for the role to be assigned dynamically and not by pre-set precedence.

geralds20:
I am also having this issue. Currently, my authenticated users all assume one role. Even though the user has a preferred role matching the group they are in my API calls are being called with this 'base' authenticated role that really doesn't have any permissions.

ahmed-sharief5:
Even i need this feature to be added to aws amplify. Please add an option to add customRoleArn in order to map the dynamic roles to the users based on their cognito groups.

ahmed-sharief5:
[...]

PR #7631, opened 2021-01-27
Add support for the customRoleArn (#1125)

HOLTHOJ:
_Fixes:_ #1125 

_Description of changes:_

The application can specify the role to assume, which will override/bypass the configured logic in the Identity Pool.
https://docs.aws.amazon.com/cognito/latest/developerguide/role-based-access-control.html#using-tokens-to-assign-roles-to-users

#### A customRoleArn property has been added to the Auth configuration object
This property is a function so that applications can first evaluate the ID token before making a role selection (e.g. lookup the available roles in the `cognito:roles` attribute).

Another use case might be that an application contains multiple security zones (e.g. www.enterprise.com/public and www.enterprise.com/private). The role to assume is defined based on the zone the user is in.

Another similar use case might be a splash screen that allows you to login into different accounts. For this the role to assume also needs to be dynamically calculated.

#### All calls to GetCredentialsForIdentity have been updated
If the property is provided, the function will be evaluated when the request for credentials is made.


ahmed-sharief5:
Hi @elorzafe, Could you please review and merge this PR. As our development work is dependent on this feature.

austinwu:
Hello @elorzafe @sammartinez @ashika01 would it be possible to consider this PR?

elorzafe:
Apologies for not looking into this issue until now. I will be looking the security implications of this.

HOLTHOJ:
Hi @elorzafe ,

It's been a while since I looked into this, 
But this is a feature from the underlying AWS api, so we should not restrict users from using it.

From what I recall there are two scenarios;
 - Server side role selection: https://docs.aws.amazon.com/cognito/latest/developerguide/role-based-access-control.html#using-rules-to-assign-roles-to-users
 - Client side role selection: https://docs.aws.amazon.com/cognito/latest/developerguide/role-based-access-control.html#using-tokens-to-assign-roles-to-users

For the client side I described some scenarios in my opening comment.

On the flip side, there doesn't seem too much demand/support for this. I also think Identity pools are simply not used that much anymore. Especially since you can do third-party signins with the user pool as well.

elorzafe:
@HOLTHOJ 

Thanks for replying so quickly!

I understand that you would like to use features that the underlying AWS SDK has, we could provide an escape hatch to support this feature in the future. But I am not convinced we should make it part as a core feature on the category itself. 

Do you have experience working with other AuthN providers on how they solve this issues?

Thanks again!

KevinToala:
in our application this functionality would be very important, thanks

HuiSF:
Hi all, as discussed in https://github.com/aws-amplify/amplify-js/issues/1125 this feature requires support from Amplify Backend, which is tracked in this issue: https://github.com/aws-amplify/amplify-backend/issues/2046. Since the dependency gap has not been filled, we are closing this PR. Please follow up in the linked Amplify Backend issue if you have any further suggestions.

Issue #6731, opened 2020-09-04
ClientMetadata not passing to Cognito Trigger, with `Auth.signIn({... , validationData: {foo: 'bar'}})`


obonyojimmy:
**Describe the bug**
ClientMetadata not passing to  Cognito Trigger, when using Amplify Custom Ui components and Auth.signIn()
 `Auth.signIn({... , validationData: {foo: 'bar'}})`, even with static preconfigured ClientMetadata, 
 `Auth.configure({  authenticationFlowType: 'CUSTOM_AUTH', clientMetadata: {foo: 'bar'}  })` or 
`Amplify.configure({ Auth: {  authenticationFlowType: 'CUSTOM_AUTH', clientMetadata: {foo: 'bar'} }  })` 

**To Reproduce**

*browser js code*

```
Auth.configure({ clientMetadata: {foo: 'bar'}  })
await Auth.signIn({ username: 'someuser', validationData: {xyz: 'abc'} })
```
*lambda trigger*

```
exports.handler = (event, ctx, callback) => {
  console.log(event)
}
```

**Expected behavior**

Expect to see log of `foo` , `xyz` ClientMetadata in lambda cloudwatch logs , this are missing.
 
**Additional context**
 Amplify versions

- "aws-amplify": "^3.1.1",
- "aws-amplify-react": "^4.2.2",

**_You can turn on the debug mode to provide more info for us by setting window.LOG_LEVEL = 'DEBUG'; in your app._**

*browser_debug_log*

![Screenshot_20200905_014143 (1)](https://user-images.githubusercontent.com/14318239/92289484-4e979800-ef19-11ea-902d-cf48d0e2d105.png)

*cloudwatch log**

```
2020-09-04T22:19:05.727Z	****	INFO	{
  version: '1',
  region: 'us-west-2',
  userPoolId: '****',
  userName: '****',
  callerContext: {
    awsSdkVersion: 'aws-sdk-unknown-unknown',
    clientId: '****'
  },
  triggerSource: 'DefineAuthChallenge_Authentication',
  request: {
    userAttributes: {
      sub: '****',
      email_verified: 'true',
      'cognito:user_status': 'CONFIRMED',
      'cognito:email_alias': '****',
      given_name: '****',
      family_name: '****',
      email: '****'
    },
    session: [],
    userNotFound: false
  },
  response: { challengeName: null, issueTokens: null, failAuthentication: null }
}

```

obonyojimmy:
[...]

PR #7633, opened 2021-01-28
fix(@aws-amplify/auth, amazon-cognito-identity-js): Include clientMetadata for token refresh

elorzafe:
Include client metadata from `Auth.configure` to be send also for token refresh

### Use case that is fixing
```jsx
Auth.configure(awsconfig);
Auth.configure({
  clientMetadata: {
    test: 'this should be on refresh token as well'
  }
})

function App() {
  async function signIn() {
    const user = await Auth.signIn({
      username: 'xxxxx',
      password: 'yyyy',
    })
  }

  async function currentUser() {
    const user = await Auth.currentAuthenticatedUser({ bypassCache: true });
  }
  return (
    <div className="App">
        <button onClick={signIn}>sign in</button>
        <button onClick={currentUser}>refresh token</button>
      </header>
    </div>
  );
}
```

When refreshing the token when doing `Auth.currentAuthenticatedUser` Amplify is not sending `ClientMetadata` param to Cognito service.

_Issue #, if available:_
#6731 
_Description of changes:_
`CognitoUser` in `amazon-cognito-identity-js` has been added a `clientMetadata` param for `getSession` and `getUserData` functions in order to be used for `refreshSession`.

`Auth` category on `@aws-amplify/auth` has been added `clientMetadata` from `this._config` to be used in `getSession` and `getUserData` operations.

Also added type safety for `CognitoUserPool` type in `Auth` category

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -90,7 +90,7 @@ const dispatchAuthEvent = (event: string, data: any, message: string) => {
  */
 export class AuthClass {
 	private _config: AuthOptions;
-	private userPool = null;
+	private userPool: CognitoUserPool = null;
 	private user: any = null;
 	private _oAuthHandler: OAuth;
 	private _storage;
@@ -282,8 +282,8 @@ export class AuthClass {
 
 		let username: string = null;
 		let password: string = null;
-		const attributes: object[] = [];
-		let validationData: object[] = null;
+		const attributes: CognitoUserAttribute[] = [];
+		let validationData: CognitoUserAttribute[] = null;
 		let clientMetadata;
 
 		if (params && typeof params === 'string') {
@@ -291,9 +291,19 @@ export class AuthClass {
 			password = restOfAttrs ? restOfAttrs[0] : null;
 			const email: string = restOfAttrs ? restOfAttrs[1] : null;
 			const phone_number: string = restOfAttrs ? restOfAttrs[2] : null;
-			if (email) attributes.push({ Name: 'email', Value: email });
+
+			if (email)
+				attributes.push(
+					new CognitoUserAttribute({ Name: 'email', Value: email })
+				);
+
 			if (phone_number)
-				attributes.push({ Name: 'phone_number', Value: phone_number });
+				attributes.push(
+					new CognitoUserAttribute({
+						Name: 'phone_number',
+						Value: phone_number,
+					})
+				);
 		} else if (params && typeof params === 'object') {
 			username = params['username'];
 			password = params['password'];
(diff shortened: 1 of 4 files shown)


packages/auth/src/Auth.ts
```diff
@@ -662,6 +682,8 @@ export class AuthClass {
 	): Promise<string> {
 		const that = this;
 		return new Promise((res, rej) => {
+			const clientMetadata = this._config.clientMetadata; // TODO: verify behavior if this is override during signIn
```
amhinson, line 685: Did you mean to leave this TODO in here? Is this something you are able to test currently?

amhinson (approved):
Just one question about the TODOs in here, but other than that, everything looks good to go 👍

packages/auth/src/Auth.ts
elorzafe: The issue there is that we might need to persist `clientMetadata` on the client in order to support `Auth.signIn` parameters

ericclemmons (approved):
We approved this in bug bash today.

Thanks for teaching me about this feature!

My understanding is that we have 2 major use-cases:

1. `Auth.signIn(...)` for passing metadata for a privileged operations within lambdas.
2. `Amplify.configure(...)` for persisting metadata so that refresh calls behind-the-scenes can provide this data to lambdas.

donaldpipowitch:
FYI: This breaks TypeScript usage in a patch (?) version as the 2nd argument is now mandatory.

image

It's not clear however, if you actually _have_ to do something or if an empty object would be fine as well. If that is the case it would be nice DX, if the 2nd argument would be optional and default to an empty object.
If an empty object is not enough, please make the type more explicit (and maybe do this in a major change?).

Thanks!

amhinson:
@donaldpipowitch Thanks for bringing that to our attention! We've made a fix in #7654 to address this, and it is available on the `unstable` tag now. We will likely do another release to `latest` next week 👍

monae:
@amhinson It seems not just a type declaration failure but a bug introducing breaking change of the method interface.
In our code `user.getSession(callback)` call became to produce runtime error of null reference.
The `options` argument must be optional in js code:

```js
getSession(callback, options = {}) {
  ...
```

bluepeter:
This issue is now resulting in following error for us. (Referencing here as the search for this error was sparse.)

> TypeError : Cannot read property 'clientMetadata' of undefined Please contact us if this error is happening frequently for you.

donaldpipowitch:
Yes. Like @monae  said it's not just a breaking change from a TypeScript perspective, it also results in a runtime error, if you don't update the getSession call.

bluepeter:
This is resulting in all our users repeatedly logged out. Very concerned if such a major bug is still live in PROD. (And assumed to be nothing more than lint?) We've had to revert to an older release. 

@donaldpipowitch does the merged PR for this issue resolve the runtime err?

amhinson:
#7664 Should address the issue here. You can test it now with the `unstable` tag, if you need an immediate workaround. We are planning to get a new release out within the next day or so.

elorzafe:
@bluepeter @donaldpipowitch @monae 

Apologies for the inconvenience. 

We have just release the fix for `amazon-cognito-identity-js` version `4.5.10` that solves this issue.

bluepeter:
Thank you! :)

Issue #7364, opened 2020-12-08
double POST to cognito /token endpoint (localhost development server only)


davegravy:
**Describe the bug**
Calls to `auth.<region>.amazoncognito.com/oauth2/token` intermittently fail with 400
Underlying cause appears to be double POST request

**To Reproduce**

1. In browser, click "Sign in with AWS"
2. Click the button for the identity provider
3. Provide username & password if prompted, and click Sign In.
4. Repeat 1-3 several times, observe multiple calls in network monitor and occasionally sign-in will work, other times it will fail with 400. 

**Expected behavior**
Consistently successful sign-in experience

**Code Snippet**
Please provide a code snippet or a link to sample code of the issue you are experiencing to help us reproduce the issue. (Be sure to remove any sensitive data)

```import logo from './logo.svg';
import './App.css';
import '@aws-amplify/ui/dist/style.css';
import Amplify from "aws-amplify";
import awsExports from "./aws-exports";
import { withAuthenticator, AmplifySignOut } from '@aws-amplify/ui-react'

Amplify.configure(awsExports);

function App() {
  return (
    <div className="App">
      <header className="App-header">
        <img src={logo} className="App-logo" alt="logo" />
        <p>
          Edit <code>src/App.js</code> and save to reload.
        </p>
        <a
          className="App-link"
          href="https://reactjs.org"
          target="_blank"
          rel="noopener noreferrer"
        >
          Learn React
        </a>
      </header>
      <AmplifySignOut />
    </div>
  );
}

export default withAuthenticator(App)
```

**Screenshots**
Frequently both POST requests fail:
![](https://i.stack.imgur.com/RKUfG.png)

Occasionally one of them succeeds:
![](https://i.stack.imgur.com/MCtYO.png)

Browser 'issue' pane reports:
![image](https://user-images.githubusercontent.com/19296125/101555069-dc09de80-3985-11eb-910f-2f5ee6012308.png)
![](https://i.stack.imgur.com/CjsHQ.png)

Typically after a 400 error the user's email_verified status in the pool is reverted to `false` and a successful sign-in is met with a prompt to verify email:
![](https://i.stack.imgur.com/cBEM2m.png)

**What is Configured?**
If applicable, please provide what is configured for Amplify CLI:
* Which steps did you follow via Amplify CLI when configuring your resources.
![image](https://user-images.githubusercontent.com/19296125/101565591-5abd4680-399b-11eb-828a-a2e7511275c1.png)
[...]

PR #7650, opened 2021-01-29
fix(@aws-amplify/ui-components): BREAKING CHANGE – Move @aws-amplify/* dependencies to peerDependencies

ericclemmons:
_Issue #, if available:_ #7364

This ensures customers including UI components in their projects won't include duplicate dependencies in development.

We expect customers to install ui components via:

```shell
yarn install aws-amplify @aws-amplify/ui-components
```

See my research here:
https://github.com/aws-amplify/amplify-js/issues/7364#issuecomment-764934478

---


packages/amplify-ui-components/package.json
```diff
@@ -34,12 +34,10 @@
     "build:watch": "npm run clean && npm run stencil:watch",
     "clean": "rimraf dist .stencil"
   },
+  "peerDependencies": {
+    "aws-amplify": "3.x.x"
+  },
```
wlee221, line 39: Can you confirm that lerna doesn't pin the version back when we do a publish? I know Lerna does that for deps and dev-deps but I'm not sure about peer-dependencies.

ericclemmons: Recently we tested this logic & lerna completely ignores `peerDependencies`.

wlee221: Cool, thanks for clarifying!

wlee221 (approved):

Issue #7585, opened 2021-01-19
Default colours do not meet contrast standards


anthonyhumphreys:
**Describe the bug**
@aws-amplify/ui-react - default colours are inaccessible according to WCAG

**To Reproduce**
Assess colour contrast against WCAG standards

https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=FF9900

**Expected behavior**
The default colour scheme should always be an accessible palette

ericclemmons:
I would like to see us address this by having neutral defaults, rather than "Amplify Brand" colors (that inevitably get changed).

PR #7653, opened 2021-01-29
fix(@aws-amplify/ui-components): Use dark neutrals rather than Amplify brand colors

ericclemmons:
_Issue #, if available:_ #7585

`<AmplifyAuthenticator>` lives in _customer applications_, but we style it with Amplify's brand colors. These aren't accessible by default (contrast ratio), and stand out in many designs vs. "blending in" tot he app around it.

> Screen Shot 2021-01-29 at 3 43 21 PM

- [x] Used neutral colors by default (nearly black)
- [x] Prefer `Inter var` (from Tailwind UI), `Amazon Ember`, as these are installed deliberately by customers or devices.
- [x] Use Tailwind's `font-sans` `font-family`, which is optimized for system defaults across platforms:

    https://tailwindcss.com/docs/font-family

- [x] 🤕  I cut myself on those buttons, so they're rounded at `3px` instead of `0`.

---

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-button/amplify-button.scss
+++ b/packages/amplify-ui-components/src/components/amplify-button/amplify-button.scss
@@ -63,6 +63,7 @@
   border-width: var(--border-width);
   border-color: var(--border-color);
   border-style: var(--border-style);
+  border-radius: 3px;
 
   &:active {
     opacity: 1;
--- a/packages/amplify-ui-components/src/global/theme.ts
+++ b/packages/amplify-ui-components/src/global/theme.ts
@@ -32,10 +32,10 @@ if (browserOrNode().isBrowser) {
       --amplify-text-xxl: 2.5rem;
 
       /* Colors */
-      --amplify-primary-color: #ff9900;
+      --amplify-primary-color: #222;
       --amplify-primary-contrast: var(--amplify-white);
-      --amplify-primary-tint: #ffac31;
-      --amplify-primary-shade: #e88b01;
+      --amplify-primary-tint: #333;
+      --amplify-primary-shade: #111;
 
       --amplify-secondary-color: #152939;
       --amplify-secondary-contrast: var(--amplify-white);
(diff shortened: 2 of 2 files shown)


packages/amplify-ui-components/src/global/theme.ts
```diff
@@ -32,10 +32,10 @@ if (browserOrNode().isBrowser) {
       --amplify-text-xxl: 2.5rem;
 
       /* Colors */
-      --amplify-primary-color: #ff9900;
+      --amplify-primary-color: #222;
```
sammartinez, line 35: Let's chat on this one, I am not sure what our PMs will think on this

ericclemmons: @renebrandel Can we get your feedback on this PR?

packages/amplify-ui-components/src/components/amplify-button/amplify-button.scss
```diff
@@ -63,6 +63,7 @@
   border-width: var(--border-width);
   border-color: var(--border-color);
   border-style: var(--border-style);
+  border-radius: 3px;
```
wlee221, line 66: Somewhat unrelated -- what is your take on `rem` vs `px`? I think it's fine to have `px` here because border radius doesn't need to scale, but we should have some rule for using one versus the other as we go on.

wlee221 (commented):
LGTM, thanks @ericclemmons! I'll let @sammartinez / PMs give the final review on this.

ericclemmons:
@wlee221 raises a good point: **we should show how to customize these colors early in our docs**.

packages/amplify-ui-components/src/components/amplify-button/amplify-button.scss
ericclemmons: I normally use `rem`, but opted for `3px` because the border radius isn't based on anything relational (e.g. fonts, or another container's border), but smoothing out the sharp corner.

(If the browser font-size went up 200%, I still wanted the corner to be `3px`)

But, to be clear, this isn't some well-researched change.  It's my subjective tastes to match the otherwise-rounded nature of the authenticator and add some appearance of polish :D

ericclemmons:
Closing as to not block the v1 release with a cosmetic detail.

PR #7719, opened 2021-02-09
feat(storage): optional exact config with tests

### LINKED ISSUE #7715: Config option for Amplify Storage Exact Path/Key
I would like to be able to upload the exact key of an object into S3 using the client side API without the user UUID prepended on it. In other words, I would like to do S3 operations without the `user_identity_id` prefix. 

A configuration option called **exactKey** or **exact**. This would be a boolean option to tell the **AWSS3Provider** not to add the user UUID and/or region prefix to the key for the S3 object being uploaded, downloaded, updated, deleted or listed. So that the following code 

```js
 Storage.put(fileName, fileContents, {
      exact: true,
      contentType: 'image/jpeg',
      customPrefix: {
        private: 'someDir/'
      }
    })
      .then((res) => {
        alert("all done here!");
      })
      .catch((err) => {
        console.error(err);
      });
```
Would upload an object with **someDir/fileName.jpeg** as a key to S3. Instead of **someDir/us-east-1:c7c464af-bca0-4e00-a2f4-8c99caeb3e8b/fileName.jpeg**.

**Describe alternatives you've considered**
There aren't alternatives through Amplify. The other alternatives I'm aware of is using the S3 SDK and manually hooking up authentication to cognito Auth roles. Then automating it through cloudformation. Amplify does an excellent job with provisioning file level user permissions using cognito. I'd rather use that than a manual configuration 

**Additional context**
I'm actively working on a PR for this. This is the original discussion question and documentation. 
https://github.com/aws-amplify/amplify-js/discussions/7708
https://docs.amplify.aws/lib/storage/configureaccess/q/platform/js

Ochaun:
_Issue #, if available:_
fix #7715 
_Description of changes:_
- Allows user to set a boolean **exact** property so that they can upload S3 objects without the cognito user id added to the key. 
- Unit tests for the above functionality. 

### DIFF
--- a/packages/storage/__tests__/Storage-unit-test.ts
+++ b/packages/storage/__tests__/Storage-unit-test.ts
@@ -228,7 +228,7 @@ describe('Storage', () => {
 		test('vault level is always private', () => {
 			const storage = StorageCategory;
 			expect.assertions(3);
-			storage.vault.configure = jest.fn().mockImplementation(configure => {
+			storage.vault.configure = jest.fn().mockImplementation((configure) => {
 				expect(configure).toEqual({
 					AWSS3: { bucket: 'bucket', level: 'private', region: 'region' },
 				});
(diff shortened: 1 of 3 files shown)


Ochaun:
Not needed. Setting Storage to `public` means that users within the app have access. Not putting files on the open internet.

PR #7728, opened 2021-02-10
feat(@aws-amplify/ui-components): Remove Shadow DOM

ericclemmons:
## Why?

Read this fam:
> https://github.com/aws-amplify/amplify-js/discussions/7486

### What did it look like?

> Screen Shot 2021-02-09 at 4 38 23 PM

### What does it look like with this PR?

Not quite the same: some CSS variables and display values aren't propagating correctly with `scoped: true`.

> Screen Shot 2021-02-10 at 3 43 25 PM

## What next?

I'm considering punting on this due to time constraints & resolving this independent of v1.

Removing the Shadow DOM will allow customers to override styles, yes, but greater customization will be at the component/slot level.

With `scoped: true`, the HTML structure slightly changes, which _could_ be considered a breaking change. Same for CSS classes! (But I don't think that's reasonable because they're not documented APIs)

Without exhaustive visual regression testing (manual or otherwise), changing or removing `shadow: true`  is **risky**.

---

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-amazon-button/amplify-amazon-button.tsx
@@ -11,7 +11,6 @@ const logger = new Logger('amplify-amazon-button');
 
 @Component({
   tag: 'amplify-amazon-button',
-  shadow: true,
 })
 export class AmplifyAmazonButton {
   /** App-specific client ID from Google */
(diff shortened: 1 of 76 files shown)


ericclemmons:
Closing thanks to @wlee221's research:

> Native web component slots require shadow dom to work. Stencil allows the usage of slots without shadow DOM, but they manually modify the dom tree instead of having browser handle the logic.
>
> With Shadow DOM disabled, if a user provides slotted content to a component that doesn’t have a slot to put it in, the slotted content still renders on the page (at the top of the component). (source (https://medium.com/8451/make-your-stenciljs-web-components-faster-by-using-shadow-dom-d010a9f0cdda))
>
> Haven’t had time to test this, but this means that the following pattern will break:
> ```html
>     <amplify-authenticator>
>       Your app content
>     </amplify-authenticator>
> ```

PR #7746, opened 2021-02-14
fix(@aws-amplify/datastore): improve IDB query performance

iartemiev:
This PR fixes a bug where calling `DataStore.query(Post)` _without_ a pagination object would use the cursor-based fetch from IndexedDB, as opposed to `getAll()`, which is optimized specifically for getting all the records. This was having a negative effect on the performance of `DataStore.query` when retrieving all the items for a given model.

The pagination object was getting initialized to `{limit: undefined, page: undefined, sort: undefined}` in the `DataStore.processPagination` private method, therefore `pagination` always had a truthy value in the conditional checks that determine which retrieval method to utilize. This gets fixed [here](https://github.com/aws-amplify/amplify-js/pull/7746/files#diff-10866b5ac7d469b38dd9928d9d7056a3a402904fef591ea8cfe773b4d75d63efR1131-R1134).

Also:
* Refactored `IndexedDBAdapater.query` to make it easier to test
* Added test coverage for IndexedDBAdapter, which ensures the correct retrieval method is used for all of the different combinations of parameters passed to `IndexedDBAdapter.query`

### DIFF
--- a/packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
+++ b/packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
@@ -370,62 +370,77 @@ class IndexedDBAdapter implements Adapter {
 		await this.checkPrivate();
 		const storeName = this.getStorenameForModel(modelConstructor);
 		const namespaceName = this.namespaceResolver(modelConstructor);
-		const sortSpecified = pagination && pagination.sort;
 
-		if (predicate) {
-			const predicates = ModelPredicateCreator.getPredicates(predicate);
-			if (predicates) {
-				const { predicates: predicateObjs, type } = predicates;
-				const idPredicate =
-					predicateObjs.length === 1 &&
-					(predicateObjs.find(
-						p => isPredicateObj(p) && p.field === 'id' && p.operator === 'eq'
-					) as PredicateObject<T>);
+		const hasPredicate = !!predicate;
+		const hasSort = pagination && pagination.sort;
+		const hasPagination = pagination && pagination.limit;
 
-				if (idPredicate) {
-					const { operand: id } = idPredicate;
+		let records: T[];
 
-					const record = <any>await this._get(storeName, id);
-
-					if (record) {
-						const [x] = await this.load(namespaceName, modelConstructor.name, [
-							record,
-						]);
-
-						return [x];
-					}
-					return [];
-				}
-
-				// TODO: Use indices if possible
-				const all = <T[]>await this.db.getAll(storeName);
-
-				const filtered = predicateObjs
-					? all.filter(m => validatePredicate(m, type, predicateObjs))
-					: all;
-
-				return await this.load(
-					namespaceName,
-					modelConstructor.name,
-					this.inMemoryPagination(filtered, pagination)
-				);
-			}
-		}
-
-		if (sortSpecified) {
-			const all = <T[]>await this.db.getAll(storeName);
-			return await this.load(
-				namespaceName,
-				modelConstructor.name,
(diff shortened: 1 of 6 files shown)


amhinson (commented):
Everything looks good to me. The refactor is much easier to follow. Although what do you think about doing the same thing for `AsyncStorageAdapter` for consistency's sake?

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -370,62 +370,77 @@ class IndexedDBAdapter implements Adapter {
...
 
-		if (predicate) {
-			const predicates = ModelPredicateCreator.getPredicates(predicate);
-			if (predicates) {
-				const { predicates: predicateObjs, type } = predicates;
-				const idPredicate =
-					predicateObjs.length === 1 &&
-					(predicateObjs.find(
-						p => isPredicateObj(p) && p.field === 'id' && p.operator === 'eq'
-					) as PredicateObject<T>);
```
manueliglesias, line 383: I think this code (or a similar form of it 🤔) needs to exist to handle the case for "get by id".

This code was looking at the predicates, and if there was one with only one predicate object of `p.field === 'id' && p.operator === 'eq'` it did a `const record = <any>await this._get(storeName, id);`

It would be great to have a test case for that scenario too, btw, thanks for writing tests for this PR 😄

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -370,62 +370,77 @@ class IndexedDBAdapter implements Adapter {
...
+			);
+			records = this.inMemoryPagination(filtered, pagination);
+		} else if (hasSort) {
+			const all = await this.getAll(storeName);
+			records = <T[]>this.inMemoryPagination(all, pagination);
+		} else if (hasPagination) {
+			records = await this.enginePagination(storeName, pagination);
+		} else {
+			records = await this.getAll(storeName);
+		}
```
manueliglesias, line 393: I like how this section clearly shows which method gets called, I think it would be nice if the case for "get by id" lives here. What do you think?

manueliglesias (approved):
It looks great!

I left 2 comments but they are non-blocking, let me know your thoughts

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
iartemiev: The code is still there, it just got moved into it's own method here: https://github.com/aws-amplify/amplify-js/pull/7746/files#diff-75c76069840355a3f346f2f35c443018e4a5716ed128191602499c3b39935cfcR412-R418

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
iartemiev: "Get by id" is handled in the `hasPredicate` path. Since there's a lot of overlap between "get by id" and "get by (other) predicate", I think that breaking them up into their own methods would result in unnecessary code duplication.
I've added a unit test for "get by id" though.

iartemiev:
> Everything looks good to me. The refactor is much easier to follow. Although what do you think about doing the same thing for `AsyncStorageAdapter` for consistency's sake?

@amhinson Agreed! I've refactored `AsyncStorageAdapater.query` to follow the same pattern and have added unit tests

iartemiev:
@amhinson @manueliglesias - thanks for the thorough reviews and good suggestions, guys. I've updated the PR

amhinson (approved):
LGTM 👍

PR #7790, opened 2021-02-21
chore: consolidate .prettierrc & .prettierignore

iartemiev:
Delete .prettier config files from subdirectories. Consolidate rules in the root config files


sammartinez (approved):
LGTM 🌮

packages/amplify-ui-components/.prettierrc.js
```diff
@@ -1,11 +0,0 @@
-module.exports = {
-  printWidth: 120,
```
wlee221, line 2: Too late of a comment, but do we want to consider moving this to the root config? And can we follow this up with local prettier runs?

PR #7802, opened 2021-02-22
fix(@aws-amplify/amplify-ui-components): phone_number input issue in require-new-password

### LINKED ISSUE #7175: Changing temporary password fails due to phone number issue @aws-amplify/ui-angular
When using    "@aws-amplify/ui-angular": "^0.4.10" and logging in with a temporary password the user is prompted to enter a phone number. Whilst the UI allows you to select the appropriate country code +1 for US, +44 for UK. If you type the remainder of the phone number in to the text box for example,

+44 7777777777

It reports a failure in the phone number. Looking at what is actually sent in the post the +44 is missing:

userAttributes.phone_number | "7777777777"

The user selected portion of the phone number is not being added to the post. If you add the +44 to the text field it works.

Additionally, the text entry box implies that phone numbers should be entered as (555) 555-1234, however, the formatting rules expect there to be only numbers.

renegus:
_Issue #, if available:_ Fixes #7175 

_Description of changes:_

This issue occurs when creating a new cognito user with temporary password and without a ```phoneNumber```. If ```phoneNumber``` is an required field, the ```amplify-require-new-password``` component will contain a ```dialCode``` dropdown and a ```phoneNumber``` input. Both the dropdown and input field overrides the required ```phone_number```field, and only way to proceed is to enter the entire ```dialCode + phoneNumber``` in the input field.

This issue was fixed using same helper functions ```handlePhoneNumberChange``` and ```composePhoneNumberInput``` as used in components (```sign-up```, ```sign-in```, ```forgot-password```, ```confirm-sign-up```)


packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
```diff
@@ -58,9 +67,20 @@ export class AmplifyRequireNewPassword {
...
+    phoneNumberValue: null,
+  };
 
   private handleRequiredAttributeInputChange(attribute, event) {
-    this.requiredAttributes[attribute] = event.target.value;
+    if (attribute === 'phone_number') {
+      handlePhoneNumberChange(event, this.phoneNumber);
+      this.requiredAttributes[attribute] = composePhoneNumberInput(
+        this.phoneNumber
+      );
```
wlee221, line 80: Can we move this to when `completeNewPassword` is called below so that this is not called on every input change?

wlee221 (changes requested):
@renegus, thanks a ton for catching this! Just got a nit but this LGTM otherwise.

packages/amplify-ui-components/src/components/amplify-require-new-password/amplify-require-new-password.tsx
wlee221: And can we add try catch around this, calling `dispatchToastHubEvent(error)` on error?

wlee221:
Hi @renegus, just checking if you're able to work on this PR. I'll put this on auto-close after 2 weeks, but we can reopen or take this over as needed.

eddiekeller:
Closing this PR as the changes have been merged in here - https://github.com/aws-amplify/amplify-js/pull/8489

PR #7803, opened 2021-02-22
Missing return in AmazonAIConvertPredictionsProvider.ts

SebSchwartz:
Without this return, user will also get 'Source types other than byte source are not supported.' error.

_Issue #, if available:_

_Description of changes:_
Dont reject with another error as well.


wlee221 (approved):
This LGTM, thanks for the catch.

Issue #7824, opened 2021-02-26
Authenticated graphql requests with SSR missing custom attributes


asp3:
**Describe the bug**
We have recently switched to using SSR with Next.js, and we have been trying to modify some of our api calls to be able to be used with server side rendering. However, we ran into an issue because in our backend, we use the identity.claims in an appsync call to determine the identity of a user and get some of their common attributes, but when using `API.graphql` from `withSSRContext`, it looks like those fields were omitted. We believe this is because the type of authentication is `token_use: access` instead of `token_use: id`, but looking through the docs there does not seem to be a good way to change our server side calls to use the ID Token.

**Expected behavior**
As you can see below, when an auth is made by initializing a graphql client on client side , there is extra information in `identity.claims` such as Name, referral, classInfo, etc. However, when the same API call is made using `API.graphql()` using `const { API } = withSSRContext(context)`, the `identity.claims` has the generic values. When calling `Auth.currentAuthenticatedUser` in both SSR and on client side, the payload returns correctly, and all of the added values are visible in the payload (`user.signInUserSession.idToken.payload`). 

**Code Snippet**

API call made from client side
```
"identity": {
    "claims": {
      "sub": "...",
      "access_pool": "0",
      "cognito:groups": [
        "Teachers"
      ],
      "email_verified": true,
      "iss": "...",
      "cognito:username": "...",
      "aud": "...",
      "referral": "...",
      "classinfo": "...",
      "token_use": "id",
      "auth_time": 1595474525,
      "name": "...",
      "ID": "...",
      "exp": 1595478125,
      "iat": 1595474525,
      "access_pool_expiry": "0",
      "email": "..."
    },
    "defaultAuthStrategy": "ALLOW",
    "groups": [
      "Teachers"
    ],
    "issuer": "...",
    "sourceIp": [
      "..."
    ],
    "sub": "...",
    "username": "..."
  },
```

API Call made from SSR
```
  "identity": {
    "claims": {
      "sub": "...",
      "device_key": "...",
      "cognito:groups": [
        "Teachers"
      ],
      "token_use": "access",
      "scope": "aws.cognito.signin.user.admin",
      "auth_time": 1614323009,
      "iss": "...",
      "exp": 1614326991,
      "iat": 1614323391,
      "jti": "...",
      "client_id": "...",
      "username": "..."
    },
    "defaultAuthStrategy": "ALLOW",
    "groups": [
      "Teachers"
    ],
    "issuer": "...",
[...]

PR #7827, opened 2021-02-26
add optional `Authorization` parameter to graphql call

asp3:
#7824 

To allow for using id token authorization, add an optional parameter `Authorization` to the graphql function to allow for users to add custom auth types for authenticated requests.

### DIFF
--- a/packages/api-graphql/src/GraphQLAPI.ts
+++ b/packages/api-graphql/src/GraphQLAPI.ts
@@ -160,7 +160,7 @@ export class GraphQLAPIClass {
 			case 'AMAZON_COGNITO_USER_POOLS':
 				const session = await this.Auth.currentSession();
 				headers = {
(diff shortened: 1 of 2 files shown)


svidgen:
@asp3, thanks for taking the time to submit a PR!

Unfortunately, I'm not clear on the intention here. What specific use case are you trying to support?

If you're trying to execute queries that require different types of authorization, [you can provide `authMode` as part of your operation](https://docs.amplify.aws/lib/graphqlapi/authz/q/platform/js) and Amplify will normally do the right thing. For example:

```
let results = await API.graphql({
  query,
  variables,
  authMode: 'API_KEY'
});
```

If that doesn't do the trick, I believe you can also override the `Authorization` (or any other header) with the second method argument:

```
API.graphql({query, variables}, {
  Authorization: "YOUR CUSTOM AUTH HEADER"
});
```

Do either of these options work for you?

Thanks!

Issue #7808, opened 2021-02-23
Predictions Speech to text doesn't work in fr-FR, en-AU, en-GB and fr-CA


SebSchwartz:
**Describe the bug**
The speech to text cannot work in fr-FR, en-AU, en-GB and fr-CA as the sample rate for the streaming transcription is limited to 8 kHz and amplify always send 16 kHz in the request.
As explained in the doc -> https://docs.aws.amazon.com/transcribe/latest/dg/streaming.html

**To Reproduce**

1. Take the example from -> https://docs.amplify.aws/lib/predictions/sample/q/platform/js#sample-react-app
2. Specify the language in one of the listed one.
3. You will receive 2 errors 
   1. One that should be fixed in https://github.com/aws-amplify/amplify-js/pull/7803 -> missing return in code
   2. The real error -> **The requested language doesn't support the specified sample rate. Use the correct sample rate then try again.**
6. You could try to resample to 8 kHz as requested in the aws doc with _wavefile_ library for example (code is working for en-US)
```javascript
let wav = new WaveFile();
wav.fromBuffer(new Uint8Array(fileBlob));
wav.toSampleRate(8000);
```
7. But the request made to Streaming Transcription endpoint is hardcoded to 16 kHz as you can see in the code (L402):
https://github.com/aws-amplify/amplify-js/blob/ac0ef4b994a863fac3b5899fe7190c20a070fa16/packages/predictions/src/Providers/AmazonAIConvertPredictionsProvider.ts#L397-L402

**Expected behavior**
Amplify should detect the sample rate or allow user to pass it as param (more easy to do).

**Code Snippet**
```javascript
const speechToTextOutput = await Predictions.convert({
                      transcription: {
                        source: {
                          bytes: rawPcm,
                        },
                        language: 'fr-FR',
                      },
                    });
```

wlee221:
Hi @SebSchwartz, thanks for your research! You are right, and we should downsample the buffer to 8khz depending on the language code. We already have code for downsampling that should make this fix easier: https://github.com/aws-amplify/amplify-js/blob/ac0ef4b994a863fac3b5899fe7190c20a070fa16/packages/predictions/src/Providers/AmazonAIConvertPredictionsProvider.ts#L332

[...]

PR #7835, opened 2021-02-28
Downsizing sample for some language in speech to text

SebSchwartz:
_Issue #, if available:_ https://github.com/aws-amplify/amplify-js/issues/7808

_Description of changes:_

If language is one of (fr-FR, en-AU, en-GB and fr-CA), downsize the sample to 8kHz and change url to transcribe.


packages/predictions/src/Providers/AmazonAIConvertPredictionsProvider.ts
```diff
@@ -178,10 +180,17 @@ export class AmazonAIConvertPredictionsProvider extends AbstractConvertPredictio
 					languageCode: language,
 				});
 
+				let sourceBytes = source.bytes;
+				if (LANGUAGES_CODE_IN_8KHZ.includes(languageCode)) {
+					sourceBytes = this.downsampleBuffer({
+						buffer: source.bytes,
+						outputSampleRate: 8000,
+					});
+				}
```
wlee221, line 189: If we downsample buffer here, it will be downsampled again inside `sendDataToTranscribe` > [`sendEncodedDataToTranscribe`](https://github.com/aws-amplify/amplify-js/blob/35737d20f69ab7c0f11fdfe942623262448c75ba/packages/predictions/src/Providers/AmazonAIConvertPredictionsProvider.ts#L298) to 16khz instead of 8khz. Let's make sure `downsampleBuffer` is called only once to the correct sample rate.

wlee221: One idea on top of my head is to pass `targetSampleRate` down to `sendEncodedDataToTranscribe`, but the implementation is up to you!

packages/predictions/__tests__/Providers/AWSAIConvertPredictionsProvider-unit-test.ts
```diff
@@ -217,6 +217,31 @@ describe('Predictions convert provider test', () => {
 				predictionsProvider.convert(validSpeechToTextInput)
 			).rejects.toMatch('region not configured for transcription');
 		});
+		test('Error languageCode not configured ', () => {
```
wlee221, line 220: This is great, thanks!

packages/predictions/__tests__/Providers/AWSAIConvertPredictionsProvider-unit-test.ts
```diff
@@ -247,5 +272,35 @@ describe('Predictions convert provider test', () => {
...
+					},
+				},
+			};
+			predictionsProvider.configure(speechGenOptions);
+			jest.spyOn(Credentials, 'get').mockImplementationOnce(() => {
+				return Promise.resolve(credentials);
+			});
+
+			await predictionsProvider.convert(validSpeechToTextInput);
+			expect(downsampleBufferSpyon).toBeCalled();
```
wlee221, line 302: As per discussion below, can we check if this has been called once, and with target sample rate 8000?

wlee221 (changes requested):
@SebSchwartz, thanks a lot for this PR and even writing a missed unit test! Got one problem with downsampling that we want to address, but the rest LGTM 👍

packages/predictions/src/Providers/AmazonAIConvertPredictionsProvider.ts
SebSchwartz: Yes, you're right, I've adapted the code

packages/predictions/__tests__/Providers/AWSAIConvertPredictionsProvider-unit-test.ts
SebSchwartz: Done.

SebSchwartz:
@wlee221 I've changed according to your comments ;)

wlee221 (approved):
This LGTM, thanks a lot for your contribution! Have you tried this change on your app?

wlee221:
Since the fix is straightforward and unit test has been included, I'll go ahead and merge this. Will do simple manual testing after this hits unstable.

PR #7926, opened 2021-03-12
fix(@aws-amplify/datastore): handle sync query unauthorized

iartemiev:
Fixes a bug where the sync engine breaks when a sync query returns an `unauthorized` error. 
The expected behavior for DataStore is that an unauthorized error on a model should not break syncing for other models.

Currently, when a sync query request returns unauthorized, we get the following unhandled error
```
error on  TypeError: Cannot read property 'items' of null
```
on [this line](https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/src/sync/processors/sync.ts#L181). 

An unauthorized query response typically won't return any `items`, so this PR adds a check to see if `items` are defined before attempting to iterate over them. If not defined, we return an empty list (`items: []`) and `nextToken: null` to signal to the `retrievePage` method that it shouldn't attempt to retrieve additional pages via this query.

### DIFF
--- a/packages/datastore/src/sync/processors/sync.ts
+++ b/packages/datastore/src/sync/processors/sync.ts
@@ -142,14 +142,14 @@ class SyncProcessor {
 						variables,
 					});
 				} catch (error) {
-					if (this.partialDataFeatureFlagEnabled()) {
-						const hasItems = Boolean(
-							error &&
-								error.data &&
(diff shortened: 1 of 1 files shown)


amhinson (approved):
LGTM 👍

manueliglesias (approved):
LGTM

Issue #7840, opened 2021-03-01
Update mutations on connection is not working with latest version of amplify


TheMoums:
**Describe the bug**
Hello, I have updated all my aws-amplify packages to their latest versions as well as the amplify cli, and it seems that one particular case of update mutations is not working anymore.

A (very) simplified version of my schema.graphql 

```graphql
type Building @model {
  id: ID!
  units: [Unit!] @connection(keyName: "byBuilding", fields: ["id"])
}
```
```graphql
type Unit @model @key(name: "byBuilding", fields: ["buildingId"]) {
  id: ID!
  buildingId: ID!
  building: Building @connection(fields: ["buildingId"])
  }
```
In previous versions, whenever I would update my Unit to put it on another Building, I would do something like:
```js
await DataStore.save(
      Unit.copyOf(fromStore, (updated) => {
        updated.building = ...
      })
    );
```
Which would result in a mutation like 
![old_screen](https://user-images.githubusercontent.com/22890903/109502838-ccd49180-7a99-11eb-9c21-cc0e9bcebab4.PNG)

Basically, the previous version would send the whole record in the update mutation (Here I was updating the 'Unit4' to put it on the 'Building3').

Unfortunately, it seems that, with the latest versions of aws-amplify, some kind of optimisation is performed to send only the fields that have changed. Which is working when I update simple fields on a given record but failing when updating connections between models.

Here is the same example as before **but** with the latest version

![new_screen_not_working](https://user-images.githubusercontent.com/22890903/109503320-68fe9880-7a9a-11eb-9fcb-2488e790bb52.PNG)

As you can see, the field buildingId is ignored which results in the impossibility to update the Unit in order to put it on another Building.
 ```json
"@aws-amplify/api": "3.2.24",
 "@aws-amplify/auth": "3.4.24",
 "@aws-amplify/cache": "3.1.49",
 "@aws-amplify/core": "3.8.16",
 "@aws-amplify/datastore": "2.9.10",
 "@aws-amplify/predictions": "3.2.24",
 "@aws-amplify/storage": "3.3.24",
```
amplify --v : 4.44.1

iartemiev:
Hey @TheMoums, you're correct in that we recently modified the behavior for update mutations to where only changed fields are sent in the mutation input. This change was made in order to comply with the Automerge conflict resolution strategy, which requires this to function correctly. It looks like we missed some test cases around relations. I'll mark this as a bug and will work towards resolving it ASAP. 

[...]

PR #7935, opened 2021-03-14
fix(@aws-amplify/datastore): update mutation input - diff with DB instead of patches

iartemiev:
_Issue #, if available:_
https://github.com/aws-amplify/amplify-js/issues/7840

This PR changes the way we derive the correct update mutation input when using DataStore. Relying on Immer patches makes it difficult to resolve field data for connections and to include the correct fields when updating relational data. E.g., in the linked issue. Using a diff between the updated record and the previous record is a simpler and more reliable approach.

### DIFF
--- a/packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
+++ b/packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
@@ -206,7 +206,7 @@ class IndexedDBAdapter implements Adapter {
 	async save<T extends PersistentModel>(
 		model: T,
 		condition?: ModelPredicate<T>
-	): Promise<[T, OpType.INSERT | OpType.UPDATE][]> {
+	): Promise<[T, OpType.INSERT | OpType.UPDATE, T?][]> {
 		await this.checkPrivate();
 		const modelConstructor = Object.getPrototypeOf(model)
 			.constructor as PersistentModelConstructor<T>;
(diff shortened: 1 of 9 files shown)


amhinson (approved):
LGTM 👍

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -250,31 +250,25 @@ class IndexedDBAdapter implements Adapter {
...
+			const fromDB = <T>await this._get(store, id);
 			const opType: OpType =
-				(await this._get(store, id)) === undefined
-					? OpType.INSERT
-					: OpType.UPDATE;
+				fromDB === undefined ? OpType.INSERT : OpType.UPDATE;
 
-			// It is me
-			if (id === model.id) {
+			// Even if the parent is an INSERT, the child might not be, so we need to get its key
```
richardmcclellan, line 264: Could you explain this a little more for me?  Is this a cascading save method?  Just trying to figure out why you are talking about parent/child here.

richardmcclellan (approved):
Overall, this looks good!  Very similar to the Android implementation.  I just had one minor question for you (above), and as mentioned in Slack, please confirm that when update mutations are created while offline, they are properly sent to AppSync when they come back online!  As long as that works properly, ship it!

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
iartemiev: Yes, exactly this is cascading save logic. 

FWIW I just slightly refactored it to return the locally stored record (i.e., the previous state that we then diff against) in the [StorageAdapter].save method. That comment and most of the business logic were already there, this PR is not changing the cascading save logic itself.

richardmcclellan: Interesting.  We don't have cascading save on Android (and I don't think on iOS either).  We should probably add that to our roadmap.

Issue #7499, opened 2021-01-05
Cannot use amplify-js in browser environment


sem4phor:
**Describe the bug**

The package aws-amplify is meant to be used in the browser, isnt it? Therefore it should not rely on node builtins, which sadly is the case and prevents the usage of this package in many js build tools like rollup, snowpack or vitejs.

The @aws-cli package which is used by aws-amplify relies on various node build in modules like fs, process or http2, ....

I described the issue also here (https://github.com/vitejs/vite/issues/1374) and Evan You the creator of vuejs and vite mentioned that webpack will drop support for node build ins in version five. Because webpack is used by tools such as create-react-app  and vue-cli this will block many developers from using the amplify package soon I think!

**To Reproduce**
1. Download https://github.com/sem4phor/aws-cli-bug
2. Run npm run dev (node required)
3. See error in console

**Expected behavior**
The package works in the browser and does not import any node build ins which are not available in browser environment.

**Code Snippet**
https://github.com/sem4phor/aws-cli-bug

**Screenshots**
```
[plugin:vite:imports] Failed to resolve import "process". Does the file exist?
/Users/XXXXX/node_modules/@aws-sdk/util-user-agent-node/dist/es/index.js:1:22
1  |  import process from "process";
   |                       ^
2  |  export function defaultUserAgent(packageName, packageVersion) {
3  |      var engine = process.platform + "/" + process.version;
Error: Failed to resolve import "process". Does the file exist?
    at TransformContext.error (C:\Users\XXXXX\node_modules\vite\dist\node\chunks\dep-262f13eb.js:59552:50)
    at TransformContext.transform (C:\XXXXX\node_modules\vite\dist\node\chunks\dep-262f13eb.js:52957:30)
    at async Object.transform (C:\Users\XXXXX\node_modules\vite\dist\node\chunks\dep-262f13eb.js:59742:30)
    at async transformRequest (C:\Users\XXXXX\node_modules\vite\dist\node\chunks\dep-262f13eb.js:65977:29)
    at async C:\Users\XXXXX\node_modules\vite\dist\node\chunks\dep-262f13eb.js:66068:32
```

**What is Configured?**
Not relevant for this error.
 
- OS: Windows 10
- Browser: Chrome 87
- aws-amplify 3.3.13

Jexordexan:
[...]

PR #7970, opened 2021-03-19
Vite Bundle Size Action Example Demonstrating Vite Errors with library

ErikCH:
This PR is demonstrate the errors with Vite when running with the amplify-js library. 

#### Description of changes

This adds a Vite app to the bundle-size action directory

#### Issue #, if available

The errors scene are in #7499 

To reproduce, run `yarn dev`.
Or
`yarn build`


.github/actions/bundle-size-action/vite/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "serve": "vite preview"
+  },
+  "dependencies": {
+    "aws-amplify": "^3.3.25",
+    "vue": "^3.0.5"
+  },
+  "devDependencies": {
+    "@vitejs/plugin-vue": "^1.1.5",
+    "@vue/compiler-sfc": "^3.0.5",
+    "vite": "^2.1.0"
```
ericclemmons, line 16: Should this use `latest`? I don't know how this project is initialized, so I'm uncertain if these are `latest`, `next`, or specifically defined from some external docs.

ErikCH: This is what's configured after a `yarn create @vitejs/app`

.github/actions/bundle-size-action/vite/index.html
```diff
@@ -0,0 +1,19 @@
...
+    <link rel="icon" href="/favicon.ico" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Vite App</title>
+  </head>
+  <body>
+    <div id="app"></div>
+<script>
+	if(global === undefined){
+		global = window;
+	}
```
ericclemmons, line 14: `yarn dev` initially failed for me, then I changed this to:

```html
		<script>
			// https://docs.amplify.aws/start/getting-started/setup/q/integration/angular#angular-6-support
			window.global = window;
		</script>
```

Now `yarn dev` works!  But `yarn build` still fails...

ericclemmons:
So we learned the following:

1. `@aws-sdk/credential-provider-imds` is only required by `@aws-sdk/credential-provider-node`
2. `@aws-sdk/credential-provider-node` is defined in `runtimeConfig.ts` within the `@aws-sdk/client-*` packages.
3. AWS SDK has a `browser` field for bundlers to use `runtimeConfig.browser.ts` instead, but **Vite doesn't seem to be respecting this**:

    -    https://github.com/vitejs/vite/issues/2329
    -   https://github.com/okta/okta-auth-js/issues/641

**My question is, how can we or `aws-sdk-js-v3` define `exports` or something so that Vite picks up on it since it won't respect `browser`?**

ErikCH:
I have uploaded a newer work around, that should allow amplify to work with Vite. See -> https://github.com/aws-amplify/amplify-js/issues/7499#issuecomment-804386820

PR #7974, opened 2021-03-19
fix: user pool authDetails accept hash secret parameter

maximilianorey:
#### Description of changes
AuthenticationDetails may now receive an extra attribute called SecretHash, in the case the pool have a client_secret.  authenticateUserDefaultAuth function and authenticateUserPlainUsernamePassword were modified to send the SECRET_HASH value to the corresponding endpoints.

#### Issue #, if available
The library doesn't work when user pool have a client_secret. This library fix this, allowing client send the SECRET_HASH value corresponding for the username.
This value can be calculated with the user and client_secret with the following code:

const crypto = require("crypto");
const hashSecret = (clientSecret, username, clientId) =>
  crypto
    .createHmac("SHA256", clientSecret)
    .update(username + clientId)
    .digest("base64");

### DIFF
--- a/packages/amazon-cognito-identity-js/index.d.ts
+++ b/packages/amazon-cognito-identity-js/index.d.ts
@@ -47,6 +47,7 @@ declare module 'amazon-cognito-identity-js' {
 
 		public getUsername(): string;
 		public getPassword(): string;
+		public getSecretHash(): string | null;
 		public getValidationData(): any[];
 	}
 
@@ -70,16 +71,12 @@ declare module 'amazon-cognito-identity-js' {
 	export class CognitoUser {
 		constructor(data: ICognitoUserData);
 
-		public setSignInUserSession(
-			signInUserSession: CognitoUserSession
-		): void;
+		public setSignInUserSession(signInUserSession: CognitoUserSession): void;
 		public getSignInUserSession(): CognitoUserSession | null;
 		public getUsername(): string;
 
 		public getAuthenticationFlowType(): string;
-		public setAuthenticationFlowType(
-			authenticationFlowType: string
-		): string;
+		public setAuthenticationFlowType(authenticationFlowType: string): string;
 		public getCachedDeviceKeyAndPassword(): void;
 
 		public getSession(
@@ -164,7 +161,10 @@ declare module 'amazon-cognito-identity-js' {
 		public sendMFACode(
 			confirmationCode: string,
 			callbacks: {
-				onSuccess: (session: CognitoUserSession, userConfirmationNecessary?: boolean) => void;
+				onSuccess: (
+					session: CognitoUserSession,
+					userConfirmationNecessary?: boolean
+				) => void;
 				onFailure: (err: any) => void;
 			},
 			mfaType?: string,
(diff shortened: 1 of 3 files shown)


dbagia:
@maximilianorey it seems `SECRET_HASH` is also required in the `REFRESH_TOKEN_AUTH` flow. Can this PR be updated to handle this?

thetaPC:
Any more progress on this PR?

thetaPC:
`RespondToAuthChallengeRequest` also requires the `SECRET_HASH`. We might want to consider a different approach as there are several functions that require it.

What if we store the `SECRET_HASH` inside of `CognitoUserPool` instead of `AuthenticationDetails`?

This would minimize the amount of rework on individual functions. We wouldn't have to pass authDetails through parameters and we can call the secret hash through the pool.

haochang:
Can we have some updates on this please?

jweyrich:
@sammartinez I'll speak for myself, and on behalf of the company I work for.

We use `amazon-cognito-identity-js` to integrate our own APIs (run on Lambda) with Cognito. Our user pool has secret enabled. Without the proposed changes to `amazon-cognito-identity-js`, the only way we could make the Cognito integration to work properly was to intercept `fetch` (actually, replace it) and inject `SECRET_HASH` where it's required by Cognito.
For example, the 2 "targets" below require the secret hash for the API call to succeed when the user pool has secret enabled:
- AWSCognitoIdentityProviderService.InitiateAuth
- AWSCognitoIdentityProviderService.RespondToAuthChallenge

I understand that Amplify is designed to be used on the frontend and that it must not include the secret. However, we use `amazon-cognito-identity-js` directly on our backend, not using amplify.
So it makes sense (to me) for Amplify does not to support the secret thing, but it doesn't make sense (to me at least), that `amazon-cognito-identity-js` also doesn't support it.

abdallahshaban557:
@jweyrich - thank you for the excellent breakdown of your requirements. We will take this into consideration. Supporting Authentication with Cognito running on the backend is something that we are exploring.

jweyrich:
@abdallahshaban557 have you discussed this with the team? We'd love to hear an update on this. Any chances of getting this on the roadmap?

abdallahshaban557:
Hi @jweyrich  - We have had conversations about it, but we are still a bit far from providing a solution to this using our `Auth` feature set. We definitely have this in mind though for the revamped version of the Amplify Auth feature sets which we want to support to run in NodeJS environments in the future.

jweyrich:
Lovely! No problem. Thank you for the quick response. We'll be eager to it as soon as it's released.

abdallahshaban557:
Sounds good! Thank you for following up as well! Happy Holidays!

jweyrich:
@abdallahshaban557 do you know if there was any advance on this feature on `amazon-cognito-identity-js` ([specifically this](https://github.com/aws-amplify/amplify-js/pull/7974#issuecomment-1119867514))?
If not, should I submit a feature request? We're still stuck with the ugly workaround I cited above.

abdallahshaban557:
Hi @jweyrich  - I am not longer part of the AWS Amplify team. I am adding @haverchuck and @cwomack to help in answering this.

PR #8027, opened 2021-04-05
fix(@aws-amplify/datastore): add additional type check to util.objectsEqual

iartemiev:
Add a type check to `util.objectsEqual` to ensure we're only attempting to compare 2 objects


amhinson (approved):
LGTM 👍

PR #8044, opened 2021-04-07
chore(@aws-amplify/auth): enable type-checking in unit tests

iartemiev:
* Enabled type-checking our auth unit tests
* Fixed most of the existing test file type errors by setting using `<any>`. We can gradually convert these to actual types if/when we decide to do so, but in many of the cases here where we're testing `private` methods and mocking method implementations, so `any` is the only viable solution.
* This will enable us to write new Auth unit tests that perform type-checking


sammartinez (approved):
LGTM 🌮

nickarocho (approved):
LGTM! Thanks for making these adjustments, Ivan 🎉

amhinson (approved):
LGTM 👍

Issue #7774, opened 2021-02-18
Auth.signUp doesn't work with validationData


pafflu:
**Describe the bug**
When calling `Auth.signUp` with a username, password and a set of validationData and without a set of attributes, it fails with
> Cannot read property '0' of undefined

**To Reproduce**

1. Create a `SignUpParams` Object with username, password, validationData and nothing else
2. Call `Auth.signUp()` with said object

**Expected behavior**
No error.

**Code Snippet**
In Auth.ts, lines 318 through 334:
```
if (attrs) {
    Object.keys(attrs).map(key => {
	    attributes.push(
		    new CognitoUserAttribute({ Name: key, Value: attrs[key] })
	    );
    });
}

const validationDataObject = params['validationData'];
if (validationDataObject) {
    validationData = [];
    Object.keys(validationDataObject).map(key => {
	    validationData.push(
		    new CognitoUserAttribute({ Name: key, Value: attrs[key] })
	    );
    });
}
```
In the second if statement I believe it should be `validationDataObject[key]` instead of `attrs[key]`

evcodes:
Hello pafflu,

Can you please tell me which platform/library you are using so that we can reproduce this issue and get to the root cause of the problem (i.e. React, Vue, Angular, etc).

Thanks!

PilotLeon:
We had the same issue wir v3.3.11. Switching back to version 
`    "aws-amplify": "3.3.11",`
fixed the problem

ryanwalters:
@evcodes This issue appears to have been introduced in @aws-amplify/auth@3.4.18. Running the same code in 3.4.17 works as expected.

calumleask:
I’m also experiencing this issue since updating aws-amplify from 3.3.14 which breaks recaptcha validation in a preSignUp lambda since the recaptcha token is not longer making it to the lambda.

I don’t think the fix from the initial comment will fix the underlying issue since ValidationDataObject isn’t defined as an object, it’s an array which is why the key is being set as “0”.

I believe this line may have been incorrectly changed?

https://github.com/aws-amplify/amplify-js/commit/3a9efb0b596cf2795d7e1424f011f8e59058ecfb#diff-d7133ba15216009e91d0c1632f34aa2279f4e6a03873e701c4abf582e011ed5fL314

Unless the type here is also updated to be an object but that would be a breaking change?

https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/types/Auth.ts#L27

[...]

PR #8045, opened 2021-04-07
fix(@aws-amplify/auth): define validationData type as object literal

nickarocho:
#### Description of changes

- fixes TypeScript error when passing `validationData`
- fixes `ts-jest` linting errors when running `yarn test`
- adds a unit test for passing `validationData` to the `Auth.signUp` method

In JavaScript, users can simply pass `validationData` as an object literal with `'key': 'value'` pairs, but in TypeScript we get the following error:

Screen Shot 2021-04-05 at 12 42 27 PM

This is because the `SignUpParams` interface was defining `validationData` as an array of objects (via `ICognitoUserAttributeData[]`).

This fixes the problem for TypeScript users, but we will still need to update the documentation around any reference of `validationData` to make sure users are just passing object literals rather than an array of objects.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/7774

I made a sample app from scratch, imported Amplify (@latest), imported Auth (@latest), manually called `Auth.signUp`, and passed different sets of `validationData`. 

We do see `"SerializationException"` errors from the server response when passing the values as `integers` (i.e. `"NUMBER_VALUE can not be converted to a String"`), `arrays` (i.e. `"Start of list found where not expected"`, and `objects` (i.e. "Start of structure or map found where not expected."), so maybe it would be better to define the type as:
`validationData?: { [key: string]: string };`.

I also wrote a new test as well to validate the type passed as a param matches what we expect from TypeScript - thanks to @iartemiev's recent updates, these types are now being explicitly checked in the `auth` testing suites.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


amhinson (approved):
Nice! LGTM 👍

iartemiev (approved):
LGTM 👍🏻

WIStudent:
Looking at https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/Auth.ts#L326 I think the intended usage of `validationData` is now with an object containing string values (`CognitoUserAttribute` type expects a string as `Value`)  so I think `validationData`'s type should be `validationData?: { [key: string]: string };`

The current type `validationData?: { [key: string]: any }` does not throw a typescript error if I try to pass something like 
```typescript
validationData: [
  new CognitoUserAttribute({Name: 'foo', Value: 'bar'})
]
```
but the implementation will throw a SerializationException as runtime error. I would rather prefer a typescript error at compile time than a runtime error when a user is using my project.

PR #8050, opened 2021-04-08
Prioritize using `user.username` whenever provided

wlee221:
#### Description of changes

This properly handles `username` input on `confirmSignUp` whenever `usernameAlias` is set to `phone_number`.

__Problem__
The problem happens when we 

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-ui-components/.prettierrc
+++ b/packages/amplify-ui-components/.prettierrc
@@ -1,0 +1,5 @@
+{
+	"trailingComma": "es5",
+	"singleQuote": true,
+	"printWidth": 120
+}
--- a/packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx
@@ -1,279 +1,286 @@
 import { I18n } from '@aws-amplify/core';
 import { Component, Prop, h, State, Watch, Host } from '@stencil/core';
 import {
-  FormFieldTypes,
-  FormFieldType,
-  PhoneNumberInterface,
-  PhoneFormFieldType,
+	FormFieldTypes,
+	FormFieldType,
+	PhoneNumberInterface,
+	PhoneFormFieldType,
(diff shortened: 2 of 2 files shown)


packages/amplify-ui-components/.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
thaddmt, line 1: Is this file meant to be included / are the prettier changes meant to be included in this pr?

wlee221: Good catch, this isn't supposed to be here. I have added that to suppress prettier while I'm working on this, but I created #8136 to better address this.

packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx
```diff
@@ -237,21 +252,12 @@ export class AmplifyConfirmSignUp {
...
-					this.userInput = composePhoneNumberInput(this.phoneNumber);
-				} catch (error) {
-					dispatchToastHubEvent(error);
-				}
-			default:
-				break;
-		}
 		try {
-			if (!this.userInput) throw new Error(Translations.EMPTY_USERNAME);
-			this.userInput = this.userInput.trim();
```
hkjpotato, line 252: all looks good except one minor concern: before, it seems we always sync the local state `userInput` in the form submit action follow by

1.  trim and sync(reassign) : `this.userInput = field(phone|email|username).trim();`.
2. then directly use `this.userInput` in the payload.

now this step seems replaced by the `getUsername` function, I wonder if this will cause out of sync in `userInput` state, and affect the [auto signin follow](https://github.com/aws-amplify/amplify-js/blob/main/packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx#L267), which relies on that state?

wlee221: Nice catch! We should use `getUsername` in the auto `signIn` below as well. Will fix.

hkjpotato (approved):

PR #8088, opened 2021-04-13
Allow users to cancel put and get requests for Storage

### LINKED ISSUE #3098: How to cancel Storage file upload

### LINKED ISSUE #7984: Amplify Storage Storage.get cancel download
For many cases it's necessary cancel download for large files. When users start download there is no way for cancel this.

May be is necessary a method like Storage.cancel or Storage.stop for current download process.

jamesaucode:
#### Description of changes
This PR will allow users to cancel in-flight requests made by `Storage.put` or `Storage.get`. This borrows the idea from a previous PR made on the API module #5753 that maintains a weak map of <Promise, CancelTokenSource> in order to maintain a mapping between a request and it's cancellation token.
Similar to #5753, I removed `async` from the function, as annotating it with `async` will wrap the return into a different `Promise`, which prevents retrieving the corresponding `cancelTokenSource` from the weak map. 

Side note: The [AWS SDK way](https://aws.amazon.com/blogs/developer/abortcontroller-in-modular-aws-sdk-for-javascript/) of doing request cancellation is to pass in an `AbortSignal` to the `s3.send` command, and use the corresponding `AbortContoller`'s `.abort()`, however, because we are using a custom `AxiosHttpHandler`, that solution will not work, which is why we are using `CancelTokenSource` here.

A sample usage of it will be the following:
```javascript
try {
  const getRequest = Storage.get("a_big_file");
  ... // Assuming it takes >1 second
  setTimeout(() => {
    Storage.cancel(getRequest);
  }, 1000);
} catch (error) {
  console.error(error);
}
```
On a successful cancellation, it should throw an error (which is up to the user to handle it)  

#### Issue #, if available
fixes #3098
fixes #7984

Added unit tests to ensure 
1. Cancel tokens are generated for `Storage.put` and `Storage.get` requests
2. The cancelToken's cancel function is called
Also manually tested by uploading and downloading bigger files (e.g arch linux iso), and calling `Storage.cancel` before the request completes. Also tested by having multiple requests at the same time, cancel one of them, and the rest of them completes without problem.

Tested on both Chrome and Firefox.

### Documentation
See: https://github.com/aws-amplify/docs/pull/3141 

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


jamesaucode:
Still need to update the relevant docs

agileurbanite:
> Still need to update the relevant docs

Will that be done as part of this PR or in a separate PR?

packages/storage/src/Storage.ts
```diff
@@ -159,14 +169,47 @@ export class Storage {
...
+	 * @param {string} [message] - A message to include in the cancelation exception
+	 */
+	public cancel(request: Promise<any>, config?, message?: string) {
+		const { provider = DEFAULT_PROVIDER } = config || {};
+		const prov = this._pluggables.find(
+			pluggable => pluggable.getProviderName() === provider
+		);
+		if (prov === undefined) {
+			logger.debug('No plugin found with providerName', provider);
+			Promise.reject('No plugin found in Storage for the provider');
```
hvergara, line 197: in my opinion, we should throw a custom error type, that could be used later to differentiate it from another error.

packages/storage/src/Storage.ts
```diff
@@ -159,14 +169,47 @@ export class Storage {
...
+		}
+		const cancelTokenSource = this._cancelTokenSourceMap.get(request);
+		if (cancelTokenSource) {
+			cancelTokenSource.cancel(message);
+		}
```
hvergara, line 202: just curious: what happen if the cancel token cannot be found? should we emit a debug log at minimum?

packages/storage/src/Storage.ts
hvergara: ok, now I see that you're just replicating some logic from siblings methods. Please discard this message.

packages/storage/src/Storage.ts
```diff
@@ -175,7 +218,17 @@ export class Storage {
...
+		const cancelTokenSource = this.getCancellableTokenSource();
+		const responsePromise = prov.get(key, {
+			...config,
+			cancelTokenSource,
+		});
+		this.updateRequestToBeCancellable(responsePromise, cancelTokenSource);
+		return responsePromise;
+	}
+
+	public isCancel(error: any) {
```
hvergara, line 230:
```suggestion
	public isCancelError(error: any) {
```

jamesaucode:
> > Still need to update the relevant docs
> 
> Will that be done as part of this PR or in a separate PR?

I will probably create separate PR and link it here

packages/storage/src/Storage.ts
svidgen, line 197: Noob question: What is the intent of creating a rejection without returning it?

This pattern has probably been baked into the other methods here for ages, so I'm hesitant to say it's wrong. But, I'm unclear as to how this works or is intended to work. :/

Do we have any tests that reach this branch/line?

packages/storage/src/Storage.ts
```diff
@@ -159,14 +169,47 @@ export class Storage {
...
+
 	/**
 	 * Get a presigned URL of the file or the object data when download:true
 	 *
 	 * @param {string} key - key of the object
 	 * @param {Object} [config] - { level : private|protected|public, download: true|false }
 	 * @return - A promise resolves to either a presigned url or the object
 	 */
-	public async get(key: string, config?): Promise<String | Object> {
+	public get(key: string, config?): Promise<String | Object> {
```
sammartinez, line 212: What happens here for our current customers?

packages/storage/src/Storage.ts
jamesaucode: Good catch, I can't find any documentation that indicates this does anything. There are no unit tests that reaches this breach from I can tell, I will add a test for this.

packages/storage/src/Storage.ts
jamesaucode: I think emitting debug log that says the cancelToken can't be found for this request will be nice

packages/storage/src/Storage.ts
```diff
@@ -186,7 +239,7 @@ export class Storage {
 	 *  progressCallback: function }
 	 * @return - promise resolves to object on success
 	 */
-	public async put(key: string, object, config?): Promise<Object> {
+	public put(key: string, object, config?): Promise<Object> {
```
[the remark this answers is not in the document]
jamesaucode: > But, here and on all other instances of removing async, is removing aysnc actually correct? Does it cause problems if an exception is raised at any point prior to returning the promise?

According to [MDN's documentation on async](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function), wrapping a function with an `async` is the equivalent of wrapping a `Promise.resolve()` around your return value, quoting their example:
```javascript
async function foo() {
   return 1
}
// equivalent to
function foo() {
   return Promise.resolve(1)
}
```

The [ECMA spec](https://tc39.es/ecma262/#sec-async-function-constructor) also has a similar explanation, but not as clear as the MDN docs

So if I understand correctly, `async` function ensures that 
1. Allows users to use `await` instead of chaining `.then`s inside the async function
2. The return value will always be a `Promise`, and it allows user to use `await` inside the function. 

For 1, `prov.get` and `prov.put` both returns a `Promise`, customers can still do `Storage.get('...').then` or `await Storage.get('...')` which is demonstrated in the unit test
For 2, we don't use `await` inside the function.

Therefore I don't believe this is a breaking change. 

> Do we have any tests that throw an exception prior to the return?

Currently no, I am adding some unit tests for this scenario to make sure it works as intended.

This is a great question!

packages/storage/src/Storage.ts
jamesaucode: It should not be breaking as both `get` and `put` already returns a `Promise`, so customers can still `await` or `.then` the function without problem.
You can find a slightly more detailed answer from my [reply](https://github.com/aws-amplify/amplify-js/pull/8088#discussion_r612801006) to Jon.

packages/storage/src/Storage.ts
jamesaucode: Took the suggestion, thank you

packages/storage/src/Storage.ts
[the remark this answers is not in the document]
svidgen: Hey, to clarify, the `await` use-case is not really concerning, since you can actually safely `await` *anything*. For example, this 100% works:

```
(async () => { console.log(await "whatever you want") })();
```

And, although I haven't explicitly tested it, I assume that wrapping the `await`'d usage in a `try-catch` will catch any *exceptional* scenario.

**This is concerning the concerning usage:**

```js
Storage.put(k,o).then(...).catch(...);
```

An explicitly `async` function promises you get a `Promise` back regardless of what happens inside. But, if there's any potential for a function (like `put()`) to throw an exception (*or exit early*), it will not return a `Promise`. But, if we're sure nothing here can blow up or return early, the change is probably OK.

The still-lingering concerns I have are probably just paranoid ignorance. But, can we check all the boxes below?

- [ ] The pluggable interactions can't throw exceptions.
- [ ] The underlying `axios` cancellation token calls cannot throw exceptions.
- [ ] Nothing on the roadmap calls for adding risky code to any of these "semi-async" functions.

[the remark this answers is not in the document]
jamesaucode: Thanks for clarifying, I think I understand your concern now.

* [x]  The pluggable interactions can't throw exceptions.
I am not aware of scenarios where the pluggable interactions can throw exceptions.
* [x]  The underlying `axios` cancellation token calls cannot throw exceptions.
If by exceptions you mean any error, then yes. The cancellation token can throw exceptions, which is used for indicating if the request is aborted or not, otherwise no.
* [x]  Nothing on the roadmap calls for adding risky code to any of these "semi-async" functions.
Looking at the backlog, seems like it won't be the case. Maybe @sammartinez and @agileurbanite will have a better idea on this?

Regardless, if we are concerned about errors from `put()` or `get()` not being caught with `.catch`, we can also wrap those calls with a try catch block and explicitly return a `Promise.reject` inside the catch block. (unless if this can cause any other unintended bad behavior?)

packages/storage/src/Storage.ts
jamesaucode: Update: added a `return` for all these `Promise.reject` and added a few tests cases for these branches

svidgen (approved):
LGTM

packages/storage/src/Storage.ts
```diff
@@ -159,14 +169,47 @@ export class Storage {
...
+	 * @param {Object} [config] - { provider: string }
+	 * @param {string} [message] - A message to include in the cancelation exception
+	 */
+	public cancel(request: Promise<any>, config?, message?: string) {
+		const { provider = DEFAULT_PROVIDER } = config || {};
+		const prov = this._pluggables.find(
+			pluggable => pluggable.getProviderName() === provider
+		);
+		if (prov === undefined) {
+			logger.debug('No plugin found with providerName', provider);
```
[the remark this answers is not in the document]
jamesaucode: It seems like that's the case for the `Storage` package. We can probably add an enum with all the error messages there like how Auth does it https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/common/AuthErrorStrings.ts, but that might be more suitable for another PR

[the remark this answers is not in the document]
agileurbanite: 👍

agileurbanite (approved):
👍

manueliglesias (approved):
LGTM 🎉

packages/storage/src/types/Provider.ts
```diff
@@ -13,6 +13,9 @@
 export interface StorageProvider {
 	// you need to implement those methods
 
+	// cancel an in-flight request
+	cancel?(request: Promise<any>): void;
```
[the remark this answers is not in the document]
jamesaucode: Yes, but I think that might be more suited for another PR because that will involve typing the `get` and `put` return as well.

enriqueavg:
LGTM

packages/storage/src/Storage.ts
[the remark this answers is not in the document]
gvankeerberghen: Hello, I'm arriving here as I'm trying to nail a nasty and rare bug which seems linked to `Storage.put` not throwing when it should. When the connection breaks for instance.

In my case this started to appear only after upgrading aws-amplify, from `3.3` to `4.3`. 

See more info on my case and others comments on this issue here https://github.com/aws-amplify/amplify-js/issues/9252

Given this, do you think the risks foreseen and eliminated here might have materialised?

Issue #7740, opened 2021-02-12
Rekognition does not return ageRange & face attributes while using the Predictions Identify API


rajrajhans:
**Describe the bug**
Rekognition does not return `ageRange`, `attributes`, because of which `Predictions.identify` results in showing `ageRange` and `attributes` (smile, eyeglasses, gender, etc) as `undefined` in `entities`.

**To Reproduce**

1. Use `Predictions.identify` with a source image and log the result to console, as described in the [documentation for Predictions here](https://docs.amplify.aws/lib/predictions/identify-entity/q/platform/js#working-with-the-api).

**Expected behavior**
The result should show valid `ageRange` and `attributes` as per the source image.

**Code Snippet**
```js
// Here, the "image" is being taken from webcam, but you can also use any File as source image, as described in the documentation

if (image) {
  const base64image = image.split(",")[1];
  const binaryImage = atob(base64image);
  const imageLength = binaryImage.length;
  const imageBuffer = new ArrayBuffer(imageLength);
  const imageUintAr = new Uint8Array(imageBuffer);

  for (let i = 0; i < imageLength; i++) {
    imageUintAr[i] = binaryImage.charCodeAt(i);
  }

  const blob = new Blob([imageBuffer], {
    type: "image/jpeg",
  });

  Predictions.identify({
    entities: {
      source: {
        bytes: imageBuffer,
      },
    },
  })
    .then(({ entities }) => {
      console.log(entities);
    })
    .catch((e) => {
      console.log(e);
    });
```

Once this code is executed, it logs the following to the console:

You can see that is is able to detect things like `eyeLeft`, `eyeRight`, etc. but `ageRange` and `smile`, `gender`, `eyeglasses`, etc. shows `undefined`.
![image](https://user-images.githubusercontent.com/32734049/107754175-207b7700-6d47-11eb-9436-df8888b5d88f.png)

When I check the Network request made to Rekognition, I can see that it did not send back `ageRange` and `attributes` for some reason. 
![image](https://user-images.githubusercontent.com/32734049/107754415-72bc9800-6d47-11eb-8f73-72eba412d1c7.png)

**Additional context**
[...]

PR #8095, opened 2021-04-14
fix(predictions-identify): update detectFacesCommand parameters to include all facial attributes (including Emotions types) in the result

ZakariaNaaija:
Issue #, if available:
[#7740
](https://github.com/aws-amplify/amplify-js/issues/7740)
Description of changes:
The context for the undefined problem can be found in the [issue](https://github.com/aws-amplify/amplify-js/issues/7740). As per the documentation for Rekognition JS Client Library [here](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Rekognition.html), by default Rekognition will only send back **boundingBox** and landmarks, which is why Amplify Predictions identify result shows **undefined** in attributes, **ageRange**, and missing **Emotions** attribute. To fix this, I have made the required changes as specified in the Rekognition client documentation [here](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Rekognition.html).


packages/predictions/src/Providers/AmazonAIIdentifyPredictionsProvider.ts
```diff
@@ -462,7 +461,7 @@ export class AmazonAIIdentifyPredictionsProvider extends AbstractIdentifyPredict
 						boundingBox: makeCamelCase(detail.BoundingBox),
 						landmarks: makeCamelCaseArray(detail.Landmarks),
 						ageRange: makeCamelCase(detail.AgeRange),
-						attributes: makeCamelCase(detail, attributeKeys),
+						attributes: faceAttributes,
```
wlee221, line 464: Thanks for the simplification here!

packages/predictions/src/Providers/AmazonAIIdentifyPredictionsProvider.ts
```diff
@@ -367,8 +367,8 @@ export class AmazonAIIdentifyPredictionsProvider extends AbstractIdentifyPredict
 				return Promise.reject(err);
 			});
 
-		const param = { Image: inputImage };
-
+			const param = { Attributes: ['ALL'], Image: inputImage };
+
```
wlee221, line 371: Looks like there are some whitespace typos -- can you align them and remove the extra whitespace?

wlee221 (changes requested):
Thanks for the PR, @ZakariaNaaija! Sorry about the delay here. Just got a style nit but the actual code LGTM.

(aws-sdk-js-v3 detect faces references: [link](https://github.com/aws/aws-sdk-js-v3/blob/10db5ac2aa34d11ecf039ab51071f55985d76cf0/clients/client-rekognition/models/models_0.ts#L2872-L2882)).

packages/predictions/src/Providers/AmazonAIIdentifyPredictionsProvider.ts
wlee221:
```suggestion
		const param = { Attributes: ['ALL'], Image: inputImage };

```

wlee221 (approved):
Thanks again, @ZakariaNaaija!

PR #8038, opened 2021-04-06
fix(@aws-amplify/datastore): consecutive saves 2

iartemiev:
Re-work of https://github.com/aws-amplify/amplify-js/pull/8000
Addresses issues in comments

Issue #, if available:
#7888, #7950, #8012

* Allow consecutive updates to work correctly with the changes in the [update mutation input PR](https://github.com/aws-amplify/amplify-js/pull/7935)

* Enable the follow workflows:
 1. Create new Post, then immediately update
 2. Create new Post, then immediately update with a different field
  3. Create new Post, wait for sync to complete, then update 10 times consecutively
  4. Create new Post, then immediately delete
  5. Create new Post with Comment, reassign Comment to a different Post
  6. Consecutive updates with conditions

* Updated unit tests

* Added e2e tests for all these scenarios: https://github.com/aws-amplify/amplify-js-samples-staging/pull/212

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #8102, opened 2021-04-15
fix(@aws-amplify/datastore): handle merging queued create with incoming update

iartemiev:
Adds support for one more use case to the work in https://github.com/aws-amplify/amplify-js/pull/8038

Create Post, then create another Post and immediately update another field

e2e test PR updated to include this scenario.


manueliglesias (approved):
LGTM 👍

PR #8134, opened 2021-04-20
test(amazon-cognito-identity-js): add unit test coverage to Cognito Identity JS package

nickarocho:
#### Checklist

- [ ] PR description included
- [x] `yarn test` passes


packages/amazon-cognito-identity-js/src/CognitoUserSession.js
```diff
@@ -85,7 +85,7 @@ export default class CognitoUserSession {
 	isValid() {
 		const now = Math.floor(new Date() / 1000);
 		const adjusted = now - this.clockDrift;
-
+
```
nickarocho, line 88: Should remove this space.

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -407,10 +408,10 @@ export default class AuthenticationHelper {
 
 			/*
 			For hex strings starting with 'FF8', 'FF' can be dropped, e.g. 0xFFFF80=0xFF80=0x80=-128
-
+
```
nickarocho, line 411: Should remove this space.

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -407,10 +408,10 @@ export default class AuthenticationHelper {
 
 			/*
 			For hex strings starting with 'FF8', 'FF' can be dropped, e.g. 0xFFFF80=0xFF80=0x80=-128
-
+	
 			Any sequence of '1' bits on the left can always be substituted with a single '1' bit
 			without changing the represented value.
-
+
```
nickarocho, line 414: This too.

packages/amazon-cognito-identity-js/package.json
```diff
@@ -89,54 +90,8 @@
     "eslint-plugin-standard": "^3.0.1",
     "jsdoc": "^3.4.0",
     "react": "^16.0.0",
-    "react-native": "^0.44.0",
+    "react-native": "^0.54.0",
```
nickarocho, line 93: Do we know if this will have any downstream effects from bumping this dep?

packages/amazon-cognito-identity-js/package.json
```diff
@@ -52,7 +52,7 @@
     "doc": "jsdoc src -d docs",
     "lint": "eslint src",
     "lint2": "eslint enhance-rn.js",
-    "test": "jest -w 1",
+    "test": "jest --config ./jest.config.js",
```
nickarocho, line 55: This is a much better convention, IMO. Makes it a lot easier to maintain granular testing config.

nickarocho (commented):
Just a preliminary once-over.

sammartinez:
Can we validate the 1 broken tests? Seems like CircleCI is yelling about it.

packages/amazon-cognito-identity-js/__mocks__/mocks.js
```diff
@@ -0,0 +1,72 @@
...
+ * @param {boolean} success defines if a network request is successful
+ * @param {object?} optional data to return onSuccess, some tests requires specific object values
+ */
+export function netRequestMockSuccess(success, data = {}) {
+	if (success) {
+		jest
+			.spyOn(Client.prototype, 'request')
+			.mockImplementationOnce((...args) => {
+				args[2](null, data);
+			});
```
manueliglesias, line 21: Would you consider this more readable?
up for debate 😅 

```suggestion
			.mockImplementationOnce((...[,,callback]) => {
				callback(null, data);
			});
```

packages/amazon-cognito-identity-js/__mocks__/mocks.js
```diff
@@ -0,0 +1,72 @@
...
+	} else {
+		jest
+			.spyOn(Client.prototype, 'request')
+			.mockImplementationOnce((...args) => {
+				args[2](networkError, null);
+			});
```
manueliglesias, line 27: Same as above

packages/amazon-cognito-identity-js/__mocks__/mocks.js
```diff
@@ -0,0 +1,72 @@
...
+ * @param {string} fnName function name within the AuthenticationHelper class
+ * @returns mockImplemntation of the fnName or throw error
+ */
+export function authHelperMock(fnName) {
+	const fnNames = Object.getOwnPropertyNames(AuthenticationHelper.prototype);
+	if (fnNames.indexOf(fnName) < 0) {
+		throw new Error(
+			`${fnName} provided does not exist in Authentication Helper class`
+		);
+	}
```
manueliglesias, line 42: You can use the `in` operator to make this check without getting all the properties and looking for one.

```suggestion
  if(!(fnName in AuthenticationHelper.prototype)) {
  		throw new Error(
			`${fnName} provided does not exist in Authentication Helper class`
		);
  }
```

packages/amazon-cognito-identity-js/__mocks__/mocks.js
```diff
@@ -0,0 +1,72 @@
...
+			.mockImplementationOnce((...args) => {
+				args[2](null, null);
+			});
+	} else if (fnName === getSalt || fnName === getVerifiers) {
+		jest
+			.spyOn(AuthenticationHelper.prototype, fnName)
+			.mockImplementationOnce(() => {
+				return 'deadbeef';
+			});
+	}
```
manueliglesias, line 55: This might be more readable with a `switch` statement, and having the `mockImplementationOnce` just once

Something like:
```javascript
let implementation = () => {throw new Error('Invalid property')};

switch(fnName) {
  case genHashDevices:
    implementation = (...[,,callback]) => { callback(null, null) };
    break;
  case getSalt:
  case getVerifiers:
    implementation = (...[,,callback]) => 'deadbeef';
    break;
}


jest
	.spyOn(AuthenticationHelper.prototype, fnName)
	.mockImplementationOnce(implementation);
```

packages/amazon-cognito-identity-js/__tests__/AuthenticationHelper.test.js
```diff
@@ -1,17 +1,19 @@
+import 'regenerator-runtime/runtime';
```
manueliglesias, line 1: Why is this one needed?

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
manueliglesias: Looks like this file didn't have changes, if so, let's take it out from the PR

packages/amazon-cognito-identity-js/__tests__/AuthenticationHelper.test.js
```diff
@@ -552,3 +554,184 @@ describe('AuthenticatorHelper', () => {
...
+	test('Test generate hash devices', async () => {
+		const deviceGroupKey = instance.generateRandomString();
+		const username = instance.generateRandomString();
+		const result = await promisifyCallback(
+			instance,
+			'generateHashDevice',
+			deviceGroupKey,
+			username
+		);
+		expect(result).toBe(null);
```
manueliglesias, line 597: What is this testing? Why is a `null` expected in this case?

packages/amazon-cognito-identity-js/__tests__/AuthenticationHelper.test.js
```diff
@@ -552,3 +554,184 @@ describe('AuthenticatorHelper', () => {
...
+	});
+
+	test('Ensure that generateRandomSmallA is non deterministic', () => {
+		const firstSmallA = instance.generateRandomSmallA();
+		const secondSmallA = instance.generateRandomSmallA();
+		expect(firstSmallA).not.toEqual(secondSmallA);
+	});
+
+	test('Generate random strings', () => {
+		expect(typeof instance.generateRandomString()).toEqual('string');
```
manueliglesias, line 613: Should we also assert a length and not just the type of the return value?

packages/amazon-cognito-identity-js/__tests__/index.test.js
```diff
@@ -0,0 +1,8 @@
+import * as index from '../src/index';
+
+test('Index modules are imported/exported properly', () => {
+	const modules = Object.keys(index);
+	modules.forEach(module => {
+		expect(module).toBeTruthy();
+	});
```
amhinson, line 7: Could we do something with a snapshot here to validate what is expected as well? Something similar to this: https://github.com/aws-amplify/amplify-js/blob/main/packages/aws-amplify/__tests__/exports-test.ts

packages/amazon-cognito-identity-js/__tests__/AuthenticationHelper.test.js
```diff
@@ -552,3 +554,184 @@ describe('AuthenticatorHelper', () => {
...
+		const serverValue = new BigInteger('deadbeef', 16);
+		instance.k = new BigInteger('deadbeef', 16);
+		instance.UValue = instance.calculateU(instance.largeAValue, xValue);
+		const result = await promisifyCallback(
+			instance,
+			'calculateS',
+			xValue,
+			serverValue
+		);
+		expect(result).toMatchObject(new BigInteger());
```
manueliglesias, line 653: See comment above about expecting a specific value and not just an assertion on the type. Also, `toMatchObject` is very permissive, we need to expect a specific value

packages/amazon-cognito-identity-js/__tests__/AuthenticationHelper.test.js
```diff
@@ -552,3 +554,184 @@ describe('AuthenticatorHelper', () => {
...
+		const hexB = new BigInteger('deadbeef', 16);
+
+		const hashed = instance.hexHash(
+			instance.padHex(hexA) + instance.padHex(hexB)
+		);
+		const expected = new BigInteger(hashed, 16);
+
+		const result = instance.calculateU(hexA, hexB);
+
+		expect(expected).toEqual(result);
```
manueliglesias, line 667: Nice! yeah, this is what I mean above about asserting a value and not just the type

packages/amazon-cognito-identity-js/__tests__/AuthenticationHelper.test.js
```diff
@@ -552,3 +554,184 @@ describe('AuthenticatorHelper', () => {
...
+		expect(regEx.test(instance.hexHash(hexStr))).toBe(true);
+	});
+
+	test('Hashing a buffer returns a string', () => {
+		const buf = Buffer.from('7468697320697320612074c3a97374', 'binary');
+		expect(typeof instance.hash(buf)).toBe('string');
+	});
+
+	//treating the hkdf algorithm as a blackbox
+	test('Ensure that hkdf algorithm returns a strong hex string', () => {
```
manueliglesias, line 682: Let's reword this test's description, it is expecting the result to be an instance of `Buffer`, not verifying that it is *strong* or a hex string.

packages/amazon-cognito-identity-js/__tests__/CognitoUserPool.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+	afterAll(() => {
+		jest.restoreAllMocks();
+	});
+
+	test('Improper data supplied', () => {
+		const data = { UserPoolId: null, ClientId: clientId };
+		expect(() => {
+			new CognitoUserPool(data);
+		}).toThrowError();
```
manueliglesias, line 17: Can we check on a specific message here?

packages/amazon-cognito-identity-js/__tests__/CognitoUserPool.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+	test('Getting clientId from CognitoUserPool', () => {
+		expect(cognitoUserPool.getClientId()).toBe(clientId);
+	});
+
+	test('Getting userPoolId from CognitoUserPool', () => {
+		expect(cognitoUserPool.getUserPoolId()).toBe(userPoolId);
+	});
+
+	test('Getting user context data without advancedSecurityData', () => {
```
manueliglesias, line 28: can we add a test with advancedSecurityData?

packages/amazon-cognito-identity-js/__tests__/CognitoUserPool.test.js
```diff
@@ -0,0 +1,89 @@
...
+	test('Signing up a user happy path', async () => {
+		jest.spyOn(Client.prototype, 'request').mockImplementation((...args) => {
+			args[2](null, {});
+		});
+
+		//returns a function that records everything being done to it
+		const callback = jest.fn();
+
+		cognitoUserPool.signUp(userName, password, [], [], callback, []);
+		//mockCallback.mock.calls[1][0]).toBe(1);
```
manueliglesias, line 67: Remove this

packages/amazon-cognito-identity-js/__tests__/CognitoUserSession.test.js
```diff
@@ -0,0 +1,49 @@
+import CognitoUserSession from '../src/CognitoUserSession';
+import {
+	ivCognitoUserSession,
+	ivRefreshToken,
+	ivAccessToken,
+	ivCognitoIdToken,
+	ivRefreshToken,
```
manueliglesias, line 7: I am curious, what does `iv` stand for?

packages/amazon-cognito-identity-js/__tests__/CognitoUserSession.test.js
```diff
@@ -0,0 +1,49 @@
...
+		const accessToken = ivCognitoUserSession.getAccessToken().getIssuedAt();
+		const min = Math.min(idToken, accessToken);
+		const expectedValue = currDate - min;
+
+		expect(ivCognitoUserSession.calculateClockDrift()).toBe(expectedValue);
+	});
+
+	test('JWT Expiration was hard-coded to be a time in the past so that this is guaranteed to be an invalid token', () => {
+		expect(ivCognitoUserSession.isValid()).toBe(false);
+	});
```
manueliglesias, line 48: Can you add a test for the valid case?

packages/amazon-cognito-identity-js/__tests__/CookieStorage.test.js
```diff
@@ -0,0 +1,64 @@
+import CookieStorage from '../src/CookieStorage';
+import { cookieStorageDomain } from './constants';
+
+describe('Cookie Storage Unit Tests', () => {
+	//defining a DOM to attach a cookie to
+	Object.defineProperty(window.document, 'cookie', { writable: true });
```
manueliglesias, line 6: Can we instead use `jsdom` as environment?

https://jestjs.io/docs/configuration#testenvironment-string

packages/amazon-cognito-identity-js/__tests__/CookieStorage.test.js
```diff
@@ -0,0 +1,64 @@
...
+
+			test('setting an item', () => {
+				expect(cookieStore.setItem('domain', 'newdomain.com')).toBe(
+					'newdomain.com'
+				);
+			});
+
+			test('Clearing cookies should return an empty object', () => {
+				expect(cookieStore.clear()).toEqual({});
+			});
```
manueliglesias, line 61: Can we check they are actually deleted and not only the return value?

packages/amazon-cognito-identity-js/__tests__/DateHelper.test.js
```diff
@@ -0,0 +1,10 @@
+import DateHelper from '../src/DateHelper';
+
+// DateHelper is a utility class that provides the date in "ddd MMM D HH:mm:ss UTC YYYY" format.
+describe('DateHelper unit tests', () => {
+	test('Expect getNowString() to return the date in the correct format', () => {
+		const result = new DateHelper();
+		const year = new Date().getUTCFullYear();
+		expect(result.getNowString()).toContain(year);
```
manueliglesias, line 8: Let's have this test look at the format of the full string, not just that it containes the year.

Maybe you can mock the Date fn to return fixed values and write the expect accordingly

packages/amazon-cognito-identity-js/__mocks__/mocks.js
evcodes: javascript is a strange, strange beast

packages/amazon-cognito-identity-js/__tests__/index.test.js
nickarocho: Fixed! https://github.com/aws-amplify/amplify-js/pull/8134/commits/48fbd3bcfc1662c598e5edb977169d0b0ab8b4be

packages/amazon-cognito-identity-js/__tests__/AuthenticationHelper.test.js
```diff
@@ -552,3 +554,184 @@ describe('AuthenticatorHelper', () => {
...
+		expect(typeof instance.generateRandomString()).toEqual('string');
+	});
+
+	test('Generate random strings is non-deterministic', () => {
+		expect(instance.generateRandomString()).not.toEqual(
+			instance.generateRandomString()
+		);
+	});
+
+	test('Calculate A works successfully', async () => {
```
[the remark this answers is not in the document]
evcodes: yeah I agree, after looking at this now, it's obvious that this should have been it's own suite with unit tests for each case

packages/amazon-cognito-identity-js/__tests__/CognitoUserPool.test.js
evcodes: I did not know how to do this, here's what it said in the comments:

This method returns the encoded data string used for cognito advanced security feature.
	 * This would be generated only when developer has included the JS used for collecting the
	 * data on their client. Please refer to documentation to know more about using AdvancedSecurity
	 * features

packages/amazon-cognito-identity-js/__tests__/CognitoUserSession.test.js
evcodes: valid/invalid is just defined by the access and refresh tokens so either way it would work as expected but I changed it have tests for the valid sessions too.

packages/amazon-cognito-identity-js/__tests__/CookieStorage.test.js
evcodes: cool!

packages/amazon-cognito-identity-js/__tests__/CognitoIdToken.test.js
```diff
@@ -0,0 +1,8 @@
+import CognitoIdToken from '../src/CognitoIdToken';
+
+describe('Constructor for CognitoId Token', () => {
+	test('Constructing a CognitoID Token', () => {
+		const cognitoToken = new CognitoIdToken('testToken');
+		expect(typeof cognitoToken.jwtToken).toBe('string');
```
iartemiev, line 6: Nit: I think it's better to expect the actual value `''`. 
When checking `typeof` we don't know if maybe `testToken` actually got assigned to `jwtToken` somehow.

```suggestion
		expect(cognitoToken.jwtToken).toBe('');
```

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
iartemiev: There are some minor formatting changes it looks like, e.g., https://github.com/aws-amplify/amplify-js/pull/8134/files#diff-4e7fc32caed143fe1cc3fd688b63f9b695af679bc065c5dd6210f9be984894e2L33-R34

Maybe we had never run Prettier on it before 🤷🏻‍♂️

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
```diff
@@ -0,0 +1,1623 @@
...
+		callback.onSuccess.mockClear();
+		callback.onFailure.mockClear();
+	});
+
+	afterAll(() => {
+		jest.restoreAllMocks();
+	});
+
+	test('happy case should callback onSuccess with the token', () => {
+		netRequestMockSuccess(true);
```
jamesaucode, line 761: Another nit: I wonder if it would be beneficial or a bit overkill if we can have a mock net request function that also takes an operation name? At least for this block it's not immediately clear that `netRequestMockSuccess` means 'mock a successful request once' or that `verifySoftwareToken` does 2 calls, if it's `netRequestMockSuccess(true, 'VerifySoftwareToken)` then `netRequestMockSuccess(true, 'RespondToAuthChallenge`) then it will be easier to understand

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
```diff
@@ -0,0 +1,1623 @@
...
+	});
+
+	test('Happy case for signed in user session', () => {
+		cognitoUser.setSignInUserSession(vCognitoUserSession);
+		netRequestMockSuccess(true);
+		cognitoUser.verifySoftwareToken(totpCode, deviceName, callback);
+		expect(callback.onSuccess.mock.calls.length).toBe(1);
+	});
+
+	test('Error case for non-signed in user session', () => {
```
jamesaucode, line 789: I think you might need to set `cognitoUser`'s session to an invalid one to test the non-signed in case here

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
```diff
@@ -0,0 +1,1623 @@
...
+	afterAll(() => {
+		jest.restoreAllMocks();
+	});
+
+	afterEach(() => {
+		callback.onSuccess.mockClear();
+		callback.onFailure.mockClear();
+	});
+
+	test('signOut expected to set signinUserSession to equal null', () => {
```
jamesaucode, line 875: I think we should check if the cached user is cleared too

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
```diff
@@ -0,0 +1,1623 @@
...
+		expect(callback.onSuccess.mock.calls.length).toEqual(1);
+	});
+
+	test('when inputVerificationCode exists in the callback, call inputVerifier with the data', () => {
+		netRequestMockSuccess(true);
+		cognitoUser.getAttributeVerificationCode(...getAttrsVerifCodeDefaults);
+		expect(callback.inputVerificationCode.mock.calls.length).toEqual(1);
+	});
+
+	test('when inputVerificationCode exists in the callback, call inputVerifier with the data', () => {
```
jamesaucode, line 1116: This doesn't match what it's doing in the test, I am guessing this is the failure case?

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
```diff
@@ -0,0 +1,1623 @@
...
+	});
+
+	test('client returning an error should call onFailure', () => {
+		netRequestMockSuccess(false);
+		cognitoUser.forgotPassword(...forgotPasswordDefaults);
+		expect(callback.onFailure.mock.calls.length).toEqual(1);
+	});
+});
+
+describe('MFA test suite', () => {
```
jamesaucode, line 1187: Should we just remove the 'sendMFAcode' test block you have in line 734? That one seems redundant, and this one is a lot more robust

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
```diff
@@ -0,0 +1,1623 @@
...
+	test('having an invalid user session should callback with a new error', () => {
+		cognitoUser.setSignInUserSession(ivCognitoUserSession);
+		cognitoUser.getUserAttributes(callback);
+		expect(callback.mock.calls[0][0]).toMatchObject(
+			new Error('User is not authenticated')
+		);
+	});
+});
+
+describe('getCognitoUserSession()', () => {
```
jamesaucode, line 1498: I think this doesn't test the behavior of the function properly, `getCognitoUserSession(authResult)` uses the `authResult` to build the tokens and return them, instead of using the user's `signInUserSession`

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
```diff
@@ -0,0 +1,1623 @@
...
+			new CognitoUserSession(sessionData)
+		);
+	});
+	test('client throws an error ', () => {
+		netRequestMockSuccess(false);
+		cognitoUser.refreshSession(...refreshSessionDefaults);
+		expect(callback.mock.calls[0][0]).toMatchObject(new Error('Network Error'));
+	});
+
+	describe('getSession()', () => {
```
jamesaucode, line 1555: I think this will be better suited to be in a separate test block instead of nested in the `refreshSession` test block

jamesaucode (commented):
Great work!! This is a huge test file

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
```diff
@@ -0,0 +1,1623 @@
...
+	test('AuthHelper generateHashDevice with no error calls auth methods', () => {
+		const randomPasswordSpy = jest.spyOn(
+			AuthenticationHelper.prototype,
+			'getRandomPassword'
+		);
+		const getVerifierDevicesSpy = jest.spyOn(
+			AuthenticationHelper.prototype,
+			'getVerifierDevices'
+		);
+		const spyon3 = jest.spyOn(
```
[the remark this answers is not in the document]
evcodes: fixed, I think he meant that to be generateHashDevice instead

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
evcodes: I see your point, I propose to do something like this:

`export function netRequestMockSuccess(success, data = {}, operationName = '') {
	if (success) {
		jest
			.spyOn(Client.prototype, 'request')
			.mockImplementationOnce((...[operationName, , callback]) => {
				callback(null, data);
			});
	} else {
		jest
			.spyOn(Client.prototype, 'request')
			.mockImplementationOnce((...[, , callback]) => {
				callback(networkError, null);
			});
	}
}`
and inside the CognitoUser.test.js:
`	netRequestMockSuccess(true, {}, 'verifySoftwareToken');
	netRequestMockSuccess(true, {}, 'RespondToAuthChallenge');
`

I am not sure if you think that having this parameter is beneficial since we would have to explicitly provide an empty object for each time that we want to use the operation name, which is only helpful within the testing file. Open to other possible suggestions

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
evcodes: Sorry, this was meant to be `error case for signed in user session`

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
evcodes: agreed

packages/amazon-cognito-identity-js/__tests__/StorageHelperRN.test.js
```diff
@@ -0,0 +1,126 @@
...
+
+			const StorageHelper = require('../src/StorageHelper-rn.js').default;
+			const storageHelper = new StorageHelper();
+			const callback = jest.fn();
+
+			//localStorage would then be out of sync with AsyncStorage
+			storageHelper.getStorage().setItem('key', 'value');
+			storageHelper.getStorage().sync(callback);
+			expect(callback.mock.calls[0][1]).toEqual('SUCCESS');
+		});
```
iartemiev, line 84: I think here we should first check that AsyncStorage and MemoryStorage are actually out of sync by attempting to get a value by one of those keys and `expect` it to be `undefined`. Then perform the sync and get it again, this time `expecting` it to equal the value.
```suggestion
		test('Happy case for syncing', () => {
			const items = {
				'@MemoryStorage:key': 'value1',
			};
			jest.mock('react-native', () => ({
				AsyncStorage: {
					setItem: jest.fn((item, value) => {
						return new Promise((resolve, reject) => {
							items[item] = value;
							resolve(value);
						});
					}),
					getAllKeys: jest.fn(callback => {
						callback(null, Object.keys(items));
					}),
					multiGet: jest.fn((keys, callback) => {
						const values = keys.map(key => [key, items[key] || null]);
						callback && callback(null, values);

						return values;
					}),
				},
			}));

			const StorageHelper = require('../src/StorageHelper-rn.js').default;
			const storageHelper = new StorageHelper();
			const callback = jest.fn();

			// MemoryStorage would then be out of sync with AsyncStorage
			expect(storageHelper.getStorage().getItem('key')).toBeUndefined();
			storageHelper.getStorage().sync(callback);
			expect(callback.mock.calls[0][1]).toEqual('SUCCESS');
			expect(storageHelper.getStorage().getItem('key')).toEqual('value1');
		});
```

iartemiev (changes requested):
Just a couple minor changes requested to StorageHelper.test and StorageHelperRN.test

packages/amazon-cognito-identity-js/__tests__/StorageHelperRN.test.js
```diff
@@ -0,0 +1,126 @@
...
+			expect(storageHelper.getStorage().clear()).toEqual({});
+		});
+	});
+
+	describe('Testing StorageHelper sync', () => {
+		beforeEach(() => {
+			jest.resetModules();
+		});
+		test('Happy case for syncing', () => {
+			var items = { key: 'value1', key2: 'value2' };
```
iartemiev, line 55: Let's have `items` simulate the contents of AsyncStorage
```suggestion
			const items = {
				'@MemoryStorage:key': 'value1',
			};
```

packages/amazon-cognito-identity-js/__tests__/StorageHelperRN.test.js
```diff
@@ -0,0 +1,126 @@
...
+			}));
+
+			const StorageHelper = require('../src/StorageHelper-rn.js').default;
+			const storageHelper = new StorageHelper();
+			const callback = jest.fn();
+
+			//localStorage would then be out of sync with AsyncStorage
+			storageHelper.getStorage().setItem('key', 'value');
+			storageHelper.getStorage().sync(callback);
+			expect(callback.mock.calls[0][1]).toEqual('SUCCESS');
```
iartemiev, line 83: I think here we should first check that AsyncStorage and MemoryStorage are actually out of sync by attempting to get a value by one of those keys and expect it to be undefined. Then perform the sync and get it again, this time expecting it to equal the value.
```suggestion
			// MemoryStorage would then be out of sync with AsyncStorage
			expect(storageHelper.getStorage().getItem('key')).toBeUndefined();
			storageHelper.getStorage().sync(callback);
			expect(callback.mock.calls[0][1]).toEqual('SUCCESS');
			expect(storageHelper.getStorage().getItem('key')).toEqual('value1');
```

iartemiev (changes requested):
Re-structured one of my requested changes to make it easier to read

iartemiev:
One more minor suggestion. Let's add:
```yaml
overrides:
  files: *.test.js
  rules: jest/valid-expect
```
Right above [this line ](https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/.eslintrc.yaml#L9) to prevent ESLint from yelling at us in the test files

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
evcodes: you're right and I just changed it to pass in the session data

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
evcodes: done :)

packages/amazon-cognito-identity-js/__tests__/StorageHelperRN.test.js
evcodes:
```
// MemoryStorage would then be out of sync with AsyncStorage
expect(storageHelper.getStorage().getItem('key')).toBeUndefined();
storageHelper.getStorage().setItem('key', 'value1');
storageHelper.getStorage().sync(callback);
expect(callback.mock.calls[0][1]).toEqual('SUCCESS');
expect(storageHelper.getStorage().getItem('key')).toEqual('value1');
```

packages/amazon-cognito-identity-js/__tests__/StorageHelperRN.test.js
```diff
@@ -0,0 +1,130 @@
...
+				},
+			}));
+
+			const StorageHelper = require('../src/StorageHelper-rn.js').default;
+			const storageHelper = new StorageHelper();
+			const callback = jest.fn();
+
+			// MemoryStorage would then be out of sync with AsyncStorage
+			expect(storageHelper.getStorage().getItem('key')).toBeUndefined();
+			storageHelper.getStorage().setItem('key', 'value1');
```
iartemiev, line 84: Should be no need for this line. The `sync` should add it to MemoryStorage from AsyncStorage

iartemiev (changes requested):

packages/amazon-cognito-identity-js/__tests__/StorageHelperRN.test.js
iartemiev: Otherwise, we don't know whether the `sync` method is actually working

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
jamesaucode: I can think of a few options.
1. Make the function take operation name as well in second parameter so it becomes `netRequestMockSuccess(true, { data: someData, operationName: 'verifySoftwareToken'})`
But then you will need to change all the other tests that calls this function :/
2. Make another function called `netRequestMockWithOperationName`
3. Add more comments to clarify that the function actually does 2 separate network calls, so 2 mocks are necessary (and that each netRequestMock call only mocks once)
Up for other suggestions!

evcodes: I think I will go with the comments. Since they are mocking network calls, the details usually doesn't matter but sometimes we need to have some output with specific fields. That's why I added the optional data parameter

iartemiev (approved):
LGTM 🎉 Thank you guys, this is great!

PR #8136, opened 2021-04-21
chore(@aws-amplify/ui-components): Local prettier run on ui-components

wlee221:
#### Description of changes

This is a follow-up to https://github.com/aws-amplify/amplify-js/pull/7790, which consolidated `prettierrc` at the root. The prettier changes haven't been applied to ui-components files yet, so this pr runs prettier on them. Submitting this so that future UI Components PR are free of unrelated style changes.

`yarn test` and `eslint` passed. The only code change was disabling an eslint rule that forces `decorators` to be inline with the variable. See comment below.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-ui-components/.eslintrc.js
```diff
@@ -1,40 +1,41 @@
...
+		'@typescript-eslint/no-empty-interface': 0,
+		'@typescript-eslint/no-explicit-any': 0,
+		'@typescript-eslint/no-unused-vars': 0,
+		'@typescript-eslint/camelcase': 0,
+		'@typescript-eslint/no-inferrable-types': 0,
+		'@typescript-eslint/explicit-module-boundary-types': 0,
+		'@typescript-eslint/no-empty-function': 0,
+		'@typescript-eslint/ban-ts-comment': 0,
+		'@typescript-eslint/ban-types': 0,
+		'@stencil/decorators-style': 0
```
wlee221, line 39: This is the only code change made. With reduced printwidth from 120 to 80, decorators were moved to newlines which eslint didn't like. This added line disables that rule.

.github/actions/bundle-size-action/webpack/package.json
```diff
@@ -47,7 +47,7 @@
       },
       {
         "path": "dist/withSSRContext+Storage.js.min.js",
-        "maxSize": "181kB"
```
wlee221, line 50: One more file with actual code change! 😛

.github/actions/bundle-size-action/webpack/package.json
```diff
@@ -43,11 +43,11 @@
       },
       {
         "path": "dist/withSSRContext.js.min.js",
-        "maxSize": "146kB"
+        "maxSize": "155kB"
```
wlee221, line 46: Increasing this to 155kB as well (+9 kB)

thaddmt: Out of curiosity was there a reason that we chose +9 kB and is that likely to need to change again?

wlee221: No particular reason for +9kB. But the goal is that this should be small enough to catch unusually high increase in bundle size. This will need to be changed again eventually as our bundle size creeps up little by little, which I'm okay with as long as we're capable of catching unusual bundle size increase.

ericclemmons (approved):

svidgen (approved):

Issue #8116, opened 2021-04-18
Authentication with Amazon Cognito


afonsecac:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm install -g @aws-amplify/cli

### If applicable, what version of Node.js are you using?

v14.15.1

### Amplify CLI Version

4.48.0

### What operating system are you using?

Ubuntu

### Amplify Categories

auth

### Amplify Commands

Not applicable

### Describe the bug

(0, _getRandomBase.default) is not a function. (In '(0, _getRandomBase.default)(array.byteLength)', '(0, _getRandomBase.default)' is undefined)

### Expected behavior

User Cognito when Auth.signIn(username, password)

### Reproduction steps

1. amplify add api
2. amplify push
3. Configure Amazon Cognito by username
4. Enable MFA
5. Register User
6. Confirm User
7. Error when auth.signIn(username, password)

### GraphQL schema(s)

<details>

```graphql
# Put schemas below this line


```

</details>

Antoine-C:
It seems like a duplicate of #8113

Issue #7916, opened 2021-03-10
Upgraded amplify version broke auth


imoby:
**Describe the bug**
I've upgraded my aws-amplify version to the latest and my auth no longer seems to work and throws various errors that I can't seem to figure out where/why its going on. Reverting back to the older version works again.

Old:
"aws-amplify": "^3.0.20",
"aws-amplify-react-native": "^4.2.1",

New:
"aws-amplify": "3.3.23",
"aws-amplify-react-native": "^4.3.1",

I'm using the sign in/signup functionality the same way its mentioned in the docs online. Below are the two different errors that the sign in method is throwing at me:

`[TypeError: (0, _getRandomBase.default) is not a function. (In '(0, _getRandomBase.default)(array.byteLength)', '(0, _getRandomBase.default)' is undefined)]
`

And after getting that error and attempting to sign in again, it throws this:

`
[Error: Pending sign-in attempt already in progress]
`
**To Reproduce**

1. Upgrade from older version of amplifyjs
2. Using the same docs to sign in, attempt to sign in
3. Error is thrown

**Expected behavior**
I should be able to login normally

**Code Snippet**
Please provide a code snippet or a link to sample code of the issue you are experiencing to help us reproduce the issue. (Be sure to remove any sensitive data)

**Screenshots**

sammartinez:
Hello @imoby, Would you mind sharing a code snippet or screenshots of what you are experiencing? Also, can you provide your `package.json` along with your environment settings? 

Please use the follow command: 
```
npx envinfo --system --binaries --browsers --npmPackages --npmGlobalPackages
```

Thanks ahead of time!

sammartinez:
In testing the documentation on our docs site, I cannot reproduce this issue. @imoby have you done a `rm -rf node_modules` and re-install your dependencies? You may also need to remove your `package-lock.json` or `yarn.lock` file if you have one. Please let us know how we can help you further.

imoby:
@sammartinez Im going to give you some of the items you requested later today but I have definitely removed my node modules and lock and the issue still has happened.

imoby:
I was able to update successfully after running a clean build but only through xcode directly. Not sure why a dependency like amplify would be affected there but so far things seem to be good now. if theres any update I'll post here.

svidgen:
Is this working for you now? If not, are you using expo by chance?

The error in the OP seems the same as this one: https://github.com/aws-amplify/amplify-js/issues/8113

[...]

Issue #7679, opened 2021-02-03
React Native Expo Auth.signIn shows error


billzeng2k:
**Describe the bug**
When I use the function Auth.signIn it gives me error
`[TypeError: (0, _getRandomBase.default) is not a function. (In '(0, _getRandomBase.default)(array.byteLength)', '(0, _getRandomBase.default)' is undefined)]`
Only this function produces an error, Sign Up and Federated Sign In both work perfectly

**To Reproduce**

Initialize an expo project by following steps on https://docs.amplify.aws/lib/auth/getting-started/q/platform/js and call the Auth.signIn command

**Expected behavior**
I expect to sign in or an error

**Code Snippet**
package.json
```
{
  "main": "node_modules/expo/AppEntry.js",
  "scripts": {
    "start": "expo start",
    "android": "expo start --android",
    "ios": "expo start --ios",
    "web": "expo start --web",
    "eject": "expo eject"
  },
  "dependencies": {
    "@eva-design/eva": "^2.0.0",
    "@react-native-community/masked-view": "0.1.10",
    "@react-native-community/netinfo": "^5.9.10",
    "@react-navigation/native": "^5.9.2",
    "@react-navigation/stack": "^5.14.2",
    "@ui-kitten/components": "^5.0.0",
    "@ui-kitten/eva-icons": "^5.0.0",
    "@ui-kitten/moment": "^5.0.0",
    "amazon-cognito-identity-js": "^4.5.10",
    "aws-amplify": "^3.3.17",
    "aws-amplify-react-native": "^4.3.1",
    "expo": "^40.0.1",
    "expo-status-bar": "~1.0.3",
    "expo-web-browser": "~8.6.0",
    "lottie-react-native": "~2.6.1",
    "moment": "^2.29.1",
    "react": "16.13.1",
    "react-dom": "16.13.1",
    "react-native": "https://github.com/expo/react-native/archive/sdk-40.0.1.tar.gz",
    "react-native-gesture-handler": "~1.8.0",
    "react-native-modal": "^11.6.1",
    "react-native-reanimated": "~1.13.0",
    "react-native-safe-area-context": "3.1.9",
    "react-native-screens": "^2.15.2",
    "react-native-shimmer": "^0.6.0",
    "react-native-svg": "^12.1.0",
    "react-native-web": "~0.13.12"
  },
  "devDependencies": {
    "@babel/core": "^7.9.6",
    "@babel/plugin-proposal-decorators": "^7.12.13",
[...]

PR #8162, opened 2021-04-22
fix(amazon-cognito-identity-js): resolve missing getRandomBase64 implementation in expo >= 41 

svidgen:
#### Description of changes

Expo 41 [removes handling for `.expo.js` files](https://github.com/expo/fyi/blob/master/expo-extension-migration.md) &mdash; *they are no longer imported*. Because of this, builds using expo >= 41 were missing a `getRandomBase64` implementation, and `Auth.signIn()` was therefore failing in **production** builds. (DEV builds were also missing the implementation, but [DEV builds do not actually use it](https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/src/utils/getRandomValues.native.js#L49-L53).)

Expo *does* still import `.native.js` files. So, this change conditionally uses expo's `getRandomBase64String` implementation in the *native* import.

#### Issue #, if available

Expected to solve:

* https://github.com/aws-amplify/amplify-js/issues/8116
* https://github.com/aws-amplify/amplify-js/issues/8113
* https://github.com/aws-amplify/amplify-js/issues/7916
* https://github.com/aws-amplify/amplify-js/issues/7679

1. Created a trivial expo react-native application.
2. Added amplify auth. 
3. Observed the error in the iOS simulator.
4. Made the change.
5. Observed successful signIn in the iOS simulator.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amazon-cognito-identity-js/src/utils/getRandomBase64.native.js
```diff
@@ -1,8 +1,12 @@
...
-const getRandomBase64 = NativeModules.RNAWSCognito
+
+const getRandomBase64 = NativeModules.ExpoRandom
+	? NativeModules.ExpoRandom.getRandomBase64String
+	: NativeModules.RNAWSCognito
 	? NativeModules.RNAWSCognito.getRandomBase64
 	: () => {
-			throw Error('Validate that amazon-cognito-identity-js has been linked');
+			throw new Error(
+				'Could not find a native getRandomBase64 implementation.'
```
amhinson, line 9: It could still be helpful here to suggest that `amazon-cognito-identity-js` needs to be linked. `ExpoRandom` should always be available for Expo apps, so this error message only has the possibility of being thrown in a regular RN app. With that said, perhaps we could combine the previous message with this new one? I think the combination of the two will give more context to the customer.

svidgen: Yeah. I'm not entirely sure why I dropped the instructive part of this error. Re-adding ...

ericclemmons (approved):
This looks good!  Nothing particular to block over, but I'm good to merge if the answer is based on the following questions:

- [ ] Will this also work for Expo < 41? (Looks like it should, but I don't know when `ExpoRandom.getRandomBase64String` was introduced.
- [ ] Do we have Expo integration tests? (I know we have some React Native ones) How do we avoid any regressions for this critical-path code?

ericclemmons:
The bundle-size fixes are addressed in https://github.com/aws-amplify/amplify-js/pull/8136 via ddd82b8

amhinson (approved):
LGTM 👍

adilusman51:
Detail of these changes are provided here [https://github.com/expo/fyi/blob/master/expo-extension-migration.md](https://github.com/expo/fyi/blob/master/expo-extension-migration.md)
Kindly merge this fix so that Expo Client and Expo Go Apps can work properly with aws-amplify

sammartinez (approved):
LGTM 🌮

svidgen:
FYI. Now that this in, I've tested on expo 39 .. 41 against `unstable` in both prod and development mode, and this fix looks to have worked.

ericclemmons:
Thanks for the final validation @svidgen! Really great, thorough work!!

sreimer15:
How do you install this version for 
aws-amplify
and
aws-cognito-identity-js
when I try to install both with the git hashes it says I need a peer dependency of graphql? Do I have to wait for the next release? Thank you!

EvanBacon:
@ericclemmons 
> Expo integration tests

In Expo SDK 41 we removed all of the JS bundling differences between bare and managed workflow so you don't need to special case between Expo and React Native for this anymore. 

If we ever introduce differences in the future it will be at the bundler level, so using metro.config.js with "expo/metro-config" should suffice. https://docs.expo.io/guides/customizing-metro/

Issue #7837, opened 2021-03-01
Login inputs not treated as a single form in new Angular UI


lkowalski-smartbusiness:
**Describe the bug**
The login form is not treated as a "whole" in the new Angular UI Auth module
Resulting issue is that none of my users can use Chrome's account autosuggestion - the password just never shows up (or the login, if you go straight for the password input, which nobody ever does). Even if you try forcing a new entry via "Suggest a password" it just stubbornly stays blank.

From a brief review of the new dynamic component present in the login form I could see that the shadow-dom is used extensively. This poses an issue when the form inputs are scattered across multiple encapsulated DOMs, not connected by an ancestor `<form>` element.

**To Reproduce**
_The steps provided assume you have credentials stored in a Chromium-based browser for your local development page_
--- steps to get a valid aws-exports.js file
1. Create new amplify project
2. Select Angular as frontend component
3. Add auth
--- steps for the error
4. install `aws-amplify ^3.3.21`, `@aws-amplify/ui-components ^1.0.3` and `@aws-amplify/ui-angular ^1.0.3`
4. add AmplifyUIAngularModule to app.modules.ts
5. put `<amplify-authenticator></amplify-authenticator>` in the `app.component.html` template
6. run the app and try to autopopulate

**Expected behavior**
The user is able to autofill *both* input fields in the login page if he's saved the credentials before

**Screenshots**
![image](https://user-images.githubusercontent.com/64459849/109492688-a4459b00-7a8b-11eb-9fe1-827150ec25f2.png)

**What is Configured?**
If applicable, please provide what is configured for Amplify CLI:
* Auth
* Hosting
* Angular as frontend
 
<details>
  <summary><strong>Environment</strong></summary>

```
   System:
    OS: macOS 11.2.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 1.53 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.19.0 - ~/.nvm/versions/node/v12.19.0/bin/node
    npm: 7.6.0 - ~/.nvm/versions/node/v12.19.0/bin/npm
  npmPackages:
    @angular-devkit/build-angular: ~0.1102.0 => 0.1102.0 
    @angular/animations: ~11.2.0 => 11.2.0 
    @angular/cdk: ~11.2.0 => 11.2.0 
    @angular/cli: ~11.2.0 => 11.2.0 
[...]

Issue #6782, opened 2020-09-12
amplify-authenticator component prevents save password


srchulo:
**Describe the bug**
Browsers do not prompt to save username/password after logging in with amplify authenticator:

https://docs.amplify.aws/ui/auth/authenticator/q/framework/vue

**To Reproduce**
1. Go to a login page with amplify authenticator on it
2. Enter username and password
3. Login

**Expected behavior**
Browsers should prompt you to save username/password.

**Additional context**
LastPass also doesn't prompt to save password after logging in.

srchulo:
@mauerbac thanks! I guess thought this was a bug since I would assume it would work with however browsers normally autodetect this? Sorry for putting into docs-- didn't realize I did that!

jcardus:
It doesn't even tell the browser to remeber the email / username. I'm tryed using inputPros and set autocomplete but nothing. There's no way the browser saves the user / password...

ChristopheBougere:
I'm having the same issue (with react).
The browser was prompting well with the old system, this is a regression with the new auth component.

camhart:
Is the shadow-root stuff new?  If so, that could be the cause (see https://bugs.chromium.org/p/chromium/issues/detail?id=746593).

camhart:
> I'm having the same issue (with react).
> The browser was prompting well with the old system, this is a regression with the new auth component.

What was the old system?  Do I just need to roll a version number back?

ChristopheBougere:
> > I'm having the same issue (with react).
> > The browser was prompting well with the old system, this is a regression with the new auth component.
> 
> What was the old system? Do I just need to roll a version number back?

I mean the [legacy one](https://docs.amplify.aws/ui-legacy/q/framework/react) (`aws-amplify-react` instead of `@aws-amplify/ui-react`), but going back the legacy doesn't sound as a good idea.

For my part, I decided to get rid of amplify UI (because of this and many other bugs I've been facing).
I built a custom one using [this tutorial](https://blog.logrocket.com/authentication-react-apps-aws-amplify-cognito/), you can have a look [here](https://github.com/claptime/claptime/pull/25/files) (React with hooks + AntDesign)

ccate:
https://stackoverflow.com/questions/16420898/chrome-safari-and-opera-do-not-remember-password-on-simple-form

slewis77:
[...]

PR #8181, opened 2021-04-26
feat(@aws-amplify/ui-components): partial shadow DOM removal + pw manager workaround

wlee221:
#### Description of changes

This PR **partially** enables password managers as a workaround while we write the new ui components to address this more fully (see https://github.com/aws-amplify/amplify-js/issues/5782#issuecomment-826968238 for details). This patch triggers browsers to save password and autofill the passwords through three steps:

1. Remove all shadow dom layer between form and each input.

This wouldn't be breaking as long as we / customers do not use any slots exposed at those levels. Fortunately, the only slot exposed is `input`: https://github.com/aws-amplify/amplify-js/blob/6754981d9e079d8837905b7027b4c25dbfd65224/packages/amplify-ui-components/src/components/amplify-form-field/amplify-form-field.tsx#L55
which does not make sense to override. Removing all shadow dom in our form elements provides other html benefits too: keyboard enter event automatically submits forms and clicking labels focuses the respective input elements.

2. Set `autocomplete` and `name` attribute across our components. 

This enables browsers to detect form and submit events for password managers to work. Note that FireFox password manager will work at this point. Chrome successfully triggers password save, but it does not autofill the inputs (chromium [issue](https://bugs.chromium.org/p/chromium/issues/detail?id=649162)). 

3. Implement a container with hidden input.

To workaround shadow DOM, we implement `amplify-auth-container` component that has hidden form inputs. These inputs will be on top of `amplify-authenticator` and is not shadowed -- and thus gets autofilled. Then our actual `amplify-input` inside the authenticator can attach event listeners to the hidden form and get the values from them. Customer code will look like this:

```tsx
<amplify-auth-container>
  <amplify-authenticator></amplify-authenticator>
</amplify-auth-container>
```

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/5782, https://github.com/aws-amplify/amplify-js/issues/6782, https://github.com/aws-amplify/amplify-js/issues/7837, 

Tested that this triggers chrome / firefox password manager save and autofill. Chrome for example:
Kapture 2021-04-20 at 16 46 09 (1)

But more investigation should be put into whether this has any regressions as we are in fact removing a lot of shadows from this PR. This will be my next step.

edit: I've done some manual validations:

- [x] works with `usernameAlias="email"`, even when sub components are slotted or have custom form fields.
- [x] ditto as above, but with `usernameAlias="phone_number"`
- [x] ditto as above, but with `usernameAlias='username'` 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -33,19 +35,55 @@ export class AmplifyInput {
...
+		if (!username || !password) return;
+
+		if (closestElement('amplify-sign-in', this.el)) {
+			if (
+				this.name === 'username' ||
+				this.name === 'email' ||
+				this.name === 'phone'
+			) {
+				username.addEventListener('input', e => {
+					this.setAutoCompleteValue(e);
```
thaddmt, line 79: Dang this a super interesting way to solve this

thaddmt (approved):

jamesaucode:
Just curious, this is only a work around for browser built-in password manager, but not other password manage plugin like Bitwarden right? Otherwise LGTM

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
wlee221: Yeah, this is a pretty hacky workaround :(

wlee221:
@jamesaucode, you are right. Third-party password managers have different support regarding shadow DOM: for example [1password](https://1password.community/discussion/79137/shadow-dom-polymer-forms-do-not-fill-companion-extension-v4-7-4-now-supports-shadow-doms) traverses through shadow DOM but some managers do not. But yes, this PR is targeted at browser password managers.

packages/amplify-ui-components/src/components/amplify-auth-container/amplify-auth-container.tsx
```diff
@@ -0,0 +1,22 @@
...
+})
+export class AmplifyAuthContainer {
+	render() {
+		return (
+			<Host>
+				<form autoComplete="on" hidden>
+					<input name="username"></input>
+					<input name="password" type="password"></input>
+					<input type="submit"></input>
+				</form>
```
wlee221, line 17: This is the hidden form that will pick up the credentials from autofill.

packages/amplify-ui-components/src/components/amplify-button/amplify-button.tsx
```diff
@@ -24,48 +22,22 @@ export class AmplifyButton {
 	/** Name of icon to be placed inside the button */
 	@Prop() icon?: IconNameType;
 
-	private handleClick = (ev: Event) => {
```
wlee221, line 27: One positive feedback from removing shadow: html automatically submits the form when we click the submit button!

packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx
```diff
@@ -219,7 +220,7 @@ export class AmplifyConfirmSignUp {
 			if (!this.userInput) throw new Error(Translations.EMPTY_USERNAME);
 			this.userInput = this.userInput.trim();
 			await Auth.resendSignUp(this.userInput);
-			this.handleAuthStateChange(AuthState.ConfirmSignUp);
+			this.handleAuthStateChange(AuthState.ConfirmSignUp, this.user);
```
wlee221, line 223: Unrelated to this pr, but this fixes a bug where clicking "resend code" will remove the username field value because `this.user` was not getting sent.

packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx
```diff
@@ -113,6 +113,7 @@ export class AmplifyConfirmSignUp {
 				),
 				value: this.userInput,
 				disabled: this.userInput ? true : false,
+				inputProps: { autocomplete: 'username' },
```
wlee221, line 116: Adding `autocomplete: 'username'` to the username (whether that be `username`, `phone_number`, or `email`).

packages/amplify-ui-components/src/components/amplify-form-field/__snapshots__/amplify-form-field.spec.ts.snap
```diff
@@ -2,125 +2,119 @@
...
+    <div class="form-field-label">
+      <amplify-label>
+        label
+      </amplify-label>
+    </div>
+    <div class="form-field-description" data-test="form-field-description" id="undefined-description">
+      description
+    </div>
+    <div>
+      <slot-fb name="input">
```
wlee221, line 16: `slot-fb` are Stencil's polyfilled version of `slot`.

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -33,19 +35,55 @@ export class AmplifyInput {
...
+		if (/Firefox/.test(navigator.userAgent)) return; // firefox autofill works
+		const container = closestElement('amplify-auth-container', this.el);
+		if (!container) return;
+
+		const username: HTMLInputElement = container.querySelector(
+			"input[name='username']"
+		);
+		const password: HTMLInputElement = container.querySelector(
+			"input[name='password']"
+		);
```
wlee221, line 68: These fetch the hidden `username` and `password` form from `amplify-auth-container`.

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -33,19 +35,55 @@ export class AmplifyInput {
...
+
+		if (!username || !password) return;
+
+		if (closestElement('amplify-sign-in', this.el)) {
```
wlee221, line 72: Make sure we only autofill the input if we're in sign-in.

packages/amplify-ui-components/src/components/amplify-input/amplify-input.scss
```diff
@@ -14,6 +14,10 @@
 	--margin: 0 0 0.625rem 0;
 }
 
+[data-autocompleted] {
+	background-color: rgb(232, 240, 254) !important;
+}
```
wlee221, line 19: I directly took this color from chrome. Maybe there's a more direct way to set this? Will look into it.

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -33,19 +35,55 @@ export class AmplifyInput {
...
-	})
-	formSubmit: EventEmitter;
+	}) formSubmit: EventEmitter;
+	@State() autoCompleted;
 
-	// eslint-disable-next-line
-	@Listen('keydown')
-	handleKeyDown(ev) {
-		if (ev.key === 'Enter') {
-			this.formSubmit.emit(ev);
```
wlee221, line 48: We no longer need keydown event because html does it for us if shadow DOM isn't present :)

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.scss
```diff
@@ -4,7 +4,7 @@
  * @prop --footer-color: Font color of the footer
  * @prop --footer-weight: Font weight of the footer
  */
-:host {
+amplify-sign-up {
```
wlee221, line 7: Without shadow-dom, you should use the component name itself instead of host (stencil [docs](https://stenciljs.com/docs/host-element))

wlee221 (commented):
Added some comments to pick out the actual code changes. I will do some manual regression testing, but the code itself is ready to be reviewed.

packages/amplify-ui-components/src/components/amplify-input/amplify-input.scss
ericclemmons: Ah, so this exists so that we can make our existing fields appear to be auto-completed like Chrome does!

ericclemmons (approved):
Thank you @wlee221 !

These comments help explain a lot. Honestly, nothing in this PR seems controversial, besides potential cosmetic issues.

However, those will be trivial to address compared to the benefit of this functionality.

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -32,20 +25,83 @@ export class AmplifyInput {
...
+			/**
+			 * When we're using slots, autofilled data will persist between different authState,
+			 * even though input events are not triggered. This ends up in parent components
+			 * not picking up input values. For now, we're emptying the input on authState change
+			 * which is the existing behavior.
+			 */
+			const input = this.el.querySelector('input');
+			input.value = '';
+			this.autoCompleted = false;
+		});
```
wlee221, line 78: This is the only significant code change since the last review. Comments should be self explanatory, but this accounts for slot's unexpected behavior where it stays in DOM even when it's not rendered and lifecycle methods do not get called as expected. The implementation is getting fairly hacky and it's not a good practice that `amplify-input` is listening to authState -- but this was the simplest solution I could find without modifying all auth subcomponents. (And the need for hacky solutions also demonstrate why we're moving away from web components, too.) This shouldn't affect chatbot or storage components because `onAuthUIStateChange` does not get called anyway, but let's see how the integration tests go.

A better long-term solution would be just to use uncontrolled form throughout the subcomponents, but this is a major rewrite effort which we are already doing with the new components. I think this is an acceptable complexity for a workaround for #5782.

wlee221 (commented):
Made some edge case handling based on some testing. There's only one significant code change that I commented. The overall approach is the same and I tested this pretty throughly with different `usernameAlias` , so I don't think another review is necessary. But it's always welcome!

packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
wlee221: Alleviated this a little bit by doing this only when it's inside an authenticator.

wlee221:
Did some final testing with [gist](https://gist.github.com/wlee221/9e3c6347aa5883d2a82c54714b68248e). I'm merging this in now. Let's keep an eye for regressions after this is shipped as this is a fairly big change.

PR #8194, opened 2021-04-28
fix(@aws-amplify/ui-components): Get autofill data on `componentDidLoad`

wlee221:
#### Description of changes

On Vue 3 specifically, autofill with `amplify-auth-conatainer` was failing because it was trying to get the reference to the `input` before it has materialized. This PR moves this logic to `componentDidLoad` so that `autofill` is attempted after the component is fully loaded. Also adds a basic null check.

#### Issue #, if available

follows up #8181 

Did a `yarn link` with a vue 3 authenticator sample and validated that the inputs get autofilled.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
@@ -36,6 +36,7 @@ export class AmplifyInput {
 	 */
 	private setAutoCompleteValue(value: string) {
 		const input = this.el.querySelector('input');
+		if (!input) return;
 		input.value = value;
 		// dispatch an input event from this element to the parent form
 		input.dispatchEvent(new Event('input'));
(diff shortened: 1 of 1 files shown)


packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -36,6 +36,7 @@ export class AmplifyInput {
 	 */
 	private setAutoCompleteValue(value: string) {
 		const input = this.el.querySelector('input');
+		if (!input) return;
```
ericclemmons, line 39: Ahh, of course this may not exist yet. :(

ericclemmons (approved):
Thanks for this fix!  It's a shame the DOM behavior is so tricky that it's necessary, but it makes sense!

Issue #8158, opened 2021-04-22
Previously deleted items are being returned as part of a query after going from offline -> online network state


sconway:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

DataStore

### Amplify Categories

storage

### Environment information

<details>

```
# Put output below this line


```

</details>

### Describe the bug

We are currently using Datastore to manage overrides to an itinerary displayed in our app. We model an Override as

```
type Override @model {
    id: ID!
    createdBy: String
    date: String
    dayOfDuration: Int
    description: String
    isCreatedItem: Boolean
    isTdOnly: Boolean
    prodComponentId: ID!
    prodTourId: Int
    startTime: String
    title: String
    uniqueComponentId: ID!
}
```

Overrides can be created and deleted, but there is never more than 1 override created for a corresponding itinerary item. 

Adding and removing Datastore items in our application works as expected while a device is online, however the following scenario is giving us unexpected results:

1. Device saves an override via Datastore
2. Device deletes the override that was created in step 1
3. Device goes offline
4. Device saves a new override via Datastore
5. Device goes online
6. Device queries datastore for overrides and both the created override from step 4, and the deleted override from step 2 are returned

Example:
```
[ 
    { 
        id: '5a53e76f-9f2c-4492-9ee2-facf58626f8b',
        createdBy: null,
        date: '',
        dayOfDuration: 1,
        description: null,
        isCreatedItem: null,
        isTdOnly: true,
        prodComponentId: '16965791',
        prodTourId: 221398,
        startTime: null,
        title: '@@TEST3 -- offline Overnight Security',
        uniqueComponentId: '16965791-1',
        _version: 1,
        _lastChangedAt: 1619116075244,
        _deleted: null 
    },
    { 
        id: '2e7efb8e-2627-405c-b335-0df486b876f0',
        createdBy: null,
        date: '',
        dayOfDuration: 1,
        description: null,
        isCreatedItem: null,
        isTdOnly: true,
[...]

PR #8196, opened 2021-04-28
fix(@aws-amplify/datastore): correctly processing Delta Sync query response in RN

iartemiev:
#### Description of changes

Fixes the way a Delta Sync query response is processed in React Native.

### Current behavior

The `batchSave` method in the AsyncStorageDatabase class separates incoming records from a delta query response into saves and deletes. It then sequentially performs a `multiRemove` of all the delete operations, followed by a `multiSet` of the save ops.

So in the event that we have a create/update operation as well as a delete op for a record with the same `id` within the timespan of the delta sync record TTL, the create/update will win out locally, regardless of its `_version` or `_lastChangedAt` value.

### Change introduced

The change in this PR adds a step that iterates through the records from the delta sync response in the order they're received and creates a `Map` with a key of `id` from the records. The map is then iterated over before splitting out the records into saves and deletes. This way, only the latest version of the record for a given model `id` ends up getting persisted to the local store.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/8158

* Manual validation in sample app
* Added unit test
* e2e test 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
N/A ~Relevant documentation is changed or added (and PR referenced)~

### DIFF
--- a/packages/datastore/__tests__/AsyncStorageDatabase.test.ts
+++ b/packages/datastore/__tests__/AsyncStorageDatabase.test.ts
@@ -1,0 +1,136 @@
+import AsyncStorageDatabase from '../src/storage/adapter/AsyncStorageDatabase';
+import AsyncStorageAdapter from '../src/storage/adapter/AsyncStorageAdapter';
+import {
+	DataStore as DataStoreType,
+	initSchema as initSchemaType,
+} from '../src/datastore/datastore';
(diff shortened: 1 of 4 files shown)


manueliglesias (approved):
LGTM 👍 🎉

Issue #3330, opened 2019-05-22
What is going on with .getMFAOptions()?


davidgatti:
**Describe the bug**

Calling  .getMFAOptions() using `amazon-cognito-identity-js` as per documentation will produce a `undefined` variable, no matter if MFA is enabled or not. Thus there is no way to find out if a user has MFA enabled or not.

**To Reproduce**

Run your code at least once before committing the code to the public.

**Expected behavior**

Get a variable with some data that can help me found out the state of the MFA of a user.

# This is unacceptable 

You have a service released for the public for years now, and you have functions that do not work! How can anyone use your product? This is crazy. What do you lack AWS: Money? People? Experience? 

- This issue was already mentioned in issue #360, and got a lame reply showing the AWS employ didn't spend 5 min reading the issue, and gave a general answer that is not related to the problem. And to top this off, another Amazon employees, closed the issue thinking it is resolved 😕. Unreal!
- Then a random stranger on the internet proposes the fix to this problem on Issue #2707, someone that is payed $0 by Amazon, and dose it on his free time, and get zero acknowledgment from AWS.

And since your SDK is able to find out if a user has MFA or not during the log in process, it is proof that you can get to this data, meaning you just have to copy and paste the code form one place to another.

And so. I demand a resolution to this problem in a timely manner. People on the internet should not fix bugs for free that you produce!

haverchuck:
Hi @davidgatti, thanks for raising the issue, and apologies for the time you (and others) have had to spend on this.  

I am not sure from your post whether you are using the amazon-cognito-identity-js SDK by itself or with the Amplify JS framework.  If you are using the Amplify JS framework and your users have the option of choosing their MFA type (which I believe is your case if I’m reading your issue correctly), you can try using[ Auth.getPreferredMFA](https://aws-amplify.github.io/docs/js/authentication#retrieving-current-preferred-mfa-type) to see if the user has already setup their MFA preference.  

If you are using amazon-cognito-identity-js only, then yes - the PR you cited seems to be required.  We have tested the PR and will be discussing it internally before merging.

Please let us know if you require anything further.  We will update this issue once the PR you’ve cited has been merged.

davidgatti:
@haverchuck I updated the original message, to make it more clear.

haverchuck:
@davidgatti  - Thank you.  We made some headway on this problem today and will continue to update this issue.

HoraceShmorace:
> Run your code at least once before committing the code to the public.

[...]

PR #2707, opened 2019-02-14
Fix getMFAOptions data extraction

sjawhar:
*Description of changes:*
`cognitoUser.getMFAOptions()` is currently broken. It always returns undefined, because `userData.MFAOptions` is not defined. Instead, there are two keys in `userData` containing MFA info: `UserMFASettingList` and `PreferredMfaSetting`. So we should return those instead.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #8197, opened 2021-04-28
chore(amazon-cognito-identity-js): deprecate getMFAOptions

svidgen:
#### Description of changes

PR #2707 from Feb. 2019 (two years ago) indicated that `getMFAOptions()` was not returning meaningful values. Currently, [`MFAOptions`](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_GetUser.html#CognitoUserPools-GetUser-response-MFAOptions) is no longer supported *at all*. Hence, the method needs to be deprecated.

The [corresponding method in `auth`](https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/Auth.ts#L652-L658) is *already* deprecated.

#### Issue #, if available

Two incidents where customers were expecting this method to behave meaningfully: #2707, #3330

No functional changes.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] ~~Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)~~ (no functional changes)

Notably, [tests do still exist for these methods](https://github.com/aws-amplify/amplify-js/search?q=getMFAOptions&type=code). Since these methods are simply marked as deprecated for now, I'm leaving those tests in on the off-chance any customers are actually calling these methods. (If anyone is depending on the broken behavior, let's not allow it to break *differently.*)

### DIFF
--- a/packages/amazon-cognito-identity-js/README.md
+++ b/packages/amazon-cognito-identity-js/README.md
@@ -724,16 +724,28 @@ cognitoUser.forgetDevice({
 
 ```
 
-**Use case 24.** Retrieve the MFA Options for the user in case MFA is optional.
+**Use case 24.** Retrieve the MFA settings for the user.
 
 ```javascript
-cognitoUser.getMFAOptions(function(err, mfaOptions) {
+cognitoUser.getUserData((err, data) => {
 	if (err) {
 		alert(err.message || JSON.stringify(err));
 		return;
 	}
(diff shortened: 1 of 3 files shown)


sammartinez (approved):
LGTM 🌮

PR #8227, opened 2021-05-05
correct the unmount lifecyle for amplify-authenticator

### LINKED ISSUE #6840: AmplifyAuthenticator memory leak
**Describe the bug**

I'm following the documented example for recommended usage (https://docs.amplify.aws/ui/auth/authenticator/q/framework/react#recommended-usage).

If I `console.log` inside `onAuthUIStateChange` then the logged message increases by one on every sign out, indicating a new function is created while the old function(s) are also maintained.

These duplicate functions are all called on signing in again, so state is updated x times, resulting in x calls to render the app.

**To Reproduce**

As above.

**Expected behavior**

`onAuthUIStateChange` to be called once per auth state change.

**Code Snippet**

I'm using the code snippet found in the docs here:
https://docs.amplify.aws/ui/auth/authenticator/q/framework/react#recommended-usage

hkjpotato:
#### Description of changes

Fix the memory leak issue by using the correct lifecycle method to remove event listener when component is unmounted from DOM.

It seems there is no such lifecycle hook called `componentWillUnload`. Based on the stencil doc https://stenciljs.com/docs/component-lifecycle#disconnectedcallback, it should be `disconnectedCallback`.

#### Issue #, if available

Fixes #6840

- run it locally to see the lifecycle is called and `handleExternalAuthEvent` is being removed from Hub, and thus fix the memory leak issue.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
+++ b/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx
@@ -206,7 +206,7 @@ export class AmplifyAuthenticator {
 		return <slot name={slotName}>{slotIsEmpty && authComponent}</slot>;
 	}
 
-	componentWillUnload() {
+	disconnectedCallback() {
 		Hub.remove(AUTH_CHANNEL, this.handleExternalAuthEvent);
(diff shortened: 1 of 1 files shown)


ericclemmons (approved):
Oh my gosh, I can't believe this was missed!  Thanks for the fix @hkjpotato !

PR #8233, opened 2021-05-06
Add new integ test for amazon-cognito-identity-js to CCI jobs

nickarocho:
The "old" `amazon-cognito-identity-js` integ test is now called `amazon-cognito-identity-js-cookie-storage` and the "new" `amazon-cognito-identity-js` is added (though it appears like it was already in the `config.yml`).

Passing CCI build - https://app.circleci.com/pipelines/github/aws-amplify/amplify-js?branch=amazon-cognito-identity-js-integ-cci

- [x] PR description included
- [x] `yarn test` passes


iartemiev (approved):
LGTM! 👍🏻

.circleci/config.yml
```diff
@@ -680,7 +696,7 @@ jobs:
     steps:
       - prepare_test_env
       - integ_test_node_js:
-          test_name: 'amazon-cognito-identity-js'
+          test_name: 'amazon-cognito-identity-js-cookie-storage'
```
amhinson, line 699: I think this one can stay, since this is the node test

amhinson (changes requested):

.circleci/config.yml
nickarocho: For sure, I thought I'd just change the name of the test since it functionally only tests `cookie-storage` stuff in node. I can change it back though, shouldn't have any effect either way.

iartemiev: Yep, I thought it would make sense to rename the `test_name` here to align it with what's being tested. We can update the sample dir in the samples repo as well and then rename `sample_name`

amhinson: There were two sample apps & tests named `amazon-cognito-identity-js`, one for `react` and another for `node`, but the renaming that we already did only changed the `react` app that focused on cookie storage. The node app has a different focus and still has this same name. 

With that said, we could change the sample name to something different to be more specific, but that might be more effort than its worth for now until we add another node sample.

iartemiev: Ah, got it. That's my mistake then. I had assumed it was testing the same thing, although now I feel dumb, because what would cookies be doing in a nope app 🤦🏻‍♂️

nickarocho: @amhinson fixed!

amhinson (approved):
LGTM 👍

Issue #7714, opened 2021-02-09
DataStore does not sync updated Models where keys are part of Composite keys


meducati:
**Describe the bug**
Saving an updated model works locally, but sync to backend fails when the updated keys are also part of a composite key.

**To Reproduce**
1. Create schema model with @key and fields: ["alpha", "beta"]
2. Update an existing DataStore item with at least one of the field keys (Datastore.save, modelname.copyof...)
3. DataStore will update as expected
4. Sync will throw a warning (see below), due to the auto-generated resolver. The reason however, is that  the "outboxMutationEnqueued" element is not including all updated keys in the "data" variable

**Expected behavior**
Update should work

**Code Snippet**
Schema (part of):
type CADFeature
@model
@key(name: "FeaturesByFile", fields: ["fileId","tool","status"], queryField: "getFeaturesByFileId")
@auth(rules: [
    {allow: owner},
    {allow: groups, groups: ["admin"] },
    {allow: groups, groups: ["users"], operations: [read] },
    {allow: private, provider: iam, operations: [read] }
  ])
{
id: ID!
status: FeatureStatus!
characteristic: String
notes: String
tool: FeatureTool!
...

Example template ("status" and "tool" are fields in the "FeaturesByFile" @key):
```javascript
await DataStore.save(
    CADFeature.copyOf(features[0], updated => {
        updated.status = FeatureStatus.DELETED;
        updated.tool = features[0].tool;
        updated.characteristic = 'whatever';
        updated.notes = 'whatever2';
    })
);
```
**Does NOT work: (EDIT: These are both keys that are part of a composite key)**
1. updated.status = FeatureStatus.DELETED only
2. updated.tool = features[0].tool only
3. updated.status AND updated.tool

**Does work: (EDIT: These are both keys that are NOT part of a composite key)**
1. updated.characteristic only
2. updated.notes only
3. updated.characteristic AND updated.status

**Screenshots**
This DataStore warning originates in the Resolver which is correct. (In my previous build, this check was not generated, so must have come from in a recent aws-amplify version)). _Note that I updated BOTH status and tool for this example, but only status is submitted in the sync operation._
![DatastoreWarning](https://user-images.githubusercontent.com/63409162/107383821-a7cfad00-6af1-11eb-9a9e-c73abb34e8d0.JPG)

As seen here, "tool" is never included in the pending mutation (but as mentioned before, tool has been updated in the local DataStore).
[...]

PR #8253, opened 2021-05-11
fix(@aws-amplify/datastore): updates with composite keys

iartemiev:
#### Description of changes

Include all fields that comprise a composite key in the update mutation request input if any of those fields was changed

For example:
```gql
type Post
  @model
  @key(name: "titleSort", fields: ["title", "created", "sort"]) {
  id: ID!
  title: String!
  description: String
  created: String
  sort: Int
}
```

Performing the following update:

```js
await DataStore.save(Post.copyOf(post, updated => {
  updated.title = 'Updated';
}));
```

Should include `title`, `created`, and `sort` in the GraphQL mutation input even though the latter 2 fields are unchanged.

#### Issue #, if available

#7714 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
~- [ ] Relevant documentation is changed or added (and PR referenced)~

### DIFF
--- a/packages/datastore/src/types.ts
+++ b/packages/datastore/src/types.ts
@@ -40,6 +40,9 @@ export type SchemaModel = {
 	attributes?: ModelAttributes;
 	fields: ModelFields;
 	syncable?: boolean;
+	compositeKeys?: {
+		[key: string]: string[];
+	};
 };
 export function isSchemaModel(obj: any): obj is SchemaModel {
 	return obj && (<SchemaModel>obj).pluralName !== undefined;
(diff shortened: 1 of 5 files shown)


packages/datastore/src/types.ts
```diff
@@ -78,6 +81,55 @@ export function isTargetNameAssociation(
...
+			string,
+			string,
+			string?,
+			string?,
+			string?,
+			string?,
+			string?,
+			string?,
+			string?
+		];
```
amhinson, line 106: Where does this specific type come from?

amhinson (approved):
LGTM 👍

packages/datastore/src/types.ts
iartemiev: This was my way of specifying that a composite key can have >= 3 fields. There may be a better way to do this in TS, definitely open to suggestions

iartemiev: I know how to limit array size to a specific number, e.g., 3 but not sure how to do 3 or more

iartemiev: I changed that part of the type slightly to `fields: [string, string, string, string?, string?]`.

Issue #3238, opened 2019-05-07
Cognito - User is not authorized to get auth details


raquila:
** Which Category is your question related to? **
SAML authentication

** What AWS Services are you utilizing? **
amazon-cognito-identity-js

** Provide additional details e.g. code snippets **
I'm getting an error when trying to authenticate using a login that is tied to a saml provider.  If I try to log in through Cognito UI login screen I don't have any issues.  I'm not sure I need to do something different since non saml provider users work fine using the code below.  I'm not able to find any detailed documentation on this.

![Capture1](https://user-images.githubusercontent.com/124973/57334759-a5574c80-70ee-11e9-8374-9650b90a4e3c.PNG)

powerful23:
@raquila if you want to login in with a saml provider, then you have to use the Cognito Hosted UI as this is the only way for now.

raquila:
I'm doing this within an electron application and trying to integrate an SSO solution.  There is no other way to do this without having to use the Hosted UI?

powerful23:
@raquila you can try to use Amplify and  the `Auth.federatedSignIn({provider: ${THE_PROVIDER_NAME}})` to directly open the login page of that SAML provider.  Please take a look at this doc: https://aws-amplify.github.io/docs/js/authentication#social-provider-federation

raquila:
@powerful23  Do you know if its possible to covert a set of users that were created in cognito already to use an identity provider?  When I test, it creates a new user in cognito with the same email address but different username.

PR #8295, opened 2021-05-18
Adding download progress tracker for Storage.get

### LINKED ISSUE #4734: Storage.get - Download progress tracking
** Which Category is your question related to? **
 Storage.
** What AWS Services are you utilizing? **
Cognito, Auth, Storage.

I am using aws amplify Auth and storage in web application. I am using Storage.get to download a file from S3 bucket. How to track progress of the download?

jamesaucode:
#### Description of changes
This adds a download tracker for `.get` similar to `.put`. Also changed `SEND_PROGRESS` to `SEND_UPLOAD_PROGRESS` to remove ambiguity.
Added a small test to verify the cancelToken from the last PR's `.cancel` API gets properly added to the request

Doc change PR in [#3238](https://github.com/aws-amplify/docs/pull/3238)

#### Issue #, if available
fixes #4734 

Added relevant unit tests, manually tested.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are added (https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)


packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -160,6 +166,18 @@ export class AWSS3Provider implements StorageProvider {
...
+					if (progressCallback) {
+						if (typeof progressCallback === 'function') {
+							progressCallback(progress);
+						} else {
+							logger.warn(
+								'progressCallback should be a function, not a ' +
+									typeof progressCallback
+							);
+						}
+					}
```
ericclemmons, line 179: This will `warn` _every time progress updates_.  I think what you're looking for is a `typeof progressCallback` check _prior_ to this, so you can safely call `emitter.on(SEND_DOWNLOAD_PROGRESS_EVENT, progressCallback`) only.

jamesaucode: good catch, I updated that part of the code

svidgen (approved):
Pulled and tested (manually). Seems to work as advertised!

Is there an associated integ test PR?

jamesaucode:
> Pulled and tested (manually). Seems to work as advertised!
> 
> Is there an associated integ test PR?

Thanks! Not yet, will add it and reference it here!

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -265,6 +271,18 @@ export class AWSS3Provider implements StorageProvider {
 		if (download === true) {
 			const getObjectCommand = new GetObjectCommand(params);
 			try {
+				if (progressCallback) {
```
elorzafe, line 274: This is axios downloader not the browser downloader right?, where is this stored I don't remember how this worked?

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -265,6 +271,18 @@ export class AWSS3Provider implements StorageProvider {
...
+					if (typeof progressCallback === 'function') {
+						emitter.on(SEND_DOWNLOAD_PROGRESS_EVENT, progress => {
```
elorzafe, line 276: Do you need to remove the event listener?

elorzafe (commented):
This is looking good @jamesaucode !

I have a couple of questions if you can review those

Thanks again!

packages/storage/src/providers/AWSS3Provider.ts
jamesaucode: Yes, this is for the use case of when user calls `Storage.get('key', { download:true })` where the user wants the `Blob` directly.

`progressCallback` will be called whenever the `onDownloadProgress` event fires from `axios` (which seems to be the progress event from `XMLHttpRequest` https://github.com/axios/axios/blob/master/lib/adapters/xhr.js#L159)

PR #8296, opened 2021-05-18
feat(@aws-amplify/datastore): include custom pk in update/delete mutations

iartemiev:
#### Description of changes

Expands on this PR: https://github.com/aws-amplify/amplify-js/pull/8253

If the model has a custom primary key defined, always include it in the update and delete mutation input.

#### Issue #, if available

N/A

* Added and ran unit tests 
* Validated against sample apps

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/helpers.ts
+++ b/packages/datastore/__tests__/helpers.ts
@@ -48,12 +48,19 @@ export declare class Profile {
 	public readonly lastName: string;
 }
 
-export declare class PostComposite {
+export declare class PostKeys {
 	public readonly id: string;
 	public readonly title: string;
 	public readonly description: string;
 	public readonly created: string;
 	public readonly sort: number;
+}
+
+export declare class PostCustomPK {
+	public readonly id: string;
+	public readonly postId: number;
+	public readonly title: string;
+	public readonly description?: string;
 }
 
 export function testSchema(): Schema {
@@ -306,8 +313,8 @@ export function testSchema(): Schema {
 					},
 				],
 			},
-			PostComposite: {
-				name: 'PostComposite',
+			PostKeys: {
+				name: 'PostKeys',
 				fields: {
 					id: {
 						name: 'id',
@@ -346,7 +353,7 @@ export function testSchema(): Schema {
 					},
 				},
 				syncable: true,
-				pluralName: 'PostComposites',
+				pluralName: 'PostKeys',
 				attributes: [
 					{
 						type: 'model',
@@ -355,7 +362,7 @@ export function testSchema(): Schema {
 					{
 						type: 'key',
 						properties: {
-							name: 'titleSort',
+							name: 'titleCreatedSort',
 							fields: ['title', 'created', 'sort'],
 						},
 					},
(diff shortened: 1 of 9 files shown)


packages/datastore/__tests__/storage.test.ts
```diff
@@ -490,45 +491,143 @@ describe('Storage tests', () => {
 					})
 				);
 
+				// `sort` is part of the key's composite sort key.
+				// `created` should also be included in the mutation input
```
nickarocho, line 495: Thank you for leaving helpful comments! Could definitely use more of this throughout.

packages/datastore/src/datastore/datastore.ts
```diff
@@ -157,10 +157,13 @@ const initSchema = (userSchema: Schema) => {
 	};
 
 	Object.keys(schema.namespaces).forEach(namespace => {
-		schema.namespaces[namespace].relationships = establishRelation(
+		const [relations, keys] = establishRelationAndKeys(
 			schema.namespaces[namespace]
 		);
 
+		schema.namespaces[namespace].relationships = relations;
+		schema.namespaces[namespace].keys = keys;
```
nickarocho, line 165: Can we be positive `schema.namespaces[namespace]` will never be `falsy`?

packages/datastore/src/sync/processors/mutation.ts
```diff
@@ -425,9 +426,19 @@ class MutationProcessor {
 
 		const { _version, ...parsedData } = <ModelInstanceMetadata>JSON.parse(data);
 
+		// include all the fields that comprise a custom PK if one is specified
+		const deleteInput = {};
+		if (primaryKey) {
```
nickarocho, line 431: Do we also need to check for `&& primaryKey.length` like you are checking for in `storage.ts`?

nickarocho (approved):
LGTM! Just had a couple questions (mostly out of curiosity). Nice stuff @iartemiev!

packages/datastore/__tests__/helpers.ts
```diff
@@ -355,15 +376,156 @@ export function testSchema(): Schema {
...
+				pluralName: 'PostCustomPKComposites',
+				attributes: [
+					{
+						type: 'model',
+						properties: {},
+					},
 					{
 						type: 'key',
 						properties: {
-							name: 'descSort',
```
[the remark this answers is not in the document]
iartemiev: In general, omitting the `name` property causes `@key` to create a primary index (something that you can only specify when creating the model for the first time). See the argument table here for more info: https://docs.amplify.aws/cli/graphql-transformer/key#key

I don't think I'm actually removing it per se, it was probably just a weird diff from my adding a few models to that testSchema

packages/datastore/src/datastore/datastore.ts
iartemiev: Yes, we can be certain that those exist, since this code is inside a `forEach` that's iterating over those very namespaces. 

For context: https://github.com/aws-amplify/amplify-js/blob/6ea11c0e1f642e3e955b531ec669304672c70440/packages/datastore/src/datastore/datastore.ts#L159-L165

manueliglesias (approved):
LGTM 🎉 
Thanks for the additional tests!

packages/datastore/__tests__/helpers.ts
```diff
@@ -355,15 +376,156 @@ export function testSchema(): Schema {
...
+							fields: ['id', 'postId'],
+						},
+					},
+				],
+			},
+			PostCustomPKComposite: {
+				name: 'PostCustomPKComposite',
+				fields: {
+					id: {
+						name: 'id',
```
vgribok, line 483: What's the benefit of having a composite primary key when partition key is a unique GUID? Is this PR just to fix mutations when people (seemingly pointlessly) use composite PKs with a GUID as a partition key? Or will this make non-unique partition key allowed in the PK?

Issue #7565, opened 2021-01-17
DataStore not storing updated model when setting optional attribute as 'undefined'


mgm1313:
**Describe the bug**
When trying to update a DataStore instance, I'm unable to set a field as `undefined`. 

**To Reproduce**
```graphql
type S3Object {
  bucket: String!
  region: String!
  key: String!
}

type Post @model {
  id: ID!
  title: String!
  featuredImage: S3Object
}
```

First create a new Post, populated with a featuredImage object:
```js
DataStore.save(new Post({
    title: 'My first post', 
    featuredImage: {
      bucket: 'myS3bucket',
      region: 'myS3region',
      key: 'MyS3ObjectKey',
    }
  }))
```

Then we update this created instance by it's id, and try setting `featuredImage` as `undefined` again:
```js
  const original = await DataStore.query(Post, id)

  DataStore.save(Post.copyOf(original, updated => {
    updated.featuredImage = undefined
  }))
```

**Expected behavior**
I would expect that after the update `featuredImage` would be `undefined` again both locally and in the cloud. 
Locally it works, I checked this by immediately logging the updated instance through:
```js
  DataStore.save(Post.copyOf(original, updated => {
    updated.featuredImage = undefined
  })).then(updated => console.log(updated))
```
But it doesn't upstream to the cloud, so the moment the local store is synced again the `featuredImage` field gets overwritten again with the old value. 

I tried setting the field as `null` instead of `undefined`, and this does work! But is giving me a TypeScript error since it doesn't expect `featuredImage` to be null: `Type 'null' is not assignable to type 'DeepWritable<S3Object> | undefined'.`. 

**Additional context**
Updating using `undefined`
```js
[...]

PR #8301, opened 2021-05-18
fix(@aws-amplify/datastore): coerce undefined field values to null

nickarocho:
#### Description of changes

A customer pointed out that if we define the value of a model's field as `undefined`, the mutation doesn't persist. However, if passed as `null`, it would work fine. So this one-liner just checks to see if a pre-existing field's passed value is `undefined` and coerces that value to `null`.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/7565

I tested the changes linked to a local sample app. Screenshots attached.

*Example* (with requested change) of explicitly setting the value of an existing field to `undefined`:

Screen Shot 2021-05-18 at 1 08 13 PM

*Example* (with requested change) of explicitly setting the value of a NEW field - (doesn't coerce to `null` because the key doesn't exist on the model yet):

Screen Shot 2021-05-18 at 1 12 35 PM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/src/storage/storage.ts
+++ b/packages/datastore/src/storage/storage.ts
@@ -332,6 +332,7 @@ class StorageClass implements StorageFacade {
 
 			// check field values by value. Ignore unchanged fields
 			if (!valuesEqual(source[key], originalElement[key])) {
+				// if the field was updated to 'undefined', replace with 'null' for compatibility with JSON and GraphQL
 				updatedElement[key] =
 					originalElement[key] === undefined ? null : originalElement[key];
 				if (key in compositeKeys) {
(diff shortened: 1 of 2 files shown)


amhinson (commented):
Looks good! Maybe we could also add a test in `storage.test.ts`?

packages/datastore/src/storage/storage.ts
```diff
@@ -332,8 +332,8 @@ class StorageClass implements StorageFacade {
 
 			// check field values by value. Ignore unchanged fields
 			if (!valuesEqual(source[key], originalElement[key])) {
-				updatedElement[key] = originalElement[key];
-
+				updatedElement[key] =
+					originalElement[key] === undefined ? null : originalElement[key];
```
iartemiev, line 336: Looks great! Let's just leave a comment above 335. Something to the effect of 
`// if the field was updated to `undefined`, replace with `null` for compatibility with JSON and GraphQL`

iartemiev (approved):
Thank you, Nick! LGTM, just one minor suggestion 👍🏻

amhinson (approved):

PR #8335, opened 2021-05-24
fix(@aws-amplify/datastore): fix default error/conflict handler

iartemiev:
#### Description of changes

The default error and conflict handler in DataStore were getting overwritten and set to `undefined`. This was causing console errors when mutation errors and unhandled mutation conflicts occurred (unless the handlers were explicitly set via `DataStore.configure`): 
> `this.errorHandler` is not a function

This PR fixes the default behavior, preventing the default handlers from being overwritten

Also: allow the `errorHandler` to be async (like `conflictHandler`) so that customers can perform custom async handling logic if needed before the MutationProcessor attempts to process the next record in the queue.

#### Issue #, if available

N/A

Tested in a sample app. No longer getting the `this.errorHandler` is not a function error when a custom error handler is not explicitly set via `DataStore.configure`. Same goes for the conflict handler.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/src/datastore/datastore.ts
+++ b/packages/datastore/src/datastore/datastore.ts
@@ -833,7 +833,7 @@ class DataStore {
 		const conflictHandlerIsDefault: () => boolean = () =>
 			this.conflictHandler === defaultConflictHandler;
 
-		if (configDataStore) {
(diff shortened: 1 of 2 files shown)


amhinson (approved):
LGTM 👍

PR #8336, opened 2021-05-24
Fix React Native Storage.put out of memory error & UI freezes

### LINKED ISSUE #6419: Storage.put fail with large files on Android with react-native
Storage.put need a blob to upload file.

But react-native has a fetch bug on android with "big" files. You can see [here](https://github.com/react-native-community/discussions-and-proposals/issues/109)

When you try to uplaod video with size of 160mb or more, `const response = await fetch(videoUri)` get next error: `TypeError: Network request failed`(It isnt network error, with small sizes it work).

I have tried to generate the blob with https://github.com/joltup/rn-fetch-blob but I have not been able to make it works.

Is there a way to get it working or an alternative way to upload large files?

### LINKED ISSUE #6888: React native Ui block when uploading
**Describe the bug**
Upload to using 

**To Reproduce**

1. npm install aws-amplify --save
2. npm install aws-amplify-react-native --save
3. import { Storage } from "aws-amplify";
4.Storage.put(key, blob, {
    contentType: blob.type,
    metadata: { payload: JSON.stringify(meta) },
  });

**Expected behavior**
Using  "aws-amplify": "3.3.1",

Upload file in background thread with out making block in UI

**Code Snippet**
.Storage.put(key, blob, {
    contentType: blob.type,
    metadata: { payload: JSON.stringify(meta) },
  });

**What is Configured?**
amplifyConfig = {
  Auth: {
    identityPoolId: XX-XXXX,
    region: AWS.REGION
  },
  Storage: {
    bucket: AWS.S3_BUCKET,
    region: AWS.REGION
  }
}

- Device: [e.g. iPhone11]

jamesaucode:
#### Description of changes
This fixes the memory usage issue for React Native.

The issue was with `streamCollector`. It will load everything from the file into memory before uploading, which will cause an out of memory exception if the file is large enough, slows down upload a lot and it also blocks the UI in the process.

It stems from the problem where Axios identifies the type of data using `Object.prototype.toString`, see: [utils#L147](https://github.com/axios/axios/blob/3958e9fbc8ebc0f72a8ce568cbf86f391d6869f3/lib/utils.js#L147) and  On React Native, calling `Object.prototype.toString` on Blob returns `'[object Object]'` instead of `'[object Blob]'`, it will then call a `JSON.stringify` on it instead of handling a Blob correctly. So we need to use a custom request transformer for React Native to correctly identify a Blob.

see: [defaultTransformer on Axios](https://github.com/axios/axios/blob/3958e9fbc8ebc0f72a8ce568cbf86f391d6869f3/lib/defaults.js#L39)

`createParts` is modified to partition the entire file instead of partition it on the fly due to a GC issue. In React Native Android, when slice of the blob gets GCed, it GCs the entire blob as well, so we need to partition the entire file before actually uploading.

Also removed some `any`s as I go through the code.

I did a simple benchmark with the original implementation versus this new implementation:
Upload speed is 92Mbps

| size | old | new |
| -- | -- | -- |
| 10mb | 43s | 1.6s |
| 20mb | 76s | 2.4s |
| 50mb | 212s | 5.4s |
| 75mb | ERROR | 8.1s |

Did not test anything beyond 75mb as the old implementation errors out on anything 75MB+

**TLDR: there should be a significant improvement in upload speed & UI won't be blocking for React Native users.**

Demo (20MB file):

https://user-images.githubusercontent.com/40404256/119397524-18f33480-bc8b-11eb-91ca-bfa14bebda98.mov

**(The demo downloads then upload, but it only calculates the time elapsed for the upload)**

#### Issue #, if available
fixes #6419
fixes #6888

Added unit tests
Manually tested on React Native android & ios

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


novakzaballa:
@jamesaucode and @amhinson, do you have an ETA for the final merge of this PR? I am currently blocked on a project because of this issue and would prefer to use this final solution instead of using one of the current workarounds.
Thanks in advance!

packages/storage/src/providers/axios-http-handler.ts
```diff
@@ -97,6 +133,9 @@ export class AxiosHttpHandler implements HttpHandler {
 
 		// From gamma release, aws-sdk now expects all response type to be of blob or streams
 		axiosRequest.responseType = 'blob';
+		if (Platform.isReactNative) {
```
amhinson, line 136: Could you add some comments here as to why this is necessary on React Native?

amhinson (approved):
This looks great @jamesaucode! Just one small note about adding some context around why this change is needed, but otherwise LGTM 👍

Thanks!

packages/storage/src/providers/axios-http-handler.ts
jamesaucode: Good suggestion, added

jamesaucode:
> @jamesaucode and @amhinson, do you have an ETA for the final merge of this PR? I am currently blocked on a project because of this issue and would prefer to use this final solution instead of using one of the current workarounds.
> Thanks in advance!

Hey @novakzaballa, the changes are now merged it should be available in our "unstable" branch now

novakzaballa:
@jamesaucode , After some testing using 4.0.2-unstable.21, I still have out of memory issues uploading large files. Files about 30MB are uploaded larger about 75MB cause app to suddenly close due to out of memory issues. How should I report and document the issue?

Issue #7036, opened 2020-10-22
DataStore max subscriptions reached error


nubpro:
**Describe the bug**
I found an easy and reliable method of reproducing "MaxSubscriptionsReachedError" error which is coming from AppSync. This is a React Native project.

**To Reproduce**
This is the most explicit ways I could figured out to reproduce the bug.

Updated repro methods:
https://github.com/aws-amplify/amplify-js/issues/7036#issuecomment-746942585

~~1. Setup Auth and DataStore on your app.~~
~~2. Login and wait for syncing process is to finish. (I'd use `Hub.listen('datastore')` to wait for `syncQueriesReady` event to trigger)~~
~~3. On your physical device, off the WiFi.~~
~~4. Logout.~~
~~5. Turn WiFi on.~~
~~6. Repeat steps 2 to 6. Depends on your schema, you may need to do this up to 10 times until it hits the max subscription error. In my case, I would only need to repeat 5 times as my schema contains 7 models with 21 subscriptions.~~

**Expected behavior**
Max subscription reached error shouldn't happen. When this happens, the app is unable to receive any incoming changes from the server anymore.

**Code Snippet**
I don't have any project samples that I can share at the moment.

**Screenshots**
Here's a screenshot of what I get.
![image](https://user-images.githubusercontent.com/762914/96849524-49ba8380-1488-11eb-984d-5a4d3a466fed.png)

<details>
  <summary><strong>Environment</strong></summary>

```
System:
    OS: macOS 10.15.5
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Memory: 53.96 MB / 8.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.3.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.8 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 86.0.4240.80
    Safari: 13.1.1
  npmGlobalPackages:
    @aws-amplify/cli: 4.29.5
    ios-deploy: 1.10.0
    npm: 6.14.8
    react-native-cli: 2.0.1
```

</details>

**Smartphone (please complete the following information):**

- Device: [e.g. iPhone6] iPad
- OS: [e.g. iOS8.1] iOS 13
- React Native

nubpro:
[...]

PR #8344, opened 2021-05-26
fix(@aws-amplify/datastore):  correct reachability unsubscribe behavior

iartemiev:
### Current Behavior

Currently, unsubscribing the Reachability observer in the DataStoreConnectivity class does not stop it from continuing to send messages that are consumed by the SyncEngine. This is due to a `setTimeout` that does not get canceled upon calling `DataStoreConnectivity.unsubcribe()`. (See https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/src/sync/datastoreConnectivity.ts#L46-L61 for context). 

This results in the `setTimeout` callback continuing to send messages to the SyncEngine, causing it to attempt to restart at inappropriate times. Additional context is in the linked issue, but here's a summary:

1. Network connectivity is lost
2. `DataStore.stop` is called as a result of some user action shortly thereafter (within 5 seconds of network connectivity loss)
3. SyncEngine unsubscribes all subscriptions, stops the sync process, disposes of WebSockets, etc.
4. `setTimeout` callback gets invoked, which sends a message to the SyncEngine 
5. SyncEngine attempts to start back up, even though `DataStore.stop` had been sent and there's still no network connectivity

One of the undesirable side effects is that this will cause duplicate AppSync subscriptions to be created after network connectivity is restored and `DataStore.start` is called via the app code.

### PR Changes

Cancels the `setTimeout` whenever `DataStoreConnectivity.unsubcribe()` is executed, preventing it from sending additional messages.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/7036

* Manually validated this in a sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/src/sync/datastoreConnectivity.ts
+++ b/packages/datastore/src/sync/datastoreConnectivity.ts
@@ -15,6 +15,7 @@ export default class DataStoreConnectivity {
 	private connectionStatus: ConnectionStatus;
 	private observer: ZenObservable.SubscriptionObserver<ConnectionStatus>;
 	private subscription: ZenObservable.Subscription;
+	private timeout: ReturnType<typeof setTimeout>;
 	constructor() {
 		this.connectionStatus = {
 			online: false,
@@ -38,6 +39,7 @@ export default class DataStoreConnectivity {
 			});
 
 			return () => {
+				clearTimeout(this.timeout);
 				this.unsubscribe();
 			};
 		});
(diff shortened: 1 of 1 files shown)


packages/datastore/src/sync/datastoreConnectivity.ts
```diff
@@ -15,6 +15,7 @@ export default class DataStoreConnectivity {
 	private connectionStatus: ConnectionStatus;
 	private observer: ZenObservable.SubscriptionObserver<ConnectionStatus>;
 	private subscription: ZenObservable.Subscription;
+	private timeout: ReturnType<typeof setTimeout>;
```
manueliglesias, line 18: Nice! I like the typing

manueliglesias (approved):
LGTM 🎉

Issue #1355, opened 2018-07-31
Remembering a device


saaniaki:
Hi everyone,

I tried to follow this [link](https://aws.amazon.com/blogs/mobile/tracking-and-remembering-devices-using-amazon-cognito-your-user-pools/) and remember user's devices at the login time, however, I always get this error:
` 1 validation error detected: Value null at 'deviceKey' failed to satisfy constraint: Member must not be null `
This is my code:
```
cognitoUser.setDeviceStatusRemembered({
	    onSuccess: function (result) {
	        console.log('call result: ' + result);
	    },

	    onFailure: function(err) {
	        alert(err);
	    }
	}); 
```
Is this a bug or I'm missing something?

jordanranz:
Hi @saaniaki, thanks for submitting feedback.

Seems that user does not have a deviceKey associated with their user object.

Have you tried calling getCachedDeviceKeyAndPassword to populate the deviceKey for the user?
https://github.com/aws-amplify/amplify-js/blob/master/packages/amazon-cognito-identity-js/src/CognitoUser.js#L1255

Here is a related Stack Overflow question:
https://stackoverflow.com/questions/44792164/aws-cognito-identity-js-forget-remember-do-not-remember-device]

saaniaki:
I'm using Angular, the problem was that the type mapping file doesn't have this method in it, so I couldn't call it. Anyway, I forced it to call and ignored the typescript error by adding `[x: string]: any;` to the index.d.ts file for CognitoUser class (not a good idea). Now it's working, thank you!

Hope they add it in the next update!

koladilip:
Device remembering only works in Mobile devices? or can be used for desktops? I tried using sample backend NodeJS code to sign in using Amplify it didn't remember the user.

jordanranz:
@saaniaki, this sounds like a feature request to add typescript support to amazon-cognito-identity-js yea?

@koladilip could you post the sample code?

saaniaki:
@jordanranz I would say a request to complete the typescript support since they already have it, or to say, complete the typings of https://github.com/aws-amplify/amplify-js/blob/master/packages/amazon-cognito-identity-js/index.d.ts

JustFly1984:
I got this error today, for my own personal account in my WebApp in production. We do not have mobile app, there is nothing to remember.

JustFly1984:
I could swear I had login to my app as super admin successfuly, logged out, and at attempt to login again got this error message

[...]

PR #8353, opened 2021-05-28
Device tracking

evcodes:
Implementing device tracking for JavaScript to achieve parity across platforms. Native iOS and Android have had these features for some time and we aim to provide this to our customers.

#### Issue #, if available
#1355 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/8250/workflows/11965c4a-ea47-4321-83e6-e9dd44a2f32e


jamesaucode:
Seems like it's not build because of a tslint rule, should we just turn off max line length for comments? 
```diff
diff --git a/packages/auth/tslint.json b/packages/auth/tslint.json
index 1bb9e144d..b1f7fcaed 100644
--- a/packages/auth/tslint.json
+++ b/packages/auth/tslint.json
@@ -5,7 +5,10 @@
        "jsRules": {},
        "rules": {
                "prefer-const": true,
-               "max-line-length": [true, 120],
+               "max-line-length": [true, {
+                       "limit": 120,
+                       "ignore-pattern": "// "
+               }],
                "no-empty-interface": true,
                "no-var-keyword": true,
                "object-literal-shorthand": true,
```
Or we can probably just disable the max length for the url at line 89

packages/auth/src/Auth.ts
```diff
@@ -85,6 +85,10 @@ const dispatchAuthEvent = (event: string, data: any, message: string) => {
 	Hub.dispatch('auth', { event, data, message }, 'Auth', AMPLIFY_SYMBOL);
 };
 
+// as per Cognito's Documentation 
+// https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListDevices.html#API_ListDevices_RequestSyntax
+const maxDevices = 60;
```
sammartinez, line 90: Can we make this 
```suggestion
const MAX_DEVICES = 60;
```

evcodes: done 👍🏼

packages/auth/src/Auth.ts
```diff
@@ -85,6 +85,10 @@ const dispatchAuthEvent = (event: string, data: any, message: string) => {
 	Hub.dispatch('auth', { event, data, message }, 'Auth', AMPLIFY_SYMBOL);
 };
 
+// as per Cognito's Documentation
+// https://tinyurl.com/c2a9f4hz
```
elorzafe, line 89: I would use the official link instead
```suggestion
// https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListDevices.html#API_ListDevices_RequestSyntax
```

packages/auth/src/Auth.ts
```diff
@@ -2194,6 +2201,39 @@ export class AuthClass {
 		const type = this.noUserPoolErrorHandler(this._config);
 		return Promise.reject(new NoUserPoolError(type));
 	}
+
+	public async rememberDevice(): Promise<any> {
+		const currUser = await this.currentUserPoolUser();
+		currUser.getCachedDeviceKeyAndPassword();
+
+		return new Promise((res, rej) => {
+			currUser.setDeviceStatusRemembered({ onSuccess: res, onFailure: rej });
```
elorzafe, line 2210: I would check the error messages, and what information should be returned to the customer

packages/auth/src/Auth.ts
```diff
@@ -2194,6 +2201,39 @@ export class AuthClass {
 		const type = this.noUserPoolErrorHandler(this._config);
 		return Promise.reject(new NoUserPoolError(type));
 	}
+
+	public async rememberDevice(): Promise<any> {
```
elorzafe, line 2205: Was an API review for the naming? I don't like the idea of return `any`

There are related issues for example if the customer is using a Storage mechanism that is not persisted, e.g. Memory storage in that case I think it should be fair to throw an error because it wouldn't work in practice

packages/auth/src/Auth.ts
```diff
@@ -2194,6 +2201,39 @@ export class AuthClass {
...
+		return new Promise((res, rej) => {
+			currUser.setDeviceStatusRemembered({ onSuccess: res, onFailure: rej });
+		});
+	}
+
+	public async forgetDevice(): Promise<any> {
+		const currUser = await this.currentUserPoolUser();
+		currUser.getCachedDeviceKeyAndPassword();
+		return new Promise((res, rej) => {
+			currUser.forgetDevice({ onSuccess: res, onFailure: rej });
```
elorzafe, line 2218: I would check the error message and what do we want to return to the customers

packages/auth/src/Auth.ts
```diff
@@ -2194,6 +2201,39 @@ export class AuthClass {
...
+	public async rememberDevice(): Promise<any> {
+		const currUser = await this.currentUserPoolUser();
+		currUser.getCachedDeviceKeyAndPassword();
+
+		return new Promise((res, rej) => {
+			currUser.setDeviceStatusRemembered({ onSuccess: res, onFailure: rej });
+		});
+	}
+
+	public async forgetDevice(): Promise<any> {
```
elorzafe, line 2214: I am not sure about naming, IMO is related to Cognito more than the real action, which I think is not asking for MFA code right.

packages/auth/src/Auth.ts
```diff
@@ -2194,6 +2201,39 @@ export class AuthClass {
...
+		const currUser = await this.currentUserPoolUser();
+		currUser.getCachedDeviceKeyAndPassword();
+		return new Promise((res, rej) => {
+			currUser.forgetDevice({ onSuccess: res, onFailure: rej });
+		});
+	}
+
+	public async fetchDevices(): Promise<any> {
```
elorzafe, line 2222: We should define a type that has useful information (not just Cognito returned data)

packages/auth/src/Auth.ts
```diff
@@ -2194,6 +2201,39 @@ export class AuthClass {
...
+		const currUser = await this.currentUserPoolUser();
+		currUser.getCachedDeviceKeyAndPassword();
+		return new Promise((res, rej) => {
+			const cb = {
+				onSuccess(data) {
+					res(data);
+				},
+				onFailure(err) {
+					rej(err);
```
elorzafe, line 2231: Error message should be serialized to something useful for customer not just Cognito messaging.

elorzafe (changes requested):
Thanks @evcodes ! I think this will be a great contribution.

However I have a few concerns related to API (naming and return values) and also you need to add tests for each method as well.

We can chat more offline.

Thanks!

packages/auth/src/Auth.ts
evcodes: based on the tslinting rules, the link was too long, 120 characters

hkjpotato: could consider disable the linting by adding a comment
```
// tslint:disable-next-line:max-line-length
```

packages/auth/src/Auth.ts
hkjpotato: It seems a common pattern that we just [throw the raw error](https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/Auth.ts#L696) from Cognito upwards, especially the error is from a direct api call.

We do throw [specific error message](https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/Auth.ts#L702) if we know better about what the error is.

While we should make our errors more explicit, refactoring the pattern could be in another commit to reduce the scope.

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -224,6 +226,19 @@ jest.mock('amazon-cognito-identity-js/lib/CognitoUser', () => {
...
+		return 'success';
+	};
+	CognitoUser.prototype.setDeviceStatusRemembered = callback => {
+		callback.onSuccess('success');
+	};
+	CognitoUser.prototype.forgetDevice = callback => {
+		callback.onSuccess('success');
+	};
+	CognitoUser.prototype.listDevices = (limit, paginationToken, callback) => {
+		callback.onSuccess('success');
```
elorzafe, line 239: I think this is the issue with `map` you were seeing before

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -3657,4 +3672,113 @@ describe('auth unit test', () => {
...
+				.mockImplementationOnce(
+					(
+						MAX_DEVICES,
+						none: string,
+						obj: {
+							onSuccess: (success: Object) => void;
+							onFailure: (err: any) => void;
+						}
+					) => {
+						obj.onSuccess({ Devices: [] });
```
elorzafe, line 3771: I would add test that return some data, so can it can see if that is returned successfully on `Auth.fetchDevices`

packages/auth/src/Auth.ts
```diff
@@ -2202,33 +2204,91 @@ export class AuthClass {
...
+			logger.debug('The user is not authenticated by the error', error);
+			throw new Error('The user is not authenticated');
+		}
+
 		currUser.getCachedDeviceKeyAndPassword();
 		return new Promise((res, rej) => {
 			const cb = {
 				onSuccess(data) {
-					res(data);
+					const deviceList: AuthDevice[] = data.Devices.forEach(device => {
```
elorzafe, line 2279: this should be `map` instead of `forEach`, because `forEach` doesnt return anything

elorzafe (changes requested):
Few small changes, overall looks good!

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -3780,5 +3781,46 @@ describe('auth unit test', () => {
...
+					) => {
+						obj.onSuccess({
+							Devices: mockDeviceArray,
+						});
+					}
+				);
+
+			await auth.fetchDevices().then(res => {
+				expect(spyOnCognito).toBeCalled();
+			});
```
elorzafe, line 3819: you need to validate here the response from `auth.fetchDevices()` if it matches

packages/auth/src/types/Auth.ts
```diff
@@ -220,3 +222,39 @@ export type ClientMetaData =
...
 	return !!(obj as UsernamePasswordOpts).username;
 }
+
+export interface IAuthDevice {
+	deviceCreatedDate: string;
+	deviceKey: string;
+	deviceLastAuthenticatedDate: string;
+	deviceLastModifiedDate: string;
+}
+
```
elorzafe, line 232: This should be equivalent to iOS client to keep parity with them https://aws-amplify.github.io/amplify-ios/docs/Protocols/AuthDevice.html

elorzafe (changes requested):
I've just figured out `IAuthDevice` is what Cognito is returning, that shouldn't be the case, we should keep consistency across platforms

packages/auth/src/types/Auth.ts
```diff
@@ -244,17 +242,11 @@ export type AuthDeviceProps = {
...
 	constructor(props: AuthDeviceProps) {
-		this.deviceCreatedDate = props.DeviceCreateDate;
-		this.deviceKey = props.DeviceKey;
-		this.deviceLastAuthenticatedDate = props.DeviceLastAuthenticatedDate;
-		this.deviceLastModifiedDate = props.DeviceLastModifiedDate;
-		this.deviceAttributes = props.DeviceAttributes;
+		this.id = props.DeviceKey;
+		this.name = props.DeviceAttributes[1].Value;
 	}
 }
```
manueliglesias: This file is supposed to have TypeScript types only, this class should live somewhere else, or even better, not exist, as it is basically the same as `IAuthDevice`.

```typescript
// instead of:
new AuthDevice({
    DeviceKey: '',
    DeviceAttributes: [
      { Name: 'device_status', Value: 'valid'},
      { Name: 'device_name', Value: 'myDevice'},
      // ...
    ],
});

// do this on Auth.ts:
const deviceInfo: IAuthDevice = {
  id: props.DeviceKey,
  name: props.DeviceAttributes.find(({Name}) => Name === 'device_name').Value, 
};
```

manueliglesias (changes requested):
Thanks @evcodes !

Just one change requested and it is good to go on my side

packages/auth/src/common/AuthErrorStrings.ts
```diff
@@ -9,4 +9,6 @@ export enum AuthErrorStrings {
 	INVALID_MFA = 'Invalid MFA type',
 	EMPTY_CHALLENGE = 'Challenge response cannot be empty',
 	NO_USER_SESSION = 'Failed to get the session because the user is empty',
+	NETWORK_ERROR = 'Network Error',
+	DEVICE_CONFIG = 'Device tracking has not been properly configured in Cognito User Pools',
```
elorzafe, line 13: Maybe this is more clear 
```suggestion
	DEVICE_CONFIG = 'Device tracking has not been configured in this User Pool'
```

elorzafe (approved):
This looks great @evcodes ! 🎉 

One minor comment related to error message to make it more clear.

We need to have integration in place before merging this

elorzafe:
You might also need to fix the issue with circleci

elorzafe (approved):
Lets merge this! 

Good job @evcodes ! 🌮 🎉 🥇

sammartinez (approved):
LGTM 🌮  And Approval due to the `angular-devkit/core` issue within the Bundle Size run

manueliglesias (approved):
LGTM 👍 

Thanks Eddy!

Issue #8337, opened 2021-05-24
Hosted UI returns tokens, but Hub and onAuthUIStateChange continue to give login error.


jmarshall9120:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Vue

### Amplify APIs

Authentication

### Amplify Categories

auth

### Environment information

<details>

```
  System:
    OS: Windows 10 10.0.18362
    CPU: (8) x64 Intel(R) Xeon(R) CPU E3-1505M v6 @ 3.00GHz
    Memory: 16.63 GB / 31.81 GB
  Binaries:
    Node: 10.16.3 - C:\Program Files\nodejs\node.EXE
    npm: 6.9.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 90.0.4430.212
    Edge: Spartan (44.18362.267.0)
    Internet Explorer: 11.0.18362.1
  npmPackages:
    @aws-amplify/ui-vue: ^1.0.9 => 1.0.9
    @vue/cli-plugin-babel: ~4.5.0 => 4.5.13
    @vue/cli-plugin-eslint: ~4.5.0 => 4.5.13
    @vue/cli-plugin-router: ~4.5.0 => 4.5.13
    @vue/cli-plugin-unit-mocha: ~4.5.0 => 4.5.13
    @vue/cli-plugin-vuex: ~4.5.0 => 4.5.13
    @vue/cli-service: ~4.5.0 => 4.5.13
    @vue/eslint-config-standard: ^5.1.2 => 5.1.2
    @vue/test-utils: ^1.0.3 => 1.2.0
    aws-amplify: ^4.0.2 => 4.0.2
    babel-eslint: ^10.1.0 => 10.1.0
    chai: ^4.1.2 => 4.3.4
    core-js: ^3.6.5 => 3.12.1 (2.6.12)
    dev:  1.0.0
    eslint: ^6.7.2 => 6.8.0
    eslint-plugin-import: ^2.20.2 => 2.23.3
[...]

PR #8360, opened 2021-05-30
Hosted UI vue redirect url parse #8337

jmarshall9120:
#### Issue #, if available

main isssue [here](https://github.com/aws-amplify/amplify-js/issues/8337)

I tracked my work running down the bug pretty throughouly there. 

I tested in browser, and against my current dev project for which I am trying to use amplify.  Furthermore, it passed all current tests.  I did not write any new tests, because as far (as I could tell), such tests would require a much finer scope of testing that is already present, and the current tests salsify that previous builds are not broken.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,12 +4,12 @@ Config
 lerna-debug.log
 node_modules
 packages/*/node_modules/**
-packages/**/lib/
-packages/**/lib-esm/
-packages/**/dist/
-packages/**/es/
-packages/**/esm/
-packages/**/cjs/
+# packages/**/lib/
+# packages/**/lib-esm/
+# packages/**/dist/
+# packages/**/es/
+# packages/**/esm/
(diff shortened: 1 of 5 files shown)


jmarshall9120:
I went back and added tests.  The more I thought about it, unless I build a '#/' vs '#' test into the two affected packages, it would be far to easy for someone not using hash url system of routing to undo the change in the future.

abdallahshaban557:
Hi @jmarshall9120 - we are closing this PR in response to this comment on the original [Github issue](https://github.com/aws-amplify/amplify-js/issues/8337#issuecomment-853400142). Here is additional documentation on changing the routing type to avoid this issue https://v3.router.vuejs.org/guide/essentials/history-mode.html#html5-history-mode

Issue #8199, opened 2021-04-29
No notification on selecting Preferred MFA Type


m-bilal-dev9:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication

### Amplify Categories

auth

### Environment information

<details>

```
# Put output below this line
npx: installed 1 in 4.549s

  System:
    OS: macOS 11.1
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 61.21 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.0 - ~/.nvm/versions/node/v14.16.0/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.16.0/bin/yarn
    npm: 6.14.11 - ~/.nvm/versions/node/v14.16.0/bin/npm
  Browsers:
    Chrome: 90.0.4430.93
    Firefox: 83.0
    Safari: 14.0.2
  npmGlobalPackages:
    eslint-plugin-prettier: 3.4.0
    npm: 6.14.11
    yarn: 1.22.10

```

</details>

### Describe the bug

Using AmplifySelectMfaType component for allowing users to set their preferred MFA type.
Once user clicks on verify button, API request is being but users are not getting notified.

### Expected behavior

An hub event should be fired for MfaTypeSet for both cases success & failure.
Or There should be a notification at-least in case of success/failure to notify the user.

### Reproduction steps

1. Add UI component for AmplifySelectMfaType with MFATypes props
2. Select the preferred MFA from UI and click verify

### Code Snippet

```javascript
        <AmplifySelectMfaType
          MFATypes={{ TOTP: true, SMS: true, Optional: true }}
          authData={data}
        />
```

sidverma32:
Hi @chrisbonifacio ,  I would like to pick this up. If you can guide me how to proceed from here?

eddiekeller:
[...]

PR #8394, opened 2021-06-04
feat: adding toast notification when selecting an MFA type

eddiekeller:
#### Description of changes

This PR adds a toast notification for when a user selects a MFA type. The notification text comes from the success/error logs that we already write when an MFA type is seleted. Our code contains TODO comments for adding this and it addresses this open issue - https://github.com/aws-amplify/amplify-js/issues/8199

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/8199

Loaded up a sample application to verify behavior and added unit tests for testing the component state changes.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
```diff
@@ -44,6 +46,7 @@ export class AmplifySelectMFAType {
 		this.isNoMFA = false;
 		this.isTOTP = false;
 		this.isSMS = false;
+		this.showToast = false;
```
hvergara, line 49: suggestion: to keep consistency with other boolean values, can we rename it to `isToastVisible`?

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
```diff
@@ -175,6 +175,15 @@ export class AmplifySelectMFAType {
 			<div>
 				{this.contentBuilder()}
 				{this.TOTPSetup ? <amplify-totp-setup user={this.authData} /> : null}
+				{this.showToast && this.selectMessage ? (
+					<amplify-toast
+						message={this.selectMessage}
+						handleClose={() => {
```
hvergara, line 181: can we put this handler logic into a separate method/function? makes unit testing easier

packages/amplify-ui-components/tsconfig.json
```diff
@@ -4,7 +4,7 @@
 		"allowUnreachableCode": false,
 		"declaration": false,
 		"experimentalDecorators": true,
-		"lib": ["dom", "es2017"],
+		"lib": ["dom", "es2017", "esnext.asynciterable", "ES2018.AsyncGenerator"],
```
hvergara, line 7: question: why do we need to add extra TS lib support and how is related with this change?

eddiekeller: It isn't related to this change - it is related to the CloudWatch update since that adds additional libraries to the `core` package. However, we had to revert that and get AppSec to take a look, so I can update this and remove these lib additions for now. But eventually, they will need to be added.

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
eddiekeller: Sure, can do that.

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
eddiekeller: Np

slaymance (approved):
I pulled down the changes and verified the following functionality with a sample application:

- Toast message appears for all MFA radio button selections after "Verify" is clicked
- Toast message disappears on radio button selection change
- Toast message disappears on toast close

packages/amplify-ui-components/src/components/amplify-select-mfa-type/amplify-select-mfa-type.tsx
```diff
@@ -175,6 +175,15 @@ export class AmplifySelectMFAType {
...
+							this.selectMessage = '';
```
slaymance, line 182: The class is instantiated with an initial state of `null` for `this.selectMessage`.  The property is defined as a `string` type so an empty string makes more sense as a default state, but maybe we would want to reset it to `null` here for consistency.

eddiekeller: That's fine, I can reset it to null.

PR #8395, opened 2021-06-04
fix(@aws-amplify/datastore): expose timestamp fields and prevent writing to read-only fields

nickarocho:
### Problem
Currently, if the `"addtimestampfields": true` flag is defined in an Amplify project via `cli.json`, when a user runs `amplify codegen models`, the outputted code in `schema.js` will have these timestamp fields added (with values managed by AppSync):
```js
"createdAt": {
    "name": "createdAt",
    "isArray": false,
    "type": "AWSDateTime",
    "isRequired": false,
    "attributes": [],
    "isReadOnly": true
},
"updatedAt": {
    "name": "updatedAt",
    "isArray": false,
    "type": "AWSDateTime",
    "isRequired": false,
    "attributes": [],
    "isReadOnly": true
}
```

Codegen has been updated to add these fields ([aws-amplify/amplify-codegen#114](https://github.com/aws-amplify/amplify-codegen/pull/114)) but the current implementation of Amplify JS datastore doesn't expose these fields to the customer. 

Additionally, these fields (and potentially others down the road) are defined with the `isReadOnly` flag set to `true`, meaning the user should not be able to modify the values of these fields with data input to a mutation, and we don't currently have anything in place to check for this `isReadOnly` flag on the input fields.

### Solution

We can't rely on the timestamp's field names to be `createdAt` or `updatedAt` since custom labels can be generated with code like:
```gql
type Post @model(timestamps:{createdAt: "createdOn", updatedAt: "updatedOn"} {
  id: ID!
}
```
so we need to check for the `isReadOnly` flag added by codegen on all fields regardless of their defined label.

This PR adds the `isReadOnly` property to the `@ModelField` interface - which codegen will set to `true` for timestamp fields (regardless of their label). This will ensure that they still get requested in the selection set, and read in the response, but they just won't be able to be modified.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -791,6 +791,22 @@ jobs:
           test_name: 'amazon-cognito-identity-js'
           category: auth
           sample_name: amazon-cognito-identity-js
+  integ_react_device_tracking:
+    parameters:
+      browser:
+        type: string
+    executor: js-test-executor
+    <<: *test_env_vars
+    working_directory: ~/amplify-js-samples-staging/samples/react/auth/device-tracking
+    steps:
+      - prepare_test_env
+      - integ_test_js:
+          test_name: 'cognito-device-tracking'
+          framework: react
+          category: auth
+          sample_name: device-tracking
+          spec: device-tracking
+          browser: << parameters.browser >>
   integ_rn_ios_storage:
     executor: macos-executor
     <<: *test_env_vars
@@ -1132,6 +1148,15 @@ workflows:
             - build
           filters:
             <<: *releasable_branches
+      - integ_react_device_tracking:
+          requires:
+            - integ_setup
+            - build
+          filters:
+            <<: *releasable_branches
+          matrix:
+            parameters:
+              <<: *test_browsers
       - integ_rn_ios_storage:
           requires:
             - integ_setup
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -812,19 +812,19 @@ jobs:
     steps:
       - integ_test_rn_ios
 
-  integ_rn_android_storage:
-    executor: macos-executor
-    <<: *test_env_vars
-    working_directory: ~/amplify-js-samples-staging/samples/react-native/storage/StorageApp
-    steps:
-      - integ_test_rn_android
+  # integ_rn_android_storage:
+  #   executor: macos-executor
+  #   <<: *test_env_vars
+  #   working_directory: ~/amplify-js-samples-staging/samples/react-native/storage/StorageApp
+  #   steps:
+  #     - integ_test_rn_android
 
-  integ_rn_android_storage_multipart_progress:
-    executor: macos-executor
-    <<: *test_env_vars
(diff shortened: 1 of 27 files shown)


amhinson (approved):
LGTM 👍

svidgen (commented):
Are we setting `createdAt` and/or `updatedAt` anywhere on the client side? Or are these fields only populated upon coming back from the server?

The real question being: When a customer *queries* a model, can they actually depend on these fields being populated 100% of the time? Or, suppose they want to *delete* all `T` where `createdAt <= now - 90 days` (or whatever). Will it do the right thing? Or, will the local store have a bunch of `nulls` in those fields, and therefore behave erratically?

nickarocho:
@svidgen from my understanding, the `createdAt` and `updatedAt` fields are automatically added via AppSync regardless of whether they are passed in explicitly to the app's `schema.gql` with custom keys like:

```gql
type Post @model(timestamps:{createdAt: "createdOn", updatedAt: "updatedOn"} {
  id: ID!
}
```

I haven't tested the query question you brought up yet, but I'm pretty positive we can rely on those fields always having a value as long as AppSync does its job populating the values.

@iartemiev can you double check me on this?

iartemiev:
@nickarocho @svidgen read-only timestamp fields are set in the AppSync request resolver as Nick suggested.

When you have the timestamp fields implicitly defined (i.e., not included in the model definition):
```gql
type Post @model {
  id: ID!
  title: String!
}
```
These fields will be read-only as far as the client is concerned and `post.createdAt` and `post.updatedAt` will be `undefined` immediately after creating a new record via `DataStore.save`. Once that record syncs to the cloud, those fields will contain timestamp values. 

If the developer adds the fields to their model definition explicitly:
```gql
type Post @model {
  id: ID!
  title: String!
  createdAt: AWSDateTime!
  updatedAt: AWSDateTime!
}
```
They become read/write client-side. If they're defined as non-nullable like above, they will always require a valid AWSDateTime value to be set at the time of record creation on the client side.

If they are defined as nullable (`createdAt: AWSDateTime`, i.e., without the `!`) then the client can omit them at the time of record creation and the resolver will set those values. 

> When a customer queries a model, can they actually depend on these fields being populated 100% of the time?
If the fields are implicitly defined, records that have synced to the cloud will have these fields populated 100% of the time.

> suppose they want to delete all T where createdAt <= now - 90 days (or whatever). Will it do the right thing? Or, will the local store potentially have null timestamps in some of those records and behave incorrectly?

[Deleting with a predicate](https://docs.amplify.aws/lib/datastore/sync/q/platform/js#update-and-delete-with-predicate) such as `DataStore.delete(T, c => c.createdAt('lte', 'some-date'))` applies the predicate to the remote data. Since we utilize a mutation queue that processes mutations in order, one-by-one, by the time the delete gets executed it will apply to any items created before it, even if they hadn't yet synced at the moment `DataStore.delete` was invoked.

svidgen (approved):

nickarocho:
Closing due to branch/git problems, starting fresh with https://github.com/aws-amplify/amplify-js/pull/8509

Issue #7763, opened 2021-02-17
React Native Auth SignIn: Double Clicking SignIn crashes with Error: Pending sign-in attempt already in progress


rraczae:
**Describe the bug**
In the react native Auth library the SignIn component does not have a pending state and the `Sign In` button can be tapped again before the request completes. This leads to an exception that crashes the app.

https://github.com/aws-amplify/amplify-js/blob/3e423d017a98362a0117e1218dc3fa55abbff609/packages/aws-amplify-react-native/src/Auth/SignIn.tsx#L103

**To Reproduce**
Wrap your app with the `useAuthenticator` or drop the `<Authenticator />` component anywhere and double click sign in after filling out fields with anything.

**Expected behavior**
Sign In button is disabled (and loading indicator is displayed)

**Code Snippet**
-

**Screenshots**
![image](https://user-images.githubusercontent.com/73604600/108251142-aec37480-711c-11eb-8eef-391702db2511.png)

Sandeepbadri:
On the first click of button I get no response. So I click on it again hoping  to some response but I get the same error.

leowz:
Same issue here. If I am not logged in already, it's pending sign-in. If I was logged in before, then the screen just show a 
"loading..." text and an empty screen. Some body from Amplify team could help on this issue?

cleett:
+1

austindvo:
+1

SeanRenet:
Has anyone figure this out yet?

meehawk:
From the looks of it, it seems that it is not communicated to the user that the sign in request has already been made via a UI. This leads to user tapping the button again and hence the error. Would something like disabling the button while the sign in is in progress help or appear as a potential solution?  First time contributor here

pablo-lopez-domowicz:
+1

danieloi:
+1

danieloi:
My issue was with a dependency. This solved it:
https://github.com/aws-amplify/amplify-js/issues/7191#issuecomment-728234443

BeifeiZhou:
same issue here.

bcs-gbl:
+1

PFStein:
+1

chrisbonifacio:
If you're experiencing this issue, please try the steps mentioned here https://github.com/aws-amplify/amplify-js/issues/7191#issuecomment-728238337

Please let me know if this works for you.

rraczae:
[...]

PR #8411, opened 2021-06-07
fix(@aws-amplify/aws-amplify-react-native): disable sign in button wh…

calebpollman:
…ile awaiting sign in request (#7763)

#### Description of changes

- disable sign in button of `SignIn` component while a sign in request is active
- update `SignIn.signIn` method:
  - add early return and logging when a sign in request is active
  - add `async`/`await` declaration and remove extraneous `Promise` return
  - set `hasPendingSignIn` to `true` when called with no pending sign in request, set `hasPendingSignIn` to `false` on sign in success or error

#### Issue #, if available
https://github.com/aws-amplify/amplify-js/issues/7763

Locally tested on iOS simulator and Android emulator

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/aws-amplify-react-native/src/Auth/SignIn.tsx
```diff
@@ -76,9 +84,11 @@ export default class SignIn extends AuthPiece<ISignInProps, ISignInState> {
 					this.error(err);
 				}
 			});
+		this.setState({ hasPendingSignIn: false });
```
amhinson, line 87: This seems fine to me, but just to double check, does this happen to cause the `Can only update a mounted or mounting component. ` warning? I'm mainly just asking because the `this.changeState` method will technically unmount this `SignIn` component when switching to the new screen, so there might be chance that this `setState` call happens after switching screens.

calebpollman: In my local testing did not come across that warning but it is a legit concern, thanks for checking

amhinson (approved):
LGTM 👍

packages/aws-amplify-react-native/src/Auth/SignIn.tsx
rraczae: not sure exactly whats going on in `this.error(..)` or `this.changeState(...)` so it might be a non issue but in case an error happens there `hasPendingSignIn` will remain true and the component will be stuck.

rraczae: Not sure what's happening in `this.error(...)` or `this.changeState(...)` so this might be a non-issue but in case there is an error the component will remain stuck with `hasPendingSignIn: true`.

calebpollman: Thanks for the comment @rraczae

Line 87 will run regardless of error or success as neither `this.changeState` or `this.error` prevent execution

rraczae: Hey @calebpollman thanks for responding! What I meant is if there is an exception lets say on line 82 then then the outer promise gets rejected and await will re-throw. We can enforce this by putting a `throw` statement on line 80. Wrapping it in a try catch would also prevent this but again, chances are probably minimal. Thanks again for checking this!

Rob

Issue #7675, opened 2021-02-03
Integrating Analytics (Pinpoint) with Auth (Cognito)


kylekirkby:
**Which Category is your question related to?**
Analytics (Pinpoint) / Auth (Cognito) / Plus more...

**Amplify CLI Version** 
4.21.0

**What AWS Services are you utilizing?**
Pinpoint / Cognito / plus more...

**Provide additional details e.g. code snippets**

I've recently enabled the Analytics category and I'm looking to use Pinpoint for my transactional emails. I've checked the Pinpoint console and I could not see the Endpoints for each of my Cognito users. Should this be the case?

Secondly, if the above is not the case, how do I sync my users from Cognito with Pinpoint? I've seen that Cognito offers the Analytics integration via the console (which I've enabled) and from my understanding this should create the endpoints for users in Pinpoint automatically... Is this meant to happen when a new user is created in Cognito? 

Could we get the Auth category to update when Analytics is added so that the two services "talk" to each other? I setting when adding the analytics category asking users whether to integrate with the auth category would be great!

Thoughts?

Best regards,

Kyle

abdallahshaban557:
hello everyone, we are currently investigating enabling this feature, allowing the endpoint to be correctly updated AFTER a user signs in. Which would make the endpoints a lot more usable. We are thinking of starting with updating the user-id, and maybe the name and email. What other data points would make sense to update?

cunneen:
Hi @abdallahshaban557 , that would be a great start!

mzarnawski:
I cannot believe that @gspeicher solution https://github.com/aws-amplify/amplify-js/pull/8413 has been closed without being merged. 
In my opinion his solution could be even simplified. It is not even needed to add AnalyticsMetadata as another configuration parameter for Auth class. All you need is to generate a UUID within amazon-cognito-identity-js and send it ALWAYS in AnalyticsMetadata object for a couple of Cognito events mentioned in the docs (see below). 
If Cognito-Pinpoint integration is not enabled AnalyticsMetadata will be ignored by Cognito.

https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-pinpoint-integration.html#cognito-user-pools-pinpoint-integration-client
https://aws.amazon.com/premiumsupport/knowledge-center/pinpoint-cognito-user-pool-analytics/

PR #8413, opened 2021-06-08
add support for Cognito/Pinpoint integration via AnalyticsMetadata pa…

gspeicher:
This PR is a first-cut draft with minimal support to enable Cognito Analytics integration with Pinpoint.

For Cognito Analytics to work, it is not enough to enable it on the Cognito resource; you also need to include the AnalyticsMetadata parameter in Cognito API calls. Otherwise, email/phone endpoints will not be automatically created in Pinpoint for your users when they sign up. However, Amplify abstracts away all calls to the Cognito API, and it does not currently support the inclusion of that parameter, so Amplify developers are left with no simple way to communicate with their customers using Pinpoint.

Amplify already supports the creation and maintenance of Cognito and Pinpoint resources. In order to achieve full integration, it should support enabling Analytics on the user pool, as well as automatically including the metadata parameter in its Cognito API calls. This PR addresses the latter; it enables the use of the metadata parameter as a step towards the larger goal of fully automating the complete integration.

With this changeset in place, you can manually enable Cognito Analytics in the Cognito console under General Settings / Analytics, and add the following code snippet near the top of your main app source to configure the Auth service:

```
import {Auth} from 'aws-amplify';

Auth.configure({
  analyticsMetadata: {
    AnalyticsEndpointId: awsmobile.aws_mobile_analytics_app_id,
  },
});
```

Then your Cognito users will automatically have their Pinpoint endpoints created.

#### Description of changes
This change adds the analyticsMetadata config parameter to the Auth service so that it can be configured as shown above. That value then gets pushed into the `CognitoUserPool` provided by the `amazon-cognito-identity-js` package, and the value from the pool then gets used as the `AnalyticsMetadata` parameter in Cognito API requests where it is supported. Note that many of these requests happen from methods on `CognitoUser` so it simply gets the value through its existing `pool` attribute.

This is a minimal change with low risk for breakage and high impact in terms of opening up integration between Cognito and Pinpoint.

#### Issue #, if available
Issue #7675 

To validate these changes, I manually stepped through user signUp and email address verification and confirmed that the Pinpoint endpoint was created. More thorough testing and feedback from the community would be helpful and appreciated.

`yarn test` is failing for me with an `UnhandledPromiseRejection` in `@aws-amplify/core` but I believe that has something to do with my local environment and does not appear to be related to these changes. If anyone can verify this PR as non-breaking, that would be greatly appreciated.

I cannot think of any useful automated tests to add or change since this is an optional parameter that relies on a working environment to properly test. Perhaps we could validate the inclusion of the AnalyticsMetadata value in API requests when the value has been configured in Auth, but the architecture does not appear to lend itself to that test. If anyone can think of anything useful and feasible, please let me know. Otherwise the checklist item related to addition/modification of tests can likely be removed.

Also not sure the best place to add documentation for the steps necessary to enable and configure the integration as described above. Maybe the [Authentication - Getting Started](https://docs.amplify.aws/lib/auth/getting-started/q/platform/js) page? Thoughts and contributions welcome here as well.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amazon-cognito-identity-js/index.d.ts
+++ b/packages/amazon-cognito-identity-js/index.d.ts
@@ -296,11 +296,16 @@ declare module 'amazon-cognito-identity-js' {
 		codeDeliveryDetails: CodeDeliveryDetails;
 	}
 
+	export interface ICognitoAnalyticsMetadata {
+		analyticsEndpointId: string;
+	}
+
 	export interface ICognitoUserPoolData {
 		UserPoolId: string;
 		ClientId: string;
 		endpoint?: string;
 		Storage?: ICognitoStorage;
+		AnalyticsMetadata?: ICognitoAnalyticsMetadata;
 		AdvancedSecurityDataCollectionFlag?: boolean;
 	}
 
--- a/packages/amazon-cognito-identity-js/src/CognitoUser.js
+++ b/packages/amazon-cognito-identity-js/src/CognitoUser.js
@@ -420,6 +420,7 @@ export default class CognitoUser {
 			ClientId: this.pool.getClientId(),
 			AuthParameters: authParameters,
 			ClientMetadata: clientMetaData,
+			AnalyticsMetadata: this.pool.analyticsMetadata,
 		};
 		if (this.getUserContextData(this.username)) {
 			jsonReq.UserContextData = this.getUserContextData(this.username);
@@ -618,6 +619,7 @@ export default class CognitoUser {
 			ChallengeResponses: finalUserAttributes,
 			Session: this.Session,
 			ClientMetadata: clientMetadata,
+			AnalyticsMetadata: this.pool.analyticsMetadata,
 		};
 		if (this.getUserContextData()) {
 			jsonReq.UserContextData = this.getUserContextData();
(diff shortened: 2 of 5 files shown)


gspeicher:
Update: it seems that I (and others in online forums) misinterpreted the expected value for the `AnalyticsEndpointId` parameter in Cognito API calls. The documentation is rather sparse, but the expected value appears to be the ID of the actual endpoint that Cognito will create or update in Pinpoint. This poses somewhat of a chicken-and-egg problem, since the desired endpoint ID would need to be somehow known for a user prior to SignUp/SignIn. This could be as simple as using the user-supplied value in the username/login field when using password authentication, but it gets more complicated when using OAuth to login.

In addition, I'm not sure if this is a problem with my changes to the Auth library or with the underlying Cognito/Pinpoint integration, but when I create a user and verify their contact info (email or phone), Cognito ends up creating two distinct endpoints in Pinpoint: one gets created at the time of account creation (with the requested endpoint ID, OptOut:'ALL', and no Address), and the other gets created at the time of email/phone verification (generated endpoint ID, OptOut:'NONE', and expected Address). There is no obvious way to know the generated endpoint ID for the user so that the Auth service can be configured to include it in subsequent API calls. I have an active Technical Support case open with AWS Premium support to try and get to the bottom of what's going on here and see if there is a path forward.

Those extra endpoints with no valid Address are also causing downstream problems with Pinpoint Journeys; an issue for which I have a separate open support case and which I am told will be fixed in 3–4 weeks. In the meantime, the extra endpoints are preventing any messages from being delivered to the valid endpoints corresponding to the same user.

In conclusion, given the current state of affairs I would not recommend anyone plan to use this changeset in any production capacity until the behavior is better understood and any related underlying platform issues are resolved. For now, my solution is to create/maintain Pinpoint endpoints in the Cognito PostAuth trigger.

Issue #7382, opened 2020-12-10
AmplifyAuthenticator component formFields hint prop only works for the last one


jerocosio:
**Describe the bug**
When adding a hint on the formFields prop for AmplifySignUp, AmplifySignIn, AmplifyConfirmSignUp and AmplifyConfirmSignIn only the last input displays the hint correctly, the rest don't and I can't even see it when looking at the generated HTML.

**To Reproduce**

1. Add the AmplifyAuthenticator from @aws-amplify/ui-react
2. Add AmplifySignUp as a children
3. Create more than 1 formField with a 'hint'
4. See that the hint is only displayed on the last input

**Expected behavior**
To see the hints in the different inputs that it was added to.

**Code Snippet**
```
<AmplifySignUp
        headerText=""
        slot="sign-up"
        usernameAlias="email"
        formFields={[
          {
            type: 'name',
            label: 'Nombre *',
            placeholder: 'Escribe tu nombre',
            required: true,
            hint: 'Escribe tu nombre completo.',
          },
          {
            type: 'family_name',
            label: 'Apellidos *',
            placeholder: 'Escribe tus apellidos completos',
            required: true,
            hint: 'Escribe los 2 apellidos.',
          },
          {
  type: 'email',
  label: 'Correo electrónico *',
  placeholder: 'Escribe tu correo electrónico',
  required: true,
},
          {
            type: 'phone_number',
            label: 'Teléfono celular *',
            placeholder: '10 dígitos, enviaremos un SMS para confirmar',
            required: true,
            dialCode: '+52',
            hint: 'Enviaremos un SMS para confirmar.',
          },
          {
  type: 'password',
  label: 'Contraseña *',
  placeholder: 'Escribe tu contraseña (8 caracteres mínimo)',
  required: true,
  hint: '8 caracteres mínimo',
},
        ]}
        submitButtonText="Registrarme"
      />
```
**Screenshots**
<img width="493" alt="Screen Shot 2020-12-10 at 8 36 34 AM" src="https://user-images.githubusercontent.com/933413/101785936-f695a200-3ac2-11eb-8f73-8262b0761628.png">
 

<details>
  <summary><strong>Environment</strong></summary>

```
System:
    OS: macOS 10.15.7
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 53.10 MB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
[...]

PR #8415, opened 2021-06-08
feat: adding hint text props to our auth form field mappings

eddiekeller:
#### Description of changes

This PR adds `hint` support for our auth form fields. Previously, if a user added a `hint` prop to their `formFields` definition, the `hint` would only show up if the field was a `password` or `code` field. This led to an issue being raised and confusion around why the hints were not showing up on their other form fields.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/7382

Validating locally in a sample app and ran `yarn test`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-ui-components/src/components/amplify-username-field/amplify-username-field.tsx
```diff
@@ -23,6 +23,8 @@ export class AmplifyUsernameField {
 	@Prop() inputProps?: object;
 	/** Will disable the input if set to true */
 	@Prop() disabled?: boolean;
+	/** Used for the hint text that displays underneath the input field */
+	@Prop() hint?: string | FunctionalComponent | null;
```
wlee221, line 27: Ah, I actually didn't know about `FunctionalComponent` type! I've been using `JSXBase.IntrinsicElements` to type these. I did some experimenting and it seems to handle all expected types. LGTM!

wlee221 (approved):
We have a lot of parity problems between subcomponents, don't we 😅 I believe this will be easier to maintain once we move away from `formField` to use more native composition methods. It'll be less error-prone to make all attributes available by default, than trying to define which ones are available with one-by-one props.

Anyways... changes LGTM! thanks @eddiekeller 👍

PR #8418, opened 2021-06-09
feat(amazon-cognito-identity-is, @aws-amplify/auth)

elorzafe:
#### Description of changes
Adding token revocation feature

Added cypress test in samples repo

CircleCI workflow [here]( https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/8107/workflows/c555b390-d2f0-4a14-ab64-211bf7eddc59/jobs/44033)

#### TODO
- Add `token-revocation.spec.js` on integration test suite to validate this changes after merge

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are in samples repo(https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)

### DIFF
--- a/packages/amazon-cognito-identity-js/src/CognitoUser.js
+++ b/packages/amazon-cognito-identity-js/src/CognitoUser.js
@@ -1391,9 +1391,8 @@ export default class CognitoUser {
 			return callback(null, this.signInUserSession);
 		}
 
-		const keyPrefix = `CognitoIdentityServiceProvider.${this.pool.getClientId()}.${
-			this.username
-		}`;
+		const keyPrefix = `CognitoIdentityServiceProvider.${this.pool.getClientId()}.${this.username
+			}`;
 		const idTokenKey = `${keyPrefix}.idToken`;
 		const accessTokenKey = `${keyPrefix}.accessToken`;
 		const refreshTokenKey = `${keyPrefix}.refreshToken`;
@@ -1556,9 +1555,8 @@ export default class CognitoUser {
 	 * @returns {void}
 	 */
 	cacheDeviceKeyAndPassword() {
-		const keyPrefix = `CognitoIdentityServiceProvider.${this.pool.getClientId()}.${
-			this.username
-		}`;
+		const keyPrefix = `CognitoIdentityServiceProvider.${this.pool.getClientId()}.${this.username
+			}`;
 		const deviceKeyKey = `${keyPrefix}.deviceKey`;
 		const randomPasswordKey = `${keyPrefix}.randomPasswordKey`;
 		const deviceGroupKeyKey = `${keyPrefix}.deviceGroupKey`;
@@ -1573,9 +1571,8 @@ export default class CognitoUser {
 	 * @returns {void}
 	 */
 	getCachedDeviceKeyAndPassword() {
-		const keyPrefix = `CognitoIdentityServiceProvider.${this.pool.getClientId()}.${
-			this.username
-		}`;
+		const keyPrefix = `CognitoIdentityServiceProvider.${this.pool.getClientId()}.${this.username
+			}`;
 		const deviceKeyKey = `${keyPrefix}.deviceKey`;
 		const randomPasswordKey = `${keyPrefix}.randomPasswordKey`;
 		const deviceGroupKeyKey = `${keyPrefix}.deviceGroupKey`;
@@ -1592,9 +1589,8 @@ export default class CognitoUser {
 	 * @returns {void}
 	 */
 	clearCachedDeviceKeyAndPassword() {
-		const keyPrefix = `CognitoIdentityServiceProvider.${this.pool.getClientId()}.${
-			this.username
-		}`;
+		const keyPrefix = `CognitoIdentityServiceProvider.${this.pool.getClientId()}.${this.username
+			}`;
 		const deviceKeyKey = `${keyPrefix}.deviceKey`;
 		const randomPasswordKey = `${keyPrefix}.randomPasswordKey`;
 		const deviceGroupKeyKey = `${keyPrefix}.deviceGroupKey`;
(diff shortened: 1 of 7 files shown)


packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -1966,11 +1962,104 @@ export default class CognitoUser {
...
-	signOut() {
+	signOut(revokeTokenCallback) {
+		// If tokens won't be revoked, we just clean the client data.
+		if (!revokeTokenCallback || typeof revokeTokenCallback !== "function") {
+			this.cleanClientData();
+
+			return;
+		}
+
+		this.getSession((error, _session) => {
```
hkjpotato, line 1973: Some concerns (not blocking) of using `getSession` here. Potentially it might use the `refreshToken` to [refresh the session](https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/src/CognitoUser.js#L1434). In the backend, this will generate another pair of `idToken` and `accessToken`. I feel it is a bit odd to do this while the intention is to revoke the session (and what if it fails due to "Refresh Token has been revoked"...?).

It seems the intention to call `this.getSession` here is to ensure `signInUserSession` is populated on `this`, which will be used by the following `revokeTokens`. Considering the fact that we need to check the `origin_jti` claim of the `accessToken` (as a feature flag), before making the revoke call, this might be a reasonable solution though.

packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -1966,11 +1962,104 @@ export default class CognitoUser {
...
+
+		const tokensToBeRevoked = [];
+
+		if (!this.signInUserSession) {
+			const error = new Error('User is not authenticated');
+
+			return revokeTokenCallback(error);
+		}
+
+		if (!this.signInUserSession || !this.signInUserSession.getAccessToken()) {
```
hkjpotato, line 1999: the first condition `!this.signInUserSession` is already covered by line#1993, do we still need to check it here?

packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -1966,11 +1962,104 @@ export default class CognitoUser {
...
+
+		const refreshToken = this.signInUserSession.getRefreshToken().getToken();
+		const accessToken = this.signInUserSession.getAccessToken();
+
+		if (this.isSessionRevocable(accessToken)) {
+			if (refreshToken) {
+				tokensToBeRevoked.push(refreshToken);
+			}
+
+			const tokenRevocationPromises = tokensToBeRevoked.map(token => new Promise((res, rej) => {
```
hkjpotato, line 2013: why do we use an array to store the `tokensTobeRevoked`? By the current logic only the refresh token will be in it. Is this for potential future extension?

packages/amazon-cognito-identity-js/src/CognitoUser.js
elorzafe: Is a good point!

We need to do something for sessions that are revoked, especially with global signOut. I can dive deep into that use case!

That could be a quick follow up

packages/amazon-cognito-identity-js/src/CognitoUser.js
elorzafe: Good catch!

packages/amazon-cognito-identity-js/src/CognitoUser.js
elorzafe: Good catch too, I think this is not necessary at the moment. I will refactor this, we can chat more offline

elorzafe (commented):
Thanks KJ!

amhinson (commented):
Looks great @elorzafe 👍  
Just a few small things:
- We'll need to update the TS types for `CognitoUser.signOut` to add the callback
- `integ_auth_test_cypress_no_ui` will need to be added to the `requires` section of the `deploy` job

iartemiev (approved):
I reviewed the CCI config change specifically - LGTM!

elorzafe:
> Looks great @elorzafe 👍
> Just a few small things:
> 
> * We'll need to update the TS types for `CognitoUser.signOut` to add the callback
> * `integ_auth_test_cypress_no_ui` will need to be added to the `requires` section of the `deploy` job

Great catch @amhinson !

hkjpotato (approved):
LGTM! thanks for updating it.

amhinson (approved):
LGTM 👍

sammartinez (approved):
LGTM 🌮

packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -1966,11 +1962,93 @@ export default class CognitoUser {
 	 * This is used for the user to signOut of the application and clear the cached tokens.
 	 * @returns {void}
 	 */
-	signOut() {
+	signOut(revokeTokenCallback) {
```
lawmicha, line 1965: If no callback is passed then that's the developers intention to not revoke tokens? shouldn't revoking tokens be the default behavior as part of signing out?

I'm looking at this and trying to figure out how this translates to other platforms, ie. iOS/Android

packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -1966,11 +1962,93 @@ export default class CognitoUser {
...
+
+		if (!this.signInUserSession) {
+			const error = new Error('User is not authenticated');
+
+			return revokeTokenCallback(error);
+		}
+
+		if (!this.signInUserSession.getAccessToken()) {
+			const error = new Error('No Access token available');
+
```
lawmicha, line 2001: should this first get the access token, if no available then check the id token?

Issue #7613, opened 2021-01-24
onAuthStateUiChange(authStateHandler) can call authStateHandler after unsubscribe with AuthState changes to SignedIn 


bwindsor:
**Describe the bug**
It is possible for the `authStateHandler` to be called after the unsubscribe function is called. In particular, this happens when `authState` changes to `AuthState.SignedIn`.

**To Reproduce**
1. Create an auth backend and create a test Cognito user.
1. Create a file with the **code snippet** below in it.
1. Render `MyComponent` in your app.
1. Open your app in the browser and open dev tools
1. Enter the username and password of the test user and click **Sign In**
1. Observe that the change handler is called once more after the `unsubscribe` function has been called

**Expected behavior**
I expect that `authStateHandler` is never called after `unsubscribe` is called.

**Code Snippet**
```ts
const MyComponent = () => {
    const [signedIn, setSignedIn] = React.useState<boolean>(false);

    return signedIn ? <div>You are signed in</div> : <MyAuthenticator onSignedIn={() => setSignedIn(true)}/>;
}

type MyAuthenticatorProps = {
    onSignedIn: () => void;
}
const MyAuthenticator: React.FC<MyAuthenticatorProps> = ({onSignedIn}) => {
    React.useEffect(() => {
        console.log('subscribe');
        const authStateHandler: AuthStateHandler = (nextAuthState, authData) => {
            console.log('change handler called: ' + nextAuthState);
            if (nextAuthState === AuthState.SignedIn) {
                onSignedIn();
            }
        }
        const unsubscribe = onAuthUIStateChange(authStateHandler);        return () => {
            console.log('unsubscribe');
            unsubscribe();
        }
    }, [onSignedIn]);

    return <AmplifyAuthenticator />
}
```

**Screenshots**
Observed console output. For my test user pool I had MFA enabled, hence the `confirmSignIn` step, but whether that is there or not is irrelevant to this issue.
![image](https://user-images.githubusercontent.com/22724286/105639504-9f7c3e80-5e70-11eb-9683-0b7f055f65d7.png)

**What is Configured?**
My Cognito pool is created by terraform. I am only using Amplify for the UI, not to configure backend resources.
I configure Amplify as follows
```ts
import { Auth } from '@aws-amplify/auth';
Auth.configure({
    // REQUIRED - Amazon Cognito Region
    region: config.AWS_REGION,
    // OPTIONAL - Amazon Cognito User Pool ID
[...]

PR #8428, opened 2021-06-10
Prevent side-effects from firing after unsubscribe

ericclemmons:
#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/7613

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


jacoblogan:
I've been able to manually validate the above fix as seen in the attached screenshots showing the issue before the fix but not in the identical scenario after the fix.  

BeforeFix
AfterFix

This has removed the side effects from triggering authUIStateHandler after being unsubscribed.

wlee221 (approved):
This LGTM! Thanks @jacoblogan and @ericclemmons.

I've looked at why there are two `signedIn` events. We're handling authState from two different places:

1) https://github.com/aws-amplify/amplify-js/blob/e8f9e5a9d754853b29e7bf9789973b055a517bbc/packages/amplify-ui-components/src/components/amplify-confirm-sign-in/amplify-confirm-sign-in.tsx#L116-L117

2) https://github.com/aws-amplify/amplify-js/blob/e8f9e5a9d754853b29e7bf9789973b055a517bbc/packages/amplify-ui-components/src/components/amplify-authenticator/amplify-authenticator.tsx#L68-L70

both calling the same functions. As per our discussion, I'm in favor of merging this PR as refactoring `handleExternalAuthEvent` is risky.

PR #8431, opened 2021-06-10
feat(storage): Adding copy API to Storage

### LINKED ISSUE #5998: AWS amplify StorageClass not supporting copyObject while AWS S3 SDK does
I need a feature where I can copy file within a bucket to another 'folder' 

This option is provided by https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#copyObject-property

but not with https://aws-amplify.github.io/amplify-js/api/classes/storageclass.html

Can we make this functionality available to AWS Amplify

jamesaucode:
#### Description of changes
Adding `Storage.copy`, allowing users to copy files across levels or identity id (if file is in protected)
Integ tests: https://github.com/aws-amplify/amplify-js-samples-staging/pull/236
Doc changes: https://github.com/aws-amplify/docs/pull/3309

#### Issue #, if available
fixes #5998 

Added integration tests, unit tests and manually tested

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -114,6 +113,126 @@ export class AWSS3Provider implements StorageProvider {
...
+		s3.middlewareStack.remove(SET_CONTENT_LENGTH_HEADER);
+
+		const copier = new AWSS3ProviderMultipartCopier({
+			params,
+			emitter,
+			s3client: s3,
+		});
+		if (progressCallback) {
+			if (typeof progressCallback === 'function') {
+				emitter.on(COPY_PROGRESS, progress => {
```
elorzafe, line 199: What is `emiter` sending on the callback? Like the shape of that data

packages/storage/src/providers/AWSS3ProviderMultipartCopy.ts
```diff
@@ -0,0 +1,224 @@
...
+	 */
+	public async copy(): Promise<CompleteMultipartUploadCommandOutput | CopyObjectCommandOutput> {
+		let uploadId: string = undefined;
+		try {
+			const { Size } = await this._getObjectMetadata();
+			this.totalBytesToCopy = Size;
+			// Fallback to basic CopyObject if the file is smaller than the minimum required 5MB.
+			if (this.totalBytesToCopy <= AWSS3ProviderMultipartCopier.minPartSize) {
+				const copyObjectCommand = new CopyObjectCommand(this.params);
+				const result = await this.s3client.send(copyObjectCommand);
```
elorzafe, line 98: Is this because it happen on the server side and not on the client, we can chat about this offline

packages/storage/src/types/AWSS3Provider.ts
```diff
@@ -0,0 +1,59 @@
...
+	/** Total bytes copied */
+	loaded: number;
+	/** Total bytes to copy */
+	total: number;
+};
+
+type S3ClientCopyCommandInput = S3ClientCopyCommandParams;
+
+/** A subset of S3 CopyCommand params allowed for AWSS3Provider. */
+interface S3ClientCopyCommandParams {
```
elorzafe, line 15: Sounds a little weird to add some of this, if it was a copy for example you wont need `contentEncoding` why would you like to change it?

packages/storage/src/types/AWSS3Provider.ts
```diff
@@ -0,0 +1,59 @@
...
+interface StorageCopyConfig {
+	level?: StorageLevel;
+	/** if set to true, automatically sends Storage Events to Amazon Pinpoint */
+	track?: boolean;
+	provider?: string;
+	/**
+	 * callback function that gets called on each successful part copied to track
+	 * the copy progress 
+	 **/
+	progressCallback?: (progress: CopyProgress) => any;
```
elorzafe, line 42: If is only 0/100 I dont see if we need this right now. We can add it later when it can show more significant data

packages/storage/src/types/Provider.ts
```diff
@@ -16,6 +17,9 @@ export interface StorageProvider {
 	// cancel an in-flight request
 	cancel?(request: Promise<any>): void;
 
+	// copy object from src to dest
+	copy?(src: StorageCopyTarget, dest: StorageCopyTarget, config?): Promise<any>;
```
elorzafe, line 21: Seems the target doesn't need the identityId, users can copy to its own prefix (identityId) or public. Not to others users, might be worth it have a different type for Destination without identityId. What do you think?

packages/storage/src/types/Storage.ts
```diff
@@ -19,3 +19,5 @@ export interface StorageOptions {
 	level?: string;
 	credentials?: object;
 }
+
+export type StorageCopyTarget = { key: string; [key: string]: any };
```
elorzafe, line 23: Shouldn't this have `level` and `identityId` for source?

elorzafe (changes requested):
Thanks @jamesaucode this looks great!

We can chat more offline to look on some cases

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -114,6 +113,126 @@ export class AWSS3Provider implements StorageProvider {
...
+			serverSideEncryption,
+			SSECustomerAlgorithm,
+			SSECustomerKey,
+			SSECustomerKeyMD5,
+			SSEKMSKeyId,
+		} = opt;
+		const { level: srcLevel = DEFAULT_STORAGE_LEVEL, identityId: srcIdentityId } = src;
+		const { level: destLevel = DEFAULT_STORAGE_LEVEL } = dest;
+		if (srcLevel !== 'protected' && srcIdentityId) {
+			logger.warn(`IdentityId only works if level is protected, but level is set to ${srcLevel}`);
```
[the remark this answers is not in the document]
jamesaucode: Will change the warning message here

packages/storage/src/types/AWSS3Provider.ts
jamesaucode: Will remove `contentEncoding`

packages/storage/src/common/StorageErrorStrings.ts
```diff
@@ -1,5 +1,7 @@
 export enum StorageErrorStrings {
 	NO_CREDENTIALS = 'No credentials',
+	NO_SRC_KEY = 'src should be an object with the property "key" with value of type string',
```
elorzafe, line 3: I would say `source param and destination param` instead of `src` or `dst`, `src` is only visible inside of the `copy` function

elorzafe (approved):
LGTM 🎉 🌮 

Just confirm which metadata like contentEncoding, contentType, etc are copied from original file so they are not needed.

amhinson (approved):
Looking specifically at the CircleCI config, LGTM 👍

Issue #6885, opened 2020-09-28
Sign up form validation not working with "required" formField


borisdayma:
**Describe the bug**

In the sign up form, "required" fields are only considered for email, username and password.
The form does not catch any incomplete fields that have been set to "required".

**To Reproduce**

In Vue, create a sign up form from `<amplify-sign-up>` component and add some required formFields.
Then you can still sign-up without filling in those required fields.

**Expected behavior**
We should get a notification mentioning that those fields are required

**Code Snippet**

```vue
<template>
  <amplify-authenticator username-alias="email">
    <amplify-sign-up
      slot="sign-up"
      username-alias="email"
      :form-fields.prop="formFields"
    />
  </amplify-authenticator>
</template>

<script>
export default {
  data() {
    return {
      formFields: [
        {
          type: 'email',
          required: true,
        },
        {
          type: 'given_name',
          label: 'First Name *',
          placeholder: 'Enter your first name',
          required: true,
        },
        {
          type: 'password',
          required: true,
        },
        {
          type: 'phone_number',
          required: true,
        },
      ],
    }
  },
}
</script>
```

**Screenshots**

We can create an account without filling in "First name" and "Phone Number", which are required fields.

![image](https://user-images.githubusercontent.com/715491/94489156-9670ae80-01a9-11eb-981f-2ce19f37857c.png)

Once, I got a message saying that given_name was required but it never happened again…

**Additional information**

I see there is a `validationData` parameter which appears to be empty.
I assume it should be completed based on these `required` parameters.
It would be nice to also explain how to use it for more complex scenarios.

amhinson:
@borisdayma Could you also share what version of the Amplify packages you are using?

borisdayma:
I use:
`@aws-amplify/ui v2.0.2`
`aws-amplify v3.3.1`

sammartinez:
@borisdayma Thanks for reporting this, in checking the code, I am seeing this as a miss on our side and we need to further step up validation for fields besides phone number, email, and password as you have called out. I do want to call out that we are busy with other priorities within the library and may not get to this for a little bit of time. However, if you are up for it, we would accept a contribution from you if this is something you are want to work on if you have time yourself. Thanks ahead of time!

ericclemmons:
Here are the two places that could have `required` added:

[...]

PR #8455, opened 2021-06-15
fix: fix bug where required fields were not being enforced

eddiekeller:
#### Description of changes

Previously, if a user added custom form fields and wanted them to be "required", they weren't actually being enforced as required fields. This PR changes that by adding the `required` prop to our `amplify-input`, which generates the actual `input` field on the page.

It also moves the `required` prop from the parent `amplify-form-field` component in the phone form field to the text input field, as having the `required` on the top level actually didn't enforce it properly and sometimes caused a bug where the form field was untargettable.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/6885

Test locally in UI. GIF showing how it works - 

out

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-ui-components/src/components/amplify-input/amplify-input.tsx
```diff
@@ -132,6 +134,7 @@ export class AmplifyInput {
 					class="input"
 					value={this.value}
 					disabled={this.disabled}
+					required={this.required}
```
ericclemmons, line 137: Thanks for patching this!  Such a small, easy thing that customers expected to be there by default.

You're the best @eddiekeller !

ericclemmons (approved):

Issue #2089, opened 2018-11-12
Switching Unauthenticated Users to Authenticated Users


ahmetcetin:
It's common scenario, user starts as unauthenticated, later registers. I allow user to use services as unauthenticated in identity pool. Using amplify, if I run this scenario, an identity created as unauthenticated, later user signs up, unauthenticated identity stays as it is, and new authenticated identity created. It looks there is a way to convert unauthenticated users to authenticated ones in cognito, any idea how to achieve it using amplify?

Initially Unauthenticated User
```
// set the default config object
var creds = new AWS.CognitoIdentityCredentials({
    IdentityPoolId: 'us-east-1:1699ebc0-7900-4099-b910-2df94f52a030'
});
AWS.config.credentials = creds;
``` 

Switch to Authenticated User
```
// Called when an identity provider has a token for a logged in user
function userLoggedIn(providerName, token) {
    creds.params.Logins = creds.params.Logins || {};
    creds.params.Logins[providerName] = token;

    // Expire credentials to refresh them on the next request
    creds.expired = true;
}
```
Do we have access to CognitoIdentityCredentials function through amplify?

** Which Category is your question related to? **
Auth
** What AWS Services are you utilizing? **
S3, AppSync
** Provide additional details e.g. code snippets **
https://docs.aws.amazon.com/cognito/latest/developerguide/switching-identities.html

elorzafe:
Hi @ahmetcetin thanks for your feedback, what is the use you have in mind?

jarstingstall:
👍

marwanhilmi:
Am looking for support for this feature as well, currently using `amazon-cognito-identity-js` directly but would like to switch to `Amplify.Auth`. Can we possibly workaround by manually expiring the credentials and forcing a refresh?

With `amazon-cognito-identity-js`, am currently doing `AWS.config.credentials.expired = true` before fetching new authenticated credentials.

salimhamed:
@elorzafe What would you suggest as a work around?

NikoMontana:
I cannot even let unauth users use api, from switching between unauth. and auth. are different worlds. Did anybody solve this?

powerful23:
[...]

PR #8466, opened 2021-06-17
Switching Unauthenticated Users to Authenticated Users

hkjpotato:
#### Description of changes
This PR fixes issue #2089 "Switching Unauthenticated Users to Authenticated Users" in such a scenario: transferring the unauthenticated identity into the authenticated one at the first time when the user signs in through the user pool. 

The design has been verified by the design review (thanks to @reesscot), and backend implementation(Cognito identity pool) clarification ([internal](https://paste.amazon.com/show/kaijih/1623951826)), to ensure it does not create unexpected side effect on the persistent data layer.

#### Implementation details

_where is the fix_
```
Auth.signUp -> Auth.signIn(get session) -> Credentials.setCredentialsFromSession[fix is here]
```

_quick summary_
```
setCredentialsFromSession {
  identityId
  
+  if (localCachedIdentity)
+   identityId = localCachedIdentity
+  else
    identityId = getId();
  
  getCredentialsFromIdentity(identity, sessionInfo)
}
```
The fix adds an attempt to find the locally cached guest identity, If found, that guest identity will be used for the following `getCredentialsForIdentity` call and thus link the guest identity to a first time login.

_concerns_
The major concern during implementation is to find the "right place" to apply the fix. Unlike [the opt-in fix on `Auth.ts` callback](https://github.com/aws-amplify/amplify-js/issues/2089#issuecomment-622331432), here the fix is in the `Credentials.ts`, where the interactions with identity pool happen. The reasons are:
1. In the design review, we have decided that this is a "fix", not an opt-in/opt-out feature. This will align the JS with the current IOS behavior.
2. Using the cached identity to fetch credentials actually matches the intended design of the [Cognito identity pool sdk](https://github.com/aws/aws-sdk-js-v3/blob/main/packages/credential-provider-cognito-identity/src/fromCognitoIdentityPool.ts#L30) itself, and it won't create unexpected side effect.
    - for a first-time user, the provided guest id will become the primary id for the login, and thus fix the issue #2089.
    - for a returning user, it will always keep the previous primary id associated with the login, even the guest id is provided.
    - in both cases,  the guest id will be "linked" and disqualified as an unauth identity, thus right at that moment, we remove it from localStorage.
    - if a guestId is not found (user has not tried to obtain the guest credentials before), then it remains the same as today's flow.

We no longer clean up the local cached identity at sign out (by `Credentials.clear`), but do it at a better time: when we know it is disqualified as a guest identity, and has to be removed.
  - this helps avoid it being [cleaned up too early](https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/Auth.ts#L529).

note: in the future, we might only want to attempt the linking between a guest identity and the login in the "sign up" flow. To 
do so, we need a mechanism to track the change history of the "auth state" in `Auth.ts`.

_lesson learned_
we rely on [aws-sdk-v3](https://github.com/aws/aws-sdk-js-v3/blob/main/packages/credential-provider-cognito-identity) for interacting with identity pool. It exposes declarative api (e.g. `fromCognitoIdentityPool`) and imperative api (`getCommandId`).
- The declarative API will [cached](https://github.com/aws/aws-sdk-js-v3/blob/main/packages/credential-provider-cognito-identity/src/fromCognitoIdentityPool.ts#L27) or [mutate](https://github.com/aws/aws-sdk-js-v3/blob/main/packages/credential-provider-cognito-identity/src/fromCognitoIdentity.ts#L41) the network request/response.
- If we (Credentials) want to handle the local cache and data processing ourself, we should favor the imperative api to keep single source of truth.

#### Issue #, if available
#2089

- api test (postman), and backend verification.
- yarn test (unit test)
- [integ test](https://github.com/aws-amplify/amplify-js-samples-staging/tree/master/samples/react/auth/guest-to-auth-user)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/core/src/Credentials.ts
```diff
@@ -254,16 +256,7 @@ export class CredentialsClass {
...
-		try {
-			await this._storageSync;
-			identityId = this._storage.getItem(
-				this._getCognitoIdentityIdStorageKey(identityPoolId)
-			);
-			this._identityId = identityId;
-		} catch (e) {
-			logger.debug('Failed to get the cached identityId', e);
-		}
+		const identityId = this._identityId = await this._getGuestIdentity();
```
eddiekeller, line 259: Can we toss this in a `try/catch` block? Else we risk this causing issues if an error occurs.

packages/core/src/Credentials.ts
```diff
@@ -551,6 +570,36 @@ export class CredentialsClass {
...
+	/* operations on local stored guest identity */
+	private async _getGuestIdentity() {
+		const { identityPoolId } = this._config;
+		try {
+			await this._storageSync;
+			return this._storage.getItem(
+				this._getCognitoIdentityIdStorageKey(identityPoolId)
+			);
+		} catch (e) {
+			logger.debug('Failed to get the cached identityId', e);
```
reesscot, line 584: nit: for extra clarity, this could say:
`logger.debug('Failed to get the cached guest identityId', e);`

packages/core/src/Credentials.ts
```diff
@@ -551,6 +570,36 @@ export class CredentialsClass {
...
+	private async _setGuestIdentity(identityId) {
+		const { identityPoolId } = this._config;
+		try {
+			await this._storageSync;
+			this._storage.setItem(
+				this._getCognitoIdentityIdStorageKey(identityPoolId),
+				identityId,
+			);
+		} catch (e) {
+			logger.debug('Failed to cache identityId', e);
```
reesscot, line 597: nit: for extra clarity, this could say:
`logger.debug('Failed to cache guest identityId', e);`

reesscot (changes requested):
Looking great, just a few nits for extra clarity in debug statements

reesscot:
Also, this is the sample I created to test the guest->auth scenario:

https://github.com/aws-amplify/amplify-js-samples-staging/tree/master/samples/react/auth/guest-to-auth-user

packages/core/src/Credentials.ts
```diff
@@ -430,20 +430,20 @@ export class CredentialsClass {
...
+			const guestIdentityId = await this._getGuestIdentityId();
 
-			// if no guestIdentity found, call getId with the logins
-			// for first-time users, this will return a brand new identity
-			// for returning users, this will retrieve the previous identity assocaited with the logins
-			let generatedOrRetrievedIdentity;
-			if (!guestIdentity) {
+			// if no guestIdentityId found, call getId with the logins
+			// for a first-time user, this will return a brand new identity
+			// for a returning user, this will retrieve the previous identity assocaited with the logins
```
reesscot, line 437: nit: spelling
```suggestion
			// for a returning user, this will retrieve the previous identity associated with the logins
```

hkjpotato: thanks for catching it! I wonder if I just commit it from the github UI

reesscot: I think you can do that, especially with a comment.

packages/core/src/Credentials.ts
hkjpotato: I have update the according, but still see this "requested change"...should I just comment on this

hkjpotato:
> Looking great, just a few nits for extra clarity in debug statements

addressed, please re-review it :)

reesscot (approved):
LGTM 🚀

packages/core/src/Credentials.ts
```diff
@@ -438,25 +429,60 @@ export class CredentialsClass {
...
 		*/
 		const credentialsProvider: CredentialProvider = async () => {
-			const { IdentityId } = await cognitoClient.send(
-				new GetIdCommand({
-					IdentityPoolId: identityPoolId,
-					Logins: logins,
+			// try to fetch the local stored guest identity, if found, we will associate it with the logins
+			const guestIdentityId = await this._getGuestIdentityId();
+
+			// if no guestIdentityId found, call getId with the logins
```
elorzafe, line 435: This comment is not necessary because it reads clear from the code

packages/core/src/Credentials.ts
```diff
@@ -438,25 +429,60 @@ export class CredentialsClass {
...
+			// for a first-time user, this will return a brand new identity
+			// for a returning user, this will retrieve the previous identity associated with the logins
```
elorzafe, line 437: I would move this comment before calling `GetIdCommand`

packages/core/src/Credentials.ts
```diff
@@ -438,25 +429,60 @@ export class CredentialsClass {
...
+					Expiration,
+					SecretKey,
+					SessionToken,
+				},
+				// single source of truth for the primary identity associated with the logins
+				// only if a guest identity is used for a first-time user, that guest identity will become its primary identity
+				IdentityId: primaryIdentityId,
+			} = await cognitoClient.send(
+				new GetCredentialsForIdentityCommand({
+				  IdentityId: guestIdentityId || generatedOrRetrievedIdentityId,
```
elorzafe, line 461: Why do you need to send `guestIdentityId` first and not `generatedOrRetrievedIdentityId`? I would do the opposite so you don't need to link `guestIdentityId` for that user

packages/core/src/Credentials.ts
```diff
@@ -438,25 +429,60 @@ export class CredentialsClass {
...
-			};
 
-			const credentialsFromCognitoIdentity = fromCognitoIdentity(
-				cognitoIdentityParams
-			);
+			this._identityId = primaryIdentityId;
+			if (guestIdentityId) {
+				// if guestIdentity is found and used by GetCredentialsForIdentity
+				// it will be linked to the logins provided, and disqualified as an unauth identity
+				logger.debug(`The guest identity ${guestIdentityId} has been successfully linked to the logins`);
```
elorzafe, line 470: I would add this message only on the following `if` statement

elorzafe (changes requested):
Thanks @hkjpotato for this!

I think setting `guestIdentity` for every user is not necessary. Ping me offline if you need to discuss.

Thanks again!

packages/core/src/Credentials.ts
hkjpotato: as discussed offline, this is due to the cognito implementation of these two api in the backend https://paste.amazon.com/show/kaijih/1623951826

the `getId` call will always return a brand new identity for a new customer. The brand new identity will be set as primary identity and cannot be changed.

hkjpotato: > Why do you need to send `guestIdentityId` first and not `generatedOrRetrievedIdentityId`? I would do the opposite so you don't need to link `guestIdentityId` for that user

as discussed offline, this is due to the cognito implementation of these two api in the backend https://paste.amazon.com/show/kaijih/1623951826

the getId call will always return a brand new identity for a new customer. The brand new identity will be set as primary identity and cannot be changed.

elorzafe: Thanks for the clarification @hkjpotato

packages/core/src/Credentials.ts
hkjpotato: Thanks and agree!, I wrote them down as instructions before writing the code, but should clean them up now since the good code should be self explained :)

packages/core/src/Credentials.ts
hkjpotato: yep

elorzafe (approved):
Thanks @hkjpotato 🎉

hkjpotato:
> Thanks @hkjpotato for this!
> 
> I think setting `guestIdentity` for every user is not necessary. Ping me offline if you need to discuss.
> 
> Thanks again!

Thanks @elorzafe for looking to it. I have updated accordingly and verify this design with the cognito dev.

Issue #8401, opened 2021-06-05
AmplifySignUp handleSubmit loses functionality when customizing Auth.signUp with clientMetaData


ronybot:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication

### Amplify Categories

auth, function

### Environment information

<details>

```
hi Team 
We are really struggling to get query parameters to the lambda-post-confirmation 
we ready the article there is says that we can however I can see how we can hock the query parameter in these clientmetadata values. 
this is my code 
   <AmplifyAuthenticator >
          <AmplifySignUp   />
          </AmplifyAuthenticator> 
https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-post-confirmation.html#cognito-user-pools-lambda-trigger-syntax-post-confirmation 
this is a blocker issue can you please update once you have a workaround, I need a short workaround for that.


```

</details>

### Describe the bug

hi Team 
We are really struggling to get a query parameters to the lambda-post-confirmation 
we ready the article there is says that we can however I can see how we can hock the query parameter in these clientmetadata values. 
this is my code 
   <AmplifyAuthenticator >
          <AmplifySignUp   />
          </AmplifyAuthenticator> 
https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-post-confirmation.html#cognito-user-pools-lambda-trigger-syntax-post-confirmation 
this is a blocker issue can you please update once you have a workaround , I need a short workaround for that.

### Expected behavior

clientmetadata values are set

### Reproduction steps

install amplify 
setup post confirm function 
set querystring value 
expect to see in the post confirm function log the query string values.

### Code Snippet

```javascript
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line


```

</details>

ericclemmons:
We discussed as a team how to resolve this within `AmplifySignUp` and will communicate the fix both here & internally via: https://t.corp.amazon.com/V376611509/overview

[...]

PR #8474, opened 2021-06-18
feat: add handleSignUp prop to override auth signup method

eddiekeller:
#### Description of changes

This PR adds an additional prop to the `SignUp` component - `handleSignUp`. Currently, if a user actually decides to override the existing `handleSubmit` method, it will...pretty much lose most of the functionality that is required. `handleSignUp` is meant specifically so customers can override the call to `Auth.signUp()` and provide additional data, like `clientMetadata`.

I also cleaned up the existing `signUp` method in our `amplify-sign-up` component. It was pretty large with a ton of logic packed in. I just moved some of that logic into individual methods. Same stuff happens...just a bit cleaner now.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/8401

Local testing.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
```diff
@@ -117,16 +122,17 @@ export class AmplifySignUp {
...
-		}
-		if (!Auth || typeof Auth.signUp !== 'function') {
-			throw new Error(NO_AUTH_MODULE_FOUND);
+	private async authSignUp(params: SignUpParams): Promise<ISignUpResult> {
+		try {
+			const data = await Auth.signUp(params);
+
+			return data;
+		} catch (err) {
+			throw new Error(Translations.SIGN_UP_FAILED);
```
wlee221, line 131: Should we throw `err` here so that the end users know what exactly failed?

ericclemmons (approved):
This looks like a great fix, but now we have to figure out why the build is failing. (I'm good to merge it once it it is!)

From the looks of it, my guess is that `@aws-amplify/auth` isn't being built prior to `@aws-amplify/ui-components`, due to lerna not being able to correctly infer the topological order of the dependency graph.

This is because `aws-amplify` is a peer dependency, not a direct one:
> https://github.com/eddiekeller/amplify-js/blob/amplify-ui-components/fix-signup-override/packages/amplify-ui-components/package.json

One thing you can try is add `@aws-amplify/auth` to `devDependencies` for `@aws-amplify/ui-components` (at the **same version** as `aws-amplify`), and attempt to run it again: my bet is that the order will be correct then.

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.spec.ts
```diff
@@ -34,6 +34,10 @@ describe('amplify-sign-up spec:', () => {
 		it('should render `usernameAlias` as `username` by default', () => {
 			expect(signUp.usernameAlias).toEqual('username');
 		});
+
+		it('should render `handleSignUp` as defined by default', () => {
+			expect(signUp.handleSignUp).toBeDefined();
+		});
```
wlee221, line 40: In addition to this, can we check whether `signUp.handleSignUp` has been called with its `signUpAttributes` when it's overridden?

wlee221 (dismissed):
This is a _big_ improvement, thanks a lot @eddiekeller! Just got a minor nit.

packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx
ericclemmons: That's a good catch: the previous logic didn't `try` but threw this when data was falsey:

```js
const data = await Auth.signUp(this.signUpAttributes);
			if (!data) {
				throw new Error(Translations.SIGN_UP_FAILED);
```

eddiekeller: Yep, I just kept the same `throw` here that existed before. I'm good to change it to throw the specific error though if we want to do that.

ericclemmons: The previous block was different, now that I look at it:
https://github.com/aws-amplify/amplify-js/blob/efa2482fc14053145e6db8da70ff52bd7657c80b/packages/amplify-ui-components/src/components/amplify-sign-up/amplify-sign-up.tsx#L169-L172

The difference is that the previous version would automatically bubble up any errors as part of `Auth.signUp` and would throw this default error only when `null`.

Am I reading that right?

eddiekeller: Looks like it. `null` or `undefined`. Would you rather I remove the `try` block and keep it the exact same as the previous implementation?

wlee221: I'd remove the `try` block since we're try catching it anyway. 

I'm wondering though why
```ts
if (!data) { 
 	throw new Error(Translations.SIGN_UP_FAILED); 
 } 
```
was necessary though. Looking at Auth.ts, `Auth.signUp` always returns a promise so try catching would be enough.

eddiekeller: Ok, I'll remove the `try` for now then. Just keep it as identical as the current experience as possible.

wlee221: 👍  thanks!

wlee221 (approved):

Issue #8405, opened 2021-06-06
DataStore Query Performance Very Slow


sacrampton:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

DataStore

### Amplify Categories

api

### Environment information

<details>

```
# Put output below this line


```

</details>

### Describe the bug

We are porting our AppSync/GraphQL based app to DataStore and what we are finding is that queries on the cached database are very slow. We are also finding that pagination of the queries makes no difference to the performance of the queries.

There are many tables with the largest table having about 8,000 records and most other tables having less than 1000 records. There are only 5 tables/models that have data.

Whether we are connected to the network or working offline it makes no difference to the performance.

In issue #6994 it talks of a first time slowing in performance, but we see consistently slow performance.

I have recorded a video of the performance we are seeing - both online and offline - to give you an idea of the problem we are facing.

https://user-images.githubusercontent.com/6362888/120909860-f74b7180-c6bc-11eb-97dc-300f8d19ab4d.mp4

### Expected behavior

The database is local so we should be seeing almost instant responses to our queries. The performance of our existing AppSync GraphQL cache gives us almost instant responses so we expect the same or better performance than the existing AppSync GraphQL database.

### Reproduction steps

Not sure how to replicate this as the database I have has 48 models and 200 GSI's and takes about 6 hours of incremental (manual) pushes to deploy into a new environment.

### Code Snippet

```javascript
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line


```

</details>

### Mobile Device

iPhone 7

### Mobile Operating System

14.5

### Mobile Browser

Not relevant

nubpro:
I closed my original issue as I could no longer replicate the problem anymore

Do u happen to be using the Storage module in your app?

sacrampton:
Hi @nubpro - we are using the Storage module in the app for upload of photographs.

sacrampton:
[...]

PR #8481, opened 2021-06-21
feat(@aws-amplify/datastore): add RN SQLite adapter

iartemiev:
#### Description of changes

NOTE: I may extract this Adapter into its own project/repo and then make the Storage Adapters for DS pluggable.
Opening draft PR into main to preserve context with how this fits into the library and for early reviews/comments/suggestions.

The main motivation for adding this new storage adapter is to mitigate performance issues in DataStore on React Native with larger data sets (see linked issue). 

Since AsyncStorage does not provide any built-in filtering capabilities, performing `DataStore.query` operations on tables with 10k+ records is very slow. We have to first retrieve all of the records from the store, then filter them in the library code in-memory.

SQLite for React Native allows us to perform filtering, sorting, and pagination using the DB engine, which is orders of magnitude more efficient. 

E.g., Querying a model containing 40k records with `DataStore.query(Model, c => c.someField('eq', 'someVal')` takes 20+ seconds on average in AsyncStorage.
The same query takes ~ 0.3s using this SQLite adapter.

Completed Features:
* Basic CRUD operations are fully functional on non-relational models (relational are implemented but untested)
* Sync Engine is fully functional
* String/Number predicates are enabled (converted to WHERE CLAUSE)
* Pagination & sort are fully functional

Major TODOs:

- [x] Nested predicate groups
- [x] Testing operations with relational models
- [x] Predicates on List types
- [x] Test non-model types
- [x] Fix issues with many-to-many relations (mainly done. Some occasional issues with deletes, since we're not cascading currently)
- [ ] Cascading deletes on relations - omitting this for now. Will include in a later version

**This Storage Adapter will only be compatible with React Native CLI apps / ejected Expo apps. If we see a lot of demand, we may create another adapter for Expo apps as well (it would utilize the expo-sqlite package).**

AsyncStorage will remain the default StorageAdapter for RN apps. Customers will be able to opt-in to using this SQLite Adapter if it suits their needs.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/8405

* Unit tests
* Manual Q/A in sample app

TODO:
- [ ]  Add e2e Detox test to pipeline

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/__tests__/SQLiteUtils.test.ts
```diff
@@ -0,0 +1,321 @@
...
+					field: 'lastName',
+					sortDirection: 'DESCENDING',
+				},
+			];
+
+			const limit = 10;
+			const page = 3;
+
+			const expected = [
+				`SELECT * FROM Model WHERE firstName = ? and lastName LIKE ? and sortOrder > ? ORDER BY sortOrder ASC, lastName DESC LIMIT 10 OFFSET 3`,
```
svidgen, line 114: Should `OFFSET` be 30? (`page * limit`)

packages/datastore/__tests__/SQLiteUtils.test.ts
```diff
@@ -0,0 +1,321 @@
...
+			const expected = 'LIMIT 10';
+
+			expect(limitClauseFromPagination(limit)).toEqual(expected);
+		});
+
+		it('should generate valid LIMIT clause from pagination limit and page', () => {
+			const limit = 10;
+			const page = 3;
+
+			const expected = 'LIMIT 10 OFFSET 3';
```
svidgen, line 238: Isn't OFFSET an index from the start of the result set? Should this be page * limit?

packages/datastore/src/storage/adapter/SQLiteUtils.ts
```diff
@@ -0,0 +1,326 @@
...
+	}, 'ORDER BY');
+}
+
+export function limitClauseFromPagination(
+	limit: number,
+	page: number = 0
+): string {
+	let clause = `LIMIT ${limit}`;
+	if (page) {
+		clause += ` OFFSET ${page}`;
```
svidgen, line 229: Should this be `OFFSET ${page * limit}`?

packages/datastore/src/storage/adapter/SQLiteDatabase.ts
```diff
@@ -0,0 +1,159 @@
...
+
+// TODO: make these configurable
+const DB_SIZE = 200000;
+const DB_VERSION = '1.0';
+
+/*
+
+Note: 
+I purposely avoided using arrow functions () => {} in this class,
+Because I ran into issues with them in some of the SQLite method callbacks
```
svidgen, line 22: I'm curious to know more.

packages/datastore/src/storage/adapter/SQLiteDatabase.ts
```diff
@@ -0,0 +1,159 @@
...
+		await this.db.transaction(function(tx) {
+			tx.executeSql(
+				queryStatement,
+				queryParams,
+				function(_tx, res) {
+					results = res.rows.raw();
+				},
+				logger.warn
+			);
+			tx.executeSql(deleteStatement, deleteParams, () => {}, logger.warn);
```
svidgen, line 136: This arrow function is OK?

Might be worth elaborating (above) to let us know in what situations arrow functions are problematic, if you can.

packages/datastore/src/storage/adapter/SQLiteUtils.ts
svidgen: And should it be parametrized or sanitized?

packages/datastore/src/storage/adapter/SQLiteUtils.ts
```diff
@@ -0,0 +1,326 @@
...
+		}
+		return acc + ` ${field} ${orderByDirection}`;
+	}, 'ORDER BY');
+}
+
+export function limitClauseFromPagination(
+	limit: number,
+	page: number = 0
+): string {
+	let clause = `LIMIT ${limit}`;
```
svidgen, line 227: Should this be parametrized? (Or sanitized.)

packages/datastore/src/storage/adapter/SQLiteUtils.ts
```diff
@@ -0,0 +1,326 @@
...
+	limit?: number,
+	page?: number
+): SQLStatement {
+	let statement = `SELECT * FROM ${tableName}`;
+	const params = [];
+
+	if (predicate && predicate.predicates.length) {
+		const [whereClause, whereParams] = whereClauseFromPredicate(predicate);
+		statement += ` ${whereClause}`;
+		Array.prototype.push.apply(params, whereParams);
```
svidgen, line 248: For my edification, why apply the prototype's `push` here?

packages/datastore/src/storage/adapter/SQLiteUtils.ts
```diff
@@ -0,0 +1,326 @@
...
+export function queryOneStatement(
+	firstOrLast,
+	tableName: string
+): SQLStatement {
+	if (firstOrLast === QueryOne.FIRST) {
+		return [`SELECT * FROM ${tableName} LIMIT 1`, []];
+	} else {
+		// Not very efficient, but I don't think this query gets used often
+		return [
+			`SELECT * FROM ${tableName} LIMIT 1 OFFSET ((SELECT COUNT(*) FROM ${tableName}) - 1)`,
```
svidgen, line 273: What's the default sort order in SQLite? Is it deterministic? Is it by PK? Does it make sense to sort `DESC` instead?

packages/datastore/src/storage/adapter/SQLiteUtils.ts
```diff
@@ -0,0 +1,326 @@
...
+}
+
+export function queryOneStatement(
+	firstOrLast,
+	tableName: string
+): SQLStatement {
+	if (firstOrLast === QueryOne.FIRST) {
+		return [`SELECT * FROM ${tableName} LIMIT 1`, []];
+	} else {
+		// Not very efficient, but I don't think this query gets used often
```
svidgen, line 271: I'm not too concerned about performance. I trust SQLite will do *reasonably smart* things with the query. (Though, I could be completely wrong on that!) I'm slightly more concerned about the "look" of the query and whether these results are **deterministic** &mdash; if not, should we make them so? If they are deterministic, should we make it explicit *and clear*?

packages/datastore/src/storage/adapter/SQLiteUtils.ts
```diff
@@ -0,0 +1,326 @@
...
+	tableName: string,
+	predicate?: PredicatesGroup<T>
+): SQLStatement {
+	let statement = `DELETE FROM ${tableName}`;
+	const params = [];
+
+	if (predicate && predicate.predicates.length) {
+		const [whereClause, whereParams] = whereClauseFromPredicate(predicate);
+		statement += ` ${whereClause}`;
+		Array.prototype.push.apply(params, whereParams);
```
svidgen, line 297: Again, for my edification, why apply the prototype push here?

packages/datastore/src/types.ts
```diff
@@ -197,7 +197,37 @@ export namespace GraphQLScalarType {
...
+
+	export function getSQLiteType(
+		scalar: keyof Omit<
+			typeof GraphQLScalarType,
+			'getJSType' | 'getValidationFunction'
+		>
+	): 'TEXT' | 'INT' | 'REAL' | 'BOOLEAN' {
+		switch (scalar) {
+			case 'Boolean':
+				return 'BOOLEAN';
```
svidgen, line 212: See above: `BOOLEAN` &rarr; `NUMERIC`?

packages/datastore/src/types.ts
```diff
@@ -197,7 +197,37 @@ export namespace GraphQLScalarType {
...
+			case 'AWSDateTime':
+			case 'AWSEmail':
+			case 'AWSJSON':
+			case 'AWSURL':
+			case 'AWSPhone':
+			case 'AWSIPAddress':
+				return 'TEXT';
+			case 'Int':
+			case 'AWSTimestamp':
+				return 'INT';
```
svidgen, line 226: See above: `INT` &rarr; `INTEGER`?

packages/datastore/src/types.ts
```diff
@@ -197,7 +197,37 @@ export namespace GraphQLScalarType {
...
+		}
+	}
+
+	export function getSQLiteType(
+		scalar: keyof Omit<
+			typeof GraphQLScalarType,
+			'getJSType' | 'getValidationFunction'
+		>
+	): 'TEXT' | 'INT' | 'REAL' | 'BOOLEAN' {
+		switch (scalar) {
```
svidgen, line 210: This is a mix of actual SQLite types and inferred types, if I understand the SQLite docs correctly.

https://www.sqlite.org/datatype3.html#affinity_name_examples

*Do we care?*

svidgen (commented):
I can't say I've read every line here yet &mdash; *you've done a heck of a lot of work here!* &mdash; but, I've got some initial overarching questions and a few comments for you!

* Does this need a security review?
* Do we need to consider how customers would migrate data?
* I have others questions I may ask you OOB ... look out!

iartemiev:
Thank you for taking a look, Jon!

> * Does this need a security review?

No, but I'll explain in more detail in a DM.

> * Do we need to consider how customers would migrate data?

Customer data would migrate automatically via a base sync from AppSync once customers configure this as the Storage Adapter and then start DataStore.

packages/datastore/__tests__/SQLiteUtils.test.ts
iartemiev: I think you're right. Good catch!

packages/datastore/src/storage/adapter/SQLiteUtils.ts
iartemiev: It gets sanitized by input validation before it gets to the storage adapter layer, so I don't think that's a concern

packages/datastore/src/storage/adapter/SQLiteDatabase.ts
iartemiev: Callback functions with parameters have errored out for me when using arrow functions. It seems the underlying implementation relies on passing/modifying `this` context

packages/datastore/src/storage/adapter/SQLiteDatabase.ts
iartemiev: Yea, since it's a just a no-op. I clarified above.

packages/datastore/src/storage/adapter/SQLiteUtils.ts
iartemiev: It's just concatting `params` and `whereParams`. Functionally equivaly to doing `[...params, ...whereParams]` but works without re-assigning `params`

packages/datastore/src/storage/adapter/SQLiteDatabase.ts
svidgen: Oh, duh. No-op ... That makes sense!

packages/datastore/src/types.ts
iartemiev: Good call. We might as well use the actual types. I'll update

packages/datastore/src/storage/adapter/SQLiteUtils.ts
iartemiev: This is a good call-out. There's no default order and it's not deterministic. The engine basically uses whatever column it deems appropriate unless an ORDER BY clause is specified. I had to look that up and didn't realize that's how it worked when I wrote the code.

I don't think we can rely on any of the existing columns for sorting. We may have to add something like `_insertionOrder` and sort on that, then strip the field out when the models are instantiated from the result set.

nubpro:
> Customer data would migrate automatically via a base sync from AppSync once customers configure this as the Storage Adapter and then start DataStore.

I'd imagine if there is still pending items in the outbox, and our users switched over to the new app(with sqlite db), they would lose all their unsynced data. A smoother transition over would be better

packages/datastore/src/storage/adapter/SQLiteAdapter.ts
```diff
@@ -0,0 +1,453 @@
+import { ConsoleLogger as Logger } from '@aws-amplify/core';
```
svidgen, line 1: I didn't give this file a super thorough look. But, there seem to be a lot of similarities between the adapters. Have you given any thought to how these might be DRY'd out?

iartemiev: All of the adapters implement the `Adapter` interface so they'll have the same public APIs, but other than that, there should not be many similarities. Especially between this one and the other No-SQL adapter

packages/datastore/src/storage/adapter/SQLiteDatabase.ts
iartemiev: Here's an example: https://github.com/andpor/react-native-sqlite-storage/blob/master/lib/sqlite.core.js#L550

packages/datastore/src/storage/adapter/SQLiteUtils.ts
iartemiev: Yep, will paramterize

packages/datastore/src/storage/adapter/SQLiteDatabase.ts
iartemiev: To be clear, I'm talking about the SQLite library implementation of the callback functions. Not our usage of them

packages/datastore/src/storage/adapter/SQLiteUtils.ts
iartemiev: Yes, the API will return the results in the same order they are stored in DDB. Also, the other storage adapter (IDB and AsyncStorage) will preserve insertion order, so we should have consistent behavior here as well

packages/datastore/src/storage/adapter/SQLiteUtils.ts
iartemiev: Yea, `params.push(...whereParams)` will have the same effect. I just didn't think of it at the time of writing :)

packages/datastore/package.json
```diff
@@ -70,7 +71,7 @@
 						"es2015",
 						"dom",
 						"esnext.asynciterable",
-						"es2019.object"
+						"es2019"
```
manueliglesias, line 74: What else are we using from es2019 besides the Object functions?

iartemiev: Using `Array.flatMap` here: https://github.com/aws-amplify/amplify-js/pull/8481/files#diff-6eaebb7ba4effa266c383aea86ef70fdaa6230f4aeb5f5e94251d840b3d6dc3bR42

iartemiev:
@svidgen all of the changes you requested have been implemented

manueliglesias (approved):
Approved to publish to unstable, still needs some cleanup before it makes its way to main

svidgen (approved):
Thanks Ivan!

PR #8509, opened 2021-06-28
feat(@aws-amplify/datastore): expose timestamp fields and prevent writing to read-only fields

nickarocho:
(updated PR due to messy git/branch issues)

--------
[start of original PR]

### Problem

Currently, if the `"addtimestampfields": true` flag is defined in an Amplify project via `cli.json`, when a user runs `amplify codegen models`, the outputted code in `schema.js` will have these timestamp fields added (with values managed by AppSync):
```js
"createdAt": {
    "name": "createdAt",
    "isArray": false,
    "type": "AWSDateTime",
    "isRequired": false,
    "attributes": [],
    "isReadOnly": true
},
"updatedAt": {
    "name": "updatedAt",
    "isArray": false,
    "type": "AWSDateTime",
    "isRequired": false,
    "attributes": [],
    "isReadOnly": true
}
```

Codegen has been updated to add these fields ([aws-amplify/amplify-codegen#114](https://github.com/aws-amplify/amplify-codegen/pull/114)) but the current implementation of Amplify JS datastore doesn't expose these fields to the customer. 

Additionally, these fields (and potentially others down the road) are defined with the `isReadOnly` flag set to `true`, meaning the user should not be able to modify the values of these fields with data input to a mutation, and we don't currently have anything in place to check for this `isReadOnly` flag on the input fields.

### Solution

We can't rely on the timestamp's field names to be `createdAt` or `updatedAt` since custom labels can be generated with code like:
```gql
type Post @model(timestamps:{createdAt: "createdOn", updatedAt: "updatedOn"} {
  id: ID!
}
```
so we need to check for the `isReadOnly` flag added by codegen on all fields regardless of their defined label.

This PR adds the `isReadOnly` property to the `@ModelField` interface - which codegen will set to `true` for timestamp fields (regardless of their label). This will ensure that they still get requested in the selection set, and read in the response, but they just won't be able to be modified.

Codegen will now produce a model and its associated meta data types like the following in the `index.d.ts`:
```
type PostMetaData = {
	readOnlyFields: 'createdAt' | 'updatedAt';
};

export declare class Post {
	readonly id: string;
	readonly title: string;
	readonly status: PostStatus | keyof typeof PostStatus;
	readonly color?: string;
	readonly content?: string;
	readonly comments?: (Comment | null)[];
	readonly createdAt?: string;
	readonly updatedAt?: string;
	constructor(init: ModelInit<Post, PostMetaData>);
	static copyOf(
		source: Post,
		mutator: (
			draft: MutableModel<Post, PostMetaData>
		) => MutableModel<Post, PostMetaData> | void
	): Post;
}
```

### DX Typescript Enhancements

To provide an optimal developer experience, changes have been implemented (with the dedicated help of TS gurus @manueliglesias and @iartemiev 🙏) to catch these read-only properties within Intellisense and only provide hints where it logically applies.

In the user's app (if it is a TS project), when instantiating a new model Intellisense will omit any read-only properties:
Screen Shot 2021-06-29 at 11 23 03 AM

However, when using `[Model].copyOf`, Intellisense will provide hints including these read-only properties for any use case that involves reading this data (i.e. `console.log(draft.createdAt)`:
Screen Shot 2021-06-29 at 12 21 21 PM
Screen Shot 2021-06-29 at 11 26 35 AM

But since the user won't be able to overwrite these read-only properties, we want to make sure to throw a linting error if they try:
Screen Shot 2021-06-29 at 11 25 27 AM
Screen Shot 2021-06-29 at 11 26 10 AM
Screen Shot 2021-06-29 at 11 26 20 AM

#### Description of changes

Associated Codegen Changes: https://github.com/aws-amplify/amplify-codegen/pull/185

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/DataStore.ts
+++ b/packages/datastore/__tests__/DataStore.ts
@@ -561,9 +561,9 @@ describe('DataStore tests', () => {
 				field1: 'something',
 				dateCreated: new Date().toISOString(),
 				createdAt: '2021-06-03T20:56:23.201Z',
-			});
+			} as any);
 
-			expect(DataStore.save(model)).rejects.toThrowError(
+			await expect(DataStore.save(model)).rejects.toThrowError(
 				'createdAt is read-only.'
 			);
 
@@ -573,18 +573,18 @@ describe('DataStore tests', () => {
 			});
 
 			model = Model.copyOf(model, draft => {
-				draft.createdAt = '2021-06-03T20:56:23.201Z';
+				(draft as any).createdAt = '2021-06-03T20:56:23.201Z';
 			});
 
-			expect(DataStore.save(model)).rejects.toThrowError(
+			await expect(DataStore.save(model)).rejects.toThrowError(
 				'createdAt is read-only.'
 			);
 
 			model = Model.copyOf(model, draft => {
-				draft.updatedAt = '2021-06-03T20:56:23.201Z';
+				(draft as any).updatedAt = '2021-06-03T20:56:23.201Z';
 			});
 
-			expect(DataStore.save(model)).rejects.toThrowError(
+			await expect(DataStore.save(model)).rejects.toThrowError(
 				'updatedAt is read-only.'
 			);
 		});
--- a/packages/datastore/__tests__/DataStore.ts
+++ b/packages/datastore/__tests__/DataStore.ts
@@ -528,6 +528,67 @@ describe('DataStore tests', () => {
 			expect(patches2).toMatchObject(expectedPatches2);
 		});
 
+		test('Read-only fields cannot be overwritten', async () => {
+			let model: Model;
+			const save = jest.fn(() => [model]);
+			const query = jest.fn(() => [model]);
+
+			jest.resetModules();
+			jest.doMock('../src/storage/storage', () => {
+				const mock = jest.fn().mockImplementation(() => {
+					const _mock = {
+						init: jest.fn(),
+						save,
+						query,
+						runExclusive: jest.fn(fn => fn.bind(this, _mock)()),
+					};
+
+					return _mock;
+				});
+
+				(<any>mock).getNamespace = () => ({ models: {} });
+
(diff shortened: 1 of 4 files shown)


packages/datastore/__tests__/DataStore.ts
```diff
@@ -528,6 +528,67 @@ describe('DataStore tests', () => {
...
+			});
+
+			expect(DataStore.save(model)).rejects.toThrowError(
+				'createdAt is read-only.'
+			);
+
+			model = Model.copyOf(model, draft => {
+				draft.updatedAt = '2021-06-03T20:56:23.201Z';
+			});
+
```
iartemiev, line 586:
```suggestion
      // using `as any` here and below and order to test runtime readOnly field validation
			model = new Model({
				field1: 'something',
				dateCreated: new Date().toISOString(),
				createdAt: '2021-06-03T20:56:23.201Z',
			} as any);

			expect(DataStore.save(model)).rejects.toThrowError(
				'createdAt is read-only.'
			);

			model = new Model({
				field1: 'something',
				dateCreated: new Date().toISOString(),
			});

			model = Model.copyOf(model, draft => {
				(draft as any).createdAt = '2021-06-03T20:56:23.201Z';
			});

			expect(DataStore.save(model)).rejects.toThrowError(
				'createdAt is read-only.'
			);

			model = Model.copyOf(model, draft => {
				(draft as any).updatedAt = '2021-06-03T20:56:23.201Z';
			});
```

iartemiev (changes requested):
Made a suggestion on how to test that the read-only fields are enforced at runtime

packages/datastore/__tests__/DataStore.ts
nickarocho: Good call. Added these changes, thx @iartemiev!

iartemiev (approved):
LGTM! Thanks Nick!

Issue #8493, opened 2021-06-23
Picker deprecated


loeccalon:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

Authentication

### Amplify Categories

auth

### Environment information

<details>

```
# Put output below this line
My current environment setup is:
react-native: version 0.64.1
node: version 12.16.1
```

</details>

### Describe the bug

In the Signup component of the AWS Amplify, the Phone component has a Picker component that is imported from react-native framework. The problem is the Picker component is deprecated and is no longer supported in react-native version that my team is using. React Native 0.64.1 is the most current version. When the user attempts to tap on the picker component the the app crashes on android and ios. Please advise on what steps to take to help override this behavior for the Picker component. Thanks so much.

### Expected behavior

The Picker component content should be rendered when the user taps on it.

### Reproduction steps

1. Make sure you have an AWS account.
2. run npm install -g @aws-amplify/cli
3. run **amplify configure**
4. run npx react-native init your_project_name
5. cd **your_project_name**
6. run **amplify init** with respect to https://docs.amplify.aws/start/getting-started/setup/q/integration/react-native#initialize-a-new-backend
7. run npm install aws-amplify aws-amplify-react-native amazon-cognito-identity-js @react-native-community/netinfo @react-native-async-storage/async-storage
8. run amplify add auth
9. run amplify push
10. In App.js **import withAuthenticator** as stated in https://docs.amplify.aws/start/getting-started/auth/q/integration/react-native#create-login-ui
11. run **npx react-native run-android** or **npx react-native run-ios**.
12. Attempt to Sign up as a new user.
13. Tap the Picker component with the "+1" in it.
14. And then the app crashes.

### Code Snippet

```javascript
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line


```

</details>

eddiekeller:
@loeccalon Thanks for submitting this. Looking into it now.

eddiekeller:
[...]

PR #8520, opened 2021-06-30
chore(@aws-amplify/aws-amplify-react-native): migrate to community Pi…

calebpollman:
…cker component (#8493)

#### Description of changes

**This is a preliminary PR that will publish a tagged version of _aws-amplify-react-native_ for a customer to test against, once confirmed a PR against _main_ will follow**

- migrate deprecated React Native `Picker` component to [community version](https://github.com/react-native-picker/picker)
- update `Picker` style in Android `PhoneField` component to prevent truncation of rendered value
- add temporary `publish:migrate-rn-picker` script in _package.json_

#### Issue #, if available
https://github.com/aws-amplify/amplify-js/issues/8493

Manually tested on Android and iOS using both expo and vanilla React Native

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/aws-amplify-react-native/package.json
+++ b/packages/aws-amplify-react-native/package.json
@@ -21,9 +21,9 @@
 		"@babel/preset-env": "^7.9.6",
 		"@babel/preset-react": "^7.9.4",
 		"@babel/preset-typescript": "^7.9.0",
+		"@react-native-picker/picker": "^1.9.11",
 		"@types/react-native": "^0.62.5",
 		"babel-jest": "^21.2.0",
-		"react": "^16.0.0",
 		"rimraf": "^2.6.2"
 	},
 	"dependencies": {
@@ -32,7 +32,8 @@
 	},
 	"peerDependencies": {
 		"aws-amplify": "4.x.x",
-		"graphql": "^14.0.0"
+		"graphql": "^14.0.0",
+		"react": "^17.0.1"
 	},
 	"repository": {
 		"type": "git",
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -936,6 +936,7 @@ releasable_branches: &releasable_branches
       - main
       - ui-components/main
       - 1.0-stable
+      - migrate-rn-picker
 
 test_browsers: &test_browsers
   browser: [chrome, firefox]
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
 		"publish:1.0-stable": "lerna publish --conventional-commits --yes --dist-tag=stable-1.0 --message 'chore(release): Publish [ci skip]' --no-verify-access",
 		"publish:ui-components/main": "lerna publish --canary --force-publish \"*\" --yes --dist-tag=ui-preview --preid=ui-preview --exact --no-verify-access",
 		"publish:native": "lerna publish --canary --force-publish \"*\" --yes --dist-tag=native --preid=native --exact --no-verify-access",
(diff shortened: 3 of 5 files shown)


packages/aws-amplify-react-native/package.json
```diff
@@ -32,7 +32,8 @@
 	},
 	"peerDependencies": {
 		"aws-amplify": "4.x.x",
-		"graphql": "^14.0.0"
+		"graphql": "^14.0.0",
+		"react": "^17.0.1"
```
elorzafe, line 36: Why do you need react as peerDependency and not just devDependency?

elorzafe (changes requested):
You also need to add `migrate-rn-picker` in `release_branch` on `.circleci/config.yml` as well.

packages/aws-amplify-react-native/package.json
calebpollman: Still working through the root cause of this, but the community version of the React Native `Picker` component crashes due to its usage of _react_ hooks (see below), which would make sense as _react_ hooks were introduced in version `16.8.0`, while we had version `16.0.0` listed as a devDependency prior to this PR. Upgrading the devDependency _react_ version in theory should fix this but the error persisted.

Since this package is not used for standalone development and is only used with React Native projects (which by default already contain _react_), added _react_ to the peerDependencies to warn consumers of the version needs as a workaround. This didn't feel **too** strange as we do not declare _react-native_ in the deps, devDeps, or peerDeps of this package.


That being said, do not feel strongly about this being the solution (and this may be just kicking the can to an extent without a root cause) and am definitely open to feedback :)

```
Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.

This error is located at:
    in ForwardRef(PickerAndroid) (at Picker.js:174)
    in Picker
    in RCTView (at View.js:34)
    in View
    in RCTView (at View.js:34)
    in View
    in PhoneField
    in RCTView (at View.js:34)
    in View
    in RCTView (at View.js:34)
    in View (at ScrollView.js:1124)
    in RCTScrollView (at ScrollView.js:1260)
    in ScrollView (at ScrollView.js:1286)
    in ScrollView
    in TouchableWithoutFeedback
    in Wrapper
    in SignUp
    in RCTView (at View.js:34)
    in View (at SafeAreaView.js:41)
    in ForwardRef(SafeAreaView)
    in Container
    in Authenticator
    in Wrapper (created by ExpoRoot)
    in ExpoRoot (at renderApplication.js:45)
    in RCTView (at View.js:34)
    in View (at AppContainer.js:106)
    in RCTView (at View.js:34)
    in View (at AppContainer.js:132)
    in AppContainer (at renderApplication.js:39)
at node_modules/react-native/Libraries/Core/ExceptionsManager.js:104:6 in reportException
at node_modules/react-native/Libraries/Core/ExceptionsManager.js:171:19 in handleException
at node_modules/react-native/Libraries/Core/setUpErrorHandling.js:24:6 in handleError
at node_modules/expo-error-recovery/build/ErrorRecovery.fx.js:12:21 in ErrorUtils.setGlobalHandler$argument_0
at [native code]:null in flushedQueue
at [native code]:null in invokeCallbackAndReturnFlushedQueue
```

elorzafe (approved):
Thanks @calebpollman 🌮 

This looks good, lets merge it into migrate-rn-picker branch and test it more deeply

PR #8589, opened 2021-07-14
added a condition to handle OAuth query parameters only when an OAuth flow is initiated

Khairo-kh:
#### Description of changes
When using OAuth authentication, certain keywords that are used in the OAuth authentication flow cannot be used as query parameters and will always be hijacked by the Amplify auth library. 
I used the storage object in the auth class to save a value whenever the function `federatedSignIn` is called (OAuth flow always starts with calling this function). Then, I added a condition to check if this value is set before parsing the query parameters in the URL as OAuth query parameters. The value is removed from storage after OAuth flow is done.

I tested the changes on a local app with yarn linking. Query parameters "code" and "error" work fine if they are intended to be normal query parameters and no error is thrown. OAuth authentication works as expected.  

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -1902,6 +1902,7 @@ export class AuthClass {
 		response?: FederatedResponse,
 		user?: FederatedUser
 	): Promise<ICredentials> {
+		this._storage.setItem('is-oauth-code', 'true');
 		if (!this._config.identityPoolId && !this._config.userPoolId) {
 			throw new Error(
 				`Federation requires either a User Pool or Identity Pool in config`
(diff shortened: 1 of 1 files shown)


hkjpotato:
my concern is if customer directly visit the hosted UI (by typing the url for example), it will never get a chance set the local storage flag, but would like to hear Francisco's opinion on this.

evcodes:
I am noticing that the integration tests are failing, in particular `auth unit test › handleAuthResponse test › User Pools and Identity Pools`. I reran the test suite and things are still failing so we need to make sure that these cases pass before we merge.

oferitz:
Any updates? I'm willing to help fixing the tests but @hkjpotato comment need to be addressed first, so we won't put an effort for nothing.

Khairo-kh:
@oferitz The fix here works when users initiate the OAuth flow on calling `federatedSignIn()`. However, it fails to account for the case where users visit the hosted UI directly via link (which is apparently a common use case). Unfortunately, I was not able to find a way to account for this use case with the current fix implementation which utilizes local storage/

oferitz:
@Khairo-kh Thanks for the explanation. Unfortunately, my use case falls into the common case of direct link. hope the maintainers can help here.

israx:
Thank you for submitting this PR. I'm closing it as there is already a [PR](https://github.com/aws-amplify/amplify-js/pull/10755#discussion_r1081269156) in progress that we are tracking

PR #8590, opened 2021-07-14
Storage AWSS3Provider typescript improvement

jamesaucode:
#### Description of changes

Adding more typing (internal) for `AWSS3Provider` and removing implicit and explicit `any`s.
There is **no implementation changes and all the typings here are not exposed to users**. 
Main purpose is to make the code is more easily readable by other contributors, and easier to maintain down the road.

- Added a typeguard `isWithCredentials` to assert the property `credentials` exist in `config` so that it can be used for creating a valid `S3Client`.
- Re-wrote `ensureCredentials` in async/await to make it more consistent with other functions
- Added Typescript definition for the `config` parameters of each of the operations with the name `S3Provider<operation>Config`, and the corresponding output as `S3Provider<operation>Output`
  -  For `get`, it conditionally return `GetCommandOutput` if config has `{ download: true }`, else it returns a pre-signed url.
- Removed docstrings with `@private` and instead added the `private` keyword
- Removed the `progressCallback` and the `CopyProgress` type from `copy` as it's not a supported feature yet

#### Issue #, if available
N/A

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


evcodes:
I think that these are great additions to the library and will significantly improve the developer experience.

LGTM 🌮

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
```diff
@@ -280,25 +273,19 @@ describe('StorageProvider test', () => {
...
-				.mockImplementationOnce(async params => {
-					return { Body: [1, 2] };
-				});
+			jest.spyOn(S3Client.prototype, 'send').mockImplementationOnce(async params => {
+				return { Body: [1, 2] };
+			});
 			await storage.get('key', {
 				download: true,
+				// @ts-ignore - Disable TS check here as this is intentional
 				progressCallback: 'this is not a function',
```
manueliglesias, line 286: In this case it is not a big deal, but `@ts-ignore` can mask other problems, that's why I would prefer something like:

```suggestion
        progressCallback: <S3ProviderGetConfig['progressCallback']><unknown>'this is not a function', // This is intentional
```

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -310,12 +336,11 @@ export class AWSS3Provider implements StorageProvider {
...
 		}
 
-		params.Expires = expires || 900; // Default is 15 mins as defined in V2 AWS SDK
-
 		try {
 			const signer = new S3RequestPresigner({ ...s3.config });
 			const request = await createRequest(s3, new GetObjectCommand(params));
-			const url = formatUrl((await signer.presign(request, { expiresIn: params.Expires })) as any);
+			// Default is 15 mins as defined in V2 AWS SDK
+			const url = formatUrl(await signer.presign(request, { expiresIn: expires || 900 }));
```
manueliglesias, line 343: In the previous code the comment was right next to the definition making it easier to read. It would be even better to have a constant defined for it

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -539,29 +565,31 @@ export class AWSS3Provider implements StorageProvider {
...
+	private _isWithCredentials(config: StorageOptions): config is StorageOptions & { credentials: ICredentials } {
+		return typeof config === 'object' && config.hasOwnProperty('credentials');
+	}
+
+	private _prefix(config: {
+		credentials: ICredentials;
+		level?: StorageLevel;
+		customPrefix?: CustomPrefix;
+		identityId?: string;
+	}) {
```
manueliglesias, line 592: It is better to be explicit in the return types instead of relying on TypeScript's inference as changes in implementation could inadvertently change the return type of the function.

```suggestion
	}): string {
```

manueliglesias: There are other places in the PR where we could do this too

packages/storage/src/types/AWSS3Provider.ts
```diff
@@ -1,18 +1,92 @@
...
+	PutObjectRequest,
+	CopyObjectRequest,
+	_Object,
+	DeleteObjectCommandOutput,
+} from '@aws-sdk/client-s3';
+import { StorageOptions, StorageLevel } from './Storage';
+import { CancelTokenSource } from 'axios';
 
-type StorageLevel = 'public' | 'protected' | 'private';
+type ListObjectsCommandOutputContent = _Object;
```
manueliglesias, line 12: I like this alias 😄

manueliglesias (commented):
Very minor things, this looks good, thank james 😄

manueliglesias (approved):
LGTM 👍 🎉

Issue #8256, opened 2021-05-11
Security vulnerability. Insecure Randomness affecting crypto-js package, versions >=3.3.0 <4.0.0 || <3.2.1


emma3333:
### Describe the bug

Insecure randomness vulnerability brought in through `crypto-js`.

Affected versions of this package are vulnerable to Insecure Randomness. The secureRandom() method is supposed to return a cryptographically strong pseudo-random data string, but it is biased to certain digits. An attacker could be able to guess the created digits.

More information here: https://app.snyk.io/vuln/SNYK-JS-CRYPTOJS-548472

Remidiation: 
Affecting crypto-js package, versions >=3.3.0 <4.0.0 || <3.2.1
Upgrade crypto-js to version 4.0.0, 3.2.1 or higher.

The current version of crypto-js being used in aws-amplify^3.4.0 is 3.3.0 

### Expected behavior

Update version of crypto-js to remove security vulnerability

### JavaScript Framework

Vue

### Amplify APIs

Authentication

### Amplify Categories

auth

sammartinez:
thanks for reporting this @emma3333, we will look to resolve this as soon as we can

iartemiev:
We've upgraded to CryptoJS@4.0.0 as part of [aws-amplify@4.0.1](https://www.npmjs.com/package/aws-amplify/v/4.0.1)
The previous major version will be patched as well tomorrow (3.4.2 -> 3.4.3)

iartemiev:
CryptoJS has now been updated in the previous version of aws-amplify as well ([aws-amplify@3.4.3](https://www.npmjs.com/package/aws-amplify/v/3.4.3))

Thanks again, @emma3333!

Issue #7570, opened 2021-01-18
Adding Amplify Auth to a Next.js project has a huge impact of the chunk size


Riley0111:
**Describe the bug**
I want to use Amplify Auth with my [Next.js](https://nextjs.org/) project. After I added aws-amplify to my project and imported 'Auth' to my index page the **First load size increased from 64.1 kB to 308 kB**  without adding anything else. I also tried to analyze the problem with the webpack-bundle-analyzer and noticed that a lot of crypto libraries got loaded and some of them like bn.js multiple times (see screenshot below).

**To Reproduce**
1. Create a new Next.js project
2. Add aws-amplify to your package.json
3. `import {Auth} from 'aws-amplify';` in your index.js page

**Expected behavior**
That the use of the aws-amplify Auth module will add way less than 240 kB to the first load size. Or is this an expected increase of size?

**Code Snippet**

package.json:
```
"dependencies": {
    "next": "^10.0.5",
    "react": "16.14.0",
    "react-dom": "16.14.0",
    "aws-amplify": "3.3.14"
  },
```

index.js
```
import Head from 'next/head'
import {Auth} from 'aws-amplify';

export default function Home() {
    Auth.currentAuthenticatedUser()

    return (
        <div className="container">
            <Head>
                <title>Create Next App</title>
                <link rel="icon" href="/favicon.ico"/>
            </Head>

            <main>
                <h1 className="title">
                    Welcome to <a href="https://nextjs.org">Next.js!</a>
                </h1>
            </main>
        </div>
    )
}
```

**Screenshots**
Screenshot from the webpack bundle analyzer:
<img width="1680" alt="Bildschirmfoto 2021-01-18 um 10 33 16" src="https://user-images.githubusercontent.com/77612336/104897599-d803b000-5978-11eb-95ac-dc5036db6e8d.png">

Page size **without** amplify:
<img width="707" alt="Bildschirmfoto 2021-01-18 um 10 44 11" src="https://user-images.githubusercontent.com/77612336/104898657-341b0400-597a-11eb-8919-83ccb25cff07.png">

Page size **with** amplify:
<img width="707" alt="Bildschirmfoto 2021-01-18 um 10 43 57" src="https://user-images.githubusercontent.com/77612336/104898694-3d0bd580-597a-11eb-8007-31bfff2a0057.png">

**What is Configured?**
Amplify version: 3.3.14
Next.js version: 10.0.5

jamessouth:
[...]

PR #8607, opened 2021-07-20
deprecate crypto-js

hkjpotato:
#### Description of changes

Amplify has a dependency on [crypto-js](https://github.com/brix/crypto-js), yet we only use partial of its functions (Base64, SHA256, Hmac256). However, it is [not actively maintained](https://github.com/brix/crypto-js/pull/364), and has introduced issue like [NextJS bundle](https://github.com/aws-amplify/amplify-js/issues/7570) size to Amplify. 

After reviewing available options, I decide to extract the Amplify needed codes from `cryto-js`, re-org and refactor a bit based on our own need, and put it in a single file `CryptoJSHelper` (the original partially implemented `WordArray` is replaced as well). Below is a brief summary of build size comparison.

 |        | webpack | nextjs (first load js) | auth/dist/aws-amplify-auth-js |
|--------|---------|------------------------|---|
| before | 17.1 MB | 281 kB                 | 1.5 MB |
| after  | 15.7 MB | 148 kB                 |  437K |

#### Alternatives comparison
|      | [crypto-js](https://github.com/brix/crypto-js)               | [crypto-es](https://github.com/entronad/crypto-es)                               | [aws-crypto](https://github.com/aws/aws-sdk-js-crypto-helpers)                                    | [native browser](https://stackoverflow.com/a/47332317)                                            |   
|------|-------------------------|-----------------------------------------|-----------------------------------------------|-----------------------------------------------------------|
| pros | well adopted         | es6 class based, familiar syntax to dev | already a dependency, maintained by aws                          |                                                           |  
| cons | prototype based, not actively maintained, bundle size issue | security concern due to [Math.random](https://github.com/entronad/crypto-es/blob/master/lib/core.js#L144)     | no Hmac support, does not work with WordArray | is async based, need refactor of all sync based functions |  

note: both crypto-js and crypto-es are forked and based on the original `crypto-js` from [Jeff Wott](https://code.google.com/archive/p/crypto-js/), their codes (comments/logic/typing) are almost identical. This means the core algorithms/logics are stable.  See more about [its history](https://www.mobilefish.com/download/iota/cryptojs_part21.pdf).             

This is perhaps a temporary solution before we have a better option (either crypto-js merge the [bundle size fix PR](https://github.com/brix/crypto-js/pull/364)), or when `aws-cryto` supports more algorithm and flexible data type.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/7570
https://github.com/aws-amplify/amplify-js/issues/8256

- local testing and run through the flows related to the used functions
- I just modify a previous buggy test, but not adding any new unit tests as the imported code are basically from a third party lib.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -660,6 +657,7 @@ export default class CognitoUser {
 		authParameters.USERNAME = this.username;
 		authParameters.DEVICE_KEY = this.deviceKey;
 		authenticationHelper.getLargeAValue((errAValue, aValue) => {
+			console.log('get aValue', aValue);
```
evcodes, line 660: Do you want to keep this log here?

hkjpotato: good catch, no it is for my local debug, will remove it

packages/amazon-cognito-identity-js/src/utils/CryptoJSHelper.js
```diff
@@ -0,0 +1,1083 @@
...
+		}
+
+		return new WordArray(words, nBytes);
+	}
+
+	/**
+	 * Converts this word array to a string.
+	 *
+	 * @param {Encoder} encoder (Optional) The encoding strategy to use. Default: CryptoJS.enc.Hex
+	 *
```
evcodes, line 168: I would remove the extra spacing between @param and @return

packages/amazon-cognito-identity-js/src/utils/CryptoJSHelper.js
```diff
@@ -0,0 +1,1083 @@
...
+	 *     var wordArray = CryptoJS.enc.Utf8.parse(utf8String);
+	 */
+	parse(utf8Str) {
+		return Latin1.parse(unescape(encodeURIComponent(utf8Str)));
+	},
+};
+  
+/**
+ * Abstract buffered block algorithm template. (checked, similar to crypto-es)
```
evcodes, line 415: remove "(checked, similar to crypto-es)"

packages/amazon-cognito-identity-js/src/utils/CryptoJSHelper.js
```diff
@@ -0,0 +1,1083 @@
...
+ * 
+ * since we put all classes in one file, here we use IIFE to create namespace to avoid conflicts
+ */
+const algo = {};
+
+//-------"crypto-js/hmac-sha256" and "crypto-js/sha256"------
+/**
+ * HMAC algorithm.
+ */
+(function (algo) {
```
evcodes, line 663: Can you explain what the algo parameter is in this scenario?

packages/amazon-cognito-identity-js/src/utils/CryptoJSHelper.js
```diff
@@ -0,0 +1,1083 @@
...
+}(algo));
+
+
+/**
+ * SHA-256 hash algorithm.
+ */
+(function (algo) {
+
+	// Initialization and round constants tables
+	const H = [];
```
evcodes, line 793: why are the variable names H,K?

packages/auth/__tests__/auth-federation-unit-tests.ts
```diff
@@ -1,8 +1,4 @@
-jest.mock('crypto-js/sha256', () => {
-	return {
-		default: jest.fn(() => ''),
-	};
-});
+
```
evcodes, line 1: we should still have some tests for CryptoJSHelper.SHA256 I believe

packages/amazon-cognito-identity-js/src/utils/CryptoJSHelper.js
hkjpotato: it is copied from the original codebase
https://github.com/brix/crypto-js/blob/develop/src/sha256.js#L10
https://github.com/entronad/crypto-es/blob/master/lib/sha256.js#L7

packages/amazon-cognito-identity-js/src/utils/CryptoJSHelper.js
hkjpotato: I believe it is copied from the original file, somehow I feel like each version of the crypto-js keeps the same styles to respect the original author, but I will do a clean up on the format :)

packages/amazon-cognito-identity-js/src/utils/CryptoJSHelper.js
hkjpotato: it is basically an "[IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE)" style to create namespace, because they are all put in the same file. 
This is to avoid naming conflict between SHA256algo (the algorithm) and the actual `CryptoJSHelper.SHA256` function exposed, like [this](https://github.com/brix/crypto-js/blob/develop/src/sha256.js#L52)

packages/auth/__tests__/auth-federation-unit-tests.ts
hkjpotato: I need to think more of whether we want to add tests for it, as it is basically copied from third party logic.  If we add test, we probably add similar tests like the original crypto-js. Adding test also means we want to maintain it, which I am trying to avoid.

hkjpotato:
close as the fix in crypto-js has been released https://github.com/aws-amplify/amplify-js/pull/8626

PR #8620, opened 2021-07-21
feat(Geo): getAvailableMaps and getDefaultMap - Geo milestone 1 - PR 4

TreTuna:
#### Description of changes

Geo milestone 1 - PR 4/6
Adds:
- getAvailableMaps
- getDefaultMap

Pulls this data from `aws-exports` so only to get maps that were created with the Amplify CLI.

- Builds and unit tests pass
- Validated using PoC application
image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
*Documentation will be added in a follow-up PR*

### DIFF
--- a/packages/core/src/Parser.ts
+++ b/packages/core/src/Parser.ts
@@ -52,6 +52,12 @@ export const parseMobileHubConfig = (config): AmplifyConfig => {
 			region: config['aws_project_region'],
 		};
 	}
+	// Geo
+	if (config['geo']) {
+		amplifyConfig.Geo = Object.assign({}, config.geo);
+		// storageConfig acts like a catch-all, but doesn't need the 'geo' property, so we'll delete it
+		delete storageConfig.geo;
+	}
 
 	// Geo
 	if (config['geo']) {
--- a/packages/core/src/Parser.ts
+++ b/packages/core/src/Parser.ts
@@ -52,11 +52,10 @@ export const parseMobileHubConfig = (config): AmplifyConfig => {
 			region: config['aws_project_region'],
 		};
 	}
+
 	// Geo
 	if (config['geo']) {
 		amplifyConfig.Geo = Object.assign({}, config.geo);
-		// storageConfig acts like a catch-all, but doesn't need the 'geo' property, so we'll delete it
-		delete storageConfig.geo;
 	}
 
 	// Geo
--- a/packages/geo/src/Geo.ts
+++ b/packages/geo/src/Geo.ts
@@ -117,7 +117,7 @@ export class GeoClass {
 
 		if (prov === undefined) {
 			logger.debug('No plugin found with providerName', provider);
-			return 'No plugin found in Geo for the provider';
+			throw 'No plugin found in Geo for the provider';
 		}
 
 		return prov.getAvailableMaps();
@@ -135,7 +135,7 @@ export class GeoClass {
 
 		if (prov === undefined) {
 			logger.debug('No plugin found with providerName', provider);
-			return 'No plugin found in Geo for the provider';
+			throw 'No plugin found in Geo for the provider';
 		}
 
 		return prov.getDefaultMap();
--- a/packages/geo/src/Geo.ts
+++ b/packages/geo/src/Geo.ts
@@ -17,7 +17,7 @@ import {
 } from '@aws-amplify/core';
 import { AmazonLocationServicesProvider } from './Providers/AmazonLocationServicesProvider';
 
-import { GeoConfig, GeoProvider } from './types';
+import { GeoConfig, GeoProvider, MapStyle } from './types';
 
 const logger = new Logger('Geo');
 
@@ -34,11 +34,12 @@ export class GeoClass {
 		this._config = {};
 		this._pluggables = [];
 		logger.debug('Geo Options', this._config);
-
-		this.getAvailableMaps.bind(this);
-		this.getDefaultMap.bind(this);
 	}
 
+	/**
+	 * get the name of the module category
+	 * @returns {string} name of the module category
+	 */
 	public getModuleName() {
 		return GeoClass.MODULE;
 	}
@@ -104,7 +105,12 @@ export class GeoClass {
 		return this._config;
 	}
 
-	public getAvailableMaps(provider = DEFAULT_PROVIDER) {
+	/**
+	 * Get the map resources that are currently available through the provider
+	 * @param {string} provider
+	 * @returns - Array of available map resources
+	 */
+	public getAvailableMaps(provider = DEFAULT_PROVIDER): MapStyle[] | string {
 		const prov = this._pluggables.find(
 			pluggable => pluggable.getProviderName() === provider
 		);
@@ -117,7 +123,12 @@ export class GeoClass {
 		return prov.getAvailableMaps();
 	}
 
-	public getDefaultMap(provider = DEFAULT_PROVIDER) {
+	/**
+	 * Get the map resource set as default in amplify config
+	 * @param {string} provider
+	 * @returns - Map resource set as the default in amplify config
+	 */
+	public getDefaultMap(provider = DEFAULT_PROVIDER): MapStyle | string {
 		const prov = this._pluggables.find(
 			pluggable => pluggable.getProviderName() === provider
 		);
--- a/packages/geo/src/Geo.ts
+++ b/packages/geo/src/Geo.ts
@@ -34,6 +34,9 @@ export class GeoClass {
 		this._config = {};
 		this._pluggables = [];
 		logger.debug('Geo Options', this._config);
+
+		this.getAvailableMaps.bind(this);
+		this.getDefaultMap.bind(this);
 	}
 
 	public getModuleName() {
@@ -99,6 +102,32 @@ export class GeoClass {
 			this.addPluggable(new AmazonLocationServicesProvider());
 		}
 		return this._config;
+	}
+
+	public getAvailableMaps(provider = DEFAULT_PROVIDER) {
+		const prov = this._pluggables.find(
+			pluggable => pluggable.getProviderName() === provider
+		);
+
+		if (prov === undefined) {
+			logger.debug('No plugin found with providerName', provider);
+			return 'No plugin found in Geo for the provider';
+		}
+
+		return prov.getAvailableMaps();
+	}
+
+	public getDefaultMap(provider = DEFAULT_PROVIDER) {
+		const prov = this._pluggables.find(
+			pluggable => pluggable.getProviderName() === provider
+		);
+
+		if (prov === undefined) {
+			logger.debug('No plugin found with providerName', provider);
+			return 'No plugin found in Geo for the provider';
+		}
+
+		return prov.getDefaultMap();
 	}
 }
 
--- a/packages/geo/src/Providers/AmazonLocationServicesProvider.ts
+++ b/packages/geo/src/Providers/AmazonLocationServicesProvider.ts
@@ -67,12 +67,11 @@ export class AmazonLocationServicesProvider implements GeoProvider {
 
 	/**
 	 * Get the map resources that are currently available through the provider
-	 * @param {string} provider
 	 * @returns - Array of available map resources
 	 */
-	public getAvailableMaps(): MapStyle[] | string {
+	public getAvailableMaps(): MapStyle[] {
 		if (!this._config.maps) {
-			return "No map resources found in amplify config, run 'amplify add geo' to create them and ensure to run `amplify push` after";
+			throw "No map resources found in amplify config, run 'amplify add geo' to create them and ensure to run `amplify push` after";
 		}
 
 		const mapStyles: MapStyle[] = [];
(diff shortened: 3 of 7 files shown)


packages/geo/src/Geo.ts
```diff
@@ -34,6 +34,9 @@ export class GeoClass {
 		this._config = {};
 		this._pluggables = [];
 		logger.debug('Geo Options', this._config);
+
+		this.getAvailableMaps.bind(this);
+		this.getDefaultMap.bind(this);
```
Amplifiyer, line 39: Just for my understanding, why do we need this?

packages/geo/src/Geo.ts
```diff
@@ -104,6 +107,32 @@ export class GeoClass {
 		}
 		return this._config;
 	}
+
+	public getAvailableMaps(provider = DEFAULT_PROVIDER) {
```
Amplifiyer, line 111: JS docs for public APIs

Amplifiyer, line 111: Return type should be added in the method signature (for all public APIs)

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -63,4 +65,31 @@ export class AmazonLocationServicesProvider implements GeoProvider {
 		this._config = Object.assign({}, this._config, config);
 		return this._config;
 	}
+
+	public getAvailableMaps() {
+		if (!this._config.maps) {
+			return "No map resources found, run 'amplify add geo' to create them";
```
Amplifiyer, line 71:
```suggestion
			return "No map resources found in amplify config, run 'amplify add geo' to create them and ensure to run `amplify push` after";
```

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -63,4 +65,31 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+		}
+
+		const maps = [];
```
Amplifiyer, line 74: We should use the type `MapStyle` here and everywhere else we can add types in our own internal code.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -63,4 +65,31 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+			maps.push({ mapName, style });
+		}
+
+		return maps;
+	}
+
+	public getDefaultMap() {
+		if (!this._config.maps || !this._config.maps.default) {
+			return "No default map resource found, run 'amplify add geo' to create one";
+		}
```
Amplifiyer, line 88: If map(s) exist and default doesn't it may point to a bug with amplify-cli. We should handle these cases differently (no maps available or no default map available while maps are)

packages/geo/src/types/Provider.ts
```diff
@@ -11,11 +11,18 @@
...
 export interface GeoProvider {
 	getCategory(): string;
 
 	getProviderName(): string;
 
 	configure(config: object): object;
+
+	getAvailableMaps(): string | MapStyle[];
+
+	getDefaultMap(): string | MapStyle;
```
Amplifiyer, line 27: JS docs for public APIs

Amplifiyer (commented):
Great work on this PR!

packages/geo/src/Geo.ts
TreTuna: 👍

TreTuna: 👍

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: 👍

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: 👍

packages/geo/src/types/Provider.ts
TreTuna: 👍

packages/geo/src/Geo.ts
TreTuna: Hmm... Actually, these may not need `.bind` on them. Though, I'll do a little testing here, but I believe I can remove them.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -63,4 +64,44 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+	 * Get the map resource set as default in amplify config
+	 * @param {string} provider
+	 * @returns - Map resource set as the default in amplify config
+	 */
+	public getDefaultMap(): MapStyle | string {
+		if (!this._config.maps) {
+			return "No map resources found in amplify config, run 'amplify add geo' to create them and ensure to run `amplify push` after";
+		}
+		if (!this._config.maps.default) {
+			return "No default map resource found in amplify config, run 'amplify add geo' to create one and ensure to run `amplify push` after";
```
Amplifiyer, line 99: Shouldn't this be a different error message. If I recall correctly, if there is at least one map added, there has to be a default one.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -63,4 +64,44 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+			const style = availableMaps[mapName].style;
+			mapStyles.push({ mapName, style });
+		}
+
+		return mapStyles;
+	}
+
+	/**
+	 * Get the map resource set as default in amplify config
+	 * @param {string} provider
```
Amplifiyer, line 91: Don't see parameter in the function signature.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -63,4 +64,44 @@ export class AmazonLocationServicesProvider implements GeoProvider {
 		this._config = Object.assign({}, this._config, config);
 		return this._config;
 	}
+
+	/**
+	 * Get the map resources that are currently available through the provider
+	 * @param {string} provider
+	 * @returns - Array of available map resources
+	 */
+	public getAvailableMaps(): MapStyle[] | string {
```
Amplifiyer, line 73: Don't see parameter in the function signature.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
Amplifiyer, line 99: These should be thrown not `string` returns

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -36,17 +33,21 @@ export class AmazonLocationServicesProvider implements GeoProvider {
 		this._config = config ? config : {};
 		logger.debug('Geo Options', this._config);
 
+		this.getAvailableMaps.bind(this);
+		this.getDefaultMap.bind(this);
```
Amplifiyer, line 37: Probably not needed

packages/geo/src/Geo.ts
```diff
@@ -100,6 +104,42 @@ export class GeoClass {
...
+	 * @returns - Map resource set as the default in amplify config
+	 */
+	public getDefaultMap(provider = DEFAULT_PROVIDER): MapStyle | string {
+		const prov = this._pluggables.find(
+			pluggable => pluggable.getProviderName() === provider
+		);
+
+		if (prov === undefined) {
+			logger.debug('No plugin found with providerName', provider);
+			return 'No plugin found in Geo for the provider';
```
Amplifiyer, line 138: Should be a throw

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: Correct, JSDoc on that was wrong

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: Likewise, bad JSDoc.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: I was trying to think of a better call-to-action here for what to do in this instance. Will have to think about it a little more. If you have a thought, I'd be happy to hear it!

TreTuna: Agreed

packages/geo/src/Geo.ts
TreTuna: agreed

packages/core/src/Parser.ts
```diff
@@ -58,6 +58,11 @@ export const parseMobileHubConfig = (config): AmplifyConfig => {
 		amplifyConfig.Geo = Object.assign({}, config.geo);
 	}
 
+	// Geo
```
thaddmt, line 61: Is this block added twice? I see it on line 56 as well

packages/geo/src/Geo.ts
```diff
@@ -100,6 +104,42 @@ export class GeoClass {
...
+
+	/**
+	 * Get the map resources that are currently available through the provider
+	 * @param {string} provider
+	 * @returns - Array of available map resources
+	 */
+	public getAvailableMaps(provider = DEFAULT_PROVIDER): MapStyle[] | string {
+		const prov = this._pluggables.find(
+			pluggable => pluggable.getProviderName() === provider
+		);
```
thaddmt, line 116: Nit: Seems like we can probably make this a utility function since it's also used in `getDefaultMap`

packages/core/src/Parser.ts
TreTuna: Yeah, it is, I have it removed in the final PR of this series. Think it happened in one of the rebases I've had to do on this.

TreTuna: https://github.com/aws-amplify/amplify-js/pull/8639/files#diff-481fe78bb7c6c496f7fb87dba98e6b9ff3521d6b9fafb4e8b0834f83c134ae05R58

packages/geo/src/Geo.ts
TreTuna: I can do that and tag it to the final PR. Looks like I actually have it five times overall in there when all is said in done, so good call on this.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -63,4 +64,42 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+	public getDefaultMap(): MapStyle {
+		if (!this._config.maps) {
+			throw "No map resources found in amplify config, run 'amplify add geo' to create them and ensure to run `amplify push` after";
+		}
+		if (!this._config.maps.default) {
+			throw "No default map resource found in amplify config, run 'amplify add geo' to create one and ensure to run `amplify push` after";
+		}
+
+		const mapName = this._config.maps.default;
+		const style = this._config.maps.items[mapName].style;
```
hvergara, line 101: how are we sure about the availability of these nested properties? e.g can we trust that `mapName` is always available or if exists within the `this._config.maps.items` dictionary?

Maybe using something like `immutable.get` could reduce the risk of broken references.

TreTuna: if `this._config.maps` exists, then yes, there should be a `.default` and `.maps.items` as that is the contract with the Amplify Geo CLI configuration data. If it doesn't, then that points to a problem with the CLI or the `aws-exports.js` configuration.

packages/geo/src/Geo.ts
TreTuna: Done in https://github.com/aws-amplify/amplify-js/pull/8639 
https://github.com/aws-amplify/amplify-js/pull/8639/commits/9c4672c6250c6541d42c4bd4b2324bbd37e2c92a

thaddmt (approved):

eddiekeller (approved):

wlee221 (approved):
LGTM 🚢

Milan-Shah (approved):

packages/geo/src/types/Provider.ts
```diff
@@ -11,11 +11,21 @@
...
 	getProviderName(): string;
 
+	// configure your provider
 	configure(config: object): object;
+
+	// get the available map resources
+	getAvailableMaps(): string | MapStyle[];
+
+	// get the map resource listed as default
+	getDefaultMap(): string | MapStyle;
```
manueliglesias, line 30: I know this has been merged already, but, under which circumstances is a `string` returned? I didn't see a test case for that scenario

TreTuna: That was left over from a previous iteration of this function. It is removed later in this series of PRs.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
hvergara: great, thanks for the detailed response

PR #8621, opened 2021-07-21
feat(Geo): searchByText - Geo milestone 1 - PR 5

TreTuna:
#### Description of changes

Geo milestone 1 - PR 5/6
- Adds `searchByText` API

- Builds and unit tests pass
- Validated using PoC application
image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
*Documentation will be added in a follow-up PR*


packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -92,4 +105,57 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+		options?: SearchByTextOptions
+	): Promise<Place[]> {
+		const credentialsOK = await this._getCredentials();
+		if (!credentialsOK) {
+			return Promise.reject('No credentials');
+		}
+
+		let searchInput = {
+			Text: text,
+			IndexName: this._config.place_indexes.default,
```
Amplifiyer, line 120: index name is optionally present in `options` as well and should be given priority (there should be a unit test for this)

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -92,4 +105,57 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+		};
+		if (options) {
+			const PascalOptions = convertPlaceCamelToPascal(options);
+			searchInput = {
+				...searchInput,
+				...PascalOptions,
+			};
+		}
```
Amplifiyer, line 128: It's not exactly clear what's the need for case conversion is on the options. I guess for Countries? If yes, is that the recommended option for _all the countries_ in the world? And does all the entries in the options need to go through this conversion?

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -92,4 +105,57 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+		const PascalResults = response.Results.map(result => result.Place);
+		const results = convertPlaceArrayPascalToCamel(PascalResults);
+
+		return results;
+	}
+
+	/**
+	 * @private
+	 */
+	private _getCredentials() {
```
Amplifiyer, line 146: nit: `get...` denotes that it is returning the credentials which is not the case.

You can probably rewrite it much better by using async/await in try/catch and actually returning the creds

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -92,4 +105,57 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+			};
+		}
+		const client = new LocationClient({
+			credentials: this._config.credentials,
+			region: this._config.region,
+		});
+		const command = new SearchPlaceIndexForTextCommand(searchInput);
+
+		const response = await client.send(command);
+
```
Amplifiyer, line 136: No error handling?

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -92,4 +105,57 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+		const PascalResults = response.Results.map(result => result.Place);
+		const results = convertPlaceArrayPascalToCamel(PascalResults);
```
Amplifiyer, line 138: what's the requirement for this? Also should add some inline comments to explain since it's not immediately understood the reason for this.

packages/geo/src/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+	PascalPlace.IndexName = PascalPlace.PlaceIndex;
+	delete PascalPlace.PlaceIndex;
+	return PascalPlace;
+}
+
+export function convertPlaceArrayPascalToCamel(placeArray) {
+	const camelPlace = camelcaseKeys(placeArray, { deep: true });
+	delete camelPlace.indexName;
+	return camelPlace;
+}
```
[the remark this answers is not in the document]
TreTuna: Yeah, that makes sense. Normally I make a utils to start and then break them out, just forgot in this instance.

packages/geo/src/types/Provider.ts
```diff
@@ -25,4 +28,9 @@ export interface GeoProvider {
 	getAvailableMaps(): string | MapStyle[];
 
 	getDefaultMap(): string | MapStyle;
+
+	searchByText(
+		text: string,
```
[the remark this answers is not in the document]
TreTuna: 👍

packages/geo/src/types/Provider.ts
```diff
@@ -25,4 +28,9 @@ export interface GeoProvider {
...
+		options?: SearchByTextOptions
+	): Promise<Place[] | void>;
```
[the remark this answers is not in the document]
TreTuna: That was a leftover type during implementation and can be removed.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: I need to rename the casing functions better. We are converting the case of the `keys` of the searchInput, not the `values`. All of our options have camelCased keys and Amazon Location Services wants PascalCase.

packages/geo/src/types/Geo.ts
```diff
@@ -10,6 +10,7 @@
  * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
  * and limitations under the License.
  */
+import { GeoProvider } from './Provider';
```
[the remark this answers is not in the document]
TreTuna: It's actually miss used and a good catch here. Where this is used below, it should be just a string for the provider name that the user wants to use instead of the default.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: Here we pull the default `place_index` and set it in the `serachInput`. Then, in the `if` block below, we convert out options object into the format wanted by Location Services and spread that out on the `searchInput`. With that, if a new `placeIndex` is given in the `options`, then it overwrites this default.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: Added a comment about this.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: Handled now

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: Added

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: This was mostly modeled off of the _ensureCredentials function from AWSS3Provider in the Storage category. To keep with that same functionality, I renamed this to the same as there `_ensureCredentials`. The credentials are set to `this._config`, so we don't need them returned, but need to verify they are good.

packages/geo/src/types/Geo.ts
```diff
@@ -24,3 +25,37 @@ export type MapStyle = {
...
+
+export type SWLatitude = Latitude;
+export type SWLongitude = Longitude;
+export type NELatitude = Latitude;
+export type NELongitude = Longitude;
+export type BoundingBox = [SWLatitude, SWLongitude, NELatitude, NELongitude];
+
+export type SearchByTextOptions = {
+	biasPosition?: Coordinates;
+	searchAreaConstraints?: BoundingBox;
```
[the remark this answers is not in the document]
TreTuna: There's ultimately nothing blocking someone from passing them both in at the same time. This actually may be something we want to test out more. I can see use case for passing both. Think about having a bounding box around Seattle, but you still want your search to be centered on your location inside that bounding box.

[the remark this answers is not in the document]
TreTuna: After reviewing things, you are correct. I had missed that part in their docs. I'll adjust this with a union type to make that clearer.

packages/geo/src/utils.ts
```diff
@@ -0,0 +1,15 @@
+import camelcaseKeys from 'camelcase-keys';
+import pascalcaseKeys from 'pascalcase-keys';
```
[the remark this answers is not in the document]
TreTuna: `camelcase-keys` is already in our dependency tree as a sub-dependency on something and with the way I just rewrote the data conversion functions I was able to remove `pascalcase-keys`.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -102,4 +116,95 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+		options?: SearchByTextOptions
+	): Promise<Place[]> {
+		try {
+			const credentialsOK = await this._ensureCredentials();
+			if (!credentialsOK) {
+				return Promise.reject('No credentials');
+			}
+		} catch (error) {
+			throw '';
+		}
```
hvergara, line 137: I suggest this simpler way to handle this: 
```suggestion
		const credentialsOK = await this._ensureCredentials();

		if (!credentialsOK) {
			throw 'No credentials';
		}
```

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -102,4 +116,95 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+
+			if (options.placeIndexName) {
+				locationServicesInput.IndexName = options.placeIndexName;
+			}
+
+			if (options['biasPosition']) {
+				locationServicesInput.BiasPosition = options['biasPosition'];
+			} else if (options['searchAreaConstraints']) {
+				locationServicesInput.FilterBBox = options['searchAreaConstraints'];
+			}
```
hvergara, line 162: can we use the same way to access properties as above? e.g
```suggestion
			if (options.biasPosition) {
				locationServicesInput.BiasPosition = options.biasPosition;
			} else if (options.searchAreaConstraints) {
				locationServicesInput.FilterBBox = options.searchAreaConstraints;
			}
```

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -102,4 +116,95 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+			region: this._config.region,
+		});
+		const command = new SearchPlaceIndexForTextCommand(locationServicesInput);
+
+		let response;
+		try {
+			response = await client.send(command);
+		} catch (error) {
+			logger.debug(error);
+			throw new Error(error);
```
hvergara, line 176: just curious: why do we need to create a *new* error?

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -102,4 +116,95 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+		 * The response from Location Services is a "Results" array of objects with a single `Place` item,
+		 * which are Place objects in PascalCase.
+		 * Here we want to flatten that to an array of results and change them to camelCase
+		 */
+		const PascalResults: PlaceResult[] = response.Results.map(
+			result => result.Place
+		);
+		const results: Place[] = (camelcaseKeys(PascalResults, {
+			deep: true,
+		}) as undefined) as Place[];
```
hvergara, line 189: can you explain briefly why we need to perform all these casts?

TreTuna: This is mainly due to going from PascalCase data keys returned from Location Services to camelCased data keys that we intend on returning. If you try to cast directly, TypeScript throws an error as there is not sufficient overlap between the two types, thus you cast to undefined first, then to the type we want to return in order to do the conversion.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
Amplifiyer: Sounds good, is there a unit test for this?

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: This was what I had to use due using a union type here for the `options`.

If I use the `object.key` syntax here, it throws issues saying:
```
Property 'biasPosition' does not exist on type 'SearchByTextOptions'.
  Property 'biasPosition' does not exist on type 'SearchByTextOptionsWithSearchAreaConstraints'.
```
```
Property 'searchAreaConstraints' does not exist on type 'SearchByTextOptions'.
  Property 'searchAreaConstraints' does not exist on type 'SearchByTextOptionsWithBiasPosition'.
```

However, using the `object["key"]` syntax, it does not throw.

packages/geo/src/Geo.ts
```diff
@@ -140,6 +144,20 @@ export class GeoClass {
...
+		const prov = this._pluggables.find(
+			pluggable => pluggable.getProviderName() === provider
+		);
+		if (prov === undefined) {
+			logger.debug('No plugin found with providerName', provider);
+			return Promise.reject('No plugin found in Geo for the provider');
+		}
+
+		const responsePromise = await prov.searchByText(text, options);
+		return responsePromise;
```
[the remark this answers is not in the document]
TreTuna: Yeah, I actually changed to doing this in the config refactor PR in this series already:
https://github.com/aws-amplify/amplify-js/pull/8639/files#diff-73388651ffd6ac5a23aa315c72af9908460cad862c31cfa669fe975074e4b26bR143

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: Good point. I can just change this to `throw error`. Updated this in the final PR of this series:
https://github.com/aws-amplify/amplify-js/pull/8639/commits/6814e0cfe6e3c68fd67955374b95e9d5fdb2c4c9

packages/geo/src/Geo.ts
```diff
@@ -140,6 +144,20 @@ export class GeoClass {
...
 	}
+
+	public async searchByText(text: string, options?: SearchByTextOptions) {
+		const { provider = DEFAULT_PROVIDER } = options || {};
+		const prov = this._pluggables.find(
+			pluggable => pluggable.getProviderName() === provider
+		);
+		if (prov === undefined) {
+			logger.debug('No plugin found with providerName', provider);
+			return Promise.reject('No plugin found in Geo for the provider');
```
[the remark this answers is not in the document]
TreTuna: This is updated in the new config PR with a combination of two commits:
Moved these to use reusable getPluggable method: https://github.com/aws-amplify/amplify-js/pull/8639/commits/9c4672c6250c6541d42c4bd4b2324bbd37e2c92a

Updates to throw new Error: https://github.com/aws-amplify/amplify-js/pull/8639/commits/6814e0cfe6e3c68fd67955374b95e9d5fdb2c4c9

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
```diff
@@ -102,4 +116,95 @@ export class AmazonLocationServicesProvider implements GeoProvider {
...
+	 */
+	public async searchByText(
+		text: string,
+		options?: SearchByTextOptions
+	): Promise<Place[]> {
+		try {
+			const credentialsOK = await this._ensureCredentials();
+			if (!credentialsOK) {
+				return Promise.reject('No credentials');
+			}
```
wlee221, line 134: NIT / not blocking: Looks like we have double logic determining whether credential is invalid:

1. `_ensureCredentials` resolves to false
2. `_ensureCredentials` throws and is caught by throw/catch statements

Can we combine this to a single logic? One idea I have is to refactor `_ensureCredentials` so that it returns `Promise<void>` and not `Promise<boolean>` so that we only have to worry about whether it resolves or not.

Milan-Shah (approved):
LGTM!

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
wlee221: Can we extend `SearchByTextOptions` to accept those properties?

packages/geo/src/types/Geo.ts
```diff
@@ -10,17 +10,79 @@
...
+// Geometry object for Place points
+export type PlaceGeometry = {
+	point: Coordinates;
 };
 
+// Place object with locality information
+export interface Place {
+	addressNumber?: string;
+	country?: string;
+	geometry: PlaceGeometry | undefined;
```
wlee221, line 80: Why 
```ts
geometry: PlaceGeometry | undefined
```

instead of
```ts
geometry?: PlaceGeometry
```

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
wlee221: I'm not sure about `undefined` here though as it suggests this variable is empty, maybe this can be casted to `any` / `Record<PropertyKey, any>` to better convey its type?

wlee221 (commented):
LGTM, just some typing nits but nothing blocking.

manueliglesias (approved):
I am not familiar enough with the internals of this new category but I've looked at the TypeScript related changes and I don't see any issues with that aspect. 

LGTM

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: Yeah, that makes sense, I'll update to cast to `any` instead of `undefined` in Geo PR #7

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: @wlee221 and I spoke offline about this and we're going to leave it like this for now and possibly update how this is done later.

packages/geo/src/types/Geo.ts
TreTuna: This `Place` type really came directly from `@aws-sdk/client-location`'s [Place type](https://github.com/aws/aws-sdk-js-v3/blob/main/clients/client-location/models/models_0.ts#L4530), which looks like it's a generated and this may be an anomaly from that generation. I copy/pasted their interface and just changed all the keys to camelCase instead of PascalCase. We should be able to update this to use `geometry?: PlaceGeometry`. I'll give it a go and update it in PR #7 if that's the case.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: Yes, there are two tests that pass a custom `placeIndexName` in that cover this.

packages/geo/src/types/Geo.ts
```diff
@@ -24,3 +25,37 @@ export type MapStyle = {
...
+};
+
+export type Place = {
+	addressNumber: string;
+	country: string;
+	geometry: {
+		point: Coordinates;
+	};
+	label: string;
+	municipality: string;
```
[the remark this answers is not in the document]
TreTuna: It's not and was updated to be optional.

packages/geo/src/Providers/AmazonLocationServicesProvider.ts
TreTuna: Yeah, I noticed this same thing and have a note on my desk to revisit the `_ensureCredentials` flow. Mostly, I followed how other Categories such as `storage`, `core`, and `pubsub` are the other categories with `_ensureCredentials`, which is where I grabbed this functionality from. `analytics` does the same thing as well, just with a `_getCredentials` vs `_ensureCredentials`.

wlee221 (approved):

Issue #8398, opened 2021-06-05
DataStore: Cannot create record with no value for optional @connection field


james-braund:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication, DataStore

### Amplify Categories

auth, function, api

### Environment information

<details>

```
# Put output below this line

System:
    OS: Linux 4.19 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i5-4670K CPU @ 3.40GHz
    Memory: 21.51 GB / 25.02 GB
    Container: Yes
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.15.0 - ~/.nvm/versions/node/v14.15.0/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.15.0/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v14.15.0/bin/npm
  Browsers:
    Chrome: 88.0.4324.96
  npmPackages:
    @aws-amplify/ui-react: ^1.1.0 => 1.1.0 
    @aws-sdk/client-s3: ^3.15.0 => 3.15.0 (3.6.1)
    @aws-sdk/lib-storage: ^3.15.0 => 3.15.0 
    @electron-forge/cli: ^6.0.0-beta.54 => 6.0.0-beta.54 
    @electron-forge/maker-deb: ^6.0.0-beta.54 => 6.0.0-beta.54 
    @electron-forge/maker-rpm: ^6.0.0-beta.54 => 6.0.0-beta.54 
    @electron-forge/maker-squirrel: ^6.0.0-beta.54 => 6.0.0-beta.54 
    @electron-forge/maker-zip: ^6.0.0-beta.54 => 6.0.0-beta.54 
    @electron-forge/plugin-webpack: 6.0.0-beta.54 => 6.0.0-beta.54 
[...]

PR #8623, opened 2021-07-21
fix(@aws-amplify/datastore): remove null properties from connected model instances

nickarocho:
#### Description of changes

Per [this issue](https://github.com/aws-amplify/amplify-js/issues/8398), if an app using Datastore has a schema with optional `@connection` directives between two models - currently the `traverse` method it Datastore's `util.ts` file explicitly sets the value of the model's `@key` field to `null`.

This causes problems because DynamoDB can't reference the connection successfully. The record is saved fine in the IndexedDB locally, but the GraphQL call to the server ultimately fails because of the `null` value.

This fix simply converts the ternary assignment to a conditional which checks for a truthy value first before adding the property to the connected model's instance within the `IndexedDBAdapter`.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/8398

Without the fix, we can see a warning returned from DynamoDB and can also see that the expected record is not saved in the backend DB:
Screen Shot 2021-07-21 at 1 07 32 PM

WITH the fix, there is no warning, and the record is successfully saved in the backend DB:
Screen Shot 2021-07-21 at 1 03 03 PM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/src/util.ts
+++ b/packages/datastore/src/util.ts
@@ -361,13 +361,12 @@ export const traverseModel = <T extends PersistentModel>(
 						}
 					}
 
-					(<any>draftInstance)[rItem.targetName] = draftInstance[
-						rItem.fieldName
-					]
-						? (<PersistentModel>draftInstance[rItem.fieldName]).id
-						: null;
-
-					delete draftInstance[rItem.fieldName];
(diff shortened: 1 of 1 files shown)


iartemiev (approved):
LGTM!

PR #8639, opened 2021-07-26
chore(geo): update configs for newer scoped configurations and providers - Geo milestone 1 - PR 7

TreTuna:
#### Description of changes

Update config methods to be more agnostic and scope configs to providers. The CLI's exported config to `aws-exports` was updated to nest `amazon_location_services` config under the `geo` key (before it was not scoped and just under `geo`). This updates Geo's config to handle that scoping. Also, updates to allow for more scoped provider configurations in the future, making it more agnostic to the provider.

Updated tests and ran this through my example application

image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
_Documentation PRs are next in line_

### DIFF
--- a/packages/core/src/Parser.ts
+++ b/packages/core/src/Parser.ts
@@ -56,21 +56,11 @@ export const parseMobileHubConfig = (config): AmplifyConfig => {
 	// Geo
 	if (config['geo']) {
 		amplifyConfig.Geo = Object.assign({}, config.geo);
-	}
-
-	// Geo
-	if (config['geo']) {
-		amplifyConfig.Geo = Object.assign({}, config.geo);
-	}
-
-	// Geo
-	if (config['geo']) {
-		amplifyConfig.Geo = Object.assign({}, config.geo);
-	}
-
-	// Geo
-	if (config['geo']) {
-		amplifyConfig.Geo = Object.assign({}, config.geo);
+		if (config.geo['amazon_location_services']) {
+			amplifyConfig.Geo = {
+				AmazonLocationServices: config.geo['amazon_location_services'],
+			};
+		}
 	}
 
 	amplifyConfig.Analytics = Object.assign(
--- a/packages/geo/__tests__/Geo.test.ts
+++ b/packages/geo/__tests__/Geo.test.ts
@@ -79,14 +79,16 @@ describe('Geo', () => {
 			geo.addPluggable(provider);
 			geo.removePluggable(provider.getProviderName());
 
-			expect(geo.getPluggable(provider.getProviderName())).toBeNull();
+			expect(() => geo.getPluggable(provider.getProviderName())).toThrow(
+				'No plugin found in Geo for the provider'
+			);
 		});
 	});
 
 	describe('AmazonLocationServices is used as default provider', () => {
-		test('configure with aws-exports file', () => {
+		test('creates the proper default provider', () => {
 			const geo = new GeoClass();
-			const config = geo.configure(awsConfig);
+			geo.configure(awsConfig);
 			expect(geo.getPluggable('AmazonLocationServices')).toBeInstanceOf(
 				AmazonLocationServicesProvider
 			);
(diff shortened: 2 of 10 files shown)


packages/geo/src/Geo.ts
```diff
@@ -79,7 +76,7 @@ export class GeoClass {
 		);
 		if (pluggable === undefined) {
 			logger.debug('No plugin found with providerName', providerName);
-			return null;
+			throw 'No plugin found in Geo for the provider';
```
jamesaucode, line 79: I prefer if we throw a `new Error` instead of just string

packages/geo/src/Geo.ts
```diff
@@ -141,47 +131,34 @@ export class GeoClass {
...
-			return Promise.reject('No plugin found in Geo for the provider');
-		}
+		const prov = this.getPluggable(providerName);
 
-		const responsePromise = await prov.searchByText(text, options);
-		return responsePromise;
+		try {
+			return await prov.searchByText(text, options);
+		} catch (error) {
+			throw new Error(error);
```
jamesaucode, line 146: Why do we catch the errors if we just throw them back immediately?

jamesaucode (commented):
Looks good! Just a couple nit picks

packages/core/src/Parser.ts
```diff
@@ -55,12 +55,9 @@ export const parseMobileHubConfig = (config): AmplifyConfig => {
...
 	if (config['geo']) {
-		amplifyConfig.Geo = Object.assign({}, config.geo);
-	}
-
-	// Geo
-	if (config['geo']) {
-		amplifyConfig.Geo = Object.assign({}, config.geo);
+		amplifyConfig.Geo = {
+			AmazonLocationServices: config.geo['amazon_location_services'],
+		};
```
Amplifiyer, line 60: This should be part of the Geo category implementation. i.e. Geo category when added a provider names `AmazonLocationServices` should send the config for `amazon_location_services` to it. Parser shouldn't have provider information.

packages/geo/__tests__/Geo.test.ts
```diff
@@ -133,7 +150,7 @@ describe('Geo', () => {
 			geo.configure(awsConfig);
 
 			const maps = [];
-			const availableMaps = awsConfig.geo.maps.items;
+			const availableMaps = awsConfig.geo.amazon_location_services.maps.items;
 			for (const mapName in availableMaps) {
```
Amplifiyer, line 154: Now with proper segmentation of Category and Providers, the tests should be updated as such. e.g. Geo category tests shouldn't rely on `amazon_location_services` and should just mock the provider implementation. Provider tests should test the implementation (parsing the config etc.)

Amplifiyer (commented):
Just couple of structural nits. Non blocking.

packages/geo/src/Geo.ts
TreTuna: Because I forgot to add the `logger.debug(error)` statement that should also be here. 🤦

packages/geo/src/Geo.ts
TreTuna: Yeah, makes sense. 👍

packages/core/src/Parser.ts
TreTuna: Spoke with Praveen offline about this. We are going to keep this as is for now and update it after the initial developer preview release.

packages/geo/__tests__/Geo.test.ts
TreTuna: Spoke with Praveen offline about this. We are going to keep this as is for now and update it after the initial developer preview release.

packages/geo/src/Geo.ts
```diff
@@ -41,11 +41,6 @@ export class GeoClass {
 		this._config = {};
 		this._pluggables = [];
 		logger.debug('Geo Options', this._config);
-
-		this.getAvailableMaps.bind(this);
-		this.getDefaultMap.bind(this);
-		this.searchByText.bind(this);
-		this.searchByCoordinates.bind(this);
 	}
```
wlee221, line 44: Why were these removed?

wlee221 (approved):
LGTM, just got one question.

packages/geo/src/Geo.ts
TreTuna: They were unneeded. Tested it without them and everything works fine.

Issue #8673, opened 2021-07-30
react-native hard crash on Auth.signOut() when using federated identities


andreialecu:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

Authentication

### Amplify Categories

auth

### Environment information

<details>
 irrelevant
</details>

### Describe the bug

Using the documentation from the section in "In-App Browser Setup (optional, but recommended)" at: https://docs.amplify.aws/lib/auth/social/q/platform/js#full-samples results in a crash when signing out when authenticated with a federated identity (apple sign in):

```
Exception '*** -[NSURL initWithString:relativeToURL:]: nil string parameter' was thrown while invoking openAuth on target RNInAppBrowser with params (
    "https://....auth.eu-west-2.amazoncognito.com/logout?client_id=49jjijjs1g3lle6of82imj25j0&logout_uri=zzz%3A%2F%2FsignOut",
    "<null>",
        {
        enableDefaultShare = 0;
        enableUrlBarHiding = 1;
        ephemeralWebSession = 0;
        forceCloseOnRedirection = 0;
        showTitle = 0;
    },
    17634,
    17635
)
callstack: (
	0   CoreFoundation                      0x00000001803f7978 __exceptionPreprocess + 236
	1   libobjc.A.dylib                     0x0000000180188800 objc_exception_throw + 56
	2   Foundation                          0x000000018083ac2c -[NSURL(NSURL) _cfurl] + 0
	3   zzz                               0x0000000100c72bc8 -[RNInAppBrowser openAuth:redirectURL:options:resolver:rejecter:] + 480
	4   CoreFoundation                      0x00000001803fe030 __invoking___ + 144
	5   CoreFoundation                      0x00000001803fb384 -[NSInvocation invoke] + 300
	6   CoreFoundation                      0x00000001803fb644 -[NSInvocation invokeWithTarget:] + 76
[...]

PR #8713, opened 2021-08-05
fix(@aws-amplify/auth): passing redirectUri to the urlOpener after signOut 

Samaritan1011001:
#### Description of changes
It passes in the redirectUri after sign out which is in parity with the urlOpener signature when sign in. 
The difference in parameters are evident between these two lines: 

Sign in, https://github.com/aws-amplify/amplify-js/blob/1e0887bdc45e178f04229c888051ecf10d2bfabc/packages/auth/src/OAuth/OAuth.ts#L117
Sign out, https://github.com/aws-amplify/amplify-js/blob/1e0887bdc45e178f04229c888051ecf10d2bfabc/packages/auth/src/OAuth/OAuth.ts#L294 

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/8673

Tested it on ios simulator 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/auth/src/OAuth/OAuth.ts
+++ b/packages/auth/src/OAuth/OAuth.ts
@@ -291,7 +291,7 @@ export default class OAuth {
 		);
 		logger.debug(`Signing out from ${oAuthLogoutEndpoint}`);
 
-		return this._urlOpener(oAuthLogoutEndpoint);
+		return this._urlOpener(oAuthLogoutEndpoint, signout_uri);
 	}
 
 	private _generateState(length: number) {


evcodes (approved):
Thank you for your contribution. 

This looks good to me 🌮

packages/auth/src/OAuth/OAuth.ts
```diff
@@ -291,7 +291,7 @@ export default class OAuth {
 		);
 		logger.debug(`Signing out from ${oAuthLogoutEndpoint}`);
 
-		return this._urlOpener(oAuthLogoutEndpoint);
+		return this._urlOpener(oAuthLogoutEndpoint, signout_uri);
```
calebpollman, line 294: Are there any side effects of passing `signout_uri` here?

evcodes: No because, for this ```_urlOpener```, we are getting it from the config file as highlighted by the return types:

```constructor({
		config,
		cognitoClientId,
		scopes = [],
	}: {
		scopes: string[];
		config: OAuthOpts;
		cognitoClientId: string;
	}) {
		this._urlOpener = config.urlOpener || launchUri;
```

Where the config.urlOpener takes in a string a returnURL: 
```urlOpener?: (url: string, redirectUrl: string) => Promise<any>;```

I believe that the app was crashing because the redirectUrl had not been set like in the login case.

Issue #8649, opened 2021-07-27
DataStore Update attempting to write over its own read-only property


RobertGemmaJr:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication, DataStore

### Amplify Categories

api

### Environment information

<details>

```
# Put output below this line
  System:
    OS: Windows 10 10.0.19043
    CPU: (16) x64 Intel(R) Core(TM) i7-10875H CPU @ 2.30GHz
    Memory: 4.42 GB / 15.75 GB
  Binaries:
    Node: 14.15.4 - C:\Program Files\nodejs\node.EXE
    npm: 6.14.9 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 92.0.4515.107
    Edge: Spartan (44.19041.1023.0), Chromium (92.0.902.55)
    Internet Explorer: 11.0.19041.1
  npmPackages:
    @aws-amplify/ui-react: ^1.2.6 => 1.2.6
    @material-ui/core: ^4.11.4 => 4.11.4
    @material-ui/data-grid: ^4.0.0-alpha.32 => 4.0.0-alpha.32
    @material-ui/icons: ^4.11.2 => 4.11.2
    aws-amplify: ^4.2.0 => 4.2.0
    cross-env: ^7.0.3 => 7.0.3
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
    react-scripts: 4.0.3 => 4.0.3
    web-vitals: ^1.1.2 => 1.1.2
  npmGlobalPackages:
    @aws-amplify/cli: 5.1.2
    nodemon: 2.0.4
    npm: 6.14.9
    rimraf: 3.0.2


```

</details>

### Describe the bug

Error: createdAt is read-only
DataStore.save is attempting to update a read-only property (createdAt) of my Setting model.

[...]

PR #8794, opened 2021-08-23
fix(@aws-amplify/datastore): check read-only at instance level

iartemiev:
#### Description of changes

* Check read-only fields at the instance level, instead of at `DataStore.save`
* Resolves bug where error was being thrown during .copyOf call without modifying read-only fields

#### Issue #, if available
#8649 

* unit tests
* sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/src/datastore/datastore.ts
```diff
@@ -374,13 +373,18 @@ const createModelClass = <T extends PersistentModel>(
...
 
-					const id =
-						// instancesIds is set by modelInstanceCreator, it is accessible only internally
-						_id !== null && _id !== undefined
-							? _id
-							: modelDefinition.syncable
-							? uuid4()
-							: ulid();
+					// instancesIds are set by modelInstanceCreator, it is accessible only internally
+					const isInternal = _id !== null && _id !== undefined;
```
nickarocho, line 377: Thanks for storing this value in a defined variable, makes the following code much more readable upon first glance.

nickarocho (approved):
This is so much better in every way. Thanks, Ivan! LGTM 👍

packages/datastore/src/datastore/datastore.ts
nickarocho: I see you also needed it stored to reference later, but still, readability is a great added benefit :)

ArkamJ (approved):

PR #8799, opened 2021-08-25
feat(@aws-amplify/api-graphql) Add support for string 'authmode' values in TypeScript based apps 

Khairo-kh:
#### Description of changes
In TypeScript based projects, GraphQL operations that specify `authmode` does not accept string values (example: "API_KEY")and users have to import `GRAPHQL_AUTH_MODE` which is defined as an enum and use it to specify the value of `authmode`.

I added all valid authmode values (`'API_KEY'  'AWS_IAM'  'OPENID_CONNECT'  'AMAZON_COGNITO_USER_POOLS' 'AWS_LAMBDA'`) as accepted string values for type `authmode` 

- Tested with local app (yarn linking)
- yarn test passed

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/api-graphql/src/types/index.ts
+++ b/packages/api-graphql/src/types/index.ts
@@ -14,10 +14,12 @@
 import { GraphQLError } from 'graphql/error/GraphQLError';
 import { DocumentNode } from 'graphql/language/ast';
 
+type GRAPHQL_STR_MODE = 'API_KEY' | 'AWS_IAM' | 'OPENID_CONNECT' | 'AMAZON_COGNITO_USER_POOLS' | 'AWS_LAMBDA';
+
 export interface GraphQLOptions {
 	query: string | DocumentNode;
 	variables?: object;
-	authMode?: GRAPHQL_AUTH_MODE;
+	authMode?: GRAPHQL_AUTH_MODE | GRAPHQL_STR_MODE;
 	authToken?: string;
 }
 


packages/api-graphql/src/types/index.ts
```diff
@@ -14,10 +14,12 @@
 import { GraphQLError } from 'graphql/error/GraphQLError';
 import { DocumentNode } from 'graphql/language/ast';
 
+type GRAPHQL_STR_MODE = 'API_KEY' | 'AWS_IAM' | 'OPENID_CONNECT' | 'AMAZON_COGNITO_USER_POOLS' | 'AWS_LAMBDA';
+
 export interface GraphQLOptions {
 	query: string | DocumentNode;
 	variables?: object;
-	authMode?: GRAPHQL_AUTH_MODE;
+	authMode?: GRAPHQL_AUTH_MODE | GRAPHQL_STR_MODE;
```
iartemiev, line 22: No need to create a new type. You can accomplish the same thing with `keyof typeof` for example:
```suggestion
	authMode?: GRAPHQL_AUTH_MODE | keyof typeof GRAPHQL_AUTH_MODE;
```

iartemiev (changes requested):

packages/api-graphql/src/types/index.ts
evcodes: Awesome, thank you Ivan! :)

Khairo-kh: Thank you for the feedback! I will make the change.

PR #8825, opened 2021-08-30
fix: Update .d.ts file to match the actual code

MatrixFrog:
#### Description of changes

A couple of functions take an optional `clientMetadata` parameter which is not documented in this .d.ts. This PR adds that parameter.
 
#### Description of how you validated changes

Tests pass.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #8826, opened 2021-08-30
Avoid 'any' type in a few places, and fix resulting type errors

MatrixFrog:
Relies on #8825 

#### Description of changes

Change from `CognitoUser | any` to `CognitoUser` in several places, since `CognitoUser | any` just evaluates to `any`.

I'm not sure what I should do to validate this other than making sure tests pass (which they don't as of this writing, but they should after #8825 is merged).

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


evcodes:
This likely won't work because some responses from the Cognito package either return a cognito user or some error message.

MatrixFrog:
Then would `CognitoUser | string` be correct perhaps? Or `CognitoUser | Error`?

Issue #7667, opened 2021-02-02
Storage.get should have conditional return type


avlonder:
**Describe the bug**
Storage.get currently returns a *Promise<string|Object>*, where the object case only occurs for *config.download = true*.
As such, this would benefit from conditional typing, as it now complicates the integration in a larger typescript code base.

**Additional context**
Frankly, I just started refactoring my JS code to TS code, so I'm very new to TS, and maybe I'm overlooking some stuff.

sammartinez:
Thank you for this feedback @avlonder. I will mark this as a feature request and we will take this into consideration for the future of the library.

jamesaucode:
The latest Typescript improvement for Storage has been merged and released. That includes conditional return type 😄

PR #8832, opened 2021-08-31
Storage category public API Typescript improvement

jamesaucode:
#### Description of changes
This patch will introduce Typescript support for Storage.
instead of having `any` type config and `any` or `object` return types, all APIs will have proper types, including conditional return type on `get`.
Custom provider use case will still be supported.

Example of using the default provider:
```ts
// no changes, defaults to using our AWSS3Provider implementation
const res = await storage.get('key', { download: false }); // string

const res2 = await storage.get('key', { download: true }); // GetObjectCommandOutput

// Typescript or your editor's LSP will be able to show you warning for incorrect usage now
await storage.get('key', { invalidConfig: true }); // Error: Argument of 'invalidConfig' is not assignable ....
```

Example of using custom providers:
```ts
class CustomProvider implements StorageProvider {
  getProviderName() {
    return 'customProvider' as const
  }
  get(key: string, config: { foo: boolean; bar: number; }) {
    return 'get';
  }
  ...
}

storage.get<CustomProvider>('key', {
  provider: 'customProvider',
  foo: true,
  bar: 20,
});
```

For backwards compatibility, the following call will still work for existing custom provider users
```ts
storage.get('key', {
  provider: 'myProvider',
  foo: true,
  bar: 20,
}
```

List of changes:
- Enabled type checking on all unit tests in Storage, ts-jest **will fail** now if there are any typescript errors even in unit tests
  - Added some type practicing tests by calling apis with all available configs
- Rewrote some unit tests to remove any typescript compilation error
- `AWSS3Provider`'s `getProviderName` now returns `'AWSS3'` instead of `string`, which allows us to do conditional check like `ReturnType<Provider['getProviderName']> extends 'AWSS3' ? ... : ...`
- All main Storage APIs (get, put, remove, list, copy) will accept a `StorageProvider` as generic type, defaults to `AWSS3Provider`

#### Issue #, if available

#7667
#8925 

Unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/storage/package.json
```diff
@@ -53,7 +53,9 @@
   "jest": {
     "globals": {
       "ts-jest": {
-        "diagnostics": false,
+				"diagnostics": {
```
jamesaucode, line 56: Enabling diagnostic so unit tests will also be type tested to ensure our Typescript types are not breaking in future iterations

svidgen (approved):
Neat stuff!

Not at all a blocker, but was this the feature we had discussed creating typed provider instances around?

katiegoines (approved):

jamesaucode:
> Neat stuff!
> 
> Not at all a blocker, but was this the feature we had discussed creating typed provider instances around?

Yes! But we are not doing typed provider instance yet to keep parity across categories

petercwk (approved):

evcodes (approved):
LGTM 👍🏼

jamesaucode:
Added some changes to make sure current custom provider calls still works

e.g
```ts
storage.get('key', {
  provider: 'myProvider',
  foo: true,
  bar: 20,
  prop1: 1,
  prop2: 2,
}); // config will allow anything here since provider is specified
```

if they want stricter type checking then they can pass their provider in the generic type argument
```ts
storage.get<MyProvider>('key', {
  provider: 'myProvider',
  foo: true,
  bar: 20,
});
```

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -113,7 +113,7 @@ export class AWSS3Provider implements StorageProvider {
 	/**
 	 * get provider name of the plugin
 	 */
-	getProviderName(): string {
+	getProviderName(): 'AWSS3' {
```
jamesaucode, line 116: stricten return type to 'AWSS3' so if we do conditional checks like `T extends { provider: 'AWSS3' } ? ... : ...`

packages/storage/__tests__/Storage-unit-test.ts
```diff
@@ -490,30 +569,86 @@ describe('Storage', () => {
...
+			storage.addPluggable(customProvider);
+			await storage.get<TestCustomProvider>('key', {
+				provider: 'customProvider',
+				foo: false,
+				bar: 10,
+			});
+			expect(customProviderGetSpy).toBeCalled();
+		});
+		// backwards compatible with current custom provider user
+		test('get with custom provider should work with no generic type provided', async () => {
```
jamesaucode, line 587: These tests are for backwards compatibility for custom provider users.
As long as the config has `provider: string`, it means they are using their custom provider, in that case we allow any shape for the config.
If they want better type checking then they can do `storage.get<MyProvider>(...)`

packages/storage/src/Storage.ts
```diff
@@ -200,54 +197,63 @@ export class Storage {
...
+		src: StorageCopySource,
+		dest: StorageCopyDestination,
+		config?: StorageCopyConfig<T>
+	): Promise<any>;
+	public copy<T extends StorageProvider = AWSS3Provider>(
+		src: Parameters<T['copy']>[0],
+		dest: Parameters<T['copy']>[1],
+		config?: StorageCopyConfig<T>
+	): ReturnType<T['copy']> {
+		const provider = config?.provider || DEFAULT_PROVIDER;
```
jamesaucode, line 217: Changed this from the original object deconstruction method because Typescript could not infer that `Storage<Operation>Config<T>` has `provider: string`

packages/storage/src/types/Provider.ts
```diff
@@ -18,7 +18,7 @@ export interface StorageProvider {
 	cancel?(request: Promise<any>): void;
 
 	// copy object from src to dest
-	copy?(src: StorageCopySource, dest: StorageCopyDestination, config?): Promise<any>;
+	copy(src: StorageCopySource, dest: StorageCopyDestination, config?): Promise<any>;
```
elorzafe, line 21: Do we need to make this non optional? this would be a breaking change. I would make it optional as it was before, could be an issue?

packages/storage/__tests__/Storage-unit-test.ts
```diff
@@ -17,10 +22,47 @@ const options = {
 	credentials,
 	level: 'level',
 };
+
+
+class TestCustomProvider implements StorageProvider {
```
elorzafe, line 27: I would add a provider without `copy` to make it backwards compatible (currently is optional)

elorzafe (changes requested):
Just have two comments, `copy` should be optional to not make it breaking change

Awesome work! @jamesaucode 🥇

packages/storage/src/types/Provider.ts
jamesaucode: reverted the change, it's back to being optional!

packages/storage/__tests__/Storage-unit-test.ts
jamesaucode: good point. I removed the `copy` function  from this class and moved it to another.

elorzafe (approved):
Great work @jamesaucode 🎉 🌮 🏅

manueliglesias (approved):
LGTM 🎉

PR #8864, opened 2021-09-07
feat(@aws-amplify/datastore): observeQuery API

nickarocho:
## Description of changes

- Exposes new functionality for customers to combine `DataStore.observe(...)` and `DataStore.query(...)` into one useful method
- Returns an observable with a `snapshot` (Type: `DataStoreSnapshot<T>`) which contains: `items` (an array of all the records synced so far), `isSynced` (a boolean which tells us if the local DB is full synced up with AppSync), and `itemsChanged` (an array of records which were synced since the last glimpse of `items` was provided).

### Usage
__How customers might use this:__
```js
DataStore.observeQuery(Post).subscribe(snapshot => {
  const { items, isSynced, itemsChanged } = snapshot;
  // check if `isSynced`, and do something with the items/itemsChanged, i.e. update the UI
});
```

This is particularly useful when there are many (thousands of) records to be synced from AppSync. Rather than having to wait for the `modelSynced` event (which could take many seconds), we will get batches of 1000 records as they get synced down, which provides a much smoother UX, and a quicker "perceived load time" to the end user.

For example, when querying a `Post` model with 4123 records in AppSync, it would work as follows:

#### FIRST time loading the app - i.e. there are no available records in the local DB

1. The `snapshot` returned from the observable in `observeQuery` provides:
`snapshot = { items: [], isSynced: false, itemsChanged: [] }`
2. Then the first page of results will return:
`snapshot: { items: Array(1000), isSynced: false, itemsChanged: Array(1000)}`
3. Second page:
`snapshot: { items: Array(2000), isSynced: false, itemsChanged: Array(1000)}`
4. Third page:
`snapshot: { items: Array(3000), isSynced: false, itemsChanged: Array(1000)}`
5. Fourth page:
`snapshot: { items: Array(4000), isSynced: false, itemsChanged: Array(1000)}`
6. Finally, the last page of results will return:
`snapshot: { items: Array(4123), isSynced: true, itemsChanged: Array(123)}`

#### After syncing is complete (i.e. refreshing the app without calling `DataStore.clear()`):
1. `snapshot = { items: [4123], isSynced: false, itemsChanged: [] }`
2. `snapshot = { items: [4123], isSynced: true, itemsChanged: [] }`
(the Observer.next() is called twice - once right away, and once when the `modelSynced` event via `Hub.listen` is triggered)

### Predicates/Query Options
Customers can still pass `predicates`, as well as options like `sort` order and `pagination` values. For example:
```js
DataStore.observeQuery(
  Post,
  p => p.status("eq", "PUBLISHED")
).subscribe(snapshot => {
  const { items, isSynced, itemsChanged } = snapshot;
  // update the UI
});
```
```js
DataStore.observeQuery(
  Post,
  Predicates.ALL,
  { sort: s => s.createdAt(SortDirection.DESCENDING) }
).subscribe(snapshot => {
  const { items, isSynced, itemsChanged } = snapshot;
  // update the UI
});
```
```js
DataStore.observeQuery(
  Post,
  p => p.status("eq", "PUBLISHED"),
  {
    sort: s => s.createdAt(SortDirection.DESCENDING),
    page: 0,
    limit: 100
  }
).subscribe(snapshot => {
  const { items, isSynced, itemsChanged } = snapshot;
  // update the UI
});
```

### Cancellation
Customers can cancel this subscription (and the query process itself while in-flight), by calling the `.unsubscribe()` method on the stored Observable. For example, in a React app:
```js
import { useState, useEffect } from "react";

// store the Observable in state, to unsubscribe later
const [postObserveQuery, setPostObserveQuery] = useState();


useEffect(() => {
  let postSub = DataStore.observeQuery(Post, Predicates.ALL).subscribe(snapshot => {
    const { items, isSynced, itemsChanged } = snapshot;
    // update the UI
  });
  setPostObserveQuery(postSub);

  // when the component is unmounted, unsubscribe from the observeQuery
  return () => {
    postObserveQuery.unsubscribe();
  }
}, []);

// ... or, somewhere in the UI maybe there is a "cancel" button, which could fire the unsubscribe method
const cancelObserveQuery = () => {
  postObserveQuery.unsubscribe();
};
```
This will immediately invoke the cleanup method within the Observable and unsubscribe from the established subscription.

## Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/datastore/src/datastore/datastore.ts
+++ b/packages/datastore/src/datastore/datastore.ts
@@ -1101,6 +1101,63 @@ class DataStore {
 		});
 	};
 
+	observeQuery = (
+		model,
+		criteria = undefined,
+		options = {}
+	): Observable<any> => {
+		const SYNCED_ITEMS_THRESHOLD = 1000;
+		// TODO: programmatically toggle 'isSynced' back to false when out of sync... not sure how yet
+		let isSynced = false;
+
+		return new Observable(observer => {
+			let itemsChanged = [];
+			let items = [];
+
+			const handler = this.observe(model, criteria).subscribe(({ element }) => {
+				itemsChanged.push(element);
+
+				if (itemsChanged.length > SYNCED_ITEMS_THRESHOLD) {
+					onQueryComplete();
+				}
+			});
+
+			// return any locally available records
+			(async () => {
+				items = await this.query(model, criteria, options);
+				onQueryComplete();
+			})();
+
+			const onQueryComplete = () => {
+				items = [...items, ...itemsChanged];
+				console.log('onQueryComplete', { items, itemsChanged });
+
+				observer.next({
+					items,
+					isSynced,
+					itemsChanged,
+				});
+
+				itemsChanged = [];
+			};
+
+			Hub.listen('datastore', async hubData => {
+				const { event, data } = hubData.payload;
+				if (event === 'modelSynced' && data?.model?.name === model.name) {
+					isSynced = true;
+					onQueryComplete();
+				}
+			});
+
+			return () => {
+				if (handler) {
+					handler.unsubscribe();
+					observer.next({ items, isSynced, itemsChanged });
+				}
+			};
+		});
+	};
+
 	configure = (config: DataStoreConfig = {}) => {
 		const {
 			DataStore: configDataStore,
--- a/packages/datastore/src/datastore/datastore.ts
+++ b/packages/datastore/src/datastore/datastore.ts
@@ -43,6 +43,7 @@ import {
 	SchemaNamespace,
 	SchemaNonModel,
 	SubscriptionMessage,
+	DataStoreSnapshot,
 	SyncConflict,
 	SyncError,
 	TypeConstructorMap,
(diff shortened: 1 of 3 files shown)


packages/datastore/src/datastore/datastore.ts
```diff
@@ -1146,6 +1147,95 @@ class DataStore {
...
+				observer.next(snapshot);
+				itemsChanged = new Map();
+			};
+
+			const hubCallback = async hubData => {
+				const { event, data } = hubData.payload;
+
+				if (
+					event === ControlMessage.SYNC_ENGINE_MODEL_SYNCED &&
+					data?.model?.name === model.name
```
sammartinez, line 1198: Do we allow chaining methods like this currently? @manueliglesias ?

packages/datastore/src/datastore/datastore.ts
```diff
@@ -1146,6 +1147,95 @@ class DataStore {
...
+			const handle: ZenObservable.Subscription = this.observe(
+				model,
+				// @ts-ignore TODO: fix this TSlint error
+				criteria
+			).subscribe(({ condition, element, model, opType }) => {
+				itemsChanged.set(element.id, element);
+
+				const isSynced = this.sync.getModelSyncedStatus(model);
+
+				if (itemsChanged.size >= SYNCED_ITEMS_THRESHOLD || isSynced) {
```
sammartinez, line 1217: So I should read this as if the items are greater than or equal to 1000 or the item is already Synced, run this?

nickarocho: Exactly. Might need a comment for additional context, what do you think?

packages/datastore/src/datastore/datastore.ts
iartemiev: Optional chaining is supported in TS >= 3.7, which we recently upgraded to

Issue #8919, opened 2021-09-20
(amazon-cognito-identity-js) `error` should be an Error object


naseemkullah:
callback is taking non-error object as error, this causes pain further up (down?) the stack when trying to handle such errors with tooling that treats Error objects specially

naseemkullah:
referring to:

https://github.com/aws-amplify/amplify-js/blob/419e530e88bac14fa03b9696ce32fe02f23820c7/packages/amazon-cognito-identity-js/src/Client.js#L115-L120


and


https://github.com/aws-amplify/amplify-js/blob/419e530e88bac14fa03b9696ce32fe02f23820c7/packages/amazon-cognito-identity-js/src/Client.js#L131-L137

and

https://github.com/aws-amplify/amplify-js/blob/419e530e88bac14fa03b9696ce32fe02f23820c7/packages/amazon-cognito-identity-js/src/Client.js#L143-L148

chrisbonifacio:
Hi @naseemkullah 👋 Would you like to submit a PR for the changes you're suggesting?

naseemkullah:
Hi @chrisbonifacio, sure!

naseemkullah:
Hi @chrisbonifacio, https://github.com/aws-amplify/amplify-js/pull/8923 opened, PTAL 🙏

PR #8923, opened 2021-09-20
(cognito) callback error with error objects

naseemkullah:
fixes https://github.com/aws-amplify/amplify-js/issues/8919

cc @chrisbonifacio


packages/amazon-cognito-identity-js/src/Client.js
```diff
@@ -128,27 +126,19 @@ export default class Client {
...
 						return callback(err);
 					}
 					// otherwise check if error is Network error
 				} else if (err instanceof Error && err.message === 'Network error') {
-					const error = {
-						code: 'NetworkError',
-						name: err.name,
-						message: err.message,
-					};
-					return callback(error);
```
ashika01, line 148: this return is missed. Could you add that in? also what happened to the error object here? Am I missing something?

naseemkullah: the  `else if` just above checks to see if it is an `instaceof Error` which means it is already an error object 👍 
in this case we only need to add the `code` property and then it can be returned which is done below (fallback)

naseemkullah:
friendly ping

evcodes:
Hey @naseemkullah, thanks for your contribution! You help make Amplify better!

After pulling in your PR, I found that many of the client tests are now not passing. To merge this into main we would need to update those tests. I wrote the tests so if you need help with Jest please do not hesitate to reach out.

evcodes (changes requested):
Update amazon-cognito-identity-js Client unit tests

naseemkullah:
Hey @evcodes my pleasure!

> After pulling in your PR, I found that many of the client tests are now not passing. To merge this into main we would need to update those tests. I wrote the tests so if you need help with Jest please do not hesitate to reach out.

Ah I was under the impression that  the following meant that unit tests pass, but i will try locally, thanks!
image

evcodes:
> Hey @evcodes my pleasure!
> 
> > After pulling in your PR, I found that many of the client tests are now not passing. To merge this into main we would need to update those tests. I wrote the tests so if you need help with Jest please do not hesitate to reach out.
> 
> Ah I was under the impression that the following meant that unit tests pass, but i will try locally, thanks!
> image

You're right, I noticed that the lines that were missing coverage were the last 10 lines of Client.js which means that these changes do not introduce new test cases - I am going to go ahead and merge this in.

evcodes (approved):
LGTM - looking into token revocation changes and unit tests for that.

PR #8926, opened 2021-09-21
fix(@aws-amplify/auth): tests failing in Node > 14

mzarnawski:
#### Description of changes

Some other fixes in tests, this time Auth package, needed because: 1. they had errors 2. these errors would come up in Node > 14
Plus some unintended changes in formatting.

BTW, can we, in Auth tests, change .then() to await or is this disruptive in any way ?

Ran yarn test in Auth scope

#### Checklist

- [x] PR description included
- [x] `yarn test` passes (in Auth scope)
- [x] Tests are changed


mzarnawski:
@evcodes No problem, quite understandable that you need to meet on this. Maybe some label for this in Github would be helpful ?

mzarnawski:
@evcodes 
When discussing with the team please take into consideration:
1. The tests I have found so far do contain errors so they need to be amended anyway.
2. I have no idea whether it is a problem with Node, Jest, V8 or the combination of them.
3. Even if you don't go to Node 15/16, still we need to make these tests fail with Node 14 because they are false positives and then fix them.
4. Theoretically we may have some other tests that are false positives that will come up let's say under new Jest or Node version. So when creating tests we need to make sure that they don't give us false negatives or false positives. We kind of need to test the tests.

mzarnawski:
Closing as this was fixed in #9628

Issue #61, opened 2017-12-12
[feature request] s3 multipart resumable uploads


amitport:
in the style of https://github.com/TTLabs/EvaporateJS

jcharbo:
Hi, any idea on when this might be done?
thanks

clarle:
@mlabieniec @nidsharm 

We are absolutely looking for this functionality.  What's the estimated timeline for the 0.4 release?  

Thank you!

mlabieniec:
@clarle we have a large number of feature requests and are working on grooming all of them. I'm working on re-organizing the milestones etc. as we are now using a new tracking tool. Will keep this issue updated as we move ahead, this is definitely in the backlog.

paolavness:
I'd like to add a plus one to multipart uploads with features of
- resumable
- pausable
- subscribe to upload progress
- use S3 transfer accelerattion

Thank you for all the work on this 👍

ajhool:
I was having trouble uploading a file that was larger than 5MB and realized that Amplify is not exposing the ETag header in the CORS configuration when it creates the S3 bucket. 

I have not thoroughly tested this solution but I was able to upload a 7MB file and the network traffic indicated a successful multipart upload. Previously, a 7MB file upload always failed. **If you need to upload large files in your application, simply:**

1. Find your S3 bucket (created by Amplify)
2. Go To Permissions -> CORS Configuration
3. Add `<ExposeHeader>ETag</ExposeHeader>`
4. Make sure that ETag is included in "AllowedHeader" (a wildcard * by default)

After making this backend fix, the client side code appears to handle a multi-part file upload correctly.

-- Disclaimer: I'm not sure whether Amplify will overwrite this file, or if altering it somehow messes with Amplify's CloudFormation process --

My current S3 CORS Configuration
```
<?xml version="1.0" encoding="UTF-8"?>
<CORSConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
<CORSRule>
    <ID>S3CORSRuleId1</ID>
    <AllowedOrigin>*</AllowedOrigin>
    <AllowedMethod>GET</AllowedMethod>
    <AllowedMethod>HEAD</AllowedMethod>
    <AllowedMethod>PUT</AllowedMethod>
    <AllowedMethod>POST</AllowedMethod>
    <AllowedMethod>DELETE</AllowedMethod>
    <MaxAgeSeconds>3000</MaxAgeSeconds>
[...]

PR #8935, opened 2021-09-22
feat(storage): resumable uploads

petercwk:
#### Description of changes
- Allow uploads to s3 to be paused and resumed
- Also allows uploads to be resumed in the event of a network failure or page refresh

#### Issue #, if available
[https://github.com/aws-amplify/amplify-js/issues/61](61)

Unit tests and manual testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/storage/src/types/AWSS3Provider.ts
```diff
@@ -110,3 +122,11 @@ export type S3ProviderCopyConfig = Omit<CommonStorageOptions, 'level'> & {
 export type S3ProviderCopyOutput = {
 	key: string;
 };
+
+export type PutResult = {
+	key: string;
+};
+
+export type S3PutResult<T> = T extends { resumable: true }
```
jamesaucode, line 130: 👍  ! But I don't see this used anywhere, let's use this definition and replace the `S3ProviderPutOutput` above

jamesaucode (commented):
Good work! Just a  couple changes for now. I will do more tests locally later and see if there are more changes need to be addressed.

packages/storage/src/providers/AWSS3UploadTask.ts
```diff
@@ -0,0 +1,381 @@
...
+			logger.warn('Upload task already in progress');
+		} else {
+			this.state = State.IN_PROGRESS;
+			for (let i = 0; i < this.queueSize; i++) {
+				this._startNextPart();
+			}
+		}
+	}
+
+	public abort(): void {
```
elorzafe, line 342: We need to change this to `cancel`

packages/storage/src/providers/AWSS3UploadManager.ts
```diff
@@ -0,0 +1,292 @@
...
+	 *
+	 * @param blob - Blob that should be uploaded.
+	 * @return unique key of the file.
+	 */
+	private _getFileKey(blob: Blob, bucket: string, key: string): string {
+		// We should check if it's a File first because File is also instance of a Blob
+		if (this._isFile(blob)) {
+			return [
+				blob.name,
+				blob.lastModified,
```
elorzafe, line 126: Is this available this on every browser? And on react-native?

elorzafe (changes requested):
A few comments team needs to discuss before merging.

Overall looks great!

Nice job @petercwk

packages/storage/src/providers/AWSS3UploadTask.ts
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	private _onPartUploadCompletion({
+		eTag,
+		partNumber,
+		chunk,
+	}: {
+		eTag: string;
+		partNumber: number;
+		chunk: unknown;
```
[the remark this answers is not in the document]
jamesaucode: I think part of the reason is because we call `this._isBlob` on it later so that Typescript will be able to tell `chunk` is actually Blob, but I think this might be a bit confusing, I will do some refactoring on this part.

packages/storage/src/providers/AWSS3UploadTask.ts
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	public abort(): void {
+		this.pause();
+		this.queued = [];
+		this.completedParts = [];
+		this.bytesUploaded = 0;
+		this.state = State.ABORTED;
+		this.emitter.emit(TaskEvents.ABORT);
+		this.s3client
```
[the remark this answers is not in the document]
jamesaucode: I think I will add a catch block and log the error message here

PR #8946, opened 2021-09-24
In app messages

cshfang:
#### Description of changes

This PR adds a new feature - In-App Messaging - to the Amplify JS library.

_Note_: The category naming is not yet finalized. The API naming is not yet finalized.

This new feature was tested manually with an external React Native app and an AWS account with the In-App Messaging channel enabled in Pinpoint.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/aws-amplify-react-native/src/InAppMessaging/InAppMessagingProvider.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+import React, { ReactNode, useCallback, useEffect, useState } from 'react';
+import { Notifications } from 'aws-amplify';
+
+import InAppMessagingContext from './InAppMessagingContext';
+
+// TODO: replace generic T with InAppMessage from Notifications lib
```
elorzafe, line 19: Don't forget this one

packages/aws-amplify-react-native/src/InAppMessaging/types.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+// TODO: replace generic T with InAppMessage from Notifications lib
+export type InAppMessagingContextType<T = unknown> = {
```
elorzafe, line 15: don't forget this one

packages/cache/src/Utils/CachedUuid.ts
```diff
@@ -0,0 +1,39 @@
+import Cache from '../AsyncStorageCache';
+import { v1 as uuid } from 'uuid';
+
+const uuids = {};
```
elorzafe, line 4: What happen if the cache is clear? Is expected to still resolve this from memory instead of creating a new one?

packages/notifications/.npmignore
```diff
@@ -0,0 +1,10 @@
+__mocks__/**
+__tests__/**
+Authenticator_Architecture.png
```
elorzafe, line 3: This is probably copied from somewhere else, I would keep the minimum here

elorzafe: (nit)

packages/notifications/package.json
```diff
@@ -0,0 +1,105 @@
...
+	},
+	"author": "Amazon Web Services",
+	"license": "Apache-2.0",
+	"bugs": {
+		"url": "https://github.com/aws/aws-amplify/issues"
+	},
+	"homepage": "https://aws-amplify.github.io/",
+	"dependencies": {
+		"@aws-sdk/client-pinpoint": "^3.33.0",
+		"aws-amplify": "^4.0.2",
```
elorzafe, line 46: This would be a circular dependency, I would remove this

packages/notifications/src/Notifications.ts
```diff
@@ -0,0 +1,251 @@
...
+	ConsoleLogger as Logger,
+	HubCallback,
+	HubCapsule,
+	Hub,
+	parseMobileHubConfig,
+	StorageHelper,
+} from '@aws-amplify/core';
+import flatten from 'lodash/flatten';
+import noop from 'lodash/noop';
+import { AWSPinpointProvider } from './Providers';
```
elorzafe, line 25: I might rename this to `AWSNotificationsPinpointProvider` to avoid confusion with `Analytics` What do you think?

packages/notifications/src/Notifications.ts
```diff
@@ -0,0 +1,251 @@
...
+} from './types';
+
+const STORAGE_KEY_SUFFIX = '_inAppMessages';
+
+const logger = new Logger('Notifications');
+
+class NotificationsClass {
+	private config: Record<string, any> = {};
+	private inAppMessagesHandler: InAppMessagesHandler = noop;
+	private listeningForAnalyticEvents = false;
```
elorzafe, line 42: this seems a setting for the provider?

packages/notifications/src/Notifications.ts
```diff
@@ -0,0 +1,251 @@
...
+	private storageSynced = false;
+
+	constructor() {
+		this.config = {
+			storage: new StorageHelper().getStorage(),
+		};
+	}
+
+	configure = ({
+		listenForAnalyticsEvents = true,
```
elorzafe, line 53: Can this be done by the provider?

packages/notifications/src/Notifications.ts
```diff
@@ -0,0 +1,251 @@
...
+	setInAppMessagesHandler = (
+		handler: InAppMessagesHandler
+	): InAppMessagesHandler => {
+		if (handler && this.inAppMessagesHandler !== handler) {
+			this.inAppMessagesHandler = handler;
+		}
+		return this.inAppMessagesHandler;
+	};
+
+	syncInAppMessages = async (providerName = 'AWSPinpoint'): Promise<void> => {
```
elorzafe, line 144: 'AWSPinpoint' should be a constant?

packages/notifications/src/Notifications.ts
```diff
@@ -0,0 +1,251 @@
...
+			pluggable => pluggable.getProviderName() === providerName
+		);
+		if (index === -1) {
+			logger.debug(`No plugin found with name ${providerName}`);
+		} else {
+			this.pluggables.splice(index, 1);
+		}
+	};
+
+	setInAppMessagesHandler = (
```
elorzafe, line 135: We need to add the namespace right?

packages/notifications/src/Notifications.ts
```diff
@@ -0,0 +1,251 @@
...
+		const { event, data } = payload;
+		switch (event) {
+			case 'record': {
+				this.invokeInAppMessages(data);
+				break;
+			}
+			default:
+				break;
+		}
+	};
```
elorzafe, line 200: This should be part of the provider?

packages/notifications/src/types.ts
```diff
@@ -0,0 +1,103 @@
...
+export type NotificationsCategory = 'Notifications';
+
+export type NotificationEvent = {
+	name: string;
+	attributes?: Record<string, string>;
+	metrics?: Record<string, number>;
+};
+
+export interface NotificationsConfig {
+	listenForAnalyticsEvents?: boolean;
```
elorzafe, line 24: This should be for the provider?

packages/notifications/webpack.config.js
```diff
@@ -0,0 +1,48 @@
...
+	entry: {
+		'aws-amplify-notifications.min': './lib-esm/index.js',
+	},
+	externals: [
+		'react-native',
+		'@aws-amplify/cache',
+		'@aws-amplify/core',
+		'aws-sdk/clients/pinpoint',
+		'aws-sdk/clients/kinesis',
+		'aws-sdk/clients/firehose',
```
elorzafe, line 11: Why these?

elorzafe (changes requested):
Overall looks good, Namespace for InAppMessage is missing right?

packages/cache/src/Utils/CachedUuid.ts
cshfang: Hmm, I'm not sure what you mean by
> if the cache is clear

If the current map in memory is clear it should step into the `getUuid` logic where it will attempt to retrieve from Cache. If not found, then it will create a new uuid.

iartemiev: I think Francisco is asking about how the in-memory cache (i.e. `uuids`) gets invalidated. 

For example, if I were to delete the endpointId item from local/asyncStorage via `Cache.clear` or `Cache.removeItem(cacheKey)`, it seems it will remain in `uuids[cacheKey]` and will continue to be returned by `getUuid()`

packages/notifications/src/Notifications.ts
cshfang: Actually think that we should have these category level APIs apply their work to all configured pluggables at this time. We kind of copied this from Analytics and, currently, this would result in no difference since the only available pluggable is Pinpoint. But I think, in the long run, it would be good to not require any knowledge about provider names and even if it becomes necessary in the future, this would be a simple change to add back in.

iartemiev (changes requested):

packages/notifications/src/Notifications.ts
cshfang: Hmm, I think this kind of tie in with analytics could work with any InAppMessaging provider.

packages/notifications/package.json
cshfang: Hmm this is true. I will try to remove it. @calebpollman do you know why this was added?

packages/notifications/src/Notifications.ts
cshfang: I kind of feel that having it under this package should be sufficient (after all, every package has its own package.json, for example) and we don't need to set a precedence here for any future providers to also have the `AWSNotifications` prefix.

But I also do not feel that strongly opposed. Would anyone else care to weigh in?

calebpollman:
🎉

Issue #8970, opened 2021-09-30
Error message assumes the Storage provider for amazon-cognito-identity.js is "LocalStorage"


jrobbins-LiveData:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Not applicable

### Amplify APIs

Not applicable

### Environment information

<details>

```
# Put output below this line


```

</details>

### Describe the bug

[CognitoUser.js#L1444](https://github.com/aws-amplify/amplify-js/blob/f21c2d146fcabacf0584918b286ec68a644bec20/packages/amazon-cognito-identity-js/src/CognitoUser.js#L1444) (in the `getSession` function) assumes that one is using the default `window.LocalStorage` instead of a (supported) override configured via the `Storage` key.

The code is
```javascript
new Error('Local storage is missing an ID Token, Please authenticate'),
```

It might be more generically correct for the code to be

```javascript
new Error('Storage is missing an ID Token, Please authenticate'),
```

allowing for the case where the app chooses to override the default `LocalStorage` in favor of a different approach.

### Expected behavior

```javascript
new Error('Storage is missing an ID Token, Please authenticate'),
```

### Reproduction steps

Call `cognitoUser.signOut(revokeCallback);` without an ID Token in Storage.  The `revokeCallback` function is called back with an `error` arg which contains the "overspecified" text in question:
```
'Local storage is missing an ID Token, Please authenticate'
```

### Code Snippet

```javascript
// Put your code below this line.
new Error('Local storage is missing an ID Token, Please authenticate'),
```

### Log output

<details>

```
// Put your logs below this line


```

</details>

chrisbonifacio:
Hi @jrobbins-LiveData would you like to submit a PR for this?

jrobbins-LiveData:
Happy to!

chrisbonifacio:
Hey @jrobbins-LiveData can you point that PR to this repo's `main` branch? Looks like it's trying to merge into your fork instead.

jrobbins-LiveData:
[...]

PR #8978, opened 2021-10-01
Fix error message in amazon-cognito-identity-js

jrobbins-LiveData:
See [#8970](https://github.com/aws-amplify/amplify-js/issues/8970)

#### Description of changes

Removed the word "Local" and capitalized the letter "S" of "Storage".

#### Issue #, if available

[#8970](https://github.com/aws-amplify/amplify-js/issues/8970)

Simple deletion of the word "Local" qualifying "Storage" since Storage is pluggable and only defaults to "Local" but may be overridden by the user.

#### Checklist

- [x ] PR description included

### DIFF
--- a/packages/amazon-cognito-identity-js/src/CognitoUser.js
+++ b/packages/amazon-cognito-identity-js/src/CognitoUser.js
@@ -1441,7 +1441,7 @@ export default class CognitoUser {
 			this.refreshSession(refreshToken, callback, options.clientMetadata);
 		} else {
 			callback(
-				new Error('Local storage is missing an ID Token, Please authenticate'),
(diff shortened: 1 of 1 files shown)


hkjpotato:
@jrobbins-LiveData hey thanks for submitting the PR. After some thoughts, we think it might be better to keep the current wording because:
- "Local Storage" does not necessarily means "window.localStorage". It is more of a storage local to the system.
- Just using "Storage" can be a bit confusing for the default customer, as it can mean a remote database. People might be confused on whether it is an remote data issue, or just the data is lost in the local system.

jrobbins-LiveData:
@hkjpotato, thank you for the explanation.  I see I was conflating `Local Storage` with `windows.LocalStorage`.

Issue #6896, opened 2020-10-01
Pinpoint - new endpoints are created for the user without deleting the old ones, which causes the number of endpoints to reach 10 (which is the max)


idanlo:
**Describe the bug**
In my React Native app I call the updateEndpoint function whenever the app starts so that it creates an endpoint for my user if it doesn't exist, and update data if the endpoint already exists (Demographic, Location etc). What I think happens is that it creates an endpoint every time, and using the AWS CLI to list endpoints for my user I can see a lot of endpoints that have the INACTIVE status, and then an endpoint with the same address that has an ACTIVE status, according to the docs - 
```
Amazon Pinpoint automatically sets this value to ACTIVE when you create an endpoint or update an existing endpoint. Amazon Pinpoint automatically sets this value to INACTIVE if you update another endpoint that has the same address specified by the Address property.
```
Maybe is there a way that Amplify can check if there are INACTIVE endpoints for the user then delete them? And if there are active endpoints for the user then use them and do not create new ones.

**To Reproduce**
Call Analytics.updateEndpoint with the notification token of the device when the app starts, after a few times (some re-installs, different app builds) you will see you have a lot of endpoint even though you are only using a single device.

**Expected behavior**
Analytics.updateEndpoint() function knows how to handle existing endpoints, new endpoints, old endpoints...

**Code Snippet**
```
await Analytics.updateEndpoint({
  address: pushToken,
  userId: userId,
  userAttributes: {
    ...userAttributes,
  },
  optOut: 'NONE',
});
```

**What is Configured?**
the fields that exist in the aws-exports file are 
```
aws_project_region
aws_cognito_identity_pool_id
aws_cognito_region
aws_user_pools_id
aws_user_pools_web_client_id
aws_appsync_graphqlEndpoint
aws_appsync_region
aws_appsync_authenticationType
aws_user_files_s3_bucket
aws_user_files_s3_bucket_region
aws_mobile_analytics_app_id
aws_mobile_analytics_app_region
``` 
and the custom config for Amplify.configure looks like this 
```
const config = {
  ...aws_exports,
  graphql_headers: async () => {
    try {
      const token = (await Auth.currentSession()).getIdToken().getJwtToken();
      return { Authorization: token };
    } catch (e) {
      console.error(e);
      return {};
    }
  },
  Auth: {
    storage: AmplifyCommunityAsyncStorage,
  },
  PushNotification: {
    appId: aws_exports.aws_mobile_analytics_app_id,
[...]

PR #8982, opened 2021-10-04
fix: update TTL of the cached endpoint id

hkjpotato:
We have customer complaining about the number of endpoints reaching maximum limits (https://github.com/aws-amplify/amplify-js/issues/6896, https://github.com/aws-amplify/amplify-js/issues/7251). After investigation, we notice this issue is due to the local cached endpoint id being deleted, thus amplify recreates a new one and assign it to the same user.

The cached endpoint id might be deleted manually by devTool during development, or might due to using a different browser (e.g. incognito mode), which we cannot prevent.

However, we notice customers report that the issue also happens in production. A deep dive reveals that the current endpoint id is cached with the default [3 days TTL](https://github.com/aws-amplify/amplify-js/blob/main/packages/cache/src/Utils/CacheUtils.ts#L23), meaning it will be expired and deleted automatically after 3 days.  ([getItem](https://github.com/aws-amplify/amplify-js/blob/main/packages/cache/src/BrowserStorageCache.ts#L357) will only refresh the visitedTime, not mutating the expires).

We should persist this value longer (now set to 100 years) since it does not have any value to create a new endpoint for the same device.

#### Description of changes

- set endpoint id TTL to 100 years.
- set priority level to highest 1 (default to 5). Higher priority meaning it is less likely to be deleted when cache is full.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/7251
https://github.com/aws-amplify/amplify-js/issues/6896

1. check browser locally.
2. ensure the number is way below [safe integer in JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) so it won't overflow.

#### Checklist

- [X] PR description included
- [X] `yarn test` passes

### DIFF
--- a/packages/analytics/src/Providers/AWSPinpointProvider.ts
+++ b/packages/analytics/src/Providers/AWSPinpointProvider.ts
@@ -615,9 +615,13 @@ export class AWSPinpointProvider implements AnalyticsProvider {
 		if (!endpointId) {
 			endpointId = uuid();
 			// set a longer TTL to avoid endpoint id being deleted after the default TTL (3 days)
-			const ttl = 1000 * 60 * 60 * 24 * 365 * 1000; // 1000 years
+			// also set its priority to the highest to reduce its chance of being deleted when cache is full
+			const ttl = 1000 * 60 * 60 * 24 * 365 * 100; // 100 years
 			const expiration = new Date().getTime() + ttl;
-			Cache.setItem(cacheKey, endpointId, { expires: expiration });
+			Cache.setItem(cacheKey, endpointId, {
+				expires: expiration,
+				priority: 1,
+			});
 		}
 		return endpointId;
 	}
--- a/packages/analytics/src/Providers/AWSPinpointProvider.ts
+++ b/packages/analytics/src/Providers/AWSPinpointProvider.ts
@@ -614,7 +614,10 @@ export class AWSPinpointProvider implements AnalyticsProvider {
 		);
 		if (!endpointId) {
 			endpointId = uuid();
-			Cache.setItem(cacheKey, endpointId);
+			// set a longer TTL to avoid endpoint id being deleted after the default TTL (3 days)
+			const ttl = 1000 * 60 * 60 * 24 * 365 * 1000; // 1000 years
+			const expiration = new Date().getTime() + ttl;
+			Cache.setItem(cacheKey, endpointId, { expires: expiration });
 		}
 		return endpointId;
 	}


packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -614,7 +614,10 @@ export class AWSPinpointProvider implements AnalyticsProvider {
 		);
 		if (!endpointId) {
 			endpointId = uuid();
-			Cache.setItem(cacheKey, endpointId);
+			// set a longer TTL to avoid endpoint id being deleted after the default TTL (3 days)
+			const ttl = 1000 * 60 * 60 * 24 * 365 * 1000; // 1000 years
```
elorzafe, line 618: I hope doesn't overflow to a small number lol

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -614,7 +614,10 @@ export class AWSPinpointProvider implements AnalyticsProvider {
...
+			const expiration = new Date().getTime() + ttl;
+			Cache.setItem(cacheKey, endpointId, { expires: expiration });
```
elorzafe, line 620: Maybe it will be also useful to have `priority` also? or if you have `expires` you cannot have `priority`?

elorzafe (commented):
Just open questions.

Thanks @hkjpotato !

packages/analytics/src/Providers/AWSPinpointProvider.ts
iartemiev: It will...284,565 years from now 😸

iartemiev (approved):
LGTM! Thank you, KJ!

packages/analytics/src/Providers/AWSPinpointProvider.ts
hkjpotato: good catch

packages/analytics/src/Providers/AWSPinpointProvider.ts
hkjpotato: update it to 100 years

elorzafe: Update the code the comment to 100 years ;)

sammartinez (approved):
Overall, I'm good with this, we may want to call this out on our docs as well for customers to be aware of.

LGTM 🌮

elorzafe (approved):
Thanks @hkjpotato 🌮 🏅

hkjpotato:
I heard that browser release will still clean the localStorage. In fact, since "endpoint" is a concept about device, a browser release actually means a new device. That being said, in certain use case, such as web analytics, customer might not care which browser version those info comes from. For now, there are some go arounds I can think of, for your reference (depending on your use cases):

1. If you can get a predetermined value for endpoint id based on context, then you can configure it when setup analytics. This way, same context will always lead to same endpoint id. https://github.com/aws-amplify/amplify-js/issues/7251#issuecomment-879013837
2. Pinpoint will automatically evict "Push" channel type endpoints when it reaches the limit, you can update your endpoint to one of the push channel type (e.g. "PUSH") https://github.com/aws-amplify/amplify-js/issues/7251#issuecomment-931579550

PR #9121, opened 2021-10-27
feat(@aws-amplify/datastore): lazy loading Has_Many implementation

ArkamJ:
#### Description of changes
LazyLoading HasMany case implementation using AsyncCollection class.  Unit tests included.

IndexedDB unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


iartemiev (changes requested):

packages/datastore/src/datastore/datastore.ts
```diff
@@ -599,6 +600,22 @@ const createModelClass = <T extends PersistentModel>(
...
+							return instanceMemos[field];
+						}
+						const associatedWith = association.associatedWith;
+						const relatedModel = getModelConstructorByModelName(
+							USER,
+							relatedModelName
+						);
+						const result = await instance.query(relatedModel, c =>
+							(c as any)[associatedWith].eq(this.id)
+						);
```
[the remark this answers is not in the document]
iartemiev: Weird formatting in the diff... Here're just the 2 relevant lines:
```ts
const relatedModel: PersistentModelConstructor<typeof relatedModelName> = getModelConstructorByModelName(

c[associatedWith].eq(this.id)
```

packages/datastore/src/datastore/datastore.ts
```diff
@@ -620,6 +637,50 @@ const createModelClass = <T extends PersistentModel>(
...
+				}
+				return {
+					value: null,
+					done: true,
+				};
+			},
+		};
+	}
+	async toArray(arg?): Promise<T[]> {
+		const max = arg && arg.hasOwnProperty('max') ? arg.max : this.values.length;
```
svidgen, line 669: Let's make this typed, please:

```suggestion
	async toArray({max = Number.MAX_SAFE_INTEGER}: {max?: number} = {}): Promise<T[]> {
```

svidgen (changes requested):

packages/datastore/src/datastore/datastore.ts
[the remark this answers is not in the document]
ArkamJ: Ohh got it. Added that in

packages/datastore/src/datastore/datastore.ts
ArkamJ: Got it!

packages/datastore/__tests__/indexeddb.test.ts
```diff
@@ -350,6 +373,72 @@ describe('Indexed db storage test', () => {
...
+		expect(song).toStrictEqual(song2);
+
+		// new instance of the same record will be equal by value
+		expect(song).toEqual(song3);
+
+		// but not by reference
+		expect(song).not.toBe(song3);
+	});
+
+	test('AsyncCollection toArray test', async () => {
```
svidgen, line 405: I missed this the first time through these tests, but when we're using `.then()`, we need to make sure jest waits for test completion. It's safest IMO to use call pass `done` into the test body and `done()` explicitly when the test is all wrapped up. I actually struggled with this in some of my tests and found it to be the safest, most clear way to do it.

However you want to handle it though, let's make sure tests with `.then()` have some way of making sure jest waits for test completion. It might be simplest to test only a single `.then()` to ensure we have a promise, but to pretty much use `await` everywhere else.

svidgen (changes requested):
Sorry! One more little blocker -- I'm sorry I missed this last couple times I looked at this.

packages/datastore/__tests__/indexeddb.test.ts
```diff
@@ -417,26 +421,30 @@ describe('Indexed db storage test', () => {
...
+					savedSong1,
+					savedSong2,
+					savedSong3,
+					savedSong4,
+				]);
+			});
+			songs.toArray({ max: 3 }).then(value => {
+				expect(value).toStrictEqual([savedSong1, savedSong2, savedSong3]);
+			});
+			done();
```
svidgen, line 444: Sorry. I misread the snippet you shared internally. This doesn't quite solve the problem. I'll try to find time to sync up tomorrow with you.

svidgen (changes requested):

packages/datastore/__tests__/indexeddb.test.ts
svidgen: If you want to revise in the interim, the problem is the same here. A call to `.then()` puts work "off to the side" and execution continues to the next line immediately. That means `done()` is called and the test function both return before any of the promises callbacks are made. If one of those fails, it'll likely happen outside the scope of the test. Jest will probably complain about it if they fail, but I don't think it'll record it as an actual test failure.

It may be best to separate out all of these options scenarios anyway. Look into possibly using `test.each` or dynamically creating the tests. (Unless there's opposition to that idea -- though I did it for predicates and everyone seemed OK with it!) Then you'll only have a single `.then()` to worry about, and you can just `done()` _inside_ of that.

```
songs.toArray(options).then(value => {
  expect(value).toStrictEqual(expectedValues);
  done();
});
```

svidgen (changes requested):

packages/datastore/__tests__/indexeddb.test.ts
```diff
@@ -597,6 +651,64 @@ describe('Indexed db storage test', () => {
...
+				const savedSong2 = await DataStore.save(song2);
+				const savedSong3 = await DataStore.save(song3);
+				const savedSong4 = await DataStore.save(song4);
+				const songsArray = [savedSong1, savedSong2, savedSong3, savedSong4];
+				const q1 = await DataStore.query(Album, album1.id);
+				const songs = await q1.songs;
+				const expectedValues = [];
+				for (const num of expected) {
+					expectedValues.push(songsArray[num]);
+				}
```
svidgen, line 702: Nit: This is kind of a convoluted way to get at `exectedValues`, IMO. But, it works!

svidgen (approved):
No blockers that I can see!

iartemiev (approved):
LGTM! 👍🏻  Thanks Arkam!

PR #9122, opened 2021-10-28
build: upgrade ts to 3.8 to use export type syntax

hkjpotato:
#### Description of changes

Upgrade typescript from 3.7 to 3.8, so as to facilitate the upcoming esm bundle by esbuild, __BUT it is backward compatible__ to current version using [version selection](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions). Its backward compatibility has been tested by running through all [integ tests](https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/9282/workflows/b2a3d21f-0fde-458d-a090-71ff4a4674bf), it works in IDE like vscode as well.

Before TS3.8, "type" and "value" can be exported together implicitly. 
```
export interface A // type only, dropped in runtime
export enum B // a runtime value

import { A, B} from ... // isolatedModules tsc gets confused
```

The mixing of type and value creates confusion for both human and machine, especially for build tool like ESBuild/Vite, which relies on TS compiler to work in "isolatedModules" mode.

After TS3.8, we can define the import/export more explicitly.
```
export interface A 
export enum B 

import type { A } from ...
import { B } from ...
```
See detailed from typescript announcement https://devblogs.microsoft.com/typescript/announcing-typescript-3-8/#type-only-imports-exports, See similar issue related to esbuild support: https://github.com/evanw/esbuild/issues/1398.

Related changes due to this upgrade:
- use `typesVersions` to support backward compatibility: the only file changed is the index.d.ts for API, thus we redirects it to a separate file when customer is using a version < 3.8. This change has been tested by [integ tests](https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/9282/workflows/b2a3d21f-0fde-458d-a090-71ff4a4674bf). The syntax follows the [TS doc suggestion](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#file-redirects).
- Local `tsconfig` of analytics category is removed to use the centralized one, similar to other package. This can help unblock the esbuild project as well.
   - Its jest tsconfig is updated accordingly.
- The prettier upgrade is required to accept the new TS syntax
https://github.com/prettier/prettier/issues/7263
  - to support our monorepo structure, the local version used by react-native is kept by "nohoist".

Please note: 
- This is a small steps to move to new TS version, we have other places which can be rewritten to use type import and export, but to reduce the impact, I only make change `api/index.ts` where esBuild cannot infer the "type".
 - This does not affect the other build outputs. See [TS3.8 breaking change](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#typescript-38)
- The local useless tsconfig under analytics is removed. Our build process currently uses the centralized TS config specified in `scripts/build.js`.
   - Except for `amazon-cognito-identity-js`, which still relies on babel to transpile, and will be addressed separately.

#### Issue #, if available

Vite team @sodatea has pointed out a related issue for reference https://github.com/evanw/esbuild/issues/1525

This change has been verified by hosting ESM at localhost, based on @elorzafe's origin POC. The localhost modules can be used by ESM syntax directly in browser.

```
<script type="module" crossorigin="anonymous">

import * as core from  'http://localhost:8080/packages/core/dist/core-esm.js';
import * as auth from 'http://localhost:8080/packages/auth/dist/auth-esm.js';
import * as api from 'http://localhost:8080/packages/api/dist/api-esm.js';
import * as datastore from 'http://localhost:8080/packages/datastore/dist/datastore-esm.js';
import * as geo from 'http://localhost:8080/packages/geo/dist/geo-esm.js';
import * as cache from 'http://localhost:8080/packages/cache/dist/cache-esm.js';
import * as storage from 'http://localhost:8080/packages/storage/dist/storage-esm.js';
import * as predictions from 'http://localhost:8080/packages/predictions/dist/predictions-esm.js';
import * as interactions from 'http://localhost:8080/packages/interactions/dist/interactions-esm.js';
import * as pubsub from 'http://localhost:8080/packages/pubsub/dist/pubsub-esm.js';
import * as xr from 'http://localhost:8080/packages/xr/dist/xr-esm.js';
import * as apigraphql from 'http://localhost:8080/packages/api-graphql/dist/api-graphql-esm.js';
import * as apirest from 'http://localhost:8080/packages/api-rest/dist/api-rest-esm.js';

...

```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


hkjpotato:
They both failed due to the fact that the declaration include "import/export type" syntax
Screen Shot 2021-10-29 at 2 13 47 PM

Screen Shot 2021-10-29 at 2 13 03 PM

This is because the testing environment is using a tsc version which cannot recognize this declaration file. Will look more into
- If it is ok to exclude this syntax in declaration file.
- If it is ok to upgrade the tsc used by testing.

packages/api/src/types/index.ts
```diff
@@ -1,24 +1,24 @@
...
+ * and limitations under the License.
+ */
+
+/**
+ * This exports from the types directory is a temporary workaround, since Amplify CLI currently
+ * generates code that relies on this import path https://github.com/aws-amplify/amplify-cli/issues/3863
+ * This will be removed in future release when CLI and customers moves to recommeneded import styles.
+ */
+export {
+	graphqlOperation,
```
hkjpotato, line 20: I am not sure why this file shows up on the diff, nothing should be changed

sammartinez:
> > LGTM, we need to discuss with the team the impact this will have with Angular
> 
> 
> 
> LGTM, Angular 9+ supports TS 3.7+.

@hkjpotato can we double check old yo components as well if some customers are using them. May have some issues if the angular cli is around 7

hkjpotato:
> > > LGTM, we need to discuss with the team the impact this will have with Angular
> > 
> > 
> > LGTM, Angular 9+ supports TS 3.7+.
> 
> @hkjpotato can we double check old yo components as well if some customers are using them. May have some issues if the angular cli is around 7

Yep good call. Not sure if you can access these: https://app.circleci.com/pipelines/github/aws-amplify/amplify-js?branch=kj%2Fts38integ (A bunch of failure trials on angular, because the sample test is using ts3.7. The most recent one passes finally)
Angular CLI 7 is the reason why we introduce the "typesVersions" to support existing customers (they can still use ts3.7)

ashika01 (approved):
Looks good to me 🚢

package.json
```diff
@@ -71,7 +73,7 @@
 		"jest-config": "24.8.0",
 		"json-loader": "^0.5.7",
 		"lerna": "^3.13.1",
-		"prettier": "^1.19.0",
+		"prettier": "^2.4.1",
 		"pretty-quick": "^1.11.1",
```
calebpollman, line 77: Should we update `pretty-quick` along with `prettier`?

calebpollman (approved):
LGTM, left a minor non-blocking comment

packages/api/src/types/index.ts
```diff
@@ -1,24 +1,24 @@
...
- * This will be removed in future release when CLI and customers moves to recommeneded import styles.
- */
-export {
-	graphqlOperation,
-	GraphQLAuthError,
-	GraphQLResult,
-	GRAPHQL_AUTH_MODE,
-} from '@aws-amplify/api-graphql';
+/*
+ * Copyright 2017-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sammartinez, line 2:
```suggestion
 * Copyright 2017-2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

packages/api/src/types/index.ts
sammartinez: That is strange but can we update the copyright date :)

sammartinez (approved):
LGTM 🌮  Couple small nits but Non-Blocking to merge. Great work KJ!

elorzafe (approved):
LGTM, Thanks @hkjpotato 

Can you check @calebpollman and @sammartinez comments?

package.json
hkjpotato: I haven't tested its newer version but from the current version's peerDependencies, it works with prettier >=1.8.0, so I think it might be safer to keep it now.

hkjpotato:
> LGTM, Thanks @hkjpotato
> 
> Can you check @calebpollman and @sammartinez comments?

Thank you and I have checked the comments. Will discuss offline to see if we want to merge this with tomorrow's release.

wlee221 (approved):

PR #9141, opened 2021-11-01
feat(@aws-amplify/datastore): ObserveQuery performance and type enhancements

nickarocho:
#### Description of changes

- constrains ObserveQuery options with an explicit, unique Type
- adds a time interval/limit Promise race to return data when Model datasets are very large (i.e. have lots of fields in the Schema)

I'll be re-working the sample app/integ tests to utilize this data set to account for worst-case scenarios

`yarn test`, manual testing - adding integ test and reworking sample app next

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/src/datastore/datastore.ts
```diff
@@ -1190,21 +1209,49 @@ class DataStore {
...
 						}
+
+						// kicks off every subsequent race as results sync down
+						if (!raceInFlight) racePromises();
 					});
 
-					// will return any locally-available items in the first snapshot
-					generateAndEmitSnapshot();
+					// kicks off the first race to return any initial results
+					if (!raceInFlight) racePromises();
```
iartemiev, line 1224: Why not just return all the results from the local store right away?

packages/datastore/src/datastore/datastore.ts
```diff
@@ -1190,21 +1209,49 @@ class DataStore {
...
 					observer.error(err);
 				}
 			})();
 
+			const racePromises = async (): Promise<any> => {
+				raceInFlight = true;
+				startTimer();
+
+				try {
+					await Promise.race([timerPromise, limitPromise]);
```
iartemiev, line 1235: I know this already works as expected, but I think technically, we would want `limitPromise.promise` here, i.e. the promise itself not just the object that contains that promise as one of its properties

packages/datastore/src/datastore/datastore.ts
```diff
@@ -1152,23 +1153,41 @@ class DataStore {
...
 			const items = new Map<string, T>();
 			const itemsChanged = new Map<string, T>();
 			let deletedItemIds: string[] = [];
 			let handle: ZenObservable.Subscription;
 
+			// a mechanism to return data after X amount of seconds OR after the
+			// "limit" (itemsChanged >= this.syncPageSize) has been reached, whichever comes first
+			const TIMER_INTERVAL = 2000;
+			let timer: NodeJS.Timer;
+			let timerPromise: Promise<any>;
```
iartemiev, line 1173: currently, this is `Promise<string>` but since we don't use the resolved value anywhere, we can change line 1250 to `resolve()` and make this `Promise<void>`

packages/datastore/src/datastore/datastore.ts
```diff
@@ -1152,23 +1153,41 @@ class DataStore {
...
+			function LimitPromise(): void {
```
iartemiev, line 1174: Lets rename this to something more generic and move it to util. Can be e.g. `DeferredPromise`, `OpenPromise`, etc.

packages/datastore/src/datastore/datastore.ts
nickarocho: Yup, another good call.

packages/datastore/src/datastore/datastore.ts
nickarocho: True, kind of interesting how the `Promise.race` worked with just the object and not the promise itself 🤷🏻‍♂️

nickarocho (commented):
Thanks for the feedback, Ivan! Helpful and thorough as always. I've made some adjustments if you want to give it another look.

packages/datastore/src/datastore/datastore.ts
```diff
@@ -1190,21 +1203,48 @@ class DataStore {
...
-							generateAndEmitSnapshot();
+						const limit =
+							itemsChanged.size - deletedItemIds.length >= this.syncPageSize;
+
+						if (limit || isSynced) {
+							limitPromise.resolve();
 						}
+
+						// kicks off every subsequent race as results sync down
+						if (!raceInFlight) racePromises();
```
svidgen, line 1214: Trending towards spaghetti, IMO. And I don't see any unit test changes to accompany the change in behavior. Is there an opportunity to factor out some of the time-limited-snapshot mechanism (_buffer_?) into its own little testable class? (`MessageBuffer`? `SubscriptionBuffer`?) I could see that exercise simplifying both the buffering logic and the `observeQuery` method. (I could also see it being a fool's errand!)

What do you think? Opportunity or fool's errand?

svidgen: This comment is intended to apply generally to the buffering mechanism, not just these LOC, FWIW! I just didn't want to select every line in the diff to make this comment!

packages/datastore/src/util.ts
```diff
@@ -681,3 +682,71 @@ export const isAWSIPAddress = (val: string): boolean => {
...
+		self.resolve = resolve;
+		self.reject = reject;
+	});
+}
+
+export class SubscriptionBuffer {
+	private limitPromise = new DeferredPromise();
+	private timerPromise: Promise<string>;
+	private TIMER_INTERVAL = 2000;
+	private timer: NodeJS.Timer;
```
jamesaucode, line 698: It feels weird to have a `NodeJS` type here. Do you think it's better to use `ReturnType<typeof setTimeout>` instead? 😄 This way we can just let Typescript infer the type for us.
```suggestion
	private timer: ReturnType<typeof setTimeout>;
```

packages/datastore/src/util.ts
```diff
@@ -681,3 +682,71 @@ export const isAWSIPAddress = (val: string): boolean => {
...
 	);
 };
+
+export function DeferredPromise(): void {
+	const self = this;
+	this.promise = new Promise((resolve, reject) => {
+		self.resolve = resolve;
+		self.reject = reject;
+	});
+}
```
iartemiev, line 692: We can re-write this into a class like this:

```suggestion
class DeferredPromise {
	public promise: Promise<void>;
	public resolve: () => void;
	public reject: () => void;

	constructor() {
		const self = this;
		this.promise = new Promise((resolve, reject) => {
			self.resolve = resolve;
			self.reject = reject;
		});
	}
}
```

packages/datastore/__tests__/utils.test.ts
```diff
@@ -456,4 +457,57 @@ _deleted`;
...
+			});
+
+			expect(buffer.getIsRaceInFlight()).toBe(false);
+
+			const proto = Object.getPrototypeOf(buffer);
+			const spyOnRace = jest.spyOn(proto, 'racePromises');
+
+			buffer.start();
+			expect(buffer.getIsRaceInFlight()).toBe(true);
+			expect(spyOnRace).toBeCalledTimes(1);
```
iartemiev, line 482: We should expect `bufferCallback` to be called as well

packages/datastore/__tests__/utils.test.ts
```diff
@@ -456,4 +457,57 @@ _deleted`;
...
+			buffer.resolveBasePromise();
+			buffer.close();
+		});
+
```
iartemiev, line 486: I think it would be nice if we could test both happy paths, i.e. "limit wins" as well as "timer wins". We can discuss how to accomplish that

packages/datastore/src/util.ts
```diff
@@ -681,3 +682,71 @@ export const isAWSIPAddress = (val: string): boolean => {
...
+	private callback = () => {};
+	private errorHandler: (error: string) => void;
+	private defaultErrorHandler = (msg = 'SubscriptionBuffer error'): void => {
+		throw new Error(msg);
+	};
+
+	constructor(options: SubscriptionBufferOptions) {
+		this.callback = options.callback;
+		this.errorHandler = options.errorHandler || this.defaultErrorHandler;
+		this.TIMER_INTERVAL = options.maxInterval;
```
iartemiev, line 709: Same issue as we were seeing previously with the errorHandler. `options.maxInterval` will always overwrite `TIMER_INTERVAL` here. If we want it to default to 2000, we'd want to do:
```suggestion
		this.timerInterval = options.maxInterval || 2000;
```
Also, no need to make that property all caps, since it's configurable

iartemiev (changes requested):
Looks good! Just some minor changes requested

packages/datastore/__tests__/utils.test.ts
nickarocho: Thank you for helping me think about this and implement paths for each!

iartemiev (approved):
LGTM! 🌮

packages/datastore/src/util.ts
```diff
@@ -681,3 +682,79 @@ export const isAWSIPAddress = (val: string): boolean => {
...
+		this.promise = new Promise(
+			(resolve: (value: string | PromiseLike<string>) => void, reject) => {
+				self.resolve = resolve;
+				self.reject = reject;
+			}
+		);
+	}
+}
+
+export class SubscriptionBuffer {
```
svidgen, line 701: At a glance, this looks like it successfully extracts your limits race into something testable. That's a win in my book!

That said, this didn't end up being a _buffer_, really. If we care about naming, I think we need another one for this class. If I'm thinking about how we might re-use this elsewhere, I think sticking with some "racing" or "time limiting" terminology is more appropriate. IMO, it's more of a `TimeLimitForPromise` or `PromiseTimeLimit` &mdash; _or something like that_.

packages/datastore/src/util.ts
```diff
@@ -681,3 +682,79 @@ export const isAWSIPAddress = (val: string): boolean => {
...
+
+	public start(): void {
+		if (!this.raceInFlight) this.racePromises();
+	}
+
+	public close(): void {
+		clearTimeout(this.timer);
+	}
+
+	public resolveBasePromise(): void {
```
svidgen, line 757: There's no other resolve-oriented method on this. We could probably shorten this to `resolve`.

packages/datastore/src/util.ts
```diff
@@ -681,3 +682,79 @@ export const isAWSIPAddress = (val: string): boolean => {
...
+	constructor(options: SubscriptionBufferOptions) {
+		this.callback = options.callback;
+		this.errorHandler = options.errorHandler || this.defaultErrorHandler;
+		this.maxInterval = options.maxInterval || 2000;
+	}
+
+	private startTimer(): void {
+		this.timerPromise = new Promise((resolve, reject) => {
+			this.timer = setTimeout(() => {
+				resolve('TIMER');
```
svidgen, line 722: If invoking code is expected to use this, I'd strongly suggest using an exported `const` or `enum`.

packages/datastore/src/util.ts
```diff
@@ -681,3 +682,79 @@ export const isAWSIPAddress = (val: string): boolean => {
...
+	public start(): void {
+		if (!this.raceInFlight) this.racePromises();
+	}
+
+	public close(): void {
+		clearTimeout(this.timer);
+	}
+
+	public resolveBasePromise(): void {
+		this.limitPromise.resolve('LIMIT');
```
svidgen, line 758: Ditto `const`/`enum` comment.

svidgen (commented):
Not blocking. But a few things to consider.

packages/datastore/src/util.ts
nickarocho: Renamed to `DeferredCallbackResolver` via @iartemiev 's input.

svidgen (approved):

OskarD:
My generated models stopped working after upgrading to this version in aws-amplify.

When I tried to query the model, I got this error: `DataStore - The model is not an instance of a PersistentModelConstructor`

I also got this error everywhere in my application:
```
Call Stack
DataStore.<anonymous>
file:///Users/oskar/code/minna/minna-webapp/node_modules/aws-amplify/node_modules/@aws-amplify/datastore/lib/datastore/datastore.js (589:86)
step
file:///Users/oskar/code/minna/minna-webapp/node_modules/aws-amplify/node_modules/@aws-amplify/datastore/lib/datastore/datastore.js (44:23)
Object.next
file:///Users/oskar/code/minna/minna-webapp/node_modules/aws-amplify/node_modules/@aws-amplify/datastore/lib/datastore/datastore.js (25:53)
fulfilled
file:///Users/oskar/code/minna/minna-webapp/node_modules/aws-amplify/node_modules/@aws-amplify/datastore/lib/datastore/datastore.js (16:58)
```

Downgrading to `aws-amplify@4.3.5` and `@aws-amplify/datastore@352` fixed this problem

iartemiev:
@OskarD if you have **both** `aws-amplify` and `@aws-amplify/datastore` installed, that's likely what is causing the issue. Installing `aws-amplify` includes all of the packages already, so if you also install them individually, that can cause unexpected behavior. Please uninstall `@aws-amplify/datastore` and do a fresh install of your dependencies
```
npm uninstall @aws-amplify/datastore
rm -rf node_modules package-lock.json
npm install
```
If you're still getting the error after that, please open a new GitHub issue, so that we can track it and work on a fix.

OskarD:
Thanks, it seems to be working.

I have just followed AWS guides, how do I know which `@aws-amplify/X` to require and which are included? For instance, `@aws-amplify/ui-components` and `@aws-amplify/ui-react` both need to be required in order to use their exports, right? Or is there a better way to do it?

Issue #9128, opened 2021-10-29
aws-amplify datastore sqlite-adapter issue with connections [NOT NULL constraint]


cybexai:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

DataStore

### Amplify Categories

api

### Environment information

<details>

```
# Put output below this line

"dependencies": {
    "@aws-amplify/datastore-storage-adapter": "^1.1.10",
    "@azure/core-asynciterator-polyfill": "^1.0.0",
    "@react-native-async-storage/async-storage": "^1.15.9",
    "@react-native-community/masked-view": "^0.1.10",
    "@react-native-community/netinfo": "^6.0.2",
    "amazon-cognito-identity-js": "^5.2.0",
    "aws-amplify": "^4.3.2",
    "aws-amplify-react-native": "^5.0.4",
    ...
  },
```

</details>

### Describe the bug

We're receiving NOT NULL constraint when retrieving data from datastore. These NOT NULL constraints affects entities with "connections". 

### Expected behavior

We want the data of all entities to be retrieved from datastore

### Reproduction steps

type Enterprise
  @model
  @auth(
    rules: [
      {
        allow: owner
        ownerField: "belongTo"
      }
    ]
  ) {
  id: ID!
  name: String!
  image: String
  contact: String
  Departments: [Department!]
    @connection(keyName: "byEnterprise", fields: ["id"])
  belongTo: String!
}

...

type Department
  @model
  @key(name: "byEnterprise", fields: ["enterpriseID"])
  @auth(
    rules: [
      {
        allow: owner
        ownerField: "belongTo"
      }
    ]
  ) {
  id: ID!
  name: String!
  options: AWSJSON
}

...

await DataStore.query(
Enterprise,
(enterprise) => enterprise.isDeleted("eq", false),
{
sort: (s) => s.name(SortDirection.DESCENDING),
}
)

### Code Snippet

```javascript
// Put your code below this line.

```

### Log output

<details>

```
[...]

PR #9166, opened 2021-11-05
fix(datastore-storage-adapter): don't enforce required model fields in sql

iartemiev:
#### Description of changes

Don't attempt to enforce non-nullability of model fields in SQL, since relational data for those do not get saved in the fields themselves locally. 

For example:

```gql
type Post @model {
  comments: [Comment]! @connection
}
# OR
type Post @model {
  comments: [Comment!] @connection
}
```

Should **not** result in a `NOT NULL` constraint to be added to the `comments` column, as it will always be empty in the local store.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/9128

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore-storage-adapter/src/SQLiteAdapter/SQLiteUtils.ts
+++ b/packages/datastore-storage-adapter/src/SQLiteAdapter/SQLiteUtils.ts
@@ -133,23 +133,26 @@ export function modelCreateTableStatement(
 		}
 
 		if (isModelFieldType(field.type)) {
+			let columnParam = `"${field.name}" TEXT`;
+
 			// add targetName as well as field name for BELONGS_TO relations
 			if (isTargetNameAssociation(field.association)) {
-				const required = field.isRequired ? ' NOT NULL' : '';
-
-				let columnParam = `"${field.name}" TEXT`;
 				// check if this field has been explicitly defined in the model
 				const fkDefinedInModel = Object.values(model.fields).find(
 					(f: ModelField) => f.name === field.association.targetName
 				);
 
-				// only add auto-generate it if not
+				// if the field is not defined in the model, we have to add it in order to prevent
+				// "has no column named ${targetName}" error when inserting data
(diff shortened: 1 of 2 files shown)


packages/datastore-storage-adapter/src/SQLiteAdapter/SQLiteUtils.ts
```diff
@@ -133,23 +133,26 @@ export function modelCreateTableStatement(
...
-
-				let columnParam = `"${field.name}" TEXT`;
 				// check if this field has been explicitly defined in the model
 				const fkDefinedInModel = Object.values(model.fields).find(
 					(f: ModelField) => f.name === field.association.targetName
 				);
 
-				// only add auto-generate it if not
+				// if the field is not defined in the model, we have to add it in order to prevent
+				// "has no column named ${targetName}" error when inserting data
```
svidgen, line 146: Take this with a grain of salt (I'm still sick and fog-brained), but this comment kind of *impeded* my reading of the code. It says we're adding the column to avoid an error, which had me asking myself silly questions about whether we're doing inserts wrong. But, we're really just adding the column because the model implies it, and so it will not have been created yet... right?

Is my understanding correct? The field we're adding is just a regular-old FK. It just hasn't been explicitly declared in the graphql schema, and therefore isn't in the schema JSON, and therefore will not yet exist at this point.

Is that right?

svidgen (approved):
The meta-comment is non-blocking if I understood/summarized what's going on here correctly.

packages/datastore-storage-adapter/src/SQLiteAdapter/SQLiteUtils.ts
iartemiev: Yep, that's correct. Not a problem with IDB/AsyncStorage but it is a problem here. I'll clarify that in the comment a bit more

PR #9171, opened 2021-11-05
feat(amplify-js): add CD support

david-mcafee:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


david-mcafee:
Will add unit tests prior to merging this PR

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -142,6 +142,11 @@ const START_ACK_TIMEOUT = 15000;
  */
 const DEFAULT_KEEP_ALIVE_TIMEOUT = 5 * 60 * 1000;
 
+const standardDomainPattern =
+	'^https://\\w{26}.appsync-api.\\w{2}(?:(?:-\\w{2,})+)-\\d.amazonaws.com/graphql$';
```
iartemiev, line 146: This pattern is not precise as defined. I can insert any character in place of the periods (`.`) and still get it to match. We should escape it -> `\.`. This is [called out](https://lgtm.com/projects/g/aws-amplify/amplify-js/rev/pr-8560610adab5295d45c45daa006d5278cf7389ec) by LGTM as well.

For example, the following value will return a successful match, even though it's clearly incorrect (`+` instead of `.`):
`https://xxxxxxxxxxxxxxxxxxxxxxxxxx+appsync-api+us-east-2+amazonawsscom/graphql`

Overall, I think we should use a regular expression literal here (see suggested change), instead of a string pattern. Alternatively, we can construct a RegExp object by passing it the string pattern.

```suggestion
	/^https:\/\/\w{26}\.appsync-api\.\w{2}(?:(?:-\w{2,})+)-\d\.amazonaws.com\/graphql/
```

iartemiev (changes requested):

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
david-mcafee: Thanks Ivan, this is a great catch :)

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -162,6 +167,11 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 		throw new Error('Operation not supported');
 	}
 
+	// Check if url matches standard domain pattern
+	isCustomDomain(url: string): boolean {
```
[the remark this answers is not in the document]
david-mcafee: Agree with you on this. Updated!

iartemiev (approved):
LGTM! 🎉

david-mcafee:
Will remove formatting of `Credentials.ts`

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -593,7 +602,7 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 					this.socketStatus = SOCKET_STATUS.CONNECTING;
 					// Creating websocket url with required query strings
 					const protocol = this.isSSLEnabled ? 'wss://' : 'ws://';
-					const discoverableEndpoint = appSyncGraphqlEndpoint
+					let discoverableEndpoint: string = appSyncGraphqlEndpoint
```
elorzafe, line 605: We don't have to do the replace in case `isCustomDomain`. If a customer have a custom domain that has `appsync-api` on it it will change unnecessarily.

iartemiev: Good catch

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -142,6 +142,10 @@ const START_ACK_TIMEOUT = 15000;
  */
 const DEFAULT_KEEP_ALIVE_TIMEOUT = 5 * 60 * 1000;
 
+const standardDomainPattern = /^https:\/\/\w{26}\.appsync\-api\.\w{2}(?:(?:\-\w{2,})+)\-\d\.amazonaws.com\/graphql$/i;
```
sammartinez, line 145: Does this work in al browsers we support?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
david-mcafee: Will update this, great catch!!

elorzafe (approved):
LGTM! good work @david-mcafee 🌮 🏅

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
iartemiev: No reason why it wouldn't. Legacy browser support is spotty for lookaheads and lookbehinds, but this pattern is using neither of those

iartemiev: No reason why it wouldn't. Legacy browser support is spotty for lookaheads and lookbehinds, but this pattern is using neither of those

Issue #4257, opened 2019-10-25
GraphQLResult and Observable<object> incorrect types for API.graphql


selipso:
```
Type 'GraphQLResult | Observable<object>' is not assignable to type '{ data: CreateDocumentMutation; }'.
  Type 'GraphQLResult' is not assignable to type '{ data: CreateDocumentMutation; }'.
    Types of property 'data' are incompatible.
      Type 'object | undefined' is not assignable to type 'CreateDocumentMutation'.
        Type 'undefined' is not assignable to type 'CreateDocumentMutation'.  TS2322
```
Getting the above error using typescript after the new Amplify js version.

It would be better to use `Observable<any>` in this case because using type `object` is incompatible. In TypeScript the type `object` is the opposite of `any` and is not assignable to anything.

Amplifiyer:
@selipso, can you share your code sample where you are getting this error?

selipso:
```ts
async deleteDocument(id: string): Promise<{data:DeleteDocumentMutation}> {
    return API.graphql(graphqlOperation(deleteDocument, { input: { id } }));
}
```
Fairly simple wrapper function around the graphQL operation

ffxsam:
Agreed, I don't see how `Observable` applies here. `API.graphql()` should return a `GraphQLResult` only.

ffxsam:
Also, this:

```ts
export interface GraphQLResult {
    data?: object;
    errors?: [object];
    extensions?: {
        [key: string]: any;
    };
}
```

I would probably change the `data` line to:
```ts
data?: Record<string, any>;
```

buggy:
@ffxsam It's complaining about `result.data.createFolder` because `result.data` is optional. Once TypeScript 3.7 is release this will be much easier

```typescript
return result.data?.createFolder;
```

There's a PR to add support for generics to the API's which helps with some of the stuff mentioned here. See #3748 

There's been much discussion about the return type of `API.graphql()`. Basically it's a `GraphQLResult` if you're performing a `query`/`mutation` and an `Observable` if you're performing a `subscription`

ffxsam:
@buggy That's actually not true, as you can see above I'm checking for falsey values for `result` and `result.data` (which is partially blocked by the popup message).

It still gives me that error. And yeah, I haven't started using subscriptions yet, so your comment about `Observable` makes sense to me now.

buggy:
@ffxsam I couldn't see the code behind the popup :(

[...]

PR #9197, opened 2021-11-11
fix(@aws-amplify/api): Improve API.graphql return type

jamesaucode:
#### Description of changes
- Improve the return type signature of `API.graphql`
  - Users can pass in a generic type argument to the call, if it has the wrapper type `GraphQLQuery`, it returns a `Promise<GraphQLResult>`, if it has the wrapper type `GraphQLSubscription`, it returns an `Observable<GraphQLResult<MyType>>`
- Migrated JSDoc to using proper Typescript types and TSDoc
- Added proper type to any implicit `any` type params

Example:
```ts
const res = await API.graphql<GraphQLQuery<MyType>>({ query: 'query' }); // GraphQLResult<MyType>
const res2 = API.graphql<GraphQLSubscription<MyType>>({ query: 'query' }); // Observable<GraphQLResult<MyType>>
```

#### Issue #, if available
#4257
#6369 

#### Checklist
- [X] PR description included
- [X] `yarn test` passes


svidgen:
I realize this is still in draft, but I saw this on the list and I didn't want to forget to add Ivan and Manuel to the reviewers list. :)

jiahao-c:
Hi @jamesaucode , looking forward to this being merged and released:) Would really improve the developer experience of amplify api.

packages/api/__tests__/API-test.ts
```diff
@@ -82,6 +82,6 @@ describe('API test', () => {
 			.spyOn(GraphQLAPIClass.prototype, 'graphql')
 			.mockResolvedValue('grapqhqlResponse' as any);
 		const api = new API(null);
-		expect(await api.graphql(null)).toBe('grapqhqlResponse');
+		expect(await api.graphql({ query: 'query' })).toBe('grapqhqlResponse');
```
ashika01, line 85: why is this changed? also shouldn't null work too here? Else it would be breaking right?

packages/api/src/API.ts
```diff
@@ -82,115 +85,126 @@ export class APIClass {
...
-	 * @param {string} path - The path of the request
-	 * @param {json} [init] - Request extra params
-	 * @return {Promise} - A promise that resolves to an object with response status and JSON data, if successful.
+	 * @param apiName - The api name of the request
+	 * @param path - The path of the request
+	 * @param [init] - Request extra params
+	 * @return A promise that resolves to an object with response status and JSON data, if successful.
 	 */
-	get(apiName, path, init): Promise<any> {
+	get(apiName: string, path: string, init: Record<string, any>): Promise<any> {
```
ashika01, line 93: Would just an index signature be enough in this case than a `Record`?

packages/api/src/API.ts
```diff
@@ -82,115 +85,126 @@ export class APIClass {
...
-	 * @param {string} path - The path of the request
-	 * @param {json} [init] - Request extra params
-	 * @return {Promise} - A promise that resolves to an object with response status and JSON data, if successful.
+	 * @param apiName - The api name of the request
+	 * @param path - The path of the request
+	 * @param [init] - Request extra params
+	 * @return A promise that resolves to an object with response status and JSON data, if successful.
 	 */
-	post(apiName, path, init): Promise<any> {
+	post(apiName: string, path: string, init: Record<string, any>): Promise<any> {
```
ashika01, line 104: same as above

packages/api/src/API.ts
```diff
@@ -82,115 +85,126 @@ export class APIClass {
...
-	 * @param {string} path - The path of the request
-	 * @param {json} [init] - Request extra params
-	 * @return {Promise} - A promise that resolves to an object with response status and JSON data, if successful.
+	 * @param apiName - The api name of the request
+	 * @param path - The path of the request
+	 * @param [init] - Request extra params
+	 * @return A promise that resolves to an object with response status and JSON data, if successful.
 	 */
-	put(apiName, path, init): Promise<any> {
+	put(apiName: string, path: string, init: Record<string, any>): Promise<any> {
```
ashika01, line 115: same as above

packages/api/src/API.ts
```diff
@@ -82,115 +85,126 @@ export class APIClass {
...
+	 * @param apiName - The api name of the request
+	 * @param path - The path of the request
+	 * @param [init] - Request extra params
+	 * @return A promise that resolves to an object with response status and JSON data, if successful.
 	 */
-	patch(apiName, path, init): Promise<any> {
+	patch(
+		apiName: string,
+		path: string,
+		init: Record<string, any>
```
ashika01, line 129: same

packages/api/src/API.ts
```diff
@@ -82,115 +85,126 @@ export class APIClass {
...
-	 * @param {string} path - The path of the request
-	 * @param {json} [init] - Request extra params
-	 * @return {Promise} - A promise that resolves to an object with response status and JSON data, if successful.
+	 * @param apiName - The api name of the request
+	 * @param path - The path of the request
+	 * @param [init] - Request extra params
+	 * @return A promise that resolves to an object with response status and JSON data, if successful.
 	 */
-	del(apiName, path, init): Promise<any> {
+	del(apiName: string, path: string, init: Record<string, any>): Promise<any> {
```
ashika01, line 141: same

packages/api/src/API.ts
```diff
@@ -82,115 +85,126 @@ export class APIClass {
...
 
 	/**
 	 * Checks to see if an error thrown is from an api request cancellation
-	 * @param {any} error - Any error
-	 * @return {boolean} - A boolean indicating if the error was from an api request cancellation
+	 * @param error - Any error
+	 * @return If the error was from an api request cancellation
 	 */
-	isCancel(error) {
+	isCancel(error: any): boolean {
```
ashika01, line 161: Curious, whats stopping us from using `Error` object?

packages/api/src/API.ts
```diff
@@ -82,115 +85,126 @@ export class APIClass {
...
-	 * @param {string} path - The path of the request
-	 * @param {json} [init] - Request extra params
-	 * @return {Promise} - A promise that resolves to an object with response status and JSON data, if successful.
+	 * @param apiName - The api name of the request
+	 * @param path - The path of the request
+	 * @param [init] - Request extra params
+	 * @return A promise that resolves to an object with response status and JSON data, if successful.
 	 */
-	head(apiName, path, init): Promise<any> {
+	head(apiName: string, path: string, init: Record<string, any>): Promise<any> {
```
ashika01, line 152: same

packages/api/src/API.ts
```diff
@@ -82,115 +85,126 @@ export class APIClass {
...
 		options: GraphQLOptions,
 		additionalHeaders?: { [key: string]: string }
-	): Promise<GraphQLResult> | Observable<object> {
+	): T extends GraphQLQuery<T>
+		? Promise<GraphQLResult<T>>
+		: Observable<GraphQLResult<T>>;
+	graphql<T = any>(
+		options: GraphQLOptions,
+		additionalHeaders?: { [key: string]: string }
+	): Promise<GraphQLResult<T>> | Observable<GraphQLResult<T>> {
```
ashika01, line 207: hmm.. I would like someone from datastore to step in here. So you are implying if its GQL Query then the output will have to be a Promise with GQL Result right? Or am i missing something?

jamesaucode: Yea basically, a GQL Query gives a Promise, a GQL Subscription gives an Observable. I had a discussion with @iartemiev and @manueliglesias , they can probably chime in here.

packages/api/__tests__/API-test.ts
jamesaucode: It shouldn't have been allowed in the first place if we look at the function signature: https://github.com/aws-amplify/amplify-js/blob/d9dfff3eef7e5909f76fdea6054f8c02eb7ac8a2/packages/api/src/API.ts#L183-L196

We don't do type checks in our unit test so the tests were not failing, but ideally it should align with the actual implementation.

packages/api/src/API.ts
jamesaucode: This is essentially `axios.isCancel()` and it takes `any`.

packages/api/__tests__/API-test.ts
ashika01: oh interesting... I see. then shouldn't sending in null have caused some sort of error?

nfcsportal:
olk;o;o

packages/api/__tests__/API-test.ts
ashika01: Discussed offline - I would like for us to start doing these null checks for defensive code. James did point out we dont do this in other place and we have TS check and docs in place. Since it is a nit I m good 👍

ashika01 (approved):

packages/api/src/API.ts
```diff
@@ -82,115 +85,146 @@ export class APIClass {
...
+	 * @returns An Observable if queryType is 'subscription', else a promise of the graphql result from the query.
 	 */
-	graphql(
+	graphql<T = any>(
 		options: GraphQLOptions,
 		additionalHeaders?: { [key: string]: string }
-	): Promise<GraphQLResult> | Observable<object> {
+	): T extends GraphQLQuery<T>
+		? Promise<GraphQLResult<T>>
+		: Observable<GraphQLResult<T>>;
```
iartemiev, line 223: I think we should default the return type to the same one we have now for backward compatibility when type when `T` is neither `GraphQLQuery<T>` nor `GraphQLSubscription<T>`. 

```suggestion
	): T extends GraphQLQuery<T>
		? Promise<GraphQLResult<T>>
		: T extends GraphQLSubscription <T>
    ? Observable<GraphQLResult<T>>
    : Promise<GraphQLResult> | Observable<object>;
```

iartemiev (changes requested):
Requested a minor change to retain full backward compatibility. Otherwise, PR LGTM. Thanks, James!

iartemiev (approved):
LGTM 🎉

manueliglesias (approved):
LGTM 👍

ashika01:
it should be in `aws-amplify/@latest`? do u need a specific version when this got rolled out?

naingaungphyo:
@ashika01 I installed the latest version and it works well. Sorry for late feedback.

jdanielsyeah:
I can't get this to work. I'm using `"aws-amplify": "^4.3.28",`

```js
import { GraphQLSubscription } from "@aws-amplify/api/lib/types"; // I've attempted to type this manually, since VS Code/intellisense didn't pick it up.
// import API and other code omitted for brevity

const subscription = API.graphql<GraphQLSubscription<Card>>(
      graphqlOperation(subscriptions.onUpdateCard)
    ).subscribe({ // Property 'subscribe' does not exist on type 'Promise<GraphQLResult<any>> | Observable<object>'.
  Property 'subscribe' does not exist on type 'Promise<GraphQLResult<any>>'.ts(2339)
      next: ({ provider, value }) => console.log({ provider, value }),
      error: (error) => console.warn(error),
    });
```

Can someone please show us how this is supposed to work?

TimothyBritt:
Hello, I spent the better part of an afternoon scratching my head around this issue, and I have to agree — the issue is both frustrating and confusing. However, I have found a workaround that is not throwing type errors locally, but is far from ideal:

```typescript
import { Observable } from 'zen-observable-ts';
import { Amplify, Auth, API, graphqlOperation } from 'aws-amplify';
import { GraphQLSubscription } from '@aws-amplify/api/lib/types';
import * as APIt from './API'; // API types generated by amplify codegen

function App() {
//...
     useEffect(() => {
          const subscription = API.graphql<GraphQLSubscription<APIt.Channel>>(
              graphqlOperation(subscriptions.subscribeToChannel)
          ) as Observable<unknown>;
  
          subscription.subscribe({
              next: ({ provider, value }) => {
                  setReceived(value.data.subscribeToChannel.data);
              },
              error: (error) => console.warn(error),
          });
      }, [channel]);
//...
}
```

Casting the result of `API.graphql(graphQlOperation(...))` to `Observable<unknown>` allows the passing of the `{provider, value}` object to the `next` function on the subscription (Which, consequently is of type `Observable<T>` from `zen-observable`).

A better way to extend this workaround would be to figure out exactly what the expected type is that is passed to `Observable<T>` and replace `Observable<unknown>` with that type.

PR #9232, opened 2021-11-16
chore: Update two APIs to rethrow errors

cshfang:
#### Description of changes

This PR updates two of our InAppMessaging APIs to re-throw underlying errors since they make service calls. The errors returned from the service are descriptive and pointed, and can be useful for users. We will continue to swallow storage errors as they should be extremely rare and there is essentially no action that can be taken to remedy it if it did occur.

Local unit tests + Testing in app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


calebpollman (approved):

packages/notifications/src/inAppMessaging/InAppMessaging.ts
```diff
@@ -154,9 +154,14 @@ export default class InAppMessaging {
 	syncMessages = (): Promise<void[]> =>
 		Promise.all<void>(
 			this.pluggables.map(async (pluggable) => {
-				const messages = await pluggable.getInAppMessages();
-				const key = `${pluggable.getProviderName()}${STORAGE_KEY_SUFFIX}`;
-				await this.setMessages(key, messages);
+				try {
+					const messages = await pluggable.getInAppMessages();
+					const key = `${pluggable.getProviderName()}${STORAGE_KEY_SUFFIX}`;
+					await this.setMessages(key, messages);
```
elorzafe, line 160: Might need to revisit `setMessages` that is not throwing. How does storage could be configured? Thinking on the edge case `storage` run out of space or misconfigured.

cshfang: The storage helper is internal and there's not really configuration exposed that way. AsyncStorage which backs the React Native side does have gradle configurations to increase storage limit but that is only available on Android. I think the edge case of hitting storage limits are going to be really low considering the stored messages always get replaced during a sync.

elorzafe (approved):
LGTM 🌮

PR #9251, opened 2021-11-18
chore: mark ui packages as private

wlee221:
#### Description of changes

With [aws-amplify/amplify-ui](https://github.com/aws-amplify/amplify-ui) going GA and publishing to `latest`, we need to ensure that publish to `latest` only happens on the UI repo and not on the JS repo. 

For more context, Amplify UI publishes to `@aws-amplify/ui-react@2.x.x` and sets it as `latest`. But after a JS stable release, the `latest` dist-tag will be overwritten back to `@1.x.x`. To prevent this, we're putting `private` tags on the JS repo so that they are ignored from Lerna's versioning and publishing.

Migration will be the top priority as UI repo has reached GA. This PR is just to ensure `latest` dist-tags don't get overwritten by one another until we migrate completely.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-ui-components/package.json
+++ b/packages/amplify-ui-components/package.json
@@ -1,5 +1,6 @@
 {
 	"name": "@aws-amplify/ui-components",
+	"private": "true",
 	"version": "1.9.6",
 	"description": "Core Amplify UI Component Library",
 	"module": "dist/index.mjs",
--- a/packages/amplify-ui-angular/package.json
+++ b/packages/amplify-ui-angular/package.json
@@ -1,5 +1,6 @@
 {
 	"name": "@aws-amplify/ui-angular",
+	"private": "true",
 	"version": "1.0.35",
 	"description": "Angular specific wrapper for @aws-amplify/ui-components",
 	"publishConfig": {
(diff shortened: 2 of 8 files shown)


packages/amplify-ui-components/package.json
```diff
@@ -1,5 +1,6 @@
 {
 	"name": "@aws-amplify/ui-components",
+	"private": "true",
```
wlee221, line 3: These don't get published by the UI repository directly, but left `private` just so that the version stays static until we migrate.

sammartinez (approved):
LGTM 🌮

ashika01 (approved):
Sounds good 🚢  What package names are you publishing it to? Also let try to add deprecation notice on these packages soon :)

wlee221:
> What package names are you publishing it to? 
`@aws-amplify/ui-[framework]@2.x.x`, `@aws-amplify/ui@3.x.x`

> Also let try to add deprecation notice on these packages soon :)

Good point! Will do once we fully deprecate those packages.

ericclemmons (approved):

Issue #9162, opened 2021-11-04
Typescript: Better Typings for Analytics methods


hisham:
### Is this related to a new or existing framework?

Angular

### Is this related to a new or existing API?

Analytics

### Describe the feature you'd like to request

Analytics.record is confusing, I see multiple uses for it, well first here's what the typings say: 

```
    /**
     * Record one analytic event and send it to Pinpoint
     * @param {String} name - The name of the event
     * @param {Object} [attributes] - Attributes of the event
     * @param {Object} [metrics] - Event metrics
     * @return - A promise which resolves if buffer doesn't overflow
     */
    record(event: string | object, provider?: any, metrics?: EventMetrics): Promise<unknown>;
```

So is the second parameter provider or attributes? 

And then the documentation at https://docs.amplify.aws/lib/analytics/record/q/platform/js/#record-a-custom-event-with-attributes gives a very different way to do it: 

```
Analytics.record({
    name: 'albumVisit', 
    // Attribute values must be strings
    attributes: { genre: '', artist: '' }
});

```

Plus there's zero documentation on the "standard" events and attributes/metrics that pinpoint has such as '_monetization.purchase' and _item_price. 

### Describe the solution you'd like

Good consistent documentation. 

### Describe alternatives you've considered

Fishing around google and questioning my choice of using pinpoint. 

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [x] ⚠️ This feature might incur a breaking change

sammartinez:
Thanks for this feedback and feature request @hisham. We will take this into consideration as this may lead to breaking changes. We will provide an update on a timeline for this work once we have groomed this item and gotten clarity on next steps

cwomack:
The developer preview for v6 of Amplify has officially been released with improved support for TypeScript and much more! Please check out our [announcement](https://aws.amazon.com/blogs/mobile/aws-amplify-javascripts-new-developer-preview-with-reduced-bundle-sizes-improved-typescript-and-next-js-support/?sc_channel=sm&sc_publisher=TWITTER&sc_geo=GLOBAL&sc_outcome=awareness&linkId=238088647) and updated [documentation](https://docs.amplify.aws/lib-v1/project-setup/create-application/q/platform/js/) to see what has changed.

This issue should be resolved within the dev preview and upcoming General Availability for Amplify v6, but let us know with a comment if there are further issues.

cwomack:
[...]

PR #9272, opened 2021-11-23
feat(@aws-amplify/analytics): Analytics Typescript updates

jamesaucode:
#### Description of changes
- Adding clearer Typescript types for the parameters of `Analytics.record` function
- Removing implicit `any` wherever possible

#### Issue #, if available
related #9162

yarn clean && yarn setup-dev
yarn test

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/analytics/src/Analytics.ts
+++ b/packages/analytics/src/Analytics.ts
@@ -19,7 +19,13 @@ import {
 } from '@aws-amplify/core';
 import { AWSPinpointProvider } from './Providers/AWSPinpointProvider';
 
-import { AnalyticsProvider, EventMetrics } from './types';
+import {
+	AnalyticsProvider,
+	EventAttributes,
+	EventMetrics,
+	AnalyticsEvent,
+	AutoTrackOpts,
+} from './types';
 import { PageViewTracker, EventTracker, SessionTracker } from './trackers';
 
 const logger = new Logger('AnalyticsClass');
@@ -45,6 +51,8 @@ const trackers = {
 	session: SessionTracker,
 };
 
+type TrackerType = keyof typeof trackers;
+type Tracker = typeof trackers[TrackerType];
 let _instance = null;
 
 /**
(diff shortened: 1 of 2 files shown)


packages/analytics/src/Analytics.ts
```diff
@@ -219,14 +228,26 @@ export class AnalyticsClass {
...
+	 * @param [metrics] - Event metrics
 	 * @return - A promise which resolves if buffer doesn't overflow
 	 */
 	public async record(
-		event: string | object,
-		provider?,
+		eventName: string,
+		attributes?: EventAttributes,
+		metrics?: EventMetrics
+	);
```
ashika01, line 247: is this something we allow too?

packages/analytics/src/Analytics.ts
```diff
@@ -235,24 +256,24 @@ export class AnalyticsClass {
...
 			};
 		} else {
-			params = { event, provider };
+			params = { event, provider: providerOrAttributes };
 		}
 		return this._sendEvent(params);
 	}
 
-	public async updateEndpoint(attrs, provider?) {
+	public async updateEndpoint(attrs: Record<string, any>, provider?: string) {
```
ashika01, line 270: Curious: How is record here different from `[key:string]: any`?

jamesaucode: I don't think there's a difference in this case. However, I think `{ [key: string]: any }` is more idiomatic TS so I will change to that!

packages/analytics/src/Analytics.ts
jamesaucode: I believe this is the old way of using `Analytics.record` (at least from the implementation, this is supported)

ashika01: Gotcha. I dont see documentation on it. Do u think we can add in docs with a deprecation notice for this one?

ashika01 (approved):
LGTM. Just a tiny nit - you can add in and merge :)

packages/analytics/src/Analytics.ts
```diff
@@ -45,6 +51,8 @@ const trackers = {
 	session: SessionTracker,
 };
 
+type TrackerType = keyof typeof trackers;
+type Tracker = typeof trackers[TrackerType];
```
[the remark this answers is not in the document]
jamesaucode: > 

Updated, thanks 😄

hkjpotato (approved):
LGTM and thank you for running through offline. Nice job 🎆

PR #9278, opened 2021-11-24
feat(@aws-amplify/geo): createGeofences

TreTuna:
#### Description of changes

- add `createGeofence` to Geo and AmazonLocationServiceProvider
- add validation utility functions for coordinates and all geofence data structures

- tests.... lots of tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
Documentation will be written after all geofence APIs are created


packages/geo/src/util.ts
```diff
@@ -0,0 +1,154 @@
...
+
+	// Validate all coordinates are valid
+	try {
+		linearRing.forEach(coordinates =>
+			validateCoordinates(coordinates[0], coordinates[1])
+		);
+	} catch (error) {
+		const errorString = 'One or more of the coordinates are not valid';
+		logger.warn(errorString);
+		throw new Error(errorString);
```
Amplifiyer, line 65: In this throw, we missed the underlying error, i.e. what coordinates are wrong.

packages/geo/src/util.ts
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	return true;
+}
+
+export function validatePolygon(polygon: Polygon): boolean {
+	if (!Array.isArray(polygon)) {
+		const errorString = `Polygon ${JSON.stringify(
+			polygon
+		)} is of incorrect structure`;
```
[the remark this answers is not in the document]
TreTuna: Good call here. I'll add this in another commit as I'll have to update the test as well.

packages/geo/src/util.ts
```diff
@@ -0,0 +1,154 @@
...
+		const errorString = `Polygon ${JSON.stringify(
+			polygon
+		)} geometry.polygon must have a single LinearRing array`;
+		logger.warn(errorString);
+		throw new Error(errorString);
+	}
+	return true;
+}
+
+export function validateGeofences(geofences: any[]) {
```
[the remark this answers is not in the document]
TreTuna: Definitely meant to and not even sure how this got there. Good catch!

packages/geo/src/Geo.ts
```diff
@@ -163,17 +172,43 @@ export class GeoClass {
 	public async searchByCoordinates(
 		coordinates: Coordinates,
 		options?: SearchByCoordinatesOptions
-	) {
+	): Promise<Place> {
 		const { providerName = DEFAULT_PROVIDER } = options || {};
 		const prov = this.getPluggable(providerName);
 
+		const [lng, lat] = coordinates;
+		validateCoordinates(lng, lat);
```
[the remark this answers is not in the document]
TreTuna: Reasoning on not doing that here was that we want to bubble up the error that throws inside this function already as it has the contextual information about which coordinate is erroneous, so if we did a try/catch here, we'd end up just re-throwing the same error object. We can do that if it's what we want, but that's why I did not.

packages/geo/src/util.ts
TreTuna: Yeah, I was debating fleshing this out more to show deeper context on exactly which ones and how many had failed. I'll add that in now.

packages/geo/src/types/Geo.ts
```diff
@@ -97,3 +102,53 @@ export interface Place {
...
+export type GeofenceBase = {
+	geofenceId: string;
+	createTime: Date;
+	updateTime: Date;
+};
+
+// Output object for createGeofence and updateGeofence
+export type GeofenceResults = {
+	successes: Geofence[];
+	errors: any;
```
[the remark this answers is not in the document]
TreTuna: Can do. I did this real quick because I couldn't just reuse the Location Service error type due to them being PascalCased and use running camelCase. I meant to go back and dig into this more to build the proper type, will do that now.

packages/geo/src/util.ts
```diff
@@ -0,0 +1,154 @@
...
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+import { Longitude, Latitude, Geofence, Polygon, LinearRing } from './types';
+import { ConsoleLogger as Logger } from '@aws-amplify/core';
+
+const logger = new Logger('Geo');
+
+export function validateCoordinates(lng: Longitude, lat: Latitude): boolean {
```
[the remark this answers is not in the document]
TreTuna: Sounds good. Will remove the `return true;` from these validations as we want to keep the errors with the contextual information.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -204,6 +212,8 @@ export class AmazonLocationServiceProvider implements GeoProvider {
 			throw new Error('No credentials');
 		}
 
+		validateCoordinates(coordinates[0], coordinates[1]);
```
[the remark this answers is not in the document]
TreTuna: 👍

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -286,9 +369,24 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+			(!this._config.geofenceCollections ||
+				!this._config.geofenceCollections.default) &&
+			!optionalGeofenceCollectionName
+		) {
+			const errorString =
+				'No Geofence Collections found, please run `amplify add geo` to create one and run `amplify push` after.';
 			logger.warn(errorString);
 			throw new Error(errorString);
 		}
+		// TODO: if optionalGeofenceCollectionName is given, make API call to verify it exists
```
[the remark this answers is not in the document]
TreTuna: You are right in that we don't necessarily need to here, which is why I didn't originally when I wrote these. I had this thought late in the day/evening yesterday and just marked myself a note to return to it and evaluate. I agree with you and will just remove these comments.

packages/geo/src/types/Geo.ts
```diff
@@ -22,6 +22,10 @@ export interface GeoConfig {
 			items: string[];
 			default: string;
 		};
+		geofenceCollections?: {
+			items: string[];
+			default: string;
+		};
```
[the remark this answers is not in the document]
TreTuna: Yes... It's in our `.prettierrc.js` 😅
```
module.exports = {
	trailingComma: 'es5',
	singleQuote: true,
	useTabs: true,
	arrowParens: 'avoid',
};
```

packages/geo/src/types/Geo.ts
```diff
@@ -97,3 +102,53 @@ export interface Place {
...
 }
+
+// Array of 4 or more coordinates, where the first and last coordinate are the same to form a closed boundary
+export type LinearRing = Coordinates[];
+
+// An array of one linear ring
+export type Polygon = LinearRing[];
+
+// Geometry object for Geofence Geometry
+export type GeofenceGeometry = {
```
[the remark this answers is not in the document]
TreTuna: I don't think I'd want to do that in this case because this is the type specifically for `GeofenceInput.geometry`. If we did add a `CircularGeometry` later, it would more go into a union type with `Polygon` here, however, even a circle is just considered a `Polygon` with a lot of points.

packages/geo/src/types/Geo.ts
```diff
@@ -97,3 +102,53 @@ export interface Place {
...
+// Status types for Geofences
+export type GeofenceStatus =
+	| 'ACTIVE'
+	| 'PENDING'
+	| 'FAILED'
+	| 'DELETED'
+	| 'DELETING';
+
+// Base geofence object
+export type GeofenceBase = {
```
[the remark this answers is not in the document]
TreTuna: Not necessarily, just my snippet I used in all of these has `export` tagged onto it as well. I can remove it.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -241,9 +251,82 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+			response = await client.send(command);
+		} catch (error) {
+			logger.debug(error);
+			throw error;
+		}
+
+		// Convert results to camelCase
+		const results: GeofenceResults = camelcaseKeys(response, {
+			deep: true,
+		}) as any as GeofenceResults;
```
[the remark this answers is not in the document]
TreTuna: Hmm... Yes, very valid here. I'll rethink this.

We had thought about grabbing the geofence `geometry` from the input array and adding it back into the return objects since the service doesn't return it. However, since everything is in arrays, that probably wouldn't be very efficient to iterate and add into each object after the fact.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -241,9 +251,82 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+			region: this._config.region,
+			customUserAgent: getAmplifyUserAgent(),
+		});
+		const command = new BatchPutGeofenceCommand(geofenceInput);
+
+		let response: BatchPutGeofenceCommandOutput;
+		try {
+			response = await client.send(command);
+		} catch (error) {
+			logger.debug(error);
```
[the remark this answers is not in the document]
TreTuna: Actually, to follow with the rest of what we have here, we'd use `logger.warn`, though, perhaps we should change most of these other ones to `logger.error` as well?

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -241,9 +251,82 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+		// Create the BatchPutGeofence input
+		const geofenceInput: BatchPutGeofenceCommandInput = {
+			Entries: PascalGeofences,
+			CollectionName: this._config.geofenceCollections.default,
+		};
+
+		// Map options to Amazon Location Service input object
+		if (options?.collectionName) {
+			geofenceInput.CollectionName = options.collectionName;
+		}
```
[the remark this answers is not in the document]
TreTuna: Yeah, I can do that.

packages/geo/src/types/Geo.ts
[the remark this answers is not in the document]
Amplifiyer: My reasoning is that a Geofence _has a_ geometry which can be either a polygon or a circle while polygon or circle don't necessarily have to be geofences.

Moreover in future, Circle would be represented by a center and a radius which we would expose through Geo APIs (e.g. native devices don't even support polygon geofences)

The following would look weird since circle _will not_ be a polygon
```ts
type Circle = {
        center: Coordinate,
        radius: double
};
export type GeofenceGeometry = {
	polygon: Polygon | Circle;
};
```

What I am proposing is 
```ts
type CircleGeometry = {
        center: Coordinate,
        radius: double
};
export type PolygonGeometry = {
	polygon: Polygon;
};
export type GeofenceInput = {
	geofenceId: string;
	geometry: PolygonGeometry | CircleGeometry;
};
```

packages/geo/src/Geo.ts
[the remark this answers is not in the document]
Amplifiyer: We aren't doing any different processing of error thrown from `prov.searchByCoordinates(coordinates, options);` just logging and rethrowing, so it makes sense to put it under the same try/catch to at least get the logging there?

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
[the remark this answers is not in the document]
Amplifiyer: I'm not sure if throwing also logs error on the console or wherever `logger` logs. That would be one reason to not use the `Error` level. I'd say check with the js team on what they recommend as best practice.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -241,9 +249,83 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+		const client = new LocationClient({
+			credentials: this._config.credentials,
+			region: this._config.region,
+			customUserAgent: getAmplifyUserAgent(),
+		});
+		const command = new BatchPutGeofenceCommand(geofenceInput);
+
+		let response: BatchPutGeofenceCommandOutput;
+		try {
+			response = await client.send(command);
```
Amplifiyer, line 319: There are couple of limits that exist in the service. [A batch can only have 10 geofences at a time](https://docs.aws.amazon.com/location-geofences/latest/APIReference/API_BatchPutGeofence.html#API_BatchPutGeofence_RequestBody) and second [a polygon can only have 1000 coordinates/vertices ](https://docs.aws.amazon.com/location-geofences/latest/APIReference/API_BatchPutGeofenceRequestEntry.html#API_BatchPutGeofenceRequestEntry_Contents)

The first one is critical for us to implement, second is a good to have validation in client side.

packages/geo/src/Geo.ts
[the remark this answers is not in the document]
TreTuna: Fair point... May want to chat on all the error logging we're doing here sometime soon. I'll put this in here atm though.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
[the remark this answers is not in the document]
TreTuna: Yeah, sounds good, I'll chat with someone there and see what they say.

packages/geo/src/types/Geo.ts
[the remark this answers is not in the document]
TreTuna: Ahhh, ok, I get where you're going here now. I'm good with this and will make the change.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: I completely missed the 10 geofence limit on that one. Great catch there! I'll rework that to make multiple calls if the input has more than 10 and aggregate the results.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -241,9 +249,83 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+	 *   successes: list of geofences successfully created
+	 *   errors: list of geofences that failed to create
+	 */
+	public async createGeofences(
+		geofences: GeofenceInput | GeofenceInput[],
+		options?: GeofenceOptions
+	): Promise<GeofenceResults> {
+		const credentialsOK = await this._ensureCredentials();
+		if (!credentialsOK) {
+			throw new Error('No credentials');
```
thaddmt, line 273: Nit: Does this mean `no credentials` or just invalid ones? Should this error message be more helpful?

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -241,9 +249,83 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+		}
+
+		this._verifyGeofenceCollections(options?.collectionName);
```
thaddmt, line 276: Could it be worth it to verify the default geofence collection as well? In case the user manually deletes their collection outside of amplify?

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: This credentials check is basically copy/pasted from other categories and what we've been using thus far. The message could possibly be better here, though in the `_ensureCredentials` method it does have loggers  for a bit more information that would be hit as well here.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: We currently do no verifying via API that the collections actually exist on AWS, only that there are collections and a default exists in `aws-exports`.

thaddmt (approved):

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -241,9 +248,101 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+		}
+
+		// If single geofence input, make it an array for batch API call
+		let geofenceInputArray;
+		if (!Array.isArray(geofences)) {
+			geofenceInputArray = [geofences];
+		} else {
+			geofenceInputArray = geofences;
+		}
+
```
Amplifiyer, line 289: I'm guessing this is not needed here since the only input type for the provider method is an array of geofences?

packages/geo/src/types/Geo.ts
```diff
@@ -97,3 +102,52 @@ export interface Place {
...
+	geofenceId: string;
+	geometry: PolygonGeometry;
+};
+
+// Options object for createGeofence and updateGeofence
+export type GeofenceOptions = {
+	providerName?: string;
+};
+
+// Error type for errors for Amazon Location Service batchPutGeofence API
```
Amplifiyer, line 128:
```suggestion
// Error type for errors related to Geofence API calls
```

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: Ahh, yes, I forgot to remove this when I changed the Geo method to do the conversion.

packages/geo/src/types/AmazonLocationServiceProvider.ts
```diff
@@ -1,5 +1,44 @@
...
+	collectionName?: string;
+};
+
+type AmazonLocationServiceError = GeofenceError & {
+	error: {
+		code: string;
+		message: string;
+	};
+	geofenceId: string;
+};
```
Amplifiyer, line 26: Why do we need to create this type when it's identical to `GeofenceError`, that is defined at the category level.

packages/geo/src/types/AmazonLocationServiceProvider.ts
```diff
@@ -1,5 +1,44 @@
...
+
+export type AmazonLocationServiceGeofence = Geofence & {
+	status: AmazonLocationServiceGeofenceStatus;
+};
+
+export type AmazonLocationServiceCreateUpdateGeofenceResults =
+	CreateUpdateGeofenceResults & {
+		successes: GeofenceBase[];
+		errors: AmazonLocationServiceError[];
+	};
```
Amplifiyer, line 44: Does this mean that `AmazonLocationServiceCreateUpdateGeofenceResults` will override the `errors` key of `CreateUpdateGeofenceResults`?
Nevertheless, it seems identical to `CreateUpdateGeofenceResults` so maybe we can just use that?

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -241,9 +248,101 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+		const PascalGeofences: BatchPutGeofenceRequestEntry[] = camelcaseKeys(
+			geofenceInputArray,
+			{
+				deep: true,
+				pascalCase: true,
+			}
+		);
+
+		// Create results object
+		const results = {
```
Amplifiyer, line 300: Use the type wherever possible. Especially here while constructing the return object.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -241,9 +248,101 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+		 */
+		while (PascalGeofences.length > 0) {
+			// Splice off 10 geofences from input clone due to Amazon Location Service API limit
+			const tenGeofences = PascalGeofences.splice(0, 10);
+
+			// Make API call for the 10 geofences
+			const response = await this._AmazonLocationServiceBatchPutGeofenceCall(
+				tenGeofences,
+				options?.collectionName
+			);
```
Amplifiyer, line 319: this method can throw which will break the loop and the subsequent batches will not be called. We should figure out if we want to continue with the next batches or exit out the loop with creating error objects for all the left over batches including the one that threw. And ideally we should pick one of the two based on the type of exception thrown (e.g. recoverable or not)

In the end, for this type of API, the number of items in the input request should be equal to the number of items (success + errors) in the output response in every case.

packages/geo/src/types/AmazonLocationServiceProvider.ts
TreTuna: Yeah... I think this came from wanting to split some things out, but then never finding something deeper that needed to happen. I'll move it back out.

packages/geo/src/types/Provider.ts
```diff
@@ -10,13 +10,15 @@
...
-
 import {
 	SearchByTextOptions,
 	SearchByCoordinatesOptions,
 	Coordinates,
 	Place,
 	MapStyle,
+	GeofenceInput,
+	GeofenceOptions,
+	GeofenceResults,
```
Amplifiyer, line 21: I think this needs to change to `CreateUpdateGeofenceResults`? Build is failing because of this.

packages/geo/src/types/AmazonLocationServiceProvider.ts
TreTuna: We don't, there was a nuance I was going to add here to specify the strings that Amazon Location Service returns, but we've decided not to do that and I'll remove this now.

packages/geo/src/types/Provider.ts
TreTuna: Yeah, I just updated that

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: In this case, that call should only error out if something happens to the call itself, not if a geofence fails to create, since it returns an object with both `successes` and `errors` in those cases, which we are handling that. With this, I would lean towards the path of catching the error, adding all of that specific call's geofenceIds to our `error` array with the message from the thrown error itself, still running the remaining batches, and then returning everything in the end. Does this sound like a good path forward to you?

Amplifiyer: Yup that sounds good

TreTuna: Ok, I have the implementation here, but need to write more tests.
Just verified in my sample app that it's successfully making 150 geofences in a single call.

TreTuna: Should be good to go now.

Amplifiyer (approved):
🚢

PR #9298, opened 2021-11-30
fix(@aws-amplify/datastore): consecutive updates with timestamps

iartemiev:
#### Description of changes

As of [this](https://docs.amplify.aws/cli/reference/feature-flags/#addTimestampFields) CLI feature flag timestamp fields get automatically added to the schema with a readOnly property and are added to the GraphQL selection set for queries and mutations. 

This PR adds logic to ignore the default (or custom - if configured) timestamp fields when comparing records as part of the [Consecutive Saves protocol](https://github.com/aws-amplify/amplify-js/pull/7354).

* unit tests updated
* sample app with default timestamps
* sample app with custom timestamp fields

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/src/sync/outbox.ts
+++ b/packages/datastore/src/sync/outbox.ts
@@ -12,7 +12,7 @@ import {
 	PersistentModelConstructor,
 	QueryOne,
 } from '../types';
-import { SYNC, valuesEqual } from '../util';
+import { USER, SYNC, valuesEqual } from '../util';
 import { TransformerMutationType } from './utils';
 
 // TODO: Persist deleted ids
@@ -31,9 +31,8 @@ class MutationEventOutbox {
 		mutationEvent: MutationEvent
 	): Promise<void> {
 		storage.runExclusive(async s => {
-			const mutationEventModelDefinition = this.schema.namespaces[SYNC].models[
-				'MutationEvent'
-			];
+			const mutationEventModelDefinition =
+				this.schema.namespaces[SYNC].models['MutationEvent'];
 
 			const predicate = ModelPredicateCreator.createFromExisting<MutationEvent>(
 				mutationEventModelDefinition,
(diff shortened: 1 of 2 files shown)


packages/datastore/src/sync/outbox.ts
```diff
@@ -252,6 +248,49 @@ class MutationEventOutbox {
...
+		model: string,
+		record: PersistentModel
+	): PersistentModel {
+		const CREATED_AT = 'createdAt';
+		const UPDATED_AT = 'updatedAt';
+
+		let createdTimestamp = CREATED_AT;
+		let updatedTimestamp = UPDATED_AT;
+
+		const modelAttributes = this.schema.namespaces[USER].models[
```
nickarocho, line 279: Will `this.schema.namespaces[USER].models` always be truthy here?

packages/datastore/src/sync/outbox.ts
```diff
@@ -252,6 +248,49 @@ class MutationEventOutbox {
...
+			}
+    }
+	]
+	*/
+	private removeTimestampFields(
+		model: string,
+		record: PersistentModel
+	): PersistentModel {
+		const CREATED_AT = 'createdAt';
+		const UPDATED_AT = 'updatedAt';
```
nickarocho, line 274: Nit: I actually like the concise name for this specific usage, but playing devil's advocate - would it make more sense to call this something like `CREATED_AT_LABEL` or `CREATED_AT_KEY` (same for `UPDATED_AT`)?

nickarocho (approved):
LGTM, just left a couple clarifying questions 👍 Thanks for fixing this, Ivan!

packages/datastore/src/sync/outbox.ts
iartemiev: Yes, it will always be defined.

svidgen (approved):
🚢  Just nits as far as I can see! 🚢

PR #9159, opened 2021-11-04
revert: uuid dependency upgrade

iartemiev:
This reverts commits d7b4f49359e4b3964456cdd746f5b4512b35948f, 70eaabdd45b77261f94f76cbca047ebb86da9180, and 7cfe5817be664bf09a8766d579970e3f7fe8ef6e

#### Description of changes

The UUID dependency upgrades broke DataStore on React Native. Reverting for now. We will investigate what additional changes need to be made in order to safely upgrade this dependency.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #8464, opened 2021-06-17
uuid package out of date may cause problems with random numbers


pgbradbury:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React, Next.js

### Amplify APIs

Authentication

### Amplify Categories

auth

### Environment information

<details>

```
# Put output below this line
  System:
    OS: macOS 11.3.1
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
    Memory: 49.06 MB / 8.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 15.7.0 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.4.3 - /usr/local/bin/npm
  Browsers:
    Chrome: 91.0.4472.106
    Safari: 14.1
  npmPackages:
    @ampproject/toolbox-optimizer:  undefined ()
    @aws-amplify/ui-components: ^1.3.1 => 1.4.0 
    @aws-amplify/ui-react: ^1.2.0 => 1.2.2 
    @babel/core:  undefined ()
    @babel/eslint-parser: ^7.12.1 => 7.14.5 
    @babel/eslint-plugin: ^7.12.13 => 7.14.5 
    @babel/plugin-transform-react-jsx: ^7.10.4 => 7.14.5 
    @babel/preset-env: ^7.10.4 => 7.14.5 
    @babel/preset-react: ^7.10.4 => 7.14.5 
    @fullstory/babel-plugin-annotate-react: ^2.1.0 => 2.1.2 
    @fullstory/browser: ^1.4.9 => 1.4.9 
    @lambdatest/node-tunnel: ^3.0.0 => 3.0.1 
    @material-ui/core: ^4.11.0 => 4.11.4 
    @material-ui/icons: ^4.9.1 => 4.11.2 
    @material-ui/lab: ^4.0.0-alpha.56 => 4.0.0-alpha.58 
[...]

PR #9299, opened 2021-12-01
chore: Upgrade to uuid 8.3.2

MatrixFrog:
#### Description of changes

Upgrade uuid to avoid some annoying install-time warnings about potential risks in older uuid's (#8464). Trying this again after previous attempt was rolled back (#9159)

#### Issue #, if available
https://github.com/aws-amplify/amplify-js/issues/8464

ran tests on CI

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


ashika01:
@MatrixFrog we had to revert ur earlier upgrades of uuid because of failure in RN. If you could test it out in RN and let us know how it goes, we could merge this. Or we will investigate this issue in an upcoming cycle and post on how the investigation goes.

ashika01:
@MatrixFrog I will get back on this today. there seems to be an issue in react native in datastore usage in specific. Let me get more detail and ping here.

MatrixFrog:
I think I found the error you're talking about: 

> Error: crypto.getRandomValues() not supported. See https://github.com/uuidjs/uuid#getrandomvalues-not-supported

Will try to look into it further tomorrow.

MatrixFrog:
so i guess the easiest might be to just tell users they need to add `import 'react-native-get-random-values';` before importing the datastore package, which presumably would be a breaking change. kind of annoying react-native doesn't just ship with an implementation of `crypto` that matches other environments :(

MatrixFrog:
This is the same approach taken here: https://github.com/aws/aws-sdk-js-v3/blob/main/README.md#getting-started what do you think?

GeorgeBellTMH:
Is there a way to just to make the polyfill a peerDependency and include it if crypto.getRandomValues() doesn't exist? This way it is transparent to developers.

GeorgeBellTMH:
If it is a breaking change, is there a way to put it behind a feature or version flag...so that new projects, and projects not using datastore can move forward...

mcqj:
Any plan to resolve conflicts and release the fix?

GeorgeBellTMH:
Might be easier to switch to a different uuid generator...react-native-uuid or something?

MatrixFrog:
I'm happy to resolve conflicts if it will help get this merged!

abdallahshaban557:
Hi @MatrixFrog - this change would break existing React Native customers. So we will be holding off on merging it until our next major version bump.

MatrixFrog:
> so i guess the easiest might be to just tell users they need to add `import 'react-native-get-random-values';` before importing the datastore package

is there some kind of release notes file I should add this to?

abdallahshaban557:
@MatrixFrog  - this would actually make more sense to add to our documentation on docs.amplify.aws. We can handle that once the PR is ready to be merged as part of our next major version bump. We do not have an exact ETA on it, but we are tracking this PR as a part of that effort.

cwomack:
@MatrixFrog, this upgrade to the uuid version is now out in v6 of Amplify.  Can you update and validate if the concerns in this PR are addressed?

MatrixFrog:
looks like all the places that a package depends on uuid are now at ^9.0.0

thanks!

PR #9300, opened 2021-12-01
fix(@aws-amplify/datastore): fixes observeQuery in local-only mode

nickarocho:
#### Description of changes

The current implementation of `observeQuery` does not account for using DataStore in local-only mode (no API configured). Since the `isSynced` boolean gets its value from the Sync Engine, the user was receiving the following error when no API was configured and trying to use `DataStore.observeQuery`:
```
TypeError: Cannot read properties of undefined (reading 'getModelSyncedStatus')
```

This PR (though there are a handful of diffs from our friend Prettier), just adds a null/undefined check for the Sync Engine to determine if local-only mode is being utilized. If so, we've come to the conclusion -- also amongst iOS/Android/Flutter teams, for parity-sake -- that `isSynced` should always be `false` when using DataStore locally. 

Observe query will now return snapshots for all consecutive updates, even without an API configured.

#### Issue #, if available

None that I could find

- Spun up the `observeQuery` sample react app
- Commented out the `Amplify.configure(awsconfig)` line in the app's root (to use the app to local-only mode without an API/backend)
- Linked the local repo with these changes
- The error doesn't appear and snapshots are sent accordingly

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- For tests, I'm working on modifying [the current E2E/integ test](https://github.com/aws-amplify/amplify-js-samples-staging/blob/master/cypress/integration/datastore/observe-query.spec.js) to run a test in local-only mode to verify this use-case.


iartemiev (approved):
LGTM 👍🏻

manueliglesias (approved):
LGTM 👍

PR #9308, opened 2021-12-03
feat(in-app-messaging): Add web support for session cap

cshfang:
#### Description of changes

Previously, session tracking for the purposes of Pinpoint in-app messaging campaign session caps only worked with React Native. This PR implements the session tracker for web to extend that feature to the web.

#### Issue #, if available

N/A

Created a web app and tested session cap mechanism to see that:
1. Session caps were enforced
2. Session caps were reset when browser visibility went from inactive to active

Tested with Chrome, Firefox, Safari and Edge

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/notifications/src/InAppMessaging/SessionTracker/SessionTracker.ts
+++ b/packages/notifications/src/InAppMessaging/SessionTracker/SessionTracker.ts
@@ -10,12 +10,65 @@
  * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
  * and limitations under the License.
  */
-
+import { ConsoleLogger as Logger } from '@aws-amplify/core';
 import noop from 'lodash/noop';
-import { SessionStateChangeHandler, SessionTrackerInterface } from './types';
+import {
+	SessionState,
+	SessionStateChangeHandler,
+	SessionTrackerInterface,
+} from './types';
+
+// Per https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API
+let hidden: string;
+let visibilityChange: string;
+
+if (typeof document.hidden !== 'undefined') {
+	hidden = 'hidden';
+	visibilityChange = 'visibilitychange';
+} else if (typeof document['msHidden'] !== 'undefined') {
+	hidden = 'msHidden';
+	visibilityChange = 'msvisibilitychange';
+} else if (typeof document['webkitHidden'] !== 'undefined') {
+	hidden = 'webkitHidden';
+	visibilityChange = 'webkitvisibilitychange';
+}
+
+const logger = new Logger('InAppMessagingSessionTracker');
 
 export default class SessionTracker implements SessionTrackerInterface {
-	constructor(sessionStateChangeHandler: SessionStateChangeHandler = noop) {}
-	start = noop as SessionTrackerInterface['start'];
-	end = noop as SessionTrackerInterface['end'];
+	private sessionStateChangeHandler: SessionStateChangeHandler;
+
+	constructor(sessionStateChangeHandler: SessionStateChangeHandler = noop) {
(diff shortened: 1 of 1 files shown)


packages/notifications/src/InAppMessaging/SessionTracker/SessionTracker.ts
```diff
@@ -10,12 +10,65 @@
...
+if (typeof document.hidden !== 'undefined') {
+	hidden = 'hidden';
+	visibilityChange = 'visibilitychange';
+} else if (typeof document['msHidden'] !== 'undefined') {
+	hidden = 'msHidden';
+	visibilityChange = 'msvisibilitychange';
+} else if (typeof document['webkitHidden'] !== 'undefined') {
+	hidden = 'webkitHidden';
+	visibilityChange = 'webkitvisibilitychange';
+}
```
Samaritan1011001, line 34: Small nit: `document.hidden` and `document['msHidden']` . The way the value is accessed is different. 

And a tiny question, this needs to run only once?

packages/notifications/src/InAppMessaging/SessionTracker/SessionTracker.ts
```diff
@@ -10,12 +10,65 @@
...
+
+	end = (): SessionState => {
+		document.removeEventListener(
+			visibilityChange,
+			this.visibilityChangeHandler
+		);
+		return this.getSessionState();
+	};
+
+	private getSessionState = (): SessionState => {
```
Samaritan1011001, line 58: Observation: return type is `SessionState` but this function returns a string.

Samaritan1011001 (approved):
Thank you! Looks good. Just two questions.

packages/notifications/src/InAppMessaging/SessionTracker/SessionTracker.ts
cshfang: It should only need to run once as it's setting up vendor differences for browsers only.

`document` seems to be typed so that the vendor prefixed versions are not readily available, hence the different accessors as dot notation would error.

packages/notifications/src/InAppMessaging/SessionTracker/SessionTracker.ts
cshfang: Ah the typing for SessionState is as follows:
```typescript
export type SessionState = 'started' | 'ended';
```

Samaritan1011001: Ah! right 👍

packages/notifications/src/InAppMessaging/SessionTracker/SessionTracker.ts
```diff
@@ -10,12 +10,65 @@
...
+	SessionState,
+	SessionStateChangeHandler,
+	SessionTrackerInterface,
+} from './types';
+
+// Per https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API
+let hidden: string;
+let visibilityChange: string;
+
+if (typeof document.hidden !== 'undefined') {
```
elorzafe, line 25: Can `document` be `undefined`?

elorzafe (commented):
I have some concerns if document is undefined, can it be a Native implementation on one for the browser?

packages/notifications/src/InAppMessaging/SessionTracker/SessionTracker.ts
cshfang: As this code should not run on React Native, there should be no reason it's undefined. However, there's no harm in adding a check in case it does run in some unsupported environment for some reason.

elorzafe (approved):
LGTM 🌮

chintannp:
LGTM 🚀

Issue #8781, opened 2021-08-20
Storage.put of multipart file returns successfully upon failure


andriworld:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Not applicable

### Amplify APIs

Storage

### Amplify Categories

Not applicable

### Environment information

<details>

```
 aws-amplify: ^4.2.4 => 4.2.4
```

</details>

### Describe the bug

When Storage.put fails at the end of a multipart upload (CompleteMultipartUpload), the call returns successfully and no error is thrown.

### Expected behavior

An error should be thrown or returned to the caller, indicating that the upload failed.

### Reproduction steps

1. Misconfigure your S3 bucket so that it will fail at the end of a multipart upload. For instance, as long as the CORS settings are not exposing the ETag header, files larger than 5MB will fail to upload. See https://github.com/aws-amplify/amplify-js/issues/61#issuecomment-445436658
2. Invoke Storage.put with a file that will cause the multipart upload to fail at the end. For instance, upload a file larger than 5MB to a bucket misconfigured as in step 1.
3. OBSERVATION: an error is correctly logged:
"AWSS3ProviderManagedUpload - error happened while finishing the upload. Cancelling the multipart upload Error: Request failed with status code 400"
However, the call returns successfully, and no error is thrown.

### Code Snippet

```javascript
const object = await Storage.put(uploadPath, file, {
      level: 'private',
      contentType: file.type
  })
  console.log('object: ' + JSON.stringify(object))
```

### Log output

<details>

```
18:36.583 axios-http-handler - Request failed with status code 400
18:36.584 AWSS3ProviderManagedUpload - error happened while finishing the upload. Cancelling the multipart upload Error: Request failed with status code 400
  | push../node_modules/@aws-amplify/core/lib-esm/Logger/ConsoleLogger.js.ConsoleLogger._log | @ | ConsoleLogger.js:139
-- | -- | -- | --
[...]

PR #9317, opened 2021-12-06
fix(@aws-amplify/storage): throw error if all upload parts complete but upload cannot be finished

katiegoines:
#### Description of changes

Added error thrown when multi-part upload cannot be finalized. 

#### Issue #, if available

Issue #8781 

Tested in sample app for multi-part upload (both resumable and non-resumable) 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/storage/__tests__/providers/AWSS3ProviderManagedUpload-unit-test.ts
```diff
@@ -335,13 +341,6 @@ describe('multi part upload tests', () => {
...
 		});
-		const loggerSpy = jest.spyOn(Logger.prototype, '_log');
-		const uploader = new TestClass(testParams, testOpts, new events.EventEmitter());
-		await uploader.upload();
-		expect(loggerSpy).toHaveBeenCalledWith(
-			'ERROR',
-			'error happened while finishing the upload. Cancelling the multipart upload',
-			'error'
-		);
+		expect(Promise.reject(new Error('error'))).rejects.toThrow('error');
```
jamesaucode, line 344: Why did we remove this block of code ? Also I think if you reject with an error with the message 'error' the assertion here is always true.

packages/storage/src/providers/AWSS3ProviderManagedUpload.ts
```diff
@@ -74,51 +74,56 @@ export class AWSS3ProviderManagedUpload {
...
+						parts.slice(start, start + this.queueSize)
+					);
 
-			parts.map(part => {
-				this.removeEventListener(part);
-			});
+					/** Call cleanup a second time in case there were part upload requests
+					 *  in flight. This is to ensure that all parts are cleaned up.
+					 */
+					await this.checkIfUploadCancelled(uploadId);
```
jamesaucode, line 114: Do we still need to have these calls to `this.checkIfUploadCancelled`? We can probably just move the logic of that function into the `catch` block below.

packages/storage/src/providers/AWSS3ProviderManagedUpload.ts
```diff
@@ -74,51 +74,56 @@ export class AWSS3ProviderManagedUpload {
...
+				parts.map(part => {
+					this.removeEventListener(part);
+				});
+
+				// Step 3: Finalize the upload such that S3 can recreate the file
+				return await this.finishMultiPartUpload(uploadId);
+			}
+		} catch (error) {
+			logger.error('Error. Cancelling the multipart upload. ', error.message);
+			throw error.message;
```
jamesaucode, line 126:
```suggestion
			throw error;
```

I think we should throw the error as is. correct me if im wrong but I believe there will be no stack trace if you throw a string

jamesaucode (changes requested):
Left a few comments, let me know if you any questions 😄

Legionses:
@katiegoines @jamesaucode Hello, guys! Can u complete this fix please?  Seems like error in Storage.put of multipart file is a big issue tat needed to be fixed.

katiegoines:
@jamesaucode - Having trouble with the `cleanup`. Once the upload is aborted, there is no longer an upload to find with ListParts, which is at odds with the `'error case: upload a string as body that splits in two parts but second part fails'` test. Your guidance would be greatly appreciated. Thank you!

packages/storage/__tests__/providers/AWSS3ProviderManagedUpload-unit-test.ts
```diff
@@ -320,7 +320,13 @@ describe('multi part upload tests', () => {
...
 		});
 		const uploader = new TestClass(testParams, testOpts, new events.EventEmitter());
-		await expect(uploader.upload()).rejects.toThrow('Upload was cancelled. Multi Part upload clean up failed');
+		try {
+			await uploader.upload();
+		} catch (error) {
+			expect(error).toBe(
+				'Upload was cancelled. Multi Part upload clean up failed'
+			);
+		}
```
jamesaucode, line 329: For this part, we don't need to change this to a `try/catch` block, we want to assert that the error thrown is an `Error` with `expect(...).rejects.toThrowError(...)`

packages/storage/src/providers/AWSS3ProviderManagedUpload.ts
```diff
@@ -74,51 +76,49 @@ export class AWSS3ProviderManagedUpload {
...
+					this.removeEventListener(part);
+				});
 
-			// Step 3: Finalize the upload such that S3 can recreate the file
-			return await this.finishMultiPartUpload(uploadId);
+				// Step 3: Finalize the upload such that S3 can recreate the file
+				return await this.finishMultiPartUpload(this.uploadId);
+			}
+		} catch (error) {
+			// if any error is thrown, call cleanup
```
jamesaucode, line 118: I believe with your new implementation, `this.cancel` is no longer needed. It used to act as a signal for the main `this.upload` function to do a cleanup, now instead we can just have the `catch` block do the cleanup.

packages/storage/src/providers/AWSS3ProviderManagedUpload.ts
```diff
@@ -191,7 +191,7 @@ export class AWSS3ProviderManagedUpload {
 			}
 		} catch (error) {
 			logger.error(
-				'error happened while uploading a part. Cancelling the multipart upload',
+				'Error happened while uploading a part. Cancelling the multipart upload',
```
jamesaucode, line 194: I think we should remove the `this.cancelUpload()` function call here, and also throw an error.

jamesaucode (changes requested):
Thanks for the PR 😃  just a few more changes

jamesaucode:
> @jamesaucode - Having trouble with the `cleanup`. Once the upload is aborted, there is no longer an upload to find with ListParts, which is at odds with the `'error case: upload a string as body that splits in two parts but second part fails'` test. Your guidance would be greatly appreciated. Thank you!



> @jamesaucode - Having trouble with the `cleanup`. Once the upload is aborted, there is no longer an upload to find with ListParts, which is at odds with the `'error case: upload a string as body that splits in two parts but second part fails'` test. Your guidance would be greatly appreciated. Thank you!

I think the problem is 
1. We should throw an error in `.uploadParts` as well
2. In the test file, change line 309 so `UploadPartsCommand` resolves instead of rejecting with an error.
Happy to sync with you offline if you need more information.

jamesaucode (approved):
LGTM! 🎉

packages/storage/src/providers/AWSS3ProviderManagedUpload.ts
```diff
@@ -60,6 +60,7 @@ export class AWSS3ProviderManagedUpload {
 	private completedParts: CompletedPart[] = [];
 	private cancel = false;
```
stocaaro, line 61: Looks like this variable isn't used anymore, right?

stocaaro (approved):

packages/storage/src/providers/AWSS3ProviderManagedUpload.ts
katiegoines: Correct. Thank you for catching!

Issue #9208, opened 2021-11-12
Having ?code= on the query string seems to always be caught by some Cognito code


PeteDuncanson:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication

### Amplify Categories

auth

### Environment information

<details>

```
# Put output below this line
  System:
    OS: Windows 10 10.0.19041
    CPU: (16) x64 Intel(R) Core(TM) i7-10875H CPU @ 2.30GHz
    Memory: 11.16 GB / 31.84 GB
  Binaries:
    Node: 14.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.10 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 6.14.11 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 94.0.4606.81
    Edge: Spartan (44.19041.1266.0), Chromium (95.0.1020.44)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    @asseinfo/react-kanban: ^2.1.0 => 2.1.0
    @aws-amplify/ui-react: ^1.2.21 => 1.2.21
    @brainhubeu/react-file-input: 0.7.21 => 0.7.21
    @headlessui/react: ^1.2.0 => 1.2.0
    @ideal-postcodes/api-typings: ^2.0.1 => 2.0.1
    @ideal-postcodes/core-browser: ^1.5.2 => 1.5.2
    @sentry/react: ^6.3.0 => 6.3.0
    @sentry/tracing: ^6.2.4 => 6.2.4
    @tailwindcss/aspect-ratio: ^0.2.0 => 0.2.0
    @tailwindcss/forms: ^0.2.1 => 0.2.1
    @tailwindcss/typography: ^0.4.0 => 0.4.0 (0.2.0)
    @tailwindcss/ui: ^0.6.2 => 0.6.2
    @tailwindui/react: ^0.1.1 => 0.1.1
[...]

PR #8198, opened 2021-04-29
fix(@aws-amplify/auth): Handle URIs with no path part

manueliglesias:
#### Description of changes
Handle URIs with no path part
For full context see: https://github.com/aws-amplify/amplify-js/issues/7468#issuecomment-828890848

#### Issue #, if available
Fixes #7468

#### Description of how you validated changes
- Unit test
- React Native app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #8166, opened 2021-04-23
Dispose of ?code on error and success

ericclemmons:
#### Description of changes

<http://localhost:3000/?code=...>, when an error occurs, stays in the URL. On reload, this will throw a `invalid_grant` error because the `?code` has already been used.

### 1. Notice how `?code` stays in the URL and errors are thrown:

> Screen Shot 2021-04-23 at 2 46 07 PM

### 1. With this PR, `window.location.href` matches `redirectUri`. Notice how the original request was for `/?code=...`!

> Screen Shot 2021-04-23 at 2 48 37 PM

### 2. Errors still exist in the console, as usual:

> Screen Shot 2021-04-23 at 2 48 44 PM

### 3. But now reloading the window (in Chrome) does not cause additional errors:

> Screen Shot 2021-04-23 at 2 48 49 PM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)
- [x] ~~Relevant documentation is changed or added (and PR referenced)~~

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #7468, opened 2020-12-30
Federated facebook login in react native requires two login attempts to work on IOS


erikgaas:
**Describe the bug**
I am working through the tutorial for federated login here: https://docs.amplify.aws/lib/auth/social/q/platform/js#deploying-to-amplify-console

I am using the "In-App Browser Setup (optional, but recommended)" section to federate login into facebook from cognito.

This works fine on android with no trouble. The problem is testing it on the simulator for IOS. Following the tutorial based on auth command actually doesn't work. `Auth.federatedSignIn()` is the command the docs say to use. (Again this works fine in Android) If I use federatedSignIn without arguments it leads to this error every time I try to sign in.

```Sign in failure [ValidationException: 1 validation error detected: Value '{cognito-idp.us-west-2.amazonaws.com/us-west-2_rBVomGdCO=}' at 'logins' failed to satisfy constraint: Map value must satisfy constraint: [Member must have length less than or equal to 50000, Member must have length greater than or equal to 1]]```

Now if I change this to `Auth.federatedSignIn({provider: 'Facebook'})` then if I try signing in the first time, it will log the error above, but if I sign in again then everything works fine.

I'm pretty sure this is due to some weird mismatch in authentication state. It seems as though the first attempt to login initializes something important on the frontend that is able to send the correct request on the second auth call.

I have tried to hunt down the problem, presumably there is something I could to do inspect the auth state over time, but for now I've been looking at the requests through urlOpener. The only thing that is different is the `newUrl` returned by the InAppBrowser.openAuth call.

First call it is: appname:///?code=<uuid>&state=<state_id>#_=_
and second call is
appname://?code=<uuid>&state=<state_id>

So I guess the only difference is that #_=_ at the very end which makes zero sense to me. Given that the requests are the same to the backend and only the responses change it makes me think there is actually a backend state being mismanaged...???

**To Reproduce**

1. Try federated login with facebook using the recommended but changing federated login code to specify facebook as provider
2. Login once and observe error while being returned back to login screen
3. Try logging in again and have it succeed showing user information

Alternatively try not specifying provider and always see it fail

**Expected behavior**
I expect to be successfully authenticated such that Hub is able to listen for a successful authentication

**Code Snippet**
This is the crux of where the command is flaky. In federatedSignIn
```
  return (
    <View>
      <Text>User: {user ? JSON.stringify(user.attributes) : 'None'}</Text>
      {user ? (
[...]

PR #9370, opened 2021-12-16
prevent redirect if callback is unregistered

hkjpotato:
#### Description of changes

https://github.com/aws-amplify/amplify-js/issues/9208 describes a situation when a third party `code` parameter is passed back to the web app, under a sub route different then the registered redirect URL in Cognito. In this case, we should not redirect back to root url blindly, and thus allow the customer to handle the third party code him/herself.

It should be noticed this issue was introduced after the change in https://github.com/aws-amplify/amplify-js/pull/8166, which will attempt to redirect to the route url, even error occurs (see below screenshot on why error occurs).

A careful look reveals that:
1. Instead of reverting it, we should detect the unregistered route, and only avoid redirecting in that case (thus keep the desired change in https://github.com/aws-amplify/amplify-js/pull/8166)
2. The current solution assuming such a "contract" of the `_handleCodeFlow` function.
   - As previous, if the input `code` is empty, return silently.
   - Yet if the input URL is not the registered, throw an explicit error.

Throwing an error at `_handleCodeFlow` function is essential, as it can help skip https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/Auth.ts#L2062-L2078, which attempts to set a session with empty tokens, and makes a failure call to the identity pool to fetch credentials.

Screen Shot 2021-12-16 at 5 05 26 PM
Screen Shot 2021-12-16 at 5 05 50 PM

None of the above steps should happen if the `code` parameter is attached to an unregistered callback path (why use a parameter not for Cognito to call Cognito API?). Even eventually it will fail (with a network error), it is failing too late. The invalid url should be captured and thus prevent Auth from assuming an invalid session.

Note this PR is open for early opinions, and still pending for more testing. Specifically, I would like to test on React Native, due to a related issue
https://github.com/aws-amplify/amplify-js/issues/7468
https://github.com/aws-amplify/amplify-js/pull/8198

#### Issue #, if available

9208

TBD

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -2131,10 +2131,18 @@ export class AuthClass {
 					return credentials;
 				} catch (err) {
 					logger.debug('Error in cognito hosted auth response', err);
-
+					
+					const isUnregisteredCallback =
+						err && err.message === 'Unregistered OAuth Callback';
 					// Just like a successful handling of `?code`, replace the window history to "dispose" of the `code`.
 					// Otherwise, reloading the page will throw errors as the `code` has already been spent.
-					if (window && typeof window.history !== 'undefined') {
+					// Do not redirect if the callback is not registered in oauth config, as it can be used by the third party
+					// see https://github.com/aws-amplify/amplify-js/issues/9208
+					if (
+						window &&
+						typeof window.history !== 'undefined' &&
+						!isUnregisteredCallback
+					) {
 						window.history.replaceState(
 							{},
 							null,
--- a/packages/auth/src/OAuth/OAuth.ts
+++ b/packages/auth/src/OAuth/OAuth.ts
@@ -129,7 +129,12 @@ export default class OAuth {
 		const redirectSignInPathname =
 			parse(this._config.redirectSignIn).pathname || '/';
 
-		if (!code || currentUrlPathname !== redirectSignInPathname) {
+		if (currentUrlPathname !== redirectSignInPathname) {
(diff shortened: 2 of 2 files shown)


elorzafe:
Thanks @hkjpotato 

I would do it a little different. I would prevent calling [`this._handleAuthResponse(url);`](https://github.com/aws-amplify/amplify-js/blob/da8cd4c691c92f43ee920505230c8f4a54fdb5e4/packages/auth/src/Auth.ts#L250
) if the `url` is not on OAuth configuration. 

What do you think?

dkryptr:
What's the status of this PR? It appears the original issue is still a problem. Hoping this can get some traction.

abdallahshaban557:
@hkjpotato - are you able to address the feedback left by @elorzafe ? Do you need further support to push this PR through?

abdallahshaban557:
@hkjpotato  - did you get a chance to check the feedback from @elorzafe ?

cwomack:
@hkjpotato, wanted to follow up on this PR.  It looks like this is no longer an issue tied to the most recent version of Amplify (v6 at the time of this comment) due to OAuth flows that initiate outside of Amplify no longer being supported (as captured in Feature Request #13343 to get that back in). 

However, this issue still exists in v5 and lower as captured in #9208.  I'll close out this PR since we haven't heard back since the comment from Sep 2022 above and due to the fact that we have both the existing bug in lower versions as well as the feature parity issue in the most recent version captured in existing issues.

Issue #8754, opened 2021-08-13
Cannot use an empty string as translation


ankon:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Internationalization

### Amplify Categories

Not applicable

### Environment information

<details>

```
  System:
    OS: Linux 5.13 Fedora 34 (Thirty Four) 34 (Thirty Four)
  Binaries:
    Node: 14.16.0 - ~/modules/nvs/node/14.16.0/x64/bin/node
    Yarn: 1.22.11 - ~/modules/nvs/node/14.16.0/x64/bin/yarn
    npm: 6.14.11 - ~/modules/nvs/node/14.16.0/x64/bin/npm
  Browsers:
    Brave Browser: nightly
    Chrome: 92.0.4515.131
    Firefox: 90.0.2
  npmPackages:
    @apollo/client: ^3.4.0 => 3.4.0 
    @apollo/client/cache:  undefined ()
    @apollo/client/core:  undefined ()
    @apollo/client/errors:  undefined ()
    @apollo/client/link/batch:  undefined ()
    @apollo/client/link/batch-http:  undefined ()
    @apollo/client/link/context:  undefined ()
    @apollo/client/link/core:  undefined ()
    @apollo/client/link/error:  undefined ()
    @apollo/client/link/http:  undefined ()
    @apollo/client/link/persisted-queries:  undefined ()
    @apollo/client/link/retry:  undefined ()
    @apollo/client/link/schema:  undefined ()
    @apollo/client/link/utils:  undefined ()
    @apollo/client/link/ws:  undefined ()
    @apollo/client/react:  undefined ()
    @apollo/client/react/components:  undefined ()
    @apollo/client/react/context:  undefined ()
    @apollo/client/react/data:  undefined ()
    @apollo/client/react/hoc:  undefined ()
    @apollo/client/react/hooks:  undefined ()
    @apollo/client/react/parser:  undefined ()
    @apollo/client/react/ssr:  undefined ()
[...]

PR #9403, opened 2021-12-25
fix(@aws-amplify/core): use empty string as translation

yana-katty:
#### Description of changes

This PR fixes the I18n `get` function return an empty string, if provided vocabulary contains an empty string entry.
Only fallback to the default value if the vocabulary doesn't contain such a key at all.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/8754

 * Local unit testing
 * Verified that get function return an empty string in my local app

test code
```
    I18n.setLanguage('fr');
    const dict = {
        'fr': {
            'empty string': "",
            'Sign Up': "S'inscrire"
        }
    };
    I18n.putVocabularies(dict);
    console.log(I18n.get('empty string'));// returnig empty string
    console.log(I18n.get('Sign Up'));// returnig provided vocabulary ( "S'inscrire" )
    console.log(I18n.get('Sign in'));//returnig key value('Sign in')
    console.log(I18n.get('Sign in','Se connecter'));//returnig default value ('Se connecter')
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/src/I18n/I18n.ts
+++ b/packages/core/src/I18n/I18n.ts
@@ -78,14 +78,14 @@ export class I18n {
 
 		const lang = this._lang;
 		let val = this.getByLanguage(key, lang);
-		if (val != null) {
+		if (val !== undefined) {
 			return val;
 		}
 
 		if (lang.indexOf('-') > 0) {
 			val = this.getByLanguage(key, lang.split('-')[0]);
 		}
-		if (val != null) {
+		if (val !== undefined) {
 			return val;
 		}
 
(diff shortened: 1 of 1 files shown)


evcodes (approved):
Hey @yana-katty thank you so much for this PR. I just pulled down the changes locally and this is a great fix. I see how this would be an issue and I appreciate your commit. Green light from me 🚀

evcodes (changes requested):
!= with null, will also work if val is undefined since it's not strictly unequal. I am working on a fix.

evcodes (approved):
This works now

nickarocho (approved):
LGTM! Thank you for your contribution! 🤘

PR #9419, opened 2022-01-02
fix: added type declaration to associate secret code for MFA

lousybear:
required for authenticating with MFA option enabled

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amazon-cognito-identity-js/index.d.ts
+++ b/packages/amazon-cognito-identity-js/index.d.ts
@@ -24,6 +24,7 @@ declare module 'amazon-cognito-identity-js' {
 			userAttributes: any,
 			requiredAttributes: any
 		) => void;
+		associateSecretCode?: (challengeName: any, challengeParameters: any) => void;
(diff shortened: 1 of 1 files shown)


ashika01:
@lousybear Help me understand this better. Why do we need the `associateSecretCode` added? I see the only place that is used is inside `associateSoftwareToken` and to use TOTP MFA wouldn't one be using that?

lousybear:
@ashika01 Yes, as you mentioned it is required for TOTP based MFA workflow, since it is not currently defined as a valid callback function type we might need to get this added to get the workflow running.

elorzafe:
@lousybear the README can be misleading because is passing `this` on the callback for `associateSoftwareToken` but that callback is not needed on the original call. At the end is required to do `cognitoUser.associateSoftwareToken(callbacks)`.

If that is part of the type unless you do that "binding" it wont have any effect.

Let me know what you think.

abdallahshaban557:
@lousybear - was the description from @elorzafe sufficient for you? Should we close this PR?

PR #9428, opened 2022-01-04
feat(@aws-amplify/auth): Delete user API exposed

evcodes:
This PR is in draft however, I will be writing unit and e2e tests for this feature. I will manually test within react native as well.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [in progress] Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)
- [in progress] Relevant documentation is changed or added (and PR referenced)


packages/core/src/I18n/I18n.ts
```diff
@@ -78,14 +78,14 @@ export class I18n {
 
 		const lang = this._lang;
 		let val = this.getByLanguage(key, lang);
-		if (val) {
+		if (val != null) {
```
elorzafe, line 81: I would make this into a separate PR

packages/auth/src/Auth.ts
```diff
@@ -1120,24 +1121,56 @@ export class AuthClass {
...
+						if (err) {
+							logger.debug('Failed to get the user session', err);
+							rej(err);
+						} else {
+							await user.deleteUser((err, result) => {
+								if (err) {
+									rej(err);
+								} else {
+									user.signOut();
+									logger.debug(result);
```
elorzafe, line 1163: I am not sure about the implications when delete a user after sign in with HostedUI, can you check that behavior?

I think you will ned to this after `user.signOut()`
```javascript
dispatchAuthEvent('signOut', this.user, `A user has been signed out`);
this.user = null;
```
similar to what happens [here](https://github.com/aws-amplify/amplify-js/blob/c09eec49d9169d05186c97ec1b38b198afe70cd8/packages/auth/src/Auth.ts#L1731)

elorzafe (changes requested):
Thanks @evcodes !

I would check `deleteUser` for HostedUI use case and I would add `dispatchAuthEvent` as well.

packages/core/src/I18n/I18n.ts
evcodes: agh I thought that I removed all the work from a separate branch

packages/auth/src/Auth.ts
hkjpotato: The signout flow could be a bit confusing considering different entry points/flows: some ideas here just for reference:
1. In a hosted UI signout, it will have a callback to redirect to "signout url" https://github.com/aws-amplify/amplify-js/blob/c09eec49d9169d05186c97ec1b38b198afe70cd8/packages/auth/src/Auth.ts#L1692
2. there are another concern in signout flow: remove local storage items https://github.com/aws-amplify/amplify-js/blob/c09eec49d9169d05186c97ec1b38b198afe70cd8/packages/auth/src/Auth.ts#L129
3. maybe we can consider to call the higher function, which already encapsulate all these logics. https://github.com/aws-amplify/amplify-js/blob/c09eec49d9169d05186c97ec1b38b198afe70cd8/packages/auth/src/Auth.ts#L1731
4. or as @elorafe suggested, functions can be hook up by "event". Potentially this can help us decouple the nested logics, and keep their single responsibilities. However it might requires more refactor so not necessary something we need to do now.

jamesaucode (commented):
just some nitpicks 😃

packages/auth/src/Auth.ts
```diff
@@ -1120,24 +1121,62 @@ export class AuthClass {
...
+	 * @return {Promise}
+	 **/
+	public async deleteUser(): Promise<void> {
+		return new Promise((res, rej) => {
+			if (this.userPool) {
+				const user = this.userPool.getCurrentUser();
 
+				if (!user) {
+					logger.debug('Failed to get user from user pool');
+					rej('No current user');
```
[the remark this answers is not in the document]
ashika01: +1

packages/auth/src/Auth.ts
ashika01: +1

packages/auth/src/Auth.ts
```diff
@@ -1120,24 +1121,62 @@ export class AuthClass {
...
+					user.getSession(async (err, session) => {
+						if (err) {
+							logger.debug('Failed to get the user session', err);
+							rej(err);
+						} else {
+							await user.deleteUser((err, result) => {
+								if (err) {
+									rej(err);
+								} else {
+									user.signOut();
```
ashika01, line 1162: So when u try this out in an app, does it even allow you to call `signOut`, if the user is deleted?

i think to avoid confusion here, would it be better to explicitly clear any cache and session in here ?

evcodes: funny enough the user.signOut() does a lot of work behind the scenes which I am not sure if we would want to include that explicitly:

```
clearCachedUser() {
		this.clearCachedTokens();
		this.clearCachedUserData();
	}
```

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -658,9 +661,10 @@ describe('auth unit test', () => {
 
 			await auth.resendSignUp('username');
 
-			expect(
-				await CognitoUser.prototype.resendConfirmationCode
-			).toBeCalledWith(jasmine.any(Function), { foo: 'bar' });
+			expect(await CognitoUser.prototype.resendConfirmationCode).toBeCalledWith(
+				jasmine.any(Function),
```
ashika01, line 665: i know this is not an introduced change. But we should try to move away frm using jasmine and make these pure jest tests

packages/auth/src/Auth.ts
ashika01: Sorry I kinda wasn't clear. So if u check our Auth signout - we do `cleanCachedItems` to clear cached creds. Dont we have to do that too if we are signing a customer out? Did we decide not to do that?

ashika01: Also, `user.deleteUser` already does this `clearCachedUser` clean out.

https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/src/CognitoUser.js#L1106

So what ever reminder cleanup we want to do we can do it here explicitly than calling signout i think. What are we trying to achieve to do with this signOut?

packages/auth/src/Auth.ts
```diff
@@ -1120,24 +1121,64 @@ export class AuthClass {
...
 			});
 		});
+	}
 
+	/**
+	 * Delete the current authenticated user
+	 * @return {Promise}
+	 **/
+	public async deleteUser(): Promise<void | string> {
+		// const that = this;
```
ashika01, line 1145: nit: needs to be removed?

ashika01:
It seems like a hub event for deleteUser is missed here, There is only one for sign out.

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -2728,11 +2733,109 @@ describe('auth unit test', () => {
...
+				.mockImplementation(() => {
+					return user;
+				});
+			const spy2 = jest
+				.spyOn(CognitoUser.prototype, 'getSession')
+				.mockImplementation((callback: any) => {
+					return callback(null, session);
+				});
+
+			expect(await auth.deleteUser()).toBe('SUCCESS');
```
ashika01, line 2760: could u also check for if the session is cleared out?

packages/core/src/I18n/I18n.ts
```diff
@@ -105,7 +105,7 @@ export class I18n {
 		}
 
 		const lang_dict = this._dict[language];
-		if (!lang_dict) {
+		if (typeof lang_dict === 'undefined') {
```
elorzafe, line 108: I would revert this change from the PR

packages/auth/src/Auth.ts
```diff
@@ -1120,24 +1121,64 @@ export class AuthClass {
...
+				const user = this.userPool.getCurrentUser();
+
+				if (!user) {
+					logger.debug('Failed to get user from user pool');
+					rej(new Error('No current user'));
+				} else {
+					user.getSession(async (err, session) => {
+						if (err) {
+							logger.debug('Failed to get user session', err);
+							rej(new Error('User session does not exist'));
```
elorzafe, line 1157: I would change this error, in case the client is offline is a different problem. Can you check the err message and map it to the correct one.

elorzafe (commented):
Thanks for the update, can you check the error messages from `user.getSession`?
and revert the change from I18n please

packages/auth/__tests__/auth-unit-test.ts
evcodes: This can be slated as an operational excellence item for the future as I have no context to what Jasmine is...

packages/auth/src/Auth.ts
evcodes: I just tested without the user.signOut call and it works the same so I am going to remove it for now. Thank you Ashika!

packages/auth/src/Auth.ts
ashika01: talked offline

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -2728,11 +2733,116 @@ describe('auth unit test', () => {
...
+				});
+			const spy2 = jest
+				.spyOn(CognitoUser.prototype, 'getSession')
+				.mockImplementation((callback: any) => {
+					return callback(null, null);
+				});
+
+			console.log(await auth.deleteUser());
+
+			// expect(await auth.deleteUser()).toThrowError();
```
ashika01, line 2813: this needs to be changed.

ashika01:
Fully ran integ test report here - https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/10026/workflows/dc0959e9-6573-43c0-aaf1-9a0d1aec2d71

ashika01 (approved):

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -2728,11 +2733,118 @@ describe('auth unit test', () => {
...
+			expect(await auth.deleteUser()).toBe('SUCCESS');
+
+			// checking if session is cleared
+			user.getSession((err, session) => {
+				const idToken = new CognitoIdToken({ IdToken: '' });
+				const accessToken = new CognitoIdToken({ IdToken: '' });
+				expect(session).toStrictEqual(
+					new CognitoUserSession({ IdToken: idToken, AccessToken: accessToken })
+				);
+			});
```
elorzafe, line 2769: This expect is probably not validated by jest because is inside of the callback. Also is testing the mock and not the actual code. I would remove this.

packages/auth/src/Auth.ts
```diff
@@ -1120,24 +1121,90 @@ export class AuthClass {
...
+									if (isSignedInHostedUI) {
+										this.oAuthSignOutRedirect(res, rej);
+									} else {
+										res(result);
+									}
+								}
+								dispatchAuthEvent(
+									'signOut',
+									this.user,
+									`A user has been signed out`
```
elorzafe, line 1197: This should be inside of `else` statement otherwise it can be triggered in case `deleteUser` fails.

elorzafe (changes requested):
A couple minor changes, however the promise not awaited and the dispachEvent is a blocker

Overall looks very good

packages/auth/src/Auth.ts
```diff
@@ -1120,24 +1121,90 @@ export class AuthClass {
...
+	 **/
+	public async deleteUser(): Promise<Error | string | void> {
+		new Promise(async (res, rej) => {
+			try {
+				await this._storageSync;
+			} catch (e) {
+				logger.debug('Failed to sync cache info into memory', e);
+				return rej(new Error(e));
+			}
+		});
```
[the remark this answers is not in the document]
evcodes: so should there be an `await` before the` new Promise(...`?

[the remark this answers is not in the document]
jamesaucode: I don't think this new Promise is being used anywhere. It's probably better if we just remove the whole `new Promise`, and throw an Error if it fails. Similar to how it's done here in `currentAuthenticatedUser`
https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/Auth.ts#L1373-L1378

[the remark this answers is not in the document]
hkjpotato: agree with James, a promise is not required here, just normal try-catch.

packages/auth/src/Auth.ts
```diff
@@ -1120,24 +1121,89 @@ export class AuthClass {
...
+				});
 			});
 		});
+	}
 
+	/**
+	 * Delete the current authenticated user
+	 * @return {Promise}
+	 **/
+	public async deleteUser(): Promise<Error | string | void> {
```
jamesaucode, line 1144: Do we need to specify the `Promise<Error>` in our return type? It seems inconsistent with our other API and might be harder for our customers to use.

jamesaucode: Pulled this locally, it seems like just `Promise<string>` will be sufficient instead of a union of `Error | string | void`

ashika01: hmm.. lets chat on this one. I thought void is unnecessary too
but eddy brought a point around oAuthredirect

jamesaucode (approved):
lgtm! 🎉

elorzafe (approved):
Great work!! lets ship it 🌮

PR #9450, opened 2022-01-07
refactor: Prevent show "EventEmitter.removeListener: Method has been deprecated" warning for modern react-native versions

retyui:
#### Description of changes

Fix: https://github.com/aws-amplify/amplify-js/issues/9492

I use

- `"react-native": "0.66.x",`
- `@aws-amplify/auth": "^3.4.34",`

And on every start a see this annoying warning 

```bash
 WARN  EventEmitter.removeListener('url', ...): Method has been deprecated. Please instead use `remove()` on the subscription returned by `EventEmitter.addListener`.
```

Then I check that implementation and find out that subscription impossible for function because

```ts
let handler;

export default async callback =>{
  	if (handler) { // this code make not possible to unsubscribe after first call
		return; 
	}
        // ....
        Linking.removeEventListener('url', handler);
        subscription?.remove?.();
}
```

So I decided to remove never reached unsubscription calls

covered with tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/auth/__tests__/urlListener.native-test.ts
+++ b/packages/auth/__tests__/urlListener.native-test.ts
@@ -1,0 +1,96 @@
+import { Linking, AppState } from 'react-native';
+
+jest.mock('@aws-amplify/core', () => ({
+	ConsoleLogger: class MockConsoleLogger {
+		debug = jest.fn();
+	},
+}));
+
+jest.mock('react-native', () => ({
+	Linking: {
+		addEventListener: jest.fn(),
+		getInitialURL: jest.fn(),
+	},
+	AppState: {
+		addEventListener: jest.fn(),
+	},
+}));
+
+describe('urlListener', () => {
+	const delay = (ms: number) => new Promise(ok => setTimeout(ok, ms));
+	const mockLinking = () => {
+		let linkingListener;
+		let appStateListener;
+
+		(AppState.addEventListener as jest.Mock).mockImplementation(
+			(event, callback) => {
+				appStateListener = callback;
+			}
+		);
+
+		(Linking.addEventListener as jest.Mock).mockImplementation(
+			(eventName, _listener) => {
+				linkingListener = _listener;
+			}
+		);
+
+		return {
+			emitAppStateChange: event => appStateListener(event),
+			emitLinkingChange: event => linkingListener(event),
+		};
+	};
+
+	it('should add event listeners once', done => {
+		jest.isolateModules(async () => {
+			const urlListener = require('../src/urlListener.native').default;
+
+			await urlListener(jest.fn());
+			await urlListener(jest.fn());
+
+			expect(Linking.addEventListener).toHaveBeenCalledTimes(1);
+			expect(AppState.addEventListener).toHaveBeenCalledTimes(1);
+
+			done();
+		});
+	});
+
+	it('should invoke passed callback with url when change', done => {
+		jest.isolateModules(async () => {
+			const { emitLinkingChange } = mockLinking();
+			const url = 'https://aws.amazon.com/mock_url_1';
+			const callback = jest.fn();
+			const urlListener = require('../src/urlListener.native').default;
+
+			await urlListener(callback);
+
+			emitLinkingChange({ url });
+
+			expect(callback).toHaveBeenCalledWith({ url });
+
+			done();
+		});
+	});
+
+	it('should invoke passed callback with url app when app was opened by link', done => {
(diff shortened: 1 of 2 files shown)


retyui:
**Other notes:**

When you check history https://github.com/facebook/react-native/commits/main/Libraries/Linking/Linking.js

image


you will noticed that `removeListener` was removed but then restored (for backward compatibility) 

image

so the code above will be `true` always for all **react-native** versions that currently released (`0.67.x - 0.60.x`)

packages/auth/src/urlListener.native.ts
```diff
@@ -22,29 +22,19 @@ export default async callback => {
...
 
-	// Handles backward compatibility. removeEventListener is only available on RN versions before 0.65.
-	if (Linking.removeEventListener === typeof 'function') {
-		Linking.removeEventListener('url', handler);
-		Linking.addEventListener('url', handler);
-	} else {
-		// remove() method is only available on RN v0.65+.
-		subscription?.remove?.();
-		subscription = Linking.addEventListener('url', handler);
-	}
```
calebpollman, line 47: Removing `subscription` **should** be fine, but this conditional block is required in order to prevent React Native versions prior to 0.65 (see [v0.64-stable](https://github.com/facebook/react-native/blob/0.64-stable/Libraries/Linking/Linking.js#L47)) from crashing

calebpollman (changes requested):
Thanks for submitting the PR and really appreciate the unit tests 🤩  Left some feedback on the changes.

Regarding the deprecated warning for `Linking.removeListener` when using React Native `v0.66.x` , have you considered leveraging `YellowBox.ignoreWarnings`?  e.g.
```
import { YellowBox } from 'react-native';
YellowBox.ignoreWarnings(['Warning: ...']);
```

packages/auth/__tests__/urlListener.native-test.ts
```diff
@@ -0,0 +1,96 @@
...
+		addEventListener: jest.fn(),
+		getInitialURL: jest.fn(),
+	},
+	AppState: {
+		addEventListener: jest.fn(),
+	},
+}));
+
+describe('urlListener react-native', () => {
+	const delay = (ms: number) => new Promise(ok => setTimeout(ok, ms));
```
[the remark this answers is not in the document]
retyui: it is not possible to manipulate an ordering of promise chain using `useFakeTimers`!

packages/auth/src/urlListener.native.ts
retyui: - **First of all** `Linking.addEventListenerI()` will be always return  `EmitterSubscription` object  regardless of RN version

see below [source code 0.63.x](https://unpkg.com/browse/react-native@0.63.0/Libraries/vendor/emitter/EventEmitter.js) with flow types

image

---

- **And second thing**, that I tried to explain that unsubscribe will **NEVER** be invoked 

Let try to debug what written currently

```js
let handler;

export default async callback => {
	if (handler) { // 1
		return;
	}

	handler = handler || ((args) => { /* code */ }); // 2

	Linking.removeEventListener('url', handler); // 3
	Linking.addEventListener('url', handler); // 4
       //...
};
``` 

**First invoke:**
- `1` mark, will be `false`, because `handler` no initialized yet
- `2` mark, will assign `handler` variable 
- `3` tried to unsubscribe, but it no make canes (as it was not subscribed before) 
- `4` will be successfully subscribed!

**Second invoke:**
- `1` mark, now if block is **TRUE** and immediately exit from function (as handler was initialized on first call on `2` mark)
- add now never mind what version of react-native 0.64 or 0.66! It subscribe once AND never reach unsubscribe line!

retyui:
> Thanks for submitting the PR and really appreciate the unit tests star_struck Left some feedback on the changes.
> 
> Regarding the deprecated warning for when using React Native , have you considered leveraging ? e.g.`Linking.removeListener``v0.66.x``YellowBox.ignoreWarnings`
> 
> ```
> import { YellowBox } from 'react-native';
> YellowBox.ignoreWarnings(['Warning: ...']);
> ```

It is nonsense! when you create fresh project and install latest versions of amplify to see Deprecation warnings!

retyui:
p.s. seems current build process has a memory leak

image

retyui:
@calebpollman did you have time to review this PR ?

retyui:
Close as I patch this module localy (https://github.com/ds300/patch-package)

PR #9454, opened 2022-01-09
feat(api-rest): Add ability to configure RestAPI to auto retry

ffxsam:
#### Description of changes

This PR adds retry ability to the RestAPI class. IMO, retry is crucial, as users' network connections are not always as stable as we'd like them to be, and sometimes API Gateway will timeout or return 5xx errors that are retryable.

This PR is a draft. My time is severely limited, so any help with writing tests would be appreciated!

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/api-rest/package.json
+++ b/packages/api-rest/package.json
@@ -42,7 +42,8 @@
   "homepage": "https://aws-amplify.github.io/",
   "dependencies": {
     "@aws-amplify/core": "4.3.12",
-    "axios": "0.21.4"
+    "axios": "0.21.4",
+    "axios-retry": "3.2.4"
   },
   "jest": {
     "globals": {
--- a/packages/api-rest/src/RestClient.ts
+++ b/packages/api-rest/src/RestClient.ts
@@ -21,6 +21,7 @@ import {
 
 import { apiOptions, ApiInfo } from './types';
 import axios, { CancelTokenSource } from 'axios';
+import axiosRetry from 'axios-retry';
 import { parse, format } from 'url';
 
 const logger = new Logger('RestClient');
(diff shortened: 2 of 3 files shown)


sammartinez:
Thanks for this @ffxsam ! We are going to take a look at this and see if we can get someone from our side starting to review and test.

ffxsam:
@sammartinez Thanks!

sammartinez:
Im going to have some folks this week spike on this @ffxsam. We do have some logic for retries within the library already that we are going to see about leveraging. We will keep you posted but it is seemingly like we will get back to you next week. Was curious, are you tied to wanting to use `axios-retry`? Or really the retry logic? Let us know as it'll help our spike.

ffxsam:
IMO, axios-retry is preferrable. And then we can document the options and point out that `retryDelay` is a standard option for axios-retry and can be a function, so people can write their own retry mechanisms.

sammartinez:
> IMO, axios-retry is preferrable. And then we can document the options and point out that `retryDelay` is a standard option for axios-retry and can be a function, so people can write their own retry mechanisms.

Thanks for this! We will consider this as we are thinking through this. Appreciate it, we will be in touch!

abdallahshaban557:
Hi @ffxsam - hope you are doing great - it's been a while since we've chatted! We are closing this PR since after some internal discussions we have agreed on the following points:

1. We do not want to bring in a third party dependency for logic that currently lives within our [core package](https://github.com/aws-amplify/amplify-js/blob/main/packages/core/src/Util/Retry.ts#L17-L56) for retry.
2. The dependency would increase bundle size by 1.7 kB gzipped/minified, and we are striving to reduce our bundle size as much as possible based on customer feedback. 

Would you be open to submitting another PR that uses our existing retry logic?

ffxsam:
@abdallahshaban557 Hey Abdallah! Any solution is good as long as it works. 🙂 And of course, it's documented somewhere (I didn't notice anything in the official Amplify docs, but I might've missed it).

PR #9509, opened 2022-01-24
fix(@aws-amplify/geo): bug fixes

TreTuna:
#### Description of changes

These are some bug fixes that I found on an old branch that have not made it to `main` yet:
- Remove `ensure to` from error messages - This is from DA feedback
- Remove unused `AmazonLocationServiceProvider.ts` file (this was replaced by `AmazonLocationServicesProvider.ts`)
- Fix bad type comment
- Fix implementation for failing when both `biasPosition` and `searchAreaConstraints` are given in search options

Updated tests and ran through sample application

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/geo/__tests__/Geo.test.ts
+++ b/packages/geo/__tests__/Geo.test.ts
@@ -134,7 +134,7 @@ describe('Geo', () => {
 			geo.configure({});
 
 			expect(() => geo.getAvailableMaps()).toThrow(
-				"No map resources found in amplify config, run 'amplify add geo' to create them and ensure to run `amplify push` after"
+				"No map resources found in amplify config, run 'amplify add geo' to create them and run `amplify push` after"
 			);
 		});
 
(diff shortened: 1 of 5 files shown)


packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -268,13 +274,13 @@ export class AmazonLocationServiceProvider implements GeoProvider {
 	private _verifyMapResources() {
 		if (!this._config.maps) {
 			const errorString =
-				"No map resources found in amplify config, run 'amplify add geo' to create them and ensure to run `amplify push` after";
+				"No map resources found in amplify config, run 'amplify add geo' to create them and run `amplify push` after";
```
thaddmt, line 277: I know this is not the main point of this pr but at this point it seems like we get this error more for having multiple different @aws-amplify/core packages, do we have documentation on that we can link to?

thaddmt (approved):

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: Hmm... No, we don't. You are right in that we get this more for that reason than any others.

joebuono (approved):

packages/geo/__tests__/Geo.test.ts
```diff
@@ -194,6 +194,8 @@ describe('Geo', () => {
 	});
 
 	describe('searchByText', () => {
+		const testString = 'starbucks';
```
hvergara, line 197: nit: can we use this value? it's a registered trademark

Issue #9534, opened 2022-01-31
DataStore not syncing when upgrade from aws-amplify 4.2.6 to 4.3.13


sacrampton:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

DataStore

### Amplify Categories

api

### Environment information

<details>

```
# Put output below this line
System:
    OS: macOS 11.5.2
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 214.66 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 16.13.2 - /usr/local/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 8.1.2 - /usr/local/bin/npm
    Watchman: 2022.01.10.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 97.0.4692.99
    Safari: 14.1.2
  npmPackages:
    @apollo/client: ^3.5.8 => 3.5.8
    @apollo/client/cache:  undefined ()
    @apollo/client/core:  undefined ()
    @apollo/client/errors:  undefined ()
    @apollo/client/link/batch:  undefined ()
    @apollo/client/link/batch-http:  undefined ()
    @apollo/client/link/context:  undefined ()
    @apollo/client/link/core:  undefined ()
    @apollo/client/link/error:  undefined ()
    @apollo/client/link/http:  undefined ()
    @apollo/client/link/persisted-queries:  undefined ()
    @apollo/client/link/retry:  undefined ()
    @apollo/client/link/schema:  undefined ()
    @apollo/client/link/utils:  undefined ()
    @apollo/client/link/ws:  undefined ()
    @apollo/client/react:  undefined ()
    @apollo/client/react/components:  undefined ()
    @apollo/client/react/context:  undefined ()
    @apollo/client/react/hoc:  undefined ()
    @apollo/client/react/hooks:  undefined ()
[...]

PR #9542, opened 2022-02-01
fix(datastore): correctly apply config values

iartemiev:
#### Description of changes

Some recent changes we made to DataStore introduced a regression causing certain config values (e.g. `syncPageSize` and `maxRecordsToSync`) to not be utilized correctly in the Sync Engine.

This PR restores the expected behavior.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/9534

Sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/src/datastore/datastore.ts
+++ b/packages/datastore/src/datastore/datastore.ts
@@ -744,8 +744,6 @@ class DataStore {
 				userClasses,
 				this.storage,
 				modelInstanceCreator,
-				this.maxRecordsToSync,
-				this.syncPageSize,
 				this.conflictHandler,
 				this.errorHandler,
 				this.syncPredicates,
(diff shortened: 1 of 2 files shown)


packages/datastore/src/datastore/datastore.ts
```diff
@@ -1343,21 +1342,22 @@ class DataStore {
...
-			this.fullSyncInterval ||
 			configFullSyncInterval ||
+			this.fullSyncInterval ||
 			24 * 60; // 1 day
 
 		this.storageAdapter =
 			(configDataStore && configDataStore.storageAdapter) ||
-			this.storageAdapter ||
 			configStorageAdapter ||
+			this.storageAdapter ||
```
nickarocho, line 1360: Does the swapped order make a difference here (same question with the `configSyncExpressions || this.syncExpressions` on line 1330/1331)? Asking out of curiosity.

nickarocho (approved):
LGTM! 👍 Would be interested to know how removing these config values from the signature solved the problem.

iartemiev:
@nickarocho what fixed it was changing the order of the config value assignments. 

I cleaned up the fn signature while I was in there, removing the unused params, but that was just for the sake of maintenance.

svidgen (approved):

PR #9587, opened 2022-02-14
fix(storage): axios handler error handling fix

### LINKED ISSUE #9252: Storage API doesn't throw error even if the CORS error
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Not applicable

### Amplify APIs

Storage

### Amplify Categories

storage

### Environment information

Chrome

### Describe the bug

Hi

We found a point for improvement in Storage API.
Storage API can't catch error even if CORS error occurs, the Promise is resolved.
I think it is better Storage API throw error.

#### StorageAPI-sample-code
```
async function put_s3() {
  // Specify a bucket that does not exist
  await Storage.put("test.txt", "Hello", { bucket: "XXXXX" }).then(
    (result) => {
      console.log("result!!!!");
      console.log(result)
    }
  ).catch(
    (error) => {
      console.log("error!!!!");
      console.log(error)
    }
  )
}
```

#### Browsers console log(Chrome)
```
Access to XMLHttpRequest at 'https://XXXXX.s3.ap-northeast-1.amazonaws.com/public/test.txt?x-id=PutObject' from origin 'http://localhost:3000' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
index.js:1 [ERROR] 39:07.605 axios-http-handler - Network Error
...
App.js:48 result!!!!
App.js:49 {key: 'test.txt'}
```

I think this behavior is a little strange.
I checked the behavior of [axios](https://github.com/axios/axios), they throwed error.

#### axios-sample-code
```
  axios.post('/items')
    .then((response) => {
      console.log(response);
    })
    .catch((error) => {
      console.log("error!!!");
      console.log(error);
    });
```

Browsers console log
```
localhost/:1 Access to XMLHttpRequest at 'https://XXXXX/items' from origin 'http://localhost:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
App.js:40 error!!!
App.js:41 Error: Network Error
    at createError (createError.js:16)
    at XMLHttpRequest.handleError (xhr.js:99)
xhr.js:187 POST https://XXXXX/items net::ERR_FAILED 200
```

I checked Similar this issues
https://github.com/aws-amplify/amplify-js/issues/3419

But this issues seems to be fixed.

Thanks

jamesaucode:
#### Description of changes
Added a default error response status to the axios handler.
Before, any error response that wasn't parse-able are considered a success response (if status code is `undefined`), which makes responses resolve even tho they are supposed to be errors (e.g network interruption, CORS error, basically any unexpected error while making the request)

#### Issue #, if available
fix #9252

`yarn test`
- Configured a bucket with CORS disabled
- Made request to the bucket, checked that the `Storage.put` promise is properly rejected

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


gvankeerberghen:
Hello @jamesaucode,

Guillaume from https://github.com/aws-amplify/amplify-js/issues/9252#issuecomment-1033680473 here.

I could reproduce the network error issue in my mobile app by turning my network off just before the call to `Storage.put`. Indeed the method was logging
```
[ERROR] 28:13.460 axios-http-handler - Network Error
```
but still resolving.

Then I patched the issue by changing the catch in the file you touch here. I could get `Storage.put` to throw when the file couldn't but pushed. But I didn't do it the way you do here.

Indeed, the code with the current patch here still looks too optimistic to me.

Let me explain my thinking. The catch, as it is, would silence any error axios could have good reasons to throw. I can picture:
- wrong way to call Axios' method,
- network error (no network -> no request, or dropped network -> couldn't send request),
- cors issue,
but, more importantly, I'm sure there can be a large number of error scenarii I currently don't know of. I could try and list them from axios' doc but, even if I did, other scenarii might be added in the future when axios would be upgraded.

Currently, in all these cases we know and don't know of, the code would only log and return an Http response with status code `400 Bad Request`.

So what I felt more confident with was being defensive in this catch:
- log the error in all cases but the one you know shouldn't be logged (as it is now)
- if the error contains a response then pass that HttpResponse on, which assumes other middlewares will correctly handle a responses with fail statuses. I didn't check that and assumed it was the case based on your code.
- in all other cases we do not know of, throw the error up so it can be discovered and addressed.

If there are other error cases that could be addressed here then they could be added explicitly before the default cases which should be to throw the error we don't know anything about I think.

Hope this is good food for thoughts,

Cheers

jamesaucode:
> Hello @jamesaucode,
> 
> Guillaume from [#9252 (comment)](https://github.com/aws-amplify/amplify-js/issues/9252#issuecomment-1033680473) here.
> 
> I could reproduce the network error issue in my mobile app by turning my network off just before the call to `Storage.put`. Indeed the method was logging
> 
> ```
> [ERROR] 28:13.460 axios-http-handler - Network Error
> ```
> 
> but still resolving.
> 
> Then I patched the issue by changing the catch in the file you touch here. I could get `Storage.put` to throw when the file couldn't but pushed. But I didn't do it the way you do here.
> 
> Indeed, the code with the current patch here still looks too optimistic to me.
> 
> Let me explain my thinking. The catch, as it is, would silence any error axios could have good reasons to throw. I can picture:
> 
>     * wrong way to call Axios' method,
> 
>     * network error (no network -> no request, or dropped network -> couldn't send request),
> 
>     * cors issue,
> 
> 
> but, more importantly, I'm sure there can be a large number of error scenarii I currently don't know of. I could try and list them from axios' doc but, even if I did, other scenarii might be added in the future when axios would be upgraded.
> 
> Currently, in all these cases we know and don't know of, the code would only log and return an Http response with status code `400 Bad Request`.
> 
> So what I felt more confident with was being defensive in this catch:
> 
>     * log the error in all cases but the one you know shouldn't be logged (as it is now)
> 
>     * if the error contains a response then pass that HttpResponse on, which assumes other middlewares will correctly handle a responses with fail statuses. I didn't check that and assumed it was the case based on your code.
> 
>     * in all other cases we do not know of, throw the error up so it can be discovered and addressed.
> 
> 
> If there are other error cases that could be addressed here then they could be added explicitly before the default cases which should be to throw the error we don't know anything about I think.
> 
> Hope this is good food for thoughts,
> 
> Cheers

Really good feedback here. I agree with your points here. Let me try to summarize to see if Im understanding it correctly.

From my perspective there should be some errors that are:
1. Expected (e.g an axios.cancel call from a Resumable upload object, we expect an error to be thrown there)
2. Unexpected and can be re-constructed to an `HttpResponse` and passed down to other middleware.
3. All the other unexpected errors, e.g flaky network connection.

The way we should handle this is to 
1. Check if it's an expected error
  a. yes, don't log it
  b. no, log it
2. Unexpected error, but the error response is available so it can be handle by middlewares down the middleware chain, create an `HttpResponse` out of it and let the SDK do it's thing
3. Unexpected error but response cannot be parsed, re-throw the error.

In retrospect the default error response wasn't a good idea. Originally my thought process was that we can still pass it down so the middlewares (e.g retry middleware) can handle it and might recover from it, but if the response object is unavailable its very likely that the response is unrecoverable, + we lose the actual error object from the process. 
Good callout here, thanks ! @gvankeerberghen

packages/storage/__tests__/providers/axios-http-handler.test.ts
```diff
@@ -162,8 +167,11 @@ describe('AxiosHttpHandler', () => {
 				.mockImplementation(() => Promise.reject(new Error('err')));
 			const loggerSpy = jest.spyOn(Logger.prototype, '_log');
 			const handler = new AxiosHttpHandler();
-			await handler.handle(request, options);
-			expect(loggerSpy).toHaveBeenCalledWith('ERROR', 'err');
+			try {
+				await handler.handle(request, options);
+			} catch (_error) {
+				expect(loggerSpy).toHaveBeenCalledWith('ERROR', 'err');
```
nickarocho, line 173: Would `_error.msg` just resolve to `'err'` in this case? Might be better to write a more explicit error message to avoid confusion with different error instances.

katiegoines (approved):
lgtm

packages/storage/src/providers/axios-http-handler.ts
```diff
@@ -186,15 +200,19 @@ export class AxiosHttpHandler implements HttpHandler {
...
+					// consider the request successful by mistake. In this case we should also re-throw the error.
+					if (axios.isCancel(error) || !hasErrorResponse(error)) {
 						throw error;
 					}
 					// otherwise, we should re-construct an HttpResponse from the error, so that it can be passed down to other
 					// aws sdk middleware (e.g retry, clock skew correction, error message serializing)
 					return {
 						response: new HttpResponse({
-							statusCode: error.response?.status,
+							statusCode: error.response.status,
```
nickarocho, line 215: Why remove the optional chaining here? Can we guarantee that `error.response` will be a present property based on the type declaration?

jamesaucode: we called `hasErrorResponse` to do a runtime check above to ensure `error.response.status` exist so the optional chaining is not needed anymore.

packages/storage/__tests__/providers/axios-http-handler.test.ts
jamesaucode: Yes, I have it in `try` `catch` so the thrown error doesn't fail the test. The `'err'` is simply there for me to assert that `logger` logged the error with the correct error message (the error message thrown from the `mockImplementation`)

packages/storage/src/providers/axios-http-handler.ts
nickarocho: Makes sense, just wanted to check first. Would `error.response` not also be guaranteed now too?

jamesaucode: `error.response` , and `error.response.status` are guaranteed yes

nickarocho: Got it. Just wondering if the optional chaining should be removed there for consistency sake? Not a blocker either way just want to make sure I understand the intention behind `hasErrorResponse`.

nickarocho (approved):
LGTM! Thanks James 🚢

Issue #9556, opened 2022-02-04
DataStore observeQuery function not behaving as expected on update


alharris-at:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

DataStore

### Amplify Categories

Not applicable

### Environment information

<details>

```
  System:
    OS: macOS 11.6.3
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 4.42 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.7.0 - ~/.nvm/versions/node/v16.7.0/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 7.20.3 - ~/.nvm/versions/node/v16.7.0/bin/npm
    Watchman: 2022.01.31.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 98.0.4758.80
    Firefox: 91.5.1
    Safari: 15.3
  npmPackages:
    @testing-library/jest-dom: ^5.16.2 => 5.16.2 
    @testing-library/react: ^12.1.2 => 12.1.2 
    @testing-library/user-event: ^13.5.0 => 13.5.0 
    aws-amplify: ^4.3.14 => 4.3.14 
    react: ^17.0.2 => 17.0.2 
    react-dom: ^17.0.2 => 17.0.2 
    react-scripts: 5.0.0 => 5.0.0 
    web-vitals: ^2.1.4 => 2.1.4 
  npmGlobalPackages:
    @aws-amplify/cli: 7.6.9
    npm: 7.20.3
    typescript: 4.5.2
```

</details>

### Describe the bug

When using observeQuery, it appears on local element update, only partial fields from the model are being returned, and the remaining fields are just not propagated through the snapshot callback.

### Expected behavior

I'd expect the full model to be returned, since we return all matching elements from the query as part of the response. I have confirmed that in IndexedDB the full model is updated and appears to be complete.

[...]

PR #9617, opened 2022-02-19
feat: update DataStore observe / observeQuery to return all fields in local update snapshot (#9556)

david-mcafee:
feat: update DataStore observe to return all fields in local update snapshot (#9556)

#### Issue #, if available

#9556

Manual testing with a sample app, running unit tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [N/A] Relevant documentation is changed or added (and PR referenced)


packages/datastore/src/types.ts
```diff
@@ -397,6 +397,7 @@ export type SubscriptionMessage<T extends PersistentModel> = {
 	element: T;
 	model: PersistentModelConstructor<T>;
 	condition: PredicatesGroup<T> | null;
+	originalElement?: T;
```
david-mcafee, line 400: This is the only update to this file - the rest is Prettier.

packages/datastore/src/datastore/datastore.ts
```diff
@@ -1123,9 +1123,27 @@ class DataStore {
...
+					.map((event: SubscriptionMessage<T>): SubscriptionMessage<T> => {
+						// Storage's `save` only returns updated fields - intercept the
+						// event to combine the updated fields with the original, so
+						// that first snapshot returned to the consumer contains all fields
+						const { opType, element, model, condition, originalElement } =
+							event;
+
+						const updated = {
+							opType,
+							element: { ...originalElement, ...element },
```
cshfang, line 1140: Is it ever possible for an update to contain only changes to some nested fields?

iartemiev: I believe it's safe to just set `element: originalElement` here without combining the two. `originalElement` should be a strict superset of `element`.

iartemiev (changes requested):
Looks good! Just a minor req'd change

svidgen (commented):
Can we add testing for this?

packages/datastore/src/datastore/datastore.ts
david-mcafee: My bad, I completely forgot that `originalElement` is indeed the updated element. Will rename.

packages/datastore/src/datastore/datastore.ts
```diff
@@ -1123,9 +1123,27 @@ class DataStore {
...
 					.observe(modelConstructor, predicate)
 					.filter(({ model }) => namespaceResolver(model) === USER)
+					.map((event: SubscriptionMessage<T>): SubscriptionMessage<T> => {
+						// The `element` returned by storage only contains updated fields.
+						// Intercept the event to send the `savedElement` so that the first 
+						// snapshot returned to the consumer contains all fields
+						const { opType, model, condition, savedElement } =
+							event;
+
+						const updated = {
```
cshfang, line 1138: Super nit as this is obviously a micro-optimization but I think you could just return this directly here to save a tick without really losing code clarity.

packages/datastore/src/storage/storage.ts
```diff
@@ -144,6 +144,7 @@ class StorageClass implements StorageFacade {
 				element,
 				mutator,
 				condition: ModelPredicateCreator.getPredicates(condition, false),
+				savedElement,
```
cshfang, line 147: Just making sure we do a quick sanity check: Does this property being added to the stream affect any consumers of the stream that might not expect it?

cshfang (approved):
Small nit and a small question that I think was answered offline. Just wanted to make sure we double check it real quick but otherwise lgtm!

david-mcafee:
> Can we add testing for this?

@svidgen - sure!

manueliglesias (approved):
LGTM 👍

david-mcafee:
> > LGTM 👍
> 
> Edit: I need to re-review this as it might be having unintended effects in the fields sent to the mutation input

@manueliglesias this shouldn't affect that, as the updates are only being returned to the customer, but the mutation event sent to AppSync will only send updated fields

manueliglesias:
> > > LGTM 👍
> > 
> > 
> > Edit: I need to re-review this as it might be having unintended effects in the fields sent to the mutation input
> 
> @manueliglesias this shouldn't affect that, as the updates are only being returned to the customer, but the mutation event sent to AppSync will only send updated fields

Awesome! Thank you for confirming @david-mcafee  😄

packages/datastore/src/types.ts
```diff
@@ -397,6 +397,7 @@ export type SubscriptionMessage<T extends PersistentModel> = {
 	element: T;
 	model: PersistentModelConstructor<T>;
 	condition: PredicatesGroup<T> | null;
+	savedElement?: T;
```
manueliglesias, line 400: I think we should split the types for internal notifications and customer facing notifications.

e.g.

```typescript
export type SubscriptionMessage<T extends PersistentModel> = Omit<InternalSubscriptionMessage<T>, 'savedElement'>;

export type InternalSubscriptionMessage<T extends PersistentModel> = {
	opType: OpType;
	element: T;
	model: PersistentModelConstructor<T>;
	condition: PredicatesGroup<T> | null;
	savedElement: T;
};
```

manueliglesias (changes requested):
It works great! The only thing I'd like to see changed is creating two types instead of adding an optional field to the existing type. That way we don't leak that implementation detail.

packages/datastore/src/datastore/datastore.ts
david-mcafee: Agree!

packages/datastore/src/storage/storage.ts
david-mcafee: Definitely worth mentioning! However, it does not - I've tested this change pretty extensively, and stepped through both the storage and sync engine, and have observed nothing problematic about adding this here. Also, in the next revision I'll be splitting the types between internal and customer facing notifications (see Manuel's comment below), which should provide some type safety around the notifications.

packages/datastore/src/types.ts
david-mcafee: Discussed offline, will be using `Pick` instead of `Omit`

packages/datastore/__tests__/DataStore.ts
```diff
@@ -13,38 +13,391 @@ import {
...
+		// `observeQuery()` depends on.
+		(DataStore as any).sync = {
+			// default to report that models are NOT synced.
+			// set to `true` to signal the model is synced.
+			// `observeQuery()` should finish up after this returns `true`.
+			getModelSyncedStatus: (model: any) => false,
+
+			// not important for this testing. but unsubscribe calls this.
+			// so, it needs to exist.
+			unsubscribeConnectivity: () => {},
```
cshfang, line 275: Nit: Doesn't seem like this is the case but if tests ever did need to update the mocks for these implementations, it seems more idiomatic to use `jest.fn()`.

cshfang (approved):
This lgtm. Thanks for working on this

manueliglesias (approved):
LGTM!! 👍

iartemiev (approved):
LGTM 👍🏻

david-mcafee:
Confirmed with @svidgen offline that it's alright to merge this.

PR #9628, opened 2022-02-21
fix: Invalid session cleanup

### LINKED ISSUE #4438: What is aws-amplify's Default Behavior When a Refresh Token Expires?
**Which Category is your question related to?**
Auth

**What AWS Services are you utilizing?**
AWS Cognito

**Provide additional details e.g. code snippets**
My stack is a React application using aws-amplify to authenticate with AWS Cognito identity pool.

I am wondering what happens when a user authenticates into an app that is using AWS Amplify, and the refresh token validity expires for that user? Will aws-amplify automatically send the user to AWS Cognito for re-authentication? If not, is there a Hub event to catch when a user's refresh token expires?

### LINKED ISSUE #9638: Handle expired session
### Is this related to a new or existing framework?

_No response_

### Is this related to a new or existing API?

Authentication

### Is this related to another service?

_No response_

### Describe the feature you'd like to request

Case:
When a user's session gets expired we want to show a popup modal that describes what action the user can take to continue their work. We want to show this popup only when a user tries to do some action that triggers a request to a server. To solve this problem, we want to catch the "No current user" error and handle it gracefully.

### Describe the solution you'd like

I tried to use axios interceptors to catch expired session error but Amplify Auth doesn't even trigger a request. 
This is the error I'm trying to catch: https://github.com/aws-amplify/amplify-js/blob/bf3d189c9e7e14342230fca8a0cbba6972e7b42c/packages/auth/src/Auth.ts#L1345

It's triggered internally and the only way to catch it is to handle every request individually. Would be nice to have one place where I can catch this error.

### Describe alternatives you've considered

For now I wrapped currentUserPoolUser in another function which catches any error and dispatches it thorugh Hub module.
```js
const superCurrentUserPoolUser = Auth.currentUserPoolUser;
  Auth.currentUserPoolUser = async function(...args) {
    try {
      const result = await superCurrentUserPoolUser.apply(this, args);
      return result;
    } catch (error) {
      Hub.dispatch('app', {
        event: 'currentUserPoolUserError',
        data: { error }
      });
      throw error;
    }
  };
```

### Additional context

_No response_

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

jamesaucode:
#### Description of changes
Auth will now check for invalid session error (Access token revoked, refresh token revoked etc) while calling `.getUserData` or `.getSession`.
When such errors are detected, the invalid tokens will now be automatically cleared from the user storage. If user signed in using OAuth, it will redirect the user to the OAuth signout URL. Finally it will dispatch a `signOut` `Hub` event.

#### Issue #, if available
fix #9638
fix #4438

yarn test
integration tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/src/Auth.ts
```diff
@@ -1373,16 +1373,27 @@ export class AuthClass {
 									(err, data) => {
 										if (err) {
 											logger.debug('getting user data failed', err);
-											// Make sure the user is still valid
-											if (
-												err.message === 'User is disabled.' ||
-												err.message === 'User does not exist.' ||
-												err.message === 'Access Token has been revoked' // Session revoked by another app
-											) {
```
elorzafe, line 1381: I think this is important to validate before assuming the user is signed out. For example if there is a network issue, the user can then refresh the session once is online. I think it would also be good to check that on the tests.

jamesaucode: I think that makes sense, I've added the checks back.

elorzafe (commented):
I have one more comment, can you check if there any other options to force refreshing credentials that we are not capturing?

elorzafe (commented):
Overall looks great!

Thanks @jamesaucode 🎉 🌮 

I have one final question, what happen on `user.getSession` fail to refresh?

packages/auth/src/Auth.ts
```diff
@@ -700,9 +700,16 @@ export class AuthClass {
...
 			const bypassCache = params ? params.bypassCache : false;
 			user.getUserData(
-				(err, data) => {
+				async (err, data) => {
 					if (err) {
 						logger.debug('getting preferred mfa failed', err);
+						if (this.isSessionInvalid(err)) {
+							try {
+								await this.cleanUpInvalidSession(user);
+							} catch (err) {
```
elorzafe, line 709: Nit pick: changing the name of `err` inside `catch` to be different that outside `err` to make it easy to read

packages/auth/src/Auth.ts
```diff
@@ -1292,6 +1319,80 @@ export class AuthClass {
...
+		);
+	}
+
+	private async cleanUpInvalidSession(user: CognitoUser) {
+		user.signOut();
+		this.user = null;
+		try {
+			await this.cleanCachedItems(); // clean aws credentials
+		} catch (e) {
+			// TODO: change to rejects in refactor
```
elorzafe, line 1384: I would not reject there in favor of `dispatching` the event and `oAuthSignOutRedirect` maybe you could assign that later and throw at the end? 

On the other hand maybe is more important to throw the error that session is invalid than saying there was an issue cleaning the session. How could the customer recover from the session clean up failed?, Probably is easier to recover for session was invalidated. And deal with `Hub` event to redirect to `sign in` screen.

elorzafe (approved):
LGTM, just nitpicking comments
Thanks @jamesaucode 🌮 🥇 🎉

ashika01 (approved):
LGTM

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -252,6 +260,21 @@ jest.mock('amazon-cognito-identity-js/lib/CognitoUser', () => {
...
+	setItem: jest.fn().mockImplementation((key, value) => {
+		mockLocalStorageItems[key] = value;
+	}),
+	clear: jest.fn().mockImplementation(() => {
+		mockLocalStorageItems = {};
+	}),
+	removeItem: jest.fn().mockImplementation(key => {
+		mockLocalStorageItems[key] = undefined;
+	}),
+} as unknown as Storage;
```
nickarocho: Could this be useful in other tests down the road? If so, might be worth considering abstracting it as a `util` object which we could import to other unit tests.

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -2756,11 +2824,20 @@ describe('auth unit test', () => {
...
+			try {
+				expect(await auth.deleteUser()).toBe('SUCCESS');
+				expect(userSignoutSpy).toHaveBeenCalledTimes(1);
+			} catch (err) {
+				console.error(err);
+			}
 			// TODO: test session cleared properly
 			spy1.mockClear();
 			spy2.mockClear();
+			userSignoutSpy.mockClear();
```
nickarocho, line 2866: Jest API Clarifying Q: does the `jest.clearAllMocks()` on line 2814 not clear these mocks already?

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -3980,4 +4122,170 @@ describe('auth unit test', () => {
...
+						PreferredMfaSetting: 'SMS',
+					};
+					callback(null, data);
+				});
+			const setUserMfaPreferenceSpy = jest.spyOn(user, 'setUserMfaPreference');
+			const res = await auth.setPreferredMFA(user, 'SOFTWARE_TOKEN_MFA');
+			expect(setUserMfaPreferenceSpy).toHaveBeenCalledWith(
+				null,
+				{ Enabled: true, PreferredMfa: true },
+				expect.any(Function)
```
nickarocho, line 4302: What does this assertion test? Any function is passed as a callback?

packages/auth/src/Auth.ts
```diff
@@ -700,9 +700,12 @@ export class AuthClass {
 
 			const bypassCache = params ? params.bypassCache : false;
 			user.getUserData(
-				(err, data) => {
+				async (err, data) => {
 					if (err) {
 						logger.debug('getting preferred mfa failed', err);
+						if (this.isSessionInvalid(err)) {
+							await this.cleanUpInvalidSession(user);
+						}
```
nickarocho, line 716: I know this is within a returned promise, but would wrapping this async code with a `try/catch` block potentially catch any unhandled promise rejections?

packages/auth/src/Auth.ts
```diff
@@ -1141,7 +1149,7 @@ export class AuthClass {
 	 * Delete the current authenticated user
 	 * @return {Promise}
 	 **/
-	//TODO: Check return type void
+	// TODO: Check return type void
```
nickarocho, line 1176: Is this something we could add easily to this PR?

packages/auth/src/Auth.ts
```diff
@@ -1533,9 +1615,12 @@ export class AuthClass {
...
 			logger.debug('Getting the session from this user:', user);
 			user.getSession(
-				(err, session) => {
+				async (err, session) => {
 					if (err) {
 						logger.debug('Failed to get the session from user', user);
+						if (this.isSessionInvalid(err)) {
+							await this.cleanUpInvalidSession(user);
+						}
 						reject(err);
```
nickarocho, line 1679: Nit: (and I know it's not your code), but the other Promises use `res` and `rej`. Could we change the naming here for consistency-sake?

nickarocho (approved):
LGTM! Left some questions for my own learning and very nit comments. Thanks for your robust and detailed unit tests, James! 🚢 👍

packages/auth/__tests__/auth-unit-test.ts
jamesaucode: it does, removed the redundant calls 👍

packages/auth/__tests__/auth-unit-test.ts
jamesaucode: Yea, I am mainly interested in if the call was made for `user.setUserMfaPreference` and the second param of the call, not necessarily what function was used for the callback

packages/auth/src/Auth.ts
jamesaucode: it will catch any error/reject promises from `this.cleanUpInvalidSession(user)`. In this case its only going to be if oAuthRedirect fails

packages/auth/src/Auth.ts
jamesaucode: I will keep this PR focused on the invalid session cleanup for now.

packages/auth/src/Auth.ts
jamesaucode: 👍  done, thanks for pointing that out.

packages/auth/src/Auth.ts
```diff
@@ -1353,6 +1479,17 @@ export class AuthClass {
...
 								logger.debug('Failed to get the user session', err);
+								if (this.isSessionInvalid(err)) {
+									try {
+										await this.cleanUpInvalidSession(user);
+									} catch (cleanUpError) {
+										rej(
+											new Error(
+												`Session is invalid due to: ${err.message} and failed to clean up invalid session: ${cleanUpError.message}`
+											)
+										);
```
elorzafe, line 1490:
```suggestion
										);
										return;
```

packages/auth/src/Auth.ts
```diff
@@ -1165,6 +1197,17 @@ export class AuthClass {
...
 							logger.debug('Failed to get the user session', err);
+							if (this.isSessionInvalid(err)) {
+								try {
+									await this.cleanUpInvalidSession(user);
+								} catch (cleanUpError) {
+									rej(
+										new Error(
+											`Session is invalid due to: ${err.message} and failed to clean up invalid session: ${cleanUpError.message}`
+										)
+									);
```
elorzafe, line 1208:
```suggestion
									);
									return;
```

packages/auth/src/Auth.ts
```diff
@@ -856,9 +877,20 @@ export class AuthClass {
...
 								logger.debug('getting user data failed', err);
+								if (this.isSessionInvalid(err)) {
+									try {
+										await this.cleanUpInvalidSession(user);
+									} catch (cleanUpError) {
+										rej(
+											new Error(
+												`Session is invalid due to: ${err.message} and failed to clean up invalid session: ${cleanUpError.message}`
+											)
+										);
```
elorzafe, line 891:
```suggestion
										);
										return;
```

packages/auth/src/Auth.ts
```diff
@@ -753,14 +764,24 @@ export class AuthClass {
...
 					logger.debug('getting user data failed', err);
+					if (this.isSessionInvalid(err)) {
+						try {
+							await this.cleanUpInvalidSession(user);
+						} catch (cleanUpError) {
+							rej(
+								new Error(
+									`Session is invalid due to: ${err.message} and failed to clean up invalid session: ${cleanUpError.message}`
+								)
+							);
```
elorzafe, line 778:
```suggestion
							);
							return;
```

packages/auth/src/Auth.ts
```diff
@@ -700,9 +700,20 @@ export class AuthClass {
...
 						logger.debug('getting preferred mfa failed', err);
+						if (this.isSessionInvalid(err)) {
+							try {
+								await this.cleanUpInvalidSession(user);
+							} catch (cleanUpError) {
+								rej(
+									new Error(
+										`Session is invalid due to: ${err.message} and failed to clean up invalid session: ${cleanUpError.message}`
+									)
+								);
```
elorzafe, line 714:
```suggestion
								);
								return;
```

elorzafe (commented):
Added a few suggestions related to `return` after promise has been rejected, to avoid executing unnecessary code.

Issue #9528, opened 2022-01-28
Android native code crashing when FirebaseMessaging.getInstance().getToken() fails 


RMG0:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

Push Notifications

### Amplify Categories

notifications

### Environment information

<details>

```
# Put output below this line
  System:
    OS: Windows 10 10.0.19042
    CPU: (8) x64 Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz
    Memory: 3.43 GB / 15.71 GB
  Binaries:
    Node: 16.13.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.10 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.3.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 97.0.4692.99
    Edge: Spartan (44.19041.1266.0), Chromium (97.0.1072.76)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    @aws-amplify/pushnotification: ^4.3.8 => 4.3.8
    @babel/core: ^7.16.7 => 7.16.7
    @babel/runtime: ^7.16.7 => 7.16.7
    @react-native-community/eslint-config: ^3.0.1 => 3.0.1
    @react-native-community/netinfo: ^7.1.7 => 7.1.7
    @react-native-community/push-notification-ios: ^1.10.1 => 1.10.1 (1.0.3)
    amazon-cognito-identity-js: ^5.2.3 => 5.2.3
    aws-amplify: ^4.3.11 => 4.3.11
    axios: ^0.24.0 => 0.24.0 (0.21.4)
    babel-plugin-transform-remove-console: ^6.9.4 => 6.9.4
    eslint: 8.6.0 => 8.6.0
    eventing-bus: ^2.0.1 => 2.0.1
    hermes-inspector-msggen:  1.0.0
[...]

PR #9641, opened 2022-02-24
fix(pushnotification): return the exception message rather than the whole exception object

Samaritan1011001:
#### Description of changes

The change corrects the way exception is returned to the JS side of the push notification package when fetching the device token from Firebase's `fetchToken` API. The exception message is obtained through `exception.getMessage()` method call. 

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/9528

Produced the exception by disabling Google Play Services on the test device which caused an exception when getting the token through the function described above which caused the app to crash. With this I was able to reproduce the issue linked above. Validated the solution fixes the issue.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/pushnotification/android/src/main/java/com/amazonaws/amplify/pushnotification/RNPushNotificationModule.java
```diff
@@ -77,8 +77,11 @@ public void onComplete(@NonNull Task<String> task) {
                         onSuccessCallback.invoke(token);
                     } else {
                         Exception exception = task.getException();
-                        Log.e(LOG_TAG, "Error getting token: " + exception);
-                        onErrorCallback.invoke(exception);
+                        if (exception != null) {
```
calebpollman, line 80: Just curious, in what scenarios would `exception` be `null`? Do we need to have handling in place for that possibility?

calebpollman (commented):
Thanks for looking in to this @Samaritan1011001, just left a question

packages/pushnotification/android/src/main/java/com/amazonaws/amplify/pushnotification/RNPushNotificationModule.java
Samaritan1011001: The `getException()` function is annotated as `@Nullable` which means the returned value can be `null` so in order to avoid future errors, I added the check.  
And looks like it'll return `null` when the `Task` is not complete yet. https://developers.google.com/android/reference/com/google/android/gms/tasks/Task#getException()

packages/pushnotification/android/src/main/java/com/amazonaws/amplify/pushnotification/RNPushNotificationModule.java
```diff
@@ -77,8 +77,11 @@ public void onComplete(@NonNull Task<String> task) {
...
+                            String exceptionMessage = exception.getMessage();
+                            Log.e(LOG_TAG, "Error getting token: " + exceptionMessage);
+                            onErrorCallback.invoke(exceptionMessage);
```
cshfang, line 83: Just wanna double check this - previously the callback was invoked with an exception but now it is invoked with a string? Was it always supposed to be a string?

cshfang (commented):
Thanks, just have a qq

packages/pushnotification/android/src/main/java/com/amazonaws/amplify/pushnotification/RNPushNotificationModule.java
Samaritan1011001: Yes, it is supposed to be a serialized data and not a native object. Since the message string is enough in this case, I returned the string. 

From the RN [documentation](https://reactnative.dev/docs/native-modules-android#callbacks): 
`
You can invoke the callback in your Java method, providing whatever data you want to pass to JavaScript. Please note that you can only pass serializable data from native code to JavaScript. If you need to pass back a native object you can use WriteableMaps, if you need to use a collection use WritableArrays. 
`

cshfang (approved):

ivarvermeulen:
I'm also experiencing this issue, would be great if this PR gets merged

calebpollman (approved):
LGTM 🚢

packages/pushnotification/android/src/main/java/com/amazonaws/amplify/pushnotification/RNPushNotificationModule.java
poojamat: What happens when its not successful and exception is null?

Samaritan1011001: Great question! 
The previous if clause uses `task.isSuccessful()` which according to the documentation states it returns `true` if `completed successfully` or false **otherwise** which means when we check for the exception we would have the Task be **completed** unsuccessfully (completed being the key part) so `getException` would never in-fact return `null` in this situation

@calebpollman I think this is what you were hinting towards? 

We can either add a custom message for when the `exception` is null or leave it as is.

ivarvermeulen:
When will this fix be available as it seems no new version has been released of the pushnotification package?

Samaritan1011001:
@ivarvermeulen It's now available on the latest version -- `"@aws-amplify/pushnotification": "^4.3.14"`

PR #9674, opened 2022-03-07
fix : (urlListener.native.ts) use explicit imports, remove bugs and other improvements.

chintannp:
#### Description of changes

This PR makes some improvements to existing code and solves the bug in the code as reported in this closed [PR](https://github.com/aws-amplify/amplify-js/pull/9450).

currently working on more unit tests and validations.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/auth/src/urlListener.native.ts
```diff
@@ -12,43 +12,39 @@
...
-		subscription = Linking.addEventListener('url', handler);
-	}
-	AppState.addEventListener('change', async newAppState => {
-		if (newAppState === 'active') {
-			const initialUrl = await Linking.getInitialURL();
-			handler({ url: initialUrl });
+	linkingSubscription?.remove?.();
+	linkingSubscription = Linking.addEventListener('url', handler);
+
+	appStateEventSubscription?.remove?.();
```
cshfang, line 35: I'm not as sure about this one. But if this also only work with RN 0.65+, wouldn't it be a memory leak in lower RN versions?

packages/auth/src/urlListener.native.ts
```diff
@@ -12,43 +12,39 @@
...
+				handler({ url: initialUrl });
+			}
 		}
-	});
+	);
+
+	return () => {
+		linkingSubscription?.remove?.();
+		appStateEventSubscription?.remove?.();
+	};
```
cshfang, line 49: Why does this now return a function when it previously didn't? Wouldn't this require changes to consuming code?

packages/auth/src/urlListener.native.ts
```diff
@@ -12,43 +12,39 @@
...
-	} else {
-		// remove() method is only available on RN v0.65+.
-		subscription?.remove?.();
-		subscription = Linking.addEventListener('url', handler);
-	}
-	AppState.addEventListener('change', async newAppState => {
-		if (newAppState === 'active') {
-			const initialUrl = await Linking.getInitialURL();
-			handler({ url: initialUrl });
+	linkingSubscription?.remove?.();
```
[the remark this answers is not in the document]
chintannp: So, the `removeEventListener()` method was deprecated and now they suggest us to use the `remove` method from `EmitterSubscription` object which is returned by `addEventListener` method. However, the `addEventListener` method always used to return `EmitterSubscription` object even in previous versions of RN. So, this will still work for older react-native versions. In short, this new way to remove listener is backward compatible.

packages/auth/src/urlListener.native.ts
chintannp: Ditto same as above, 

So, the removeEventListener() method was deprecated and now they suggest us to use the remove method from EmitterSubscription object which is returned by addEventListener method. However, the addEventListener method always used to return EmitterSubscription object even in previous versions of RN. So, this will still work for older react-native versions. In short, this new way to remove listener is backward compatible.

packages/auth/src/urlListener.native.ts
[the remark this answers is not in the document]
chintannp: Yes I will remove it. In the future refactor when we return the unsubscribe function, we need to handle removing the listeners.

packages/auth/src/urlListener.native.ts
```diff
@@ -12,43 +12,30 @@
...
-		subscription?.remove?.();
-		subscription = Linking.addEventListener('url', handler);
-	}
-	AppState.addEventListener('change', async newAppState => {
-		if (newAppState === 'active') {
-			const initialUrl = await Linking.getInitialURL();
-			handler({ url: initialUrl });
+	const linkingSubscription = Linking.addEventListener('url', handler);
+
+	const appStateEventSubscription = AppState.addEventListener(
```
calebpollman, line 32: When is this subscription removed?

PR #9677, opened 2022-03-08
chore: PubSub - Use prettier to rewrap to 120 characters consistently

stocaaro:
#### Description of changes
Using prettier `printWidth: 120` to reflow existing pubsub category files programmatically. 

This package is substantially wrapped to 80 characters now, which passes tslint. I have vscode configured to use prettier to format files on save and it is defaulting to 80 characters given no configured `printWidth` value.  I would recommend we add this to the `.prettierrc.js` file, however since this could be disruptive to others I have not included that change in this PR.

All changes made by prettier automatically, I made no other edits.

- `yarn test`
- Running an AppSync integrated App and IoT integrated app to manually verify no behavior changes.
- `yarn lint` succeeds both before and after this change

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/pubsub/src/Providers/AWSAppSyncProvider.ts
+++ b/packages/pubsub/src/Providers/AWSAppSyncProvider.ts
@@ -96,9 +96,9 @@ export class AWSAppSyncProvider extends MqttOverWSProvider {
 				const { mqttConnections = [], newSubscriptions } = options;
 
 				// creates a map of {"topic": "alias"}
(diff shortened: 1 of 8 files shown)


stocaaro:
Scrubbing this. Re-reading the [printWidth documentation](https://prettier.io/docs/en/options.html), 80 probably is the right level and better to normalize these files that direction as opposed to what is proposed in this change.

PR #9683, opened 2022-03-09
feat: PubSub Add Options objects for all Providers and fix: Spelling error and deprecation of old exports

stocaaro:
#### Description of changes
This PR covers a bit of housekeeping for the PubSub category. It includes:
- Adding Provider specific Options objects that will provider minor feedback, but don't in any way change the public interface.
- Add deprecation warning for `AWSAppSyncProvider` which has been entirely replaced by `AWSAppSyncRealtimeProvider`
- Fix the spelling of ProvidertOptions and MqttProvidertOptions 
  - Leaving aliases marked deprecated so we don't break anyone using these exports
- Prettier line length remapping to align with the default printWidth of 80

- Angular app with IoT integration

For each app listed above, I verified that the codebase shows deprecation warnings for deprecated resource (mostly to confirm I was pointed at the updated package). 

On the IoT integrated React app, I did deeper testing to verify that any of the ProviderOption interface type works against the IoT Provider, confirming

Unit tests run successfully.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -142,10 +143,22 @@ const START_ACK_TIMEOUT = 15000;
...
 
-const standardDomainPattern = /^https:\/\/\w{26}\.appsync\-api\.\w{2}(?:(?:\-\w{2,})+)\-\d\.amazonaws.com\/graphql$/i;
+const standardDomainPattern =
+	/^https:\/\/\w{26}\.appsync\-api\.\w{2}(?:(?:\-\w{2,})+)\-\d\.amazonaws.com\/graphql$/i;
 
 const customDomainPath = '/realtime';
 
+export interface AWSAppSyncRealTimeProviderOptions extends ProviderOptions {
+	appSyncGraphqlEndpoint?: string;
+	authenticationType?: string;
```
jamesaucode, line 153: I would make this slightly stricter. From the usage only a set of strings are supported.
```suggestion
	authenticationType?: 'API_KEY' | 'AWS_IAM' | 'OPENID_CONNECT' | 'AMAZON_COGNITO_USER_POOLS' | 'AWS_LAMBDA';
```

packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -34,11 +34,17 @@ export function mqttTopicMatch(filter: string, topic: string) {
 	return length === topicArray.length;
 }
 
-export interface MqttProvidertOptions extends ProvidertOptions {
+export interface MqttProviderOptions extends ProviderOptions {
 	clientId?: string;
 	url?: string;
+	aws_pubsub_endpoint?: string;
```
jamesaucode, line 40: It looks like we missed `aws_appsync_dangerously_connect_to_http_endpoint_for_testing?: boolean` here

jamesaucode (changes requested):
I have a few comments. Overall looks good!! 🎉

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: I had this in the interface before pushing the PR and it seemed so internal as to not be worth advertising for customer use. Thoughts? Happy to pull it back in. For completeness, I agree 100%.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -142,10 +143,22 @@ const START_ACK_TIMEOUT = 15000;
...
+	query?: string;
+	variables?: object;
+	apiKey?: string;
+	region?: string;
+	graphql_headers?: () => {};
```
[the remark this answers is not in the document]
stocaaro: Oh interesting. I'm tracing through this and it's a bit hard to parse.

[This section appears](https://github.com/aws-amplify/amplify-js/blob/main/packages/api-graphql/src/GraphQLAPI.ts#L97-L103) to guarantee that the value will be either undefined or a function. This aligns with [PubSub usage](https://github.com/aws-amplify/amplify-js/blob/main/packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts#L287) where it is only ever run as a function.

I think you're looking at [this section](https://github.com/aws-amplify/amplify-js/blob/main/packages/api-graphql/src/GraphQLAPI.ts#L294) where it is run/expanded outside of PubSub.  If I'm reading this correctly, it wouldn't override `graphql_headers` like a fulfilled promise, but more expand into other attributes.

I went down the rabbit hole of trying to find options being sent in and didn't have much luck. The interface is exposed for customers to call through API, so its hard to say there isn't different usage, but because we always call a defined value as if it where a function, I think this is safe.

Happy to sync up if I'm missing what you're seeing.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
stocaaro: Genius!  I hadn't looked at input, focusing on usage and this appears to cover the usecase. Will get an update in shortly.

packages/pubsub/src/Providers/MqttOverWSProvider.ts
jamesaucode: Good point, I definitely overlooked that. I would probably still want to have all the available `options` documented somewhere, we could still include it in the type but mark it as [internal](https://tsdoc.org/pages/tags/internal/)

Another way that might work but involves a bit more work is maybe you can keep the constructor parameter `MqttProviderOptions` as you have, but have another type definition for the class field `options`.
```typescript
protected get options(): InternalMqttProviderOptions { // just an example, you can name it something else
  return { ...this._config};
}
```
... but then we will also have to change `this._config` to `protected` in `AbstractPubSubProvider`
This way when we type `this.options.` it will give us autocomplete for the available options inside the class, but not when someone is calling the constructor. It also documents that it does have a `.aws_appsync_dangerously_connect_to_http_endpoint_for_testing` options

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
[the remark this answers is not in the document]
stocaaro: Oh, got it! I think I can imagine a scenario where a user might wire in auth in a way that requires an async method. I wonder if anyone is actually doing this in the wild. Would be good to get a test case in here that covers the use case. So to support this, I am adding `Promise<{}>` as a valid type for `graphql_headers`. Will get that out shortly.

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: We both noticed it, might as well include it. Meant to do that in my last revision, will do so with the next revision. You're right that I could approach it by having a second internal only type, I vote we keep it simple, but happy to collect more opinions.

elorzafe: I think this shouldnt be here, the idea of Mqtt provider is to not have anything related to aws on it.

elorzafe (commented):
Thanks @stocaaro 

I have a couple of comments, my concern is Mqtt provider to have aws options. We can chat more offline

packages/pubsub/src/Providers/AWSIotProvider.ts
```diff
@@ -10,12 +10,20 @@
...
  */
-import { MqttOverWSProvider } from './MqttOverWSProvider';
+import { MqttOverWSProvider, MqttProviderOptions } from './MqttOverWSProvider';
 import { Signer, Credentials } from '@aws-amplify/core';
 
 const SERVICE_NAME = 'iotdevicegateway';
 
+export interface AWSIoTProviderOptions extends MqttProviderOptions {
+	aws_pubsub_region?: string;
+}
```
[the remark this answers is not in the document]
stocaaro: Correct. Adding these interfaces gives guidance, but doesn't change the underlying structure expected since all of the keys are optional. Practically speaking, some of these inputs are non-optional, however we have to keep the common interface to not break [this interface](https://github.com/aws-amplify/amplify-js/blob/main/packages/pubsub/src/PubSub.ts#L136-L148)

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -28,15 +28,17 @@ import {
 } from '@aws-amplify/core';
 import Cache from '@aws-amplify/cache';
 import Auth from '@aws-amplify/auth';
+import { GRAPHQL_AUTH_MODE } from '@aws-amplify/api-graphql';
```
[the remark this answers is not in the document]
stocaaro: Can it? As in we might resolve the inter-dependency to different versions?  I had initially copied over the hardcoded list, then decided that it was better to refer to a canonical central list if one existed and this appears to be the closest to that.

The object itself shouldn't cause a cycle, since it's just an enum and doesn't cross refer to anything else. Can help me create the situation where this would cause the problem?

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: I go back and forth. It is currently part of the interface and I would prefer to make that explicit and then remove it at the major version as opposed to hiding it and increasing the risk of missing the removal.  Is there a pattern for deprecating named object attributes like this?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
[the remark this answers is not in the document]
elorzafe: The circular dependency is between packages `@aws-amplify/pubsub` depending on `@aws-amplify/api-graphql` and viceversa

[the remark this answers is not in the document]
stocaaro: Do you recommend moving this constant to core for use in both? Maybe this is a case where having a list of options for TS doesn't matter because its all an internal interface and we're better off falling back to a string type? We could also copy a link this direction to avoid the inter-dependence.  Which if these are you thinking is the better path to resolution?

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: Per this thread, in the most recent revision we move this option up to the IoT provider and eliminated the one with the longer name as being not something we want in the public interface - per a discussion with Francisco.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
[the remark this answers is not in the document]
stocaaro: I have moved it to core as suggested in my conversation with Francisco yesterday. Needs review from @manueliglesias  to confirm that this is the best place for this to live and not cause us problems.

elorzafe (approved):
LGTM! thanks @stocaaro 🌮

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -142,10 +144,24 @@ const START_ACK_TIMEOUT = 15000;
...
+type GraphqlAuthModes = keyof typeof GRAPHQL_AUTH_MODE;
+
+export interface AWSAppSyncRealTimeProviderOptions extends ProviderOptions {
+	appSyncGraphqlEndpoint?: string;
+	authenticationType?: GraphqlAuthModes;
+	query?: string;
+	variables?: object;
+	apiKey?: string;
+	region?: string;
+	graphql_headers?: () => {} | Promise<{}>;
```
jamesaucode, line 161: I believe `() => Promise<{}>` is also a valid usage here

jamesaucode (changes requested):
Just one comment, everything else LGTM 😄

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
stocaaro: Another excellent catch. Fixed in the latest version

stocaaro:
Got feedback from @manueliglesias . Recommended moving `GRAPHQL_AUTH_MODE` to `auth` instead of `core`. Migrated in the most recent commit.

stocaaro:
Well. Something is broken from the enum migration. Will continue debugging tomorrow.

jamesaucode (approved):
LGTM 🎉

PR #9691, opened 2022-03-10
fix: commonmarker upgrade in Gemfile

ArkamJ:
#### Description of changes
Upgraded commonmarker in GemFile to avoid dependency security alert

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


docs/Gemfile.lock
```diff
@@ -14,7 +14,7 @@ GEM
       execjs
     coffee-script-source (1.11.1)
     colorator (1.1.0)
-    commonmarker (0.17.13)
+    commonmarker (0.23.4)
```
david-mcafee, line 17: Just checking - did the install for this succeed. [Dependabot wasn't able to do this automatically](https://github.com/aws-amplify/amplify-js/security/dependabot/8) due to a conflict, so I was assuming we may need to make an update to `github-pages` and / or `jekyll-commonmark-ghpages`.

david-mcafee (commented):
Had one question on this change, thanks!

docs/Gemfile.lock
ArkamJ: Ohh I didn't check for that. How would I locally be able to check if the install succeeded?

PR #9692, opened 2022-03-11
fix: upgraded commonmarker in Gemfile

ArkamJ:
#### Description of changes
Upgraded commonmarker in Gemgile

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -14,7 +14,7 @@ GEM
       execjs
     coffee-script-source (1.11.1)
     colorator (1.1.0)
-    commonmarker (0.17.13)
+    commonmarker (0.23.4)
       ruby-enum (~> 0.5)
     concurrent-ruby (1.1.8)
     dnsruby (1.61.5)


docs/Gemfile.lock
```diff
@@ -14,7 +14,7 @@ GEM
       execjs
     coffee-script-source (1.11.1)
     colorator (1.1.0)
-    commonmarker (0.17.13)
+    commonmarker (0.23.4)
```
david-mcafee, line 17: Prior to updating, we should ensure the following tests pass:

1. `bundle install` does not fail
2. `yarn docs` succeeds.
3. The docs page runs locally.

If you've already done this, do you mind adding that to the PR description?

I'd say this manual testing is particularly important since we don't have any unit or E2E tests to validate that things are working as expected. It wouldn't be a bad idea for us to have a README and / or contributing guidelines nested in the docs folder. 

Thanks again for making this change, @ArkamJ!

david-mcafee (changes requested):
Can we exclude the output of `yarn docs` from this PR? I think it would be better to isolate the security fix from the docs generation.

docs/Gemfile.lock
david-mcafee: Hey @ArkamJ - we discussed this offline, and I know you manually tested this, but can you add a quick description of how you tested this to the description? Thanks!

iartemiev:
I agree @david-mcafee's request for changes. We should only include the dep bump, not the output of the yarn docs command. 
It’s good to test it locally to make sure the dep upgrade didn’t break anything, but the generated doc changes shouldn’t be in this PR.

david-mcafee (changes requested):
Left some comments above

iartemiev (approved):
LGTM 👍🏻

PR #9700, opened 2022-03-14
fix: Datastore/fix offline polling retry

jameswillis99:
#### Description of changes
I believe there should be three circumstances where offline should be monitored:

1. navigator.onLine (existing reachability)
2. Polling/websocket monitoring (in this PR i used polling but happy to use websockets with some pointers about how to implement)
3. Responsive to mutations e.g. if a mutation fails with a network error we float this up and set status to offline and pause mutation processing

Within this PR I am implementing point 2 & 3 above. 

As well as this we want to monitor what and if anything has been kicked from the outbox queue.  I've added a parameter on event outboxMutationProcessed called isDeadLetter which responds if an error causes a dequeue.

I also understand polling isn't what everyone would want so added an offline poll config option containing the poll interval if someone wants improved offline reachability turned on.

#### Issue #, if available
(https://github.com/aws-amplify/amplify-js/issues/9699)

#### Description of how you validated changes
Using local sample app

#### Checklist

- [ X] PR description included
- [ X] `yarn test` passes
- [ X] Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #9699, opened 2022-03-14
Datastore Sync Reachability uses Navigator.onLine resulting in data loss when incorrectly displaying online


jameswillis99:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

DataStore

### Amplify Categories

storage

### Environment information

<details>

```
  System:
    OS: macOS 11.6.4
    CPU: (8) x64 Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
    Memory: 918.55 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.0 - ~/.nvm/versions/node/v14.16.0/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v14.16.0/bin/yarn
    npm: 6.14.11 - ~/.nvm/versions/node/v14.16.0/bin/npm
  Browsers:
    Chrome: 99.0.4844.51
    Edge: 99.0.1150.39
    Safari: 15.3
  npmPackages:
    @babel/cli: 7.17.0 => 7.17.0 
    @babel/core: 7.17.2 => 7.17.2 (7.15.5, 7.17.5, 7.9.0, 7.12.9)
    @babel/preset-env: ^7.0.0 => 7.16.11 (7.3.4, 7.9.0)
    @babel/preset-react: ^7.0.0 => 7.16.7 (7.9.1)
    @types/blob-util:  1.3.3 
    @types/bluebird:  3.5.29 
    @types/chai:  4.2.7 
    @types/chai-jquery:  1.1.40 
    @types/jest: ^24.0.18 => 24.9.1 
    @types/jquery:  3.3.31 
    @types/lodash:  4.14.149 
    @types/minimatch:  3.0.3 
    @types/mocha:  5.2.7 
    @types/node: ^8.9.5 => 8.10.66 (17.0.21)
    @types/puppeteer: 1.3.0 => 1.3.0 
    @types/sinon:  7.5.1 
[...]

PR #9724, opened 2022-03-22
fix(@aws-amplify/datastore): fix mutations to retry indefinitely on network error

svidgen:
#### Description of changes

Fixes a bug wherein mutation retries due to service reachability issues eventually end, resulting in data loss. Data can become isolated to a single client. This change causes failures due to network errors to retry indefinitely until connectivity is restored.

Provides an alternative to reachability testing from #9700.

#### Issue #, if available

#9699

Tested in app by overriding DataStore reachability test to always show online and disabling network in browser. Observed retries continue "indefinitely", whereas they did not previously.

Added unit test to show that retry delay scales on large attempt values for network errors.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/mutation.test.ts
+++ b/packages/datastore/__tests__/mutation.test.ts
@@ -35,7 +35,7 @@ describe('Jittered retry', () => {
 			const lowerLimit = 2 ** attempt * 100;
 			const upperLimit = lowerLimit + 100;
 
-			if (lowerLimit < 205000) {
+			if (lowerLimit < 2 ** 12 * 100) {
 				console.log(
 					`attempt ${attempt} (${value}) should be between ${lowerLimit} and ${upperLimit} inclusively.`
 				);
--- a/packages/datastore/__tests__/mutation.test.ts
+++ b/packages/datastore/__tests__/mutation.test.ts
@@ -1,4 +1,7 @@
-import { MutationProcessor } from '../src/sync/processors/mutation';
+import {
+	MutationProcessor,
+	safeJitteredBackoff,
+} from '../src/sync/processors/mutation';
 import {
 	Model as ModelType,
 	PostCustomPK as PostCustomPKType,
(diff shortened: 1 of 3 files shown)


packages/datastore/src/sync/processors/mutation.ts
```diff
@@ -539,4 +537,37 @@ class MutationProcessor {
...
+const MAX_RETRY_DELAY_MS = 5 * 60 * 1000;
+const originalJitteredBackoff = jitteredBackoff(MAX_RETRY_DELAY_MS);
+
+/**
+ * @private
+ * Internal use of Amplify only.
+ *
+ * Wraps the jittered backoff calculation to retry Network Errors indefinitely.
+ * Backs off according to original jittered retry logic until the original retry
+ * logic hits its max. After this occurs, IFF the error is a Network Error, we
```
manueliglesias, line 549:
```suggestion
 * logic hits its max. After this occurs, if the error is a Network Error, we
```

packages/datastore/__tests__/mutation.test.ts
```diff
@@ -20,6 +23,53 @@ let Model: PersistentModelConstructor<ModelType>;
...
+
+		const backoffs = [...Array(COUNT)].map((v, i) =>
+			safeJitteredBackoff(i)
+		) as (number | boolean)[];
+
+		const isExpectedValue = (value, attempt) => {
+			const lowerLimit = 2 ** attempt * 100;
+			const upperLimit = lowerLimit + 100;
+
+			if (lowerLimit < 205000) {
```
manueliglesias, line 38: can we express this `205000` in terms of the `13` above?

manueliglesias (approved):
LGTM 🎉

packages/datastore/src/sync/processors/mutation.ts
svidgen: This was intended to be shorthand for [if and only if](https://en.wikipedia.org/wiki/If_and_only_if). But, "if" works here too. :)

packages/datastore/__tests__/mutation.test.ts
svidgen: Sure -- at some point, it might be better for tests like this to go the other way and explicitly *not* depend on repeating the math/logic. But, I opted to use the math here for clarity. So, I should use it all the way through!

The math won't actually be the same value; but that's OK.

Revision is pushed!

packages/datastore/src/sync/processors/mutation.ts
manueliglesias: TIL !! 😄 

Thanks

PR #9731, opened 2022-03-24
fix: PubSub - Resolve ts strict errors in the AWSAppSyncRealTimeProvider

stocaaro:
#### Description of changes
Using a [`tsconfig.json` set to strict](https://github.com/stocaaro/ts-starter/blob/main/tsconfig.json) as a guide, resolve all errors found in `pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts`.

Change includes a deprecation annotation for a constant that should have been included in the prior change.

`yarn test` is passing without issues

#### Checklist
- [x] PR description included
- [x] `yarn test` passes
- [x] Tests apps manually test


packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -54,7 +54,7 @@ type ObserverQuery = {
 	subscriptionState: SUBSCRIPTION_STATUS;
 	subscriptionReadyCallback?: Function;
 	subscriptionFailedCallback?: Function;
-	startAckTimeoutId?;
+	startAckTimeoutId?: NodeJS.Timer;
```
jamesaucode, line 57: It feels wrong to have a nodejs type here for a (mostly) client-side library, I would probably use `ReturnType<typeof setTimeout>` instead and let Typescript infer it.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -161,10 +161,16 @@ export interface AWSAppSyncRealTimeProviderOptions extends ProviderOptions {
...
+		canonicalUri: string;
+		payload: string;
+	};
+
 export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
-	private awsRealTimeSocket: WebSocket;
+	private awsRealTimeSocket?: WebSocket;
 	private socketStatus: SOCKET_STATUS = SOCKET_STATUS.CLOSED;
-	private keepAliveTimeoutId;
+	private keepAliveTimeoutId?: NodeJS.Timer;
```
jamesaucode, line 173: Same as above

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -280,10 +291,10 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 		// Having a subscription id map will make it simple to forward messages received
 		this.subscriptionObserverMap.set(subscriptionId, {
 			observer,
-			query,
-			variables,
+			query: query ?? '',
+			variables: variables ?? {},
 			subscriptionState,
-			startAckTimeoutId: null,
+			startAckTimeoutId: undefined,
```
jamesaucode, line 297: is there a reason why this changes from `null` to `undefined` ?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -329,16 +340,19 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
 			logger.debug({ err });
-			const { message = '' } = err;
-			observer.error({
-				errors: [
-					{
-						...new GraphQLError(`${CONTROL_MSG.CONNECTION_FAILED}: ${message}`),
-					},
-				],
-			});
+			if (err instanceof Error) {
```
jamesaucode, line 343: Looks like we are doing this to satisfy the `strict` ts config but are we certain that `err` is always going to be an instance of Error? Im worried there might be cases where the error is not an instance of error and its going to be ignored.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -788,23 +819,22 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 		apiKey,
 		region,
 		additionalHeaders,
-	}): Promise<any> {
-		const headerHandler = {
+	}: AWSAppSyncRealTimeProviderOptions): Promise<any> {
+		const headerHandler: { [key: string]: any } = {
```
jamesaucode, line 823: Nit: This look similar to the union from the last PR so we can probably use that too. So something like `{ [key in GraphQLAuthTypes]: any }` (not sure if thats the exact name)

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -874,11 +909,16 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
 			throw new Error('No credentials');
 		}
-		const creds = await Credentials.get().then(credentials => ({
-			secret_key: credentials.secretAccessKey,
-			access_key: credentials.accessKeyId,
-			session_token: credentials.sessionToken,
-		}));
+		const creds = await Credentials.get().then((credentials: any) => {
+			const { secretAccessKey, accessKeyId, sessionToken } =
+				Credentials.shear(credentials);
```
jamesaucode, line 914: is the extra `Credentials.shear` call needed here?

jamesaucode (commented):
Looks good overall, just have a few questions 😄

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
stocaaro: In the `ObserveQuery` type definition that this fits into, the `startAckTimeoutId` value is optional, which typescript permits to be undefined and `null` is an incompatible type. Opted to keep the shape and make the usage conform, however I could have resolved this by allowing null in the type, like the following:
```
startAckTimeoutId: ReturnType<typeof setTimeout> | null;
```
instead of
```
startAckTimeoutId?: ReturnType<typeof setTimeout>;
```
I don't have strong opinions, do you have a preference?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
stocaaro: Good call. Doing it.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
stocaaro: Good call. I've gone through a couple iterations in thinking about this.
```
const message = (err as Error)?.message ?? '';
```
would accomplish the same thing as the current code, but again, if semantically we're worried about it not being an Error at times, treating like an error and being paranoid that Error might not have the required message field is odd.
```
const message = (err as {message?: string})?.message ?? '';
```
This is probably closer to reality given the concern.  Thoughts / suggestions?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
stocaaro: At the time, it felt better than embedding an `as` clarification with these fields to make ts happy. Something about letting the core library carry the ambiguity as opposed to embedding an auth specific hack here.

I'll update it with the `as` version and see how the feedback goes.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
stocaaro: Good call. Have tweaked it to make this work.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
jamesaucode: I see, I think there's a minor difference between the 2. One means the key is not optional, but the value could potentially be null, the other one means the key is optional but if it exist the value has to be `ReturnType<typeof setTimeout>`.
It looks like this is just an optional key so I think using `undefined` here is correct.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -329,7 +340,7 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 			});
 		} catch (err) {
 			logger.debug({ err });
-			const { message = '' } = err;
+			const message = (err as { message?: string })?.message ?? '';
```
jamesaucode, line 343: I think this works fine, but it might be a bit harder to read with everything in line. Another idea I have is to define a type guard, something like this
```ts
// it seems like all we care is if the thrown object has a message key
function isError(error: unknown): error is { message: string } {
    return (error as { message?: string}).message !== undefined;
}
// or you can just check if its an instanceof Error
function isError(error: unknown): error is Error {
  return error instanceof Error;
}
```
Then we can do something like
```ts
const message = isError(err) ? err.message : '';
```

What do you think?

stocaaro: It's possible, but not likely that `instanceof` would cause a functionality change where an object with `message` would have worked before, but now because it doesn't match up completely with the shape of Error it no longer works.

I agree that my use of `as` is ugly and it occurs to me that a better translation of the replaced line is just using `[]` access. With TS Strict, using deconstructing assignments is unhappy unless ts knows the shape of the object being deconstructed, where it is happy as a clam accessing members directly (though those members come through as `any`, which shouldn't be a problem with primitives like `message`.

I like the clarity that comes from wrapping this behavior in a private method, but without a refactor to decompose the realtime provider into clearer chunks I think isolating smaller bits of code for clarity makes this file harder to navigate and reason around.

jamesaucode: Yea make sense, I personally prefer using typeguard but I don't have a strong opinion here, I think using `[]` looks clean enough in this case. :+1:

jamesaucode (commented):
One more comment otherwise LGTM :smiley:

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -210,7 +216,7 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 					options,
 					observer,
 					subscriptionId,
-				}).catch(err => {
+				}).catch<any>(err => {
```
elorzafe, line 219: Why is this needed?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -329,7 +340,7 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 			});
 		} catch (err) {
 			logger.debug({ err });
-			const { message = '' } = err;
+			const message = err['message']?.message ?? '';
```
elorzafe, line 343: isnt `err.message.message` that statement?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -374,27 +384,30 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 	}
 
 	// Waiting that subscription has been connected before trying to unsubscribe
-	private async _waitForSubscriptionToBeConnected(subscriptionId) {
-		const { subscriptionState } =
+	private async _waitForSubscriptionToBeConnected(subscriptionId: string) {
```
elorzafe, line 387: Maybe this could be refactor to receive `subscriptionObserver` instead of `subscriptionId`

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -437,13 +450,13 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 			setTimeout(this._closeSocketIfRequired.bind(this), 1000);
 		} else {
 			logger.debug('closing WebSocket...');
-			clearTimeout(this.keepAliveTimeoutId);
+			if (this.keepAliveTimeoutId) clearTimeout(this.keepAliveTimeoutId);
 			const tempSocket = this.awsRealTimeSocket;
 			// Cleaning callbacks to avoid race condition, socket still exists
-			tempSocket.onclose = undefined;
-			tempSocket.onerror = undefined;
+			tempSocket.onclose = null;
```
elorzafe, line 456: Is this a TS thing?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -874,11 +906,19 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
-			secret_key: credentials.secretAccessKey,
-			access_key: credentials.accessKeyId,
-			session_token: credentials.sessionToken,
-		}));
+		const creds = await Credentials.get().then((credentials: any) => {
+			const { secretAccessKey, accessKeyId, sessionToken } = credentials as {
+				secretAccessKey: any;
+				accessKeyId: any;
+				sessionToken: any;
+			};
```
elorzafe, line 914: I think we have a CredentialsInterface on Core package

elorzafe (changes requested):
Thanks @stocaaro 

I just have one minor issue to fix on that `err['message']`

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
stocaaro: I think I generally agree, but a typeguard makes me nervous in this case as I could imagine an object with a message that doesn't fully match the error type interface making a mess of that. Maybe I'm being over-paranoid or misunderstanding something about how that fits together. Probably a thing I should play around with some more.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
```diff
@@ -788,46 +815,48 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
 			AWS_IAM: this._awsRealTimeIAMHeader.bind(this),
 			OPENID_CONNECT: this._awsRealTimeOPENIDHeader.bind(this),
 			AMAZON_COGNITO_USER_POOLS: this._awsRealTimeCUPHeader.bind(this),
 			AWS_LAMBDA: this._customAuthHeader,
 		};
 
-		const handler = headerHandler[authenticationType];
-
-		if (typeof handler !== 'function') {
+		if (!authenticationType || !headerHandler[authenticationType]) {
```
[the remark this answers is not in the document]
stocaaro: Going back and forth on this one. On one hand not providing an authenticationType is different than providing an auth type that doesn't map to one of our auth functions. Typing the input would make providing an incorrect auth function name, its optional, so it could be undefined and in the current implementation, I believe this is the only corner case you can get to in TS without overriding a string to tell ts that its one of the valid inputs.

Right now, we would log either:
```
Authentication type undefined not supported
```
or
```
Authentication type invalid_option not supported
```
Which seems clear enough to me. As a user with missing or incorrect auth type information, what logged output would you prefer to see to guide you to fixing your problem?

While working through this, I was bothered that the function type on our function map had `any` for the values, so I have narrowed that to specify what we know about them.

elorzafe (approved):
LGTM! Thanks @stocaaro 🌮

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts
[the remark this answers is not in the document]
jamesaucode: Now that I look at it again I think `Authentication type undefined not supported` is clear enough to me. Another option which will make the code a bit more verbose will be to have a separate `if` to handle the case where `authenticationType` is undefined, logging something like `authenticationType must be defined`.

jamesaucode:
Dont think the `authenticationType` thing is blocking, LGTM 🎉 !

jamesaucode (approved):

PR #9732, opened 2022-03-24
fix: PubSub - Resolve ts strict errors in the MqttOverWSProvider

stocaaro:
#### Description of changes
Using a [`tsconfig.json` set to strict](https://github.com/stocaaro/ts-starter/blob/main/tsconfig.json) as a guide, resolve all errors found in `pubsub/src/Providers/MqttOverWSProvider.ts`.

`yarn test` is passing without issues

#### Checklist
- [x] PR description included
- [x] `yarn test` passes
- [x] Tests apps manually test


packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -171,7 +181,10 @@ export class MqttOverWSProvider extends AbstractPubSubProvider {
 	}
 
 	protected async disconnect(clientId: string): Promise<void> {
-		const client = await this.clientsQueue.get(clientId, () => null);
+		const client = await this.clientsQueue.get(
+			clientId,
+			() => new Promise(() => undefined)
```
jamesaucode, line 186: Is this meant to be a placeholder Promise to satisfy the type constraint, or should it actually resolve to an `undefined` ?

jamesaucode (commented):
LGTM overall just 1 question

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: The ts error that prompted this change is:
```
Type 'null' is not assignable to type 'Promise<any>'.ts(2322)
```

The complete interface is:
```
ClientsQueue.get(clientId: string, clientFactory: (input: string) => Promise<any>): Promise<any>
```

This was probably the wrong fix, the clientsQueue appears to exist to avoid having multiple instances of the same client. In this instance, we're getting the client in order to disconnect and remove it. If it doesn't exist, this code will add it mapped to a null producing promise which we will try to disconnect and then remove. We don't need to add it in this case if it doesn't exist. Instead of this fix, I am changing the queue above, making the factory optional and returning undefined in the case where it doesn't exist, which will be wrapped in a promise by the async method handler.

packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -47,24 +47,27 @@ export type MqttProvidertOptions = MqttProviderOptions;
 class ClientsQueue {
 	private promises: Map<string, Promise<any>> = new Map();
 
-	async get(clientId: string, clientFactory: (string) => Promise<any>) {
-		let promise = this.promises.get(clientId);
-		if (promise) {
-			return promise;
+	async get(clientId: string, clientFactory?: (input: string) => Promise<any>) {
```
elorzafe, line 50: Why is optional?

packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -47,24 +47,27 @@ export type MqttProvidertOptions = MqttProviderOptions;
...
+		if (clientFactory) {
+			const newPromise = clientFactory(clientId);
 
-		this.promises.set(clientId, promise);
+			this.promises.set(clientId, newPromise);
 
-		return promise;
+			return newPromise;
+		}
+		return undefined;
```
elorzafe, line 63: I am not sure if this could break something.

elorzafe (commented):
Have a couple of question we can discuss offline. The rest looks good!

Thanks @stocaaro

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: Sure. James and I had a good conversation about how I had resolved the clientFactory requirement down on ~195 (see earlier revision and conversation). This caused me to read the code for intent and recognize that the diconnect logic only wants to get a connection if it exists, where the clientFactory it was providing was intended to be a no-op.

Recognizing this caused me to propose this refactor that makes clientFactory optional and returns undefined explicitly to make the disconnect usecase more straightforward.

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: Lets talk about it later today.

elorzafe (approved):
LGTM, thanks for clarifying @stocaaro 🌮

PR #9743, opened 2022-03-28
feat(@aws-amplify/geo): add geofence APIs

TreTuna:
#### Description of changes

- Bring Geofence developer preview code into `main` for GA release.
- Code was reviewed and released under the `@geo` tag from `geo/main` for developer preview since 1/11/22
- Adds Geofence APIs
  - `saveGeofences`
  - `getGeofence`
  - `listGeofences`
  - `deleteGeofences`
- Adds validation utility functions for geoJSON data structures
- 2/3 of the code in this PR is tests and types.

- Code has been tested and released since 1/11/22.
- Unit testing
- E2E tests
- Sample applications

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/geo/src/types/Geo.ts
```diff
@@ -100,3 +105,68 @@ export interface Place {
...
+
+// Options object for listGeofence
+export type ListGeofenceOptions = GeofenceOptions & {
+	nextToken?: string;
+};
+
+// Results options for listGeofence
+export type ListGeofenceResults = {
+	entries: Geofence[];
+	nextToken: string;
```
elorzafe, line 165: isnt this optional?

packages/geo/src/util.ts
```diff
@@ -0,0 +1,171 @@
...
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
+ * the License. A copy of the License is located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+import booleanClockwise from '@turf/boolean-clockwise';
```
elorzafe, line 13: Is the open source license in good shape? does it have dependencies with license in good shape?

packages/geo/src/util.ts
```diff
@@ -0,0 +1,171 @@
...
+	GeofenceInput,
+	GeofencePolygon,
+	LinearRing,
+} from './types';
+
+export function validateCoordinates(lng: Longitude, lat: Latitude): void {
+	if (!Number.isFinite(lng) || !Number.isFinite(lat)) {
+		throw new Error(`Invalid coordinates: [${lng},${lat}]`);
+	}
+	if (lat < -90 || lat > 90) {
```
elorzafe, line 27: Should this be `<=` and `>=`?

packages/geo/src/util.ts
```diff
@@ -0,0 +1,171 @@
...
+		throw new Error('Latitude must be between -90 and 90 degrees inclusive.');
+	} else if (lng < -180 || lng > 180) {
```
elorzafe, line 29: Should this be `<=` and `>=`?

packages/geo/src/util.ts
```diff
@@ -0,0 +1,171 @@
...
+	if (booleanClockwise(linearRing)) {
+		throw new Error(
+			`${geofenceId}: LinearRing coordinates must be wound counterclockwise`
+		);
+	}
+}
+
+export function validatePolygon(
+	polygon: GeofencePolygon,
+	geofenceId?: string
```
elorzafe, line 94: Should this be required?

packages/geo/src/util.ts
```diff
@@ -0,0 +1,171 @@
...
+	if (!geofenceIdRegex.test(geofenceId)) {
+		throw new Error(
+			`Invalid geofenceId: '${geofenceId}' - IDs can only contain alphanumeric characters, hyphens, underscores and periods.`
+		);
+	}
+}
+
+export function validateLinearRing(
+	linearRing: LinearRing,
+	geofenceId?: string
```
elorzafe, line 49: Should this be required?

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -333,6 +358,331 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+		const results: SaveGeofencesResults = {
+			successes: [],
+			errors: [],
+		};
+
+		const batches = [];
+
+		while (PascalGeofences.length > 0) {
+			// Splice off 10 geofences from input clone due to Amazon Location Service API limit
+			batches.push(PascalGeofences.splice(0, 10));
```
elorzafe, line 408: I would define a constant for the limit

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -333,6 +358,331 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+						batch,
+						options?.collectionName || this._config.geofenceCollections.default
+					);
+				} catch (error) {
+					// If the API call fails, add the geofences to the errors array and move to next batch
+					batch.forEach(geofence => {
+						results.errors.push({
+							geofenceId: geofence.GeofenceId,
+							error: {
+								code: 'APIConnectionError',
```
elorzafe, line 427: This is for network error right? Maybe define a constant, is this doing exponential backoff?

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -333,6 +358,331 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+			logger.debug(error);
+			throw error;
+		}
+
+		// Create Amazon Location Service Client
+		const client = new LocationClient({
+			credentials: this._config.credentials,
+			region: this._config.region,
+			customUserAgent: getAmplifyUserAgent(),
+		});
```
elorzafe, line 557: You could have a function that creates the client (I also saw the same code on other places)

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -333,6 +358,331 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+	/**
+	 * Delete geofences from a geofence collection
+	 * @param geofenceIds: string|string[]
+	 * @param options?: GeofenceOptions
+	 * @returns {Promise<DeleteGeofencesResults>} - Promise that resolves to an object with:
+	 *  successes: list of geofences successfully deleted
+	 *  errors: list of geofences that failed to delete
+	 */
+	public async deleteGeofences(
+		geofenceIds: string[],
```
elorzafe, line 618: What happens if `geofenceIds` is an empty Array

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -333,6 +358,331 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+	/**
+	 * Create geofences inside of a geofence collection
+	 * @param geofences - Array of geofence objects to create
+	 * @param options? - Optional parameters for creating geofences
+	 * @returns {Promise<AmazonLocationServiceSaveGeofencesResults>} - Promise that resolves to an object with:
+	 *   successes: list of geofences successfully created
+	 *   errors: list of geofences that failed to create
+	 */
+	public async saveGeofences(
+		geofences: GeofenceInput[],
```
elorzafe, line 370: What happen if the Array is empty?

packages/geo/src/Geo.ts
```diff
@@ -185,17 +197,124 @@ export class GeoClass {
...
+	): Promise<SaveGeofencesResults> {
+		const { providerName = DEFAULT_PROVIDER } = options || {};
+		const prov = this.getPluggable(providerName);
+
+		// If single geofence input, make it an array for batch call
+		let geofenceInputArray;
+		if (!Array.isArray(geofences)) {
+			geofenceInputArray = [geofences];
+		} else {
+			geofenceInputArray = geofences;
```
elorzafe, line 234: Maybe should check here if the Array is empty? or is by design that if the Array is empty is ok to not fail?

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -333,6 +358,331 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+		};
+		const command = new GetGeofenceCommand(commandInput);
+
+		// Make API call
+		let response: GetGeofenceCommandOutput;
+		try {
+			response = await client.send(command);
+		} catch (error) {
+			logger.debug(error);
+			throw error;
```
elorzafe, line 511: If the `geofence` is not found it would fail with a friendly error message?

elorzafe (commented):
In general looks very good, just have some comments we can discuss.

Thanks @TreTuna 🌮

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: Hmmm... Good point here. I covered all validations except for an empty array. So, at the moment, it would make a call with the empty array and just not do anything.

I'll add a validation here to nix out quick if the array is empty.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: I'm good with that.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: So, yes and no... 
This is for network errors, but it's not due to rate limits, so there's no exponential backoff.

We are creating an error array to save errors that come back from the API data structure, which is:
```
{
  successes: [],
  errors: []
}
```

 However, if the API call itself fails, we do not get this object, but still need to add them to the error object that we return to the customer so that they know what Geofences failed. Thus, this is adding those error messages to the return object.

 I'm not sure that a constant would be good here because it needs to have information about exactly what Geofences were included in the failed network call, so it needs to be somewhat dynamic.

packages/geo/src/util.ts
TreTuna: No, this should not be inclusive. -90 and 90 are valid `lat` coordinates.

packages/geo/src/util.ts
TreTuna: No, this should not be inclusive. -180 and 180 are valid `lng` coordinates. This looks for numbers outside of that range.

packages/geo/src/util.ts
TreTuna: The other way to write this would be:
```
if (lat >= -90 && lat <= 90)
```
Would that be more clear for you?

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: Yes, I've thought about that as a refactor. I'm going to add an `engineering excellence` ticket in our backlog to do this in the future.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: The error message currently would come straight through from Amazon Location Service, which is:

```
Uncaught (in promise) ResourceNotFoundException: no geofence with given name found
```

We could catch that and make it a little more explicit with the exact given `geofenceId` if we wanted to, however, considering this only accepts one, I left it as is.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: Added a check for this.

packages/geo/src/types/AmazonLocationServiceProvider.ts
```diff
@@ -1,5 +1,60 @@
...
+	status: AmazonLocationServiceGeofenceStatus;
+};
+
+// List Geofences
+export type AmazonLocationServiceListGeofenceOptions = ListGeofenceOptions & {
+	collectionName?: string;
+};
+
+// Delete Geofences
+export type AmazonLocationServiceBatchGeofenceErrorMessages =
```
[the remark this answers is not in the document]
TreTuna: These specific ones are pushed straight through to the user in an array of errors for batch calls. We do not change these specific error messages, but add them to the errors array that is returned to the user so that they can see what geofences failed for exactly what reason here.

packages/geo/src/types/Geo.ts
TreTuna: The `nextToken` is optional in the input, but not the results.

packages/geo/src/util.ts
TreTuna: Yes
License for all Turf libraries is MIT: https://github.com/Turfjs/turf/blob/master/LICENSE
And this package only has dependencies on other Turf libraries and `tslib`, which is BSD Zero Clause: https://github.com/microsoft/tslib/blob/main/LICENSE.txt

packages/geo/src/Geo.ts
TreTuna: We could do that here, however, I left it to the `provider` to handle validating inputs.

packages/geo/src/util.ts
TreTuna: I'd rather it's not, in order to make it more universal. However, I understand why you are questionsing this, so I'm going to update the messages instead.

TreTuna: I would rather it not be required. Will update the function to make this work better with optional `geofenceId`.

packages/geo/src/util.ts
TreTuna: I would rather it not be required. Will update the function to make this work better with optional `geofenceId`.

packages/geo/src/types/Geo.ts
elorzafe: But if is `undefined`?

packages/geo/src/util.ts
elorzafe: You are right!

packages/geo/src/types/Provider.ts
```diff
@@ -35,10 +41,33 @@ export interface GeoProvider {
...
+		options?: GeofenceOptions
+	): Promise<SaveGeofencesResults>;
+
+	// get a single geofence
+	getGeofence(
+		geofenceId: string,
+		options?: ListGeofenceOptions
+	): Promise<Geofence>;
+
+	// 	list all geofences
```
jamesaucode, line 65: nit^2: there's an extra space here

packages/geo/src/types/Provider.ts
```diff
@@ -35,10 +41,33 @@ export interface GeoProvider {
 	// get the map resource listed as default
 	getDefaultMap(): MapStyle;
 
+	// search by a text string and return a list of places
```
jamesaucode, line 44: I would prefer using JSDoc/TSDoc here, which helps with integrating with tools like `TypeDoc` + it shows you the comment when you look at the function signature with an IDE. What do you think?

packages/geo/src/Geo.ts
```diff
@@ -163,17 +176,124 @@ export class GeoClass {
...
+			return await prov.saveGeofences(geofenceInputArray, options);
+		} catch (error) {
+			logger.debug(error);
+			throw error;
+		}
+	}
+
+	/**
+	 * Get a single geofence by geofenceId
+	 * @param geofenceId: string
```
jamesaucode, line 226: Nit: It seems like these params doesn't have a proper description. Also we probably don't need to specify the types here inside a Typescript file in comments (compared to JSDoc), we already have the type definitions in code. If we change 1 in the future we will need to change both the code & the comment which makes this slightly harder to maintain.

jamesaucode (commented):
Looks good overall, mostly nit picks 😄

packages/geo/src/types/Geo.ts
TreTuna: On the last page, `nextToken: undefined` would be the return.
image

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: Hmmm... Yeah, this is a good thought and we could do something like that. This current method was basically taken from the flow of how to get credentials I saw in another category.

packages/geo/src/types/Provider.ts
TreTuna: Sync'd offline to verify: Since we use JS/TSDoc comments in the implementation of the Providers themselves, which is where TypeDoc and IDEs would pull the signatures from anyway. Thus, this is all good.

packages/geo/src/types/Provider.ts
TreTuna: Oh... My... Goodness... This is unacceptable. I'm so sorry! I'll take care of this RIGHT AWAY! 🙃

packages/geo/src/types/Geo.ts
elorzafe: correct, so in that case the accurate type should be `string | undefined`

packages/geo/src/types/Geo.ts
```diff
@@ -97,3 +102,68 @@ export interface Place {
...
+};
+
+// Results object for deleteGeofence
+export type DeleteGeofencesResults = {
+	successes: string[];
```
elorzafe, line 167: Are these ids?

packages/geo/src/util.ts
```diff
@@ -0,0 +1,173 @@
...
+	GeofenceInput,
+	GeofencePolygon,
+	LinearRing,
+} from './types';
+
+export function validateCoordinates(lng: Longitude, lat: Latitude): void {
+	if (!Number.isFinite(lng) || !Number.isFinite(lat)) {
+		throw new Error(`Invalid coordinates: [${lng},${lat}]`);
+	}
+	if (lat < -90 || lat > 90) {
```
elorzafe, line 27: nit, it reads better IMO like this

`if (lat < -90 || 90 < lat)`

packages/geo/src/util.ts
```diff
@@ -0,0 +1,173 @@
...
+		throw new Error('Latitude must be between -90 and 90 degrees inclusive.');
+	} else if (lng < -180 || lng > 180) {
```
elorzafe, line 29: nit, it reads better IMO like this

`if (lng < -90 || 90 < lng)`

elorzafe (commented):
A couple of nit comments and I would update one type.

Also not sure about using `successes`, can you tell me more about that?

packages/geo/src/util.ts
TreTuna: I get that. No problem there!

packages/geo/src/util.ts
TreTuna: I get that. No problem there!

packages/geo/src/util.ts
```diff
@@ -0,0 +1,173 @@
...
+			throw new Error(`Geofence '${geofenceId}' is missing geometry.polygon`);
+		}
+		const { polygon } = geometry;
+
+		// Validate polygon length and structure
+		try {
+			validatePolygon(polygon, geofenceId);
+		} catch (error) {
+			if (
+				error.message === `Polygon has more than the maximum 1000 vertices.`
```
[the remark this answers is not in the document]
TreTuna: Yep, you are correct there, this was from before I started doing the prefixes. I should be able to fix this by using `error.message.includes()` instead.

Also, you found the one validation it looks like I failed to write a test for! 🤦‍♂️ Will add that as well.

packages/geo/src/types/Geo.ts
TreTuna: Yes, they are. I get how that's not apparent though. Should we add a comment here or make a `GeofenceID` type?

packages/geo/src/types/Geo.ts
TreTuna: You are correct, updated!

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
TreTuna: Marked these notes in a ticket on our engineering excellence board.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -254,6 +279,340 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+					},
+				};
+			}
+		);
+		const results: SaveGeofencesResults = {
+			successes: [],
+			errors: [],
+		};
+
+		const batches = [];
```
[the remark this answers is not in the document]
TreTuna: Added better variable names and types.

packages/geo/src/util.ts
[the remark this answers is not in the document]
TreTuna: Updated and added a test for this

jamesaucode:
LGTM 🎉

elorzafe (approved):
LGTM! Thanks @TreTuna 🌮 🎉

thaddmt (approved):

PR #9753, opened 2022-03-31
feat(data): Datastore Docs

david-mcafee:
#### Description of changes

# **NOTE: TO SEE FORMATTED CHANGES AS WELL AS RENDERED DIAGRAMS, [TAKE A LOOK AT THE BRANCH](https://github.com/aws-amplify/amplify-js/blob/datastore-docs/packages/datastore/README.md) 🙂**

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/docs/conflict-resolution.md
```diff
@@ -0,0 +1,6 @@
+# Conflict Resolution
+- **AppSync is the source of truth for conflict resolution**
+   1. In the event AppSync fails to resolve a conflict, the network response will contain an error message (`conflict unhandled`). This is how we give customers the chance to make an update, or try again.
+   2. We use jittered retry (10x).
```
svidgen, line 4: Example of a place that could use a little more detail in the future, plus some links to how this retry logic occurs, especially since the logic is pretty scattered -- _and layered_ -- with different layers of this giant request processing stack performing different types of retries and potentially interacting with each other in weird ways.

packages/datastore/docs/conflict-resolution.md
```diff
@@ -0,0 +1,6 @@
...
+- We err on the side of not deleting customer data when performing conflict resolution.
+- Auto-merge is the default resolution strategy. This relies on the version, and will attempt to merge fields that changed when possible.
```
svidgen, line 6: Another example, could maybe just stand to have a link to a section in the appsync documentation.

svidgen (approved):
Overall, I think this is a really good foundation! Thank-you for doing this!! (And, I really like the addition of diagrams right in the markdown like this!)

There are a few places I'd love to see folks adding more detail in the future. I've noted just a couple of examples in the comments. Over time I think we'll also want sections dedicate to special configurations around sync intervals, "unsupported hacks", etc. It may be a little touchy to document some of these things correctly, because there are some workarounds we don't want to broadcast b/c some hacks are not supported and may disappear or change at any time, but we need folks who work on the codebase to be aware of.

If you were to change anything for this PR though, I'd just ask for a big message at the top of the README to *please* update the docs any time they're found to be incorrect or lacking. In particular if a line in the docs prompts a question, I'd love for folks to figure out the answer and update the docs with the necessary detail.

packages/datastore/docs/conflict-resolution.md
david-mcafee: > Overall, I think this is a really good foundation! Thank-you for doing this!! (And, I really like the addition of diagrams right in the markdown like this!)
> 
> There are a few places I'd love to see folks adding more detail in the future. I've noted just a couple of examples in the comments. Over time I think we'll also want sections dedicate to special configurations around sync intervals, "unsupported hacks", etc. It may be a little touchy to document some of these things correctly, because there are some workarounds we don't want to broadcast b/c some hacks are not supported and may disappear or change at any time, but we need folks who work on the codebase to be aware of.
> 
> If you were to change anything for this PR though, I'd just ask for a big message at the top of the README to _please_ update the docs any time they're found to be incorrect or lacking. In particular if a line in the docs prompts a question, I'd love for folks to figure out the answer and update the docs with the necessary detail.

Thank you for your kind words!! 

So that these items don't get lost, I'll add some `TODO`s. Re: "unsupported hacks", perhaps we can sync offline to discuss the best path forward there. 

Also, re: the callout on the primary README, I'll add that in another commit, now.

david-mcafee: > Example of a place that could use a little more detail in the future, plus some links to how this retry logic occurs, especially since the logic is pretty scattered -- _and layered_ -- with different layers of this giant request processing stack performing different types of retries and potentially interacting with each other in weird ways.

Will add TODO

packages/datastore/docs/conflict-resolution.md
david-mcafee: Done.

dpilch (approved):
This really helped my understanding!

david-mcafee:
> This really helped my understanding!

Thanks!!

manueliglesias (approved):
LGTM 😄 

Thank you for putting this together

PR #9775, opened 2022-04-06
New err handler

ArkamJ:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/datastore/src/types.ts
```diff
@@ -658,7 +658,7 @@ export type DataStoreConfig = {
 	DataStore?: {
 		authModeStrategyType?: AuthModeStrategyType;
 		conflictHandler?: ConflictHandler; // default : retry until client wins up to x times
-		errorHandler?: (error: SyncError) => void; // default : logger.warn
+		errorHandler?: (error: SyncError<any>) => void; // default : logger.warn
```
dpilch, line 661: Is it possible to use `PersistentModel` here to avoid the use of `any`.

packages/datastore/src/types.ts
```diff
@@ -775,13 +775,39 @@ export type SyncConflict = {
...
+	| 'BadRecord'
+	| 'Unauthorized'
+	| 'Transient'
+	| 'Unknown';
+
+type ProcessName =
+	| 'sync'
+	| 'mutate'
+	| 'subscribe'
+	| 'save()'
```
dpilch, line 801: nit: consistent use of parentheses after process name. Either remove or add to all.

packages/datastore/src/sync/utils.ts
```diff
@@ -40,6 +40,22 @@ enum GraphQLOperationType {
 	GET = 'query',
 }
 
+export type OperationName = 'create' | 'update' | 'delete' | 'list' | 'get';
+export function TransformerMutationTypeToOperationName(
+	operation: TransformerMutationType
+): OperationName {
+	switch (operation) {
```
dpilch, line 47: Could simplify this with `toLowerCase`.

packages/datastore/src/sync/processors/sync.ts
```diff
@@ -232,6 +230,25 @@ class SyncProcessor {
 							);
 
 							if (error.errors) {
+								for (const err of error.errors) {
```
dpilch, line 233: Consider changing this to `Promise.all` or `Promise.allSettled` to allow the async to run in parallel rather than in sequence.

```js
await Promise.all(error.error.map(async (err) => { ...

```

dpilch (changes requested):

svidgen:
This was used as the basis for another PR, which has been merged.

PR #9778, opened 2022-04-07
feat(pubsub): Add test coverage for the AWSAppSyncRealTimeProvider

stocaaro:
#### Description of changes
This change increases the unit test coverage of the AWSAppSyncRealTimeProvider from ~18% to ~95%.

To accomplish this makes minor changes to allow websocket creation to be replaced/spied on with a helper interface for exchanging fake graphql content with the provider.

This adds ~25 seconds to the unit test execution time.

Changes:
- Constants moved into created provider folder in separate file to allow for mocking over timeouts in some tests.
- Provider code moved to `index.ts` in the new provider folder
- `newWebsocket` method added to provider to expose interface for mocking over
- `helper.ts` file added to test to wrap constant mocking method, fake web socket and the interface for testing with the fake websocket
- Lots of new tests to cover most of the realtime provider behavior

Running the PubSub unit tests a large number of times and the full package unit tests several times to ensure the test suite continues to be stable / reproducible.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are added to cover behavior


packages/pubsub/__tests__/helpers.ts
```diff
@@ -0,0 +1,185 @@
+import * as constants from '../src/Providers/AWSAppSyncRealTimeProvider/constants';
+
+export class FakeWebSocketInterface {
```
cshfang, line 3: Have you considered using a community built web socket mock? I'm concerned that maybe
1. This is could be a non-trivial amount of mock logic to maintain
2. It is a potentially non-intuitive mock since it sort of behaves like both a websocket provider and server as opposed to delineating them
3. Relying on timeouts can be pretty fragile and generally not all that desirable for unit tests. Is this due to our Jest version being out of date and therefore we cannot reliable control time within the Jest sandbox?

packages/pubsub/__tests__/helpers.ts
```diff
@@ -0,0 +1,185 @@
...
+		});
+		let closeResolver: (value: PromiseLike<undefined>) => void;
+		this.hasClosed = new Promise((res, rej) => {
+			closeResolver = res;
+		});
+		this.webSocket = new FakeWebSocket(closeResolver);
+	}
+
+	async triggerOpen() {
+		// After an open is triggered, it takes a few ms for it to enact provider behavior
```
cshfang, line 22: What does this mean?

packages/pubsub/__tests__/helpers.ts
```diff
@@ -0,0 +1,185 @@
...
+	async closeInterface() {
+		//
+		await this?.triggerClose();
+		// Wait for either hasClosed or a half second has passed
+		await new Promise(res => {
+			// The interface is closed when the socket "hasClosed"
+			this.hasClosed.then(() => res(undefined));
+
+			// The provider can get pretty wrapped around itself,
+			// but its safe to continue after half a second, even if it hasn't closed the socket
```
cshfang, line 45: What would cause the interface fail to "close"? I'm kind of concerned that this is masking (or will potentially one day mask) issues either with the code or with the testing.

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -0,0 +1,879 @@
...
+					await fakeWebSocketInterface?.triggerOpen();
+					await fakeWebSocketInterface?.handShakeMessage();
+					await fakeWebSocketInterface?.sendDataMessage({
+						type: MESSAGE_TYPES.GQL_DATA,
+						payload: { data: {} },
+					});
+					await subscription.unsubscribe();
+					await fakeWebSocketInterface?.triggerClose();
+
+					expect(fakeWebSocketInterface.hasClosed).resolves.toBe(undefined);
```
cshfang, line 534: Nit: You could use the built-in function
```suggestion
					expect(fakeWebSocketInterface.hasClosed).resolves.toBeUndefined();
```

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -0,0 +1,879 @@
...
+							'DEBUG',
+							'Authenticating with AWS_IAM'
+						);
+					});
+
+					test('authenticating with AWS_IAM without credentials', async () => {
+						expect.assertions(1);
+
+						jest.spyOn(Credentials, 'get').mockImplementation(() => {
+							return Promise.resolve(undefined);
```
cshfang, line 647: Nit question: Why explicitly pass `undefined` instead of just doing `Promise.resolve()`?

packages/pubsub/__tests__/helpers.ts
```diff
@@ -0,0 +1,185 @@
+import * as constants from '../src/Providers/AWSAppSyncRealTimeProvider/constants';
+
+export class FakeWebSocketInterface {
+	readonly webSocket: FakeWebSocket;
+	readyForUse: Promise<undefined>;
+	hasClosed: Promise<undefined>;
+
+	private readyResolve: (value: PromiseLike<undefined>) => void;
```
cshfang, line 8: Nit: Can these be `any` as the generic instead of `undefined`? Seems unnecessary to force these to resolve specifically to `undefined`

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -0,0 +1,879 @@
...
+							.subscribe({});
+						await fakeWebSocketInterface?.readyForUse;
+
+						expect(loggerSpy).toBeCalledWith(
+							'DEBUG',
+							'Authenticating with AWS_LAMBDA'
+						);
+					});
+
+					test('authenticating with AWS_LAMBDA without Autorization', async () => {
```
cshfang, line 851: Nit typo:
```suggestion
					test('authenticating with AWS_LAMBDA without Authorization', async () => {
```

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -0,0 +1,879 @@
...
+							'Authenticating with API_KEY'
+						);
+					});
+
+					test('authenticating with AWS_IAM', async () => {
+						expect.assertions(1);
+
+						jest.spyOn(Credentials, 'get').mockImplementation(() => {
+							return Promise.resolve({});
+						});
```
cshfang, line 618: Nit: You should be able to use `mockResolvedValue` or `mockRejectedValue` to make this (and other places this pattern appears) more readable 🍬 
```suggestion
						jest.spyOn(Credentials, 'get').mockResolvedValue({});
```

cshfang (commented):
Unit testing can be a drag so thank you for raising the bar on this! Came across a bunch of nits - mostly typo kind of things. But also had some questions

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -0,0 +1,879 @@
...
+		test('returns an observable', () => {
+			const provider = new AWSAppSyncRealTimeProvider();
+			expect(provider.subscribe('test', {})).toBeInstanceOf(Observable);
+		});
+
+		describe('returned observer', () => {
+			test('returns error when no appSyncGraphqlEndpoint is provided', done => {
+				expect.assertions(1);
+				const provider = new AWSAppSyncRealTimeProvider();
+				let observerError: { errors: GraphQLError } | undefined = undefined;
```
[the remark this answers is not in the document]
stocaaro: Yep. Left over from when I didn't know I could use `expect.assertions(1)` to verify the right number of assertions get hit.  I had a messy manual way of making sure the assertion happened and this is left over from that.

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -0,0 +1,879 @@
...
+			});
+
+			describe('connection logic with mocked websocket', () => {
+				let fakeWebSocketInterface: FakeWebSocketInterface;
+				let loggerSpy: jest.SpyInstance;
+				let provider: AWSAppSyncRealTimeProvider;
+
+				beforeEach(async () => {
+					fakeWebSocketInterface = new FakeWebSocketInterface();
+					loggerSpy = jest.spyOn(Logger.prototype, '_log');
```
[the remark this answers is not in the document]
stocaaro: I like it.

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -0,0 +1,879 @@
...
+					provider = new AWSAppSyncRealTimeProvider();
```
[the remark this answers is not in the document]
stocaaro: Maybe I'm incorrect, but I think if I moved it out of the beforeEach closure it would only happen once.  It's really important for test isolation that we get a fresh provider for each test.

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -0,0 +1,879 @@
...
+
+					// Saving this spy and resetting it by hand causes badness
+					//     Saving it causes new websockets to be reachable across past tests that have not fully closed
+					//     Resetting it proactively causes those same past tests to be dealing with null while they reach a settled state
+					jest.spyOn(provider, 'newWebSocket').mockImplementation(() => {
+						fakeWebSocketInterface.newWebSocket();
+						return fakeWebSocketInterface.webSocket;
+					});
```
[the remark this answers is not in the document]
stocaaro: In this comment I'm trying to summarize ~4 hours of troubleshooting where having the spy stored to a variable caused new webSockets to be assigned to providers from tests that have run previously. The provider is layers of asyncs and observables where the websockets `onerror`, `onmessage` and `onclose` methods are overwritten at different stages in the provider lifecycle. As a result the old test provider would override the new tests methods with its own.

This is my attempt to document this and give a clue if similar issues come up again without writing an essay inline. Happy for suggestions.

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -0,0 +1,879 @@
...
+
+					const observer = provider.subscribe('test', {
+						appSyncGraphqlEndpoint: 'ws://localhost:8080',
+					});
+
+					const subscription = observer.subscribe({
+						// Succeed only when the first message comes through
+						next(x) {
+							done();
+						},
```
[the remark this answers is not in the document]
stocaaro: Excellent! This is what I wanted. Thank you, applying to all the cases that rely on done

packages/pubsub/__tests__/helpers.ts
stocaaro: I started out with the two prominent solutions `jest-websocket-mock` and `mock-socket`.

jest-websocket-mock creates a local socket server listening on an ip for your app to interact with during testing.  This has a lot of appeal, except that the issues with bleed over between tests means we need a fresh websocket for each test and the added time to bring up an independent websocket server is impractical for unit tests.

mock-socket's recommended integration pattern is to assign over `window.WebSocket` or `global.WebSocket`, neither of which worked for me. The `newWebsocket()` mocking pattern I've adopted here got me further, however I repeatedly wished for more control which led me to investigate the approach that I landed on, to fill in the parts of the interface I needed for testing. I would have preferred not to do this and if there's a well supported alternative, I'm all for it.

1. Totally agreed and open to suggestions.
2. Your right that the tests act as the websocket server when they push data through the `FakeWebSocketInterface` and as the provider client when they make calls to the client. It is necessary to interact from both vantage points to exercise the logic and maybe there's a better naming convention to make that clear. Looking back at `mock-socket` they use the `Server` name to represent what I'm calling an Interface.
3. The timeout business here is very fragile and undesirable. Far better would be to have observers or promises to await for resolution with each action, but the provider doesn't offer a surface to interact with these. I experimented with jest's ability to control timeouts, but it involved adding a lot of extra time control logic to the majority of these tests. Above all, this is a strong case for refactoring this provider to offer a more introspection on the state changes, but wanting to not mix test coverage with a refactor, I settled on adding timeouts, which I have rigorously verified are the right length for reliability without being too long. I'm very new to jest and if there's a better way, I'm eager to learn it.

packages/pubsub/__tests__/helpers.ts
stocaaro: Seems reasonable.

packages/pubsub/__tests__/helpers.ts
stocaaro: Changing to the following for clarity (and updating the other instances to be similar)
```
// After an open is triggered, the provider has logic that must execute
//   which changes the function resolvers assigned to the websocket
```

To repeat, I would prefer to have events to await from the provider while it is in state transition, but that isn't surfaced by the current implementation.

packages/pubsub/__tests__/helpers.ts
stocaaro: Good question. The reason to attempt to close the interface is to avoid socket events in prior tests overwriting the resolver methods during future tests. I'm a little bit mystified that this kind of cross test side-effect is possible the way this is written, but I have observed it and this was the final line of defense that stopped it from happening.

Calling the onclose method isn't guaranteed to close the socket. There are 3 different onclose methods that can be assigned to the close resolver we trigger. 2 of them reject a promise that bubbles up the stack where the third directly closes out the socket. In all cases they lead to final states where the resolvers will no longer be overwritten, but they do not always lead to the websocket itself being closed, triggering the `hasClosed` behavior this is awaiting.

I'd be really happy for a second set of eyes on this code if you would like a tour. I'm trying to do the simplest possible thing around provider code that can be a bit mysterious to me after hours of walking around it. As you've pointed out above, there are multiple places where I got too close to find the obvious right thing and maybe theres some of that here.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/constants.ts
```diff
@@ -0,0 +1,107 @@
+import { ZenObservable } from 'zen-observable-ts';
```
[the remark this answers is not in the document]
stocaaro: This probably shouldn't have been brought across the the constants file. Moving the dependent code back over.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -176,6 +83,10 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 	private subscriptionObserverMap: Map<string, ObserverQuery> = new Map();
 	private promiseArray: Array<{ res: Function; rej: Function }> = [];
 
+	newWebSocket(url, protocol) {
```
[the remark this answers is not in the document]
stocaaro: Updated to add get as a prefix like you suggested. Kept new because I'm replacing a call to `new` and it just feels right.

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
stocaaro: This appears to do what I need, so I made the recommended change

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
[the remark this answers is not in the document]
cshfang: Oh I mean that you could do this with a top-level `beforeEach`

stocaaro:
Chris and I have had a good back and forth about using [`mock-socket`](https://github.com/thoov/mock-socket) in place of the custom mock code included in this PR. Chris rewrote some of these tests and I've extended his work. With `mock-socket` I believe we can achieve ~70% coverage. The difference is that this provider includes a substantial amount of custom logic to handle the connection/handshake order of events where `mock-socket` couples all of this into a single action that can't be decomposed / run as individual pieces.

`mock-socket` is built for testing applications that communicate over websockets, such as chat apps. In our case, this code is focused primarily on connecting and disconnecting from appsync where handeling messages represents a small portion of what the provider does. With `mock-socket` we can get happy path coverage of connecting and disconnecting.

Using `mock-socket`, we could add a second implementation of their websocket api that allows us to decompose the open/handshake event process. The existing implementation is ~200 lines of code, so I would expect this to add roughly that to our codebase.

On the other hand, the mock logic I've included in this change is minimal and made to fit our use/need precisely. It is about ~200 lines of code itself and gives us 95% code coverage where there are only a few behavioral sections of the code that aren't reached by these tests.

This line of inquiry has led to several valuable improvements in this code, but I would propose that spending an additional 4-8 hours of dev-time to achieve the same results with more lines of code in our codebase (or accepting 70% coverage to live within the `mock-socket` framework constraints) would be a less optimal result than what we already have in this PR.

cshfang:
As @stocaaro mentioned above, we had several conversations about this. My concerns are essentially as follows:
1. We are adding an API to the provider in order to allow the dependency injection necessary to inject a mocked websocket object.
2. With a custom fake approach, we are adopting maintenance overheads of making sure the fake is accurate and working to spec.
3. We potentially mask deficiencies or extraneous code by coercing our fake implementation to behave in ways that a socket connection may not behave in reality. If a code path is difficult to test - maybe it does not need to exist or maybe it needs to be refactored
4. Timeouts in unit tests can be fragile and unnecessarily increase runtime. This is something that Aaron has tried to address with later commits but not quite able to completely eliminate.

In the end, my take is that less coverage with more real-world tests would be preferable over more coverage if we end up massaging our test utilities to achieve it. To be clear, this is not an indictment on the accuracy of Aaron's implementation or approach. But I do think it potentially sets up future abuse of the utility (e.g. hey this code path is not being called, I need to add x to the utility) and a more "standardized" approach inspires more confidence to me that we at least don't our fake utils is not the source of some perceived bug.

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -0,0 +1,870 @@
...
+						error(err) {
+							expect(err.errors[0].message).toEqual(
+								'Subscribe only available for AWS AppSync endpoint'
+							);
+							mockError();
+						},
+					});
+
+					// Resolve the subscription
+					await Promise.resolve();
```
elorzafe, line 101: Why not use `done` instead  or wrap `provider.subscriber` into a `new Promise`?

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -0,0 +1,870 @@
...
+				});
+
+				test('subscription waiting for onopen with ws://localhost:8080 goes untranslated', async () => {
+					expect.assertions(1);
+
+					const newSocketSpy = jest
+						.spyOn(provider, 'getNewWebSocket')
+						.mockImplementation(() => {
+							fakeWebSocketInterface.newWebSocket();
+							return fakeWebSocketInterface.webSocket;
```
elorzafe, line 113: Maybe you can use `done` inside of the mock implementation so you dont have to await

elorzafe (commented):
Thanks for doing this @stocaaro 

We can chat more offline but my only concern is maybe you can use `done` inside of mocks to avoid adding delay on your test.

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
stocaaro: These used to be timeouts. Chris showed me that a Promise resolve yields the main thread long enough to execute through a bit of code in the provider, guaranteed. You're correct that in this case and a few others, I can wrap the subscribe or message send in a promise resolve to achieve the same effect.  I've made this change.

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
stocaaro: The trouble is getting it to wait long enough that the provider code to be initialized with the endpoint. I tried using the promise.resolve pattern here, but the init isn't triggered. I wonder if I have or can add a promise to the socket that triggers after it is initialized. I'll take a look at this and get back to you.

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
stocaaro: Adopted the wrapping of subscribers to remove some of these.  In the remaining timeout dependent cases, these are the only way I have found to yield the thread for long enough that the provider code passes through a phase transition to reach the next state needed for test progression. Wherever possible these have been replaced with Promise resolves, but there's probably still room for improvement.

More than anything, it would be helpful to have an event stream to monitor and await in testing. This will be a feature proposed in reconnect and with that addition, I would propose to revist/refactor these tests.

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
stocaaro: Good news on this one. The `readyForUse` promise on the fake socket interface works in place of this wait.

elorzafe (approved):
LGTM thanks for doing this @stocaaro 🌮 🎉

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/constants.ts
```diff
@@ -0,0 +1,95 @@
...
+) as Symbol;
+
+export const AWS_APPSYNC_REALTIME_HEADERS = {
+	accept: 'application/json, text/javascript',
+	'content-encoding': 'amz-1.0',
+	'content-type': 'application/json; charset=UTF-8',
+};
+
+/**
+ * Time in milleseconds to wait for GQL_CONNECTION_INIT message
```
cshfang, line 83: Nit: Here and others down below
```suggestion
 * Time in milliseconds to wait for GQL_CONNECTION_INIT message
```

cshfang (approved):
Thanks for raising the coverage

PR #9786, opened 2022-04-07
feat: Upgrade uuid and refactor all references to core for ease of future upgrades

stocaaro:
#### Description of changes
Our use of uuid 3.X.X across the amplify library raises error in Angular due to this preventing optimization and leading to larger bundle sizes. While upgrading to 8.X.X, lets also consolidate our usage to core such that future upgrades can be made in a single place.

Change includes:
- Add the latest uuid version to `@aws-amplify/core`
- Use `uuid` name to encapsulate behavior as opposed to adding entire uuid interface directly to core
- Remove all direct package dependence on uuid
- Migrate package use to encapsulated syntax `uuid` => `uuid.v4`
- Fix for analytics test that mocks over the uuid library

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/6202

Unit tests are happy.

Verdaccio testing locally and will be running the circle suite with Ivan's help tomorrow.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/analytics/src/Providers/AWSPinpointProvider.ts
+++ b/packages/analytics/src/Providers/AWSPinpointProvider.ts
@@ -19,6 +19,7 @@ import {
 	JS,
 	Hub,
 	getAmplifyUserAgent,
+	uuid,
 } from '@aws-amplify/core';
 import {
 	EventsBatch,
@@ -37,13 +38,13 @@ import {
 	EventObject,
 	EndpointFailureData,
 } from '../types';
-import { v1 as uuid } from 'uuid';
 import EventsBuffer from './EventBuffer';
 
-const AMPLIFY_SYMBOL = (typeof Symbol !== 'undefined' &&
-typeof Symbol.for === 'function'
-	? Symbol.for('amplify_default')
-	: '@@amplify_default') as Symbol;
+const AMPLIFY_SYMBOL = (
+	typeof Symbol !== 'undefined' && typeof Symbol.for === 'function'
+		? Symbol.for('amplify_default')
+		: '@@amplify_default'
+) as Symbol;
 
 const dispatchAnalyticsEvent = (event, data) => {
 	Hub.dispatch('analytics', { event, data }, 'Analytics', AMPLIFY_SYMBOL);
@@ -165,7 +166,7 @@ export class AWSPinpointProvider implements AnalyticsProvider {
 		const timestamp = new Date().getTime();
 		// attach the session and eventId
 		this._generateSession(params);
-		params.event.eventId = uuid();
+		params.event.eventId = uuid.v1();
 
 		Object.assign(params, { timestamp, config: this._config });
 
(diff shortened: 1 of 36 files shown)


packages/datastore/src/sync/datastoreReachability/index.native.ts
```diff
@@ -1,4 +1,7 @@
 import { Reachability } from '@aws-amplify/core';
 import { default as NetInfo } from '@react-native-community/netinfo';
+import Observable from 'zen-observable-ts';
 
-export const ReachabilityMonitor = new Reachability().networkMonitor(NetInfo);
+export const ReachabilityMonitor = new Reachability().networkMonitor(
+	NetInfo
+) as Observable<{ online: boolean }>;
```
stocaaro, line 7: This seems unrelated and unnecessary, but TS won't compile without it.  If anyone can help explain why this was required with this change, I would appreciate knowing what happened here.

david-mcafee (commented):
Are we certain that every package necessarily will always require Amplify Core? For instance, when DataStore is used in local-only mode, it's unnecessary to call `Amplify.configure(awsExports)` (since there isn't a provisioned backend). In this case, I'm almost certain there is no reliance on Amplify Core (will verify). I'm wondering if there are other packages that also do not necessarily rely on the Core package.

ashika01:
Plus one. Also, we did try to merge an uuid upgrade but there was also some issues with DS and react native. @iartemiev can talk more to this..

stocaaro:
Excellent points. Maybe I should revert the DS portion of this. I worked with Ivan earlier to run the integ suite against this change and it turned up 2 tests that break:

https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/10565/workflows/cd15078c-e61d-4e22-ad18-57b5f41f6595

I haven't had capacity to investigate them today, but will be chasing them down early next week.

packages/analytics/src/Providers/AWSPinpointProvider.ts
```diff
@@ -206,14 +207,14 @@ export class AWSPinpointProvider implements AnalyticsProvider {
...
 	private _generateSession(params) {
-		this._sessionId = this._sessionId || uuid();
+		this._sessionId = this._sessionId || uuid.v1();
 		const { event } = params;
 
 		switch (event.name) {
 			case SESSION_START:
 				// refresh the session id and session start time
 				this._sessionStartTimestamp = new Date().getTime();
-				this._sessionId = uuid();
```
ashika01, line 216: why `uuid.v1` instead of `v4`? if multiple session gets generate from client app wouldn't there be a slighest possibility that timestamp might be same?

ashika01: NVM. we are already doing that? might be question for wider audience.

ashika01: My understanding of ID was wrong. NVM

ashika01:
Can we get an answer to this one? @stocaaro 
> Are we certain that every package necessarily will always require Amplify Core?

I dont think it might be good idea to move `uuid` to core.it will tie the all packages to core. Whats the purpose for moving this to `core`?

stocaaro:
> Can we get an answer to this one? @stocaaro
> 
> > Are we certain that every package necessarily will always require Amplify Core?
> 
> I dont think it might be good idea to move `uuid` to core.it will tie the all packages to core. Whats the purpose for moving this to `core`?

I had a conversation with Sam where he indicated that centralizing UUID so that we could manage the version in one place was the way to go. Had asked that I coordinate with James to make that happen, but James hadn't heard the same.

Core seems pretty interwoven into everything, is there an effort to move amplify away from core use? If so, who should I be talking to to ensure I'm inline with this? Is this captured in a best practices doc somewhere?

stocaaro:
> issues with DS and react native

This change causes an issue in one RN test that I am still troubleshooting. Will get back to you with more later this week.

manu2504:
Could this move forward, so the amplify users do not feel ashamed that the only package Angular complains heavily about is from their IaaS provider?

Issue #9470, opened 2022-01-12
Error saving post TypeError: Cannot read properties of undefined (reading 'version')


goldwareJapan:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

DataStore

### Amplify Categories

storage

### Environment information

<details>

```
  System:
    OS: macOS 12.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 20.88 GB / 64.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.22.1 - ~/.nvm/versions/node/v12.22.1/bin/node
    Yarn: 1.22.17 - ~/.yvm/shim/yarn
    npm: 8.3.0 - ~/.nvm/versions/node/v12.22.1/bin/npm
    Watchman: 2021.10.18.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 97.0.4692.71
    Firefox: 95.0.2
    Safari: 15.2
  npmGlobalPackages:
    @aws-amplify/cli: 7.6.8
    @nestjs/cli: 8.1.5
    aws-cdk: 2.3.0
    expo-cli: 4.12.10
    npm: 8.3.0
    nx: 13.2.3
    pm2: 5.1.2
    prisma: 3.5.0
    typescript: 4.5.2



```

</details>

### Describe the bug

DataStore.Query not working. 

It was working at the office. I walked home and it gives this error. 

`Error saving post TypeError: Cannot read properties of undefined (reading 'version')`

![Screen Shot 2022-01-12 at 19 45 16](https://user-images.githubusercontent.com/35557320/149126652-3be674b7-f14a-47ca-9124-474a88f176af.png)

schema is 

```
type Project @model @auth(rules:  [
    # Admin users can access any operation.
    { allow: groups, groups: ["Admin"] }

    # Each record may specify which groups may read them.
    { allow: groups, groupsField: "company", operations: [read] }
  ]) {
[...]

PR #9854, opened 2022-04-29
Peer deps

dpilch:
#### Description of changes

This change is adding peer deps to direct dependencies of `aws-amplify`. See below for reasoning for why this is necessary.

The change also adds the necessary scripts for automatically updating the peer deps with a release because this is not directly supported by lerna: https://github.com/lerna/lerna/issues/955. The version update script is needed twice due strange behavior with lerna TODO.

##### Overview

Adding aws-amplify and @aws-amplify/datastore independently can result in some dependencies being installed twice. This results in various bugs with the DataStore singleton.

This class of bugs is expected to affect any category with a singleton that is independently installed.

##### Reproduction steps

1. Create a react app (or any front-end app)
2. yarn add aws-amplify@4.3.20
3. yarn add @aws-amplify/datastore@3.9.0
    1. The installed version must be outside of the semver range in aws-amplify
    2. https://github.com/aws-amplify/amplify-js/blob/main/packages/aws-amplify/package.json#L36-L50
    3. The semver range for aws-amplify@4.3.20 is set to exactly @aws-amplify/datastore@3.10.0
    4. This an issue with all versions. The versions listed are just an example.
4. Interact with DataStore
5. Observe the error in console.

```
IndexedDBAdapter.ts:115 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'namespaces')
    at IndexedDBAdapter.<anonymous> (IndexedDBAdapter.ts:115)
    at step (InMemoryStore.ts:41)
    at Object.next (InMemoryStore.ts:41)
    at InMemoryStore.ts:41
    at new Promise (<anonymous>)
    at __awaiter (InMemoryStore.ts:41)
    at upgrade (IndexedDBAdapter.ts:113)
    at IDBOpenDBRequest.<anonymous> (index.js:16)
```

##### Root Cause

These errors are happening because the schema has not been initialized. Schema initialization is in codegen and is added to a customer project. `src/models/index.js`.

```
import { DataStore, initSchema } from '@aws-amplify/datastore';
import { schema } from './schema';
const { Todo } = initSchema(schema);
```

initSchema also has a side effect of setting the schema variable that is available via closure in the DataStore instance. 
schema variable: https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/src/datastore/datastore.ts#L96
DataStore instance: https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/src/datastore/datastore.ts#L1586-L1589

The above error occurs when two instances of DataStore are created, but only one has the initialized schema. In the case were npm does not dedupe @aws-amplify/datastore the package will exist twice in node_modules.

These two import statements will pull the package from two distinct locations in node_modules. Therefore the DataStores that are imported are different instances.

```
import { DataStore } from '@aws-amplify/datastore';
// node_modules/@aws-amplify/datatstore
import { DataStore } from 'aws-amplify';
// node_modules/aws-amplify/node_modules/@aws-amplify/datastore
```

DataStore imported via @aws-amplify/datastore has the initialized schema because initSchema was imported via @aws-amplify/datastore (the same package location). 

I have found 2 scenarios where npm does not dedupe the @aws-amplify/datastore package.

1) Using npm/yarn link will sometimes not dedupe. This is only an issue for local development of amplify-js so solutions to this will not be discussed in this doc.

2) Explicitly installing incompatible versions of the packages.
If the version of @aws-amplify/datastore is outside the semver range specified in aws-amplify npm will not dedupe the packages. For example aws-amplify@4.3.20 specifies @aws-amplify/datastore@3.10.0 as a dependency. If the customer has the exact version @aws-amplify/datastore@3.9.0 listed in their own dependencies both version 3.9.0 and 3.10.0 will be installed. npm will not attempt to dedupe different versions of the same package.

In the above case @aws-amplify/datastore in the customers dependencies is behind the required version from aws-amplify. The inverse will also cause the issue. i.e. @aws-amplify/datastore@3.10.0 and aws-amplify@4.3.19.

If the customer used a wider range of @aws-amplify/datastore version, npm would recognize the possibility for using the same version and dedupe.

Example Installs

*No dedupe (broken):*
```
"dependencies": {
  "aws-amplify": "4.3.20", // requires @aws-amplify/datastore@3.10.0
  "@aws-amplify/datastore": "3.9.0",
}
```

```
> npm list @aws-amplify/datastore
multiple@0.1.0 /Users/dppilche/apps/multnpm
├── @aws-amplify/datastore@3.9.0
└─┬ aws-amplify@4.3.20
  └── @aws-amplify/datastore@3.10.0
```

*No dedupe (broken):*

```
"dependencies": {
  "aws-amplify": "4.3.19", // requires @aws-amplify/datastore@3.9.0
  "@aws-amplify/datastore": "3.10.0",
}
```

```
> npm list @aws-amplify/datastore
multiple@0.1.0 /Users/dppilche/apps/multnpm
├── @aws-amplify/datastore@3.10.0
└─┬ aws-amplify@4.3.19
  └── @aws-amplify/datastore@3.9.0
```

*Has dedupe (working):*

```
"dependencies": {
  "aws-amplify": "4.3.20", // requires @aws-amplify/datastore@3.10.0
  "@aws-amplify/datastore": "^3.9.0", // installs 3.10.0 because npm recognizes opportunity to dedupe
}
```

```
> npm list @aws-amplify/datastore
multiple@0.1.0 /Users/dppilche/apps/multnpm
├── @aws-amplify/datastore@3.10.0
└─┬ aws-amplify@4.3.20
  └── @aws-amplify/datastore@3.10.0 dedupe
```

#### Issue #, if available

#9470 
Likely others as well.

To do: add testing logic to prevent install.

Testing that each release script correctly updates the peer deps:
- [ ] publish:main
- [x] publish:beta
- [x] publish:release
- [ ] publish:1.0-stable
- [ ] publish:ui-components/main
- [ ] publish:verdaccio
- [ ] publish:geo/main

#### Checklist

- [ ] PR description included


package.json
```diff
@@ -27,7 +27,9 @@
 		"publish:1.0-stable": "lerna publish --conventional-commits --yes --dist-tag=stable-1.0 --message 'chore(release): Publish [ci skip]' --no-verify-access",
 		"publish:ui-components/main": "lerna publish --canary --force-publish \"*\" --yes --dist-tag=ui-preview --preid=ui-preview --exact --no-verify-access",
 		"publish:verdaccio": "lerna publish --no-push --canary minor --dist-tag=unstable --preid=unstable --exact --force-publish --yes --no-verify-access",
-		"publish:geo/main": "lerna publish --canary --force-publish \"*\" --yes --dist-tag=geo --preid=geo --exact --no-verify-access"
+		"publish:geo/main": "lerna publish --canary --force-publish \"*\" --yes --dist-tag=geo --preid=geo --exact --no-verify-access",
+		"version": "scripts/update-peer-deps.sh",
```
nickarocho, line 31: Could this script name be a bit more descriptive to the actions it represents, as well as follow the naming convention of the other scripts? Maybe `version:peerDeps` or `update-peer-deps` or `version:peers` or something like that? `version` alone seems a little vague to me.

nickarocho (commented):
Thanks so much for this fix and the detailed explanation, @dpilch!

**A couple questions:**
1. Would `yarn` and `pnpm` throw the same errors as `npm` to help customers catch this problem? (i.e. would `yarn add [explicit @aws-amplify/* version]` trigger the same warning as `npm install ...` if adding a dep outside of the peer range?)
2. The Lerna bug that requires us to run the script twice seems odd to me. What would happen if (theoretically) the Lerna bug disappears? Would the script running twice bump the version of the peerDeps twice as well, resulting in version mismatches?
3. Would it be worthwhile to add an integ test to the pipeline to automatically run the manual testing you listed in the PR description? I would think it would be very beneficial being that this directly impacts our release pipeline.

I left a comment about the name of the `version` script being a bit too vague. LMK if you disagree or want some help brainstorming more descriptive names.

Thanks again! This will be a big step forward for that entire class of bugs you've identified.

scripts/update-peer-deps.sh
```diff
@@ -0,0 +1,13 @@
+# Updates the peer dependency version for all direct dependencies of aws-amplify
+# Lerna does not automatically update peer dependency versions: https://github.com/lerna/lerna/issues/955
+# This script runs with the version and prepublishOnly lifecycle events
+# The script needs to run on both lifecycle events:
+# 1. lerna increments version numbers
+# 2. update-peer-deps.sh updates the peer deps to new version (version lifecycle event)
+# 3. lerna creates release commit. i.e. chore(release) ...
+# 4. lerna changes the peers deps back to old version (appears to bug in lerna)
```
nickarocho, line 8:
```suggestion
# 4. lerna changes the peers deps back to old version (appears to be a bug in lerna)
```

package.json
dpilch: `version` is necessary to tap into the lifecycle events. https://github.com/lerna/lerna/tree/main/commands/publish#lifecycle-scripts

We could make a second more descriptive script name, but we would still need to set `version` to call the other.

dpilch:
1. I tested with just npm, so I'll need to get back to you on yarn.
2. There is no harm to script running again because it pull the version number from `packages/aws-amplify/package.json` after lerna does the auto versioning based on conventional commits. The peer deps script itself doesn't bump any versions, just copies from the package.json.
3. Good idea, I'll work on adding a test.

packages/auth/package.json
```diff
@@ -46,6 +46,9 @@
     "amazon-cognito-identity-js": "5.2.8",
     "crypto-js": "^4.1.1"
   },
+	"peerDependencies": {
+		"aws-amplify": "4.3.20"
```
reesscot, line 50: Maybe I'm missing something, but doesn't this create a circular dependency?

dpilch: No, imo peer dependencies is a misleading name. This is specifying that version 4.5.2 of `@aws-amplify/auth` only works with version 4.3.20 of `aws-amplify` not that `@aws-amplify/auth` actually depends on `aws-amplify`. https://nodejs.org/es/blog/npm/peer-dependencies/

dpilch:
yarn only provides a warning on failed peer deps (and not a very loud one either). I'll look into if we can make the install fail.

```
> yarn add @aws-amplify/datastore@3.11.1
yarn add v1.22.18
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
warning " > @aws-amplify/datastore@3.11.1" has incorrect peer dependency "aws-amplify@4.3.22".
warning "@aws-amplify/datastore > @aws-amplify/pubsub@4.4.1" has incorrect peer dependency "aws-amplify@4.3.22".
warning "@aws-amplify/datastore > @aws-amplify/core@4.5.4" has incorrect peer dependency "aws-amplify@4.3.22".
warning "@aws-amplify/datastore > @aws-amplify/api@4.0.40" has incorrect peer dependency "aws-amplify@4.3.22".
warning "@aws-amplify/datastore > @aws-amplify/auth@4.5.4" has incorrect peer dependency "aws-amplify@4.3.22".
warning "@aws-amplify/datastore > @aws-amplify/pubsub > @aws-amplify/cache@4.0.42" has incorrect peer dependency "aws-amplify@4.3.22".
[4/4] 🔨  Building fresh packages...
success Saved lockfile.
success Saved 1 new dependency.
info Direct dependencies
└─ @aws-amplify/datastore@3.11.1
info All dependencies
└─ @aws-amplify/datastore@3.11.1
✨  Done in 7.50s.
```

pnpm also only provides a warning.

```
> pnpm install @aws-amplify/datastore@3.11.1
 WARN  Issues with peer dependencies found
.
└─┬ @aws-amplify/datastore
  ├── ✕ unmet peer aws-amplify@4.3.22: found 4.3.21
  └─┬ @aws-amplify/api
    ├── ✕ unmet peer aws-amplify@4.3.22: found 4.3.21
    └─┬ @aws-amplify/api-graphql
      ├─┬ @aws-amplify/api-rest
      │ └─┬ @aws-amplify/core
      │   └── ✕ unmet peer aws-amplify@4.3.22: found 4.3.21
      ├─┬ @aws-amplify/auth
      │ ├── ✕ unmet peer aws-amplify@4.3.22: found 4.3.21
      │ └─┬ @aws-amplify/cache
      │   └── ✕ unmet peer aws-amplify@4.3.22: found 4.3.21
      └─┬ @aws-amplify/pubsub
        └── ✕ unmet peer aws-amplify@4.3.22: found 4.3.21
```

dpilch:
I'm not seeing a way to cause `pnpm` or `yarn` to fail on unmet peer deps. We will need to decide if we continue with this approach. I think having only a warning is better than not.

dpilch:
I do have two other solutions to this problem that could be used in any combination. 

### Amplify CLI Check

The Amplify CLI would add a dependency check to an existing command. For example the check could run on push/pull or codegen. The dependency check would get all versions of @aws-amplify/* packages installed. It would then compare all versions for equality. i.e. @aws-amplify/datastore@3.10.0 != @aws-amplify/datastore@3.9.0

The check takes around 1.5 seconds total.

```
const { execSync } = require("child_process");

const npmList = JSON.parse(execSync("npm list --json --depth=1").toString());
const { dependencies: directDependencies } = npmList;
const { dependencies: transientDependencies } =
  directDependencies["aws-amplify"];
const amplifyPackages = Object.keys(transientDependencies).filter(
  (dependency) => dependency.startsWith("@aws-amplify/")
);

const packageVersions = amplifyPackages.map((packageName) => {
  const directVersion = directDependencies[packageName]?.version;
  const transientVersion = transientDependencies[packageName].version;
  return { packageName, directVersion, transientVersion };
});

const disparateVersions = packageVersions
  .filter(
    ({ packageName, directVersion, transientVersion }) =>
      directVersion !== undefined && directVersion !== transientVersion
  )
  .map(
    ({ packageName, directVersion, transientVersion }) =>
      `${packageName}: ${directVersion} and ${transientVersion} installed.`
  );

if (disparateVersions.length > 0) {
  throw new Error(
    [
        "Multiple version of Amplify packages installed. This could lead to problems with your app. See <docs webpage> on how to fix the versions.",
    ].concat(disparateVersions).join("\n")
  );
}
```

### Runtime Check

Add logic to the affected @aws-amplify/* packages to check if that singletons have not been properly initialized before utilizing any variable that could be empty. For example, with DataStore the schema variable may not be initialized but the code assumes that it has been initialized. If the schema variable is not initialized provide a meaningful error message and steps for remediation.

https://github.com/aws-amplify/amplify-js/compare/main...dpilch:multiple-versions

iartemiev:
I'm in favor of a runtime check over the other proposed solutions. 

I'm not sure that establishing a peer dependency relationship between the scoped packages (e.g., `@aws-amplify/core`, `/api`, `/datastore`, etc.) and the shared package, `aws-amplify` is the correct approach. Peer dependencies allow plugin packages to designate which host package (and versions thereof) the plugin is compatible with. The implication is that the host package of a particular version is necessary for the plugin to function. That's why NPM@7 or greater will automatically install peer dependencies and throw an error in the event that there's an incompatible version already installed. 
I don't think that behavior is desirable in our case. None of our scoped packages require `aws-amplify` to run. If someone just wants to install the Auth category, I don't think we should automatically install `aws-amplify` for them as well.

dpilch:
Closing this PR in favor of the runtime check.

PR #9860, opened 2022-05-02
feat: Selective TypeScript strict builds

stocaaro:
#### Description of changes
As we adopt TS Strict for our packages, to maintain this high standard, our build process should compile with the expected TS strict configuration and IDE feedback should align with the.

- Since we compile for both ES5 and ES6, we need to have the option of turning TS Strict on for these builds
- The IDE will pay attention to a tsconfig.json file in the package or higher, so adding a minimal config to get IDE feedback
- Fixing two PubSub strict issues that complication caught

I headed down the path of consuming `tsconfig.json` files merged into the `build.js` script to keep these in parity, but the ts version used by vscode is not the same as the version used in build. While these are different, I am recommending we keep the build config separate from the config that will be referenced by vscode.

#### Issue #, if available
https://github.com/aws-amplify/amplify-js/issues/7188

`yarn build` succeeds - particularly for pubsub

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/pubsub/tsconfig.json
+++ b/packages/pubsub/tsconfig.json
@@ -1,0 +1,14 @@
+//WARNING: The tsconfig used to build amplify is inline in ../../scripts/build.js, this file exists only to provide IDE feedback while modifying the library.
+//         This file intentionally sets ts to strict globally as an aspiration where not all packages have strict turned on in the actual build process.
+{
+	"compilerOptions": {
+		"rootDir": "./",
+		"target": "es5",
+		"module": "commonjs",
+		"moduleResolution": "node",
+		"strict": true,
+		"noImplicitAny": false,
+		"useUnknownInCatchVariables": false
+	},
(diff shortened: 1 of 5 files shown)


packages/pubsub/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+//         This file intentionally sets ts to strict globally as an aspiration where not all packages have strict turned on in the actual build process.
+{
+	"compilerOptions": {
+		"rootDir": "./",
+		"target": "es5",
+		"module": "commonjs",
+		"moduleResolution": "node",
+		"strict": true,
+		"noImplicitAny": false,
+		"useUnknownInCatchVariables": false
```
jamesaucode, line 11: any reason why we have these settings off? Will this make the IDE feedback inconsistent with the actual build script?

PR #9861, opened 2022-05-02
feat: rework error handler

dpilch:
#### Description of changes

Rework error handler to catch more errors as well as providing more useful information to the customer.

To do:
* Add additional error mappings
* Add information to docs on what errors mean
* Add unit testing

Create the scenario where the error should appear and check the error is in the console. For example, 

1. Create an App with API + DataStore
2. Add a model

```
type BrokenModel @model {
  id: ID!
  value: String
}
```
4. amplify push
5. Launch the app
6. Attempt to save.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/src/sync/processors/mutation.ts
+++ b/packages/datastore/src/sync/processors/mutation.ts
@@ -24,7 +24,7 @@ import {
 	PersistentModelConstructor,
 	SchemaModel,
 	TypeConstructorMap,
-	ErrorHandlerType,
+	ProcessName,
 } from '../../types';
 import { exhaustiveCheck, USER } from '../../util';
 import { MutationEventOutbox } from '../outbox';
@@ -33,9 +33,7 @@ import {
 	createMutationInstanceFromModelOperation,
 	getModelAuthModes,
 	TransformerMutationType,
-	TransformerMutationTypeToOperationName,
 	getTokenForCustomAuth,
-	OperationName,
 	ErrorMap,
 	mapErrorToType,
 } from '../utils';
@@ -376,68 +374,26 @@ class MutationProcessor {
 
 								throw new NonRetryableError('RetryMutation');
 							} else {
-								const newOperation: OperationName =
-									TransformerMutationTypeToOperationName(operation);
-								let errorHandlerResult: ErrorHandlerType;
 								try {
 									const errorMap = {
-										BadRecord: [/^Cannot return \w+ for [\w-_]+ type/],
-										ConfigError: [
-											// (error: Error) => { return true }
-										],
-
-										// will be defaulted within mapper function.
-										// no need to specify any matchers here.
-										Unknown: [],
+										BadRecord: error =>
+											/^Cannot return \w+ for [\w-_]+ type/.test(error.message),
 									} as ErrorMap;
-									errorHandlerResult = await this.errorHandler({
-										// modelInstanceCreator not necessary
-										// localModel: this.modelInstanceCreator(
-										// 	modelConstructor,
-										// 	variables.input
-										// ),
+									await this.errorHandler({
 										localModel: variables.input,
 										message: error.message,
-										operation: newOperation,
+										operation,
 										errorType: mapErrorToType(errorMap, error),
 										errorInfo: error.errorInfo,
-										process: 'mutate',
+										process: ProcessName.mutate,
 										cause: error,
 										remoteModel: error.data
 											? this.modelInstanceCreator(modelConstructor, error.data)
 											: null,
 									});
 								} catch (err) {
-									logger.warn('failed to execute errorHandler', err);
+									logger.warn('Mutation error handler failed with:', err);
 								} finally {
-									switch (errorHandlerResult) {
-										case 'ContinueSync':
-											logger.debug('error handled by errorHandler', error);
-											return error.data
-												? [
-														{ data: { [opName]: error.data } },
-														opName,
-														modelDefinition,
-												  ]
-												: [];
-											break;
-										case 'Retry':
-											logger.debug('error is being retried', error);
-											throw new Error('errorHandler retry');
-											break;
-										case 'StopSync':
-											logger.debug('errorHandler stopped sync', error);
-											throw new NonRetryableError('errorHandler stopped sync');
-											break;
-										default:
-											logger.error(
-												'Invalid errorHandler response: ',
-												errorHandlerResult
-											);
-											throw new NonRetryableError(
-												'Invalid errorHandler response'
-											);
-									}
 									// Return empty tuple, dequeues the mutation
 									return error.data
 										? [
--- a/packages/datastore/src/sync/processors/mutation.ts
+++ b/packages/datastore/src/sync/processors/mutation.ts
@@ -42,6 +42,7 @@ const MAX_ATTEMPTS = 10;
 
 const logger = new Logger('DataStore');
 
+// TODO: add additional error maps
 const errorMap = {
 	BadRecord: error => /^Cannot return \w+ for [\w-_]+ type/.test(error.message),
 } as ErrorMap;
@@ -70,8 +71,8 @@ class MutationProcessor {
 		private readonly MutationEvent: PersistentModelConstructor<MutationEvent>,
 		private readonly amplifyConfig: Record<string, any> = {},
 		private readonly authModeStrategy: AuthModeStrategy,
-		private readonly conflictHandler?: ConflictHandler,
-		private readonly errorHandler?: ErrorHandler
+		private readonly errorHandler: ErrorHandler,
+		private readonly conflictHandler?: ConflictHandler
 	) {
 		this.generateQueries();
 	}
--- a/packages/datastore/src/sync/processors/mutation.ts
+++ b/packages/datastore/src/sync/processors/mutation.ts
@@ -42,6 +42,10 @@ const MAX_ATTEMPTS = 10;
 
 const logger = new Logger('DataStore');
 
+const errorMap = {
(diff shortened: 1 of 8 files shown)


packages/datastore/src/sync/processors/sync.ts
```diff
@@ -223,15 +230,36 @@ class SyncProcessor {
...
 							result.data[opName].items = result.data[opName].items.filter(
 								item => item !== null
 							);
-
 							if (error.errors) {
+								const errorMap = {
+									BadRecord: [/^Cannot return \w+ for [\w-_]+ type/],
+									ConfigError: [],
+									Unknown: [],
+								} as ErrorMap;
```
svidgen: In my local, I had started moving these maps out of the loops and up to the top of the file with the intent on making these blocks shorter, quicker reads down the road. Thoughts?

packages/datastore/src/sync/processors/sync.ts
```diff
@@ -223,15 +229,35 @@ class SyncProcessor {
...
-
 							if (error.errors) {
+								const errorMap = {
+									BadRecord: error =>
+										/^Cannot return \w+ for [\w-_]+ type/.test(error.message),
+								} as ErrorMap;
+								await Promise.all(
+									error.error.map(async err => {
+										try {
+											await this.errorHandler({
```
svidgen, line 246:
```suggestion
											await this.errorHandler({
												recoverySuggestion: "Ensure app code is up to date, auth directives exist and are correct on each model, and that server-side data has not been invalidated by a schema change. If the problem persists, search for or create an issue: https://github.com/aws-amplify/amplify-js/issues",
```

packages/datastore/src/sync/processors/mutation.ts
```diff
@@ -372,21 +375,24 @@ class MutationProcessor {
 								throw new NonRetryableError('RetryMutation');
 							} else {
 								try {
+									const errorMap = {
+										BadRecord: error =>
+											/^Cannot return \w+ for [\w-_]+ type/.test(error.message),
+									} as ErrorMap;
 									await this.errorHandler({
```
svidgen, line 382:
```suggestion
											await this.errorHandler({
												recoverySuggestion: "Ensure app code is up to date, auth directives exist and are correct on each model, and that server-side data has not been invalidated by a schema change. If the problem persists, search for or create an issue: https://github.com/aws-amplify/amplify-js/issues",
```

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -470,6 +475,37 @@ class SubscriptionProcessor {
...
+																	error: { errors: [{ message = '' } = {}] } = {
+																		errors: [],
+																	},
+																} = givenError;
+																const regex = /Connection failed.+Unauthorized/;
+																return regex.test(message);
+															},
+														} as ErrorMap;
+														try {
+															await this.errorHandler({
```
svidgen, line 490:
```suggestion
											await this.errorHandler({
												recoverySuggestion: "Ensure app code is up to date, auth directives exist and are correct on each model, and that server-side data has not been invalidated by a schema change. If the problem persists, search for or create an issue: https://github.com/aws-amplify/amplify-js/issues",
```

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -470,6 +475,37 @@ class SubscriptionProcessor {
...
+															Unauthorized: (givenError: any) => {
+																const {
+																	error: { errors: [{ message = '' } = {}] } = {
+																		errors: [],
+																	},
+																} = givenError;
+																const regex = /Connection failed.+Unauthorized/;
+																return regex.test(message);
+															},
+														} as ErrorMap;
```
svidgen, line 488: For each file, might be easier to read down the line to move these error maps out of the block into the top of the file(s) and/or a utils.

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -470,6 +475,37 @@ class SubscriptionProcessor {
...
+														try {
```
svidgen, line 489: This is partly what was I deliberating on earlier. I think we need to move this out near the top of the `error` callback. Let's log everything we can that comes back from appsync to start with here. We can cut back if we find we're being too verbose.

packages/datastore/src/sync/processors/sync.ts
```diff
@@ -223,15 +229,35 @@ class SyncProcessor {
...
 							const result = error;
 							result.data[opName].items = result.data[opName].items.filter(
 								item => item !== null
 							);
-
 							if (error.errors) {
+								const errorMap = {
+									BadRecord: error =>
+										/^Cannot return \w+ for [\w-_]+ type/.test(error.message),
+								} as ErrorMap;
```
svidgen, line 242: Ditto. Probably move up/out.

packages/datastore/src/sync/utils.ts
```diff
@@ -40,6 +41,21 @@ enum GraphQLOperationType {
 	GET = 'query',
 }
 
+export type ErrorMap = {
+	[key in ErrorType]: (error: Error) => boolean;
+};
+
+export function mapErrorToType(errorMap: ErrorMap, error: Error): ErrorType {
```
svidgen, line 48:
```suggestion
/**
 * Categorizes an error with a broad error type, intended to make
 * customer error handling code simpler.
 * @param errorMap Error names and a list of patterns that indicate them (each pattern as a regex or function)
 * @param error The underying error to categorize.
 */
export function mapErrorToType(errorMap: ErrorMap, error: Error): ErrorType {
```

packages/datastore/src/types.ts
```diff
@@ -775,15 +775,31 @@ export type SyncConflict = {
...
 
-export type SyncError = {
+export type SyncError<T extends PersistentModel> = {
 	message: string;
-	errorType: string;
-	errorInfo: string;
-	localModel: PersistentModel;
-	remoteModel: PersistentModel;
+	errorType: ErrorType;
+	errorInfo?: string;
```
svidgen, line 781:
```suggestion
	errorInfo?: string;
	recoverySuggestion?: string;
```

svidgen (changes requested):

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -436,12 +453,31 @@ class SubscriptionProcessor {
...
+														model: modelDefinition.name,
+														operation: operation,
+														errorType: mapErrorToType(
+															errorMap,
+															subscriptionError
+														),
+														process: ProcessName.subscribe,
+														remoteModel: null,
+														cause: subscriptionError,
+													});
```
svidgen, line 477: Doh!

```
@aws-amplify/datastore: ERROR: src/sync/processors/subscription.ts:467:15 - Expected property shorthand in object literal ('{message}').
@aws-amplify/datastore: ERROR: src/sync/processors/subscription.ts:469:15 - Expected property shorthand in object literal ('{operation}').
```

```suggestion
													await this.errorHandler({
														recoverySuggestion:
															'Ensure app code is up to date, auth directives exist and are correct on each model, and that server-side data has not been invalidated by a schema change. If the problem persists, search for or create an issue: https://github.com/aws-amplify/amplify-js/issues',
														localModel: null,
														message,
														model: modelDefinition.name,
														operation,
														errorType: mapErrorToType(
															errorMap,
															subscriptionError
														),
														process: ProcessName.subscribe,
														remoteModel: null,
														cause: subscriptionError,
													});
```

svidgen (changes requested):

packages/datastore/src/sync/processors/mutation.ts
```diff
@@ -33,12 +34,18 @@ import {
...
 } from '../utils';
 
 const MAX_ATTEMPTS = 10;
 
 const logger = new Logger('DataStore');
 
+const errorMap = {
+	BadRecord: error => /^Cannot return \w+ for [\w-_]+ type/.test(error.message),
+} as ErrorMap;
+
```
svidgen, line 48: As-is, I think this PR gives Studio will have better errors to collect. But, we should make sure we have a TODO to expand this mapping and the similar mapping on the other sync engine processes. I think the missing auth directive errors should surface here too, for example.

Looking at this regex, I probably should have provided a sample in a comment of what this is supposed to match. I can circle back on that tomorrow if desired.

svidgen (approved):
Looking at all 3 error type maps together, I think we still have more error classification work to do. But, I think this is sufficient to improve visibility into errors that are currently just *silent*.

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -436,12 +453,31 @@ class SubscriptionProcessor {
...
 											},
-											error: subscriptionError => {
+											error: async subscriptionError => {
 												const {
 													error: { errors: [{ message = '' } = {}] } = {
 														errors: [],
 													},
 												} = subscriptionError;
+												try {
+													await this.errorHandler({
```
nickarocho, line 463: If the `errorHandler` is optional in the constructor, is it possible a call to `this.errorHandler` could result in an exception?

nickarocho (approved):
LGTM 🤙🏼 just left a clarifying question

packages/datastore/src/sync/processors/subscription.ts
svidgen: Yep. And, it is intended to be supplied by the user, so every time `errorHandler` is called, it should be done in a try-catch.

dpilch: I'm thinking we change `errorHandler` to required since it is always passed via SyncEngine.

iartemiev (approved):
LGTM 👍🏻

PR #9874, opened 2022-05-05
fix: add warning for when schema is not initialized

dpilch:
#### Description of changes

Add a warning when `schema` is not initialized with remediation recommendations.

See #9854 for additional details.

#### Issue #, if available

#9470

Follow steps in #9854 to get the application to a bad state. Then start the app and observe the error.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


svidgen:
I'm debating with myself over whether this should really be an error instead. My concern is that a developer might read past warnings if there are errors present. (As I often do.)

dpilch:
I'll change it to an error.

packages/datastore/__tests__/DataStore.ts
```diff
@@ -412,6 +412,22 @@ describe('DataStore tests', () => {
 		({ initSchema, DataStore } = require('../src/datastore/datastore'));
 	});
 
+	test('warning on schema not initialized on start', async () => {
```
iartemiev, line 415: nit: ~warning~ error on schema not initialized on start. Same thing for the other test

packages/datastore/src/datastore/datastore.ts
```diff
@@ -94,6 +94,7 @@ declare class Setting {
 const SETTING_SCHEMA_VERSION = 'schemaVersion';
 
 let schema: InternalSchema;
+let schemaInitialized = false;
```
iartemiev, line 97: Do we have to introduce a new variable for this as opposed to just checking whether `schema` is `undefined`?

dpilch: It is not necessary. I did this because I thought it made the intent clearer, but I can just use `schema` if you think that is better.

iartemiev: I think just checking if `schema === undefined` on line [235](https://github.com/aws-amplify/amplify-js/pull/9874/files#diff-10866b5ac7d469b38dd9928d9d7056a3a402904fef591ea8cfe773b4d75d63efR235) should be clear enough, especially since the error message that follows is nice and explicit

iartemiev (approved):
LGTM 👍🏻  Thanks Dane!

svidgen (approved):
Woops! Sorry &mdash; I thought I approved this with my comment already.

Approved!

Issue #9325, opened 2021-12-07
observeQuery improvement


maxludovicohofer:
### Is this related to a new or existing framework?

Angular, React, React Native, Vue, Web Components, Next.js, New framework

### Is this related to a new or existing API?

DataStore

### Describe the feature you'd like to request

observeQuery should apply the provided filter predicate on observe also, to check for subsequent results.
Also, it would be nice if observeQuery had paging support.

### Describe the solution you'd like

```typescript
// Should check if isValid is still true when query items change
DataStore.observeQuery(
    type,
    ({ isValid }) => isValid("eq", true),
    {
        // Should support page and limit
        page,
        limit,
        sort
    }
);
```

### Describe alternatives you've considered

Getting items through `observeQuery` first; when synced, observing the model with `observe`.
If there is a mutation, running `query` (with the same filter predicate) only on the mutated item, and removing it or adding it or changing it.

For paging, I use `query` and then `observe` the model. If there is a mutation, I rerun the `query` on the model.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

nickarocho:
Hi @maxludovicohofer, thanks for your question.

Observe query does support dynamic sorting, filtering, and limiting the amount of records received in snapshots, so I'd like to fully understand your individual use case better to provide some personalized recommendations with the current implementation of the `observeQuery` API.

Though I'm sure you've found our documentation on this API already, for those who aren't fully familiar - here is the [Observe Query documentation](https://docs.amplify.aws/lib/datastore/real-time/q/platform/js/#observe-query-results-in-real-time) which describes common use cases like sorting, filtering, and configuring limits for the sync page size.

---

You mentioned that `observeQuery`:
> Should check if isValid is still true when query items change

Could you please describe a little more about what `isValid` would be referring to in your app and what you mean by "when query items change"? Is `isValid` a field on a given model? How might the query items change?

__re: the `limit` (i.e. max records or `items` to sync within each returned snapshot)__

you can configure this by adjusting the `syncPageSize` via:
```js
DataStore.configure({
  syncPageSize: 100, // default is 1000
});
```

__re: pagination__

Could you also describe why you would need pagination while using `observeQuery`? The API behaves just  as `observe`, which also doesn't take a pagination argument, but if you could provide a little more insight into your specific use case we might be able to help you achieve what you are looking to do in your app.

[...]

PR #9879, opened 2022-05-06
fix(@aws-amplify/datastore): fixes observeQuery not removing newly-filtered items from snapshot

svidgen:
#### Description of changes

`observeQuery()` was not removing items from the snapshot if they later become invalid through an update. Those updates were being pre-filtered by an inner `observe`, leaving `observeQuery` with no signal to remove them from or emit a new snapshot.

#### Issue #, if available

#9325 #9573

New tests added that simulate two update-to-remove scenarios; one during generation of the initial snapshot(s), and another after sync is done.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/__tests__/DataStore.ts
```diff
@@ -294,6 +296,27 @@ describe('DataStore observeQuery, with fake-indexeddb and fake sync', () => {
...
+		//      ( cover your eyes )
+		//
+		// this prevents pollution between tests that may include observe() calls.
+		// This is a cheap solution let DataStore "settle" before clearing it and
+		// starting the next test. If we don't do this, we get "spooky"
+		// contamination between tests.
+		//
+		// NOTE: If you know of a better way to isolate these tests, please
+		// replace this pause!
+		//
```
manueliglesias, line 312: Can we avoid this naughtiness by using different models for each test?

It may make the test more verbose but will make them more isolated from each other

manueliglesias (changes requested):
The logic looks good to me, I would love to have more robust test even if that means defining a model per test to prevent cross talk

packages/datastore/__tests__/DataStore.ts
svidgen: I don't think that helps with the errors I was seeing. But, I'd be happy to pair on it if you like. As ugly and "naughty" as this feels, it *does* solve the problem, whereas separate models would be unscalable, IMO, and would not be expected to solve the issue I was seeing. (Internal DataStore/storage engine `undefined` refs -- I didn't note what they were specifically, but we can pair up and take a look!)

packages/datastore/src/datastore/datastore.ts
```diff
@@ -1205,41 +1217,83 @@ class DataStore {
...
+
 			(async () => {
 				try {
 					// first, query and return any locally-available records
 					(await this.query(model, criteria, sortOptions)).forEach(item =>
 						items.set(item.id, item)
 					);
 
-					// observe the model and send a stream of updates (debounced)
-					handle = this.observe(
```
david-mcafee, line 1216: According to the PR description: "observeQuery() was not removing items from the snapshot if they later become invalid through an update. Those updates were being pre-filtered by an inner observe, leaving observeQuery with no signal to remove them from or emit a new snapshot" - does this mean that `observe` is not filtering the items as we'd expect it to? i.e. is `observe` broken, or am I perhaps misunderstanding?

david-mcafee (commented):
Left a question re: `observe`

packages/datastore/src/datastore/datastore.ts
svidgen: Heh. It is confusing! The problem was that the criteria were being passed to `observe` and `observe` *was doing exactly what it's supposed to*.

The TLDR is that `observeQuery` was calling `observe` and passing the criteria through. So, if an item was in an initial result set, updated, and then *not* in the original set, `observe` would never tell `observeQuery` about it, because the updated item would not match `observe`'s give criteria.

In a little more detail, if it helps:

Old flow:

1. app calls `observeQuery(Model, m => m.x('eq', 'abc'))`
2. `observeQuery` calls `query(Model, m => m.x('eq', 'abc'))`
3. `observeQuery` calls `observe(Model, m => m.x('eq', 'abc'))`
4. app calls `DataStore.save(new Model({x: 'abc'}))`
5. `observe` sees it
6. `observe` checks to see if it matches (`m => m.x('eq', 'abc')`)
7. `observe` passes it to `observeQuery`
8. `observeQuery` adds it to the snapshot
9. app calls `DataStore.save(Model.copyOf(model, d => d.x = 'NOT abc!!!'))`
10. `observe` sees it
11. `observe` checks to see if it matches (`m => m.x('eq', 'abc')`)
12. **`observe` discards it**
13. `observeQuery` gets no update: **the snapshot now incorrectly still includes the outdated `{x: 'abc'}` item**

The new flow moves the criteria test.

1. app calls `observeQuery(Model, m => m.x('eq', 'abc'))`
2. `observeQuery` calls `query(Model, m => m.x('eq', 'abc'))`
3. `observeQuery` calls `observe(Model, Predicates.ALL)` (*effectively*)
4. app calls `DataStore.save(new Model({x: 'abc'}))`
5. `observe` sees it 
7. `observe` passes it to `observeQuery`
8. `observeQuery` checks to see if it matches `m => m.x('eq', 'abc')`
9. `observeQuery` adds it to the snapshot
10. app calls `DataStore.save(Model.copyOf(model, d => d.x = 'NOT abc!!!'))`
11. `observe` sees it
12. `observe` passes it to `observeQuery`
13. `observeQuery` checks to see if it matches `m => m.x('eq', 'abc')`
14. `observeQuery` **correctly removes the item from the snapshot**

packages/datastore/__tests__/DataStore.ts
undefobj: Would a middle ground be to rely on the stop/start events rather than using the pause? I always find that sleep solutions bite us long term because you get into a non-deterministic state, which reduces psychological confidence levels from Engineers later when they run tests. You're basically introducing a type of isolation with the pause->stop->start workflow as a "reset" of sorts, with one of them being ambiguous. Could you just remove that piece and have an event driven sequence to check the state of the system to move on?

packages/datastore/src/datastore/datastore.ts
undefobj: Reading this update what occurs to me is that the foundational data structure here which this change is managing bookkeeping is the "snapshot". Is this an in-memory array? While the above updated flow makes sense in the workflow of the app notifications to observers, it's a bit different from the Mutation queue in the outbox that `DataStore.save()` also writes to in this sequence. In the case of the outbox flow you have a CREATE followed by an UPDATE WITH CONDITION which means the original item will be written to a key and then the update will be appended to the end of the queue. This is to be expected as it gives us a causal chain of events. It's worth thinking about if we have two logic streams now - one publishing the event stream to the cloud, one publishing the subscriber stream to local observers - with differing logic and filtering rules, and if this will cause maintenance or logic race conditions in the future. Not enough to block the merge of this PR, but we should whiteboard this and look at the longer term ramifications.

packages/datastore/__tests__/DataStore.ts
svidgen: @undefobj Yeah, I was hoping the big banner would attract a fairly concrete "use this event" or "use this call pattern". And actually, as I look at this again, it also looks like I might have dropped a `stop()` at some point. That gives me a few things to poke at.

svidgen: Sure enough: I'm an idiot. I dropped a `stop()` from my `beforeEach`. Adding it back in appears to render the whole `afterEach` cleanup moot. It's been removed.

david-mcafee (approved):
LGTM! 🎉

manueliglesias (approved):
LGTM 👍

PR #9905, opened 2022-05-14
fix: docs for amazon-cognito-identity-js

mrspence:
#### Description of changes

Use-case 4 '_[Authenticating a user and establishing a user session with the Amazon Cognito Identity service](https://github.com/svidgen/amplify-js/blob/ae9a634c2def68bf4f001560d8a2ca7b31e08c7f/packages/amazon-cognito-identity-js/README.md)_' had invalid JS syntax for defining a dynamic object key. 

This commit adds two simple square brackets in the appropriate places!

Using corrected demo code in a private environment.

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/amazon-cognito-identity-js/README.md
+++ b/packages/amazon-cognito-identity-js/README.md
@@ -313,7 +313,7 @@ cognitoUser.authenticateUser(authenticationDetails, {
 			IdentityPoolId: '...', // your identity pool id here
 			Logins: {
 				// Change the key below according to the specific region your user pool is in.
-				'cognito-idp.<region>.amazonaws.com/<YOUR_USER_POOL_ID>': result
+				['cognito-idp.<region>.amazonaws.com/<YOUR_USER_POOL_ID>']: result
 					.getIdToken()
 					.getJwtToken(),
 			},


jamesgreen-0795:
Thanks for spotting this, I spent ages trying to debug the problem!

ashika01:
@mrspence the change you called out is not technically a dynamic object key we just want you to put your user pool id there. I can take this change in but you would have to define a variable with userpool id at the top so we can make this a dynamic object key

mrspence:
This PR is a simple change so the example 'works' using a dynamic key, as the doc insinuates. I assume the docs are not recommending values to be hardcoded in.

I agree that the whole example could use a touch-up, but it only _needs_ this 1 small change to make it work for developers who are following along! When I get time, I'll put forward a re-write example too, if this would help?

abdallahshaban557:
Hi @mrspence  - The example provided in our Wiki page is only providing a psuedo code example of how to use the `amazon-cognito-identity-js` package, and not meant to be a fully functioning example. We are going to be closing out this PR for now. In the future. Thank you for your contribution!

PR #9918, opened 2022-05-19
feat(datastore): add error maps for error handler

dpilch:
#### Description of changes

Add additional errors maps for error handler

#### Issue #, if available

N/A

Tested each scenario.

* Server errors will have a 5xx response
* Timeout
* Newly required field with outdated client will give error like `Variable 'field' has coerced Null value for NonNull type`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/src/sync/processors/errorMaps.ts
```diff
@@ -4,35 +4,59 @@ export type ErrorMap = Partial<{
...
 export const syncErrorMap: ErrorMap = {
+	BadModel: () => false,
 	BadRecord: error => /^Cannot return \w+ for [\w-_]+ type/.test(error.message),
 	ConfigError: () => false,
-	Transient: () => false,
+	Transient: error => connectionTimeout(error) || serverError(error),
 	Unauthorized: () => false,
 };
 
+function unwrapObserbableError(observableError: any) {
```
svidgen, line 50: Let's start adding good docstrings more obsessively, please. I want us to be able to see a helpful description of a method by hovering over its name wherever it's being called. I realize this one is pretty local _for now_. But, any time we add a function, class, etc.. -- *or find ourselves having to F12 to see what one's doing* -- let's get a good enough docstring in place to never have to hit F12 again to understand what's going on in the caller code.

svidgen (commented):
Left a comment about docstrings. Can we also please circle back on this with some tests?

packages/datastore/__tests__/mutation.test.ts
```diff
@@ -22,6 +23,7 @@ let modelInstanceCreator: any;
 let Model: PersistentModelConstructor<ModelType>;
 let PostCustomPK: PersistentModelConstructor<PostCustomPKType>;
 let PostCustomPKSort: PersistentModelConstructor<PostCustomPKSortType>;
+let axiosError;
```
svidgen, line 26: Non-blocking, but could use a comment here explaining how this is used / where this is thrown &mdash; or even just that fact that a Rest mock is throwing whatever error is present here.

svidgen (approved):

packages/datastore/src/sync/processors/errorMaps.ts
svidgen: Thanks!

svidgen:
Thanks for the tests! 🥳

david-mcafee (approved):
LGTM! 🎉

Issue #8969, opened 2021-09-30
Better typings


jordanmonier:
### Is this related to a new or existing API?

Authentication

### Describe the feature you'd like to request

The typings aren't correct, they only have to methods and they often returns `any`. Maybe we are missing something but in a Typescript environment this is very painful to deal with.

### Describe the solution you'd like

Includes all the property in the class declaration.

```ts
export class CognitoUserPool {
	constructor(
		data: ICognitoUserPoolData,
		wrapRefreshSessionCallback?: (
			target: NodeCallback.Any
		) => NodeCallback.Any
	);

        ...
       
        public advancedSecurityDataCollectionFlag: boolean
        public clientId: string
        public userPoolId: string

	...
}
```

### Describe alternatives you've considered

Declaring our own typings for `@aws-amplify/auth`

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

jordanmonier:
I just wrote some types if you want to use them : 
```ts
declare module 'amazon-cognito-identity-js' {
  enum AuthenticationFlowType {
    USER_SRP_AUTH = 'USER_SRP_AUTH',
    USER_PASSWORD_AUTH = 'USER_PASSWORD_AUTH',
    CUSTOM_AUTH = 'CUSTOM_AUTH',
  }

  interface CognitoUser {
    attributes: {
      email: string
      email_verified: true

      family_name: string
      given_name: string
      phone_number: string
      phone_number_verified: boolean

      sub: string
    }

    pool: CognitoUserPool

    authenticationFlowType: AuthenticationFlowType
    keyPrefix: string
    signInUserSession: CognitoUserSession
    userDataKey: 'CognitoIdentityServiceProvider'
    username: string
  }

  interface CognitoUserPool {
    advancedSecurityDataCollectionFlag: boolean
    clientId: string
    userPoolId: string
  }

  interface CognitoUserSession {
    clockDrift: number
    idToken: CognitoIdToken
    accessToken: CognitoAccessToken
    refreshToken: CognitoRefreshToken
  }

  interface CognitoIdToken {
    jwtToken: string
    payload: {
      sub: string
      email_verified: boolean
      iss: string
      phone_number_verified: boolean
      'cognito:username': string
      given_name: string
      origin_jti: string
      aud: string
      event_id: string
      token_use: string
      auth_time: number
      phone_number: string
      exp: number
      iat: number
[...]

PR #9935, opened 2022-05-25
refactor(amazon-cognito-identity-js): refactor issue 8969 better typings

Yoshiitaka:
#### Issue #https://github.com/aws-amplify/amplify-js/issues/8969, 
**Better type definitions for amazon-cognito-identity-js.**

#### Checklist

- [ ] PR description included


ShawnCockburn:
+1

chintannp:
Hey @Yoshiitaka ,
                         Thanks for your contribution to the AWS Amplify JS Library. We are actively reviewing your PR and will provide feedback soon if needed.

chintannp:
Hello @Yoshiitaka , Can you resolve the merge conflicts so that we can review the PR?

chintannp:
@Yoshiitaka ,
                     I won't be able fix the merge conflict as your forked branch is behind. Can you fix the merge conflicts so that we can review the PR and be able to merge it?

Yoshiitaka:
@chintannp
Thank you check this PR.
I updated the forked branch and fixed the conflicts part :)
Please Could you review this PR?

packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -366,31 +413,81 @@ declare module 'amazon-cognito-identity-js' {
...
+			aud: string;
+			event_id: string;
+			token_use: string;
+			auth_time: number;
+			phone_number: string;
+			exp: number;
+			iat: number;
+			family_name: string;
+			jti: string;
+			email: string;
```
elorzafe, line 466: A problem here is that this payload can modified ona userpool lambda trigger

packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -366,31 +413,81 @@ declare module 'amazon-cognito-identity-js' {
         public config: AWS.CognitoIdentityServiceProvider.Types.ClientConfiguration;
     }
     */
-	export class CognitoAccessToken {
-		payload: { [key: string]: any };
 
+	export interface CognitoAccessToken {
+		jwtToken: string;
+		payload: {
```
elorzafe, line 419: A the problem here is that `decodePayload` return type is `{ [id: string]: any }`

packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -84,7 +90,26 @@ declare module 'amazon-cognito-identity-js' {
...
+		attributes: {
+			email: string;
+			email_verified: boolean;
+			family_name: string;
+			given_name: string;
+			phone_number: string;
+			phone_number_verified: boolean;
+			sub: string;
+		};
+
```
elorzafe, line 103: those got lazily load, and after that those are available, the other problem is that a user can have more `attributes`

packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -353,9 +388,21 @@ declare module 'amazon-cognito-identity-js' {
...
+	}
+
+	export class CognitoUserSession implements CognitoUserSession {
 		constructor(data: ICognitoUserSessionData);
 
+		public clockDrift: number;
+		public idToken: CognitoIdToken;
+		public accessToken: CognitoAccessToken;
+		public refreshToken: CognitoRefreshToken;
+
```
elorzafe, line 405: why do you need to expose this values if there are already getters?

elorzafe (commented):
@Yoshiitaka thanks for this PR, 

I have some concerns with typing some information that is available for getters, Another big concern is that `attributes` and `token` payload can be configured on the UserPool and could break the type.

What do you think?

chintannp:
@Yoshiitaka, In order to get this change out quickly, Can you please review and address the feedback provided by @elorzafe ?

Yoshiitaka:
@elorzafe
Thank you for reviewing.
This PR was dealt with a long time ago, So I will check the points to be pointed out and corrections from now.

Yoshiitaka:
@chintannp
Sorry for the late confirmation and reply.
I will check the review pointed out from this and correct it.

packages/amazon-cognito-identity-js/index.d.ts
Yoshiitaka: I'm see... That is as you pointed out. 
there are already created the getter, 
there was no need to expose public values.
Thank you :)

packages/amazon-cognito-identity-js/index.d.ts
Yoshiitaka: hmm...
In other words, does this place have to be as follows, not a strict type?

`payload: { [key: string]: any };`

packages/amazon-cognito-identity-js/index.d.ts
Yoshiitaka: hmm..
In other words, does this place have to be as follows, not a strict type too?

`payload: { [key: string]: any };`

packages/amazon-cognito-identity-js/index.d.ts
Yoshiitaka: In that case, I began think that it is better not to create strict type information for Cognito User attributes.

chintannp:
@Yoshiitaka , Can you address the feedback provide by @elorzafe ?

Yoshiitaka:
@chintannp
I have already responded to @elorzafe 's  feedback comment, So Please could you check it?

packages/amazon-cognito-identity-js/index.d.ts
elorzafe: Correct

packages/amazon-cognito-identity-js/index.d.ts
elorzafe: Interesting, the interface could be more restricted. I need to verify if those claims are same across every userpool

packages/amazon-cognito-identity-js/index.d.ts
elorzafe: Do you want to revert that change?

packages/amazon-cognito-identity-js/index.d.ts
elorzafe: I agree, do you want to revert that change?

Yoshiitaka: Thank you for comment. yes, undo this changes.

packages/amazon-cognito-identity-js/index.d.ts
Yoshiitaka: yes, revert this changes too.

aminya (approved):

aminya:
When I use `Hub.listen` and get the `payload.data`, the found `CognitoUser` doesn't have any methods. That's because it is serialized. So, the properties should be still added. The getters are not enough.

Simone319:
Closing as the PR is stale.

PR #9999, opened 2022-06-16
fix: update axios

david-mcafee:
#### Description of changes

Axios update

#### Issue #, if available

None

Manual testing of API Rest, API GraphQL, DataStore, and Storage.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/storage/src/providers/axios-http-handler.ts
```diff
@@ -151,11 +154,11 @@ export class AxiosHttpHandler implements HttpHandler {
 			}
 		}
 		if (emitter) {
-			axiosRequest.onUploadProgress = function(event) {
+			axiosRequest.onUploadProgress = function (event) {
```
ashika01, line 157: this is failing the unit test. Weird.

packages/storage/src/providers/axios-http-handler.ts
```diff
@@ -62,8 +64,9 @@ const normalizeHeaders = (
 	}
 };
 
-export const reactNativeRequestTransformer: AxiosTransformer[] = [
-	function(data, headers) {
+// TODO:
```
ashika01, line 67: sorry what the todo here?

packages/storage/src/providers/axios-http-handler.ts
david-mcafee: Very weird!

packages/storage/src/providers/axios-http-handler.ts
```diff
@@ -48,7 +64,7 @@ function hasErrorResponse(error: any): error is ErrorWithResponse {
 }
 
 const normalizeHeaders = (
-	headers: Record<string, string>,
+	headers: AxiosRequestHeaders,
```
jamesaucode, line 67: I think this is fine, I had it as `Record<string, string>` but I don't think the code depends on the assumption the value is always of type `string`.

david-mcafee: Discussed all of the comments on the previous revision offline, this will work as expected.

svidgen (approved):

jamesaucode (approved):
LGTM, thanks for fixing this!!

ashika01 (approved):
LGTM

PR #10012, opened 2022-06-20
feat(@aws-amplify/auth): auto sign in after sign up or confirm sign u…

helgabalashova:
…p implementation

#### Description of changes

Changes add confirm sign up listener to initiate auto sign in once account is confirmed. It happens if autoSignIn is set to true in params for SignUp function. 
Since customers can choose either code, link, or autoConfirm, the solution use three different ways to auto sign in. autoConfirm implementation immediately calls on confirmSignUp function. Link implementation uses polling to sign user in since there is no explicit way to know if user clicked the link. Polling is done within 3 minutes (time verification link is valid) every 5 seconds. Code implementation is using hub listening event. Once user confirms account with code, it dispatches listening event. SignUp function is listening for this event to initiate SignIn process.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -44,6 +44,8 @@ import {
 	JS,
 	UniversalStorage,
 	urlSafeDecode,
+	HubCallback,
+	HubPayload,
 } from '@aws-amplify/core';
 import {
 	CookieStorage,
@@ -406,11 +408,12 @@ export class AuthClass {
 									this.onConfirmSignUp(authDetails, null, autoSignInPolling);
 								}, 5000);
 							} else {
-								Hub.listen('auth', ({ payload }) => {
+								const listenEvent = ({ payload }) => {
 									if (payload.event === 'confirmSignUp') {
-										this.onConfirmSignUp(authDetails, payload);
+										this.onConfirmSignUp(authDetails, listenEvent);
 									}
-								});
+								};
+								Hub.listen('auth', listenEvent);
 							}
 						}
 						resolve(data);
@@ -423,8 +426,8 @@ export class AuthClass {
 
 	private async onConfirmSignUp(
 		authDetails: AuthenticationDetails,
-		payload?,
-		autoSignInPolling?
+		listenEvent?: HubCallback,
+		autoSignInPolling?: NodeJS.Timer
 	) {
 		const user = this.createCognitoUser(authDetails.getUsername());
 		try {
@@ -438,8 +441,8 @@ export class AuthClass {
 							value,
 							`${authDetails.getUsername()} has signed in successfully`
 						);
-						if (payload) {
-							Hub.remove('auth', payload);
+						if (listenEvent) {
+							Hub.remove('auth', listenEvent);
 						}
 						if (autoSignInPolling) {
 							clearInterval(autoSignInPolling);
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -95,6 +95,8 @@ const dispatchAuthEvent = (event: string, data: any, message: string) => {
 // https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListDevices.html#API_ListDevices_RequestSyntax
 const MAX_DEVICES = 60;
 
+const MAX_AUTOSIGNIN_POLLING_TIME = 180000;
+
 /**
  * Provide authentication steps
  */
@@ -382,18 +384,34 @@ export class AuthClass {
 							`${username} has signed up successfully`
 						);
 						if (autoSignIn) {
-							Hub.listen('auth', ({ payload }) => {
-								if (payload.event === 'confirmSignUp') {
-									const authDetails = new AuthenticationDetails({
-										Username: username,
-										Password: password,
-										ValidationData: validationData,
-										ClientMetadata: clientMetadata,
-									});
-
-									this.onConfirmSignUp(authDetails, payload);
-								}
+							const authDetails = new AuthenticationDetails({
+								Username: username,
+								Password: password,
+								ValidationData: validationData,
+								ClientMetadata: clientMetadata,
 							});
+							if (data.userConfirmed) {
+								this.onConfirmSignUp(authDetails);
+							} else if (this._config.verificationMethod === 'link') {
+								const start = Date.now();
+								const autoSignInPolling = setInterval(() => {
+									if (Date.now() - start > MAX_AUTOSIGNIN_POLLING_TIME) {
+										clearInterval(autoSignInPolling);
+										dispatchAuthEvent(
+											'SignIn',
+											null,
+											'Sorry, you have timed out...'
+										);
+									}
+									this.onConfirmSignUp(authDetails, null, autoSignInPolling);
+								}, 5000);
+							} else {
+								Hub.listen('auth', ({ payload }) => {
+									if (payload.event === 'confirmSignUp') {
+										this.onConfirmSignUp(authDetails, payload);
+									}
+								});
+							}
 						}
 						resolve(data);
 					}
@@ -403,7 +421,11 @@ export class AuthClass {
 		});
 	}
 
-	private async onConfirmSignUp(authDetails: AuthenticationDetails, payload) {
+	private async onConfirmSignUp(
+		authDetails: AuthenticationDetails,
+		payload?,
+		autoSignInPolling?
+	) {
 		const user = this.createCognitoUser(authDetails.getUsername());
 		try {
 			await user.authenticateUser(
@@ -412,11 +434,16 @@ export class AuthClass {
 					user,
 					value => {
 						dispatchAuthEvent(
-							'AutoSignIn',
+							'SignIn',
 							value,
 							`${authDetails.getUsername()} has signed in successfully`
 						);
-						Hub.remove('auth', payload);
+						if (payload) {
+							Hub.remove('auth', payload);
+						}
+						if (autoSignInPolling) {
+							clearInterval(autoSignInPolling);
+						}
 					},
 					error => {
 						logger.error(error);
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -31,7 +31,6 @@ import {
 	FederatedSignInOptions,
 	AwsCognitoOAuthOpts,
 	ClientMetaData,
-	AutoSignInResponse,
 } from './types';
 
 import {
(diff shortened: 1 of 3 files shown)


packages/auth/src/Auth.ts
```diff
@@ -379,7 +406,60 @@ export class AuthClass {
...
+							let times = 0;
+							const user = this.createCognitoUser(username);
+							const authDetails = new AuthenticationDetails({
+								Username: username,
+								Password: password,
+								ValidationData: validationData,
+								ClientMetadata: clientMetadata,
+							});
+							let result = null;
+							const interval = setInterval(async () => {
```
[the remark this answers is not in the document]
helgabalashova: 1. it needs to be run multiple times because we don't know when user clicks verification link. It checks every 5 seconds if the link was clicked. It seems like there is no other way to check than trying to sing user in. Once sign in is successful, it stops doing it.
2. I am not aware of any other way. But I am open to suggestions.

packages/auth/src/Auth.ts
```diff
@@ -285,23 +286,37 @@ export class AuthClass {
...
+		let confirmSignUp = true;
+		let confirmSignUpOption = 'code';
 
 		if (params && typeof params === 'string') {
 			username = params;
 			password = restOfAttrs ? restOfAttrs[0] : null;
 			const email: string = restOfAttrs ? restOfAttrs[1] : null;
 			const phone_number: string = restOfAttrs ? restOfAttrs[2] : null;
 
+			if (restOfAttrs && restOfAttrs[3] && restOfAttrs[3] === 'true') {
```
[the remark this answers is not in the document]
helgabalashova: restOfAttributes is used when typeOfParams is a string. It is used to pass username, password and other things necessary to SignUp. It is not something that I came up with initially but in case of customer using restOfAttrs instead of params, he can pass values necessary for autoSignIn in there.

packages/auth/src/Auth.ts
```diff
@@ -285,23 +286,37 @@ export class AuthClass {
...
+				autoSignIn = true;
+			}
+
+			if (restOfAttrs && restOfAttrs[4] && restOfAttrs[4] === 'false') {
```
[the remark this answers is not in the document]
helgabalashova: it is defined into variable in the next line

packages/auth/src/Auth.ts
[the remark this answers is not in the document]
jamesaucode: So if I'm not mistaken they can either call `Auth.signUp({ ..., autoSignIn: true })` or `Auth.signUp(..., 'true')` ?

packages/auth/src/Auth.ts
```diff
@@ -379,7 +406,60 @@ export class AuthClass {
...
+										? resolve({
+												operationResult: 'success',
+												user: result,
+												userConfirmed: true,
+										  })
+										: resolve({
+												operationResult: 'success',
+												user: null,
+												userConfirmed: false,
+										  });
```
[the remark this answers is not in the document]
jamesaucode: `userConfirmed` can be further simplified to `userConfirmed: !!result` or `userConfirmed: Boolean(result)` 😄

packages/auth/src/types/Auth.ts
```diff
@@ -50,6 +54,9 @@ export interface AuthOptions {
 	identityPoolRegion?: string;
 	clientMetadata?: any;
 	endpoint?: string;
+	autoSignIn?: boolean;
+	confirmSignUp?: boolean;
+	confirmSignUpOption?: string;
```
jamesaucode, line 59: Can we narrow this down to a union of a few constant strings or an enum? At a glance the only options are `'link'` and `'code'`

jamesaucode (commented):
Great start 😄 I wonder if we should even have the `autoSignIn` option for `confirmSignUp` ? auto sign in with `confirmSignUp` depends on a successful `signUp` call, so if I have `signUp` with `autoSignIn` set to true I would expect the `confirmSignUp` to do the auto sign in. 

I'm also thinking about scenarios where the user is doing a `confirmSignUp` later without the password, but they passed in the `autoSignIn: true` option, it won't be able to `autoSignIn` for them because we don't have the password to do the `signIn` for them, how would we handle that case?

packages/auth/src/Auth.ts
```diff
@@ -345,6 +346,7 @@ export class AuthClass {
 					);
 				});
 			}
+			autoSignIn = params['autoSignIn'] ? params['autoSignIn'] : false;
```
nickarocho, line 349: Nit: for brevity, you could do:

```suggestion
			autoSignIn = params['autoSignIn'] || false;
```

packages/auth/src/Auth.ts
```diff
@@ -387,6 +404,30 @@ export class AuthClass {
 		});
 	}
 
+	private async onConfirmSignUp(authDetails?) {
```
jamesaucode, line 407: why is `authDetails` here optional? From the usage it seems like this is required, and should be of type `AuthenticationDetails`

packages/auth/src/Auth.ts
```diff
@@ -379,6 +381,21 @@ export class AuthClass {
 							data,
 							`${username} has signed up successfully`
 						);
+						if (autoSignIn) {
+							Hub.listen('auth', resp => {
+								let { payload } = resp;
```
nickarocho, line 386: Nit: you could destructure inline like the `AuthClass` does [here](https://github.com/aws-amplify/amplify-js/blob/5e8264963c19b50069e9acc2b692d8e6a1a0a201/packages/auth/src/Auth.ts#L122).
```suggestion
							Hub.listen('auth', ({ payload }) => {
```

packages/auth/src/types/Auth.ts
```diff
@@ -14,6 +14,7 @@
 import {
 	ICookieStorageData,
 	ICognitoStorage,
+	CognitoUser,
```
nickarocho, line 17: Where's this being used?

nickarocho (commented):
Great improvements, Olya! Left a couple non-blocking suggestions.

packages/auth/src/Auth.ts
```diff
@@ -387,6 +404,30 @@ export class AuthClass {
...
+	private async onConfirmSignUp(authDetails?) {
+		const user = this.createCognitoUser(authDetails.getUsername());
+		try {
+			await user.authenticateUser(
+				authDetails,
+				this.authCallbacks(
+					user,
+					value => {
+						dispatchAuthEvent(
+							'AutoSignIn',
```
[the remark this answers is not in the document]
nickarocho: Good point. Though I could see how providing _both_ events might actually be beneficial. Or maybe a sub-set of the `signIn` event?

packages/auth/src/Auth.ts
```diff
@@ -379,6 +383,36 @@ export class AuthClass {
...
+									if (Date.now() - start > MAX_AUTOSIGNIN_POLLING_TIME) {
+										clearInterval(autoSignInPolling);
+										dispatchAuthEvent(
+											'SignIn',
+											null,
+											'Sorry, you have timed out...'
+										);
+									}
+									this.onConfirmSignUp(authDetails, null, autoSignInPolling);
+								}, 5000);
```
nickarocho, line 407: Your PR description is still blank, so I don't really understand the intent behind the polling logic. Can you add this context to the PR description describing why this is necessary and how/why you chose `180000` as the `MAX_AUTOSIGNIN_POLLING_TIME` and `5000` as the interval?

packages/auth/src/Auth.ts
```diff
@@ -379,6 +383,36 @@ export class AuthClass {
...
+								this.onConfirmSignUp(authDetails);
+							} else if (this._config.verificationMethod === 'link') {
+								const start = Date.now();
+								const autoSignInPolling = setInterval(() => {
+									if (Date.now() - start > MAX_AUTOSIGNIN_POLLING_TIME) {
+										clearInterval(autoSignInPolling);
+										dispatchAuthEvent(
+											'SignIn',
+											null,
+											'Sorry, you have timed out...'
```
nickarocho, line 403: This message could probably be a bit more descriptive. Timed out of what? How do I fix it? What might have caused this? etc.

packages/auth/src/Auth.ts
```diff
@@ -387,6 +421,40 @@ export class AuthClass {
 		});
 	}
 
+	private async onConfirmSignUp(
+		authDetails: AuthenticationDetails,
+		payload?,
+		autoSignInPolling?
```
nickarocho, line 427: Would probably be good to add well-defined types to both of these params.

nickarocho (commented):
Nice additions!

My main bit of feedback is requesting a detailed PR description to provide context to those reviewing your PR who weren't involved in the high-level design of your feature.

packages/auth/src/Auth.ts
```diff
@@ -387,6 +421,40 @@ export class AuthClass {
...
+				this.authCallbacks(
+					user,
+					value => {
+						dispatchAuthEvent(
+							'SignIn',
+							value,
+							`${authDetails.getUsername()} has signed in successfully`
+						);
+						if (payload) {
+							Hub.remove('auth', payload);
```
jamesaucode, line 442: Is this actually clearing the listener from `Hub` ? Looking at the implementation of Hub, it requires the (reference of) the callback function in order to actually remove it from Hub.

packages/auth/src/Auth.ts
```diff
@@ -95,6 +96,8 @@ const dispatchAuthEvent = (event: string, data: any, message: string) => {
 // https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListDevices.html#API_ListDevices_RequestSyntax
 const MAX_DEVICES = 60;
 
+const MAX_AUTOSIGNIN_POLLING_TIME = 180000;
```
ashika01, line 99: Curious how did we arrive with this number?

packages/auth/src/Auth.ts
ashika01: +1. Something atleast along the lines of it is the autoSignIn that has timed out after so-so time

packages/auth/src/Auth.ts
```diff
@@ -429,6 +499,11 @@ export class AuthClass {
 					if (err) {
 						reject(err);
 					} else {
+						dispatchAuthEvent(
+							'confirmSignUp',
```
ashika01, line 503: So how does this hub event get printed in ur sample app?
1. `...` has signed in successfully
2. `...` has been confirmed successfully

it would help if u can add a screenshot to visualize

packages/auth/src/Auth.ts
nickarocho: +1, can you add this to your PR description? Or even better - as a block comment somewhere in this file describing the intent behind the polling action?

packages/auth/src/Auth.ts
```diff
@@ -379,6 +384,37 @@ export class AuthClass {
...
+						if (autoSignIn) {
+							const authDetails = new AuthenticationDetails({
+								Username: username,
+								Password: password,
+								ValidationData: validationData,
+								ClientMetadata: clientMetadata,
+							});
+							if (data.userConfirmed) {
+								this.onConfirmSignUp(authDetails);
+							} else if (this._config.verificationMethod === 'link') {
```
nickarocho, line 396: This might be an appropriate place to leave a detailed block-level comment describing the need for the `setInterval` and explaining the general purpose of the following code.

helgabalashova:
New PR: https://github.com/aws-amplify/amplify-js/pull/10064

PR #10037, opened 2022-06-28
chore: DataStore docstrings

svidgen:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/datastore/src/types.ts
```diff
@@ -35,14 +35,31 @@ export type SchemaNamespaces = Record<string, SchemaNamespace>;
 export type SchemaNamespace = UserSchema & {
 	name: string;
 };
+
 export type SchemaModels = Record<string, SchemaModel>;
+
+/**
+ * something
```
dpilch, line 42: Haha, it certainly is something. Dropping a comment here so it's not forgotten.

packages/datastore/src/util.ts
```diff
@@ -712,6 +869,12 @@ export class DeferredPromise {
 	}
 }
 
+/**
+ * Essentially a timeout for a callback.
+ *
+ * A callback is provided, which is called within `maxInterval` or when `resolve()`
+ * on the `DeferredCallbackResolver` is called &mdash; whichever happens first.
```
dpilch, line 876: Dash encoded for some reason.

dpilch (commented):
I've made it half way through. So far really good information!

packages/datastore/src/util.ts
svidgen: Yes. If you use markdown in docstrings, VSCode will mostly render it correctly on hover. I assume most docs site generators will do the same. Hence the entity encoding.

packages/datastore/src/types.ts
svidgen: Oh man ... I think I might just strike some of the less complete docstrings for this pass. I want to try to tack my fix onto this PR, contrary to what I claimed I would do earlier this week.

Docstring burnout.

packages/core/src/Util/JobContext.ts
```diff
@@ -0,0 +1,251 @@
+import Observable from 'zen-observable-ts';
+
+/**
+ * @private For internal Amplify use.
+ *
+ * A scope to collect background outstanding jobs, subscriptions, promises, and
```
svidgen, line 6: Note: Docstring is out of date, as it was speculative before TDDing. Don't read yet!

PR #10045, opened 2022-06-30
fix(auth): Missing type in configure function

micksatana:
#### Description of changes
Screen Shot 2565-06-30 at 23 35 40
Missing type when using `Auth.configure`

Screen Shot 2565-07-01 at 00 28 20

In VSCode

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


decebal (dismissed):

micksatana:
Hi guys, can we have one more approver here? It's a simple change, won't take long to review.

chintannp:
@micksatana, Thanks for your contribution and also updating the docs at the same time.

chintannp (dismissed):
LGTM!  🚀

chrisbonifacio:
Looks like we need to fix one of the tests to account for the added type

```js
● Test suite failed to run
@aws-amplify/auth:     TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
@aws-amplify/auth:     __tests__/auth-configure-test.ts:16:19 - error TS2345: Argument of type '{ userPoolId: string; userPoolWebClientId: string; region: string; identityPoolId: string; mandatorySignIn: boolean; storage: {}; }' is not assignable to parameter of type 'AuthOptions'.
@aws-amplify/auth:       Types of property 'storage' are incompatible.
@aws-amplify/auth:         Type '{}' is missing the following properties from type 'ICognitoStorage': setItem, getItem, removeItem, clear
@aws-amplify/auth:     16  		auth.configure(opts);
```

packages/auth/src/Auth.ts
```diff
@@ -139,7 +139,7 @@ export class AuthClass {
 		return 'Auth';
 	}
 
-	configure(config?) {
+	configure(config?: AuthOptions) {
```
elorzafe, line 142: This would be a breaking change unfortunately. `Auth.configure` can also use the content of `aws-exports.js` file. Maybe we could do something to overload this function? @jamesaucode what do you think?

micksatana: Hi @chrisbonifacio

`Type '{}' is missing the following properties from type 'ICognitoStorage'` because the test case use `{}` instead of just left it out when storage not exists. The test mentioning `throw error when storage is empty`, so it should throw and it does. When I ran test locally, it got
```
@aws-amplify/auth: PASS __tests__/auth-configure-test.ts
@aws-amplify/auth:   ● Console
@aws-amplify/auth:     console.error ../core/lib/Logger/ConsoleLogger.js:129
@aws-amplify/auth:       [ERROR] 40:02.742 AuthClass - The storage in the Auth config is not valid!
```
And
```
@aws-amplify/auth: Test Suites: 11 passed, 11 total
@aws-amplify/auth: Tests:       200 passed, 200 total
@aws-amplify/auth: Snapshots:   1 passed, 1 total
@aws-amplify/auth: Time:        14.282s
```

I'm not sure how the same test behave differently in CI though.

micksatana: I think the test case want to test the fail case when it's an empty object. But it conflicts with type testing as the `ICognitoStorage` type has all required properties. So to pass this test, either fix the test case to remove `storage: {}` out, or make all `ICognitoStorage` props optional. If the test case intent to test logic not type, I think we shouldn't test with empty object (without ts-ignore). It should test with `undefined` instead. Or keep the empty object `storage` but use `ts-ignore` because we intentionally make the type failed.

abdallahshaban557: Hi @micksatana - Just wanted to give you a heads up we are planning on pushing a bunch of changes to our Authentication category, and are planning on include changes in your PR as a part of that. Just wanted to reach out to make sure you know we did not abandon your contribution.

micksatana: @abdallahshaban557 acknowledged with thanks.

abdallahshaban557:
Hi @micksatana - thank you for this contribution! We are currently working on improving our TS types, including adding support for typing the configure function. You can check out our [RFC proposal here](https://github.com/aws-amplify/amplify-js/issues/11113)

Issue #9824, opened 2022-04-20
Improve developer experience for handling subscriptions, WebSocket connections, etc (GraphQL, PubSub, etc)


chrisbonifacio:
### Is this related to a new or existing API?

GraphQL API, PubSub

### Describe the feature you'd like to request

Many customers have reported several issues with GraphQL/PubSub subscriptions. Opening this feature request to consolidate the issues for simpler context searching and facilitate community discussion for workarounds and/or guidance from the team.

Related issues:

- [ ] https://github.com/aws-amplify/amplify-js/issues/3039 - PubSub reconnect (highest customer activity)
- [x] https://github.com/aws-amplify/amplify-js/issues/9818 - no callback triggered when WebSocket connection is closed
- [ ] https://github.com/aws-amplify/amplify-js/issues/9805 - `unsubscribe()` does not trigger `complete` callback on signout
- [ ] https://github.com/aws-amplify/amplify-js/issues/9749 - unable to reconnect after network interruption
- [ ] https://github.com/aws-amplify/amplify-js/issues/9185 - needing to patch `paho-mqtt` package
- [x] https://github.com/aws-amplify/amplify-js/issues/4733 - subscriptions sometimes fail, no way to detect
- [ ] https://github.com/aws-amplify/amplify-js/issues/4459 - how do I detect if WebSocket closed and reconnect?
- [x] https://github.com/aws-amplify/amplify-js/issues/6064 - feature request for `subscribe` callback (observer accepts `start` but not triggered on connection/start_ack?)
- [ ] https://github.com/aws-amplify/amplify-js/issues/7156 - force close existing WebSocket connections when calling `Amplify.configure`
- [x] https://github.com/aws-amplify/amplify-js/issues/2692 - how do I catch PubSub errors if MQTT connection is lost or broken?

### Describe the solution you'd like

- [ ] WebSocket reconnection would be handled automatically by our library.
- [ ] Expose the closed connections through the `error` callback (at the moment, we do see some network errors such as `timeout disconnect`).
- [ ] Expose whether unsubscribing was successful either by accepting callbacks for `unsubscribe` or through the observer's `complete` callback, etc.

### Describe alternatives you've considered

List of Workarounds or PRs:

API GraphQL: A PR was merged to expose network errors and address https://github.com/aws-amplify/amplify-js/issues/1448 but maybe we can provide guidance or an example on how to handle re-subscribing when this error occurs.

[...]

PR #10063, opened 2022-07-06
GH-9824 - PubSub Connection state tracking for AppSyncRealtime

stocaaro:
#### Description of changes

This change includes all behavior and testing to implement tracking connection state changes for AppSync/API subscriptions as Hub events for all connection conditions discussed.

This has been extended to cover missed keep alive messages, which demonstrates how a health check could be implemented in the future.

**Why such a big change?**
This change aims to implement both the internal state tracking and use AppSync Realtime as a fully implemented usecase with complete unit testing of both in a single step.

To help you review this change, I have:
- Identified the parts and pieces in the list below with contextual links.
- Listed the specific areas of contention I might expect.

Changes:
- Add `ConnectionStateMonitor` to track connection health (derivative from DataStoreConnectivity) ([diff](https://github.com/aws-amplify/amplify-js/pull/10063/files#diff-b91b95b97628107fe184ef346a246657bc6c7eb2643a144e7d08de842cff776e))
- Add tests to cover all states/transitions of `ConnectionStateMonitor`. ([diff](https://github.com/aws-amplify/amplify-js/pull/10063/files#diff-4a284c90a011ec2a14b985e9cd4f58a99b843cf3bf5d3fb655cf8446bc40cfe7))
- Augment `AWSAppSyncRealTimeProvider` with `ConnectionStateMonitor` dispatching `Hub` events on connection state changes. ([diff](https://github.com/aws-amplify/amplify-js/pull/10063/files#diff-0d2e918c494282a8a6184d7bb1919ab3c7cc7188836d75f477d79a56aa6b2479R99-R114))
- Use `Hub` connection states instead of timeouts in AppSyncRealtimeProvider testing (17.6s => 5.4s) ([diff](https://github.com/aws-amplify/amplify-js/pull/10063/files#diff-ad695112cfc27f3467761561eda5e3cd25bb8f3c725dca9b9ab49abf33548970))
- Add missed keep alive state tracking on a 65s delay. ([diff](https://github.com/aws-amplify/amplify-js/pull/10063/files#diff-0d2e918c494282a8a6184d7bb1919ab3c7cc7188836d75f477d79a56aa6b2479R480-R483))

Areas of interest/concern:
- Are the the `ConnectionState` names the right ones?
- Should `ConnectionState` be a string literal or an enum?
- Once accessed both `Reachability().networkMonitor()` and `this.connectionStateMonitor.connectionStateObservable` subscriptions are never disposed of as even when all subscriptions are closed, the listen should still be in place to track the "Disconnected" state. I couldn't come up with a case where these should be unsubscribed, so this code doesn't capture the subscriptions.

```ts
Hub.listen('api', (data: any) => {
	const { payload } = data;
	if (payload.event === CONNECTION_STATE_CHANGE) {
		const connectionState = payload.data.connectionState as ConnectionState;
		console.log(connectionState);
	}
});
```

What connection states will be emitted?:
```ts
export type ConnectionState =
	| 'Connected'
	| 'ConnectedPendingNetwork'
	| 'ConnectionDisrupted'
	| 'ConnectionDisruptedPendingNetwork'
	| 'Connecting'
	| 'ConnectedPendingDisconnect'
	| 'Disconnected'
	| 'ConnectedPendingKeepAlive';
```

#### Issue #, if available
#9824

- Unit tests added
- All state behavior manually tested in custom example app

#### Connection States Diagram
ConnectionStateDiagram

#### Example state transitions
Example app monitoring state transitions where the connection is formed and then a network disruption times the connection out. The red line is the connection failure where the rest of the events are from Hub.

ExampleStateTransitions

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


elorzafe:
Thanks @stocaaro this is great!

Is it possible to add here a screenshot or a written diagram, to understand more the transitions? Maybe that help clarify the names a little bit more.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -371,14 +395,21 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 			setTimeout(this._closeSocketIfRequired.bind(this), 1000);
 		} else {
 			logger.debug('closing WebSocket...');
-			if (this.keepAliveTimeoutId) clearTimeout(this.keepAliveTimeoutId);
+			this.connectionStateMonitor.disconnecting();
```
elorzafe, line 398: I am not sure about sending two events so close together, in that case if there are no more subscriptions I would send the `disconnecting` event but not just before calling `disconnected`

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -371,14 +395,21 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
+				clearTimeout(this.keepAliveAlertTimeoutId);
+			}
 			const tempSocket = this.awsRealTimeSocket;
 			// Cleaning callbacks to avoid race condition, socket still exists
 			tempSocket.onclose = null;
 			tempSocket.onerror = null;
 			tempSocket.close(1000);
 			this.awsRealTimeSocket = undefined;
 			this.socketStatus = SOCKET_STATUS.CLOSED;
+			this.connectionStateMonitor.disconnected();
```
elorzafe, line 412: Would make more sense `closed` instead of `disconnected`?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -432,17 +463,24 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 					subscriptionFailedCallback,
 				});
 			}
+			this.connectionStateMonitor.connectionEstablished();
 
 			// TODO: emit event on hub but it requires to store the id first
```
elorzafe, line 468: Maybe we can remove this comment!

elorzafe (commented):
I couldnt finish reviewing but I gave partial comments

Thank you so much for doing this @stocaaro 🌮 🥇

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
stocaaro: Good catch. In the current implementation, `closing` (formerly `disconnection` updates the intended state when the connection is closed on purpose. `closed` (formerly `disconnect`) only updates the connection state, so when it appears separate from `disconnecting` it signals a `ConnectionDisrupted` state instead of the final `Disconnected` state.

Without reconnection, this is a distinction without a difference, where when reconnection logic gets mixed into this, it will be common to see connection `ConnectionDisrupted` transition back to `Connecting` => `Connected`.

For this comment, I am recognizing that I want to move `closing` out of the buffer wait, which may meet your comment. If you do not think we should have both state changes here, then the fix would be to have a new call that can change both the intended and connection state to disconnected at the same time to avoid the extra event.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -30,19 +30,21 @@ import {
 import Cache from '@aws-amplify/cache';
 import Auth, { GRAPHQL_AUTH_MODE } from '@aws-amplify/auth';
 import { AbstractPubSubProvider } from '../PubSubProvider';
-import { CONTROL_MSG } from '../../index';
+import { CONNECTION_HEALTH_CHANGE, CONTROL_MSG } from '../../index';
```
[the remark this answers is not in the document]
stocaaro: Yeah. I was too lazy here.  Had used ConnectionState in the internal implementation of the ConnectionStateMonitor and was using ConnectionHealth as an alt name for the string literal in the public interface. This is bad and wrong and I've gone back to do some internal renaming to reclaim this name for use in the constant, external interface such that...

Watching for `CONNECTION_STATE_CHANGE` Hub events will receive `ConnectionState` typed values generated internally by the `ConnectionStateMonitor` interface.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/constants.ts
```diff
@@ -93,3 +93,8 @@ export const START_ACK_TIMEOUT = 15000;
  * Default Time in milleseconds to wait for GQL_CONNECTION_KEEP_ALIVE message
  */
 export const DEFAULT_KEEP_ALIVE_TIMEOUT = 5 * 60 * 1000;
+
+/**
+ * Default Time in milleseconds to alert for missed GQL_CONNECTION_KEEP_ALIVE message
+ */
+export const KEEP_ALIVE_ALERT_TIMEOUT = 65 * 1000;
```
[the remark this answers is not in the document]
stocaaro: Sure. Had considered this, but rejected using DEFAULT as this will not be overridden. Will change it as I suppose we could add an override in the future.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
stocaaro: Removed

packages/pubsub/__tests__/ConnectionStateMonitor.tests.ts
```diff
@@ -0,0 +1,225 @@
...
+		});
+
+		test('when the network recovers, keep alives become the concern until one is seen', () => {
+			monitor.openingConnection();
+			monitor.connectionEstablished();
+			reachabilityObserver?.next?.({ online: false });
+			monitor.keepAliveMissed();
+			monitor.keepAliveMissed();
+			reachabilityObserver?.next?.({ online: true });
+			monitor.keepAlive();
```
svidgen, line 196: *Just hanging this comment here because a good number of these monitor methods are gathered here ...*

This may also be too nit-picky for an internal interface. But, did you consider any alternative naming schemes for these methods? 😅  I'm a little distracted by the non-verb method naming. I feel like I should be seeing `set`, `record`, `emit`, or `push` or something here. (And throughout the implementation, of course.)

```js
monitor.recordConnectionOpening();
monitor.recordConnectionEstablished();
monitor.recordKeepAliveMissed();
monitor.recordKeepAliveMissed();
monitor.recordKeepAliveReceived();
```

I could also get behind a single verb that accepts a state event, which could even contain the newly intended connection/network/* field values.

```js
monitor.record(CONNECTION.OPENING);
monitor.record(CONNECTION.ESTABLISHED);
monitor.record(CONNECTION.KEEP_ALIVE_MISSED);
monitor.record(CONNECTION.KEEP_ALIVE_MISSED);
monitor.record(CONNECTION.KEEP_ALIVE_RECEIVED);
```

🤷 

@elorzafe does this contradict opinions you've given?

packages/pubsub/__tests__/helpers.ts
```diff
@@ -108,11 +154,40 @@ export class FakeWebSocketInterface {
...
+	}
+
+	async waitUntilConnectionStateIn(connectionStates: ConnectionState[]) {
+		return new Promise<void>((res, rej) => {
+			if (connectionStates.includes(this.currentConnectionState)) {
+				res(undefined);
+			}
+			res(this.waitForConnectionState(connectionStates));
+		});
+	}
```
svidgen, line 185: Even though this is test code, docstrings would be help here for tooltips when reading the tests to help me quickly distinguish between the behavior of these two.

Also, unless I'm misreading, these look like they could be helpful to customers *in some form or another*. I wonder if there's opportunity to generalize these into a lite design / addition to Hub...

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -0,0 +1,196 @@
...
+	 * Translate the ConnectionState structure into a specific ConnectionState string literal union
+	 */
+	private linkedConnectionStatesToConnectionState({
+		connectionState,
+		networkState,
+		intendedConnectionState,
+		keepAliveState,
+	}: LinkedConnectionStates): ConnectionState {
+		if (connectionState === 'connected' && networkState === 'disconnected')
+			return 'ConnectedPendingNetwork';
```
svidgen, line 167: Loosely related to a comment above, it may be helpful to us to put these values into an enum with docstrings:

```js
enum CONNECTION_STATES {
  /*
   * Indicates that the physical network has been disconnected. A socket
   * disconnect will follow shortly unless network connectivity is restored.
   */
   ConnectionPendingNetwork = "ConnectionPendingNetwork"
}
```

These will show as tooltips in VS Code.

Alternatively, and much more related to a prior comment, would it make sense to return a structure here that has the short string *and a description* that can be included in the hub events?

svidgen (commented):
Happy to chat offline about my comments. I don't think I have super strong opinions on any of my nits. Just a question or two I think should be answered mostly, and then a bunch of knee-jerk reactions!

packages/pubsub/src/utils/ConnectionStateMonitor.ts
ashika01: +1

ashika01:
> > Should ConnectionState be a string literal or an enum?
> 
> Unless we have bundle size concern for PubSub. The pros for a string union I see here is the string union is strictly on the Typescript level so it won't bloat the bundle size. But overall I'm in favor of `enum` as it _usually_ gives a better editor experience. (types `ConnectionState.` and see a list of all the possible connection state).

+1. Not just editor experience, removes a lot of mistyping issues for development and makes things more readable and less distracting.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -0,0 +1,196 @@
...
+
+	private updateConnectionState(
+		statusUpdates: Partial<LinkedConnectionStates>
+	) {
+		// Maintain the socket state
+		const newSocketStatus = { ...this._connectionState, ...statusUpdates };
+
+		this._connectionState = { ...newSocketStatus };
+
+		this._connectionStateObserver.next({ ...this._connectionState });
```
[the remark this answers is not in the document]
stocaaro: Yep. You're right. Updating.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -0,0 +1,196 @@
...
+		});
+	}
+
+	/**
+	 * Get the observable that allows us to monitor the connection state
+	 *
+	 * @returns {Observable<ConnectionState>} - The observable that emits ConnectionState updates
+	 */
+	public get connectionStateObservable(): Observable<ConnectionState> {
+		// Translate from connection states to ConnectionStates, then remove any duplicates
```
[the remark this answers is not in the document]
stocaaro: Good question. This object aggregates state changes to any of the network, connection, intendedConnection, keepAliveHealth states. Some states will change these independent states without changing the overall connection state, such as keep alive misses during a network outage, both of with translate to `ConnectedPendingNetwork` while the connection hasn't been disrupted.

In other cases, it might happen that a connection becomes disrupted in a way that causes two async events that both update the state with the same value.  I had originally had these deduplicated in the `updateConnectionState` function, but refactored deuplication to only happen in this one place.

packages/pubsub/__tests__/ConnectionStateMonitor.tests.ts
stocaaro: I really like this suggestion. Verbs are good. Going with `monitor.record(CONNECTION_CHANGE.CLOSED)`

packages/pubsub/src/utils/ConnectionStateMonitor.ts
stocaaro: Updated

packages/pubsub/__tests__/helpers.ts
stocaaro: Docstring added. I've got a revision of this extracting the Hub specific stuff. I think I would prefer to not expose it cross package until there's a specific need for that, but could be persuaded.  Either way, lets hold off for the refector.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -89,8 +91,27 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
+		// Monitor the connection state and pass changes along to Hub
+		this.connectionStateMonitor.connectionStateObservable.subscribe(
+			ConnectionState => {
+				dispatchApiEvent(
+					CONNECTION_STATE_CHANGE,
+					{
+						provider: this,
+						connectionState: ConnectionState,
+					},
+					`Connection state is ${ConnectionState}`
```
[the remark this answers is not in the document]
stocaaro: Parts of this discussion happened in the Design/API documents/reviews. I haven't prepared document updates related to this change yet. For the purpose of this conversation here are 3 use-cases customers have discussed.

**Connection Status**
Lets assume a customer wants to display an icon to their users reflecting the connection health, they might do so like this:
```ts
const brokenStates = [ConnectionState.ConnectionDisrupted, ConnectionState.ConnectionDisruptedPendingNetwork, ConnectionState.Disconnected]
const unhealthyStates = [ConnectionState.ConnectedPendingDisconnect, ConnectionState.ConnectedPendingKeepAlive, ConnectionState.ConnectedPendingNetwork]

...

Hub.listen('api', (data: any) => {
	const { payload } = data;
	if (payload.event === CONNECTION_STATE_CHANGE) {
		const connectionState = payload.data.connectionState as ConnectionState;
		if (brokenStates.includes(connectionState)) {
			setConnectionHealth('broken');
		} else if (unhealthyStates.includes(connectionState)) {
			setConnectionHealth('unhealthy');
		} else {
			setConnectionHealth('healthy');
		}
	}
});
```
**Run an action whenever the connection becomes connected**
1. Emit user active status
2. Get most recent content (ensure all content is current)

In some cases this customer may track the prior state and only action on specific state transitions.

```ts
Hub.listen('api', (data: any) => {
	const { payload } = data;
	if (payload.event === CONNECTION_STATE_CHANGE) {
		const connectionState = payload.data.connectionState as ConnectionState;
		if (connectionState === ConnectionState.Connected) {
			// Take action
                }
	}
});
```

**Self managed reconnection**
Since reconnection will follow this, I wouldn't recommend we robustly document this use-case.

packages/pubsub/__tests__/helpers.ts
stocaaro: Like that thought. There's work out there to get better typing for Hub that might make a good opportunity to look at adding some helper infrastructure/interfaces around Hub.

elorzafe:
@stocaaro 

The diagram is great!! We could implement AppSync real time using state machine and following the same diagram!

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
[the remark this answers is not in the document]
elorzafe: Maybe we could add a helper function that converts the enum into health status so the customer doesnt need to implement that logic?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -89,8 +95,27 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
 	private keepAliveTimeout = DEFAULT_KEEP_ALIVE_TIMEOUT;
+	private keepAliveAlertTimeoutId?: ReturnType<typeof setTimeout>;
 	private subscriptionObserverMap: Map<string, ObserverQuery> = new Map();
 	private promiseArray: Array<{ res: Function; rej: Function }> = [];
+	private readonly connectionStateMonitor = new ConnectionStateMonitor();
+
+	constructor(options: ProviderOptions = {}) {
+		super(options);
+		// Monitor the connection state and pass changes along to Hub
+		this.connectionStateMonitor.connectionStateObservable.subscribe(
```
elorzafe, line 106: could be an issue that never unsubscribes?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -366,19 +394,28 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 			this.socketStatus = SOCKET_STATUS.CLOSED;
 			return;
 		}
+
+		this.connectionStateMonitor.record(CONNECTION_CHANGE.CLOSING);
```
elorzafe, line 398: This might be recorded multiple times if the buffer is not empty

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -0,0 +1,165 @@
...
+
+		this._linkedConnectionState = { ...newSocketStatus };
+
+		this._linkedConnectionStateObserver.next(this._linkedConnectionState);
+	}
+
+	/*
+	 * Translate the ConnectionState structure into a specific ConnectionState string literal union
+	 */
+	private linkedConnectionStatesToConnectionState({
```
elorzafe, line 129: I think we should use xstate, it would be so much better to visualize and maintain when we implement the provider using that machine. I dont see as a blocker but maybe if is worth the investment.

elorzafe (commented):
@stocaaro pr looks great, I would spend a little time to try xstate. It could be so much easy to understand the state. You already have the diagram ready, is just convert it to xstate. Let me know what you think

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
stocaaro: Agreed. This should be fine as the resulting observable is deduplicated. Made it intentionally to avoid needing to protect against duplicate protection complicating usage.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
stocaaro: Good catch/call. I originally had a subscription instance var, but this object is never disposed of / cleaned up, so when I went about testing that the subscription was cleaned up, I was considering adding a surface to allow this when I couldn't come up with a realworld usecase.

Even when all subscriptions are removed and the connection is closed, the Disconnected state is still relevant and there's nothing stopping this same provider from being used again, forming a new connection, right?

stocaaro:
@elorzafe Responding to comments here that didn't have the reply option

> Maybe we could add a helper function that converts the enum into health status so the customer doesn't need to implement that logic?

In an upcoming reconnect PR, I have a private variable that retains this for our use managing reconnection. We could make that public/add it to the provider interface, which would help in the MQTT use-case, but for AppSync, the provider isn't easy for the developer to access, right?  Are you thinking that we would plumb that out into the API or datastore surface somehow? I'm not clear on what developer experience you're proposing, would appreciate a bit more detail to make sure I'm thinking in the right direction.

>  I would spend a little time to try xstate. It could be so much easy to understand the state. You already have the diagram ready, is just convert it to xstate. Let me know what you think

I had looked briefly at state machines to model this, but they seemed like a poor fit / something that would add fragility and complexity. The named states are more of a developer experience normalization of the underlying states. From an internal implementation standpoint, it seemed more intuitive to manage the different state types independently and translate them together into a named state as late as possible in the process.

Are there good examples of where we've used xstate to accomplish things like this elsewhere in our code that I might model this after?  On its face, as a single layer of state, I imagine when the connection is lost, we might need to condition on the network condition to pick the right state transition to match the behavior (or capture this in a layer of abstraction). Either way, moving this logic to the head of the state transition process seems more complicated than leaving denormalizing the states to a single name late in the process as it is now.

I think @svidgen made a run at migrating this to xstate while reviewing it earlier and may have other thoughts in response.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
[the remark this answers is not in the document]
stocaaro: Yeah. I could see adding helper logic like this, but would prefer to get a round of customer feedback first to validate this is a need/requirement.

One thought that would make this a now vs later type thing might be to replace the enum with objects and add health state to those objects.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
stocaaro: I've made a run at this and it seemed like a poor fit. Do you have an xstate example or the time to put together a playground example that models out what your thinking? Interested to know if I've missed something here.

svidgen:
> I think @svidgen made a run at migrating this to xstate while reviewing it earlier and may have other thoughts in response.

It wasn't me... At least that I remember! 😅

stocaaro:
I've made a [run at rewriting the ConnectionStateMonitor behavior in xState](https://gist.github.com/stocaaro/69274ab0d7e4e83f34606f2f85487527).

Benefits of this approach:
- It's shorter (although the prototype is missing types)
- Each transition is very clear

This code was written to be a bit flexible/permissive such that odd or out of order state transitions work just fine, which I have attempted to capture in this state machine. 

I considered modeling this is one parent machine that contains 4 child machines as this code captures the logic, but couldn't find any examples about how this would work and doing this the parent machine loses visibility into the behavior of the child machines, so I dropped it.

This prototype probably has unintended bugs for my not being able to think through all the cases clearly enough. If we where to go this direction, I think I would try to generate the state transitions from the existing code. In the existing code, there are cases of hidden transition (such as missed keep alive messages) where `ConnectedPendingNetwork` can transition to either `Connected` or `ConnectedPendingKeepAlive` depending on whether a KA message has been missed. I rejected making a special `ConnectedPendingNetoworkAndKeepAlive` type state to represent this difference to the customer.

My conclusion is that this code models 4 very simple states that are then being denormalized into a single meta-state. Modeling this as a single state-machine adds complexity and room for bugs that are avoided in the current implementation. Modeling the problem as many very small state machines could be done, but adds complexity while serving no apparent value.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -0,0 +1,165 @@
...
+
+import { Reachability } from '@aws-amplify/core';
+import Observable, { ZenObservable } from 'zen-observable-ts';
+import { ConnectionState } from '../index';
+
+// Internal types for tracking different connection states
+type LinkedConnectionState = 'connected' | 'disconnected';
+type LinkedHealthState = 'healthy' | 'unhealthy';
+type LinkedConnectionStates = {
+	networkState: LinkedConnectionState;
```
elorzafe, line 22: Why do we need `networkState`, the problem with this is that `connected` can have false positives. I think the most important is `keepAliveState` because is testing the connection. Do you think this is very important to have?

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -0,0 +1,165 @@
...
+	keepAliveState: LinkedHealthState;
+};
+
+export const CONNECTION_CHANGE: {
+	[key in
+		| 'KEEP_ALIVE_MISSED'
+		| 'KEEP_ALIVE'
+		| 'CONNECTION_ESTABLISHED'
+		| 'CONNECTION_FAILED'
+		| 'CLOSING'
```
elorzafe, line 34: Why not `CLOSING_CONNECTION` just to be consistent with `OPENING_CONNECTION` or viceversa

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -0,0 +1,165 @@
...
+		| 'OPENING_CONNECTION'
+		| 'CLOSED'
+		| 'ONLINE'
+		| 'OFFLINE']: Partial<LinkedConnectionStates>;
+} = {
+	KEEP_ALIVE_MISSED: { keepAliveState: 'unhealthy' },
+	KEEP_ALIVE: { keepAliveState: 'healthy' },
```
elorzafe, line 41: Would this also mean `connectionState` is `connected`?

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -0,0 +1,165 @@
...
+	private _linkedConnectionState: LinkedConnectionStates;
+	private _linkedConnectionStateObservable: Observable<LinkedConnectionStates>;
+	private _linkedConnectionStateObserver: ZenObservable.SubscriptionObserver<LinkedConnectionStates>;
+
+	constructor() {
+		this._linkedConnectionState = {
+			networkState: 'connected',
+			connectionState: 'disconnected',
+			intendedConnectionState: 'disconnected',
+			keepAliveState: 'healthy',
```
elorzafe, line 70: I would prefer to start as everything is `disconnected` and `unhealthy` unless I can prove otherwise

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -0,0 +1,165 @@
...
+				connectionStateObserver.next(this._linkedConnectionState);
+				this._linkedConnectionStateObserver = connectionStateObserver;
+			});
+
+		// Maintain the network state based on the reachability monitor
+		new Reachability().networkMonitor().subscribe(({ online }) => {
+			this.record(
+				online ? CONNECTION_CHANGE.ONLINE : CONNECTION_CHANGE.OFFLINE
+			);
+		});
```
elorzafe, line 84: I would initialize this after someone is subscribed (that means you might need to count the subscribers)

I usually see this for Observable pattern so this can be cleaned up and is not adding stuff in memory that no one is using

What do you think?

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -0,0 +1,165 @@
...
+			});
+	}
+
+	/*
+	 * Updates local connection state and emits the full state to the observer.
+	 */
+	record(statusUpdates: Partial<LinkedConnectionStates>) {
+		// Maintain the socket state
+		const newSocketStatus = {
+			...this._linkedConnectionState,
```
elorzafe, line 117: Cool, so is overwritting only the changes 👍. Can happen you end up on some inconsistent state? Maybe it would be nice to define an invariant that could be type safe? What do you think?

For example `networkState: 'disconnected', keepAliveState: 'healthy'` shouldn't be possible

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -0,0 +1,165 @@
...
+			return ConnectionState.ConnectionDisruptedPendingNetwork;
+
+		if (
+			connectionState === 'disconnected' &&
+			intendedConnectionState === 'connected'
+		)
+			return ConnectionState.ConnectionDisrupted;
+
+		if (connectionState === 'connected' && keepAliveState === 'unhealthy')
+			return ConnectionState.ConnectedPendingKeepAlive;
```
elorzafe, line 158: Would it read more easily if instead of the flat `if` structure make it nested? It could read like a decision tree instead.

elorzafe (changes requested):
Thanks @stocaaro 🥇 

I think this is very close, my only blocking comment would be initializing the observables on the constructor, the others I leave it to you.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
stocaaro: These states are independent, so I would expect `networkState: 'disconnected', keepAliveState: 'healthy'` to be quite common where keepAlive will then timeout to unhealthy after 65 seconds, but because ka is ignored for disconnected states it won't add an event.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
stocaaro: I started down that path initially and refactored it into this for readability, aligning with the school of thought that flat conditionals / guard clauses are easier to read and reason about than nested conditionals. If there's a contrary philosophy at play here, I'd be happy to refactor this to fall in line with a different pattern. Do you have an article or code sample I can use to ensure I'm aligned with what you're going for?

packages/pubsub/src/utils/ConnectionStateMonitor.ts
stocaaro: Not necessarily. If we received a ka message while the connection was PendingDisconnect or Disconnected, that could hypothetically be async callbacks catching up, so flipping the connection state on a ka or a missed ka might cause bugginess (I think). My bias has been to try to keep state updates as independent as possible to avoid odd use-cases.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
stocaaro: I like this a lot. I'm going to see what I can do to adopt it.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
stocaaro: I think you're right from a cleanliness standpoint. In code, creating and disposing of this with connectivity adds a layer of interaction that could easily be buggy/error prone. I'll give it a 1-2 hour go today and update with further thoughts.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
stocaaro: Sure. That makes sense. Will make the update.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
stocaaro: Network state is very important to reconnection as it is usually the signal of when we should attempt to reconnect.

Eg. If the network is down and we fail ka after 60 seconds, then close the connection at 5 minutes, neither of these mark when we should attempt reconnect. Only once the network recovers is it valuable to attempt to reconnect again.

This pattern has been adopted from directly from the datastore implementation and has some decent production attletesting in that code.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
stocaaro: Running into two problems.
1. I'd like to start the networkState as whatever `Reachability` would tell me the current network state is. `Reachability` only fires when the state changes. For web, it is reading the value out of the global object, which I can expose.  RN doesn't have this same readable value. For RN we could setup some sort of global observer to track this state, but that seems worse than assuming the app is online at the time it is loaded.
2. Since KeepAlive arrives after a connection is formed, defaulting KA to unhealthy means that we get an extra `PendingKeepAlive` event after each new `Connection`. Since `PendingKeepAlive` is really only valuable after a 60 second timeout, I would not recommend injecting all these extra events or adding complexity to suppress them.

I think either of these could be worked around by adding complexity / fragility and in the case of reachability adding surface to core, but I don't think there is much value gained in being rigidly correct in these cases.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
stocaaro: I think I found a pattern of starting network monitoring with the first subscription and disabling it after the connection closes that works and I can live with. Interested to get your thoughts.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
elorzafe: I would check the lifecycle of this when using API graphql without subscriptions just to check if this constructor gets called (I think it does).

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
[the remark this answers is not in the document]
elorzafe: I am good with that

packages/pubsub/src/utils/ConnectionStateMonitor.ts
elorzafe: That makes sense, thanks!

packages/pubsub/src/utils/ConnectionStateMonitor.ts
elorzafe: Makes sense, thanks for clarifying!

packages/pubsub/src/utils/ConnectionStateMonitor.ts
elorzafe: If is that problematic I am fine with this, we can revisit this logic in the future.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
elorzafe: Makes sense, thanks for clarifying

packages/pubsub/src/utils/ConnectionStateMonitor.ts
elorzafe: Is more like a personal preference tbh, when I see the same statement multiple times I tend to only have it once. Could also be more readable if is spread into smaller functions. would the function name be more readable if is `connectionStatesReducer`or `connectionStatesTranslator` what do you think?

stocaaro: Name changed.

I get that. We're all aiming for code that looks like code we've had a good experience maintaining before.

I just did 2 more short experiments.
1. Building a translation object to wrap the connectedStates. My translation methods ended up with the same names as my `ConnectedState` enum, which made me think of a map.
2. Since this is order of operation guard clauses, a map is the wrong structure and its more of an array of tuples. I started down the path of code that looks like the following, however, I think this trades clarity for brevity, which seems like the wrong direction.
```ts
const CONNECT_STATE_TRANSLATION_MAPPINGS = [
	[
		ConnectionState.ConnectedPendingNetwork,
		{ connectionState: 'connected', networkState: 'disconnected' },
	],
        ...
];
```

Based on this further exploration, I've left it the way it was again.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
elorzafe: Yes! I totally good with doing that

jamesaucode:
> > I think @svidgen made a run at migrating this to xstate while reviewing it earlier and may have other thoughts in response.
> 
> It wasn't me... At least that I remember! 😅

That was me ... 😄

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -144,6 +144,8 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 	): Observable<any> {
 		const appSyncGraphqlEndpoint = options?.appSyncGraphqlEndpoint;
 
+		this.connectionStateMonitor.enableNetworkMonitoring();
+
```
elorzafe, line 148: I dont like to have this logic outside of the `connectionStateMonitor` I would do it inside of the observable

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -75,13 +77,30 @@ export class ConnectionStateMonitor {
 				connectionStateObserver.next(this._linkedConnectionState);
 				this._linkedConnectionStateObserver = connectionStateObserver;
 			});
+	}
 
+	/**
+	 * Turn network state monitoring on
+	 */
+	public enableNetworkMonitoring() {
```
elorzafe, line 85: I would have this logic inside of `_linkedConnectionStateObservable` to initialize the network monitor and on the `return` I would  unsubscribe the network monitor if there is no more subscribers.

elorzafe (changes requested):
@stocaaro I think the logic of enableNetworkMonitor should be inside of the observable. We can chat more offline if you want.

svidgen (approved):

elorzafe (approved):
Great work @stocaaro 🚢 🌮 🏅 🎉

Issue #6320, opened 2020-07-14
Sign In with Auth.confirmSignUp


mousedownmike:
**Is your feature request related to a problem? Please describe.**
This feature is related to closed issue #2562, that request was to support signing in a user with the Auth class when a successful `Auth.confirmSignUp` request was made.  The ticket was closed based on a UI implementation that does not solve the issue for users of the Auth class directly.

**Describe the solution you'd like**
The Auth library should have the option to automatically sign-in a user when they successfully complete the `Auth.confirmSignUp` request.  This should be made possible without requiring the calling application to persist the user's credentials (username and password) between requests.

The following would be my ideal use case:

1. A new site user is successfully registered with `Auth.signUp`.  The site/app **does NOT persist the user's password** so it is no longer known by the local client.  
1. The site presents the code delivery details (e.g. SMS, Email...) to the new user so they know where to find their confirmation code. 
1. The site presents a form with a field for the username/email (pre-populated if the user has not left the flow) and an empty field for the confirmation code.
1. The user retrieves the confirmation code from the indicated delivery method.
1. The user enters the valid confirmation code (and username if not pre-populated) and submits the form.
1. The site successfully performs `Auth.confirmSignUp` with the supplied username and code.
1. The Auth library validates the username and code and upon receiving a valid response is able to generate a fully realized/signed-in CognitoUser locally.  No further calls to `Auth.signIn` are required.

This flow is essentially using the confirmation code as a one-time password which seems reasonable. 

**Describe alternatives you've considered**
Multiple alternatives are discussed in #2562 and the React UI implementation is one alternative although it appears to require persisting the user's password locally between requests which seems like a potential security risk.

**Additional context**
My assumption based on reading #2562 was that this would require some changes to the Cognito service.  I imagine that is the case so that `confirmSignUp` returns a valid token but I can't find any tickets that represent that change request nor do I even know where to look for such a request.

jordanranz:
Hey @mousedownmike,

To clarify on the closing of the other issue, this is now implemented in all framework UI implementations of Authenticator and not just React. 

Alternatively, if you would like this functionality using only the Auth category, you would just call Auth.signIn on a successful `Auth.confirmSignUp`. That being said, maybe the functionality could be baked into the `Auth.confirmSignUp`, is that what you are asking for?

mousedownmike:
Thank you for the response @jordanranz. I've updated the original post to remove mention of other UI frameworks and focus just on the use of the `Auth` class.  

[...]

Issue #6018, opened 2020-05-01
Flow of the user login


BernhardSmuts:
Authentication

I find it extremely difficult to add a custom user auth flow "out the box" without heavy config.

Current:
-User enters details on auto created UI (Username, Email, password, phone number)
-User gets email
-User needs to enter email code into other waiting page (two pages open, what if user closed the first page)
-User then needs to re-enter login details (on the first page)
-Still has email page open

Ideal:
-User enters only details I need (email, password, confirm password)
-User gets email with verification link
-User clicks on link, is auto signed in and starts using app...

Why is this not the way it's done when this is how 90% of sites do it?

elorzafe:
@BernhardSmuts 

Today we released `autoSignIn` feature on Amplify JS. We will be updating the docs tomorrow, but the feature is available now on version `aws-amplify@4.3.29`

If you are using code verification for new users you only have to do.

```javascript
 Auth.signUp({
      username: 'xxxxxx',
      password: '*********,
      attributes: {
        email: 'xxxxxxxxxx'
      },
      autoSignIn: {
        enabled: true
      }
    })
````

The library will login in the user automatically after the user is confirmed on the App when doing `Auth.confirmSignUp`


If you are using link verification the library also handle that case but you need to configure the library is using `link` verification on `Amplify.configure` like this

```javascript
Amplify.configure({
    Auth: {
       // OPTIONAL
       signUpVerificationMethod: 'link'

        // REQUIRED only for Federated Authentication - Amazon Cognito Identity Pool ID
        identityPoolId: 'XX-XXXX-X:XXXXXXXX-XXXX-1234-abcd-1234567890ab',
        
        // REQUIRED - Amazon Cognito Region
        region: 'XX-XXXX-X',

        // OPTIONAL - Amazon Cognito Federated Identity Pool Region 
        // Required only if it's different from Amazon Cognito Region
        identityPoolRegion: 'XX-XXXX-X',

        // OPTIONAL - Amazon Cognito User Pool ID
        userPoolId: 'XX-XXXX-X_abcd1234',

        // OPTIONAL - Amazon Cognito Web Client ID (26-char alphanumeric string)
        userPoolWebClientId: 'a1b2c3d4e5f6g7h8i9j0k1l2m3',

        // OPTIONAL - Enforce user authentication prior to accessing AWS resources or not
        mandatorySignIn: false,

    }
});
```

Also if the user is autoConfirmed the library will sign in the user as well automatically.

[...]

Issue #3882, opened 2019-08-21
After successful registration and confirmation, how can we login the user automatically?


damianesteban:
**Is your feature request related to a problem? Please describe.**
This feature was requested in the original Cognito SDK for JavaScript over three years ago:

https://github.com/amazon-archives/amazon-cognito-identity-js/issues/186

**Describe the solution you'd like**
We would like the user to be able to automatically sign in after confirming the user’s email. 

**Describe alternatives you've considered**
There are a number of solutions that could be implemented but seeing as how Cognito is an identity provider it should provide this functionality. 

**Additional Context:**

Again, this was requested over three years ago. It’s great that Amplify can interface with Polly and Sumerian and everything else, but the identity provider portion is severely lagging behind the competition (not only Auth0, but even Firebase). 

Had this been brought up at all in discussions about features?

potatodepaulo:
This seems like it could be prioritized higher! Pretty common pattern, right?

abdallahshaban557:
We agree! We are definitely working on enabling this in the near future. Once we have exact timelines, we will share our updates on Github.

elorzafe:
@damianesteban @potatodepaulo 

Today we released `autoSignIn` feature on Amplify JS. We will be updating the docs tomorrow, but the feature is available now on version `aws-amplify@4.3.29`

If you are using code verification for new users you only have to do.

```javascript
 Auth.signUp({
      username: 'xxxxxx',
      password: '*********,
      attributes: {
        email: 'xxxxxxxxxx'
      },
      autoSignIn: {
        enabled: true
      }
    })
````

The library will login in the user automatically after the user is confirmed on the App when doing `Auth.confirmSignUp`


If you are using link verification the library also handle that case but you need to configure the library is using `link` verification on `Amplify.configure` like this

```javascript
Amplify.configure({
    Auth: {
       // OPTIONAL
       signUpVerificationMethod: 'link'

        // REQUIRED only for Federated Authentication - Amazon Cognito Identity Pool ID
        identityPoolId: 'XX-XXXX-X:XXXXXXXX-XXXX-1234-abcd-1234567890ab',
        
        // REQUIRED - Amazon Cognito Region
        region: 'XX-XXXX-X',

        // OPTIONAL - Amazon Cognito Federated Identity Pool Region 
        // Required only if it's different from Amazon Cognito Region
        identityPoolRegion: 'XX-XXXX-X',

        // OPTIONAL - Amazon Cognito User Pool ID
[...]

Issue #3631, opened 2019-07-11
Simple way to force login after a user enters their confirmation code (instead of forcing the user to type in the email/password again).


ajhool:
** Which Category is your question related to? **
Auth

** What AWS Services are you utilizing? **
aws-amplify and aws-amplify-react

** Provide additional details e.g. code snippets **

Currently, when a user Registers (provides email + password) and is sent a confirmation code (eg. 12345 is sent to their email/phone) which they enter into an aws-amplify-react Confirmation page. Then, the user is forced to enter their email name and password, again, into a SignIn form. However, the natural flow would be to automatically log them into their account once they have Registered and Confirmed.

This is bad friction at a critical point in a user's first use of the website. I'm aware that the team isn't going to touch aws-amplify-react components while they do the Stencil revamp, but does anybody have a working patch for the time being?

houmark:
I think partly the reason for this is to re-confirm the password right after signup, as the signup form only has one password field (no repeat).

Considering that you'd have a hard time capturing the password, there's no workaround besides making your own component which is not that hard actually. And you could enable other benefits in the process, such as a working ENTER submit and also keeping a more persistent step in local storage.

We had people signing up, getting the verification code, closing the browser in the meantime and losing the actual step, so they have to figure out to sign in, and then get greeted by the missing verification code that would then be sent once more (which renders the old one they just copied from their mail invalid). To prevent all of this, we did our own component which uses Auth to sign the user up and in, and also keeps track of their process in local storage, so if you reload the page or close the browser, to come back later, then the verification form would load up as you enter, ready to receive the code.

Doing each component is a few hours depending on what UI components you have already ready to support this.

And one more very important thing (I'm was using the React component):
If a user signs up with My@email.com, then that's a different user than my@email.com and you have no way of making sure it's lowercased, so with that you have serious problems waiting to happen.

Want one more? Sometimes it can be nice to send along some signup information, what page did the user sign up from, which button did they click etc. That's not possible in the core component either, but you would be able to hook that in using `attributes` and have a Lambda function react on that value after "signup confirmation".

This got way longer than I wanted it to be :) But here's my feedback after trying to use the core components, only to realize that the only way I could get to where we needed was by making my own. The core ones were a nice quick/kick starter though but for production use, I honestly don't think anyone should use them.

ajhool:
Thanks for the thoughtful response. Can't believe that cognito doesn't have a flag for case-insensitive emails in the userpool settings.

[...]

PR #10064, opened 2022-07-06
feat(@aws-amplify/auth): Auto sign in after sign up

helgabalashova:
#### Description of changes
Changes add confirm sign up listener to initiate auto sign in once account is confirmed. It happens if autoSignIn is set to true in params for SignUp function.
Since customers can choose either code, link, or autoConfirm, the solution use three different ways to auto sign in. autoConfirm implementation immediately calls on confirmSignUp function. Link implementation uses polling to sign user in since there is no explicit way to know if user clicked the link. Polling is done within 3 minutes (time verification link is valid) every 5 seconds. Code implementation is using hub listening event. Once user confirms account with code, it dispatches listening event. SignUp function is listening for this event to initiate SignIn process.

#### Issue #, if available
https://github.com/aws-amplify/amplify-js/issues/6320
https://github.com/aws-amplify/amplify-js/issues/3882
https://github.com/aws-amplify/amplify-js/issues/3631
https://github.com/aws-amplify/amplify-js/issues/6018

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/auth/__tests__/auth-unit-test.ts
+++ b/packages/auth/__tests__/auth-unit-test.ts
@@ -600,7 +600,7 @@ describe('auth unit test', () => {
 					phone_number: 'phone_number',
 					otherAttrs: 'otherAttrs',
 				},
-				autoSignIn: true,
+				autoSignIn: { enabled: true },
 			};
 			expect(await auth.signUp(attrs)).toBe(signUpResult);
 			expect(signInSpyon).toHaveBeenCalledTimes(1);
@@ -624,7 +624,7 @@ describe('auth unit test', () => {
 					phone_number: 'phone_number',
 					otherAttrs: 'otherAttrs',
 				},
-				autoSignIn: true,
+				autoSignIn: { enabled: true },
 			};
 			expect(await auth.signUp(attrs)).toBe('signUpResult');
 			expect(await auth.confirmSignUp('username', 'code')).toBe('Success');
@@ -647,7 +647,7 @@ describe('auth unit test', () => {
 					phone_number: 'phone_number',
 					otherAttrs: 'otherAttrs',
 				},
-				autoSignIn: true,
+				autoSignIn: { enabled: true },
 			};
 			expect(await auth.signUp(attrs)).toBe('signUpResult');
 			jest.advanceTimersByTime(11000);
@@ -675,7 +675,7 @@ describe('auth unit test', () => {
 					phone_number: 'phone_number',
 					otherAttrs: 'otherAttrs',
 				},
-				autoSignIn: true,
+				autoSignIn: { enabled: true },
 			};
 			expect(await auth.signUp(attrs)).toBe('signUpResult');
 			try {
--- a/packages/auth/__tests__/auth-unit-test.ts
+++ b/packages/auth/__tests__/auth-unit-test.ts
@@ -8,6 +8,7 @@ import {
 	CognitoIdToken,
 	CognitoAccessToken,
 	NodeCallback,
+	ISignUpResult,
 } from 'amazon-cognito-identity-js';
 
 const MAX_DEVICES: number = 60;
@@ -319,6 +320,14 @@ const authOptionsWithHostedUIConfig: AuthOptions = {
 		responseType: 'code',
 	},
 };
+const authOptionConfirmationLink: AuthOptions = {
+	userPoolId: 'awsUserPoolsId',
+	userPoolWebClientId: 'awsUserPoolsWebClientId',
+	region: 'region',
+	identityPoolId: 'awsCognitoIdentityPoolId',
+	mandatorySignIn: false,
+	verificationMethod: 'link',
+};
 
 const authOptionsWithClientMetadata: AuthOptions = {
 	userPoolId: 'awsUserPoolsId',
@@ -342,6 +351,13 @@ const userPool = new CognitoUserPool({
 	UserPoolId: authOptions.userPoolId,
 	ClientId: authOptions.userPoolWebClientId,
 });
+
+const signUpResult: ISignUpResult = {
+	user: null,
+	userConfirmed: true,
+	userSub: 'userSub',
+	codeDeliveryDetails: null,
+};
 
 const idToken = new CognitoIdToken({ IdToken: 'idToken' });
 const accessToken = new CognitoAccessToken({ AccessToken: 'accessToken' });
@@ -555,6 +571,122 @@ describe('auth unit test', () => {
 			expect.assertions(2);
 			expect(auth.signUp('username').then()).rejects.toThrow(AuthError);
 			expect(auth.signUp('username').then()).rejects.toEqual(errorMessage);
+		});
+	});
+
+	describe('autoSignInAfterSignUp', () => {
+		test('happy case auto confirm', async () => {
+			const spyon = jest
+				.spyOn(CognitoUserPool.prototype, 'signUp')
+				.mockImplementationOnce(
+					(
+						username,
+						password,
+						signUpAttributeList,
+						validationData,
+						callback,
+						clientMetadata
+					) => {
+						callback(null, signUpResult);
+					}
+				);
+			const signInSpyon = jest.spyOn(CognitoUser.prototype, 'authenticateUser');
+			const auth = new Auth(authOptions);
+			const attrs = {
+				username: 'username',
+				password: 'password',
+				attributes: {
+					email: 'email',
+					phone_number: 'phone_number',
+					otherAttrs: 'otherAttrs',
+				},
+				autoSignIn: true,
+			};
+			expect(await auth.signUp(attrs)).toBe(signUpResult);
+			expect(signInSpyon).toHaveBeenCalledTimes(1);
+			spyon.mockClear();
+			signInSpyon.mockClear();
+		});
+
+		test('happy case confirmation code', async () => {
+			const spyon = jest.spyOn(CognitoUserPool.prototype, 'signUp');
+			const confirmSpyon = jest.spyOn(
+				CognitoUser.prototype,
+				'confirmRegistration'
+			);
+			const signInSpyon = jest.spyOn(CognitoUser.prototype, 'authenticateUser');
+			const auth = new Auth(authOptions);
+			const attrs = {
+				username: 'username',
+				password: 'password',
+				attributes: {
+					email: 'email',
+					phone_number: 'phone_number',
+					otherAttrs: 'otherAttrs',
+				},
+				autoSignIn: true,
+			};
+			expect(await auth.signUp(attrs)).toBe('signUpResult');
+			expect(await auth.confirmSignUp('username', 'code')).toBe('Success');
+			expect(signInSpyon).toHaveBeenCalledTimes(1);
+			spyon.mockClear();
+			confirmSpyon.mockClear();
+			signInSpyon.mockClear();
+		});
+
+		test('happy case confirmation link', async () => {
+			jest.useFakeTimers();
+			const spyon = jest.spyOn(CognitoUserPool.prototype, 'signUp');
+			const signInSpyon = jest.spyOn(CognitoUser.prototype, 'authenticateUser');
+			const auth = new Auth(authOptionConfirmationLink);
+			const attrs = {
+				username: 'username',
+				password: 'password',
+				attributes: {
+					email: 'email',
+					phone_number: 'phone_number',
+					otherAttrs: 'otherAttrs',
+				},
+				autoSignIn: true,
+			};
+			expect(await auth.signUp(attrs)).toBe('signUpResult');
+			jest.advanceTimersByTime(11000);
+			expect(signInSpyon).toHaveBeenCalledTimes(2);
+			spyon.mockClear();
+			signInSpyon.mockClear();
+		});
+
+		test('fail confirmation code', async () => {
+			const spyon = jest.spyOn(CognitoUserPool.prototype, 'signUp');
+			const confirmSpyon = jest
+				.spyOn(CognitoUser.prototype, 'confirmRegistration')
+				.mockImplementationOnce(
+					(confirmationCode, forceAliasCreation, callback) => {
+						callback('err', null);
+					}
+				);
+			const signInSpyon = jest.spyOn(CognitoUser.prototype, 'authenticateUser');
+			const auth = new Auth(authOptions);
+			const attrs = {
+				username: 'username',
+				password: 'password',
+				attributes: {
+					email: 'email',
+					phone_number: 'phone_number',
+					otherAttrs: 'otherAttrs',
+				},
+				autoSignIn: true,
+			};
+			expect(await auth.signUp(attrs)).toBe('signUpResult');
+			try {
+				await auth.confirmSignUp('username', 'code');
+			} catch (e) {
+				expect(e).toBe('err');
+			}
+			expect(signInSpyon).toHaveBeenCalledTimes(0);
+			spyon.mockClear();
+			confirmSpyon.mockClear();
+			signInSpyon.mockClear();
 		});
 	});
 
@@ -1517,6 +1649,7 @@ describe('auth unit test', () => {
 	describe('currentSession', () => {
 		afterEach(() => {
 			jest.clearAllMocks();
+			jest.useRealTimers();
 		});
 		test('happy case', async () => {
 			const auth = new Auth(authOptions);
--- a/packages/auth/__tests__/auth-unit-test.ts
+++ b/packages/auth/__tests__/auth-unit-test.ts
@@ -326,7 +326,7 @@ const authOptionConfirmationLink: AuthOptions = {
 	region: 'region',
 	identityPoolId: 'awsCognitoIdentityPoolId',
 	mandatorySignIn: false,
-	verificationMethod: 'link',
+	signUpVerificationMethod: 'link',
 };
 
 const authOptionsWithClientMetadata: AuthOptions = {
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -275,6 +275,7 @@ export class AuthClass {
 					AuthErrorTypes.AutoSignInError
 				);
 			}
+			this._storage.removeItem('pollingStarted');
 		}
 		return this._config;
 	}
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -267,7 +267,9 @@ export class AuthClass {
 			!this.autoSignInInitiated &&
 			typeof this._storage['getItem'] === 'function'
 		) {
-			const pollingInitiated = this.isTrueStorageValue('pollingStarted');
+			const pollingInitiated = this.isTrueStorageValue(
+				'amplify-polling-started'
+			);
 			if (pollingInitiated) {
 				dispatchAuthEvent(
 					'autoSignIn_failure',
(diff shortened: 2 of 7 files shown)


packages/auth/src/Auth.ts
```diff
@@ -257,6 +261,20 @@ export class AuthClass {
 			null,
 			`The Auth category has been configured successfully`
 		);
+
+		if (
+			!this.autoSignInInitiated &&
+			typeof this._storage['getItem'] === 'function'
+		) {
+			const pollingInitiated = this._storage.getItem('pollingStarted') || false;
```
jamesaucode, line 269: should we check `this._storage.getItem('pollingStarted') === 'true'` instead? The value here could be `'true'` or `'false'` (the string, not the boolean), which is still going to be a truthy value

packages/auth/src/Auth.ts
```diff
@@ -387,6 +443,41 @@ export class AuthClass {
...
+							'AutoSignIn',
+							value,
+							`${authDetails.getUsername()} has signed in successfully`
+						);
+						if (listenEvent) {
+							Hub.remove('auth', listenEvent);
+						}
+						if (autoSignInPolling) {
+							clearInterval(autoSignInPolling);
+							this._storage.setItem('pollingStarted', false);
```
jamesaucode, line 468: Make sure we are checking if `this._storage.getItem('pollingStarted') === 'true'` since boolean will be stored as string in `localStorage`. Since this could be `'false'`, if we don't check it's `=== 'true'` a `'false'` still gives us a truthy value.

packages/auth/src/Auth.ts
jamesaucode: nit: I would define this key as a `const` somewhere 😄

packages/auth/src/Auth.ts
```diff
@@ -387,6 +443,41 @@ export class AuthClass {
 		});
 	}
 
+	private async onConfirmSignUp(
+		authDetails: AuthenticationDetails,
+		listenEvent?: HubCallback,
+		autoSignInPolling?: NodeJS.Timer
```
jamesaucode, line 449: a NodeJS type here seems a bit out of place. I would use `ReturnType<typeof setTimeout>` instead to make Typescript happy.

jamesaucode (changes requested):
Looks good 😄 Left a few comments

packages/auth/src/types/Auth.ts
```diff
@@ -50,6 +51,7 @@ export interface AuthOptions {
 	identityPoolRegion?: string;
 	clientMetadata?: any;
 	endpoint?: string;
+	verificationMethod?: string;
```
nickarocho, line 54: Should this be an `enum` if there are only three possible values?

nickarocho (commented):
Great work, Olya! Left one high-level comment and a clarifying TS question. Also wanted to ask before approving - what is the plan with adding unit/integ tests to this feature?

packages/auth/src/Auth.ts
```diff
@@ -345,6 +364,10 @@ export class AuthClass {
 					);
 				});
 			}
+			autoSignIn = params['autoSignIn'] || false;
```
elorzafe, line 367:
```suggestion
			autoSignIn = params.autoSignIn ?? false;
```

packages/auth/src/Auth.ts
```diff
@@ -379,6 +402,39 @@ export class AuthClass {
...
+									if (Date.now() - start > MAX_AUTOSIGNIN_POLLING_TIME) {
+										clearInterval(autoSignInPolling);
+										dispatchAuthEvent(
+											'AutoSignInFail',
+											null,
+											'Failed to Auto Sign In. Please confirm account and try to sign in again.'
+										);
+									}
+									this.onConfirmSignUp(authDetails, null, autoSignInPolling);
+								}, 5000);
```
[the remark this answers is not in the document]
elorzafe: I agree with @nickarocho can look cleaner by exporting this into a function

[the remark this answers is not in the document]
elorzafe: Also split the logic for link and code separately

packages/auth/src/Auth.ts
```diff
@@ -379,6 +402,39 @@ export class AuthClass {
...
+								const autoSignInPolling = setInterval(() => {
+									if (Date.now() - start > MAX_AUTOSIGNIN_POLLING_TIME) {
+										clearInterval(autoSignInPolling);
+										dispatchAuthEvent(
+											'AutoSignInFail',
+											null,
+											'Failed to Auto Sign In. Please confirm account and try to sign in again.'
+										);
+									}
+									this.onConfirmSignUp(authDetails, null, autoSignInPolling);
```
elorzafe, line 427: Would you need to `return` on the `if` or make it an `else` statement for `this.onConfirmSignUp`

packages/auth/src/Auth.ts
```diff
@@ -387,6 +443,41 @@ export class AuthClass {
...
+							'AutoSignIn',
+							value,
+							`${authDetails.getUsername()} has signed in successfully`
+						);
+						if (listenEvent) {
+							Hub.remove('auth', listenEvent);
+						}
+						if (autoSignInPolling) {
+							clearInterval(autoSignInPolling);
+							this._storage.setItem('pollingStarted', false);
```
elorzafe, line 468: Why not delete?

packages/auth/src/Auth.ts
```diff
@@ -429,6 +520,19 @@ export class AuthClass {
 					if (err) {
 						reject(err);
 					} else {
+						dispatchAuthEvent(
+							'confirmSignUp',
+							data,
+							`${username} has been confirmed successfully`
+						);
+						const autoSignIn = this._storage.getItem('autoSignIn') || false;
+						if (autoSignIn === 'true' && !this.autoSignInInitiated) {
```
elorzafe, line 529: Should this be `true` instead, serializing data to boolean from storage seem a little hacky, maybe having a helper function that allows to treat this better? What do you think?

packages/auth/src/types/Auth.ts
elorzafe: or `'code'` | `'link'`

packages/auth/src/Auth.ts
```diff
@@ -95,6 +96,8 @@ const dispatchAuthEvent = (event: string, data: any, message: string) => {
 // https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListDevices.html#API_ListDevices_RequestSyntax
 const MAX_DEVICES = 60;
 
+const MAX_AUTOSIGNIN_POLLING_TIME = 180000;
```
elorzafe, line 99: I would add the unit like `MAX_AUTOSIGNIN_POLLING_MS`, and split into minutes like this `3 * 60 * 1000`

packages/auth/src/Auth.ts
```diff
@@ -443,6 +421,54 @@ export class AuthClass {
...
+				this.onConfirmSignUp(authDetails, listenEvent);
+			}
+		};
+		Hub.listen('auth', listenEvent);
+	}
+
+	private handleLinkAutoSignIn(authDetails: AuthenticationDetails) {
+		this._storage.setItem('pollingStarted', true);
+		const start = Date.now();
+		const autoSignInPolling = setInterval(() => {
```
elorzafe, line 459: Could be more clear `autoSignInPollingIntervalId`?

elorzafe: or `pollingIntervalId`

packages/auth/src/Auth.ts
```diff
@@ -541,6 +569,10 @@ export class AuthClass {
 		});
 	}
 
+	private isTrueStorageValue(value: any) {
```
elorzafe, line 572: what if instead of passing the value you pass the key of the storage or rename the function because it seems that is getting the value from Storage which isnt

elorzafe (commented):
Just minor comments,

Have a callout related to `clientMetadata` and `validationData` because what happens if the customer wants to use that for `autoSignIn`?

oemer-sellwerk:
When can we expect this to be merged?

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -558,6 +574,122 @@ describe('auth unit test', () => {
...
+						password,
+						signUpAttributeList,
+						validationData,
+						callback,
+						clientMetadata
+					) => {
+						callback(null, signUpResult);
+					}
+				);
+			const signInSpyon = jest.spyOn(CognitoUser.prototype, 'authenticateUser');
```
elorzafe, line 593: I would add a mock for `authenticateUser` in order to not make the network request inside of the test (it wont impact the test now but maybe it will impact the whole suite)

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -558,6 +574,122 @@ describe('auth unit test', () => {
...
+			signInSpyon.mockClear();
+		});
+
+		test('happy case confirmation code', async () => {
+			const spyon = jest.spyOn(CognitoUserPool.prototype, 'signUp');
+			const confirmSpyon = jest.spyOn(
+				CognitoUser.prototype,
+				'confirmRegistration'
+			);
+			const signInSpyon = jest.spyOn(CognitoUser.prototype, 'authenticateUser');
```
elorzafe, line 617: I would add a mock for this to avoid network request

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -558,6 +574,122 @@ describe('auth unit test', () => {
...
+				autoSignIn: { enabled: true },
+			};
+			expect(await auth.signUp(attrs)).toBe(signUpResult);
+			expect(signInSpyon).toHaveBeenCalledTimes(1);
+			spyon.mockClear();
+			signInSpyon.mockClear();
+		});
+
+		test('happy case confirmation code', async () => {
+			const spyon = jest.spyOn(CognitoUserPool.prototype, 'signUp');
```
elorzafe, line 612: I think this might need a mock Implementation. Maybe is using another mock that responds with `userConfirmed: false`?

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -558,6 +574,122 @@ describe('auth unit test', () => {
...
+			const confirmSpyon = jest.spyOn(
```
elorzafe, line 613: You might need a mock for this, or is using a global mock?

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -558,6 +574,122 @@ describe('auth unit test', () => {
...
+			expect(signInSpyon).toHaveBeenCalledTimes(1);
+			spyon.mockClear();
+			confirmSpyon.mockClear();
+			signInSpyon.mockClear();
+		});
+
+		test('happy case confirmation link', async () => {
+			jest.useFakeTimers();
+			const spyon = jest.spyOn(CognitoUserPool.prototype, 'signUp');
+			const signInSpyon = jest.spyOn(CognitoUser.prototype, 'authenticateUser');
```
elorzafe, line 640: Maybe you need mocks for this or is using a global mock?

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -558,6 +574,122 @@ describe('auth unit test', () => {
...
+			};
+			expect(await auth.signUp(attrs)).toBe('signUpResult');
+			jest.advanceTimersByTime(11000);
+			expect(signInSpyon).toHaveBeenCalledTimes(2);
+			spyon.mockClear();
+			signInSpyon.mockClear();
+		});
+
+		test('fail confirmation code', async () => {
+			const spyon = jest.spyOn(CognitoUserPool.prototype, 'signUp');
```
elorzafe, line 660: Is this using a global mock?

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -1517,6 +1649,7 @@ describe('auth unit test', () => {
 	describe('currentSession', () => {
 		afterEach(() => {
 			jest.clearAllMocks();
+			jest.useRealTimers();
```
elorzafe, line 1652: Why is this needed?

packages/auth/src/Auth.ts
```diff
@@ -257,6 +262,20 @@ export class AuthClass {
...
+		if (
+			!this.autoSignInInitiated &&
+			typeof this._storage['getItem'] === 'function'
+		) {
+			const pollingInitiated = this.isTrueStorageValue('pollingStarted');
+			if (pollingInitiated) {
+				dispatchAuthEvent(
+					'autoSignIn_failure',
+					null,
+					'Please use your credentials to sign in.'
```
elorzafe, line 275: You can add this on the `AuthError.ts` file

packages/auth/src/Auth.ts
```diff
@@ -257,6 +262,20 @@ export class AuthClass {
 			null,
 			`The Auth category has been configured successfully`
 		);
+
+		if (
+			!this.autoSignInInitiated &&
+			typeof this._storage['getItem'] === 'function'
+		) {
+			const pollingInitiated = this.isTrueStorageValue('pollingStarted');
+			if (pollingInitiated) {
```
elorzafe, line 271: I think you need to clear `pollingStarted`

packages/auth/src/Auth.ts
```diff
@@ -295,6 +314,9 @@ export class AuthClass {
 		const attributes: CognitoUserAttribute[] = [];
 		let validationData: CognitoUserAttribute[] = null;
 		let clientMetadata;
+		let autoSignIn: AutoSignInOptions = { enabled: false };
+		const autoSignInValidationData: CognitoUserAttribute[] = null;
```
elorzafe, line 318: Are you sure that is the right type?

packages/auth/src/Auth.ts
```diff
@@ -429,6 +559,19 @@ export class AuthClass {
...
+							'confirmSignUp',
+							data,
+							`${username} has been confirmed successfully`
+						);
+						const autoSignIn = this.isTrueStorageValue('autoSignIn');
+						if (autoSignIn && !this.autoSignInInitiated) {
+							dispatchAuthEvent(
+								'autoSignIn_failure',
+								null,
+								'Please use your credentials to sign in.'
```
elorzafe, line 572: I would reuse the same constant from `AuthErrors.ts`

packages/auth/src/Auth.ts
```diff
@@ -345,6 +367,22 @@ export class AuthClass {
 					);
 				});
 			}
+
+			autoSignIn = params.autoSignIn ?? { enabled: false };
+			if (autoSignIn.enabled) {
+				this._storage.setItem('autoSignIn', true);
```
elorzafe, line 373: We are not removing this. Can you check where this should be removed?

elorzafe (changes requested):
It looks very good, just some adjustments need to be made, we can chat more offline.

Thanks @helgabalashova !

packages/auth/__tests__/auth-unit-test.ts
helgabalashova: it is a global mock

packages/auth/__tests__/auth-unit-test.ts
helgabalashova: This one is also a global mock

packages/auth/__tests__/auth-unit-test.ts
helgabalashova: yes

packages/auth/__tests__/auth-unit-test.ts
helgabalashova: it breaks all other tests that rely on time to pass

packages/auth/src/Auth.ts
```diff
@@ -387,6 +426,90 @@ export class AuthClass {
...
+		const start = Date.now();
+		const autoSignInPollingIntervalId = setInterval(() => {
+			if (Date.now() - start > MAX_AUTOSIGNIN_POLLING_MS) {
+				clearInterval(autoSignInPollingIntervalId);
+				dispatchAuthEvent(
+					'autoSignIn_failure',
+					null,
+					'Please confirm your account and use your credentials to sign in.'
+				);
+			}
```
elorzafe, line 472: This could be a bug, if this is true it should return otherwise it could throw the error and also try to signIn

packages/auth/src/Auth.ts
```diff
@@ -387,6 +426,90 @@ export class AuthClass {
...
+			this.onConfirmSignUp(authDetails, null, autoSignInPollingIntervalId);
+		}, 5000);
+	}
+
+	private async onConfirmSignUp(
```
elorzafe, line 477: ` signInAfterUserConfirmed` ?

seems odd using `this.onConfirmSignUp` from polling use case, what do you think?

elorzafe (changes requested):
It looks very good and is super close, can you also add `amplify-` as a prefix for all the items persisted on localStorage?
Thanks @helgabalashova !

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -558,6 +574,122 @@ describe('auth unit test', () => {
...
+				password: 'password',
+				attributes: {
+					email: 'email',
+					phone_number: 'phone_number',
+					otherAttrs: 'otherAttrs',
+				},
+				autoSignIn: { enabled: true },
+			};
+			expect(await auth.signUp(attrs)).toBe(signUpResult);
+			expect(signInSpyon).toHaveBeenCalledTimes(1);
```
jamesaucode, line 606: Can we also test that the storage is clearing the flag properly, and `Hub` is removing the callback function?

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -558,6 +574,122 @@ describe('auth unit test', () => {
...
+				attributes: {
+					email: 'email',
+					phone_number: 'phone_number',
+					otherAttrs: 'otherAttrs',
+				},
+				autoSignIn: { enabled: true },
+			};
+			expect(await auth.signUp(attrs)).toBe('signUpResult');
+			expect(await auth.confirmSignUp('username', 'code')).toBe('Success');
+			expect(signInSpyon).toHaveBeenCalledTimes(1);
```
jamesaucode, line 631: Same as above

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -558,6 +574,122 @@ describe('auth unit test', () => {
...
+				attributes: {
+					email: 'email',
+					phone_number: 'phone_number',
+					otherAttrs: 'otherAttrs',
+				},
+				autoSignIn: { enabled: true },
+			};
+			expect(await auth.signUp(attrs)).toBe('signUpResult');
+			jest.advanceTimersByTime(11000);
+			expect(signInSpyon).toHaveBeenCalledTimes(2);
```
jamesaucode, line 654: same as above

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -558,6 +574,122 @@ describe('auth unit test', () => {
...
+				},
+				autoSignIn: { enabled: true },
+			};
+			expect(await auth.signUp(attrs)).toBe('signUpResult');
+			try {
+				await auth.confirmSignUp('username', 'code');
+			} catch (e) {
+				expect(e).toBe('err');
+			}
+			expect(signInSpyon).toHaveBeenCalledTimes(0);
```
jamesaucode, line 686: same as above

jamesaucode (commented):
Looks good to me for the most part, good job @helgabalashova  😄  For the unit tests, can we also test that the flag inside storage is cleared properly, and also the `Hub` event is removed?

elorzafe (approved):
Looks great! Thanks @helgabalashova 🏅 🌮

PR #10086, opened 2022-07-15
Datastore/feat user agent suffix

erinleigh90:
Enriches the user-agent to include information on whether requests to AppSync are coming from the Datastore category.
Accomplished by appending "/DataStore" to the user-agent for calls coming from datastore through the GraphQL API.
To achieve this, an additional parameter, userAgentSuffix, is added to the GraphQLOptions sent into GraphQLApiClass graphql() method. Within Datastore, any calls to graphql should contain the userAgentSuffix parameter with a value of "/DataStore."

Placed the method to append the suffix in the core library so we can implement this elsewhere on future calls as we work to fine-tune our metrics and understand what libraries are most actively used.

Some variable declarations in mutation.test.ts and subscription.test.ts had to be moved above imports because of the way jest hoists global mocks. If we want to be able to validate what is sent to the mocked methods, we have to declare the variables at the very top of the file to be recognized as in-scope by the mock function.

Steps followed to verify expected results: 
1. Linked updated libraries to a datastore sample app
2. Monitored Network to verify 
    a. The user agent had the datastore suffix appended to it for calls originating from the datastore library. 
    b. The user agent did not have the datastore suffix appended to it for calls originating from the api library.
4. Abdallah Shaban confirmed that the suffix is coming through in our metrics.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


stocaaro (commented):
Good stuff. I think I can follow this all the way through. Have a couple minor comments below.

packages/datastore/src/util.ts
```diff
@@ -29,6 +29,21 @@ import {
 } from './types';
 import { WordArray } from 'amazon-cognito-identity-js';
 
+export enum NAMESPACES {
```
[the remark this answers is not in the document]
erinleigh90: I moved them up because it seemed like a good idea to have constants at the top (i couldn't find constants until later and realized they were hidden at the bottom) I suppose I should have called that out in a commit message.

packages/core/src/Platform/index.ts
```diff
@@ -35,8 +35,8 @@ if (typeof navigator !== 'undefined' && navigator.product) {
 	}
 }
 
-export const getAmplifyUserAgent = () => {
-	return Platform.userAgent;
+export const getAmplifyUserAgent = (content?: string | null) => {
```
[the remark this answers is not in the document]
erinleigh90: I found that i needed to be able to actively pass null at some point, potentially because I was thinking I couldn't set to undefined. I will look into where I thought this was necessary and verify.

packages/datastore/__tests__/mutation.test.ts
```diff
@@ -143,6 +150,55 @@ describe('MutationProcessor', () => {
...
+				},
+				signerServiceInfo: { region: 'us-west-2', service: 'appsync' },
+			};
+			const expectedChild = {
+				headers: {
+					Authorization: null,
+					'X-Api-Key': 'da2-xxxxxxxxxxxxxxxxxxxxxx',
+					'x-amz-user-agent': 'aws-amplify/4.5.9 js/DataStore',
+				},
+			};
```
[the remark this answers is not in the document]
erinleigh90: That is what that looks like. Removing. Thank you for catching that!

dpilch (approved):

packages/core/__tests__/Platform-test.ts
```diff
@@ -1,3 +1,4 @@
+import { getAmplifyUserAgent } from '../lib-esm';
```
ashika01, line 1: Change this import from `src/core/platform` shouldnt be `lib-esm`

packages/datastore/__tests__/mutation.test.ts
```diff
@@ -1,3 +1,4 @@
+const mockRestPost = jest.fn();
```
ashika01, line 1: nit: i would adjust these order to have `imports` at top. Should have been a lint config no?

packages/api-graphql/__tests__/GraphQLAPI-test.ts
```diff
@@ -1237,7 +1237,63 @@ describe('API test', () => {
 				},
 			});
 		});
-	});
+
+		test('happy-case-query with userAgentSuffix', async () => {
```
ashika01, line 1241: nit: change description to not say happy-case instead maybe say `userAgent on request`

packages/datastore/__tests__/subscription.test.ts
```diff
@@ -1,7 +1,10 @@
+import Observable from 'zen-observable-ts';
+let mockObservable = new Observable(() => {});
```
ashika01, line 2: same nit as above imports followed by variable declarations

packages/datastore/__tests__/subscription.test.ts
```diff
@@ -16,14 +19,13 @@ import {
 	InternalSchema,
 	PersistentModelConstructor,
 } from '../src/types';
-
-let mockObservable = new Observable(() => {});
+import { USER_AGENT_SUFFIX_DATASTORE } from '../lib-esm/util';
```
ashika01, line 22: no lib-esm imports

packages/datastore/__tests__/sync.test.ts
```diff
@@ -281,6 +281,41 @@ describe('Sync', () => {
...
+
+			await SyncProcessor.jitteredRetry({
+				query: defaultQuery,
+				variables: defaultVariables,
+				opName: defaultOpName,
+				modelDefinition: defaultModelDefinition,
+			});
+
+			expect(mockGraphQl).toHaveBeenCalledWith(
+				expect.objectContaining({ userAgentSuffix: '/DataStore' })
```
ashika01, line 316: nit: u can use ur `const` created

ashika01 (changes requested):

packages/datastore/__tests__/mutation.test.ts
erinleigh90: Since jest hoists mock calls, I had to put the constant at the top of the file for it to be in scope. I attempted to put below the imports and the constant was not recognized.

packages/datastore/__tests__/subscription.test.ts
erinleigh90: Fixed

packages/datastore/__tests__/sync.test.ts
erinleigh90: Fixed

david-mcafee (changes requested):
Left a few minor comments - overall looks good!! :)

packages/datastore/__tests__/sync.test.ts
```diff
@@ -1,5 +1,5 @@
 // These tests should be replaced once SyncEngine.partialDataFeatureFlagEnabled is removed.
-
+let mockGraphQl;
```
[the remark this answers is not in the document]
erinleigh90: Actually I tested to verify in this file and it appeared to work either way. In the above files pointed out by Ashika, when I tested with them previously the tests did not recognize the variable unless it was at the very top. I wonder if that is related to the number or size of imports? I will validate in other files as well, but will change in this file.

packages/datastore/__tests__/mutation.test.ts
erinleigh90: Validated that the test fails when const is moved directly under imports.

packages/api-graphql/__tests__/GraphQLAPI-test.ts
erinleigh90: done

ashika01 (approved):
Great work @erinleigh90 I am good with this PR 👍

packages/datastore/__tests__/sync.test.ts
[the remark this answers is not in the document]
ashika01: Yup i think the error is happening when the mock needs to run through the imported module. Something along those lines :)

packages/core/src/Platform/index.ts
```diff
@@ -35,8 +35,8 @@ if (typeof navigator !== 'undefined' && navigator.product) {
 	}
 }
 
-export const getAmplifyUserAgent = () => {
-	return Platform.userAgent;
+export const getAmplifyUserAgent = (content?: string) => {
```
david-mcafee, line 38: I could have sworn I put a comment on this before - but could `content` here, ever be empty? And if so, when?

david-mcafee (commented):
Quick follow-up comment

david-mcafee (approved):
LGTM! 🎉

packages/core/src/Platform/index.ts
david-mcafee: Discussed offline

Issue #10027, opened 2022-06-27
Error: No current user | getStaticProps & getStaticPaths W/ AMAZON_COGNITO_USER_POOLS


temrb:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication, GraphQL API, Storage

### Amplify Categories

auth, storage, api

### Environment information

<details>

```
# Put output below this line

Amplify 8.3.1
NextJS

```

</details>

### Describe the bug

Trying to setup ISR with posts from DynamoDB though keep getting error:

Server Error
Error: No current user

This error happened while generating the page. Any console logs will be displayed in the terminal window.
Call Stack
GraphQLAPIClass.<anonymous>

Term:
error - Error: No current user

I have authcontext with signed in user... using authMode: GRAPHQL_AUTH_MODE.AMAZON_COGNITO_USER_POOLS..
I cannot use APIKEY authmode as these cards are setup for only allow:owner along with almost everything else in my schema.

I have 
```javascript
Amplify.configure({ ...awsconfig, ssr: true });
```

What is the issue ? Or any further setup to get ISR with nextjs?

### Expected behavior

get related querued ID cards on ISR page

I.e. 

http://localhost:4242/card/c8762039-8a6f-477d-8819-ffc9d311dd9b

### Reproduction steps

n/a

### Code Snippet

```javascript
// Put your code below this line.
import React, { ReactElement } from 'react';
import { GetStaticProps, GetStaticPaths } from 'next';
import { withSSRContext, API } from 'aws-amplify';
import { listCards, getCard } from '../../graphql/queries';
import { GRAPHQL_AUTH_MODE } from '@aws-amplify/api';
import { Card, ListCardsQuery, GetCardQuery } from '../../API';

interface Props {
  card: Card;
}

export default function IndividualCard({ card }: Props): ReactElement {
  console.log('card', card);
  return <div></div>;
}

export const getStaticProps: GetStaticProps = async ({ params }) => {
[...]

Issue #8792, opened 2021-08-23
Amplify DataStore creating data only locally on NextJS API Route


casperibo:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Next.js

### Amplify APIs

DataStore

### Amplify Categories

api

### Environment information

<details>

```
# Put output below this line

System:
    OS: macOS 11.4
    CPU: (8) arm64 Apple M1
    Memory: 88.58 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.2.0 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.13.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 92.0.4515.159
    Firefox: 91.0.1
    Safari: 14.1.1
  npmPackages:
    @ampproject/toolbox-optimizer:  undefined ()
    @babel/core:  undefined ()
    @tailwindcss/typography: ^0.4.1 => 0.4.1 
    @types/react: ^17.0.15 => 17.0.19 
    amphtml-validator:  undefined ()
    arg:  undefined ()
    async-retry:  undefined ()
    async-sema:  undefined ()
    autoprefixer: ^10.3.1 => 10.3.2 
    aws-amplify: ^4.2.2 => 4.2.4 
    axios: ^0.21.1 => 0.21.1 
    bfj:  undefined ()
    cacache:  undefined ()
    ci-info:  undefined ()
    comment-json:  undefined ()
    compression:  undefined ()
    conf:  undefined ()
    content-type:  undefined ()
    cookie:  undefined ()
    css-loader:  undefined ()
    debug:  undefined ()
    devalue:  undefined ()
    easymde: ^2.15.0 => 2.15.0 
    escape-string-regexp:  undefined ()
    eslint: 7.32.0 => 7.32.0 
    eslint-config-next: 11.0.1 => 11.0.1 
    file-loader:  undefined ()
    find-cache-dir:  undefined ()
[...]

PR #10088, opened 2022-07-18
fix(datastore): preserve ssr context

iartemiev:
#### Description of changes

The Amplify library exports the `withSSRContext` function (https://docs.amplify.aws/lib/ssr/q/platform/js/#withssrcontext) that allows a Next.js application to use client-side credentials in an isolated context on the server. The function creates new instances of the Auth, API, Credentials, and DataStore classes and configures this instance of Auth with the client-side cookies injected into it from the client-side request.

Currently, a library bug is preventing DataStore from using these instances of Auth and API while running in an SSR Context. The reference to the instance does not get preserved in the DataStore implementation. Therefore, DataStore simply imports new default instances of these classes lacking the client-side auth configuration. This breaks all SSR flows that depend on client-side credentials, such as the ones in the linked issues.

This PR adds a context object that stores references to Auth and API categories in the top-level DataStore class (`datastore.ts`) and passes the context object by reference to the descendent classes that use these categories. This allows DataStore to perform authenticated requests while running in the SSR context, fixing the linked GH issues.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/8792
https://github.com/aws-amplify/amplify-js/issues/10027

1. Created new sample app and test script that validates DataStore with owner Auth with SSR Context: https://github.com/aws-amplify/amplify-js-samples-staging/pull/366 - **_must be merged before merging this PR_**
2. Ran all integ tests (including the new one, integ_next_datastore_owner_auth) against the changes in this branch: https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/12228/workflows/5e01d94a-0456-4448-bdff-f7bdb320b5da
3. Updated unit test for `ssrContext`
4. Tested locally with additional sample apps

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/src/authModeStrategies/multiAuthStrategy.ts
+++ b/packages/datastore/src/authModeStrategies/multiAuthStrategy.ts
@@ -5,6 +5,7 @@ import {
 	ModelAttributeAuthProperty,
 	ModelAttributeAuthProvider,
 	ModelAttributeAuthAllow,
+	AmplifyContext,
 } from '../types';
 
 function getProviderFromRule(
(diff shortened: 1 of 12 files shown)


dpilch (approved):

packages/datastore/src/authModeStrategies/multiAuthStrategy.ts
```diff
@@ -121,29 +122,31 @@ function getAuthRules({
...
-		currentUser = await Auth.currentAuthenticatedUser();
-	} catch (e) {
-		// No current user
-	}
+export const multiAuthStrategy: (
+	amplifyContext: AmplifyContext
+) => AuthModeStrategy =
+	(amplifyContext: AmplifyContext) =>
+	async ({ schema, modelName }) => {
+		amplifyContext.Auth = amplifyContext.Auth || Auth;
```
elorzafe, line 130: You dont need L1 anymore?

packages/datastore/src/datastore/datastore.ts
```diff
@@ -1417,6 +1433,10 @@ class DataStore {
 	};
 
 	configure = (config: DataStoreConfig = {}) => {
+		this.amplifyContext.Auth = this.Auth;
+		this.amplifyContext.API = this.API;
+		this.amplifyContext.Cache = this.Cache;
+
```
elorzafe, line 1439: Is this something that can have implications on how you are importing the library?

For example the order of the imports or if you are when is `amplifyContext` been modified?

elorzafe (commented):
Just a general question related about `amplifyContext`, I think that gets generated when doing `withSSRContext`, but what happens if you dont do that? or if someone calls that after doing `DataStore.start`?

iartemiev:
> Just a general question related about `amplifyContext`, I think that gets generated when doing `withSSRContext`, but what happens if you dont do that? or if someone calls that after doing `DataStore.start`?

@elorzafe with the change in this PR, `amplifyContext` always gets created. But it's a no-op if you're outside of the SSR Context. It'll just reference the default instances of Auth and API.

packages/datastore/src/authModeStrategies/multiAuthStrategy.ts
iartemiev: What's L1 referring to here?

packages/datastore/src/datastore/datastore.ts
iartemiev: > Is this something that can have implications on how you are importing the library? For example the order of the imports 

No, I don't think so. It just retains the reference either to the default instance of the module or to the class instantiated inside of withSSRContext implementation. It doesn't matter how you import the library. 

> or if you are when is amplifyContext been modified?
`amplifyContext` doesn't get modified anywhere outside of the DataStore class. We pass it into the descendent classes as a `readonly` property.

svidgen (approved):
I think this is reasonable. Thanks for fixing this!

Issue #6838, opened 2020-09-20
Storage - more than 1000 keys


marksteele:
**Is your feature request related to a problem? Please describe.**
There are many use cases which involve listing out a more than 1000 keys, however the storage api provides no means of doing so (nor any work-arounds that I could find).

**Describe the solution you'd like**
A mechanism to list out keys supporting paging/offsets/prefixes and delimiters.

**Describe alternatives you've considered**
Would also be possible to do with the raw S3 client if that was accessible from the Storage api.

**Additional context**

harrysolovay:
Hi @marksteele –– I'd recommend creating a custom storage adapter for [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API).

Storage adapters abide by the following interface.

```ts
interface Storage {
  // set item with the key
  static setItem(key: string, value: string): string;
  // get item with the key
  static getItem(key: string): string;
  // remove item with the key
  static removeItem(key: string): void;
  // clear out the storage
  static clear(): void;
  // If the storage operations are async(i.e AsyncStorage)
  // Then you need to sync those items into the memory in this method
  static sync(): Promise<void>;
}
```

And can be provided via a `storage` field in your Amplify config.

In auth, for instance:

```ts
import {Auth} from "@aws-amplify/auth";
import Config from "./aws-exports";
import {MyStorageAdapter} from "./wherever";

Auth.configure({
  ...config,
  storage: new MyStorageAdapter,
});
```

marksteele:
Why would I want to use indexdb? The reason I'm using the Storage API is to leverage S3.

I have users who want to store thousands of files in S3. I'd like to be able to leverage Amplify (makes handling the authentication/token handling bits so simple, it's great!). I understand the reasons for dumbing down the API for some use cases, but to cripple it to the point where it removes some pretty basic functionality seems like a very odd design choice (nobody thought: "what if a user has more than 1000 files?).

Why can't additional functionality be added to the list function to expose more of the underlying S3 API? Is there some fundamental reason you can just add all the options from ListObjectsV2 to the Storage.list calls?

ex:
```
Storage.list('photos/', { level: 'private', ... other options .., listv2: {
ContinuationToken: 'asdf',
  Delimiter: 'asdf/',
  MaxKeys: 1000,
  StartAfter: 'STRING_VALUE'
}
}})
// Prefix would be auto-derived from access level and prefix passed in
/* returns something like
[{
key: 'photos/foo',
[...]

PR #10095, opened 2022-07-19
feat(@aws-amplify/storage): Access all files from S3 with List API

kvramyasri7:
#### Description of changes

In storage list api, users can pass 'ALL' as maxKeys in configuration. This change enables users to access all files from S3.

#### Issue #, if available

#7133 
#6838

- Unit test added for the change
- Tested change in the sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -699,28 +729,34 @@ export class AWSS3Provider implements StorageProvider {
...
+				Prefix: final_path,
+				MaxKeys: maxKeys,
+				ContinuationToken: token,
+			};
+			if (maxKeys === 'ALL') {
+				do {
+					params.ContinuationToken = token;
+					params.MaxKeys = 1000;
+					listResult = await this._list(params, opt, prefix);
+					listResult.contents.map(ele => {
```
jamesaucode, line 748: Looks like we are not using the return of `.map`, we can probably use `.forEach` instead.
Or we can simplify it to 

```js
list.push(...listResult)
```

jamesaucode (changes requested):
Looks pretty good. Left a few suggestions 😄

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -88,6 +89,10 @@ interface AddTaskInput {
 	params?: PutObjectCommandInput;
 }
 
+type S3ProviderListOutputWithToken = {
```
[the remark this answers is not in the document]
kvramyasri7: Fixed

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -678,6 +683,31 @@ export class AWSS3Provider implements StorageProvider {
 			throw error;
 		}
 	}
+	private async _list(
+		params: any,
```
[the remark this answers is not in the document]
kvramyasri7: Fixed

packages/storage/src/providers/AWSS3Provider.ts
kvramyasri7: Fixed

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -694,34 +722,36 @@ export class AWSS3Provider implements StorageProvider {
...
+				Prefix: final_path,
+				MaxKeys: 1000,
+				ContinuationToken: token,
+			};
+			if (maxKeys === 'ALL') {
+				do {
+					params.ContinuationToken = token;
+					params.MaxKeys = 1000;
+					listResult = await this._list(params, opt, prefix);
+					list = [...listResult.contents];
```
jamesaucode, line 746: wouldn't this override the `list` every time it calls `this._list` ? We should be concatenating all the listResults we gathered.

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -694,34 +722,36 @@ export class AWSS3Provider implements StorageProvider {
...
+					if (listResult.nextToken) token = listResult.nextToken;
+				} while (listResult.nextToken);
+			} else {
+				maxKeys < 1000 || typeof maxKeys === 'string'
```
jamesaucode, line 750: Did we discuss in API review what happens if someone put maxKeys > 1000 ? I think at least having a `logger.warn` telling the user `.list` only list at most 1000 keys or they will need to use 'ALL' will be helpful here.

kvramyasri7: Fixed

packages/storage/src/providers/AWSS3Provider.ts
kvramyasri7: Fixed

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
```diff
@@ -1063,13 +1072,48 @@ describe('StorageProvider test', () => {
...
+					lastModified: 'lastmodified',
+					size: 'size',
+				},
+			]);
+			expect(spyon.mock.calls[0][0].input).toEqual({
+				Bucket: 'bucket',
+				Prefix: 'public/path',
+				ContinuationToken: undefined,
+				MaxKeys: 1000,
+			});
```
stocaaro, line 1111: Can we add a test that receives multiple requests so that we observe a continuation token being used? Mocking a sequence of things is a bit of a pain, but it will be less fragile if we have this usecase covered.  Also, will be valuable for testing the token passing version later.

kvramyasri7: Thank you for the suggestion.Done adding this test.

packages/storage/src/types/AWSS3Provider.ts
```diff
@@ -94,15 +95,24 @@ export type S3ProviderRemoveConfig = CommonStorageOptions & {
...
 
 export type S3ProviderListConfig = CommonStorageOptions & {
 	bucket?: string;
-	maxKeys?: number;
+	maxKeys?: number | 'ALL';
 	provider?: 'AWSS3';
 	identityId?: string;
 };
 
+export type NewS3ClientOptions = StorageOptions & {
```
jamesaucode, line 112: It's a little unclear what this type means. What does `NewS3ClientOptions` here mean?

jamesaucode (commented):
LGTM for the most part, just have a few minor comments 😄

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
```diff
@@ -1070,6 +1062,65 @@ describe('StorageProvider test', () => {
...
+			const storage = new StorageProvider();
+			storage.configure(options);
+			const listResultObj = {
+				Key: 'public/path/itemsKey',
+				ETag: 'etag',
+				LastModified: 'lastmodified',
+				Size: 'size',
+			};
+			let methodCalls = 1;
+			let continuationToken = 'token';
```
stocaaro, line 1082: Can this be "TEST_TOKEN" or something like that?

kvramyasri7: Thanks for the suggestion. Done.

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
```diff
@@ -1070,6 +1062,65 @@ describe('StorageProvider test', () => {
...
+				});
+			const storage = new StorageProvider();
+			storage.configure(options);
+			const listResultObj = {
+				Key: 'public/path/itemsKey',
+				ETag: 'etag',
+				LastModified: 'lastmodified',
+				Size: 'size',
+			};
+			let methodCalls = 1;
```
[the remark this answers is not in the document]
kvramyasri7: Yeah. Good Catch. Changed it.

stocaaro (approved):
Like that the test is one long narrative piece now.  No hidden logic. There's probably room to improve/clarify the test a bit, but it works and that can wait for the next change where we'll have more complexity to test against.

jamesaucode (approved):
LGTM 🎉

Issue #9863, opened 2022-05-02
Signer.sign({body:"hello"}, {method:"POST"}) should use `body` attribute like fetch()


maxerbubba:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication

### Amplify Categories

auth

### Environment information

Typescript page, with React, and I am using the Signer.sign() library directly (Sorry not using the full Amplify toolset)
```
import { Signer } from 'aws-amplify';

```

### Describe the bug

1. Run this code
```
Signer.sign({body:"hello"}, {method:"POST"}); // <---- Notice body attribute, where API docs use data.
```
2. output:
```
headers: Content-Length: 0
```

### Expected behavior

One possible solution:
```
headers: Content-Length: 5
```

Another possible solution:
```
throw "attribute `body` found in Signer, but `data` is missing. Please use attribute `data`"
```

### Reproduction steps

create-react-app
import { Signer } from 'aws-amplify';
Signer.sign({body:"hello"}, {method:"POST"})

### Code Snippet

```javascript
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line


```

</details>

ahilashsasidharan:
@maxerbubba  I was looking for clarification in the exact code you ran to get the error. As far as I can tell your code seems to be a more reduced version of code sent to the sign method is it possible to send a more detailed code snippet as sending just the code you sent with the two objects results in an error due to missing information, or some more detail in exactly what you would like implemented. Also could you provide some clarification on the second possible solution```typescript
throw "attribute `body` found in Signer, but `data` is missing. Please use attribute `body`"``` on what you would like implemented for this solution to work.

maxerbubba:
Whoops, I meant to put "data". (I edited the message above)
```
throw "attribute `body` found in Signer, but `data` is missing. Please use attribute `data`"
```

[...]

PR #10113, opened 2022-07-25
feat(@aws-amplify/core): Use body attribute on request object in Sign…

amehi0index:
…er.sign()

#### Description of changes
User can pass in body attribute on the request object when using the sign method from Signer class in Core package

#### Issue #, if available
#9863

Unit test added for the change
Tested change in the sample app

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/__tests__/SignerBody-test.ts
+++ b/packages/core/__tests__/SignerBody-test.ts
@@ -1,0 +1,51 @@
+import Signer from '../src/Signer';
+import { DateUtils } from '../src';
+
+jest.mock('@aws-sdk/util-hex-encoding', () => ({
+	...jest.requireActual('@aws-sdk/util-hex-encoding'),
+	toHex: () => {
+		return 'encrypt';
+	},
+}));
+
+describe('Signer test', () => {
+	describe('sign test', () => {
+		test('happy case', () => {
+			const url = 'https://host/some/path';
+
+			const request = {
+				url,
+				headers: {},
+			};
+			const access_info = {
+				session_token: 'session_token',
+			};
+
+			const spyon = jest
+				.spyOn(Date.prototype, 'toISOString')
+				.mockReturnValueOnce('0');
+
+			const getDateSpy = jest.spyOn(DateUtils, 'getDateWithClockOffset');
+
+			const res = {
+				headers: {
+					Authorization:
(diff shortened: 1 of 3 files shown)


packages/core/__tests__/SignerBody-test.ts
```diff
@@ -0,0 +1,51 @@
...
+jest.mock('@aws-sdk/util-hex-encoding', () => ({
+	...jest.requireActual('@aws-sdk/util-hex-encoding'),
+	toHex: () => {
+		return 'encrypt';
+	},
+}));
+
+describe('Signer test', () => {
+	describe('sign test', () => {
+		test('happy case', () => {
```
ashika01, line 13: could we also add checking the logic of body in request.data in the test?

ashika01: i dont see how this is different than one in signer-test?

packages/core/src/Signer.ts
```diff
@@ -290,6 +284,8 @@ export class Signer {
 	static sign(request, access_info, service_info = null) {
 		request.headers = request.headers || {};
 
+		if (request.body && !request.data) request.data = request.body;
```
ashika01, line 287: curious: what benefit does we get from overriding the request.data with body over just throwing an error to make the usage uniform across the lib?

ashika01 (changes requested):

packages/core/tslint.json
```diff
@@ -17,7 +17,7 @@
 		"object-literal-shorthand": true,
 		"no-eval": true,
 		"space-before-function-paren": [
-			true,
+			false,
```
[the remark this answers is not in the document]
chrisbonifacio: we were running into issues committing locally due to the linter, flipping back to true directly from this PR

packages/core/src/Signer.ts
chrisbonifacio: The customer that requested this change did not express a preference, so throwing an error should be fine and makes sense to stay consistent with the rest of the library. We will do that instead!

packages/core/__tests__/SignerBody-test.ts
chrisbonifacio: We will update the test to check for whether there is a `body` and not `data` and assert that the proper error is thrown.

amehi0index:
I am creating a new pull request for this issue. Hope that's okay.

abdallahshaban557:
of course @amehi0index ! We appreciate your contributions!!

PR #10115, opened 2022-07-26
Update Errors.ts

cargallo:
Change 'AuthError' for type of error. Otherwise the error type is overwrited with 'AuthError' and this makes that becames imposible to translate error messages based on error code in the client.


abdallahshaban557:
Hi @cargallo - we are in the middle of re-writing our Auth category including providing better type support. This change has been noted internally on our end, and we will address it as part of our refactor. I am closing this PR for now - thank you for your contribution!

PR #10136, opened 2022-07-29
GH-9824 - feat: PubSub Connection state tracking for MQTT and IoT providers

stocaaro:
#### Description of changes
Add Hub events for MQTT connection state changes. Building on the work to [add connection states for AppSync](https://github.com/aws-amplify/amplify-js/pull/10063), this applies the same events to the MQTT interface, which will include the IoT provider.

Changes:
- Add ConnectionStateMonitor and events to the MQTT provider
- Refactor the test helper to expose the hub event tracking for use with the MQTT tests
- Add MQTT tests to cover the new behavior

#### Issue #, if available
#9824

- Unit tests where added
- Manually tested against a sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -149,6 +174,7 @@ export class MqttOverWSProvider extends AbstractPubSubProvider {
 	public async newClient({ url, clientId }: MqttProviderOptions): Promise<any> {
 		logger.debug('Creating new MQTT client', clientId);
 
+		this.connectionStateMonitor.record(CONNECTION_CHANGE.OPENING_CONNECTION);
```
elorzafe, line 177: on L196 is missing `record`

elorzafe: on L188 could be something similar to the `ka` messages?

elorzafe (changes requested):
It seems we could add more events on the monitor, we can chat more offline

I think is very close
Thanks @stocaaro 🌮

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: Alright. The onDisconnect record event is added.

Re: Keep alive. The only place in the [paho public surface](https://github.com/eclipse/paho.mqtt.javascript/blob/master/src/paho-mqtt.js) that keep alive appears to be logged is in the `trace` behavior.  It looks like this is available in the public interface if you receive it through the `traceFunction` callback. This isn't offered the the TS interface, but I made a run at using it and couldn't make it respond with the expected callback events.

I'm pretty sure I could turn this into something in another hour or two, but I'm off the end of my timebox and I think the better path would be to settle for not offering KA input for MQTT at this time.

packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -72,13 +78,32 @@ class ClientsQueue {
...
 
 export class MqttOverWSProvider extends AbstractPubSubProvider {
 	private _clientsQueue = new ClientsQueue();
+	private readonly connectionStateMonitor = new ConnectionStateMonitor();
 
 	constructor(options: MqttProviderOptions = {}) {
 		super({ ...options, clientId: options.clientId || uuid() });
+
+		// Monitor the connection health state and pass changes along to Hub
+		this.connectionStateMonitor.connectionStateObservable.subscribe(
```
elorzafe, line 95: What happens if the connection is closed, is this going to be emitting events right?

Shouldn't in that case this be only when there is a subscription open ?

packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -293,6 +330,10 @@ export class MqttOverWSProvider extends AbstractPubSubProvider {
 					this._clientIdObservers.get(clientId)?.delete(observer);
 					// No more observers per client => client not needed anymore
 					if (this._clientIdObservers.get(clientId)?.size === 0) {
+						this.connectionStateMonitor.record(
+							CONNECTION_CHANGE.CLOSING_CONNECTION
+						);
```
elorzafe, line 335: This should also remove the subscription for `connectionStateMonitor` right?

elorzafe (changes requested):
Sorry I didn't see this before. It seems `connectionStateMonitor` should be unsubscribe after the las subscription is unsubscribed

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: Good news. When [CONNECTION_FAILED or CLOSING_CONNECTION changes](https://github.com/aws-amplify/amplify-js/blob/main/packages/pubsub/src/utils/ConnectionStateMonitor.ts#L44-L48) are recorded, the monitor automatically [turns off the Reachability subscription](https://github.com/aws-amplify/amplify-js/blob/main/packages/pubsub/src/utils/ConnectionStateMonitor.ts#L136-L140).

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: Yep, since `CLOSING_CONNECTION` turns the `intended _connection_state` to disconnected, this will automatically turn off the Reachability sub.

packages/pubsub/src/Providers/MqttOverWSProvider.ts
elorzafe: oh nice, ok cool

packages/pubsub/src/Providers/MqttOverWSProvider.ts
elorzafe: Cool!

elorzafe (approved):
Great work

Thanks @stocaaro 🌮 🎉 🏅

packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -168,17 +194,27 @@ export class MqttOverWSProvider extends AbstractPubSubProvider {
...
+				onFailure: () => {
+					reject();
+					this.connectionStateMonitor.record(
+						CONNECTION_CHANGE.CONNECTION_FAILED
+					);
+				},
 			});
 		});
 
+		this.connectionStateMonitor.record(
```
Samaritan1011001, line 214: Should this be inside the onSuccess block of the connect call? Does it not make a difference?

Samaritan1011001, line 214: Should this be inside the onSuccess block of the connect call? Does it not make a difference?

packages/pubsub/__tests__/PubSub-unit-test.ts
```diff
@@ -306,6 +316,123 @@ describe('PubSub', () => {
...
+			test('test happy case connect -> disconnect cycle', async () => {
+				const pubsub = new PubSub();
+
+				const awsIotProvider = new AWSIoTProvider({
+					aws_pubsub_region: 'region',
+					aws_pubsub_endpoint: 'wss://iot.mymockendpoint.org:443/notrealmqtt',
+				});
+				pubsub.addPluggable(awsIotProvider);
+
+				const sub = pubsub.subscribe('topic', { clientId: '123' }).subscribe({
```
Samaritan1011001, line 351: Curious why the subscribe calls are chained here.

Samaritan1011001 (approved):
Looking good. Just two clarification, non-actionable questions. 
Thank you.

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: We want this to fire as soon as the connection is available for use. In the above `.connect`, we could add this in the `onSuccess` handler.  Since this resolves the outer promise that is being awaited, having the state recorded right after the await keeps this in the right place without burying it down in a callback, but you're right to call out that the code doesn't immediately tell you that this is the moment where the connection has finally been established.

packages/pubsub/__tests__/PubSub-unit-test.ts
stocaaro: The PubSub API uses the `.subscribe` keyword to setup an `Observable` where these `zen-observable` objects also have a `.subscribe` method that allows you to setup independent callback for `next` events, `complete` and `error` events. Both `error` and `complete` events close the observable, so will only ever be called once.

Samaritan1011001: Nice! thanks for both the explanations.

PR #10140, opened 2022-07-31
async adapter bug fixes; observe and observeQuery bug fixes; empty key value bug fix

david-mcafee:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/datastore/src/datastore/datastore.ts
+++ b/packages/datastore/src/datastore/datastore.ts
@@ -63,6 +63,7 @@ import {
 	__modelMeta__,
 	isIdentifierObject,
 	AmplifyContext,
+	isModelAttributePrimaryKey,
 } from '../types';
 import {
 	DATASTORE,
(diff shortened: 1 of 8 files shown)


packages/datastore/src/datastore/datastore.ts
```diff
@@ -1347,6 +1357,13 @@ class DataStore {
 					criteria
 				);
 			} else {
+				// observeQuery does not accept OL, okay to remove, Manuel?
```
david-mcafee, line 1360: Discussed with Manuel offline, will remove

packages/datastore/src/util.ts
```diff
@@ -31,8 +31,13 @@ import {
 	IndexesType,
 } from './types';
 import { WordArray } from 'amazon-cognito-identity-js';
+
 const ID = 'id';
 
+export const errorMessages = {
+	idEmptyString: 'Field of type `ID` cannot be an empty string',
```
iartemiev, line 38: This should read: "An index field cannot contain an empty string value". 

The field's type is not important, we can have an optional field with type `ID`. But any field (regardless of type) that's part of a primary key or other index cannot be empty.

iartemiev (changes requested):
Looks good! Just a couple tiny call outs

packages/datastore/src/util.ts
david-mcafee: Updated!

iartemiev (approved):
LGTM! 👍🏻

dpilch (approved):
LGTM, just one non-blocking style nit

packages/datastore/src/datastore/datastore.ts
```diff
@@ -987,6 +989,14 @@ class DataStore {
 
 		const modelDefinition = getModelDefinition(modelConstructor);
 
+		for (const key of extractPrimaryKeyFieldNames(modelDefinition)) {
```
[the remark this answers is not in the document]
dpilch: nvm, looks like this has changed since I started the review.

PR #10161, opened 2022-08-05
feat(@aws-amplify/storage): pagination for storage list api using Token

kvramyasri7:
#### Description of changes
To enable users to paginate through storage using List api. Enabled users to pass a pageToken to the list Api so that they can get a next set of files.

#### Issue #, if available

#7133 
#6838 

- Unit test added for the change
- Tested change in the sample app

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -711,56 +712,66 @@ export class AWSS3Provider implements StorageProvider {
...
+					list.results.push(...listResult.results);
+					if (listResult.nextPageToken)
+						continuationToken = listResult.nextPageToken;
+				} while (listResult.nextPageToken);
 			} else {
-				maxKeys < 1000 || typeof maxKeys === 'string'
-					? (params.MaxKeys = maxKeys)
+				pageSize < 1000 || typeof pageSize === 'string'
+					? (params.MaxKeys = pageSize)
 					: (params.MaxKeys = 1000);
```
stocaaro, line 757: My brain wants this block of code structured differently. Saying I don't like the aesthetic of it isn't helpful, so let me see if I can step into what I might propose as better and why.

```js
params.maxKeys = 1000
if (typeof pageSize === 'number') {
  params.maxKeys = Math.min(params.maxKeys, pageSize);
}
```

Now, what do I like about this, better than what you've got here?
- Ternaries shouldn't span multiple lines, ideally, but with multiple conditions this would be too much for one line.
- Guarding for `number` as opposed to against `string` seems more inline with what is happening here.
- I like using 1000 as the default value and only overriding it when we find an applicable override.

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -711,56 +712,66 @@ export class AWSS3Provider implements StorageProvider {
...
 				Bucket: bucket,
 				Prefix: final_path,
 				MaxKeys: 1000,
+				ContinuationToken: pageToken,
 			};
-			if (maxKeys === 'ALL') {
+			if (pageSize === 'ALL' || pageSize === undefined) {
 				do {
-					params.ContinuationToken = token;
+					params.ContinuationToken = continuationToken;
```
stocaaro, line 747: can we get away with not having a different `continuationToken` variable? Instead, just referring to `listResult?nextPageToken`?

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -711,56 +712,66 @@ export class AWSS3Provider implements StorageProvider {
...
 					params.MaxKeys = 1000;
```
stocaaro, line 748: This never changes, right?  Do we need to reset it to 1000 each loop?

packages/core/src/Signer.ts
```diff
@@ -290,6 +284,12 @@ export class Signer {
 	static sign(request, access_info, service_info = null) {
 		request.headers = request.headers || {};
 
+		if (request.body && !request.data) {
```
stocaaro, line 287: This signer stuff must have gotten mixed up in our next merge stuff.  Can you see if this was removed and we're adding it back vs the next branch was missing this diff?

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
```diff
@@ -955,26 +956,63 @@ describe('StorageProvider test', () => {
...
+		};
+		const inputResult = {
+			Bucket: 'bucket',
+			Prefix: 'public/listALLResultsPath',
+			MaxKeys: 1000,
+			ContinuationToken: 'TEST_TOKEN',
+		};
+		let listResult = [];
+		let pages = 3;
+		for (let i = 0; i < pages; i++) listResult.push(resultObj);
```
stocaaro, line 974: This testing stuff looks a bit odd, loops in open describe blocks. Long-ish tests with lots of layered behavior. There's got to be a better way, but its hard for me to think of exactly which direction to explore for improvement. Can we pair on this some today or Monday?

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -711,56 +712,66 @@ export class AWSS3Provider implements StorageProvider {
...
-				if (maxKeys > 1000)
+				list.results.push(...listResult.results);
+				list.hasNextPage = listResult.hasNextPage;
+				listResult.nextPageToken
+					? (list.nextPageToken = listResult.nextPageToken)
+					: (list.nextPageToken = null);
+				if (pageSize > 1000)
 					logger.warn(
-						"maxkeys can be from 0 - 1000 or 'ALL'. To list all files you can set maxKeys to 'ALL'."
+						"makeys can be from 0 - 1000 or 'ALL'. To list all files you can set maxKeys to 'ALL'."
```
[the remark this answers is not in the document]
kvramyasri7: This is a good catch. Thank you.

packages/storage/src/providers/AWSS3Provider.ts
kvramyasri7: Yeah, we can place it out side of the loop. Thanks for the suggestion.

packages/core/__tests__/Signer-test.ts
```diff
@@ -73,3 +73,60 @@ describe('Signer test', () => {
...
+		};
+
+		const spyon = jest
+			.spyOn(Date.prototype, 'toISOString')
+			.mockReturnValueOnce('0');
+
+		expect(() => {
+			Signer.sign(request_data, access_info, {
+				service: 'aservice',
+				region: 'aregion',
```
elorzafe, line 126: I think this change is because the other branch needs rebasing right?

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -711,56 +711,67 @@ export class AWSS3Provider implements StorageProvider {
...
+			let continuationToken: string = pageToken;
 			let listResult: S3ProviderListOutputWithToken;
 			const params: ListObjectsV2Request = {
 				Bucket: bucket,
 				Prefix: final_path,
 				MaxKeys: 1000,
+				ContinuationToken: pageToken,
 			};
-			if (maxKeys === 'ALL') {
+			console.log('Helloo  from list');
```
elorzafe, line 744: Need to remove this

packages/storage/src/types/AWSS3Provider.ts
```diff
@@ -96,17 +96,19 @@ export type S3ProviderRemoveConfig = CommonStorageOptions & {
...
+	nextPageToken?: string;
+	hasNextPage: boolean;
 };
 
 export type S3ProviderRemoveOutput = DeleteObjectCommandOutput;
 
 export type S3ProviderListConfig = CommonStorageOptions & {
 	bucket?: string;
-	maxKeys?: number | 'ALL';
+	pageSize?: number | 'ALL';
```
elorzafe, line 108: What happens if `pageSize` has changed after first `list` operation? Is that possible?

elorzafe: I would have a type with `pageSize` (the first list), then another with `pageToken`.

I would also remove `ALL` it seems to be the default anyway. And also feels odd that a page is `ALL`.

elorzafe (changes requested):
it seems `ALL` vs list page is a little unclear on the type level. We can discuss that more offline

packages/storage/src/providers/AWSS3Provider.ts
kvramyasri7: Thanks. Done.

packages/storage/src/types/AWSS3Provider.ts
stocaaro: Agree that we can drop ALL if this is the default behavior.

For `pageSize` I think we have tested that it is possible to change the pageSize after the first operation, given a token.  Since without having an explicit page size, we would need to retain it somewhere, are you thinking that we would include this as part of the nextPageToken so that we know the pagesize to keep things consistent moving forward? I can see customer usecases where the end user might change the pagesize mid-pagination and expect not to be reset. Presumably if the developer doesn't want to enable this, they will be able to keep the pagesize static. Are you sure about removing the pageSize input when there is a token?

packages/storage/src/providers/AWSS3Provider.ts
kvramyasri7: Yeah we can do that. Thanks for the suggestion.

packages/storage/src/providers/AWSS3Provider.ts
kvramyasri7: Removing string or number as we are removing 'ALL' in total. Thanks for the suggestion.

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
kvramyasri7: Done. Thanks for the suggestion.

packages/core/src/Signer.ts
kvramyasri7: Done. Thanks

packages/core/__tests__/Signer-test.ts
kvramyasri7: Removed the Changes. Done. Thanks.

kvramyasri7:
Closing this PR and opened a new PR with the pagination changes.

PR #10200, opened 2022-08-13
fix(pubsub): Connection Ack verification bug

stocaaro:
#### Description of changes

The connection updates for API show disconnected after 15 seconds when the connection is working just fine. Thats a problem.  Turns out that connection ack verification is working with old state. This fixes it so that it will use current state.

#### Issue #, if available
#9824

- Unit tests pass
- Verified on sample app, confirming that a disconnect no longer happens after 15 seconds

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


stocaaro:
Talking with Francisco, a fixed bug should include a test that verifies the behavior was broken. I've added a test that fails when I revert the fix.

While adding this, I ran into side effects between tests that I have fixed by tightening connection failure timeouts for two tests.

Also found a wording log error that has been fixed.

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -662,7 +663,43 @@ describe('AWSAppSyncRealTimeProvider', () => {
...
+					expect.assertions(2);
+					await replaceConstant('CONNECTION_INIT_TIMEOUT', 20, async () => {
+						const observer = provider.subscribe('test', {
+							appSyncGraphqlEndpoint: 'ws://localhost:8080',
+						});
+
+						const subscription = observer.subscribe({ error: () => {} });
+
+						await fakeWebSocketInterface?.readyForUse;
+						Promise.resolve();
```
elorzafe, line 676: Why do you need this? Is something just to make jest happy?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -747,20 +747,20 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
-						setTimeout(checkAckOk.bind(this, ackOk), CONNECTION_INIT_TIMEOUT);
-					}
+						const checkAckOk = (ackOk: boolean) => {
+							if (!ackOk) {
+								this.connectionStateMonitor.record(
+									CONNECTION_CHANGE.CONNECTION_FAILED
+								);
+								rej(
+									new Error(
+										`Connection timeout: ack from AWSRealTime was not received after ${CONNECTION_INIT_TIMEOUT} ms`
```
elorzafe, line 757:
```suggestion
										`Connection timeout: ack from AWSAppSyncRealTime was not received after ${CONNECTION_INIT_TIMEOUT} ms`
```

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -690,7 +727,7 @@ describe('AWSAppSyncRealTimeProvider', () => {
 							'error on bound ',
 							expect.objectContaining({
 								message: expect.stringMatching(
-									'Connection timeout: ack from AWSRealTime'
+									'Connection timeout: ack from AWSRealTime was not received after'
```
elorzafe, line 730:
```suggestion
									'Connection timeout: ack from AWSAppSyncRealTime was not received after'
```

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -662,7 +663,43 @@ describe('AWSAppSyncRealTimeProvider', () => {
...
+							CS.Connecting
+						);
+
+						// Watching for raised exception to be caught and logged
+						expect(loggerSpy).not.toBeCalledWith(
+							'DEBUG',
+							'error on bound ',
+							expect.objectContaining({
+								message: expect.stringMatching(
+									'Connection timeout: ack from AWSRealTime was not received after'
```
elorzafe, line 695:
```suggestion
									'Connection timeout: ack from AWSAppSyncRealTime was not received after'
```

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -662,7 +663,43 @@ describe('AWSAppSyncRealTimeProvider', () => {
...
+						// Wait no less than 20 ms
+						await delay(20);
+
+						// Wait until the socket is automatically disconnected
+						await expect(fakeWebSocketInterface?.currentConnectionState).toBe(
+							CS.Connecting
+						);
+
+						// Watching for raised exception to be caught and logged
+						expect(loggerSpy).not.toBeCalledWith(
```
elorzafe, line 690: If you revert the fix can you see the test failing?

elorzafe (commented):
Overall looks good, just a couple of suggestions

packages/pubsub/__tests__/AWSAppSyncRealTimeProvider.test.ts
stocaaro: Correct, I tested this test without the fix and it fails. The failure is against the prior expect where the state is Disconnected.

In this failure mode, becuase we're working with old ackOk state, the outer promise is resolved, so can't be rejected, but we still maintain the connection monitor before we attempt and fail to fire the reject, which triggers the message log.  Technically, without testing the connection state, this test would pass both ways as the resolved promise prevents this message from being logged.

elorzafe (approved):
Looks good!

Thanks @stocaaro 🎉 🏅

katiegoines (approved):

Issue #10124, opened 2022-07-28
AppSync Subscription connection closing and unable to reestablish


cdellinger:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication, GraphQL API

### Amplify Categories

auth, api

### Environment information

<details>

```
# Put output below this line
  System:
    OS: macOS 12.0.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 487.03 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.0 - /usr/local/bin/node
    npm: 8.1.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 103.0.5060.134
    Safari: 15.1
  npmPackages:
    @aws-amplify/ui-react: ^3.2.1 => 3.2.1
    @aws-amplify/ui-react-internal:  undefined ()
    @aws-amplify/ui-react-legacy:  undefined ()
    @emotion/react: ^11.9.3 => 11.9.3
    @emotion/styled: ^11.9.3 => 11.9.3
    @mui/icons-material: ^5.8.4 => 5.8.4
    @mui/material: ^5.9.2 => 5.9.2
    @mui/x-date-pickers: ^5.0.0-beta.2 => 5.0.0-beta.2
    @testing-library/jest-dom: ^5.16.4 => 5.16.4
    @testing-library/react: ^13.3.0 => 13.3.0
    @testing-library/user-event: ^13.5.0 => 13.5.0
    @types/jest: ^27.5.2 => 27.5.2
    @types/node: ^16.11.43 => 16.11.43
    @types/react: ^18.0.15 => 18.0.15
    @types/react-dom: ^18.0.6 => 18.0.6
    @welldone-software/why-did-you-render: ^7.0.1 => 7.0.1
    aws-amplify: ^4.3.28 => 4.3.28
    date-fns: ^2.29.1 => 2.29.1
    randomcolor: ^0.6.2 => 0.6.2
[...]

Issue #9749, opened 2022-03-30
Unable to reconnect AppSync after connection interruption (web) or app going into background (ios/ android)


CSHaze:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Angular

### Amplify APIs

Authentication, GraphQL API, DataStore, Storage

### Amplify Categories

auth, storage, function, api, hosting

### Environment information

<details>

```

  System:
    OS: Windows 10 10.0.19042
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
    Memory: 4.43 GB / 31.96 GB
  Binaries:
    Node: 16.13.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.10 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.4.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 100.0.4896.60
    Edge: Spartan (44.19041.1266.0), Chromium (99.0.1150.55)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @angular-devkit/build-angular: ~13.2.1 => 13.2.5
    @angular-eslint/builder: 13.0.1 => 13.0.1
    @angular-eslint/eslint-plugin: 13.0.1 => 13.0.1
    @angular-eslint/eslint-plugin-template: 13.0.1 => 13.0.1
    @angular-eslint/schematics: 13.0.1 => 13.0.1
    @angular-eslint/template-parser: 13.0.1 => 13.0.1
    @angular/cli: ~13.2.1 => 13.2.5
    @angular/common: ~13.2.0 => 13.2.5
    @angular/common/http:  undefined ()
    @angular/common/http/testing:  undefined ()
    @angular/common/testing:  undefined ()
    @angular/common/upgrade:  undefined ()
    @angular/compiler: ~13.2.0 => 13.2.5
    @angular/compiler-cli: ~13.2.0 => 13.2.5
[...]

Issue #8513, opened 2021-06-28
[Feedback] Amplify subscriptions stop when app goes to sleep. Don't resubscribe


spmatthews03:
**Page**: [`/guides/q/platform/js`](https://docs.amplify.aws/guides/q/platform/js)

**Feedback**:

I'm trying to have a simple subscription and it works until the phone goes to sleep, or if the app is in the background for too long. I would like a way to resubscribe to the data so my UI and my server are not out of sync.

chrisbonifacio:
@spmatthews03 are you using `API` or `DataStore` in your app?

spmatthews03:
@chrisbonifacio API. I semi fixed it buy a listener on AppState and if subscriptions are null then I reload ALL subscriptions. just 8 of them soits not terrible and only if their componentn is mounted

dave-vazquez:
@spmatthews03 I'm currently running into the same issue. Can you share your implementation for listening to AppState, checking if subscriptions are null and reloading all subscriptions?

chrisbonifacio:
@spmatthews03 Nice workaround! I'm going to find out if this issue is expected behavior so I know if it's either a bug or a feature request that can be built in.

dave-vazquez:
> @spmatthews03 Nice workaround! I'm going to find out if this issue is expected behavior so I know if it's either a bug or a feature request that can be built in.

@spmatthews03 Thank you, this would be super-helpful if you could find out.

chrisbonifacio:
Hey @spmatthews03 and @dave-vazquez, so I did talk to the team about this. At the moment re-establishing connections to subscriptions is only handled by Amplify in the DataStore category. Unfortunately, this functionality is not yet available out of the box from the API category as explained here https://github.com/aws-amplify/amplify-js/issues/7057#issuecomment-716843165

That being said, until we do implement subscription reconnection I'm going to label this as a feature request. I'm personally interested in this and I know there are examples of how to handle this out there in the wild so I'll try to find that and post it here as well as make sure the docs are updated with a code snippet that achieves that.

dave-vazquez:
Thanks @chrisbonifacio ! Very much apprrciated

CSHaze:
Is this still being worked on?

rudalson:
Hello. I need to solve this issue. It happens to frequently in ios. @spmatthews03  Can you also share your workaround to me?

stocaaro:
[...]

Issue #4459, opened 2019-11-25
PubSub Javascript: How to detect websocket closed and reconnect?


markmacau:
** Which Category is your question related to? **
PubSub Javascript(vue)

** What AWS Services are you utilizing? **
PubSub, IoT, Cognito(Unauth)

** Provide additional details e.g. code snippets **
Somehow my MQTT connection drops even after successfully subscribed a topic and send/receive messages.

1) How can I detect/monitor MQTT connection status? In other word, How can I access AWSIoTProvider created by Amplify.addPluggable?

2) After MQTT disconnect and even I unsubscribe then subscribe topic again, I cannot publish message any more with error "AMQJS0011E Invalid state not connected". 

So far only solution is to reload page. Is there any better way to recover session?

3) Above "AMQJS0011E Invalid state not connected." error can be a trigger to reconnect socket. How to set eventlisteners on such errors?

escar:
Having this issue too, any way to overcome this problem?

houmark:
It's really starting to become a problem for us, that we cannot in any way check the current connection status and reconnect in any reliable way.

The simplest use case is a browser tab staying open while the user putting her computer to sleep, then resumes work without refreshing the UI/tab. The UI will stop providing realtime updates, and there's no way we have found, where we can even request a connection status.

kevin-mitchell:
Also interested in tracking this!

dcoellarb:
any updates on this?

nitindatta:
anyone found workaround for this.

pieterbergmans:
This seems to happen quite often.  Any solutions?

stocaaro:
Some good news. Detecting socket status changes [just got a whole lot easier](https://www.google.com/maps/d/edit?mid=1R7JcHfuNwVnet-g6OBs1XdMU5h7s8Rw&usp=sharing) with new Hub messages.

Work is inflight to [add automated reconnection](https://github.com/aws-amplify/amplify-js/pull/10235).

nubpro:
> Some good news. Detecting socket status changes [just got a whole lot easier](https://www.google.com/maps/d/edit?mid=1R7JcHfuNwVnet-g6OBs1XdMU5h7s8Rw&usp=sharing) with new Hub messages.
> 
> Work is inflight to [add automated reconnection](https://github.com/aws-amplify/amplify-js/pull/10235).

might want to check on the link you posted...

stocaaro:
[...]

PR #10235, opened 2022-08-19
feat(pubsub): Automated reconnect for IoT and API

stocaaro:
#### Description of changes
Automated reconnect for IoT and API.

This change doesn't alter the interface for either IoT or API, however when the connection is lost or disrupted it will automatically attempt to re-form the connection and resubscribe for new events.

This changes what behavior developers can expect and may conflict with workarounds. Based on this, the change will be merged into the `next` tag for inclusion in the next major version.

#### Expected behavior
- When an existing connection is severed by sleep/backgrounding/airplane mode, it will reconnect 5 seconds after the app/site is resumed.
- When the network is unavailable past socket expiration and JS can detected network reachability, when the network recovers, then the reconnection will be restarted 5 seconds later.
- When the network is unavailable past socket expiration and JS cannot detected network reachability, then reconnection will be attempted every 60 seconds until the network recovers and it can succeed.
- When the connection is started while the network is unavailable, the reconnection behavior will be used to recover when the network becomes available.

#### Bugs fixed
- Sending messages through MQTT would initialize and cache the connection, but didn't integrate with the observables that started the connection in ways that caused issues  and/or interfered with the reconnection process. Refactored to allow publish to use the connection, but not to create a missing connection.
- MQTT connection caching retaining failed, errored connections. Refactored to more proactively clear the cache of failed connections.

#### Issue #, if available
#10124
#9824
#9749
#8513
#4459
#3039
#2692
#2372
#1016

- Manually tested and retested in both Safari, Chrome and Firefox for the Mac then also Chrome and Brave on Android
- Unit tests added
- Integ testing needed

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -800,14 +842,16 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
+				`Authentication type ${String(authenticationType)} not supported`
+			);
 			return '';
 		} else {
 			const handler = headerHandler[authenticationType];
 
 			const { host } = url.parse(appSyncGraphqlEndpoint ?? '');
 
-			logger.debug(`Authenticating with ${authenticationType}`);
+			logger.debug(`Authenticating with ${String(authenticationType)}`);
```
elorzafe, line 854: Why do you need this?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -525,14 +574,7 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
 		logger.debug(`Disconnect error: ${msg}`);
-		this.subscriptionObserverMap.forEach(({ observer }) => {
-			if (observer && !observer.closed) {
-				observer.error({
-					errors: [{ ...new GraphQLError(msg) }],
-				});
-			}
-		});
-		this.subscriptionObserverMap.clear();
+
```
elorzafe, line 577: Would this impact the connection state change? Like reconnecting status?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -157,26 +182,50 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
-							},
-						],
-					});
-					this.connectionStateMonitor.record(CONNECTION_CHANGE.CLOSED);
-					observer.complete();
+				const startSubscription = () => {
+					if (!subscriptionStartActive) {
+						subscriptionStartActive = true;
+						const startSubscriptionPromise =
+							this._startSubscriptionWithAWSAppSyncRealTime({
```
elorzafe, line 191: What would happen if when start you have a network issue?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -157,26 +182,50 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
+								options,
+								observer,
+								subscriptionId,
+							}).catch<any>(err => {
```
elorzafe, line 195: Is it ok to do a `catch` when you are assigning  the Promise? It reads a little odd

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -157,26 +182,50 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
+								this.connectionStateMonitor.record(CONNECTION_CHANGE.CLOSED);
+								observer.complete();
+							});
+						startSubscriptionPromise.finally(() => {
+							subscriptionStartActive = false;
+						});
+					}
+				};
+
+				startSubscription();
```
elorzafe, line 214: Maybe this should happen after the reconnect subscription is ready?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -157,26 +182,50 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
+
+				let reconnectSubscription: ZenObservable.Subscription;
+
+				// Add an observable to the reconnection list to manage reconnection for this subscription
+				reconnectSubscription = new Observable(observer => {
+					this.reconnectObservers.push(observer);
+				}).subscribe(() => {
+					startSubscription();
```
elorzafe, line 222: Do we want to add some sort of exponential back off?

elorzafe (commented):
More questions than feedback.

We can chat more offline.

Thanks for doing this @stocaaro !

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
stocaaro: Alright. I found another instance (not my computers strange Safari). On the Brave mobile browser, when I stop the network by either disabling wifi/mobile or turning on airplane mode, this places the the connection state in `ConnectionDisrupted` and attempts to reconnect immediately.  Later when the network returns, the status doesn't change from ConnectionDisrupted and reconnect is never reattempted.

This is rooted in the same issue I was seeing in my buggy safari, without Reachability, the network state remains on at all times. As a result, when disruption occurs, reconnection is attempted immediately.

Is we discussed leading to this comment, I recommend adding a periodic (once per minute) reconnection attempt after `ConnectionDisrupted` is observed. The interval will be reset if a connection is formed or a new `ConnectionDisrupted` event occurs. This guarantees that a browser that doesn't support Reachability will still get a connection back.

Thinking this through makes me want to diagram the expected recovery pattern for different browsers against a timeline for clarity.

packages/pubsub/src/utils/ReconnectionMonitor.ts
```diff
@@ -0,0 +1,50 @@
+import { Observer } from 'zen-observable-ts';
+import { RECONNECT_DELAY, RECONNECT_INTERVAL } from '../Providers/constants';
+
+export enum ReconnectEvent {
+	RECONNECT = 'RECONNECT',
+	HALT_RECONNECT = 'HALT_RECONNECT',
+}
+
+export class ReconnectionMonitor {
```
stocaaro, line 9: Add basic docstrings for this new class

packages/pubsub/src/utils/ReconnectionMonitor.ts
```diff
@@ -0,0 +1,50 @@
...
+				}, RECONNECT_DELAY);
+			}
+		}
+
+		if (event === ReconnectEvent.HALT_RECONNECT) {
+			if (this.reconnectIntervalId) {
+				clearInterval(this.reconnectIntervalId);
+				this.reconnectIntervalId = undefined;
+			}
+		}
```
stocaaro, line 42: On Android Chrome, when a connection is started then the phone put into airplane mode, then both IoT and AppSync connections flap between `ConnectionDisruptedPendingNetwork` and `Connecting` every minute.

To start this sequence `ConnectionDisrupted` precedes the first `ConnectionDisruptedPendingNetwork` starting the reconnect cycle. Reconnect is initially triggered, then the state moves on before the setTimeout happens. Since moving on clears the interval, but not the timeout, reconnect proceeds and this becomes a minute over minute cycle.

Halting reconnection clears the 60 second interval, but not the setTimeout. Clearing the setTimeout will properly halt reconnection for this case.

packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -50,21 +58,21 @@ export interface MqttProviderOptions extends ProviderOptions {
...
+		if (cachedPromise) return cachedPromise;
 
 		if (clientFactory) {
 			const newPromise = clientFactory(clientId);
-
 			this.promises.set(clientId, newPromise);
-
+			newPromise.then(v => this.promises.set(clientId, v));
+			newPromise.catch(v => this.promises.delete(clientId));
+			newPromise.finally();
```
stocaaro, line 72: Left over finally?

packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -50,21 +58,21 @@ export interface MqttProviderOptions extends ProviderOptions {
...
-			return cachedPromise;
-		}
+		if (cachedPromise) return cachedPromise;
 
 		if (clientFactory) {
 			const newPromise = clientFactory(clientId);
-
 			this.promises.set(clientId, newPromise);
-
+			newPromise.then(v => this.promises.set(clientId, v));
```
stocaaro, line 70: I think the important part is the delete on catch. Are we sure we need to replace the promise with the value?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
stocaaro: I have been testing with 2 browsers that don't [honor Reachability](https://caniuse.com/online-status). I've updated the implementation to support this case by re-attempting reconnection every 60 seconds until an event outside the reconnection cycle is observed. For most browsers, this code goes unexercised and reconnection follows 5 seconds after recovery. In the less common cases where reachability isn't working reconnection will happen within 60 seconds of recovery using the interval.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -157,26 +187,44 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
 				observer.complete();
 			} else {
+				let subscriptionStartActive = false;
 				const subscriptionId = uuid();
-				this._startSubscriptionWithAWSAppSyncRealTime({
-					options,
-					observer,
-					subscriptionId,
-				}).catch<any>(err => {
-					observer.error({
```
stocaaro, line 166: These errors should be debugs instead.

stocaaro: This is a recoverable error, not a big scary error.

stocaaro:
403 erros shouldn't be retried. Also if the subscription is malformed / invalid then we should not retry.

packages/pubsub/src/utils/ConnectionStateMonitor.ts
```diff
@@ -84,6 +95,9 @@ export class ConnectionStateMonitor {
 	 * Turn network state monitoring on if it isn't on already
 	 */
 	private enableNetworkMonitoring() {
+		// If no initial network state was discovered, stop trying
+		this._initialNetworkStateSubscription?.unsubscribe();
```
elorzafe, line 99: I am having trouble understanding this ☕ 

L78 tries to get the initial state and then after it gets the status is unsubscribe, I think we chatted about this and that it didn't return the status immediately? After looking on ReachabilityMonitor code https://github.com/aws-amplify/amplify-js/blob/main/packages/core/src/Util/Reachability.ts#L21 it seems it does. 

I am probably missing something here

packages/pubsub/src/utils/ReconnectionMonitor.ts
```diff
@@ -0,0 +1,64 @@
+import { Observer } from 'zen-observable-ts';
```
elorzafe, line 1: I am having a hard time to take this dependency out from core, is it necessary or can be defined "by hand"?

packages/pubsub/src/utils/ReconnectionMonitor.ts
```diff
@@ -0,0 +1,64 @@
...
+	 * Add reconnect observer to the list of observers to alert on reconnect
+	 */
+	addObserver(reconnectObserver: Observer<void>) {
+		this.reconnectObservers.push(reconnectObserver);
+	}
+
+	/**
+	 * Given a reconnect event, start the appropriate behavior
+	 */
+	record(event: ReconnectEvent) {
```
elorzafe, line 24: What happens if this gets called multiple times, can be a race condition?

packages/pubsub/src/utils/ReconnectionMonitor.ts
```diff
@@ -0,0 +1,64 @@
+import { Observer } from 'zen-observable-ts';
+import { RECONNECT_DELAY, RECONNECT_INTERVAL } from '../Providers/constants';
+
+export enum ReconnectEvent {
+	START_RECONNECT = 'START_RECONNECT',
+	HALT_RECONNECT = 'HALT_RECONNECT',
+}
+
+export class ReconnectionMonitor {
```
elorzafe, line 9: I would explain a little more why do we need this.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -98,23 +102,50 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
+							connectionState,
+						},
+						`Connection state is ${connectionState}`
+					);
+					this.connectionState = connectionState;
+
+					// Trigger reconnection when the connection is disrupted
+					if (connectionState === ConnectionState.ConnectionDisrupted) {
+						this.reconnectionMonitor.record(ReconnectEvent.START_RECONNECT);
+					} else if (connectionState !== ConnectionState.Connecting) {
```
elorzafe, line 129: I would prefer here to make an exhaustive check just in case we are missing a state in the future, maybe explicitly check the states that goes to halt state for reconnection and if we add another state in the future we could have a compilation error because it was not handle properly

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -676,9 +713,6 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 						logger.debug(`WebSocket connection error`);
 					};
 					newSocket.onclose = () => {
```
elorzafe, line 715: This was remove because of Reachability monitor?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -289,21 +338,36 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
+					logger.debug(`${CONTROL_MSG.CONNECTION_FAILED}: ${message}`);
+				}
+
+				const { subscriptionFailedCallback } =
+					this.subscriptionObserverMap.get(subscriptionId) || {};
+
+				// Notify concurrent unsubscription
+				if (typeof subscriptionFailedCallback === 'function') {
+					subscriptionFailedCallback();
+				}
```
elorzafe, line 370: Maybe I would export some of this into a separate function

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -157,26 +188,44 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
+				let reconnectSubscription: ZenObservable.Subscription;
+
+				// Add an observable to the reconnection list to manage reconnection for this subscription
+				reconnectSubscription = new Observable(observer => {
+					this.reconnectionMonitor.addObserver(observer);
+				}).subscribe(() => {
+					startSubscription();
 				});
 
+				startSubscription();
```
elorzafe, line 223: Could have a race condition? with L220?

elorzafe (commented):
Partial review (didnt review the tests) more questions than changes

packages/pubsub/src/utils/ConnectionStateMonitor.ts
stocaaro: From your guidance in the connection state monitoring PR, we don't want to leave reachability running for every provider in perpetuity, but only want to monitor when there are active subscriptions. So now, the monitor is turned on and off, however, this means we don't turn it on at the beginning to get the state, which we otherwise are defaulting to being on.

Now in this code, we need to know that the network is off before the first subscription is formed to get the right reconnection states for cold start. To accomplish this, we are turning it on, reading the current state and then turning it off again to be later restarted when the first subscription begins.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
stocaaro: You could, although in the case where it gets run twice `subscriptionStartActive` causes the second call to abort if it is run before the first call reaches its `finally` block.

Hypothetically of both calls execute their `if (!subscriptionStartActive) {` before the assignment on the next line, then they could be running in parallel, however I've read that this isn't how the JS GIL operates, although if this last week has taught me anything its that there is diversity in the ecosystem.

Sufficiently safe or is there a better locking mechanism to prevent any chance of double execution?

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
stocaaro: Yeah good call. This bit has gotten extra out of hand.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
stocaaro: This was triggering failed too early and getting in the way of reconnection.

It gets re-thrown on 815, retried then finally rejecting the promise on 676 with ultimately logged in the top level promise catch block I just refactored into its own method. Connection is recorded closed on 373 after I post the update.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
stocaaro: Maybe. Updating to have an exhaustive list for halt as requested.

packages/pubsub/src/utils/ReconnectionMonitor.ts
stocaaro: Sure. Adding a comment above this class to make it clear.

packages/pubsub/src/utils/ReconnectionMonitor.ts
stocaaro: Had a bug with this earlier and this pay enough attention to the retained `reconnectSetTimeoutId`. Fixing this, but assuming the thread isn't being dropped between the conditional block and assignment, this should be safe. Interested to hear your thoughts.

packages/pubsub/src/utils/ReconnectionMonitor.ts
stocaaro: This is imported 19 times across Pubsub, Core, DataStore, Storage and ApiGraphQL.  I'm sure there are alternatives, but I don't think it is in the scope of this change to remove this dependency.

stocaaro:
I spent several hours chasing the DataStore with API cold start issue and have confirmation that they aren't compatible. AppSync appears to only honor one subscription per unique query, last subscriber wins. As my demo app adds subscriptions normally, through luck the datastore subscribers are setup first and when I start the API subscriptions they replace the datastore ones leaving everything apparently working.

For the cold start case, Datastore doesn't attempt to start its subscriptions until the network exists while API sets up its subscriptions and internally waits for the network to connect. In this situation when the network returns the API subscriptions are overridden the the datastore subscriptions.

Long term, if deduplicate on queries and send messages to every subscription that is watching a given query, then it won't matter who's subscription identifier the payload was sent to. Until we support multiple subscribers in this, or some other, way, the behavior our tests are observing is expected.

katiegoines (approved):
Thank you for taking the time to go through my questions with me! LGTM

Issue #10275, opened 2022-08-31
Hosted-UI issue when using with electron


evanrbowers:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication

### Amplify Categories

auth

### Environment information

<details>

```
# Put output below this line
  System:
    OS: macOS 12.5
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 2.20 GB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 14.18.3 - ~/.nvm/versions/node/v14.18.3/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v14.18.3/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v14.18.3/bin/npm
  Browsers:
    Brave Browser: 104.1.42.97
    Chrome: 104.0.5112.101
    Firefox: 92.0
    Safari: 15.6
  npmPackages:
    @aws-amplify/auth: ^4.6.3 => 4.6.4 
    @aws-amplify/core: ^4.7.1 => 4.7.2 
    @date-io/date-fns: ^1.3.13 => 1.3.13 (2.14.0)
    @emotion/react: ^11.9.3 => 11.9.3 
    @emotion/styled: ^11.9.3 => 11.9.3 
    @fortawesome/fontawesome-svg-core: ^6.1.1 => 6.1.2 
    @fortawesome/free-brands-svg-icons: ^6.1.1 => 6.1.2 
    @fortawesome/pro-duotone-svg-icons: ^6.1.1 => 6.1.2 
    @fortawesome/pro-light-svg-icons: ^6.1.1 => 6.1.2 
    @fortawesome/pro-regular-svg-icons: ^6.1.1 => 6.1.2 
    @fortawesome/pro-solid-svg-icons: ^6.1.1 => 6.1.2 
[...]

PR #10276, opened 2022-08-31
fix-10275 - handle electron/app redirects for hosted-ui

evanrbowers:
#### Description of changes

There is in issue when using Hosted-UI with electron. It appears the library is trying to push the redirect url on the history. However, the history of the app is http:localhost but the redirect is an app name and path when using in electron. This errors and prevents some final logic to execute including setting the "amplify-signin-with-hostedUI" key. This then prevents the app from fully signing out via the hostedUI. The library is also not removing the "amplify-signin-with-hostedUI" key after signout.

This fix checks for http/https url and otherwise skips the adding to the history.  It also added the appropriate hook to remove "amplify-signin-with-hostedUI" on oauth signout. 

#### Issue #, if available

[10275](https://github.com/aws-amplify/amplify-js/issues/10275)

Built our own app and linked int the changes. Ran inside electron and outside.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -133,6 +133,9 @@ export class AuthClass {
 				case 'signOut':
 					this._storage.removeItem('amplify-signin-with-hostedUI');
 					break;
+				case 'oAuthSignOut':
+					this._storage.removeItem('amplify-signin-with-hostedUI');
+					break;
 				case 'cognitoHostedUI':
 					this._storage.setItem('amplify-signin-with-hostedUI', 'true');
 					break;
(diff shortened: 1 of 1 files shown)


packages/auth/src/Auth.ts
```diff
@@ -2480,11 +2483,17 @@ export class AuthClass {
 					currentUser.setSignInUserSession(session);
 
 					if (window && typeof window.history !== 'undefined') {
-						window.history.replaceState(
-							{},
-							null,
-							(this._config.oauth as AwsCognitoOAuthOpts).redirectSignIn
-						);
+						if (
```
tannerabread, line 2486: I believe this would break anything that's not electron, and would make the library tailored to electron only.

E.g. `https://yourdomain.com/login/` would now not use the replaceState method here

cwomack:
@evanrbowers, the corresponding issue that this PR was for was consolidated into #12737.  That issue was also closed as resolved per [this comment](https://github.com/aws-amplify/amplify-js/issues/12737#issuecomment-1906748970) regarding the fix being released in more recent version of Amplify.  

We'll close this PR because of this, but let us know if there are any questions.  Thanks!

PR #10293, opened 2022-09-06
Feat: add cleanup to build and scripts

israx:
#### Description of changes

Build cleanup 

This change doesn't alter the current build process. However, It adds a `tsconfig.json` file on every package that can be consumed by libraries that analyze the use of typescript (TS).

This change adds the possibility to add independent `compilerOptions` on every package.

#### Issue #, if available
none

- the build process works on every package after running yarn build

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amazon-cognito-identity-js/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+	"include": ["index.d.ts"]
```
stocaaro, line 2: Interesting. This is different than the others.  Looking at the `package.json` the build process is using bable directly instead of the build script.  Guessing this will not be used in that process or will not impact it.  Is this right?  Guessing this is added only to support the coverage work. Am I right or is something else happening here?

scripts/build.js
```diff
@@ -36,6 +36,9 @@ const pkgRollUpOutputFile = path.join(pkgRootPath, packageInfo.main);
 const es5TsBuildInfoFilePath = path.join(pkgTscES5OutDir, '.tsbuildinfo');
 const es6TsBuildInfoFilePath = path.join(pkgTscES6OutDir, '.tsbuildinfo');
 
+const defaultCompilerOptions =
+	require('./tsconfig.settings.json').compilerOptions;
```
stocaaro, line 40: I think in an earlier version you had the settings included directly in the package tsconfig files. What prompted relocating this to here only? Does this mean that the settings config isn't used as part of the coverage reporting?

packages/storage/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
+	"compilerOptions": { "composite": true, "strict": false },
+	"include": ["lib*/**/*.ts", "src"]
```
stocaaro, line 3: All of these are essentially the same. Any way to share some ore most of the config so that general changes can be made to all at once?  Not sure if this kind of dedup is worth it, happy to understand your thoughts.

stocaaro (commented):
This looks like many steps in the right direction. Interested to understand how the settings file works/doesn't work with the coverage reporting. My guess is that it isn't included, but maybe I'm not thinking it through or there's some something I don't fully understand.

packages/amazon-cognito-identity-js/tsconfig.json
israx: Yes. You are right. This will not impact the build process, however this is a route for "ts-coverage-report" to analyze the typescript code.

scripts/build.js
israx: Yes. I had the settings extended to every package but the build was not behaving well. The problem was that the extended object containing the "compilerOptions" was not being parsed correctly by build.js.

packages/storage/tsconfig.json
israx: Yes, basically all the tsconfig files have the same structure. However, we can add additional `compilerOptions` to each  package. 

At the end, the main purpose of having tsconfig files on every package is to enable the usage of "ts-coverage-report"  library and to add extra compiler options per package if desired.

packages/analytics/build.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
 const build = require('../../scripts/build');
+const tsconfig = require('./tsconfig.json');
```
stocaaro, line 4: Instead of require, could we use `convertCompilerOptionsFromJson`? My theory is that if we did so, this would composite the extends behavior and allow us to use the actual tsconfig instead of just the flat file, but I haven't tested this theory.

Like this:

https://github.com/plantain-00/type-coverage/blob/215c923540d161a285be2d47da6060ab1f981f6e/packages/core/src/tsconfig.ts#L14

israx: Thank you. I will make updates on that and also do more research in case it doesn't work out

elorzafe:
Did you get the TS report?

packages/amazon-cognito-identity-js/tsconfig.build.json
```diff
@@ -0,0 +1,7 @@
+{
+	"extends": "../tsconfig.settings.json",
+	"compilerOptions": {
+		"composite": true
```
stocaaro, line 4: What is composite doing for us here? Reading over [the documentation](https://www.typescriptlang.org/tsconfig#composite), I can't figure out why this was important to add.

israx: Yes, you are right, the composite flag doesn't do anything for us. It would add value if the entire file is used for compilation. In this case, we just need the **extends** flag. I just removed that flag on all `tsconfig.build` files

stocaaro (approved):
LGTM

stocaaro:
Closing, as I believe this was part of the changes now merged onto next.

Issue #10295, opened 2022-09-07
Amplify and PubSub imports break React applications since `>4.3.30`


louis-young:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

PubSub

### Environment information

I have tested this on MacOS, Windows and statically deployed to S3.

<details>

```
  System:
    OS: macOS 12.5.1
    CPU: (8) arm64 Apple M1
    Memory: 130.52 MB / 8.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.13.1 - /opt/homebrew/Cellar/node@16/16.13.1/bin/node
    Yarn: 1.22.17 - /opt/homebrew/bin/yarn
    npm: 8.1.2 - /opt/homebrew/Cellar/node@16/16.13.1/bin/npm
  Browsers:
    Chrome: 105.0.5195.52
    Safari: 15.6.1
  npmPackages:
    @aws-amplify/pubsub: ^4.5.2 => 4.5.2 
    aws-amplify: ^4.3.34 => 4.3.34 
    react: ^18.2.0 => 18.2.0 (18.1.0)
    react-dom: ^18.2.0 => 18.2.0 
    react-scripts: 5.0.1 => 5.0.1 
  npmGlobalPackages:
    gatsby-cli: 4.4.0
    n: 7.3.1
    npm: 8.5.5
    serverless: 2.59.0
    ts-node: 9.1.1
    typescript: 4.2.4
    yarn: 1.22.17
```
</details>

### Describe the bug

An error is thrown that prevents anything from being rendered in React applications when importing the `Amplify` and `AWSIoTProvider` classes. This has happened since `aws-amplify` version `>4.3.30`.

### Expected behavior

The application renders as it did before with `aws-amplify` version `<4.3.30`.

### Reproduction steps

1. Clone https://github.com/louis-young/aws-amplify-import-issue-investigation.
2. Run `npm install` in the project root.
3. Run `npm start` in the project root.
[...]

Issue #10287, opened 2022-09-02
[Typescript] Error: No pubsub module applied for subscription when attempting to create subscription.


Bluepuff71:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Vue

### Amplify APIs

GraphQL API

### Amplify Categories

auth, api

### Environment information

<details>

```
# Put output below this line
  System:
    OS: Windows 10 10.0.22000
    CPU: (32) x64 AMD Ryzen 9 5950X 16-Core Processor
    Memory: 116.57 GB / 127.90 GB
  Binaries:
    Node: 16.15.0 - E:\x64\nodejs\node.EXE
    npm: 8.18.0 - E:\x64\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (104.0.1293.70)
    Internet Explorer: 11.0.22000.120
  npmPackages:
    @aws-amplify/ui-vue: ^2.4.19 => 2.4.19
    @popperjs/core: ^2.11.6 => 2.11.6
    @rushstack/eslint-patch: ^1.1.4 => 1.1.4
    @types/node: ^16.11.47 => 16.11.56
    @vitejs/plugin-vue: ^3.0.1 => 3.0.3
    @vue/eslint-config-prettier: ^7.0.0 => 7.0.0
    @vue/eslint-config-typescript: ^11.0.0 => 11.0.0
    @vue/tsconfig: ^0.1.3 => 0.1.3
    aws-amplify: ^4.3.34 => 4.3.34
    bootstrap: ^5.2.0 => 5.2.0
    bootswatch: ^5.2.0 => 5.2.0
    eslint: ^8.21.0 => 8.23.0
    eslint-plugin-vue: ^9.3.0 => 9.4.0
    npm-run-all: ^4.1.5 => 4.1.5
    pinia: ^2.0.17 => 2.0.21
    prettier: ^2.7.1 => 2.7.1
    typescript: ~4.7.4 => 4.7.4
[...]

Issue #10260, opened 2022-08-25
GraphQL subscriptions fail when using modular api and pubsub packages


pfried:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication, GraphQL API, PubSub

### Amplify Categories

auth, api

### Environment information

<details>

```
# Put output below this line
System:
    OS: Windows 10 10.0.19044
    CPU: (24) x64 AMD Ryzen 9 3900X 12-Core Processor
    Memory: 6.71 GB / 31.95 GB
  Binaries:
    Node: 16.10.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 3.2.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 7.21.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 104.0.5112.101
    Edge: Spartan (44.19041.1266.0), Chromium (104.0.1293.63)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @aws-amplify/api: ^4.0.51 => 4.0.51
    @aws-amplify/auth: ^4.6.4 => 4.6.4
    @aws-amplify/core: ^4.7.2 => 4.7.2
    @aws-amplify/pubsub: ^4.5.1 => 4.5.1
    @casl/ability: ^5.4.4 => 5.4.4
    @casl/ability/extra:  undefined ()
    @casl/react: ^3.0.0 => 3.0.1
    @craco/craco: ^6.4.3 => 6.4.5
    @headlessui/react: ^1.6.6 => 1.6.6
    @heroicons/react: ^1.0.6 => 1.0.6
    @hookform/resolvers: 2.8.0 => 2.8.0
    @juggle/resize-observer: ^3.3.0 => 3.4.0
    @mapbox/mapbox-gl-style-spec:  13.26.0
    @nantis/grafana-data: * => 0.0.0
    @nantis/gridknight-measurements: * => 0.0.0
[...]

PR #10298, opened 2022-09-07
fix(pubsub): Fix vite build issue

stocaaro:
#### Description of changes
There is an issue related to the object load order that impacts the Vite build process (and maybe others). This [stack overflow](https://stackoverflow.com/questions/53122751/typeerror-object-prototype-may-only-be-an-object-or-null-undefined) question helped identify the nature of the problem.  In this case, the PubSub `index` files loads providers, which in turn load `CONNECTION_STATE_CHANGE` from further down `index` creating a cycle.

Moving this constant to a file so that the load order is acyclic fixes the issue.

#### Issue #, if available
#10295
#10287
#10260

I was able to reproduce the issue by following the Vue project instructions from #10287, however had trouble getting linking working to test a fix. The example React project from #10295 was very helpful in proving and testing that this fix works.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/pubsub/src/Providers/MqttOverWSProvider.ts
+++ b/packages/pubsub/src/Providers/MqttOverWSProvider.ts
@@ -21,8 +21,7 @@ import {
 	ConnectionStateMonitor,
 	CONNECTION_CHANGE,
 } from '../utils/ConnectionStateMonitor';
-import { AMPLIFY_SYMBOL } from './constants';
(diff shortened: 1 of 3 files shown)


elorzafe (approved):
Looks good

Thanks @stocaaro

Samaritan1011001 (approved):
LGTM. Thanks!

PR #10302, opened 2022-09-08
Add PlaceId support for Geo Autocomplete feature

sreeramsama:
#### Description of changes
- Update existing SearchForSuggestions API to return PlaceId (if available) along with text (title, address) of the places.
- Implement a new API called SearchForLocationById which will be a wrapper around AmazonLocationService’s GetPlace API

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/geo/__tests__/Geo.test.ts
+++ b/packages/geo/__tests__/Geo.test.ts
@@ -62,6 +62,7 @@ LocationClient.prototype.send = jest.fn(async command => {
 			Results: [
 				{
 					Text: 'star',
+					PlaceId: 'a1b2c3d4',
 				},
 				{
 					Text: 'not star',
@@ -333,7 +334,15 @@ describe('Geo', () => {
 
 	describe('searchForSuggestions', () => {
 		const testString = 'star';
-		const testResults = ['star', 'not star'];
+		const testResults = [
+			{
+				text: 'star',
+				placeId: 'a1b2c3d4',
+			},
+			{
+				text: 'not star',
+			},
+		];
 
 		test('should search with just text input', async () => {
 			jest.spyOn(Credentials, 'get').mockImplementationOnce(() => {
--- a/packages/geo/__tests__/Geo.test.ts
+++ b/packages/geo/__tests__/Geo.test.ts
@@ -18,6 +18,7 @@ import {
 	SearchPlaceIndexForPositionCommand,
 	GetPlaceCommand,
 } from '@aws-sdk/client-location';
+import camelcaseKeys from 'camelcase-keys';
 
 import { GeoClass } from '../src/Geo';
 import { AmazonLocationServiceProvider } from '../src/Providers/AmazonLocationServiceProvider';
@@ -340,9 +341,7 @@ describe('Geo', () => {
 
 	describe('searchByPlaceId', () => {
 		const testPlaceId = 'a1b2c3d4';
-		const testResults = {
-			Place: TestPlacePascalCase,
-		};
+		const testResults = camelcaseKeys(TestPlacePascalCase, { deep: true });
 
 		test('should search with PlaceId as input', async () => {
 			jest.spyOn(Credentials, 'get').mockImplementationOnce(() => {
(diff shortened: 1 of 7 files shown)


packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -292,14 +296,59 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+			text: result.Text,
+			placeId: result.PlaceId,
+		}));
 
 		return results;
 	}
 
+	private _verifyPlaceId(placeId: string) {
+		// if (isValid(placeId)) { - if we have more validation rules for placeId
+		if (placeId.length === 0) {
```
thaddmt, line 311: We can probably check for undefined here as well. Have we asked the location team if there is a regex or something in place for the placeId? They seem to list regex for other things on their reference page - https://docs.aws.amazon.com/location/latest/APIReference/API_DescribePlaceIndex.html

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -292,14 +296,59 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
+
+		const client = new LocationClient({
+			credentials: this._config.credentials,
+			region: this._config.region,
+			customUserAgent: getAmplifyUserAgent(),
+		});
+
+		const searchForLocationByIdInput: GetPlaceCommandInput = {
+			PlaceId: placeId,
+			IndexName: options?.searchIndexName,
```
thaddmt, line 337: This should default to `this._config.search_indices.default` and be options.searchIndexName if set right? Maybe something like

```
options?.searchIndexName || this._config.search_indices.default
```

thaddmt (commented):
LGTM so far, just waiting on unit tests :)

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
sreeramsama: Did not check for undefined as this function only accepts placeId as a string.
Checked with the Location Services team and they said there is no regex - and they just check for undefined or empty strings.

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
sreeramsama: Yes. Thanks for catching that. Will update it.

packages/geo/__tests__/Geo.test.ts
```diff
@@ -332,6 +338,45 @@ describe('Geo', () => {
 		});
 	});
 
+	describe('searchByPlaceId', () => {
+		const testPlaceId = 'a1b2c3d4';
+		const testResults = {
+			Place: TestPlacePascalCase,
+		};
+
+		test('should search with PlaceId as input', async () => {
```
thaddmt, line 347: Can we also test some error cases here?

packages/geo/__tests__/Providers/AmazonLocationServiceProvider.test.ts
```diff
@@ -478,6 +484,56 @@ describe('AmazonLocationServiceProvider', () => {
 		});
 	});
 
+	describe('searchByPlaceId', () => {
+		const testPlaceId = 'a1b2c3d4';
+		const testResults = {
+			Place: TestPlacePascalCase,
+		};
+
+		test('should search with PlaceId as input', async () => {
```
thaddmt, line 493: Can we also test some error cases or invalid inputs here?

thaddmt (approved):

thaddmt:
Can you ask in the JS channel if we can run integration / unit tests on a forked pr? if not maybe we should move this to a branch on amplify-js direclty so the tests run

packages/geo/__tests__/Geo.test.ts
ioanabrooks: +1

thaddmt (approved):

PR #10306, opened 2022-09-09
chore(datastore): add cms auth use case test

iartemiev:
#### Description of changes

Adds unit test that ensures default auth mode is always respected (when not using multi-auth), regardless of MIPR auth metadata

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/__tests__/CMSUseCases.test.ts
```diff
@@ -0,0 +1,185 @@
...
+								},
+							],
+						},
+					},
+				],
+			},
+		},
+		version: '2',
+	};
+}
```
david-mcafee, line 185: Thoughts on adding these to our [helpers](https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/__tests__/helpers.ts) so they can be used in other tests, as well?

david-mcafee (approved):
Added a minor comment, otherwise LGTM!

alharris-at (approved):

packages/datastore/__tests__/CMSUseCases.test.ts
iartemiev: The schemas are pretty specific to this test and would likely be worthless outside of it, but if we run into other use cases down the road, I'm happy to extract them out.

packages/datastore/__tests__/CMSUseCases.test.ts
```diff
@@ -0,0 +1,185 @@
...
+		await DataStore.start();
+		await new Promise(res => setTimeout(res, 10));
+
+		expect(gqlMock).toHaveBeenCalledTimes(2);
+		expect(gqlMock).toHaveBeenCalledWith(
+			expect.objectContaining({
+				authMode: 'AWS_IAM',
+			})
+		);
+	});
```
svidgen, line 65: While I don't imagine they'd be different, can we get some mutations in here to ensure the correct auth mode is used there as well?

How about subscriptions?

Issue #6906, opened 2020-10-03
Amplify Personalize Record Events For MediaAutoTrack are Failing 


rkarthikr:
**Describe the bug**
MediaAutoTrack Events Using Amplify SDK version - 3.3.3,3.3.2 is triggering events to tracker but they are failing due to bad payload . See the Sample Request/Response section at the bottom

https://docs.amplify.aws/lib/analytics/personalize/q/platform/js#working-with-the-api
https://aws.amazon.com/blogs/aws/amazon-personalize-is-now-generally-available/

**To Reproduce**
1. Record Event for MediaAutoTrack Using the Dom ID for video player 
2. Play the Video
3. Check Tracker Events are sent to Personalize 
4. Check Request/Response in Developer Tools 
5. HTTP 400 Error Code

**Expected behavior**
Events are successfully submitted . HTTP 200 response code for Put Events

**Desktop (please complete the following information):**
 - OS: macOS Catalina
 - Browser: Firefox
 - Version: 68.8.0esr (64-bit)

**Additional context**

 **Sample Request**

```
{
    "eventList": [
        {
            "eventId": "01cfc340-0539-184387240-0526-11eb-aa3a-29f4dbbd8521",
            "eventType": "TimeWatched",
            "properties": "{\"itemId\":\"593\",\"timestamp\":139.472841,\"duration\":597,\"eventValue\":0.2336}",
            "sentAt": 1601702130
        },
        {
            "eventId": "01cfc340-0539-184387240-0526-11eb-aa3a-29f4dbbd8521",
            "eventType": "TimeWatched",
            "properties": "{\"itemId\":\"593\",\"timestamp\":139.472841,\"duration\":597,\"eventValue\":0.2336}",
            "sentAt": 1601702130
        },
        {
            "eventId": "01cfc340-0539-184387240-0526-11eb-aa3a-29f4dbbd8521",
            "eventType": "TimeWatched",
            "properties": "{\"itemId\":\"593\",\"timestamp\":139.472841,\"duration\":597,\"eventValue\":0.2336}",
            "sentAt": 1601702130
        },
        {
            "eventId": "01cfc340-0539-184387240-0526-11eb-aa3a-29f4dbbd8521",
            "eventType": "TimeWatched",
            "properties": "{\"itemId\":\"593\",\"timestamp\":139.472841,\"duration\":597,\"eventValue\":0.2336}",
            "sentAt": 1601702130
        }
    ],
[...]

PR #10329, opened 2022-09-15
bug(analytics): change timestamp to camelCase

katiegoines:
#### Description of changes
Changed the case styling of `timestamp` to `timeStamp` [per documentation](https://docs.aws.amazon.com/personalize/latest/dg/API_UBS_Event.html#:~:text=The%20keys%20use%20camel%20case%20names%20that%20match%20the%20fields%20in%20the%20Interactions%20schema.%20In%20the%20above%20example%2C%20the%20numberOfRatings%20would%20match%20the%20%27NUMBER_OF_RATINGS%27%20field%20defined%20in%20the%20Interactions%20schema). 

#### Issue #, if available
https://github.com/aws-amplify/amplify-js/issues/6906

Manual testing.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


AllanZhengYP (approved):

jimblanc (commented):
Could this be considered a breaking change? It looks like the existing naming has been in place for a few years (even though it disagrees with the docs).

stocaaro:
Timestamp can be one word or two. The error from the related issue is what matters. I just tried to trace out the API interface that needs this to be camelcase and couldn't find how this buffer is cleared. I could repro, but I suspect that this is what you mean by "manual testing". Can you add more detail about what you did to test this and the behavior you observed?

In this case, unless there's an API definition we can use to test inputs against, unit testing this would be brittle and not especially helpful, so I wouldn't advocate for that at this time, unless there is an interface for this that I missed.

jimblanc (approved):

jimblanc (changes requested):
Ran into some issues while testing this PR, blocking merge until I can update.

PR #10378, opened 2022-09-27
fix: correct no-op write detection

iartemiev:
#### Description of changes

The relevant change is in `packages/datastore/src/storage/storage.ts` - exit early when there are no changes introduced in the update mutation, instead of adding the PK to the mutation input when no other fields have been changed.

The changes to the `storage.test.ts` test file are there simply to allows us to run tests individually (via `.only`) without side-effects

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/src/storage/storage.ts
+++ b/packages/datastore/src/storage/storage.ts
@@ -408,15 +408,15 @@ class StorageClass implements StorageFacade {
 			}
 		});
 
+		if (Object.keys(updatedElement).length === 0) {
+			return null;
+		}
+
 		// include field(s) from custom PK if one is specified for the model
 		if (primaryKey && primaryKey.length) {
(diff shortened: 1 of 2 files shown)


packages/datastore/src/storage/storage.ts
```diff
@@ -408,17 +408,17 @@ class StorageClass implements StorageFacade {
 			}
 		});
 
+		if (Object.keys(updatedElement).length === 0) {
```
david-mcafee, line 411:
```suggestion
        /* Exit early when there are no changes introduced in the update mutation 
        (instead of adding the PK to the mutation input when no other fields have 
        been changed) */
		if (Object.keys(updatedElement).length === 0) {
```

david-mcafee (approved):
LGTM!

PR #10379, opened 2022-09-27
feat: add TS coverage report per package

israx:
#### Description of changes

This PR creates a TS report per package after running yarn ts-coverage at the root of the workspace or at every package.

The following changes are:
-Adds current changes from [https://github.com/aws-amplify/amplify-js/pull/10293](url)
-Adds ts-coverage script to every package
-Adds lerna ts-coverage script at the root of the workspace 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/amazon-cognito-identity-js/tsconfig.build.json
```diff
@@ -0,0 +1,7 @@
+{
+	"extends": "../tsconfig.settings.json",
+	"compilerOptions": {
+		"composite": true
```
jimblanc, line 4: Should this be moved up to `tsconfig.settings.json`? Seems like it appears in every package

scripts/build.js
```diff
@@ -153,32 +157,24 @@ async function buildES5(typeScriptCompiler, watchMode) {
...
-		incremental: true,
+		jsx: jsx,
 		tsBuildInfoFile: es5TsBuildInfoFilePath,
 		typeRoots,
-		// temporary fix
-		types: ['node'],
 		outDir: pkgTscES5OutDir,
 	};
 
+	if ('composite' in compilerOptions) delete compilerOptions['composite'];
```
jimblanc, line 176: Could you add a small comment explaining why we're doing this?

jimblanc (commented):
LGTM besides some nitpicks, thanks for setting this up!

packages/amazon-cognito-identity-js/tsconfig.build.json
israx: Actually, I'm removing this "composite" flag, there's not use to it

scripts/build.js
israx: As I'm removing the "composite" flag, that code is not needed any more

jimblanc (approved):

PR #10409, opened 2022-10-03
Create method that verificates if an instance was created, in order t…

miranciandt:
Hello. I've opened this pull request due to an issue we faced on our project while using amplify, where the graphql class could not read the getCancellableToken variable

 
We found out that this was happening because an instance was not being created before calling the graphql method, once the verification if the instance exists was happening after the assignment of the cancellableToken variable.

#### Description of changes
To work around this, I put the if statement that verificates if an instance was created into a method, in order to make it reusable. In this case, it was necessary to use it before the assignment of the cancellableToken variable.

[VScode debugger]
Code block behavior

Screenshot from 2022-10-03 14-03-35

- Before the implementation:

https://user-images.githubusercontent.com/93326804/193637213-7bb3c531-23e4-47a3-a715-8655a2eb0017.mp4

- After the implementation:

https://user-images.githubusercontent.com/93326804/193653243-4d8dcd9a-eae5-49a9-ac66-44df4692642b.mp4

#### Checklist
- [✓] PR description included
- [✓] `yarn test` passes
- [✓] Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)
- [ x ] Relevant documentation is changed or added (and PR referenced)


packages/api-graphql/__tests__/GraphQLAPI-test.ts
```diff
@@ -1290,11 +1294,21 @@ describe('API test', () => {
...
 
-			await api.graphql(graphqlOperation(GetEvent, variables, authToken, userAgentSuffix));
+			await api.graphql(
+				graphqlOperation(GetEvent, variables, authToken, userAgentSuffix)
+			);
 
 			expect(spyon).toBeCalledWith(url, init);
 		});
 
+		test('call isInstanceCreated', () => {
```
dpilch, line 1304: Could you create a test to assert that `createInstance` is no called when an instance already exists?

packages/api-graphql/src/GraphQLAPI.ts
```diff
@@ -359,6 +357,12 @@ export class GraphQLAPIClass {
 		return response;
 	}
 
+	async isInstanceCreated() {
```
dpilch, line 360: This reads more like it should return a boolean. Maybe something like `createInstanceIfNotCreated` would be a better fit here.

dpilch (commented):
Thanks for the contribution! Overall LGTM, have a couple comments.

packages/api-graphql/__tests__/GraphQLAPI-test.ts
miranciandt: Done!

packages/api-graphql/src/GraphQLAPI.ts
miranciandt: Done! I agree with the name, thanks for the suggestion :slightly_smiling_face:

miranciandt:
@dpilch fixed

stocaaro (approved):

stocaaro:
Thank you for this contribution! We've merged your change to main and expect it to go out in the release tomorrow.

PR #10416, opened 2022-10-04
feat: add bundle_size job in CCI flow

israx:
#### Description of changes

-adds bundle_size job in CCI flow
-upgrades node version in build executor

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


jimblanc (approved):
Nice, good first step! Could you confirm that you got a clean integ test run (pretty big cypress upgrade)? Also we should hold off on merging to main until we discuss this afternoon

israx:
> Could you confirm that you got a clean integ test run (pretty big cypress upgrade)? I think the integration tests are running on another executor that has a different node version.

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ version: 2.1
 executors:
   build-executor:
     docker:
-      - image: cypress/base:12.22.8
+      - image: cypress/base:14.15.0
```
haverchuck, line 8: Have we verified that all other tests run successfully with this version bump?

packages/auth/package.json
```diff
@@ -28,8 +28,15 @@
...
     "clean": "rimraf lib-esm lib dist",
     "format": "echo \"Not implemented\"",
-    "lint": "tslint '{__tests__,src}/**/*.ts'"
+    "lint": "tslint '{__tests__,src}/**/*.ts'",
+    "bundle-size": "size-limit"
   },
+   "size-limit": [
+       {
+        "limit": "41.633 kB",
+        "path": "dist/aws-amplify-auth.min.js"
```
haverchuck, line 37: This is testing the size of the minified file only.  If that's what we are going to do for now, maybe we should change the command names to "size-test" or something, and not use the term "bundle".

Have you looked into testing the actual impact of the auth/core modules on a web app bundled with webpack?

package.json
```diff
@@ -85,6 +87,7 @@
 		"rollup-plugin-node-resolve": "^4.0.0",
 		"rollup-plugin-sourcemaps": "^0.4.2",
 		"rollup-plugin-typescript": "^1.0.0",
+		"size-limit": "^8.1.0",
```
haverchuck, line 90: This is a CLI tool, yeah? Have other teams looked at this and vetted it with security?

haverchuck (changes requested):
See inline comments.

package.json
israx: The UI is using the same dev dependency

.circleci/config.yml
israx: Unit tests run and work but need to verify if integration tests are using this version bump

iartemiev: Integ tests use different executors altogether, so they won't be affected

packages/auth/package.json
iartemiev: +1

abdallahshaban557:
Closing this - confirmed with @jimblanc

Issue #10357, opened 2022-09-22
`DataStore.save` silently failing for nested type arrays


cwoolum:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

DataStore

### Amplify Categories

api

### Environment information

<details>

```
# Put output below this line
  System:
    OS: Linux 5.4 Amazon Linux 2
    CPU: (16) x64 Intel(R) Xeon(R) Platinum 8175M CPU @ 2.50GHz
    Memory: 3.40 GB / 61.37 GB
    Container: Yes
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.20.0 - ~/.asdf/installs/nodejs/lts-fermium/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.16.0/bin/yarn
    npm: 8.19.2 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Firefox: 91.10.0esr
  npmPackages:
    @aws-amplify/ui-react: ^3.5.2 => 3.5.2 
    @aws-amplify/ui-react-internal:  undefined ()
    @aws-amplify/ui-react-legacy:  undefined ()
    @types/react: ^18.0.17 => 18.0.20 
    @types/react-dom: ^18.0.6 => 18.0.6 
    @vitejs/plugin-react: ^2.1.0 => 2.1.0 
    aws-amplify: ^4.3.35 => 4.3.35 
    react: ^18.2.0 => 18.2.0 (18.1.0)
    react-dom: ^18.2.0 => 18.2.0 
    react-router: ^6.4.0 => 6.4.0 
    react-router-dom: ^6.4.0 => 6.4.0 
    typescript: ^4.6.4 => 4.8.3 
    vite: ^3.1.0 => 3.1.1 
  npmGlobalPackages:
    npm: 8.19.2
    nx: 14.7.5

```

</details>

### Describe the bug

[...]

PR #10417, opened 2022-10-04
fix: validate non models when using object literal

dpilch:
#### Description of changes

Validate non model fields when using object literal syntax.

Non models can be created with an object literal instead of a constructor. When the object literal is used the validation is skipped. This results in `DataStore.save` successfully saving locally, but the subsequent mutation fails to sync with the cloud. The failed sync is pushed to `errorHandler`, but is otherwise lost to the user.

The following are equivalent methods of creating a new Post.

With constructor:

```
new Post({
  name: 'My Post',
  metadata: new Metadata({
    author: 'Some author',
    tags: undefined,
  }),
})
```

Without constructor:

```
new Post({
  name: 'My Post',
  metadata: { // no constructor
    author: 'Some author',
    tags: undefined,
  },
})
```

In the second example there is no validation. This would allow the user to enter something like `author: 5` without an easily noticeable failure. If they were to use TypeScript they would get a build failure. When not using TypeScript there is no error.

#### Issue #, if available

Resolves [#10357](https://github.com/aws-amplify/amplify-js/issues/10357)

* Unit testing
* Tested in an app with the following schema

```
type Parent @model {
  id: ID!
  name: String
  requiredChildren: [Child]!
  requiredChild: Child!
  children: [Child]
  child: Child
}

type Child {
  id: ID!
  name: String!
  description: String
}
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/__tests__/DataStore.ts
```diff
@@ -1516,295 +1516,596 @@ describe('DataStore tests', () => {
...
 		});
 
-		test('Instantiation validations', async () => {
-			expect(() => {
-				new Model({
-					field1: undefined,
-					dateCreated: new Date().toISOString(),
-				});
-			}).toThrowError('Field field1 is required');
+		describe.only('Instantiation validations', () => {
```
svidgen, line 1519: You've got a `.only` hanging out here!

packages/datastore/src/datastore/datastore.ts
```diff
@@ -394,6 +397,56 @@ const validateModelFields =
 						`Field ${name} should be of type ${type}, validation failed. ${v}`
 					);
 				}
+			} else if (isNonModelFieldType(type)) {
```
svidgen, line 400: Not a blocker ⬇️ 

Would be good to think about whether some of these larger methods can be elegantly decomposed down the line.

svidgen (approved):

packages/datastore/src/datastore/datastore.ts
```diff
@@ -394,6 +397,56 @@ const validateModelFields =
...
+					const modelValidator = validateModelFields(subNonModelDefinition);
+
+					if (isArray) {
+						let errorTypeText: string = type.nonModel;
+						if (!isRequired) {
+							errorTypeText = `${type.nonModel} | null | undefined`;
+						}
+						if (!Array.isArray(v)) {
+							throw new Error(
+								`Field ${name} should be of type [${errorTypeText}], ${typeof v} received. ${v}`
```
david-mcafee, line 414: Nit (and not a blocker): could place these error messages [here](https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/src/util.ts#L54-L62)

david-mcafee (approved):
LGTM! Just left one minor comment.

PR #10470, opened 2022-10-13
test: add await to expect.rejects

dpilch:
#### Description of changes

Some tests are missing `await` on `expect.rejects` resulting in some tests passing when they should not on Node < 14. Continuation of https://github.com/aws-amplify/amplify-js/pull/8927.

#### Issue #, if available

N/A

With node < 14 and node >= 14.

```
yarn test
```

Observe that there is no error message in test results.
```
(node:30943) UnhandledPromiseRejectionWarning: Error: expect(received).rejects.toThrowError(expected)
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/__tests__/GraphQLAPI-test.ts
```diff
@@ -755,7 +755,7 @@ describe('API test', () => {
 					variables,
 					authMode: GRAPHQL_AUTH_MODE.OPENID_CONNECT,
 				})
-			).rejects.toThrowError('No federated jwt');
+			).rejects.toThrowError('No current user');
```
micksatana, line 758: The previous test case intents to test when able to getItem but no `token` prop in [this line](https://github.com/aws-amplify/amplify-js/blob/36de9014320f2a39caba492ad156da4f6a79575a/packages/api-graphql/src/GraphQLAPI.ts#L171)

So we may not want to change the expected result here but change the mock to be correctly mocked instead. Perhaps, change the `jest.spyOn(Cache, 'getItem').mockReturnValue(null);` to `jest.spyOn(Cache, 'getItem').mockReturnValue({});` should make it correctly tested.

dpilch: Fixed. I need to make some changes to `GraphQLAPI.ts` as well because the try/catch was swallowing the `No federated jwt` error.

PR #10518, opened 2022-10-21
feat: add ts-coverage threshold to categories

israx:
#### Description of changes

- Analyses internal typescript usage
- Adds a 0 threshold for ts-coverage per package
- Adds ts-coverage script to tests 

#### Checklist

- [X] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/analytics/package.json
+++ b/packages/analytics/package.json
@@ -16,7 +16,7 @@
 		"access": "public"
 	},
 	"scripts": {
-		"test": "tslint 'src/**/*.ts' && jest -w 1 --coverage",
+		"test": "tslint 'src/**/*.ts' && jest -w 1 --coverage && npm run ts-coverage",
 		"test:watch": "tslint 'src/**/*.ts' && jest -w 1 --watch",
 		"build-with-test": "npm run clean && npm test && tsc && webpack",
 		"build:cjs": "node ./build es5 && webpack && webpack --config ./webpack.config.dev.js",
--- a/packages/analytics/package.json
+++ b/packages/analytics/package.json
@@ -27,7 +27,7 @@
 		"clean": "rimraf lib-esm lib dist",
 		"format": "echo \"Not implemented\"",
 		"lint": "tslint 'src/**/*.ts' && npm run ts-coverage",
-		"ts-coverage":  "typescript-coverage-report -p ./tsconfig.build.json -t 0"
+		"ts-coverage":  "typescript-coverage-report -p ./tsconfig.build.json -t 69.26"
 	},
 	"react-native": {
 		"./lib/index": "./lib-esm/index.js",
--- a/packages/analytics/package.json
+++ b/packages/analytics/package.json
@@ -16,7 +16,7 @@
 		"access": "public"
 	},
 	"scripts": {
-		"test": "tslint 'src/**/*.ts' && jest -w 1 --coverage && npm run ts-coverage",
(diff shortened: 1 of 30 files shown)


packages/analytics/package.json
```diff
@@ -16,7 +16,7 @@
 		"access": "public"
 	},
 	"scripts": {
-		"test": "tslint 'src/**/*.ts' && jest -w 1 --coverage",
+		"test": "tslint 'src/**/*.ts' && jest -w 1 --coverage && npm run ts-coverage",
```
jimblanc, line 19: I think this might make more sense to do within the `lint` script, also not your change but `test` should probably be using that internally to reduce duplication, e.g.:

```
"test": "npm run lint && jest -w 1 --coverage",
...
"lint": "tslint 'src/**/*.ts' && npm run ts-coverage",
```

Ditto for other packages in this PR.

packages/analytics/package.json
```diff
@@ -27,7 +27,7 @@
 		"clean": "rimraf lib-esm lib dist",
 		"format": "echo \"Not implemented\"",
 		"lint": "tslint 'src/**/*.ts'",
-		"ts-coverage":  "typescript-coverage-report -p ./tsconfig.build.json -t 68"
+		"ts-coverage":  "typescript-coverage-report -p ./tsconfig.build.json -t 0"
```
jimblanc, line 30: I'm curious why we can't just enable this now with some lower threshold, maybe something like `<current coverage>-2%`?

jimblanc (commented):
Thanks for setting this up!

packages/analytics/package.json
stocaaro: If we have this info or it's easy to get, I'd be a fan. Can easily be part of the package by package improvement work. Would it take more than 2 hours to add a reasonable % with this change?

packages/analytics/package.json
stocaaro: Seems reasonable. This is essentially a lint.

packages/analytics/package.json
israx: good idea. I will define the new thresholds then based on the current ones

israx: It's updated it

packages/auth/package.json
```diff
@@ -19,7 +19,7 @@
     "access": "public"
   },
   "scripts": {
-    "test": "yarn lint --fix && jest -w 1 --coverage",
+    "test": "yarn lint --fix && jest -w 1 --coverage ",
```
jimblanc, line 22: Nit:  ` ` (few other instances in PR)

packages/storage/package.json
```diff
@@ -28,8 +28,8 @@
     "build": "npm run clean && npm run build:esm && npm run build:cjs",
     "clean": "rimraf lib-esm lib dist",
     "format": "echo \"Not implemented\"",
-    "lint": "tslint 'src/**/*.ts'",
-    "ts-coverage":  "typescript-coverage-report -p ./tsconfig.build.json"
+    "lint": "tslint 'src/**/*.ts'  && npm run ts-coverage",
```
jimblanc, line 31: Nit: `  ` (double space in lint script)

jimblanc (commented):
Change looks good but few nits around spacing.

packages/storage/package.json
israx: just updated it. Thanks for checking that

PR #10523, opened 2022-10-24
Migrate `auth` and `amazon-cognito-identity-js` to use `@aws-crypto/sha256-js`

jimblanc:
#### Description of changes
This change removes the dependency on `crypto-js` from the `auth` & `amazon-cognito-identity-js` packages in favor of `@aws-crypto/sha256-js` which is already used by `core` & `cache`. This results in a parsed bundle size improvement of 7.48 kB for the `auth` category.

##### A note on backwards compatibility
Originally this change was thought to be a breaking change in Node & RN, however the final implementation uses the pure JS version of the crypto utilities which vends a small inline polyfill for `Buffer`. No issues were encountered during manual testing in RN & Node.

See: https://github.com/aws/aws-sdk-js-crypto-helpers/blob/master/packages/util/src/convertToBuffer.ts#L7

##### Other changes
Updated `amazon-cognito-identity-js`'s UglifyJS plugin version to resolve a build issue.

#### Issue #, if available
NA

- Unit tests
- Clean integ test run against v5: WIP
- Manually tested updated code-paths in React Native & Node 14

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)


packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
```diff
@@ -4,11 +4,8 @@
  */
 
 import { Buffer } from 'buffer';
-import CryptoJS from 'crypto-js/core';
-import 'crypto-js/lib-typedarrays'; // necessary for crypto js
-import SHA256 from 'crypto-js/sha256';
-import HmacSHA256 from 'crypto-js/hmac-sha256';
 import WordArray from './utils/WordArray';
+import { Sha256 as jsSha256 } from '@aws-crypto/sha256-js';
```
cshfang, line 8: Why do we need an alias here (and in the other files)? Since `Sha256` is a class I think I'd actually expect it to be capitalized.

cshfang (approved):
Just a question/suggestion but not really blocking. Lgtm

elorzafe (commented):
Looks good to me, just curious about the `uglify` plugin dependency that wasn't before

packages/amazon-cognito-identity-js/src/AuthenticationHelper.js
jimblanc: Made the change

Issue #10514, opened 2022-10-21
DataStore macOS Safari and iOS and iPadOS browsers - ConstraintError and undefined model[f] error


ebeltran1981:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Angular

### Amplify APIs

DataStore

### Amplify Categories

api

### Environment information

<details>

```
# Put output below this line

  System:
    OS: macOS 12.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 7.41 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.18.0 - ~/.nvm/versions/node/v16.18.0/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v16.18.0/bin/npm
  Browsers:
    Chrome: 106.0.5249.119
    Firefox: 106.0.1
    Safari: 16.0
  npmPackages:
    @angular-devkit/build-angular: ~13.3.2 => 13.3.9 
    @angular/animations: ~13.3.2 => 13.3.11 
    @angular/animations/browser:  undefined ()
    @angular/animations/browser/testing:  undefined ()
    @angular/cdk: ~13.3.2 => 13.3.9 
    @angular/cdk/a11y:  undefined ()
    @angular/cdk/accordion:  undefined ()
    @angular/cdk/bidi:  undefined ()
    @angular/cdk/clipboard:  undefined ()
    @angular/cdk/coercion:  undefined ()
    @angular/cdk/collections:  undefined ()
    @angular/cdk/drag-drop:  undefined ()
    @angular/cdk/keycodes:  undefined ()
    @angular/cdk/layout:  undefined ()
    @angular/cdk/observers:  undefined ()
    @angular/cdk/overlay:  undefined ()
    @angular/cdk/platform:  undefined ()
    @angular/cdk/portal:  undefined ()
[...]

Issue #10508, opened 2022-10-20
undefined is not an object evaluating model[f] or model[f] 


laclance:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React, Next.js

### Amplify APIs

DataStore

### Environment information

<details>

```
# Put output below this line
  System:
    OS: Windows 10 10.0.22621
    CPU: (4) x64 Intel(R) Core(TM) i3-10110U CPU @ 2.10GHz
    Memory: 3.05 GB / 7.84 GB
  Binaries:
    Node: 14.20.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 6.14.17 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.674.0), Chromium (106.0.1370.47)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @ampproject/toolbox-optimizer:  undefined ()
    @babel/core:  undefined ()
    @babel/helper-call-delegate: ^7.12.1 => 7.12.13
    @babel/runtime:  7.15.4
    @capacitor/core: ^4.3.0 => 4.3.0
    @capacitor/device: ^4.0.1 => 4.0.1
    @edge-runtime/primitives:  1.1.0-beta.31
    @hapi/accept:  undefined ()
    @napi-rs/triples:  undefined ()
    @next/react-dev-overlay:  undefined ()
    @react-pdf/renderer: ^3.0.0 => 3.0.0
    @segment/ajv-human-errors:  undefined ()
    @types/crypto-js: ^4.0.2 => 4.1.1
    @types/intl-tel-input: ^17.0.5 => 17.0.5
    @types/node: 18.11.0 => 18.11.0 (18.7.6)
    @types/qrcode.react: ^1.0.1 => 1.0.2
    @types/react: ^18.0.21 => 18.0.21
    @types/react-datepicker: ^4.4.1 => 4.4.2
[...]

PR #10527, opened 2022-10-24
fix(datastore): CPK on Safari

iartemiev:
#### Description of changes

Amplify 4.3.38 added client-side support for Custom Primary Keys (currently behind a CLI feature flag). With this change we needed to modify the index key paths for our IndexedDB stores in order to support Composite PKs (multi-field key). All store indexes now use array keys instead of strings. 

IndexedDB on Safari supports array key index fields, however, if the array key index is composed of a single field, Safari's IDB implementation treats it as a scalar key instead of an array key. It then expects that we pass a single field when returning or updating data in an index, i.e. `index.get('my-id')` instead of `index.get(['my-id])`. All other major browsers support the latter syntax, which is ECMAScript spec-compliant. This is unexpected behavior from Safari. However, we can easily remediate this in the DataStore implementation of our IndexedDB storage adapter by only passing an array when the size of the array is greater than 1, otherwise passing a scalar field.

This PR adds a detection and mitigation mechanism for this issue.

Here's a web app app written in Vanilla JS that demonstrates the issue when using the native IndexedDB web APIs:  https://github.com/iartemiev/safari-indexeddb-bug
I will open a bug with the WebKit/Safari team.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/10514
https://github.com/aws-amplify/amplify-js/issues/10508

* automated tests (unit & integ)
* manual smoke testing against sample apps

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/src/storage/adapter/IndexedDBSafariAdapter.ts
```diff
@@ -0,0 +1,1073 @@
...
+const DB_NAME = 'amplify-datastore';
+
+const keyArrOrSingleKey = (keyArr: string[]) => {
+	return keyArr.length > 1 ? keyArr : keyArr[0];
+};
+
+// IndexedDB on Safari has a bug where an array key composed of a single field
+// Has to be treated like a scalar key
+// issue for reference: https://github.com/aws-amplify/amplify-js/issues/10514
+class IndexedDBAdapterSafari implements Adapter {
```
stocaaro, line 50: This file is mostly a copy of the existing [IndexDBAdapter](https://github.com/aws-amplify/amplify-js/blob/0e3fc3e7926a4726ee9cdbd969d68400d3601fd5/packages/datastore/src/storage/adapter/IndexedDBAdapter.ts), right? I understand expediting a fix by making a full copy, so long as there's a plan to come back and factor out the redundancy. Is there a plan / timeline around this?

Any reason this is easier / safer than elevating functions that need to be overridden changed from private to protected and the inheriting instead of copying all the duplicate code?

iartemiev: This is safer because it guarantees we don't break other browsers. 

We're going to discuss a more long-term solution tomorrow.

david-mcafee (dismissed):
LGTM! We'll be discussing ways to improve on this after we get out the fix.

david-mcafee (dismissed):
LGTM!

packages/datastore/src/util.ts
```diff
@@ -417,6 +417,87 @@ export const isPrivateMode = () => {
 	});
 };
 
+let safariCompatabilityModeResult;
+
+export const isSafariCompatabilityMode: () => Promise<boolean> = async () => {
+	try {
+		const dbName = uuid();
+		const storeName = 'store';
```
svidgen, line 425: It *might* be worthwhile to put meaningful names here on the off-chance an one of these databases gets "stuck". `indexedDBFeatureProbeStore` or something. 🤷

packages/datastore/src/util.ts
```diff
@@ -417,6 +417,87 @@ export const isPrivateMode = () => {
...
+
+				const store = db.createObjectStore(storeName, {
+					autoIncrement: true,
+				});
+
+				store.createIndex(indexName, ['id']);
+			};
+		});
+
+		if (!db) return false;
```
svidgen, line 456: Memoize here too?

svidgen (dismissed):
OK. I think I grok it, and I'm OK with it as long as it's passing integ tests.

I'd really like to see unit tests that mock the Safari IndexedDB behavior though, if possible. Long term, if we can codify that behavior, I'd like to see all tests from the indexeddb tests + commonAdapterTests run against the "Safari personality" in addition to the "compliant personality".

david-mcafee (approved):
LGTM!

svidgen (approved):

svidgen (dismissed):

david-mcafee (approved):
LGTM!

svidgen (approved):

packages/datastore/src/storage/adapter/IndexedDBSafariAdapter.ts
iartemiev: Update: had some time today to refactor this and get rid of the redundant adapter. Resolving

PR #10540, opened 2022-10-26
Fixed Conflict handler doesn't run when a non-nullable @belongsTo fie…

ncarvajalc:
…ld exists and Optimistic Concurrency is used

#### Description of changes

`getConnectionFields` was changed replacing `result.push(${name} { id _deleted });` with `result.push(${association.targetName});` because the `jitteredRetry` function was getting an `error.data` of value `null` when `result.push(${name} { id _deleted });` was being used. This change solves the issue and the conflict handler gets executed normally. For more details on the issue see #10487 

#### Issue #, if available

#10487 

Tested in a sample app replicating what mentioned in #10487. Debugged to see if there were any braking changes with the change in the function as tests had to be changed. I validated both with subscriptions and non model fields to see if there were any breaking changes but there weren't. 

All tests in datastore pass after the change in `__tests__/graphql.ts`. 

#### Checklist

- [x] PR description included
- [x] `yarn run test --scope @aws-amplify/datastore ` passes (as tests in storage are not working)

### DIFF
--- a/packages/datastore/src/sync/utils.ts
+++ b/packages/datastore/src/sync/utils.ts
@@ -181,7 +181,7 @@ function getConnectionFields(
 							result.push(`${name} { ${keyFieldSelectionSet} _deleted }`);
 						} else {
 							// backwards-compatability for schema generated prior to custom primary key support
-							result.push(`${name} { id _deleted }`);
+							result.push(`${association.targetName}`);
 						}
 					}
 					break;
--- a/packages/datastore/__tests__/graphql.ts
+++ b/packages/datastore/__tests__/graphql.ts
@@ -28,14 +28,8 @@ metadata {
 _version
 _lastChangedAt
 _deleted
-reference {
-	id
-	_deleted
-}
-blog {
-	id
-	_deleted
-}
+referencePostId
+postBlogId
 `;
 
 const ownerAuthPostSelectionSet = `id


packages/datastore/src/sync/utils.ts
```diff
@@ -181,7 +181,7 @@ function getConnectionFields(
 							result.push(`${name} { ${keyFieldSelectionSet} _deleted }`);
 						} else {
 							// backwards-compatability for schema generated prior to custom primary key support
-							result.push(`${name} { id _deleted }`);
+							result.push(`${association.targetName}`);
```
alharris-at, line 184: N.p, but there's no need to wrap this in a string and deref here.

Can just enter `result.push(association.targetName);`

alharris-at:
@david-mcafee - could you take a look at this PR? I'm not sure if we're losing information by dropping the _deleted field, for example.

alharris-at:
@iartemiev - would you mind taking a look at this one as well?

packages/datastore/src/sync/utils.ts
david-mcafee: Was this tested with a previous version of Codegen? If we're entering the `else` block with the latest Codegen, that's a problem.

ncarvajalc: @david-mcafee This was tested 21 days ago with the latest version of amplify-cli. The details on how to reproduce it are in #10487

iartemiev:
@ncarvajalc - thank you for diving deep on this issue and opening the PR. 

Unfortunately, there are side effects from modifying the mutation selection set that may not be obvious at first. Specifically, this is a concern for multi-platform applications, e.g. JS + iOS/Android/Flutter DataStore clients interacting with the same AppSync API. Modifying the mutation selection set on JS will affect the client's ability to receive subscription messages on the other platforms, unless the subscription selection set is also modified on those. To summarize, the selection sets have to be kept in parity across platforms to allow for interoperability. 

There's currently a known issue with the iOS selection set which is causing problems when using iOS + another platform (https://github.com/aws-amplify/amplify-android/issues/1772, https://github.com/aws-amplify/amplify-ios/issues/1792#issuecomment-1182329582, etc.) The change in this PR would further increase the disparity and break interoperability between JS and all other platforms, not just iOS. 

In order to arrive at a solution, we'll need to update all platforms to use consistent selections sets. The iOS DataStore implementation in particular relies on a nested selection set of the related entity with all required fields present. The team is currently evaluating solutions to this problem, i.e. which selection set to use on all platforms.

Closing this PR, as we're unable to merge this change without cross-platform coordination.

PR #10546, opened 2022-10-27
fix: Standardize `cache` named export to preserve interoperability with RN

jimblanc:
#### Description of changes
This change standardizes the `cache` named export to vend a default implementation in order to preserve interoperability with React Native.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)


packages/cache/src/index.ts
```diff
@@ -1,21 +1,24 @@
...
+ */
+
+import { Amplify } from '@aws-amplify/core';
+import { BrowserStorageCache } from './BrowserStorageCache';
+import { InMemoryCache } from './InMemoryCache';
+import { CacheConfig } from './types';
+
+export { BrowserStorageCache, InMemoryCache, CacheConfig };
+
+// Standard `Cache` export to maintain interoperability with React Native
```
cshfang, line 21: Nit: I don't think we really need this comment here. Exporting as Cache seems pretty intuitive imo.

packages/cache/src/reactnative.ts
```diff
@@ -14,6 +14,9 @@
 import { Amplify } from '@aws-amplify/core';
 import { Cache, AsyncStorageCache } from './AsyncStorageCache';
 
-export { Cache, AsyncStorageCache };
+export { AsyncStorageCache };
+
+// Standard `Cache` export to maintain interoperability with React Native
```
cshfang, line 19: Nit: Ditto here but even more so we don't really need to explain that this is for React Native when this is the RN entry point imo.

cshfang (approved):
Lgtm thanks :D

packages/cache/src/index.ts
```diff
@@ -1,21 +1,24 @@
...
-import { Amplify } from '@aws-amplify/core';
-import { BrowserStorageCache } from './BrowserStorageCache';
-import { InMemoryCache } from './InMemoryCache';
-import { CacheConfig } from './types';
-
-export { BrowserStorageCache, InMemoryCache, CacheConfig };
-
-Amplify.register(BrowserStorageCache);
+/*
+ * Copyright 2017-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
elorzafe, line 2: nit, change it `2017-2022`

elorzafe (approved):
Changes looks good thanks @jimblanc 🎉 

Can we change `sideEffects` to `BrowserStorageCache` instead of `index.ts` that could be on a separate PR if you want

jimblanc:
> Changes looks good thanks @jimblanc 🎉
> 
> Can we change `sideEffects` to `BrowserStorageCache` instead of `index.ts` that could be on a separate PR if you want

I have this change in a separate branch, would prefer to take care of it in another PR.

elorzafe (approved):

PR #10562, opened 2022-10-28
chore: Move `cache` sideEffects to align with other packages

jimblanc:
#### Description of changes
This change moves the side effects in `cache` out of `index.ts` (and `reactnative.ts`) to align with the other categories. It also removes `cache` as a default static member (currently unused) of the Amplify singleton in order to reduce bundle size for use cases that don't directly use `cache`. Finally, it removes `cache` as a dependency of `auth`.

**Observed bundle size improvements in a barebones CRA app:**
Parsed: -15.1 kB
GZipped: -2.06 kB

Local testing.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


elorzafe (approved):
Looks good!

Thanks @jimblanc 🔥

packages/aws-amplify/src/amplifySingleton.ts
```diff
@@ -3,13 +3,11 @@
 
 import { Amplify, ServiceWorker } from '@aws-amplify/core';
 import { Auth } from '@aws-amplify/auth';
-import { Cache } from '@aws-amplify/cache';
 
 /** Always importing Auth when users import Amplify such that
 	for unauthenticated access (no sign in and sign up),
 	users don't have to import Auth explicitly **/
 Amplify.Auth = Auth;
-Amplify.Cache = Cache;
```
cshfang, line 12: Is this technically a breaking change if someone were accessing the Cache via the singleton?

packages/cache/src/AsyncStorageCache.ts
```diff
@@ -1,11 +1,11 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { Amplify, ConsoleLogger as Logger } from '@aws-amplify/core';
+import AsyncStorage from '@react-native-async-storage/async-storage';
```
cshfang, line 5: ✨

packages/aws-amplify/src/amplifySingleton.ts
cshfang: Just making sure we call out this breaking change in v5 changelog

cshfang (approved):

packages/aws-amplify/src/amplifySingleton.ts
jimblanc: Yep it would be, it'll have to go into the changelog (this usage pattern has been deprecated for a while I believe).

PR #10572, opened 2022-10-31
Change maven repository order to use google first

iOrcohen:
#### Description of changes

In the past day, jCenter had troubles in their Maven repository. That makes the android ```amplify-js:pushnotification``` package impossible to build, because there is a google-services dependency from jCenter which causes a timeout during builds that use the ```amplify-js:pushnotification``` package in the react native app.

image

In this PR I change the order of the maven repository to make google maven repository first, and solve the issue.

I run a build for the pushnotification package, which succeded. 
Also I make this change inside my react-native app and it worked.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/pushnotification/android/build.gradle
+++ b/packages/pushnotification/android/build.gradle
@@ -2,8 +2,8 @@
 
 buildscript {
     repositories {
-        jcenter()
-				google()
+        google()
+        jcenter()		
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.5.1'
@@ -13,10 +13,8 @@ buildscript {
 
 allprojects {
     repositories {
+        google()
         jcenter()
-        maven {
-            url 'https://maven.google.com'
-        }
     }
 }
 


packages/pushnotification/android/build.gradle
```diff
@@ -2,8 +2,8 @@
 
 buildscript {
     repositories {
-        jcenter()
-				google()
+        google()
+        jcenter()
```
cshfang, line 6:
```suggestion
        mavenCentral()
```

packages/pushnotification/android/build.gradle
```diff
@@ -13,10 +13,8 @@ buildscript {
 
 allprojects {
     repositories {
+        google()
         jcenter()
```
cshfang, line 17:
```suggestion
        mavenCentral()
```

cshfang (commented):
Hey @iOrcohen, thanks for submitting this PR. The change looks fine to me but I'm wondering if you agree we might as well just deprecate jcenter and migrate to mavenCentral.

tannerabread:
Closing this due to inactivity. This should be handled by the upcoming versions of push notification

PR #10596, opened 2022-11-02
fix(@aws-amplify/core): Fix guest credentials without Auth module

elorzafe:
#### Description of changes

Removing dependency on Auth for getting guest credentials

Sample app that will be included on samples staging repo in short, it was probe to be working with Analytics category on a react app.

```javascript
import { Analytics } from '@aws-amplify/analytics';

Amplify.configure(awsconfig);

function App() {
  Analytics.record({
    name: 'working'
  })
// ....
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/__tests__/Credentials-test.ts
+++ b/packages/core/__tests__/Credentials-test.ts
@@ -52,7 +52,7 @@ describe('Credentials test', () => {
 			expect(credentials.Auth).toBeUndefined();
 
 			expect(credentials.get()).rejects.toMatchInlineSnapshot(
-				`"No Auth module registered in Amplify"`
+				`"No Cognito Identity pool provided for unauthenticated access"`
 			);
 		});
 
--- a/packages/core/__tests__/Credentials-test.ts
+++ b/packages/core/__tests__/Credentials-test.ts
@@ -1,6 +1,6 @@
 import { CredentialsClass as Credentials } from '../src/Credentials';
 import { Amplify } from '../src/Amplify';
-
+import { Hub } from '../src/Hub';
 const session = {};
 
 const user = {
@@ -70,10 +70,21 @@ describe('Credentials test', () => {
 	});
 
 	describe('configure test', () => {
-		test('happy case', () => {
+		test('happy case', done => {
+			expect.assertions(1);
 			const config = {
 				attr: 'attr',
 			};
+
+			Hub.listen('core', ({ channel, payload, source }) => {
+				if (
+					channel === 'core' &&
+					payload?.event === 'credentials_configured' &&
+					source === 'Credentials'
+				) {
+					done();
+				}
+			});
 
 			const credentials = new Credentials(null);
 			expect(credentials.configure(config)).toEqual({
--- a/packages/core/src/Credentials.ts
+++ b/packages/core/src/Credentials.ts
@@ -118,7 +118,7 @@ export class CredentialsClass {
 		const { Auth = Amplify.Auth } = this;
 
 		if (!Auth || typeof Auth.currentUserCredentials !== 'function') {
-			// If Auth module is not import will do a best effort to get guest credentials
+			// If Auth module is not imported, do a best effort to get guest credentials
 			return this._setCredentialsForGuest();
 		}
 
--- a/packages/core/src/Credentials.ts
+++ b/packages/core/src/Credentials.ts
@@ -18,6 +18,7 @@ import {
 	GetCredentialsForIdentityCommand,
 } from '@aws-sdk/client-cognito-identity';
 import { CredentialProvider } from '@aws-sdk/types';
+import { parseAWSExports } from './parseAWSExports';
 
 const logger = new Logger('Credentials');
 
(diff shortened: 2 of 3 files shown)


jimblanc (commented):
Change looks good, but I think we would also need to remove the sideEffect in `aws-amplify` to realize any bundle size improvements here?

elorzafe:
> 

Can we do it in a separate PR?

jimblanc (approved):

packages/core/src/Credentials.ts
```diff
@@ -117,7 +118,8 @@ export class CredentialsClass {
 		const { Auth = Amplify.Auth } = this;
 
 		if (!Auth || typeof Auth.currentUserCredentials !== 'function') {
-			return Promise.reject('No Auth module registered in Amplify');
+			// If Auth module is not import will do a best effort to get guest credentials
```
cshfang, line 121: Nit:
```suggestion
			// If Auth module is not imported, do a best effort to get guest credentials
```

packages/core/src/Credentials.ts
```diff
@@ -233,7 +235,19 @@ export class CredentialsClass {
 
 	private async _setCredentialsForGuest() {
 		logger.debug('setting credentials for guest');
-		const { identityPoolId, region, mandatorySignIn } = this._config;
+		let { identityPoolId, region, mandatorySignIn } = this._config || {};
+
+		if (!identityPoolId) {
```
cshfang, line 240: If you modify this to be
```ts
if (!this._config?.identityPoolId) {
```
Could you move the destructuring assignment below the if clause and keep it as a `const`?

stocaaro (approved):

packages/core/src/Credentials.ts
elorzafe: I like that! thanks @cshfang

packages/core/src/Credentials.ts
```diff
@@ -233,7 +235,19 @@ export class CredentialsClass {
...
+			// If Credentials are not configured thru Auth module,
+			// doing best effort to check if the library was configured
+			this._config = Object.assign(
+				{},
+				this._config,
+				parseAWSExports(this._config || {}).Auth,
+				this._config
```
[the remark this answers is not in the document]
elorzafe: good catch, let me see why we are doing this on others

elorzafe:
I am doing a small change, please dont merge yet

packages/core/src/Credentials.ts
```diff
@@ -233,7 +256,17 @@ export class CredentialsClass {
...
+		if (!this._config?.identityPoolId) {
+			// If Credentials are not configured thru Auth module,
+			// doing best effort to check if the library was configured
+			this._config = Object.assign(
+				{},
+				this._config,
+				parseAWSExports(this._config || {}).Auth
+			);
+		}
+		const { identityPoolId, region, mandatorySignIn } = this._config || {};
```
cshfang, line 268: I don't think you need the `|| {}` since an empty config would have fallen into the if clause, thus assigning something to the config

packages/core/__tests__/Credentials-test.ts
```diff
@@ -70,11 +70,21 @@ describe('Credentials test', () => {
...
+					channel === 'core' &&
+					payload?.event === 'credentials_configured' &&
+					source === 'Credentials'
+				) {
+					done();
+				}
+			});
+
 			const credentials = new Credentials(null);
 			expect(credentials.configure(config)).toEqual({
```
cshfang, line 89: Can we double check that this is running before the test exits via the `done()` above? Maybe add an ` expect.assertions(1);` at the top?

packages/core/__tests__/Credentials-test.ts
```diff
@@ -70,11 +70,21 @@ describe('Credentials test', () => {
...
 
 	describe('configure test', () => {
-		test('happy case', () => {
+		test('happy case', done => {
 			const config = {
 				attr: 'attr',
 			};
 
+			Hub.listen('core', ({ channel, payload, source }) => {
+				if (
```
cshfang, line 79: Instead of the if, maybe we should turn this into assertions so that this test fails if these criteria aren't fulfilled instead.

jimblanc (approved):

packages/core/__tests__/Credentials-test.ts
elorzafe: Good catch! it failes if `done` is not called but I will add `expect.assertions`

packages/core/src/Credentials.ts
elorzafe: good catch!

packages/core/__tests__/Credentials-test.ts
elorzafe: if `done` is not called the test fails via timeout

ARJUNSV01:
autoSignIn doesn't work if Auth.confirmSignUp was called in a different session than Auth.signUp. Is there any solution to this?

PR #10611, opened 2022-11-04
fix(aws-amplify): Fix remove side effect

elorzafe:
#### Description of changes

Making `aws-amplify` sideEffect free

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/aws-amplify/src/index.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-export { Amplify } from './amplifySingleton';
```
stocaaro, line 4: Is this a breaking change? If not, can it wait until after the v5 launch?

I'm ironically worried about potential side effects of the change and needing to reset testing to give us high assurance.

stocaaro (commented):
Can you add context to the overview about how this change was tested?

jimblanc (approved):
Good stuff! Change looks good pending any discussion around v5 timing.

packages/aws-amplify/src/index.ts
elorzafe: The breaking part is the `"sideEffects": false`, this change is because that file is not longer needed, the whole purpose was the side effect

PR #10621, opened 2022-11-05
chore: Update package version numbers

stocaaro:
#### Description of changes
In preparation for the v5 rollout, updating the impacted package version numbers.

Since there are `files` changes in all packages or dependency on other packages with breaking changes, I've incremented everything except  `notifications` which is new in v5.

We discussed leaning on automation for this, where in theory a commit with the semver keyword "break:" should work, but did not during the last major version (from Francisco's memory). An explicit commit will be more predictable, although the first version of the released packages will be `X.0.1` instead of `X.0.0`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


cshfang (approved):

svidgen (approved):

.circleci/config.yml
```diff
@@ -369,8 +369,7 @@ jobs:
           # https://github.com/TryGhost/node-sqlite3#source-install
           command: |
             cd packages/datastore-storage-adapter
-            npm install --build-from-source
-            rm -rf node_modules/@aws-amplify node_modules/@aws-sdk
+            npm install sqlite3 --build-from-source
```
stocaaro, line 372: This build step was installing core 5.0.0 from npm, which doesn't exist yet. Jon changed it to not include all sources in the build process as they aren't needed for this step.

elorzafe (approved):
Looks good to me

Thanks @stocaaro 🎉

PR #10623, opened 2022-11-05
fix: sync expression types and shorthand handling

svidgen:
#### Description of changes

1. Swaps new predicate syntax types in place of old predicate types for sync expressions.
2. Fixes missing functional support for "shorthand" syntax.

Two idiosyncrasies:

1. TS cannot determine in "shorthand" syntax that the predicate builder is always defined. This will require a `?.` at the start of the shorthand builder, like so:

```
syncExpression(Post, post => post?.title.contains('cleaning'))
```

2. TS successfully gates return values from `async` "condition producers", but cannot provide auto-completion in this context. E.g.,

```
syncExpression(Post, async () => post => post.title.contains('cleaning'))
```

TS identifies this as correct and complains if it becomes incorrect, but does not provide autocompletion.

In my testing, the noted idiosyncrasies are consistent with the legacy predicate behavior and are not regressions.

1. Added unit tests. Red/Greened types in these tests as well as functionality.
2. Added missing syntaxes in integ test app.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/connectivityHandling.test.ts
+++ b/packages/datastore/__tests__/connectivityHandling.test.ts
@@ -1,6 +1,13 @@
 import { Observable } from 'zen-observable-ts';
 import { parse } from 'graphql';
-import { pause, getDataStore, waitForEmptyOutbox } from './helpers';
+import {
+	pause,
+	getDataStore,
+	waitForEmptyOutbox,
+	waitForDataStoreReady,
+} from './helpers';
+import { Predicates } from '../src/predicates';
+import { syncExpression } from '../src/types';
 
 describe('DataStore sync engine', () => {
 	// establish types :)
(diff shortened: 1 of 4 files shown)


packages/datastore/src/datastore/datastore.ts
```diff
@@ -496,11 +496,9 @@ const checkSchemaCodegenVersion = (codegenVersion: string) => {
...
 
 	if (!isValid) {
-		const message =
-			`Models were generated with an unsupported version of codegen. Codegen artifacts are from ${
-				codegenVersion || 'an unknown version'
-			}, whereas ^${majorVersion}.${minorVersion}.0 is required. ` +
-			"Update to the latest CLI and run 'amplify codegen models'.";
+		const message = `Models were generated with an unsupported version of codegen. Codegen artifacts are from ${
+			codegenVersion || 'an unknown version'
+		}, whereas ^${majorVersion}.${minorVersion}.0 is required. Update to the latest CLI and rerun codegen.`;
```
svidgen, line 501:
```suggestion
		const message =
			`Models were generated with an unsupported version of codegen. Codegen artifacts are from ${
				codegenVersion || 'an unknown version'
			}, whereas ^${majorVersion}.${minorVersion}.0 is required. ` +
			"Update to the latest CLI and run 'amplify codegen models'.";
```

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -226,4 +233,100 @@ describe('DataStore sync engine', () => {
...
+			// This is working in integ tests. Need to dive on why
+			// fake graphql service isn't handling Predicates.All.
+			// expect(records.length).toBe(5);
+
+			// leaving test in to validate the type.
+		});
+
+		test('basic contains() filtering', async () => {
+			await resyncWith([
+				syncExpression(Post, post => post?.title.contains('cleaning')),
```
stocaaro, line 304: Per your comment in the title, it would be really nice to get away from optionals in the top level interface. Begs the questions "When would the input predicate be `undefined`?", which seems like a big ask to pass along to our customers.

All that said, only those applying ts-strict at the highest levels (eg. Angular users and those that opt in) are going to encounter this required optional syntax, so it seems like a fine surface to launch with.

stocaaro (approved):
LGTM

packages/datastore/__tests__/connectivityHandling.test.ts
svidgen: Agreed. I struggled for longer that I care to admit trying to remove this little `?`. I'd be eager to pair with a TS guru to sort this one out at some point.

PR #10658, opened 2022-11-10
fix(datastore): upgrade after destructive schema change

iartemiev:
#### Checklist

- [ ] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/indexeddb.upgrade.test.ts
+++ b/packages/datastore/__tests__/indexeddb.upgrade.test.ts
@@ -1,0 +1,121 @@
+import Dexie from 'dexie';
+import 'dexie-export-import';
+import 'fake-indexeddb/auto';
+import * as idb from 'idb';
+import {
+	initSchema,
+	ModelInit,
+	MutableModel,
+	PersistentModelConstructor,
+	DataStore,
+} from '../src/index';
+import { Schema } from '../src/types';
+
+const DB_VERSION = 3;
+
+let db: idb.IDBPDatabase;
+const indexedDB = require('fake-indexeddb');
+const IDBKeyRange = require('fake-indexeddb/lib/FDBKeyRange');
+Dexie.dependencies.indexedDB = indexedDB;
+Dexie.dependencies.IDBKeyRange = IDBKeyRange;
+
+/* 
+	This schema has only one of the models that exists in the v1 schema we're upgrading from
+	below './v1schema.data.json'
+
+	Previously, this would cause an error in the IDB upgrade process
+	This test validates that the db upgrade works as expected
+*/
+const schema: Schema = {
+	models: {
+		Blog: {
+			syncable: true,
+			name: 'Blog',
+			pluralName: 'Blogs',
+			fields: {
+				id: {
+					name: 'id',
+					isArray: false,
+					type: 'ID',
+					isRequired: true,
+					attributes: [],
+				},
+			},
+			attributes: [
+				{
+					type: 'model',
+					properties: {},
+				},
+			],
+		},
+	},
+	enums: {},
+	nonModels: {},
+	version: 'a66372d29356c40e7cd29e41527cead7',
+	codegenVersion: '3.2.0',
+};
+
+declare class BlogModel {
+	readonly id: string;
+	readonly name: string;
+	constructor(init: ModelInit<BlogModel>);
+	static copyOf(
+		source: BlogModel,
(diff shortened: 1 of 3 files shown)


packages/datastore/__tests__/indexeddb.upgrade.test.ts
```diff
@@ -0,0 +1,121 @@
...
+		}
+
+		const exportedJSON = await readBlob(exportedBlob);
+		const exported = JSON.parse(exportedJSON);
+
+		for (const { schema } of exported.data.tables) {
+			expect(schema.split(',')).toContain('&[id]');
+		}
+
+		expect(exported).toMatchSnapshot(`v${DB_VERSION}-destructive-schema`);
```
svidgen, line 119: Checking indexedDB is OK. But, can we ensure DataStore is actually operational against the database post-upgrade?

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -182,6 +182,16 @@ class IndexedDBAdapter implements Adapter {
...
 
+									const storeExists = !!Object.values(
+										this.schema.namespaces[namespaceName].models
+									)?.find(m => m.name === modelName);
+
+									if (!storeExists) {
+										// delete original
+										db.deleteObjectStore(tmpName);
+										continue;
+									}
```
manueliglesias, line 193:
```suggestion
									const storeExists = !!this.schema.namespaces[namespaceName].models[modelName];

									if (!storeExists) {
										// delete original
										db.deleteObjectStore(origStore.name);
										continue;
									}
```

packages/datastore/__tests__/indexeddb.upgrade.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+	test(`Migration from v1 to v${DB_VERSION}`, async () => {
+		const v1Data = require('./v1schema.data.json');
+
+		const blob = new Blob([JSON.stringify(v1Data)], {
+			type: 'application/json',
+		});
+
+		// Import V1
+		(await Dexie.import(blob)).close();
```
svidgen, line 85: I understand that you TDD'd the fix, and maybe this is just my ignorance, but is there a sanity check we can add here to ensure Dexie loaded what we expect into indexedDB?

E.g., if someone starts adding tests and "tweaks" the dexie DI at the top, is there any possibility this test will test something *other* than what we expect it to?

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
manueliglesias: - `this.schema.namespaces[namespaceName].models`'s keys are the model names, no need to extract the values and iterate over them.
- `db.deleteObjectStore(origStore.name);` is IMO more clear in its intention

manueliglesias: `const storeExists = modelName in this.schema.namespaces[namespaceName].models;` would work too

elorzafe (approved):
LGTM, 

maybe @manueliglesias suggestion are worth to use to make it more clear, what do you think @iartemiev

manueliglesias (dismissed):
LGTM

svidgen (dismissed):
Agree with @manueliglesias's comments. And, I think I'd like to see some extra protection against erosion of the test in the future. But, it looks like it'll do the job. If addressing the asks are very time-consuming, I'm favor shipping sooner rather than later on this one.

packages/datastore/__tests__/indexeddb.upgrade.test.ts
iartemiev: Yea, can get by one of the items from the v1 store

packages/datastore/__tests__/indexeddb.upgrade.test.ts
iartemiev: If we add tests to this, we'll want to do some cleanup in a before/after. I'll add a comment for future reference

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
iartemiev: Yep, I ended up doing the `modelName in` approach - updated

elorzafe (approved):
Thanks @iartemiev

manueliglesias (approved):
LGTM 👍

PR #10679, opened 2022-11-16
fix: predicate logic against bools

svidgen:
#### Description of changes

Predicate filtering logic (used by `observe` and `observeQuery`) was not working correctly against bools, because we were casting field values to strings before comparison. This appears to be cruft from POC work &mdash; I can't otherwise tell why this cast would have been in place.

Added tests that were red. Made the change. Tests turned green.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/DataStore.ts
+++ b/packages/datastore/__tests__/DataStore.ts
@@ -1733,12 +1733,14 @@ describe('Model behavior', () => {
 		expect(await disconnectedParent.child).toBeUndefined();
 	});
 
-	test('removes items update out of the from the snapshot with an eq predicate on boolean field', done => {
+	test('removes no-longer-matching items from the snapshot when using an eq() predicate on boolean field', done => {
 		(async () => {
 			const { DataStore, ModelWithBoolean } = getDataStore();
 			try {
+				// the number of records we expect in each snapshot
 				const expecteds = [5, 4];
 
+				// initial data set, 5 records that will match our predicate.
 				for (let i = 0; i < 5; i++) {
 					await DataStore.save(
 						new ModelWithBoolean({
@@ -1750,10 +1752,10 @@ describe('Model behavior', () => {
 				const sub = DataStore.observeQuery(ModelWithBoolean, m =>
 					m.boolField.eq(true)
 				).subscribe(({ items, isSynced }) => {
+					// we don't actually expect 0 records in our snapshots after our list runs out.
+					// we just want to make TS happy.
 					const expected = expecteds.shift() || 0;
 					expect(items.length).toBe(expected);
-
-					console.log({ items });
 
 					for (let i = 0; i < expected; i++) {
 						expect(items[i].boolField).toEqual(true);
@@ -1765,12 +1767,16 @@ describe('Model behavior', () => {
 					}
 				});
 
+				// update an item to no longer match our criteria.
+				// we want to see a snapshot come through WITHOUT this item.
 				const itemToUpdate = (await DataStore.query(ModelWithBoolean)).pop()!;
 				await DataStore.save(
 					ModelWithBoolean.copyOf(itemToUpdate, m => {
 						m.boolField = false;
 					})
 				);
+
+				// advance time to trigger another snapshot.
 				jest.advanceTimersByTime(2000);
 			} catch (error) {
 				done(error);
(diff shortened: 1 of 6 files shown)


packages/datastore/__tests__/helpers.ts
```diff
@@ -1478,6 +1482,20 @@ export declare class LegacyJSONComment {
 	): LegacyJSONComment;
 }
 
+export declare class ModelWithBoolean {
+	public readonly id: string;
+	public readonly boolField: boolean;
```
svidgen, line 1487:
```suggestion
	public readonly boolField?: boolean;
```

packages/datastore/__tests__/model.ts
```diff
@@ -138,6 +138,7 @@ declare class EditorModel {
 
 declare class AuthorModel {
 	readonly id: string;
+	readonly isActive: boolean;
```
svidgen, line 141:
```suggestion
	readonly isActive?: boolean;
```

iartemiev (approved):
LGTM 👍🏻

dpilch (changes requested):
Can you add tests for `eq` and `ne` on number fields. I found that casting to string also affects numbers. https://github.com/dpilch/predicate-tests

packages/datastore/__tests__/DataStore.ts
```diff
@@ -1731,6 +1732,96 @@ describe('Model behavior', () => {
 		expect(disconnectedParent.hasOneParentChildId).toEqual(child.id);
 		expect(await disconnectedParent.child).toBeUndefined();
 	});
+
+	test('removes items update out of the from the snapshot with an eq predicate on boolean field', done => {
```
stocaaro, line 1736: The wording on this test  description is hard to parse.

packages/datastore/__tests__/DataStore.ts
```diff
@@ -1731,6 +1732,96 @@ describe('Model behavior', () => {
...
+		(async () => {
+			const { DataStore, ModelWithBoolean } = getDataStore();
+			try {
+				const expecteds = [5, 4];
+
+				for (let i = 0; i < 5; i++) {
```
stocaaro, line 1742: Why the 5 iterations here? I suppose we're just making 5 discreet records?

dpilch (approved):

packages/datastore/__tests__/DataStore.ts
```diff
@@ -1731,6 +1732,96 @@ describe('Model behavior', () => {
...
+						new ModelWithBoolean({
+							boolField: true,
+						})
+					);
+				}
+
+				const sub = DataStore.observeQuery(ModelWithBoolean, m =>
+					m.boolField.eq(true)
+				).subscribe(({ items, isSynced }) => {
+					const expected = expecteds.shift() || 0;
```
stocaaro, line 1753: So, we get a sequence of true models where the first one has 5 items and the second one has 4 items, then every event after that has 0 items?

I'm not quite sure how these item counts come out this way.

iartemiev (approved):

packages/datastore/__tests__/DataStore.ts
```diff
@@ -1731,6 +1732,96 @@ describe('Model behavior', () => {
...
+
+					console.log({ items });
+
+					for (let i = 0; i < expected; i++) {
+						expect(items[i].boolField).toEqual(true);
+					}
+
+					if (expecteds.length === 0) {
+						sub.unsubscribe();
+						done();
```
stocaaro, line 1764: Got it, so after the second event arrives, exit the test.

packages/datastore/__tests__/DataStore.ts
```diff
@@ -1731,6 +1732,96 @@ describe('Model behavior', () => {
...
+					})
+				);
+				jest.advanceTimersByTime(2000);
+			} catch (error) {
+				done(error);
+			}
+		})();
+	});
+
+	test('removes items update out of the from the snapshot with a ne predicate on boolean field', done => {
```
stocaaro, line 1781: This test is very similar. Could some of the setup logic be moved to a beforeFilter?  Would this make the tests easier or harder to follow?

packages/datastore/__tests__/DataStore.ts
svidgen: Yes.

packages/datastore/__tests__/DataStore.ts
svidgen: I'm going to think about this one for followup. If I can templatize these `observeQuery` tests, that would be awesome and make it a bit easier to build more of them up if-needed. I'm not 100% sure how this would affects readability yet. But, I'm struggling to think of how to do this without making things harder to trace through ATM.

packages/datastore/__tests__/DataStore.ts
svidgen: Sprinkled a few comments throughout that hopefully help narrate the test. 🤷

dpilch (approved):

stocaaro (approved):

iartemiev (approved):

Issue #10718, opened 2022-11-28
DataStore cannot manage required belongsTo/hasOne's using SQLiteAdapter in V5


svidgen:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

DataStore

### Amplify Categories

api

### Environment information

<details>

Discovered in unit test suite for `datastore-storage-adapters`.

</details>

### Describe the bug

Create a model with a required related-model field. The MIPR from this model was imported into the test helpers, with tests added to the common adapter tests.

```graphql
type MtmLeft @model {
  id: ID! @primaryKey
  content: String
  rightOnes: [MtmRight] @manyToMany(relationName: "mtmJoin")
}

type MtmRight @model {
  id: ID! @primaryKey
  content: String
  leftOnes: [MtmLeft] @manyToMany(relationName: "mtmJoin")
}
```

When attempting to save a join record `MtmJoin`, the operation fails with required field errors.

This appears to occur because the SQLiteAdapter creates columns for all fields, including the derived columns. However, SQLite doesn't have an `undefined` state, on re-serialization it attempts to set the related-model fields to `null`, triggering the error.

E.g., this comes back from the SQLite adapter is in put to instantiate `MtmJoin`:

```
  console.log ../datastore/lib/datastore/datastore.js:495
    initializeInstance {
      id: '82d542d9-9386-4deb-b517-51b98df27c56',
      mtmLeftId: '441012b4-6c7c-418f-8d46-7d7ca5347f4a',
      mtmRightId: 'b599ebd7-dc4d-403e-b245-5959987b823c',
      mtmLeft: null,
      mtmRight: null,
      createdAt: null,
      updatedAt: null,
      _version: null,
      _lastChangedAt: null,
      _deleted: null
    }
```

### Expected behavior

The SQLite Adapter should not attempt to populate related-model fields at all during `load`. It should only populate value fields (including FK's) and allow DataStore to lazy-load related models as designed.

### Reproduction steps

[...]

PR #10720, opened 2022-11-28
fix(@aws-amplify/datastore-storage-adapter): SQLiteAdapter fails on required related models

svidgen:
#### Description of changes

Adding `manyToMany` tests revealed that the SQLiteAdapter was failing to handle required related models correctly. As it was loading from the database, the adapter was attempting to initialize records with `null` values for the required related-model fields, rather than allowing these fields to be lazy-loaded using the given FK's. This was causing operations against models with required `belongsTo` and `hasOne` fields to throw errors, due to the fields being initialized `null`.

I'm not completely sure why these fields are present in the SQLite schema to begin with, so I'm leaving them in for now. However, this change fixes the issue by filtering related model fields out of the initialization entirely. Only the FK's should be populated.

#### Issue #, if available

#10718

Added unit tests. Red/Green'd the tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/__tests__/commonAdapterTests.ts
```diff
@@ -52,6 +56,10 @@ export function addCommonQueryTests({
 	getMutations: any;
 	clearOutbox: any;
 }) {
+	beforeAll(() => {
+		DataStore.configure({ storageAdapter });
+	});
```
svidgen, line 61:
```suggestion
	// This MUST be in a `beforeAll` or `beforeEach` scoped at this level to
	// ensure all down-the-line tests use the given adapter. It is otherwise
	// VERY EASY to forgot to include this, which can cause some of these tests
	// to run against a different adapter and thus test the wrong things.
	beforeAll(() => {
		DataStore.configure({ storageAdapter });
	});
```

packages/datastore/__tests__/commonAdapterTests.ts
```diff
@@ -686,6 +690,11 @@ export function addCommonQueryTests({
 		let modelNamesToTest = Object.keys(schema.models);
 		if (isSQLiteAdapter()) {
 			modelNamesToTest = modelNamesToTest.filter(n => n.startsWith('Legacy'));
+		} else {
+			// Our manyToMany example requires both side of the relationship to be
+			// populated, so we need a different testing pattern.
+			// also notably, this is "extra" test work. MTM is just two hasMany's.
```
svidgen, line 696:
```suggestion
			// Originally thought to be "extra" testing, but showed that we were
			// not previously testing *required* relationships.
```

packages/datastore/__tests__/commonAdapterTests.ts
```diff
@@ -1059,4 +1068,240 @@ export function addCommonQueryTests({
 			}
 		}
 	});
+
+	/**
+	 * Given the way `manyToMany` works, these tests are *mostly* superfluous.
+	 * `manyToMany` is just a helper to create `belongsTo`/`hasMany` models. So,
```
svidgen, line 1074:
```suggestion
	 * (They *do* however stress *required* relationship handling. A prior gap.)
	 * `manyToMany` is just a helper to create `belongsTo`/`hasMany` models. So,
```

iartemiev (approved):
LGTM 👍🏻

dpilch (approved):

Issue #9134, opened 2021-10-31
Return code delivery details list after updateUserAttributes


jvliwanag:
### Is this related to a new or existing API?

Authentication

### Is this related to another service?

Cognito

### Describe the feature you'd like to request

UpdateUserAttributes when called directly on the Cognito API returns a code delivery details list in its response as documented [here](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_UpdateUserAttributes.html#API_UpdateUserAttributes_ResponseElements). It would be useful to expose these details directly on the API for any UI bindings to direct to the respective verification step.

### Describe the solution you'd like

Return codeDeliveryDetails on updateUserAttributes API.

### Describe alternatives you've considered

Guess if code verification is required based on prior knowledge of the user pool configuration.

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

sammartinez:
Thanks for opening up this feature request @jvliwanag. Can you provide a little more information on your use case. We do have this functionality built into our UI components with sending the event to a `ConfirmSignUp` which will change the UI state to the confirmation code section. Currently, the code is sent via email for users to put into.

jvliwanag:
Hi @sammartinez . The code delivery details list I believe is the indicator what, if any, are the required confirmation steps right after calling updateAttributes. If there are cognito lambda hooks to auto verify, then there's no need to prompt the user for verification. If there are none, then updating special fields such as email/mobile can put the identifiers to an awaiting verification state, which the client can know of through the code delivery details list.

nolde:
I've made a PR to support this partially, allowing it to pass though `amazon-cognito-identity-js`. Further work would be needed to allow it to pass through `@aws-amplify/auth`.

helgabalashova:
Hi @jvliwanag,
we've added a feature that can help you identify what attributes require verification. You can read more about it in our docs in [this](https://docs.amplify.aws/lib/auth/manageusers/q/platform/js/#update-user-attributes) section. Let us know if you have further questions.

nolde:
Hello @helgabalashova ,

Why not merge my PR in? It just returns the details straight in from the function call. It shouldn't conflict with the hub events, and it currently returns void, so no one is expecting it anyway.

helgabalashova:
Hi @nolde,
[...]

PR #10731, opened 2022-12-01
feat(auth): adds hub events to updateUserAttributes api

helgabalashova:
#### Description of changes
adds hub event to `Auth.updateUserAttributes` to determine if verification of the attribute is required. Feature requested by UI team. Decision to implement it using Hub events was made by Auth and UI team to avoid breaking change. 

#### Issue #, if available
#9134 

#### Checklist

- [X] PR description included
- [x] `yarn test` passes


packages/auth/src/Auth.ts
```diff
@@ -1425,16 +1426,55 @@ export class AuthClass {
...
 						}
 					},
 					clientMetadata
 				);
 			});
 		});
 	}
+
+	private createUpdateAttributesResultList(
+		attributes: object,
```
israx, line 1444: `object` is a JS type that can be either an `array` , `function` and `map`. I think is better to use a more constrained type like `Record<string, string>`

packages/auth/src/Auth.ts
```diff
@@ -1425,16 +1426,55 @@ export class AuthClass {
...
 					clientMetadata
 				);
 			});
 		});
 	}
+
+	private createUpdateAttributesResultList(
+		attributes: object, 
+		codeDeliveryDetailsList?: CodeDeliveryDetails []
+	): object {
```
israx, line 1446: same above

packages/auth/src/Auth.ts
```diff
@@ -1425,16 +1426,55 @@ export class AuthClass {
...
+		const attrs = {};
+		Object.keys(attributes).map(key => {
```
israx, line 1448: I think using `forEach` is the best option as `map` will return a new array

packages/auth/src/Auth.ts
```diff
@@ -1425,16 +1426,55 @@ export class AuthClass {
...
+			attrs[key] = {
+				isUpdated: true
+			};
+			if (codeDeliveryDetailsList) {
+				const codeDeliveryDetails = this.findCodeDeliveryDetails(key, codeDeliveryDetailsList);
```
israx, line 1453: I thinks there's no need to create another function, using `find` will do the job. 

```
const codeDeliveryDetails = codeDeliveryDetailsList.find(codeDeliveryDetails.AttributeName === key)
```

packages/auth/src/Auth.ts
```diff
@@ -1425,16 +1426,55 @@ export class AuthClass {
...
+							return resolve('SUCCESS');
 						}
 					},
 					clientMetadata
 				);
 			});
 		});
 	}
+
+	private createUpdateAttributesResultList(
```
israx, line 1443: I think this function should return a `Promise` as it could be expensive due to the nested `array` methods

packages/auth/src/Auth.ts
```diff
@@ -1425,16 +1426,55 @@ export class AuthClass {
 					attributeList,
 					(err, result) => {
 						if (err) {
+							dispatchAuthEvent('updateUserAttributes_failure', err, 'Failed to update attributes');
 							return reject(err);
 						} else {
-							return resolve(result);
+							const attrs = this.createUpdateAttributesResultList(attributes, result.CodeDeliveryDetailsList);
```
israx, line 1432: it should `await` as the comment bellow

israx (changes requested):

packages/auth/src/Auth.ts
haverchuck: @israx How exactly does returning a promise help with the expensiveness of this function?

helgabalashova: I personally don't think it is expensive. even if they update 100 attributes (which I doubt), it will still be super fast

haverchuck: @israx Running it async doesn't make it any less expensive - it is still all running on a single thread. So what is the real benefit of making this a Promise?

haverchuck (changes requested):
Can we create and/or update tests?

AllanZhengYP (dismissed):
Looks good to me! ⛴

packages/amazon-cognito-identity-js/src/CognitoUser.js
```diff
@@ -1125,13 +1125,13 @@ export default class CognitoUser {
...
 				ClientMetadata: clientMetadata,
 			},
-			err => {
+			(err,result) => {
 				if (err) {
 					return callback(err, null);
 				}
 
 				// update cached user
-				return this.getUserData(() => callback(null, 'SUCCESS'), {
```
haverchuck, line 1134: What is in the result object? If it is anything other than 'SUCCESS', this would actually be breaking change for anyone using `amazon-cognito-identity-js`, right?

Could we just continue to return the same thing here?

packages/auth/src/Auth.ts
elorzafe: This is a good discussion, I don't have the right response but just to add more data here.

The code is running async because is inside of a callback, the callback will run on the event loop when the `user.updateAttributes` invoke that.  If I understand correctly at some point that would be taking the single thread and in case there is a huge object I would expect to use something that split this in chunks and release the thread, otherwise even if you wrap that with a `Promise` will still take the thread until is completed.

I think that would be a very advance scenario which in that case I would solve it if a customer has a really huge attribute list which I think Cognito service doesn't support right now.

packages/auth/src/Auth.ts
```diff
@@ -1436,6 +1443,28 @@ export class AuthClass {
...
+		attributes: Record<string, string>, 
+		codeDeliveryDetailsList?: CodeDeliveryDetails []
+	): Record<string, string> {
+		const attrs = {};
+		Object.keys(attributes).forEach(key => {
+			attrs[key] = {
+				isUpdated: true
+			};
+			if (codeDeliveryDetailsList) {
+				const codeDeliveryDetails = codeDeliveryDetailsList.find(value => value.AttributeName === key);
```
elorzafe, line 1457: This could be shorter if using optional chaining instead?

`if (codeDeliveryDetails?.find(....)) {`

elorzafe (changes requested):
Can you add the change on the `d.ts` file because is a public facing API, I also have a nit comment.

Overall looks good.
Thanks @helgabalashova

packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -217,16 +217,16 @@ declare module 'amazon-cognito-identity-js' {
 			}
 		): void;
 		public getUserAttributes(
-			callback: NodeCallback<Error, CognitoUserAttribute[]>
+			callback: NodeCallback<Error, CognitoUserAttribute[], any>
 		): void;
 		public updateAttributes(
 			attributes: (CognitoUserAttribute | ICognitoUserAttributeData)[],
-			callback: NodeCallback<Error, string>,
+			callback: NodeCallback<Error, string, any>,
```
elorzafe, line 224: I think we only needed the update for the this one, not for the others

packages/auth/src/Auth.ts
```diff
@@ -1453,14 +1453,11 @@ export class AuthClass {
...
-			if (codeDeliveryDetailsList) {
-				const codeDeliveryDetails = codeDeliveryDetailsList.find(value => value.AttributeName === key);
-				if (codeDeliveryDetails) {
-					attrs[key].isUpdated = false;
-					attrs[key].codeDeliveryDetails = codeDeliveryDetails;
-				}
+			const codeDeliveryDetails = codeDeliveryDetailsList?.find(value => value.AttributeName === key);
+			if (codeDeliveryDetails) {
+				attrs[key].isUpdated = false;
+				attrs[key].codeDeliveryDetails = codeDeliveryDetails;
```
elorzafe, line 1459: 👍

elorzafe (changes requested):
Check the `d.ts` file, I would create another type for `updateUserAttribute` callback

jimblanc (approved):

haverchuck (dismissed):

israx (dismissed):
lgtm

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
```diff
@@ -1657,4 +1657,73 @@ describe('refreshSession()', () => {
...
+			   { 
+				  'AttributeName': 'email',
+				  'DeliveryMedium': 'EMAIL',
+				  'Destination': 'e***@e***'
+			   }
+			]
+		};
+		const spyon = jest.spyOn(CognitoUser.prototype, 'updateAttributes')
+			.mockImplementationOnce((attrs, callback) => {
+				callback(null, 'SUCCESS', codeDeliverDetailsResult);
```
elorzafe, line 1673: This mock is not testing anything in particular, is just testing the mock. What you should do here is mock the cognito client instead. Like this https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js#L384

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
```diff
@@ -1657,4 +1657,73 @@ describe('refreshSession()', () => {
...
```
elorzafe, line 1707: Same as the other, this is testing the mock not the actual code look on this one 
https://github.com/aws-amplify/amplify-js/blob/main/packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js#L384

packages/amazon-cognito-identity-js/index.d.ts
```diff
@@ -2,6 +2,7 @@ declare module 'amazon-cognito-identity-js' {
 	//import * as AWS from "aws-sdk";
 
 	export type NodeCallback<E, T> = (err?: E, result?: T) => void;
+	export type UpdateAttributesNodeCallback<E, T, K> = (err?: E, result?: T, details?: K) => void;
```
elorzafe, line 5: nice

packages/amazon-cognito-identity-js/__tests__/CognitoUser.test.js
elorzafe: For testing there are no regressions for the version that has two arguments we need to have a TypeScript test that validates is not introducing a regression. I will approve the PR but before launching to `latest` we need the integ test in place

elorzafe (dismissed):
Add the integ test to validate the 2 and 3 version arguments before launching this to `latest`

elorzafe:
The commit message should be 

feat(@aws-amplify/auth,amazon-cognito-identity-js): return code delivery details for updateUserAttributes on Hub event

israx (approved):

PR #10734, opened 2022-12-02
feature(@aws-amplify/datastore): IndexedDB query, nested predicate optimization

svidgen:
#### Description of changes

This PR updates automatic index leveraging in IndexedDB initially written for nested predicates pre-CPK to work with composite keys. The performance difference is negligible for small datasets, but scales very poorly without this update on larger data sets and with the number of joins performed.

This PR also includes a fix to composite key join condition generation discovered during implementation, wherein CPK join conditions were OR'd together instead of AND'd together, _potentially_ resulting in over-selection. (Note that this is only a theoretical bug at this point. No over-selection was actually observed, and existing CPK join tests exist and are all passing in `main`. The observed buggy behavior was that the query optimization step failed to "condense" these conditions into a single storage predicate. This is why I chose to include the fix in this PR instead of a separate fix.) To remedy, I swapped the join-condition generation logic to use the easier-to-read, and more-easily-validated _AST to storage predicate_ translation function.

Created benchmarking tests. Measured query times with and without tests in the unit test suite, noting an order-of-magnitude improvement (~150ms down to ~20ms) on a set of 12x nested queries against a pair of tables, each containing 250 records. (This against our *in-memory* IndexedDB fake.)

I also temporarily added logging and manually inspected key interactions to ensure queries were being generated as expected and that indexes were being found and properly selected for those queries.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/src/predicates/next.ts
```diff
@@ -351,6 +359,82 @@ export class GroupCondition {
...
+	 *
+	 * This allows `fetch()` to pass both the `id` and `name` conditions to the adapter
+	 * together, which can then decide what index to use based on both fields together.
+	 *
+	 * @param preserve Whether to preserve the current node and to explicitly not eliminate
+	 * it during optimization. Used internally to preserve the root node and children of
+	 * `not` groups. `not` groups will always have a single child, so there's nothign to
+	 * optimize below a `not` (for now), and it makes the query logic simpler later.
+	 */
+	optimized(preserve = true): UntypedCondition {
```
david-mcafee, line 398: Considering if `preserve` could have a slightly more explicit name? `preserveNode`, for instance. On the fence.

packages/datastore/src/predicates/next.ts
```diff
@@ -351,6 +359,82 @@ export class GroupCondition {
...
+		const operands = this.operands.map(o =>
+			o instanceof GroupCondition ? o.optimized(this.operator === 'not') : o
```
david-mcafee, line 400: There are a several places where you check against `not`, `and`, and `or`, and wondering if helper functions make sense here to 1) make the condition a little more explicit, and 2) to make the checks reusable (and future updates to the check would only need to be updated in one place).

packages/datastore/src/predicates/next.ts
```diff
@@ -434,34 +522,26 @@ export class GroupCondition {
 				const relationship = ModelRelationship.from(this.model, g.field);
 
 				if (relationship) {
-					const relativesPredicates: ((
-						p: RecursiveModelPredicate<any>
-					) => RecursiveModelPredicate<any>)[] = [];
-					for (const relative of relatives) {
-						const individualRowJoinConditions: FieldCondition[] = [];
+					// type T isn't known here.
+					const allJoinConditions: any = [];
```
david-mcafee, line 526: Is a better type possible here? Same for line 529.

packages/datastore/src/predicates/next.ts
```diff
@@ -434,34 +522,26 @@ export class GroupCondition {
...
+						const relativeConditions: any = [];
 						for (let i = 0; i < relationship.localJoinFields.length; i++) {
-							// rightHandValue
-							individualRowJoinConditions.push(
-								new FieldCondition(relationship.localJoinFields[i], 'eq', [
-									relative[relationship.remoteJoinFields[i]],
-								])
-							);
+							relativeConditions.push({
+								[relationship.localJoinFields[i]]: {
```
david-mcafee, line 532: Is there a reason we're not updating the `FieldCondition` constructor instead, here?

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -481,106 +481,176 @@ class IndexedDBAdapter implements Adapter {
...
+		for (const predicate of predicates) {
+			predicateIndex.set(String(predicate.field), predicate);
+		}
+
 		const store = transaction.objectStore(storeName);
 		for (const name of store.indexNames) {
 			const idx = store.index(name);
-			if (idx.keyPath === fieldName) {
-				return idx;
+			const keypath = Array.isArray(idx.keyPath) ? idx.keyPath : [idx.keyPath];
```
david-mcafee, line 509: Why is it sometimes an array, and other times not? That seems potentially smelly.

david-mcafee (changes requested):

packages/datastore/src/predicates/next.ts
svidgen: I could see some value in creating semantically named functions, especially if it allows me to strike comments like the one on the conditional below. But, I'm not sure what that looks like yet. I'm a little concerned that addressing this would result in either zero added clarity at the call sites and no LOC saved **or** lots of added clarity at the call sites, but lots of similar/redundant methods.

I've created a task on the backlog internally to revisit this one. If you have a more concrete suggestion in the meantime, let's discuss!

packages/datastore/src/predicates/next.ts
svidgen: Whoops. I was lazy in the moment and failed to circle back. I'll see what I can do!

packages/datastore/src/predicates/next.ts
svidgen: I may be having a dense moment. What do you mean? What do you want to see?

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -481,106 +481,176 @@ class IndexedDBAdapter implements Adapter {
 		return keyValues.length === keyPath.length ? keyValues : undefined;
```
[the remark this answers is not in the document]
svidgen: It might be worth putting a story in the backlog. But, my impression when I did the *original* optimization was that AsyncAdapter was on the way out. And since the SQLiteAdapter will handle these optimizations in the SQL engine layer, they're not needed there. So, IndexedDB was the only one I assumed we needed at the time.

Happy to revisit it separately.

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
svidgen: I might have traced through things incorrectly to arrive at this point, but IIRC this accounts for the Safari implementation discrepancy that we had to fix urgently a few months back. See [this method](https://github.com/aws-amplify/amplify-js/blob/a85171384db4bca202bb5cb5f832a8e319ec89d4/packages/datastore/src/storage/adapter/IndexedDBAdapter.ts#L1072-L1086).

packages/datastore/src/predicates/next.ts
```diff
@@ -665,7 +756,7 @@ export function recursivePredicateFor<T extends PersistentModel>(
 	// next steps will be to add or(), and(), not(), and field.op() methods.
 	const link = {} as any;
 
-	registerPredicateInternals(baseCondition, link);
+	const baseQueryInternals = registerPredicateInternals(baseCondition, link);
```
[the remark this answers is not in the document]
svidgen: Oh, correct! I think I'd started down a refactor path that I failed to fully abandon. Cleaning it up...

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -481,106 +481,176 @@ class IndexedDBAdapter implements Adapter {
...
+	private async baseQueryIndex<T extends PersistentModel>(
 		storeName: string,
 		predicates: PredicatesGroup<T>
 	) {
 		let { predicates: predicateObjs, type } = predicates;
 
 		// the predicate objects we care about tend to be nested at least
 		// one level down: `{and: {or: {and: { <the predicates we want> }}}}`
-		// so, we unpack and/or groups until we find a group with more than 1
+		// and these containers
```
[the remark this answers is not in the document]
svidgen: Your guess is as good as mine here at this point. I'll revert this bit. 😅

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -505,82 +546,109 @@ class IndexedDBAdapter implements Adapter {
...
+			// for now. it's higher complexity than is worth tackling for now.
+			if (predicateObjs.length > fieldPredicates.length) {
+				result = {
+					groupType: null,
+					indexedQueries: [] as (() => Promise<T[]>)[],
+				};
+			} else {
+				result = {
+					groupType: 'or',
+					indexedQueries: fieldPredicates.reduce((agg, p) => {
```
dpilch, line 585: nit: flatMap may be a little more clear here

dpilch (approved):

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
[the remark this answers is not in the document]
david-mcafee: Could we get an item in the backlog, in the event the Async Adapter sticks around?

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
david-mcafee: Tell me if this is overkill, but is that worth calling out, here? I think DataStore in general has been prone to a lot of these kinds of historical, unexplainable, spaghetti situations. Ideally, when / if the Safari implementation is fixed, we could remove this.

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
svidgen: I'm digging on this a little more. Now that I'm looking at it again, this actually looks like a test coverage gap I'll want to fill before changing (and merging).

packages/datastore/src/predicates/next.ts
david-mcafee: Previously, we had:

```ts
individualRowJoinConditions.push(
    new FieldCondition(relationship.localJoinFields[i], 'eq', [
	    relative[relationship.remoteJoinFields[i]],
    ])
);
```

which was, with this PR, replaced with:

```ts
relativeConditions.push({
    [relationship.localJoinFields[i]]: {
        eq: relative[relationship.remoteJoinFields[i]],
    },
});
```

So we've removed the call to [the `FieldCondition` constructor](https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/src/predicates/next.ts#L141). By not using the constructor, we miss out on [the validation step](https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/src/predicates/next.ts#L213). 

It's also possible I may be having a dense moment?

svidgen: Oh I see. Good question!

The `FieldCondition` validation is intended primarily to validate customer predicate construction at runtime (mainly JS contexts) and provide meaningful guidance if they supply something like this:

```
await DataStore.query(Model, m => m.name.eq('a', 'b', 'c'))
```

In this example, we don't have a defined interpretation of what `eq` does when given multiple operands. So, the `FieldCondition` validator will issue an error intended to tell the customer what to fix.

In our case, we're building the query in a predictable fashion from existing data + metadata, so I wasn't concerned about losing this validation. And it therefore seemed simpler to skip the object construction.

But, I don't think there's any reason I couldn't drop the constructor back in as an intermediary:

```
	relativeConditions.push(new FieldCondition(
		relationship.localJoinFields[i],
		'eq',
		relative[relationship.remoteJoinFields[i]]
	).toAST());
```

I'm happy to make that change, unless I'm forgetting something!

packages/datastore/__tests__/IndexedDBAdapter.test.ts
```diff
@@ -289,3 +293,265 @@ describe('IndexedDBAdapter tests', () => {
...
+	afterEach(async () => {
+		await DataStore.clear();
+	});
+
+	const benchmark = async (f, iterations = 100) => {
+		const start = new Date();
+		for (let i = 0; i < iterations; i++) {
+			await f();
+		}
+		const end = new Date();
```
iartemiev, line 331: Nit: can use [peformance.now()](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#examples) for improved benchmark accuracy

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
svidgen: This is actually a little more involved than I realized. As it's pretty much EOD here, this will take me into next week to keep digging. But, here's where I'm at so far:

Join conditions are coming through with additional levels of nesting that I failed to account for. This may just be a CPK thing -- TBD on that -- I haven't actually probed into that yet. But, I'm going to try to make this branch a little smarter and unpack those join conditions one level deeper next week. I got a small start there, but am out of time for today.

I'll also timebox the effort to my "free" time this sprint. The changes in this branch reduce the number of table scans needed and execution times overall. But, but `or` branch isn't adding the intended value. So, I may end up having to trash the branch if I can't shim it in before the end of the sprint.

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
iartemiev: FWIW even when Safari fixes the IDB keypath bug, we'll have to keep our workarounds in place for a couple of years to ensure compatibility with older versions of Safari. I think adding a comment here should suffice for now, but if we have a chance to refactor this later, we should try to abstract any repeated workarounds out into single method like the one Jon linked above.

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -481,21 +481,62 @@ class IndexedDBAdapter implements Adapter {
...
+			for (const field of keypath) {
+				const p = predicateIndex.get(field);
+				if (p) {
+					matchingPredicateValues.push(String(p.operand));
+				} else {
+					break;
+				}
+			}
+
+			// if we have a matchind predicate field for each component of this index,
```
david-mcafee, line 521: nit:
```suggestion
			// if we have a matching predicate field for each component of this index,
```

david-mcafee (dismissed):
Left a question re: using the `FieldCondition` constructor, and a minor spelling suggestion, otherwise LGTM.

packages/datastore/src/predicates/next.ts
david-mcafee: Thanks for clarifying, Jon! In that case, seems there isn't any value in adding it back in :)

iartemiev (approved):
LGTM pending responses to the questions David raised in his review.

packages/datastore/__tests__/IndexedDBAdapter.test.ts
svidgen: 👍

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
svidgen: OK. I've added tests that stress this branch, which I reworked a bit to successfully trigger on CPK join conditions. In doing so, this particular line is replaced by a handful of maps and filters above.

david-mcafee (approved):
LGTM! 🎉

PR #10753, opened 2022-12-08
fix(core): add cache-control header to cognito identity client

haverchuck:
#### Description of changes
Adds the cache-control 'no-store' header via middleware to the various instantiations of the Cognito Identity Pool Client.

Please note: the `Credentials` file instantiates the cognito client multiple times - most likely needlessly. I've left this in place, though, to minimize the scope of the change.

Manual testing.

#### Checklist

- [X] PR description included
- [X] `yarn test` passes

### DIFF
--- a/packages/amazon-cognito-identity-js/src/Client.js
+++ b/packages/amazon-cognito-identity-js/src/Client.js
@@ -86,7 +86,7 @@ export default class Client {
 			headers,
 			method: 'POST',
 			mode: 'cors',
-			cache: 'no-store',
+			cache: 'no-cache',
 			body: JSON.stringify(params),
 		});
 
--- a/packages/amazon-cognito-identity-js/src/Client.js
+++ b/packages/amazon-cognito-identity-js/src/Client.js
@@ -49,15 +49,20 @@ export default class Client {
 	requestWithRetry(operation, params, callback) {
 		const MAX_DELAY_IN_MILLIS = 5 * 1000;
 
-		jitteredExponentialRetry((p) => new Promise((res, rej) => {
-			this.request(operation, p, (error, result) => {
-				if (error) {
-					rej(error);
-				} else {
-					res(result);
-				}
-			});
-		}), [params], MAX_DELAY_IN_MILLIS)
+		jitteredExponentialRetry(
+			p =>
+				new Promise((res, rej) => {
+					this.request(operation, p, (error, result) => {
+						if (error) {
+							rej(error);
+						} else {
+							res(result);
+						}
+					});
+				}),
+			[params],
+			MAX_DELAY_IN_MILLIS
+		)
 			.then(result => callback(null, result))
 			.catch(error => callback(error));
 	}
@@ -81,7 +86,7 @@ export default class Client {
 			headers,
 			method: 'POST',
 			mode: 'cors',
-			cache: 'no-cache',
+			cache: 'no-store',
 			body: JSON.stringify(params),
 		});
 
@@ -112,9 +117,9 @@ export default class Client {
 				// Taken from aws-sdk-js/lib/protocol/json.js
 				// eslint-disable-next-line no-underscore-dangle
 				const code = (data.__type || data.code).split('#').pop();
-				const error = new Error(data.message || data.Message || null)
-				error.name = code
-				error.code = code
+				const error = new Error(data.message || data.Message || null);
+				error.name = code;
+				error.code = code;
 				return callback(error);
 			})
 			.catch(err => {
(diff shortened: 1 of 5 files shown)


packages/core/src/Credentials.ts
```diff
@@ -296,6 +297,16 @@ export class CredentialsClass {
 			customUserAgent: getAmplifyUserAgent(),
 		});
 
+		cognitoClient.middlewareStack.add(
```
palpatim, line 300: Is there ever a case when we would not want the cache control header to be added? If not, could we just bake this into the CognitoClient constructor?

packages/core/src/Credentials.ts
```diff
@@ -296,6 +297,16 @@ export class CredentialsClass {
...
+			(next, _) => (args: any) => {
+				args.request.headers['cache-control'] = 'no-store';
```
jimblanc, line 302: Nit: Might want to do this immutably to keep the middleware side-effect free.

```
const updatedArgs = {
  ...args,
  request: {
    ...args.request,
    headers: { 
      ...args.request.headers,
      'cache-control': 'no-store'
    }
  }
};
```

Also might be worth breaking into a utility function to improve re-usability & test-ability?

packages/amazon-cognito-identity-js/src/Client.js
```diff
@@ -49,15 +49,20 @@ export default class Client {
 	requestWithRetry(operation, params, callback) {
 		const MAX_DELAY_IN_MILLIS = 5 * 1000;
```
haverchuck, line 51: Changes to this file were made by Prettier when I made a change. I reverted the change, but the prettification remains.

AllanZhengYP (dismissed):

packages/amazon-cognito-identity-js/src/Client.js
stocaaro: Missed that you made no changes. Might recommend checking out the original file to simplify your commit.

eg `git co origin/main packages/amazon-cognito-identity-js/src/Client.js && git commit -am 'chore: Revert prettier changes'`

cshfang: +1

stocaaro (approved):

packages/core/src/Util/CognitoIdentityClient.ts
```diff
@@ -0,0 +1,45 @@
...
+import { Provider } from '@aws-sdk/types';
+import { getAmplifyUserAgent } from '../Platform';
+
+/**
+ * Returns a CognitoIdentityClient with middleware
+ * @param {string} region
+ * @return {CognitoIdentityClient}
+ */
+export function createCognitoIdentityClient(
+	region: string | Provider<string> | undefined
```
cshfang, line 14: Nit: Use the type provided by SDK
```js
import { CognitoIdentityClient, CognitoIdentityClientConfig } from '@aws-sdk/client-cognito-identity';

...

region: CognitoIdentityClientConfig["region"]
```

cshfang (approved):
Minor comments otherwise lgtm

packages/core/__tests__/Util-test.ts
```diff
@@ -51,6 +57,60 @@ describe('Util', () => {
 		});
 	});
 
+	describe('cognito identity client test', () => {
+		test('client should be instantiated', async () => {
+			const cognitoClient = createCognitoIdentityClient('us-west-1');
+			expect(cognitoClient).toBeTruthy();
+		});
+
+		test('headers should be added by middleware on GetIdCommand', async () => {
```
elorzafe, line 66: I would add `expect.assertions` and also use `done` in case the tests ends before doing the assertions

packages/core/__tests__/Util-test.ts
```diff
@@ -51,6 +57,60 @@ describe('Util', () => {
...
+			});
+
+			return await client.send(
+				new GetIdCommand({
+					IdentityPoolId: 'us-west-1:12345678-1234-1234-1234-123456789000',
+				})
+			);
+		});
+
+		test('headers should be added by middleware on GetCredentialsForIdentityCommand', async () => {
```
elorzafe, line 89: Same as the other just for been careful the test is not finishing before doing the assertions I would add `expect.assertions` and use `done` callback

packages/core/src/Util/CognitoIdentityClient.ts
```diff
@@ -0,0 +1,45 @@
...
+		{
+			step: 'build',
+			name: 'cacheControlMiddleWare',
+		}
+	);
+
+	return client;
+}
+
+function middlewareArgs(args: { request: { headers: any }; input: any }) {
```
elorzafe, line 34: I would add a test for this function

elorzafe (changes requested):
What do you think of having an E2E for checking the headers instead of doing it on the unit tests? I would try to test the logic on unit tests but test the actual requests headers I would do it on cypress on a real app.

Thanks @haverchuck 🎖️

packages/core/__tests__/Util-test.ts
haverchuck: Yeah we should have both.

packages/core/src/Credentials.ts
haverchuck: The answer to your question is 'no'; and while we don't control the client constructor itself (it's an sdk construct) we've moved instantiation up into a separate utility that bakes in this middleware.

packages/core/__tests__/Util-test.ts
haverchuck: I don't think the `done` is necessary if you are using async callback and awaiting, but I added the expect.assertions. Could callout, thanks.

cshfang (approved):

AllanZhengYP (dismissed):

jimblanc (dismissed):

stocaaro (dismissed):

elorzafe (approved):

cshfang (approved):

stocaaro (approved):

AllanZhengYP (approved):

PR #10755, opened 2022-12-09
fix: avoid handing it as auth url when current url is not redirectSignIn

### LINKED ISSUE #9208: Having ?code= on the query string seems to always be caught by some Cognito code

fossamagna:
#### Description of changes

I prevent calling [this._handleAuthResponse(url);](https://github.com/aws-amplify/amplify-js/blob/4725b5f2df332c006571e87efdc6ebd5fb3dbe4e/packages/auth/src/Auth.ts#L249) if the url is not on OAuth configuration.

#### Issue #, if available

fix #9208

- `yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


fossamagna:
@stocaaro I found  message like `This workflow is awaiting approval from a maintainer in #10755` in CodeQL workflow.
https://github.com/aws-amplify/amplify-js/actions/runs/3907730242
Would you approve this workflow.

israx:
Hello @fossamagna. Thank you for submitting this PR. 

After doing some manual tests I see the issue is still on place. I think what could solve the problem is to match the current response after calling the [login](https://docs.aws.amazon.com/cognito/latest/developerguide/login-endpoint.html) end point. 

This is the current response that we want to match against:

 `Location: https://YOUR_APP/redirect_uri?code=AUTHORIZATION_CODE&state=STATE]`

In addition to the current changes, probably something like: 

`const isLoginEndPointResponse = url.includes('code') && url.includes('state');`

fossamagna:
@israx Thank you for your review and doing manual tests.  I corrected it in accordance with your suggestions.

packages/auth/src/Auth.ts
```diff
@@ -237,11 +237,25 @@ export class AuthClass {
 				cognitoClientId: cognitoAuthParams.cognitoClientId,
 			});
 
+			const isLoginEndPointResponse = (url: string) => {
+				return (
+					oauth &&
+					isCognitoHostedOpts(oauth) &&
+					url.startsWith(oauth.redirectSignIn) &&
+					url.includes('code') &&
```
haverchuck, line 245: Although the Authorization code grant is definitely preferred, I believe that we also [handle implicit grant](https://github.com/aws-amplify/amplify-js/blob/e1b0b5be3e8ccb3c76e8e2e2f43f910d40d73254/packages/auth/src/OAuth/OAuth.ts#L233), which doesn't seems to be considered here.

fossamagna: @haverchuck Thank you for your review. I add  consideration for implicit flow.

israx:
Hi @fossamagna. Thanks for keep reviewing this PR. 

I highly suggest testing these changes manually. I think we went back to the previous problem after testing this out with a react application.

We need to scope and match the right response coming from cognito. So, we need to consider the following scenarios:

1. Code grant flow
2. Implicit grant flow
3. Error responses


### **Code grant flow**
 **-Response**: https://YOUR_APP/redirect_uri?code=AUTHORIZATION_CODE&state=STATE
 **-Pattern** to match: `code` & `state`

### **Implicit grant flow:**
**-Response:** https://YOUR_APP/redirect_uri#access_token=ACCESS_TOKEN&token_type=bearer&expires_in=3600&state=STATE
 **-Pattern to match:** `access_token` & `token_type` & `expires_in` & `state`

### **Error responses:**
Cognito can send error responses in different circumstances, for more information check the [Examples for negative responses](https://docs.aws.amazon.com/cognito/latest/developerguide/authorization-endpoint.html) section.

**-Pattern to match:** 
error=invalid_request
error=unauthorized_client
error=invalid_scope
error=server_error


It would be cool to include test suites that handle these scenarios as well.

fossamagna:
@israx Thank you again for your review! I test manualy and I add unit tests for the following scenarios.
- Code grant flow
- Implicit grant flow
- Error responses

israx:
> @israx Thank you again for your review! I test manualy and I add unit tests for the following scenarios.
> 
> * Code grant flow
> * Implicit grant flow
> * Error responses

Thank you so much @fossamagna . I'll review these changes and discuss them with the team to find the next steps. Any update will be posted here. Thanks again

israx:
Hello @fossamagna. Thank you for the patience.

After reviewing these changes with the team, we realized that we support OAuth flows that are not triggered from Amplify, which means that `state` will not be included in the query string. Thus, implementing this solution will be a breaking change.

However we are still discussing possible approaches to solve the problem attached to this PR, and we will consider the solution implemented here.

fossamagna:
@israx Thank you for the information. Please let me know if any changes are needed in this RP. I will respond to it.

ryuki-kaneko:
@israx @haverchuck 
Hi!
I also have the bug. So I want you to merge this PR as soon as possible.
Are you still discussing the issue?


@fossamagna @israx @haverchuck 
Thank you for your handling.

fossamagna:
@israx Could you give us information on the current situation? Thanks.

abdallahshaban557:
Hi @fossamagna - since this PR has a breaking change, it will have to be merged to our next major version bump. We are still trying to finalize timelines, but this PR is on our list of issues to address when releasing V6.

fossamagna:
@abdallahshaban557 Thank you for your reply. I understand this RP status and release.

israx:
Hello @fossamagna . Thank you so much for this contribution. This PR has helped us to address this issue in the current v6 version. Currently the library will only capture a `code` and `state` query params only if there was an oauth flow initiated by the library. 

If you were able to confirm the fix of this issue using the current v6 version of the library, we would like to hear your feedback!

Issue #10732, opened 2022-12-01
Next.js app with DataStore works locally but not when deployed to Amplify Hosting


tombray:
### Before opening, please confirm:

- [X] I have checked to see if my question is addressed in the [FAQ](https://github.com/aws-amplify/amplify-hosting/blob/master/FAQ.md).
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-hosting/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-hosting/blob/master/CONTRIBUTING.md).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [X] I have removed any sensitive information from my code snippets and submission.

### App Id

d8dd1y47jar7i

### AWS Region

us-west-2

### Amplify Hosting feature

Not Applicable

### Describe the bug

I've created a simple Next.js app that uses a DataStore.query to fetch some data I generated in Amplify Studio. 

It works locally but when deployed to Amplify Hosting it throws an exception before posting the query.

Here's the exception:

> TypeError: Cannot read properties of undefined (reading 'length')
>     at n.copyOf (zen-observable.js:444:2)
>     at uq (zen-observable.js:444:2)
>     at e.<anonymous> (zen-observable.js:444:2)
>     at zen-observable.js:444:2
>     at Object.next (zen-observable.js:444:2)
>     at s (zen-observable.js:444:2)

### Expected behavior

Here's how the page renders locally:

<img width="288" alt="image" src="https://user-images.githubusercontent.com/311060/205121264-98a71961-1e86-41e0-8a05-202301acf739.png">

### Reproduction steps

Here's my repo: https://github.com/tombray/amplify-test

The DataStore.query is in pages/index.js:

```
export default function Home() {
  const [stations, setStations] = useState([]);

  useEffect(() => {
    const getStations = async () => {
      console.log("before datastore query");
      try {
        const stations = await DataStore.query(Stations);
        console.log(stations);
        setStations(stations);
      } catch (e) {
        console.log(e);
      }
    };

    getStations();
  }, []);

```
Running that locally works for me but when I deploy it with Amplify Hosting, it throws an exception which you can see here: https://main.d8dd1y47jar7i.amplifyapp.com/

### Log output

<details>

```
# Put your logs below this line


```

</details>

svidgen:
Hey @tombray,

[...]

PR #10765, opened 2022-12-12
fix(@aws-amplify/datastore): nextjs/swc broken build workaround

svidgen:
#### Description of changes

NextJS 13 enables SWC by default, which sometimes mangles the the build. This PR does two things:

1. Adds integ tests for sample apps that attempt to trigger bad nextjs builds
2. Adds a workaround for one known case, where SWC was over-aggressively dropping a variable declaration, resulting in an exception several lines later.

Note: Depends on https://github.com/aws-amplify/amplify-js-samples-staging/pull/494

#### Issue #, if available

#10732

Runtime failures were seen as in #10732 with NextJS@13 *prod* builds. The breakage was isolated by inserting dummy console logs with unique strings to help us determine where we were in the resulting minified build. By doing so, we could see that a variable declaration and later assignment were being eliminated entirely. Setting the var to a default value caused SWC to leave the declaration in and resolved runtime issues.

One of the newly added integ tests (the "basic" one) successfully triggers the error in CI on prod builds without the fix. The fix makes it pass in CI.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2050,7 +2050,7 @@ workflows:
             - integ_react_datastore_selective_sync
             - integ_react_datastore_nested_predicate
             - integ_next_datastore_13_basic
-            - integ_next_datastore_13
+            # - integ_next_datastore_13
             - integ_next_datastore_13_js
       - post_release:
           filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -391,6 +391,7 @@ jobs:
             echo $SSH_HOST_PUBLIC_KEY >> ~/.ssh/known_hosts
             git clone $AMPLIFY_JS_SAMPLES_STAGING_URL
             cd amplify-js-samples-staging
+            git checkout ds-nextjs-13
             branchExists=true
             git ls-remote --heads origin <<pipeline.git.branch>> | grep <<pipeline.git.branch>> >/dev/null || branchExists=false
             if [ "$branchExists" = true ]; then git checkout <<pipeline.git.branch>>; fi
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1354,6 +1354,57 @@ jobs:
...
+    parameters:
+      browser:
+        type: string
+    executor: js-test-executor
+    <<: *test_env_vars
+    working_directory: ~/amplify-js-samples-staging/samples/next/datastore/next-13-basic
+    steps:
+      - prepare_test_env
+      - integ_test_js:
+          test_name: 'DataStore - Nextjs basic'
```
jimblanc, line 1367: Nit: Should we add "SWC" to these test names just to make it a little more clear what the purpose is? E.g. `DataStore - Nextjs SWC basic`

.circleci/config.yml
```diff
@@ -1910,7 +1961,33 @@ workflows:
...
+      - integ_next_datastore_13_basic:
+          requires:
+            - integ_setup
+            - build
+          filters:
+            <<: *releasable_branches
+          matrix:
+            parameters:
+              <<: *test_browsers
+      # - integ_next_datastore_13:
```
jimblanc, line 1973: Is this test a WIP? What's the difference between `integ_next_datastore_13_basic`?

jimblanc (dismissed):

.circleci/config.yml
svidgen: The TS version of the test is failing for different nodejs compatibility reasons that I just didn't have time to dig into. Not sure if it's better to leave the uncommented here to re-add later or remove entirely.

jimblanc (approved):

.circleci/config.yml
svidgen: Renamed.

iartemiev (approved):
LGTM 👍🏻

dpilch (approved):

PR #10774, opened 2022-12-13
Sign up function

helgabalashova:
#### Description of changes
Implements sign up function (without auto sign in, hub, and errors implementations; TODOs are in place).

#### Checklist

- [X] PR description included
- [ ] `yarn test` passes


package.json
```diff
@@ -126,6 +126,9 @@
 			}
 		]
 	},
+	"dependencies": {
+		"@aws-sdk/client-cognito-identity-provider": "^3.198.0"
```
jimblanc, line 130: Would prefer to peg this to a specific version of the SDK while we try to get our bundle size ironed out. Also does this align with our other SDK dependencies? I see 3.6.1 in a lot of packages.

packages/auth/src/Auth.ts
```diff
@@ -90,9 +89,11 @@ export class AuthClass {
 		this._pluggable = null;
 	}
 
-	signUp(): Promise<any> {
+	signUp<PluginOptions extends AuthPluginOptions>(req: SignUpRequest<string, PluginOptions>): Promise<AuthSignUpResult<string>>;
+	signUp<PluginOptions extends AuthPluginOptions>(req:SignUpRequest<CognitoUserAttributeKey, PluginOptions>):Promise<AuthSignUpResult<CognitoUserAttributeKey>>
```
jimblanc, line 93: Nit: Missing some white-space in the params here, e.g. `req:SignUpRequest` & `:Promise`)

packages/auth/src/Providers/AmazonCognitoProvider.ts
```diff
@@ -16,7 +16,21 @@ import {
...
+	AuthSignUpStep, 
+	DeliveryMedium, 
+	CognitoSignUpOptions, 
+	SignUpRequest, 
+	AuthSignUpResult, 
+	CognitoUserAttributeKey, 
+	ValidationData, 
+	AuthProvider 
+} from '../types';
+import { AttributeType, CognitoIdentityProviderClient, SignUpCommand } from '@aws-sdk/client-cognito-identity-provider';
```
jimblanc, line 30: Nit: Generally I think we should try to put non-relative imports before the relative ones, although I realize we should have a linter rule for this instead.

packages/auth/src/Providers/AmazonCognitoProvider.ts
```diff
@@ -31,12 +45,80 @@ export class AmazonCognitoProvider implements AuthProvider {
...
+			userAttr = req.options.userAttributes.map(obj => ({
+				Name: obj.userAttributeKey,
+				Value: obj.value
+			}));
+		}
+
+		let validationData: AttributeType[] | undefined;
+		let clientMetadata: Record<string, string> | undefined;
+		if (req.options?.pluginOptions) {
+			const pluginOptions = req.options.pluginOptions;
```
jimblanc, line 80: Nit: Could we re-order this to reduce duplication, e.g.

```
const pluginOptions = req.options?.pluginOptions;
if (pluginOptions) {
```

packages/auth/src/Providers/AmazonCognitoProvider.ts
```diff
@@ -31,12 +45,80 @@ export class AmazonCognitoProvider implements AuthProvider {
...
+		}
+
+		const password: string = req.password;
+		if (!password) {
+			throw new AuthError(AuthErrorTypes.EmptyPassword); // TODO change when errors are defined
+		}
+
+		let userAttr: AttributeType[] | undefined;
+		if (req.options?.userAttributes) {
+			userAttr = req.options.userAttributes.map(obj => ({
```
jimblanc, line 71: Ditto here about reducing duplication

packages/auth/src/Providers/AmazonCognitoProvider.ts
```diff
@@ -31,12 +45,80 @@ export class AmazonCognitoProvider implements AuthProvider {
...
+					Value: value
+				}));
+			}
+			clientMetadata = pluginOptions['ClientMetadata'];
+		}
+
+		const signUpCommand: SignUpCommand = createSignUpCommand(clientId, username, password, userAttr, validationData, clientMetadata);
+
+		try {
+			const signUpCommandOutput: any = await sendCommand(this._client, signUpCommand);
```
jimblanc, line 94: Any way to properly type this output?

packages/auth/src/Providers/AmazonCognitoProvider.ts
```diff
@@ -31,12 +45,80 @@ export class AmazonCognitoProvider implements AuthProvider {
...
+				};
+			} else {
+				result = {
+					isSignUpComplete: false,
+					nextStep: {
+						signUpStep: AuthSignUpStep.CONFIRM_SIGN_UP,
+						codeDeliveryDetails: {
+							deliveryMedium: signUpCommandOutput.CodeDeliveryDetails?.DeliveryMedium as DeliveryMedium,
+							destination: signUpCommandOutput.CodeDeliveryDetails?.Destination as string,
+							attributeName: signUpCommandOutput.CodeDeliveryDetails?.AttributeName as CognitoUserAttributeKey
```
jimblanc, line 111: Should we avoid optional chaining here? If the next step is `CONFIRM_SIGN_UP`, would it ever be valid for `CodeDeliveryDetails` to be undefined? Also nit-pick but we could extract `signUpCommandOutput.CodeDeliveryDetails` into a const to reduce duplication.

packages/auth/src/types/index.ts
```diff
@@ -1,4 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-export { AuthProvider } from './models';
+export * from './models';
+export * from './options';
+export * from './request';
+export * from './result';
+export * from './aws-plugins/cognito-plugin/types/models';
+export * from './aws-plugins/cognito-plugin/types/options';
```
jimblanc, line 9: Nit-pick but I think I'd prefer to export all types for a given plugin from that plugin's index file, and then re-export from that here (just to keep the organization cleaner)

packages/auth/src/utils/CognitoIdentityProviderClientUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+	return config.userPoolId;
+}
+
+export const createSignUpCommand = (
+	clientId: string, 
+	username: string, 
+	password: string,
+	userAttributes?: AttributeType[],
+	validationData?: AttributeType[],
+	clientMetadata?
```
jimblanc, line 50: Should this be typed?

packages/auth/src/utils/CognitoIdentityProviderClientUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+	if (clientMetadata) {
+		signUpCommandInput.ClientMetadata = clientMetadata;
+	}
+	return new SignUpCommand(signUpCommandInput);
+}
+
+export const sendCommand = async (client:CognitoIdentityProviderClient, command) => {
+	try {
+		const commandOutput = await client.send(command);
+		return commandOutput;
```
jimblanc, line 72: Should be able to just return the promise directly: `return client.send(command);`

jimblanc (commented):
Some comments below. Also in the interest of keeping PRs smaller okay with leaving unit tests out for now, but we should make sure to get those included.

packages/auth/src/types/aws-plugins/cognito-plugin/types/models/cognitoUserAttributeKey.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { AuthStandardAttributeKey } from '../../../../models';
+
+export type CognitoUserAttributeKey = 
+	| AuthStandardAttributeKey
+	| `custom:${string}`
```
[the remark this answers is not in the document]
israx: please also double check the type of `AuthUserAttributeKey` as well

packages/auth/src/Providers/AmazonCognitoProvider.ts
helgabalashova: I think the only way to type this output would be to create a separate function for each send command and the only difference between them would be the command input type and return type.

packages/auth/src/Providers/AmazonCognitoProvider.ts
helgabalashova: by optional chaining you mean to make codeDeliveryDetails required part of the result object? If so, it is not an option because we don't return codeDeliveryDetails if next step is DONE

packages/auth/src/Providers/AmazonCognitoProvider.ts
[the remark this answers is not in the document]
israx: I checked the util `sendCommand` function. I think it's not needed, it's redundant to have a double `try and catch` clause

packages/auth/src/utils/CognitoIdentityProviderClientUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+	if (validationData) {
+		signUpCommandInput.ValidationData = validationData;
+	}
+	if (clientMetadata) {
+		signUpCommandInput.ClientMetadata = clientMetadata;
+	}
+	return new SignUpCommand(signUpCommandInput);
+}
+
+export const sendCommand = async (client:CognitoIdentityProviderClient, command) => {
```
israx, line 69: this function shouldn't be needed, there's no reusability benefits

packages/auth/src/utils/assertPluginAvailable.ts
```diff
@@ -12,7 +12,7 @@
  */
 
 import { AuthErrorStrings } from '../constants/AuthErrorStrings';
-import { AuthProvider } from '../types/AuthProvider';
+import { AuthProvider } from '../types';
```
israx, line 15: `AuthProvider` name should be change as it would conflict with other type name

israx (changes requested):

packages/auth/src/Providers/AmazonCognitoProvider.ts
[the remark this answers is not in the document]
helgabalashova: it has a SignUpCommandOutput but we need to decide if we want to have a separate function for each send command or just one general? If we choose to proceed with general then we won't be able to define type here

packages/auth/src/utils/CognitoIdentityProviderClientUtils.ts
helgabalashova: this is what @huverchuck asked me to do. We are going to have all sdk calls as functions in this file. And in case of reusability, if we keep it this way, it can be used for all other send calls (i.e. send confirmSignUpCommand)

packages/auth/src/Providers/AmazonCognitoProvider.ts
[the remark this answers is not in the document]
helgabalashova: and having sendCommand was Dustin's decision to move all aws-sdk commands to utils files (that's why it is there) and in case of keeping it general, it will be reused for other function. We will use it from each Auth function

packages/auth/src/Auth.ts
```diff
@@ -90,9 +89,11 @@ export class AuthClass {
 		this._pluggable = null;
 	}
 
-	signUp(): Promise<any> {
+	signUp<PluginOptions extends AuthPluginOptions>(req: SignUpRequest<string, PluginOptions>): Promise<AuthSignUpResult<string>>;
```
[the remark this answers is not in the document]
helgabalashova: discussed offline. SignUpRequest is in tact with current version of approved API

packages/auth/src/Providers/AmazonCognitoProvider.ts
helgabalashova: I think it is usually in alphabetical order (at least when you let VS Code add imports for you).

packages/auth/src/types/models/AuthProvider.ts
```diff
@@ -11,7 +11,7 @@ export interface AuthProvider {
 	// return the name of you provider
```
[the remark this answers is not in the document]
helgabalashova: let's see what @haverchuck has to say and then I will change it if needed.

packages/auth/src/Providers/AmazonCognitoProvider.ts
```diff
@@ -31,12 +45,77 @@ export class AmazonCognitoProvider implements AuthProvider {
...
+				Value: obj.value
+			}));
+		}
+
+		let validationData: AttributeType[] | undefined;
+		let clientMetadata: Record<string, string> | undefined;
+		const pluginOptions = req.options?.pluginOptions;
+		if (pluginOptions) {
+			const validationDataObject: ValidationData = pluginOptions['ValidationData'];
+			if (validationDataObject) {
```
elorzafe, line 83: What if we create a separate function for this kinds of things?

elorzafe (changes requested):
Overall looking good, I would try to create smaller functions so the code can be easier to test.

Are you creating Unit tests in a follow up PR?

Can you fix the `export *`?

Thanks!

packages/auth/src/Providers/AmazonCognitoProvider.ts
```diff
@@ -31,12 +45,77 @@ export class AmazonCognitoProvider implements AuthProvider {
...
+				Name: obj.userAttributeKey as string,
+				Value: obj.value
+			}));
+		}
+
+		let validationData: AttributeType[] | undefined;
+		let clientMetadata: Record<string, string> | undefined;
+		const pluginOptions = req.options?.pluginOptions;
+		if (pluginOptions) {
+			const validationDataObject: ValidationData = pluginOptions['ValidationData'];
```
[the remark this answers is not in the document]
israx: I think the reason is that `pluginOptions` variable doesn't have a type yet. So something to add later is an **assertion guard** for plugin options, so then `pluginOptions` will show a type

packages/auth/src/types/index.ts
israx: @elorzafe I think that wouldn't be a problem as this are just types. At build time types will be lost anyways

packages/auth/src/types/models/AuthProvider.ts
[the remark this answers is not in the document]
israx: yeah the thing is that the convention with the other libraries is `AuthProvider`

packages/auth/src/utils/CognitoIdentityProviderClientUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+	switch(challengeNameType) {
+		// case ChallengeNameType.ADMIN_NO_SRP_AUTH:
+		case ChallengeNameType.CUSTOM_CHALLENGE:
+			return AuthSignInStep.CONFIRM_SIGN_IN_WITH_CUSTOM_CHALLENGE;
+		// case ChallengeNameType.DEVICE_PASSWORD_VERIFIER:
+		// case ChallengeNameType.DEVICE_SRP_AUTH:
+		// case ChallengeNameType.MFA_SETUP:
+		case ChallengeNameType.NEW_PASSWORD_REQUIRED:
+			return AuthSignInStep.CONFIRM_SIGN_IN_WITH_NEW_PASSWORD_REQUIRED;
+		// case ChallengeNameType.PASSWORD_VERIFIER:
```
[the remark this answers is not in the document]
israx: I'm wondering if the AWS SDK has a type for `ChallengeName` that can be imported, could you check that @helgabalashova  ?

packages/auth/src/utils/CognitoIdentityProviderClientUtils.ts
israx: If we are keeping this function then I think we can make it generic, for various output commands

packages/auth/src/types/models/AuthProvider.ts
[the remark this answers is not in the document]
helgabalashova: it just makes more sense to keep it as `AuthProvider` and change what you suggest to `oAuthProvider` or something like that.

packages/auth/src/Providers/AmazonCognitoProvider.ts
```diff
@@ -51,48 +63,68 @@ export class AmazonCognitoProvider implements AuthProvider {
...
-				validationData = Object.entries(validationDataObject).map(([key, value]) => ({
-					Name: key,
-					Value: value
-				}));
-			}
-			clientMetadata = pluginOptions['ClientMetadata'];
+			// TODO: change to pluginOptions.ValidationData if type of PluginOptions is mapped
+			validationData = this.mapValidationDataObject(pluginOptions['validationData']);
+			// TODO: change to pluginOptions.ClientMetadata if type of PluginOptions is mapped 
+			clientMetadata = pluginOptions['clientMetadata'];
```
elorzafe, line 91: Is it PascalCase or camelCase? (`validationData` or `ValidationData`)

packages/auth/src/Providers/AmazonCognitoProvider.ts
```diff
@@ -51,48 +63,68 @@ export class AmazonCognitoProvider implements AuthProvider {
...
+	):AttributeType[] | undefined {
+		if (attrs) {
+			return attrs.map(obj => ({
+				Name: obj.userAttributeKey as string,
+				Value: obj.value
+			}));
+		}	
+	}
+
+	private mapValidationDataObject(
```
elorzafe, line 113: nit: is it `map` the right description, it looks more convert object to array

elorzafe (commented):
A few nit pick comments and one concern with camelCase vs PascalCase value

packages/auth/src/Providers/AmazonCognitoProvider.ts
helgabalashova: it is `validationData` according to [API definition](https://quip-amazon.com/rfNLAS8ns64b/Amplify-JS-Auth-API-Definition-2022#temp:C:IeK0ddcb020529e4579a56dffdb1)

elorzafe (approved):

elorzafe (approved):

packages/auth/src/Auth.ts
```diff
@@ -90,9 +96,15 @@ export class AuthClass {
 		this._pluggable = null;
 	}
 
-	signUp(): Promise<any> {
-		assertPluginAvailable(this._pluggable);
-		return this._pluggable.signUp();
+	signUp<PluginOptions extends AuthPluginOptions>(
```
haverchuck, line 99: We need to decide how we are going to document category-level API methods and plugin implementation methods. Could you work with @israx on that? For now, please include documentation for both following TS best practices.

packages/auth/src/Providers/AmazonCognitoProvider.ts
```diff
@@ -31,13 +57,99 @@ export class AmazonCognitoProvider implements AuthProvider {
...
+	private _client: CognitoIdentityProviderClient;
 
 	constructor(config?) {
 		this._config = config ? config : {};
+		this._client = createCognitoIdentityProviderClient(config);
 	}
-	signUp(): Promise<any> {
-		throw new Error('Method not implemented.');
+
+	async signUp<PluginOptions extends AuthPluginOptions = CognitoSignUpOptions>(
```
haverchuck, line 67: See above comment about documentation.

packages/auth/src/Providers/AmazonCognitoProvider.ts
```diff
@@ -31,13 +57,99 @@ export class AmazonCognitoProvider implements AuthProvider {
...
+		req: SignUpRequest<CognitoUserAttributeKey, PluginOptions>
+	): Promise<AuthSignUpResult<CognitoUserAttributeKey>> {
+		const clientId: string = getUserPoolId(this._config);
```
haverchuck, line 70: Couldn't the CognitoIdentityProviderClient add this  internally?

packages/auth/src/Providers/AmazonCognitoProvider.ts
```diff
@@ -31,13 +57,99 @@ export class AmazonCognitoProvider implements AuthProvider {
...
+		let clientMetadata: Record<string, string> | undefined;
+		const pluginOptions = req.options?.pluginOptions;
+		if (pluginOptions) {
+			// TODO: change to pluginOptions.ValidationData if type of PluginOptions is mapped
+			validationData = this.convertValidationDataObjectToArray(pluginOptions['validationData']);
+			// TODO: change to pluginOptions.ClientMetadata if type of PluginOptions is mapped 
+			clientMetadata = pluginOptions['clientMetadata']; 
+		}
+
+		const signUpCommand: SignUpCommand = createSignUpCommand(
```
haverchuck, line 94: I'm not entirely sure exporting these command constructors is the cleanest way to go, but I think it's ok for now.

packages/auth/src/types/models/AuthProvider.ts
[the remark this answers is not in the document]
haverchuck: We are trying to match the APIs in other libraries as closely as possible, so we should probably change this file name to `AuthPluginProvider` or something similar, since we will want to reserve this name for classes/types/enums that are exposed in the APIs.

haverchuck (changes requested):
Left a couple of minor suggestions, but otherwise looks great.

haverchuck (approved):

Issue #10764, opened 2022-12-12
DataStore.clear()  Error : [TypeError: undefined is not a function]


santhoshpriya0715:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

DataStore

### Amplify Categories

Not applicable

### Environment information

<details>

```
# Put output below this line

  System:
    OS: Windows 10 10.0.22000
    CPU: (12) x64 11th Gen Intel(R) Core(TM) i5-11400F @ 2.60GHz
    Memory: 3.14 GB / 15.84 GB
  Binaries:
    Node: 16.13.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.18 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.18.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (108.0.1462.46)
    Internet Explorer: 11.0.22000.120
  npmPackages:
    @alessiocancian/react-native-actionsheet: ^3.2.0 => 3.2.0 
    @aws-amplify/datastore-storage-adapter: ^2.0.5 => 2.0.5 
    @aws-amplify/pushnotification: ^5.0.5 => 5.0.5 
    @aws-sdk/client-codeguruprofiler: ^3.226.0 => 3.226.0 
    @aws-sdk/client-xray: ^3.226.0 => 3.226.0 
    @azure/core-asynciterator-polyfill: ^1.0.2 => 1.0.2 
    @babel/core: ^7.20.5 => 7.20.5 
    @babel/runtime: ^7.20.6 => 7.20.6 
    @react-native-async-storage/async-storage: ^1.17.11 => 1.17.11 
    @react-native-community/eslint-config: ^3.2.0 => 3.2.0 
    @react-native-community/netinfo: ^9.3.7 => 9.3.7 
    @react-native-community/push-notification-ios: ^1.10.1 => 1.10.1 (1.0.3)
[...]

PR #10799, opened 2022-12-20
fix: create custom Promise.allSettled to work on react native

dpilch:
#### Description of changes

Replace `Promise.allSettled` with a non-fully-spec implementation. In current versions of React Native `Promise.allSettled` is not available resulting in an error. (See https://github.com/aws-amplify/amplify-js/issues/10764#issuecomment-1359993927 for full details).

Another option would be to add a shim from an npm package. However, this would add additional deps.

IMO, the custom implementation serves our needs without introducing additional deps, but I could be convinced to use a shim from npm.

TODO: add a lint rule to prevent the use of `Promise.allSettled`.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/10764

* Unit test
* Test in react native app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/core/src/Util/BackgroundProcessManager.ts
```diff
@@ -361,7 +361,7 @@ export class BackgroundProcessManager {
 
 			// Use `allSettled()` because we want to wait for all to finish. We do
 			// not want to stop waiting if there is a failure.
-			this._closingPromise = Promise.allSettled(
+			this._closingPromise = allSettled(
```
david-mcafee, line 364: Thoughts on adding `BackgroundProcessManager.native.ts` alongside this file (as we do for `Reachability`) and only using this implementation for React Native?

packages/core/src/Util/BackgroundProcessManager.ts
```diff
@@ -447,3 +447,30 @@ type JobEntry = {
 	 */
 	description?: string;
 };
+
+/**
+ * Custom non-spec implementation for Promise.allSettled.
```
david-mcafee, line 452: I did a little bit of digging, but I'm sure you've done more than I have - are there spec-compliant implementations we could pull in here?

david-mcafee (changes requested):

packages/core/src/Util/BackgroundProcessManager.ts
dpilch: I know of at least one https://www.npmjs.com/package/promise.allsettled. It just depends on if we want to add an additional dependency or not.

packages/core/src/Util/BackgroundProcessManager.ts
dpilch: I wasn't aware of this option. Is there a way we can easily do this without repeating the other code in this file?

PR #9763, opened 2022-04-04
Fix(storage): Remove unhandled promise rejection for _verifyFileSize()

Pickleboyonline:
#### Description of changes

The function `_verifyFileSize()` is not awaited in the `_completeUpload()` function located in `AWSS3UploadTask.ts`. This can cause unhandled promise rejections and the `_completeUpload()` to not behave as intended. This PR aims to solve this by awaiting the `_verifyFileSize()` while only rejecting if `ListObjectsV2Command`  fulfills and the remote and local file sizes are different. The change is due to Amplify allowing users to use pre-existing S3 instances and the permission `S3:ListBucket` is not listed as a prerequisite for uploading a file. Enforcing the `ListObjectsV2Command` can break some developers' app functionality if they did not have this action permitted.  

#### Description of how you validated changes
Made a test that expects the error to be called with the message 'File size does not match between local file and file on s3' when remote and local file sizes do not match.

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)
- [X] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #10859, opened 2023-01-11
chore(storage): Apply fixes from merged PR

stocaaro:
#### Description of changes
There where a handful of minor comments on [another PR](https://github.com/aws-amplify/amplify-js/pull/9763) that has been merged. This change covers the minor refactors that where requested.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/storage/src/providers/AWSS3UploadTask.ts
```diff
@@ -375,12 +375,8 @@ export class AWSS3UploadTask implements UploadTask {
...
 		} catch (e) {
 			logger.log('Could not get file on s3 for size matching: ', e);
-			// Allow to proceed to not break previous implementations.
-			// Could have authentication error caught since the `s3:getObject`
-			// action is not a listed as a prerequisite for Storage.put().
-			// Users can have this missing due to how one can use an
-			// existing S3 bucket.
-			// ? We can throw warning if DEV environments
+			// Don't gate verification on auth or other errors
+			// unrelated to file size verification
```
stocaaro, line 379: Was going to pull the code below up into the try block to allow `obj` to be const, but in this case, the `throw` in the code below would be caught which changes behavior by not throwing out of the function when file sizes are different because the error is caught.

Tried a couple refactors, but they weren't apparently better, so I reverted this and made a run at clarifying the comment.

cshfang: I see what you're saying and yeah I can't see anything significantly improving this block. If I were to nitpick, I do think refactoring it slightly so that the `let` is on the `valid` instead of the obj would be a little more preferable as it is more idiomatic (and microscopically safer?) to have a variable boolean as opposed to an object.

I.e. 

```
let valid
try {
  const obj = ...
  valid = Boolean(obj && obj.Size === this.file.size);
} ...
if (!valid) {
  throw
...
  
```

cshfang (approved):
Lgtm. Just a minor probably _subjectively_ bar-raising feedback?

AllanZhengYP (approved):

packages/storage/src/providers/AWSS3UploadTask.ts
stocaaro: I can get behind that. Updated.

cshfang (approved):

AllanZhengYP (approved):

Issue #10866, opened 2023-01-13
Error when deleting (Error: Namespace Resolver for 'Object' not found! This is probably a bug in '@amplify-js/datastore'.)


madmaxaus:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React, Next.js

### Amplify APIs

DataStore

### Amplify Categories

Not applicable

### Environment information

<details>

```
# Put output below this line

  System:
    OS: Windows 10 10.0.22000
    CPU: (16) x64 AMD Ryzen 7 Microsoft Surface (R) Edition
    Memory: 3.46 GB / 15.45 GB
  Binaries:
    Node: 16.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.11.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (108.0.1462.76)
    Internet Explorer: 11.0.22000.120
  npmGlobalPackages:
    @aws-amplify/cli: 10.0.0
    amplify: 0.0.11
    azure-functions-core-tools: 4.0.4915
    yarn: 1.22.19
```

</details>

### Describe the bug

When i try to delete a record using datastore i get the error: (other non-delete queries work)

        
       Error: Namespace Resolver for 'Object' not found! This is probably a bug in '@amplify-js/datastore'.
    at IndexedDBAdapter.namespaceResolver (datastore.js?38c2:94:1)
    at IndexedDBAdapter.getIndexKeyValuesFromModel (IndexedDBAdapter.js?e5aa:38:1)
    at IndexedDBAdapter.eval (IndexedDBAdapter.js?e5aa:1031:1)
    at step (tslib.es6.js?6654:100:1)
    at Object.eval [as next] (tslib.es6.js?6654:81:46)
    at fulfilled (tslib.es6.js?6654:71:43)

### Expected behavior

Delete the object like it used to

### Reproduction steps

[...]

PR #10880, opened 2023-01-19
fix(datastore): cascade deleted for nested Has Many

iartemiev:
#### Description of changes

Fixes bug when deleting a record of a model with > 1 nested Has Many descendants
Ex:
```graphql
type Blog @model {
  id: String!
  posts: [Post] @hasMany
}

type Post @model {
  id: String!
  comments: [Comment] @hasMany
}

type Comment @model {
  id: String!
}
```
Behavior before the change in this PR:
1. `await DataStore.delete(Blog, 'some-id')`
2. `deleteTraverse` retrieves all `Post` records from the client database that are related to the Blog record and recursively calls `deleteTraverse`, passing the retrieved post records as POJOs ([link to code](https://github.com/aws-amplify/amplify-js/pull/10880/files#diff-75c76069840355a3f346f2f35c443018e4a5716ed128191602499c3b39935cfcR1076-R1080))
3. since `Post` has a Has Many relationship to `Comment`, `deleteTraverse` attempts to retrieve the related `Comment` records, but first has to extract PK metadata from the posts. Since the posts are POJOs, instead of model instances, they don't contain the necessary data and cause the error in the linked issue.

Note: If the child model does not have descendant relationships defined in the schema (e.g. if Post did not have a Has Many relationship to Comment) we wouldn't attempt to retrieve its descendants in step 3. so the bug would not show up with a relational graph of depth 2. Only 3+.

The change in this PR instantiates models from the records retrieved in step 2 and passes the models in the recursive call, which fixes the bug.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/10866

* Red/green unit test
* Manual testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/commonAdapterTests.ts
+++ b/packages/datastore/__tests__/commonAdapterTests.ts
@@ -19,10 +19,10 @@ import {
 	Model,
 	User,
 	Profile,
+	Blog,
 	Post,
 	Comment,
 	testSchema,
-	CompositePKParent,
 	HasOneParent,
 	HasOneChild,
 	MtmLeft,
@@ -377,6 +377,7 @@ export function addCommonQueryTests({
 
 	describe('common `delete()` cases', () => {
 		let Comment: PersistentModelConstructor<Comment>;
+		let Blog: PersistentModelConstructor<Blog>;
 		let Post: PersistentModelConstructor<Post>;
 		let HasOneParent: PersistentModelConstructor<HasOneParent>;
 		let HasOneChild: PersistentModelConstructor<HasOneChild>;
@@ -387,6 +388,7 @@ export function addCommonQueryTests({
 			const classes = initSchema(testSchema());
 			({
 				Comment,
+				Blog,
 				Post,
 				HasOneParent,
 				HasOneChild,
@@ -394,6 +396,7 @@ export function addCommonQueryTests({
 				DefaultPKHasOneChild,
 			} = classes as {
 				Comment: PersistentModelConstructor<Comment>;
+				Blog: PersistentModelConstructor<Blog>;
 				Post: PersistentModelConstructor<Post>;
 				HasOneParent: PersistentModelConstructor<HasOneParent>;
 				HasOneChild: PersistentModelConstructor<HasOneChild>;
@@ -500,6 +503,35 @@ export function addCommonQueryTests({
 					comment.id
 				);
 				expect(retrievedDeletedComment).toBeUndefined();
+			}
+		);
+
+		(isSQLiteAdapter() ? test.skip : test)(
+			'deleting nested hasMany cascades',
+			async () => {
+				const blog = await DataStore.save(
+					new Blog({
+						title: 'my blog',
+					})
+				);
+
+				const post = await DataStore.save(
+					new Post({
+						title: 'my post',
+						blogId: blog.id,
+					})
+				);
+
+				const retrievedBlog = await DataStore.query(Blog, blog.id);
+				expect(retrievedBlog!.id).toEqual(blog.id);
+
+				const posts = await retrievedBlog!.posts.toArray();
+				expect(posts.length).toEqual(1);
+				expect(posts[0].id).toEqual(post.id);
+
+				await DataStore.delete(Blog, blog.id);
+
+				expect(await DataStore.query(Post, post.id)).toBeUndefined();
 			}
 		);
 
--- a/packages/datastore/__tests__/commonAdapterTests.ts
+++ b/packages/datastore/__tests__/commonAdapterTests.ts
@@ -506,6 +506,13 @@ export function addCommonQueryTests({
 			}
 		);
 
+		/**
+		 * Distinct from other hasMany delete cascade tests in that the model
+		 * being deleted has *potential* grandchildren. Hence, delete traversals need
+		 * to be capable of successfully querying the child for its children,
+		 * which has seen a regression at least once!
+		 * See: https://github.com/aws-amplify/amplify-js/issues/10866
+		 */
 		(isSQLiteAdapter() ? test.skip : test)(
 			'deleting nested hasMany cascades',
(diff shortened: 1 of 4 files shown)


packages/datastore/__tests__/commonAdapterTests.ts
```diff
@@ -503,6 +506,35 @@ export function addCommonQueryTests({
...
+
+				const posts = await retrievedBlog!.posts.toArray();
+				expect(posts.length).toEqual(1);
+				expect(posts[0].id).toEqual(post.id);
+
+				await DataStore.delete(Blog, blog.id);
+
+				expect(await DataStore.query(Post, post.id)).toBeUndefined();
+			}
+		);
```
svidgen, line 536: What's the relevant difference between what this tests and [the test just above it](https://github.com/aws-amplify/amplify-js/pull/10880/files#diff-de9d5dae30ad1e8fd1b9539e2f4cfdd40c71894df2fbb38c779a409f3d8a4908R472-R507)?

Why was the existing test passing and this one red-green-able? Is the pre-existing test for this broken?

svidgen (commented):
I think I see why this fixes the problem. But, I have one concern, noted in diff.

I'm also very curious whether this problem becomes simpler if handled outside the adapter. I'm thinking we should have an adapter-independent `DeleteTraversal` class or util function, and the adapters should just accept a list of items to delete (transactionally if supported, for the efficiency's sake if nothing else).

packages/datastore/__tests__/commonAdapterTests.ts
iartemiev: The reason is the same as the Note in the PR description. You have to have at least 3 models in a relational graph for this bug to occur.

packages/datastore/__tests__/helpers.ts
```diff
@@ -530,8 +536,7 @@ class FakeGraphQLService {
 			this.tables.set(model.name, new Map<string, any[]>());
 			let CPKFound = false;
 			for (const attribute of model.attributes || []) {
-				// Pretty sure the first key is the PK.
-				if (attribute.type === 'key') {
+				if (isModelAttributePrimaryKey(attribute)) {
```
svidgen, line 539: Oh, thanks for fixing this!

iartemiev:
> I'm also very curious whether this problem becomes simpler if handled outside the adapter. I'm thinking we should have an adapter-independent DeleteTraversal class or util function, and the adapters should just accept a list of items to delete (transactionally if supported, for the efficiency's sake if nothing else).

Abstracting out parts of the storage adapter implementations would avoid repetition, but the change in this PR would still apply.

packages/datastore/__tests__/helpers.ts
iartemiev: Np! The "key" difference with that util is that - it's not the first key attribute that's the PK. It's the key attribute without a `name` property

packages/datastore/__tests__/commonAdapterTests.ts
svidgen: I'm probably being dense. But, how does the three-level-nesting matter if the test is only going two levels deep?

iartemiev: And with this unit test I added a `Blog` model that has a Has Many relationship to `Post`, which has a Has Many to `Comment`. Elsewhere in the unit tests we only test models with at most 1 descendant.

svidgen:
> > I'm also very curious whether this problem becomes simpler if handled outside the adapter. I'm thinking we should have an adapter-independent DeleteTraversal class or util function, and the adapters should just accept a list of items to delete (transactionally if supported, for the efficiency's sake if nothing else).
> 
> Abstracting out parts of the storage adapter implementations would avoid repetition, but the change in this PR would still apply.

Woops. I posted a half-completed thought here! 🧠  Yes. I agree! And this is also outside of scope for the PR anyway. Seeing redundant changes and *yet another* "not relevant for SQLiteAdapter" test bumps the urgency level of some refactoring here. So, I wanted to plant the seed and looking for a +/- reaction to the idea. If it inspired a followup, I wouldn't be opposed. I also wouldn't oppose to adding an item to the backlog.

iartemiev:
That's fair. I think we should merge this to unblock folks that are depending on it first (provided there aren't any issues with the PR). But I'll create a backlog task to refactor. It'll be a non-trivial effort because despite a lot of high level similarity, there are many adapter-specific code differences that we'll need to find a way to express elegantly.

packages/datastore/__tests__/commonAdapterTests.ts
```diff
@@ -503,6 +506,35 @@ export function addCommonQueryTests({
 			}
 		);
 
+		(isSQLiteAdapter() ? test.skip : test)(
```
svidgen, line 509:
```suggestion
		/**
		 * Distinct from other hasMany delete cascade tests in that the model
		 * being deleted has *potential* grandchildren. Hence, delete traversals need
		 * to be capable of successfully querying the child for its children,
		 * which has seen a regression at least once!
		 * See: https://github.com/aws-amplify/amplify-js/issues/10866
		 */
		(isSQLiteAdapter() ? test.skip : test)(
```

svidgen (commented):
Thanks for your patience explaining this to me. Does the suggested comment below on the new test make sense?

packages/datastore/__tests__/helpers.ts
svidgen: 😅

packages/datastore/__tests__/commonAdapterTests.ts
svidgen: Thanks for explaining it and also making me feel less dumb about asking!

svidgen (approved):

david-mcafee (approved):
LGTM!

PR #10915, opened 2023-01-31
fix(@aws-amplify/datastore): strictly define `null` vs `undefined` behavior on models

svidgen:
#### Description of changes

Given:

```graphql
type Model @model {
  id: ID!
  value: String!
  otherValue: String
}
```

If we save an instance:

```js
    value: 'some value'
}));
```

We'll have this returned:

```json
{
  "value": "some value",
}
```

```json
{

  // LOOK!
  "otherValue": null, 

  "owner": "test",
  "_version": 1,
  "_deleted": null
}
```

2. Verified in sample apps.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/src/predicates/next.ts
```diff
@@ -558,17 +592,41 @@ export class GroupCondition {
...
+			// 		{
+			// 			a: this,
+			// 			// b: this.toAST(),
+			// 			c: this.withFieldConditionsOnly(negateChildren),
+			// 			d: this.withFieldConditionsOnly(negateChildren).toAST(),
+			// 		},
+			// 		null,
+			// 		2
+			// 	)
+			// );
```
svidgen, line 628: Pending cleanup.

packages/datastore/src/predicates/next.ts
```diff
@@ -558,17 +592,41 @@ export class GroupCondition {
...
 				p =>
 					p[operator](c =>
 						applyConditionsToV1Predicate(c, conditions, negateChildren)
 					)
 			);
 
-			resultGroups.push(
-				await storage.query(this.model.builder, predicate as any)
-			);
+			const predicateB =
```
svidgen, line 603: pending cleanup.

packages/datastore/src/predicates/next.ts
```diff
@@ -359,6 +376,24 @@ export class GroupCondition {
 		return [copied, extractedCopy];
 	}
 
+	withFieldConditionsOnly(negate: boolean) {
```
svidgen, line 379: TODO: docstring.

packages/datastore/src/predicates/next.ts
```diff
@@ -169,6 +171,21 @@ export class FieldCondition {
 		};
 	}
 
+	negated(model: ModelMeta<any>) {
```
svidgen, line 174: TODO: docstring.

packages/datastore/__tests__/indexeddb.test.ts
```diff
@@ -600,6 +608,8 @@ describe('Indexed db storage test', () => {
 			}
 		);
 
+		console.log({ sortedPersons });
```
svidgen, line 611: TODO: remove.

packages/datastore/__tests__/indexeddb.test.ts
```diff
@@ -586,6 +586,14 @@ describe('Indexed db storage test', () => {
 		await DataStore.save(p2);
 		await DataStore.save(p3);
 
+		// TEMP
+		await DataStore.save(
+			new Person({
+				firstName: 'Decoy first',
+				lastName: 'Decoy last',
+			})
+		);
```
svidgen, line 595: Remove.

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -495,7 +506,16 @@ class IndexedDBAdapter implements Adapter {
...
 			const predicateObj = predicateObjs.find(
-				p => isPredicateObj(p) && p.field === key && p.operator === 'eq'
+				p =>
+					// it's a relevant predicate object only if it's an equality
+					// operation for a key field from the key:
+					isPredicateObj(p) &&
+					p.field === key &&
+					p.operator === 'eq' &&
+					// it's only valid if it's not nullish.
+					// (IDB will throw a fit if it's nullist.)
```
svidgen, line 516: Typo. "nullist" -> "nullish"

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -443,6 +443,17 @@ class IndexedDBAdapter implements Adapter {
 		const hasPagination = pagination && pagination.limit;
 
 		const records: T[] = (await (async () => {
+			//
+			// NOTE: @wirej explored removing this and letting query() take care of automatic
```
svidgen, line 447: @wirej -> @svidgen

svidgen:
Converting back to draft for now. Tests against the sample app show that `readonly` fields are being included as inputs to the model input for create mutations. This wasn't caught by any tests. Adding tests and fixing ...

dpilch (dismissed):

iartemiev:
From reading the code it seems like this doesn't address the scenario where I set a value to `undefined` in an update. Is that correct? 

For example:
Save new record, set optional value
```js
  value: 'some value'
  otherValue: 'optional value'
}));
```
Update the optional value to `undefined`
```js
const updated = await DataStore.save(Model.copyOf(saved, (updated) => {
  updated.otherValue = undefined;
});

console.log(updated); // {  ..., otherValue: undefined }

// wait for sync
console.log(await DataStore.query(Model, saved.id)); // {  ..., otherValue: null }
```

svidgen:
> From reading the code it seems like this doesn't address the scenario where I set a value to `undefined` in an update. Is that correct?

Digging on this now. There was definitely a gap in test-cases for `copyOf()`. So, I'm filling those cases in, and I'll see about the update behavior. I'd forgotten we may have had existing bugs around that as well. But, it seems closely-related enough to take a look.

svidgen:
Putting this back into a draft state while I investigate your question @iartemiev. I'm seeing some odd behavior trying to add tests for it, and I want to rule out the possibility of introducing a regression on updates. 😓 

Thanks for poking at it!

svidgen:
@iartemiev I've added test cases to cover your question. Everything is working as expected -- I had reverted to a draft temporarily here out of an abundance of caution because I was seeing something unexpected in a test. But, it was a mistake in the test.

So, to answer your question, that case [*is* covered](https://github.com/aws-amplify/amplify-js/pull/10915/files#diff-10866b5ac7d469b38dd9928d9d7056a3a402904fef591ea8cfe773b4d75d63efR880). And, now it's tested as well. 😅  (See [DataStore.ts:2733](https://github.com/aws-amplify/amplify-js/pull/10915/files#diff-37f5c3d40431d210a7024204c661c85211ef5d7bb578d7a823a86375f75fcd18R2733-R2754) and [connectivityHandling.test.ts:134](https://github.com/aws-amplify/amplify-js/pull/10915/files#diff-68fad1b028a70ed30e99e616e38d45ec29a2993942226f28ff94cdb241bb3b25R134-R161).)

packages/datastore/src/datastore/datastore.ts
```diff
@@ -741,20 +740,56 @@ const castInstanceType = (
...
+ *
+ * @param modelDefinition Definition for the draft. Used to discover all fields.
+ * @param draft The instance draft to apply normalizations to.
+ */
+const normalize = <T extends PersistentModel>(
+	modelDefinition: SchemaModel | SchemaNonModel,
+	draft: Draft<T>
+) => {
+	for (const k of Object.keys(modelDefinition.fields)) {
+		if (draft[k] === undefined) (<any>draft)[k] = null;
```
david-mcafee, line 789: Should we include some kind of warning, here? For instance, if the customer is explicitly setting this to `undefined`, and thus would be expecting it to be `undefined`?

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -495,7 +506,16 @@ class IndexedDBAdapter implements Adapter {
...
+				p =>
+					// it's a relevant predicate object only if it's an equality
+					// operation for a key field from the key:
+					isPredicateObj(p) &&
+					p.field === key &&
+					p.operator === 'eq' &&
+					// it's only valid if it's not nullish.
+					// (IDB will throw a fit if it's nullish.)
+					p.operand !== null &&
+					p.operand !== undefined
```
david-mcafee, line 518: What about the Async Adapter?

packages/datastore/src/predicates/next.ts
```diff
@@ -141,7 +143,7 @@ export class FieldCondition {
 	constructor(
 		public field: string,
 		public operator: string,
-		public operands: string[]
+		public operands: any[]
```
david-mcafee, line 146: Why `any`?

david-mcafee (commented):
Left a few comments, mostly questions :)

packages/datastore/src/datastore/datastore.ts
svidgen: I could see cases where this might be helpful. But, it'll also be *very* noisy, and I feel that it normalizes using warnings to tell the customer "normal things are happening" instead of "things you might not have meant to do!" Any time a customer saves an object and omits *any optional fields*, for example, they'd see the warning.

I still feel like I can be convinced this is a good idea. But, it feels like a hard sell at the moment.

packages/datastore/src/predicates/next.ts
svidgen: That's really odd. I have literally no recollection of why I changed this.

Good catch!

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
svidgen: The error was specifically an indexed DB error. The largest barrage of failures that led to this change were the related model tests. And those tests are also executed in the Async Adapter suite. So, there's no Async Adapter impact that I could see.

packages/datastore/__tests__/DataStore.ts
```diff
@@ -2699,15 +2699,58 @@ describe('DataStore tests', () => {
...
+			});
+
+			test('copyOf() setting optional field to null', () => {
+				const original = new Model({
+					field1: 'someField',
+					dateCreated: new Date().toISOString(),
+					optionalField1: 'defined value',
+				});
+				const copied = Model.copyOf(original, d => (d.optionalField1 = null));
+				expect(copied.optionalField1).toBe(null);
```
iartemiev, line 2740: I already tested this locally, so I know it works as expected, but I think we should also ensure that  optional fields with a value that are omitted from the `copyOf` aren't accidentally nullified.

```suggestion
				const emailsVal = ['test@test.test'];
				const original = new Model({
					field1: 'someField',
					dateCreated: new Date().toISOString(),
					optionalField1: 'defined value',
					emails: emailsVal,
				});
				const copied = Model.copyOf(original, d => (d.optionalField1 = null));
				expect(copied.optionalField1).toBe(null);
				expect(copied.emails).toEqual(emailsVal);
```

iartemiev (dismissed):
LGTM 👍🏻

packages/datastore/src/datastore/datastore.ts
dpilch: I agree warning would be too noisy. I think docs update would be sufficient.

dpilch (approved):

iartemiev (approved):

packages/datastore/src/datastore/datastore.ts
svidgen: Sounds good. I've added a sprint item for the docs update.

Issue #10922, opened 2023-02-03
DataStore clear, start, clear results in runtime error


dpilch:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Not applicable

### Amplify APIs

DataStore

### Amplify Categories

api

### Environment information

<details>

```
# Put output below this line

  System:
    OS: macOS 13.2
    CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
    Memory: 116.51 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 14.18.0 - ~/.nvm/versions/node/v14.18.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v14.18.0/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v14.18.0/bin/npm
    Watchman: 2023.01.30.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 109.0.5414.119
    Firefox: 102.7.0
    Safari: 16.3
  npmPackages:
    @aws-amplify/ui-react: ^3.5.11 => 3.6.1 
    @aws-amplify/ui-react-internal:  undefined ()
    @aws-amplify/ui-react-legacy:  undefined ()
    @testing-library/jest-dom: ^5.16.5 => 5.16.5 
    @testing-library/react: ^13.4.0 => 13.4.0 
    @testing-library/user-event: ^13.5.0 => 13.5.0 
    @types/jest: ^27.5.2 => 27.5.2 (29.4.0)
    @types/node: ^16.18.3 => 16.18.11 (18.11.18)
    @types/react: ^18.0.24 => 18.0.27 
    @types/react-dom: ^18.0.8 => 18.0.10 
    aws-amplify: unstable => 5.0.12-unstableV5.10+bbf4a4cbf 
    react: ^18.2.0 => 18.2.0 
    react-dom: ^18.2.0 => 18.2.0 
[...]

PR #10923, opened 2023-02-03
fix: prevent double full sync at start

dpilch:
#### Description of changes

Prevent a double full sync at start by setting correct time to next full sync.

On DataStore start a second full sync was immediately scheduled. On a full sync the time to next full sync is calculated. On the first full sync one of the variables in the calculation is `null` which is then cast to 0. This resulted in a negative time to next sync. When supplying a negative number to `setTimeout` the callback is executed at next event cycle. 

This resulted in a second fully sync scheduled immediately. However, when running clear, start, clear in succession (like in https://github.com/aws-amplify/amplify-js/issues/10922) the idb is deleted by clear. This causes a runtime error because the full sync expects the `sync_ModelMetadata` table to exist.

I think this is getting deeper into JS event loop. There are 5 async events that are scheduled.

* Full sync A
* Full sync B
* Clear A
* Clear B
* Start

First glance the order would be:

1. Clear A
2. Start
3. Full Sync A
4. Full Sync B (due to negative timeout)
5. Clear B

It appears the order is actually scheduled as:

1. Clear A
2. Start
3. Full Sync A
4. Clear B
5. Full Sync B

I am still not sure as to why the clear does not terminate the full Sync B.

#### Issue #, if available

Resolves https://github.com/aws-amplify/amplify-js/issues/10922

Using minimum repro from issue.

Todo: unit test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


dpilch:
I'm not able to reproduce the error in unit tests so I will need to make a E2E test.

dpilch:
E2E tests are in a PR for when this is merged.

packages/datastore/src/sync/index.ts
```diff
@@ -721,10 +721,15 @@ export class SyncEngine {
...
+						// Assume newestStartedAt is is also newest full sync.
 						const msNextFullSync =
-							newestFullSyncStartedAt! +
-							theInterval! -
-							(newestStartedAt! + duration!);
+							newestFullSyncStartedAt! === null
+								? theInterval! - duration!
+								: newestFullSyncStartedAt! +
+								  theInterval! -
+								  (newestStartedAt! + duration!);
```
iartemiev, line 732: Non-blocking suggestion: the change makes sense, but I think this ternary is getting a bit unruly. 

Might be out of scope here, but the naming of those variables is pretty confusing as well IMO. I would change `theInterval` -> `syncInterval`
`duration` -> `syncDuration`
`newestFullSyncStartedAt` -> `lastFullSyncStartedAt`
`newestStartedAt` -> `lastStartedAt`

Then break apart the ternary like:

```suggestion
						let msNextFullSync;

						if (lastFullSyncStartedAt! === null) {
							msNextFullSync = syncInterval! - syncDuration!;
						} else {
							msNextFullSync =
								lastFullSyncStartedAt! +
								syncInterval! -
								(lastStartedAt! + syncDuration!);
						}
```

iartemiev (approved):

packages/datastore/src/sync/index.ts
```diff
@@ -721,10 +721,19 @@ export class SyncEngine {
...
-						const msNextFullSync =
-							newestFullSyncStartedAt! +
-							theInterval! -
-							(newestStartedAt! + duration!);
+						// null is cast to 0 resulting in unexpected behavior
+						// If lastFullSyncStartedAt is null this is the first sync.
+						// Assume lastStartedAt is is also newest full sync.
+						let msNextFullSync;
+
+						if (lastFullSyncStartedAt! === null) {
```
iartemiev, line 729: Can this ever be `undefined`? Or is it guaranteed to be either `null` or a Number?

dpilch: Looks like it could be `undefined` if something happened to how model metadata was saved.

So unlikely but possible, just like the original runtime error that lead to this fix...

I'll update the condition.

iartemiev (approved):

svidgen (approved):
I'm OK approving on the grounds that the fix looks consistent with the existing sync timeout math.

But, do you see _any_ path forward to unit testing this in any way? Are there any existing unit tests?

I'd really really like to see tests that represent the table of expected `current sync status -> next sync time / timeout` values that clearly demonstrates what our sync interval math is intended to do. If this is too far out of reach for this PR, perhaps a backlog item would be appropriate.

dpilch:
We currently don't have any direct coverage of this file for unit tests. It should be possible, but starting from scratch would be a large effort for this change.

I think something on the backlog would be a good idea though.

Issue #10785, opened 2022-12-16
Sync Expressions in DataStore in Amplify@5 with old API's break the sync


svidgen:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Not applicable

### Amplify APIs

DataStore

### Amplify Categories

api

### Environment information

<details>

```
# Put output below this line


```

</details>

### Describe the bug

I'm not sure what version/flags of the CLI is necessary to create this scenario. But, I have an "old" sample app wherein the FK fields from some connected models are not available for in the GQL sync query selection set.

I have a `Blog` model where I can provide a `blogOwnerId` at mutation time:

<img width="192" alt="image" src="https://user-images.githubusercontent.com/8375502/208199658-39c06c9a-78a5-4fbd-b9f9-619ad02b5122.png">

However, `blogOwnerId` is not available in the selection set.

<img width="200" alt="image" src="https://user-images.githubusercontent.com/8375502/208199428-acb0bd8a-9e3a-4286-bc84-fd467600a9a5.png">

When I add a `syncExpression` on `Blog`, `blogOwnerId` is included to the selection set on sync, and the sync therefore fails.

### Expected behavior

DataStore should only include these connection/FK fields in the selection set if they're explicitly defined.

### Reproduction steps

🤷

### Code Snippet

```javascript
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line


```

</details>

micstepper:
Any updates @svidgen ? Did you find a solution?

svidgen:
@micstepper I started poking at it, but no solution yet. I'm planning on focusing more on this over the next couple weeks.

svidgen:
This fix should be out in yesterday's fix.

PR #10929, opened 2023-02-06
fix(@aws-amplify/datastore): extra fields in selective sync selection set on API's from older CLI

svidgen:
#### Description of changes

Fixes an issue where the selection for selective sync expressions was incorrect for apps build with some older versions of the CLI. Related-model key fields were included in the selection set, but were not recognized by cloud storage. This issue was introduced in with the nested predicates feature, which added the implicitly created FK fields to the table metadata `fields` list so they could be seen by the predicate builder.

This PR removes the logic that augments the `fields` list and add a new `allFields` list. All pre-existing (prior to lazy loading/nested predicates) will revert to use only the explicitly defined `fields` list. The predicates builder will use `allFields`.

#### Issue #, if available

#10785

1. Added testing.
2. *Pending.* Manual sample apps validation.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


svidgen:
Addressed in https://github.com/aws-amplify/amplify-js/pull/10949

PR #10932, opened 2023-02-06
fix(@aws-amplify/datastore): strictly define null vs undefined behavior on models

svidgen:
#### Resubmitted

Originally submitted as #10915, which caused failing integ tests, due to `owner: null` being submitted in mutations when the the `owner` field was made explicit. AppSync rejects `owner: null` as "Unauthorized."

This updated PR omits the owner field in mutations (or custom owner field per model `attributes.auth`) when the value would be `null` or `undefined`.

Added tests for `owner` handling are in [connectivityHandling.test.ts:102-168](https://github.com/aws-amplify/amplify-js/pull/10932/files#diff-68fad1b028a70ed30e99e616e38d45ec29a2993942226f28ff94cdb241bb3b25R102-R168), with owner field validation occurring in [the graphql fake in helper.ts](https://github.com/aws-amplify/amplify-js/pull/10932/files#diff-d80854c979fb9c08302352685e794dcd2cd3c7a4d29234d90b7757061d14ca5eR779-R781).

The fix is for `owner` handling is in `mutation.ts` on [lines 487-490](https://github.com/aws-amplify/amplify-js/pull/10932/files#diff-caf2490eda4f0fb658881cae74f9e9b5db5e70363b8d0a737bd7bdaca51e053aR487-R490) and [lines 526-529](https://github.com/aws-amplify/amplify-js/pull/10932/files#diff-caf2490eda4f0fb658881cae74f9e9b5db5e70363b8d0a737bd7bdaca51e053aR526-R529).

You can also see the bulk of the `owner` field fix and tests together by looking at the [2nd commit in the PR](https://github.com/aws-amplify/amplify-js/pull/10932/commits/bc756645a40317454cdfa6e2e94c4952a93efe09).

#### Description of changes

Given:

```graphql
type Model @model {
  id: ID!
  value: String!
  otherValue: String
}
```

If we save an instance:

```js
    value: 'some value'
}));
```

We'll have this returned:

```json
{
  "value": "some value",
}
```

```json
{

  // LOOK!
  "otherValue": null, 

  "owner": "test",
  "_version": 1,
  "_deleted": null
}
```

2. Verified in sample apps.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


svidgen:
Spot-checked failing integ tests. They are now passing locally.

image

packages/datastore/src/sync/processors/mutation.ts
```diff
@@ -484,6 +484,11 @@ class MutationProcessor {
 		const modelDefinition = this.schema.namespaces[namespaceName].models[model];
 		const { primaryKey } = this.schema.namespaces[namespaceName].keys![model];
 
+		const auth = modelDefinition.attributes?.find(a => a.type === 'auth');
+		const ownerField = auth?.properties?.rules.find(
+			rule => rule.ownerField
+		)?.ownerField;
```
iartemiev, line 490: Seems like this only handles a single owner field, but models can have multiple owner fields configured, e.g. 
```gql
type Post @model @auth(rules: [ 
  { allow: owner }, 
  { allow: owner, ownerField: editor, operations: [read, update] } 
]) {
  id: ID!
  owner: String
  editor: String
}
```
I think we should do `auth?.properties?.rules.filter` and then iterate over the resulting array below

packages/datastore/package.json
```diff
@@ -24,8 +24,7 @@
 		"build:cjs:watch": "node ./build es5 --watch",
 		"build:esm:watch": "rimraf lib-esm && node ./build es6 --watch",
 		"build": "yarn clean && yarn build:esm && npm run build:cjs",
-		"clean": "npm run clean:size && rimraf lib-esm lib dist",
-		"clean:size": "rimraf dual-publish-tmp tmp*",
```
dpilch, line 28: Accidental removal?

svidgen: Ah yes. Bad merge. Fixing...

packages/datastore/src/sync/processors/mutation.ts
svidgen: That's a pretty important gap to call out! 😅 Thank-you!

svidgen: Fixed.

dpilch (approved):

iartemiev (approved):

PR #10933, opened 2023-02-06
bug(analytics): Remove timestamp property from MediaAutoTrack

stocaaro:
#### Description of changes
Removing timestamp from Personalization properties list as it is set service side and raises an exception when provided by the client.

#### Issue #, if available
https://github.com/aws-amplify/amplify-js/issues/6906

Manual testing.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/analytics/src/Providers/AmazonPersonalizeHelper/MediaAutoTrack.ts
```diff
@@ -173,20 +173,18 @@ export class MediaAutoTrack {
 		const newParams = Object.assign({}, this._params);
 		const { eventData } = newParams;
 		eventData.eventType = eventType;
+		let currentPlayTime;
```
jimblanc, line 176: Do we want to surface this value in `eventData.properties` somewhere as well so customers can get the value that was formerly in `timestamp`?

jimblanc (approved):

packages/analytics/src/Providers/AmazonPersonalizeHelper/MediaAutoTrack.ts
elorzafe, line 176: Should this be typed ?

stocaaro: Good call. Updated

elorzafe (approved):
🚢 

Thanks @stocaaro 🎖️

jimblanc (approved):

PR #10941, opened 2023-02-08
chore(notifications): Extract common functionalities

cshfang:
#### Description of changes
This PR extracts common functionalities that will be used by both the existing In-App and upcoming Push Notification channels.

All changes in this PR are *additive* and thus introduces code duplication. Future PRs will update existing implementation to reference the extracted functionalities while removing the duplication.

Tested changes locally. Ran `yarn test`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/notifications/src/common/AWSPinpointProviderCommon/index.ts
+++ b/packages/notifications/src/common/AWSPinpointProviderCommon/index.ts
@@ -1,0 +1,288 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+	ClientDevice,
+	ConsoleLogger,
+	Credentials,
+	getAmplifyUserAgent,
+	StorageHelper,
+	transferKeyToUpperCase,
+} from '@aws-amplify/core';
+import { Cache } from '@aws-amplify/cache';
+import {
+	Event as AWSPinpointAnalyticsEvent,
+	UpdateEndpointCommand,
+	UpdateEndpointCommandInput,
+	PinpointClient,
+	PutEventsCommand,
+	PutEventsCommandInput,
+} from '@aws-sdk/client-pinpoint';
+import { v4 as uuid } from 'uuid';
+
+import {
+	NotificationsCategory,
+	NotificationsSubCategory,
+	NotificationsProvider,
+	UserInfo,
+} from '../../types';
+import { AWSPinpointUserInfo } from './types';
+
+export default abstract class AWSPinpointProviderCommon
+	implements NotificationsProvider
+{
+	static category: NotificationsCategory = 'Notifications';
+	static providerName = 'AWSPinpoint';
+
+	protected clientInfo;
+	protected config: Record<string, any> = {};
+	protected endpointInitialized = false;
+	protected initialized = false;
+	protected logger: ConsoleLogger;
+
+	constructor(logger) {
+		this.config = { storage: new StorageHelper().getStorage() };
+		this.clientInfo = ClientDevice.clientInfo() ?? {};
+		this.logger = logger;
+	}
+
+	/**
+	 * get the category of the plugin
+	 */
+	getCategory() {
+		return AWSPinpointProviderCommon.category;
+	}
+
+	/**
+	 * get the sub-category of the plugin
+	 */
+	abstract getSubCategory(): NotificationsSubCategory;
+
+	/**
+	 * get provider name of the plugin
+	 */
+	getProviderName(): string {
+		return AWSPinpointProviderCommon.providerName;
+	}
+
+	configure(config = {}): Record<string, any> {
+		this.config = { ...this.config, ...config };
+		this.logger.debug(
+			`configure ${this.getProviderName()}${this.getSubCategory()}Provider`,
+			this.config
+		);
+		return this.config;
+	}
+
+	identifyUser = async (userId: string, userInfo: UserInfo): Promise<void> => {
+		if (!this.initialized) {
+			await this.init();
+		}
+		try {
+			await this.updateEndpoint(userId, userInfo);
+		} catch (err) {
+			this.logger.error('Error identifying user', err);
+			throw err;
+		}
+	};
+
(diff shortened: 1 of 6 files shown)


packages/notifications/src/common/index.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export { default as AWSPinpointProviderCommon } from './AWSPinpointProviderCommon';
+export * from './eventListeners';
```
jimblanc, line 5: We should avoid * exports from barrel files to ensure optimal tree shaking across different bundlers.

packages/notifications/src/common/AWSPinpointProviderCommon/index.ts
```diff
@@ -0,0 +1,288 @@
...
+					OptOut: optOut ?? endpointInfo.optOut,
+					User: {
+						UserId:
+							userId ?? endpointInfo.userId ?? currentCredentials.identityId,
+						UserAttributes: attributes ?? endpointInfo.userAttributes,
+					},
+				},
+			};
+			const command: UpdateEndpointCommand = new UpdateEndpointCommand(input);
+			this.logger.debug('updating endpoint', input);
```
jimblanc, line 224: Nit: I think this was previously approved, but any risk leaking info in logs here from the input?

jimblanc (commented):
Looks good, some minor comments.

jimblanc (approved):

packages/notifications/src/common/AWSPinpointProviderCommon/index.ts
```diff
@@ -0,0 +1,288 @@
...
+		const cachedEndpointId = await Cache.getItem(cacheKey);
+		// Found in cache, just return it
+		if (cachedEndpointId) {
+			return cachedEndpointId;
+		}
+		// Otherwise, generate a new ID and store it in long-lived cache before returning it
+		const endpointId = uuid();
+		// Set a longer TTL to avoid endpoint id being deleted after the default TTL (3 days)
+		// Also set its priority to the highest to reduce its chance of being deleted when cache is full
+		const ttl = 1000 * 60 * 60 * 24 * 365 * 100; // 100 years
```
stocaaro, line 266: Should this constant be pulled to the top of the file?

packages/notifications/src/common/AWSPinpointProviderCommon/index.ts
```diff
@@ -0,0 +1,288 @@
...
+					}),
+					Metrics: {
+						...endpointInfo.metrics,
+						...metrics,
+					},
+					OptOut: optOut ?? endpointInfo.optOut,
+					User: {
+						UserId:
+							userId ?? endpointInfo.userId ?? currentCredentials.identityId,
+						UserAttributes: attributes ?? endpointInfo.userAttributes,
```
stocaaro, line 219: I'm sure this code is just moved, but it makes me wonder if we've integ tested all the various ways these can collapse down to make sure they all work and are providing value in composing request inputs.

stocaaro: Also, don't we early exit when `userId` is falsy? Is it possible that it would be `undefined | null` now such that we would exercise the other options?

stocaaro (commented):
Handful of nits. What does this change do to test coverage? Guessing in the usage PR, these get exercised, removing existing code, and we get roughly the same coverage post both changes. Any value in testing these common components in isolation from their usage?

stocaaro:
My testing comment looks largely covered by [this pr](https://github.com/aws-amplify/amplify-js/pull/10953)

packages/notifications/src/common/AWSPinpointProviderCommon/index.ts
```diff
@@ -0,0 +1,288 @@
...
+	protected clientInfo;
+	protected config: Record<string, any> = {};
+	protected endpointInitialized = false;
+	protected initialized = false;
+	protected logger: ConsoleLogger;
+
+	constructor(logger) {
+		this.config = { storage: new StorageHelper().getStorage() };
+		this.clientInfo = ClientDevice.clientInfo() ?? {};
+		this.logger = logger;
```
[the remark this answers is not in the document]
cshfang: This was the result of sharing code through inheritance, i.e.
InAppProvider subclasses CommonProvider
PushProvider subclasses CommonProvider

Since CommonProvider is actually also a concrete class, it too wants to log - but it should be logging subclassed logs. We currently just don't do this anywhere else.

packages/notifications/src/common/AWSPinpointProviderCommon/index.ts
cshfang: This was copied over from Analytics and normally I'd agree extracting makes sense, but in this case I feel the TTL is so specific to this section of code that having to look it up elsewhere actually becomes a burden especially since it's never reused anywhere else.

Lmk if you disagree, I am not averse to changing but I wanted to share a different perspective

packages/notifications/src/common/AWSPinpointProviderCommon/index.ts
cshfang: I believe the early exit has other qualifying criteria so there are scenarios where userId can be falsy but still arrive here. You're definitely right that this was more or less copied over from Analytics (when we did InApp) though. *This* is really just *moved* from InApp

stocaaro: You're right. AND vs OR

packages/notifications/src/common/AWSPinpointProviderCommon/index.ts
stocaaro: Yeah. Searching through this isn't the only embedded constant, but most constants are pulled to the top. Would vote for moving it for consistency, but its a minor nit and our codebase isn't consistent. Not going to gate approval on this.

stocaaro (approved):

PR #10947, opened 2023-02-08
fix(auth): sends current user with attributes to hub

haverchuck:
#### Description of changes
This addresses an issue in which the `signIn` hub event after `confirmSignIn` invocation contains a `CognitoUser` that does not have attributes. The issue is that we were broadcasting the user _passed_ to the `confirmSignIn` method, which represented the state of the user in the midst of the sign in flow. We solved this by calling `currentUserPoolUser` immediately before dispatching the event and adding the attributes to existing user object (we also pass the updated user to the `resolve` in the auth callback). We then add the missing value to the existing object. We maintain the same object though. 

Finally, there are other APIs  with signIn Hub dispatch sites that could potentially benefit from this fix.

Manual Testing

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/auth/src/Auth.ts
```diff
@@ -1182,14 +1186,13 @@ export class AuthClass {
 						} catch (e) {
 							logger.debug('cannot get cognito credentials', e);
 						} finally {
-							that.user = user;
-
+							const currentUser = await this.currentUserPoolUser();
```
elorzafe, line 1189: You might want to wrap that into a `try-catch` in case that method throws?

What do you think?

haverchuck: We don't do that in signIn API, but makes sense.

jimblanc (approved):

elorzafe (approved):
I am good with this. Only comment is if you can add a unit test where the `this.currentUserPoolUser` fails and you can still see the user on the hub event

packages/auth/src/Auth.ts
```diff
@@ -1183,7 +1183,13 @@ export class AuthClass {
 							logger.debug('cannot get cognito credentials', e);
 						} finally {
 							that.user = user;
-
+							let currentUser;
```
elorzafe, line 1186: Nit: this is not a `const` inside of try-catch block?

elorzafe (approved):
Looks good to me

Just a minor nit pick comment

jimblanc (approved):

elorzafe (approved):

jimblanc (approved):

PR #10949, opened 2023-02-08
fix(@aws-amplify/datastore): extra fields in selective sync selection set on API's from older CLI

svidgen:
#### Description of changes

Fixes an issue where the selection set for selective sync expressions was incorrect for apps built with some older versions of the CLI. Related-model key fields were included in the selection set, but were not recognized by cloud storage. This issue was introduced in with the nested predicates feature, which added the implicitly created FK fields to the table metadata `fields` list so they could be seen by the predicate builder.

This PR removes the logic that augments the `fields` list and add a new `allFields` list. All pre-existing logic and components (prior to lazy loading/nested predicates) will revert to use only the explicitly defined `fields` list. The predicates builder will then use `allFields`.

~To support this, storage predicate construction is refactored slightly to avoid using the storage predicate (v1 predicate) builder. The old builder complains when a field from `allFields` shows up, even though the field exist. So, we are instead constructing an AST and transforming it.~

~The resulting storage predicate can, at times, contain joins against `null` or `undefined` values. This happens when a relationship is optional. So, the IndexedDBAdapter was also updated to avoid attempting a `index.get()` on these values, which otherwise causes IndexedDB to throw a fit.~

(Refactors around storage predicate generation and related IndexedDB fixes were handled upstream.)

#### Issue #, if available

#10785

1. Added testing.
4. Manual sample apps validation: Red/Green'd `many-to-many` app with added selective syncs. [That PR](https://github.com/aws-amplify/amplify-js-samples-staging/pull/530) can be merged after this one.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/src/predicates/next.ts
```diff
@@ -169,6 +187,37 @@ export class FieldCondition {
...
+	 * @param model The model meta to use when construction a new `GroupCondition`
+	 * for cases where the negation requires multiple `FieldCondition`'s.
+	 */
+	negated(model: ModelMeta<any>) {
+		if (this.operator === 'between') {
+			return new GroupCondition(model, undefined, undefined, 'or', [
+				new FieldCondition(this.field, 'lt', [this.operands[0]]),
+				new FieldCondition(this.field, 'gt', [this.operands[1]]),
+			]);
+		} else {
```
dpilch, line 212: I think I missed this on a previous review. What about `beginsWith`?

svidgen: Looks like an existing bug. I'm adding a bunch of negation tests and will fix this in another PR ASAP.

svidgen: https://github.com/aws-amplify/amplify-js/issues/10957

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
```diff
@@ -443,6 +443,17 @@ class IndexedDBAdapter implements Adapter {
 		const hasPagination = pagination && pagination.limit;
 
 		const records: T[] = (await (async () => {
+			//
```
david-mcafee, line 446: What about the other storage adapters? Is only IDB affected?

packages/datastore/src/predicates/next.ts
svidgen: Fix: https://github.com/aws-amplify/amplify-js/pull/10962

svidgen: Fixed upstream.

packages/datastore/src/storage/adapter/IndexedDBAdapter.ts
svidgen: Discussed offline and incorporated upstream. But, for posterity, these changes to IndexedDBAdapter handle idiosyncrasies specific to IndexedDB, which throws a fit if you try to `index.get([null])`. The changes here would prevent that query pattern, and force execution through an `index.queryAll(<predicate>)` line of execution instead.

dpilch (approved):

iartemiev (approved):

PR #10951, opened 2023-02-09
chore(notifications): Refactor in-app messaging to use common code

cshfang:
#### Description of changes
This PR refactors the existing Notifications and InAppMessaging feature to use common code introduced by #10941 

Tested feature continued to work in the expected manner with no changes to DX

Changes necessitated unit test updates that will be added in a future PR

#### Checklist
- [x] PR description included


packages/notifications/src/InAppMessaging/Providers/AWSPinpointProvider/index.ts
```diff
@@ -155,14 +129,14 @@ export default class AWSPinpointProvider implements InAppMessagingProvider {
 			const command: GetInAppMessagesCommand = new GetInAppMessagesCommand(
 				input
 			);
-			logger.debug('getting in-app messages', input);
+			this.logger.debug('getting in-app messages');
```
stocaaro, line 132: Thought this looked odd, but made a related comment in the prior PR that adopts this pattern

stocaaro (approved):

packages/notifications/src/Notifications.ts
```diff
@@ -2,14 +2,16 @@
...
 import { NotificationsCategory, NotificationsConfig } from './types';
+import InAppMessaging from './InAppMessaging';
+import { PushNotificationInterface as PushNotification } from './PushNotification/types';
 
 const logger = new Logger('Notifications');
 
 class NotificationsClass {
 	private config: Record<string, any> = {};
 	private inAppMessaging: InAppMessaging;
+	private pushNotification?: PushNotification;
```
jimblanc, line 14: Should we support the same optional initialization for the in-app sub-category?

jimblanc (approved):

packages/notifications/src/Notifications.ts
jimblanc: Discussed offline

packages/notifications/package.json
```diff
@@ -67,6 +67,9 @@
 			"limit": "70.5 kB"
 		}
 	],
+	"devDependencies": {
+		"@aws-amplify/rtn-push-notification": "0.0.1"
```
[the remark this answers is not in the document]
cshfang: Yeah I think we generally have this pattern where we use devDependencies for RN only especially since peerDependencies like to install themselves on NPM.

packages/notifications/src/Notifications.ts
cshfang: Chatted offline about this but the rationale for this was 2-fold:
1. The conditional require doesn't save anything in bundle size anyway
2. Not wanting to break the current InApp use case

Our docs currently suggest doing:
```
import { Notifications } = 'aws-amplify';
const { InAppMessaging } = Notifications;
```

Which is a miss since that would only work if InAppMessaging is constructed at the same time as Notifications, which would not be the case if we used conditional require. I think the above was admittedly not a good pattern for a getter in general :(

PR #10954, opened 2023-02-09
feat(pubsub): Typescript improvements and cleanup

stocaaro:
#### Description of changes
Changes made:
- Use `typescript-coverage-report` output to identify and fix ts issues
  - TS coverage is now 94.83% and gated at regression below 93%
- Re-add coverage protection for the build and update to omit `paho.js`
- Removed the unused code in `AWSAppSyncProvider.ts`
- Removed paho as a dependency (we're using the vendor lib version right now).

TS Coverage:
- [Before](https://before.dvo88zydva5gf.amplifyapp.com/) ~**88%** (ignoring Paho), shows 79.9% with Paho counted
- [After](https://after.dvo88zydva5gf.amplifyapp.com/) **94.83%**

- Fixed unit testing issues
- Manually tested against IoT/API/Datastore

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/pubsub/src/Providers/AWSIotProvider.ts
```diff
@@ -15,8 +15,8 @@ export class AWSIoTProvider extends MqttOverWSProvider {
 		super(options);
 	}
 
-	protected get region() {
-		return this.options.aws_pubsub_region;
+	protected get region(): string | undefined {
```
cshfang, line 18: This is a good example of when we would want `undefined` for sure

packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -37,19 +37,39 @@ export function mqttTopicMatch(filter: string, topic: string) {
 export interface MqttProviderOptions extends ProviderOptions {
 	clientId?: string;
 	url?: string;
+	aws_pubsub_endpoint?: string;
+}
+
+interface PahoClient {
```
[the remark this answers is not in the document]
cshfang: Should we use the official typing for paho-mqtt?
https://www.npmjs.com/package/@types/paho-mqtt

packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -106,10 +126,10 @@ export class MqttOverWSProvider extends AbstractPubSubProvider {
 	}
 
 	protected get clientId() {
-		return this.options.clientId;
+		return this.options.clientId!;
 	}
 
-	protected get endpoint() {
+	protected get endpoint(): string | undefined | Promise<string> {
```
cshfang, line 132: In what use case would this be a promise? This isn't a public API but it seems like if it's sometimes a Promise, it'd have to be awaited so why not just make it always a Promise?

packages/pubsub/src/PubSub.ts
```diff
@@ -144,7 +144,7 @@ export class PubSubClass {
 
 		const providers = this.getProviders(options);
 
-		return new Observable(observer => {
+		return new Observable<any>(observer => {
```
cshfang, line 147: Can this be `null` or `undefined`?

cshfang (commented):
I left some comments. There's nothing blocking but this is a Typescript improvement PR after all so I was being somewhat nit-picky. Some of it probably just stems from my unfamiliarity with pubsub so thank you for bearing with me.

packages/pubsub/src/PubSub.ts
stocaaro: Good call. This can be `object`, which is nominally better.

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: This function is overridden in the [IoT Provider](https://github.com/aws-amplify/amplify-js/blob/main/packages/pubsub/src/Providers/AWSIotProvider.ts#L27), which returns a `Promise<string>`

packages/pubsub/src/Providers/MqttOverWSProvider.ts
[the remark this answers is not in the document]
stocaaro: Excellent call. Pulling it in.

packages/pubsub/src/PubSub.ts
stocaaro: And now I've seen the object thoughts above and seen the error of my ways 😂

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -165,7 +170,11 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 		throw new Error('Not used here');
 	}
 
-	public async publish(_topics: string[] | string, _msg: any, _options?: any) {
+	public async publish(
+		_topics: string[] | string,
+		_msg: object,
+		_options?: AWSAppSyncRealTimeProviderOptions
```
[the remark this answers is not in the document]
stocaaro: I'm not sure if its idiomatic in JS/TS, but in other languages it is used to denote variables that are not used. In this case, I read it as, we take the standard inputs, but then no matter what raise, so these variables don't matter.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -48,7 +48,7 @@ import {
 
 const logger = new Logger('AWSAppSyncRealTimeProvider');
 
-const dispatchApiEvent = (event: string, data: any, message: string) => {
+const dispatchApiEvent = (event: string, data: {}, message: string) => {
```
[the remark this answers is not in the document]
stocaaro: I think I have adopted this convention in all cases.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -260,8 +269,9 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 	}
 
 	protected get isSSLEnabled() {
-		return !this.options
-			.aws_appsync_dangerously_connect_to_http_endpoint_for_testing;
+		return !this.options[
+			'aws_appsync_dangerously_connect_to_http_endpoint_for_testing'
```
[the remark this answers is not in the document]
stocaaro: I made a run at rewriting the `ProviderOptions` interface as a `Record` type. Turned into cascading problems that couldn't be teased out in the 20 minutes I gave it. I'm sure it can be done, but since this is a key customer facing interface, I would prefer to leave it or wrap it in more testing before retooling as a `type`.

In this case, we have added all other variables as fields on the interface. We opted not to add this one as the name isn't something we want folks autocompleting to / seeing in the public surface. Which is why TS prefers not to directly access it.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -495,10 +509,19 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
+		}
 		logger.debug(
 			`subscription message from AWS AppSync RealTime: ${message.data}`
 		);
-		const { id = '', payload, type } = JSON.parse(message.data);
+		const {
+			id = '',
+			payload,
+			type,
+		}: { id: string; payload: DataObject; type: string } = JSON.parse(
```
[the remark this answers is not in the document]
stocaaro: Ack

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -655,15 +678,15 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
...
+						(await this._awsRealTimeHeaderBasedAuth({
 							authenticationType,
 							payload: payloadString,
 							canonicalUri: '/connect',
 							apiKey,
 							appSyncGraphqlEndpoint,
 							region,
 							additionalHeaders,
-						})
+						})) ?? ''
```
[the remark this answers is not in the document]
stocaaro: Fixed.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -69,11 +73,29 @@ const customDomainPath = '/realtime';
 
 type GraphqlAuthModes = keyof typeof GRAPHQL_AUTH_MODE;
 
+type DataObject = {
```
cshfang, line 76: 🤩

packages/pubsub/src/Providers/MqttOverWSProvider.ts
cshfang: Eesh. In that case, would it be better to align with that and wrap the below in a resolved Promise? If not possible then oh well 😞

cshfang (approved):
Have a couple lingering comments but I don't think they're blocking

packages/pubsub/src/types/Provider.ts
```diff
@@ -4,26 +4,33 @@ import Observable from 'zen-observable-ts';
...
+		topics: string[] | string,
+		msg: Record<string, unknown> | string,
+		options?: ProviderOptions
+	): void;
 
 	subscribe(
 		topics: string[] | string,
 		options?: ProviderOptions
-	): Observable<any>;
+	): Observable<Record<string, unknown>>;
```
elorzafe, line 35: Is this always the case, can be returned something different? like a `string`?

packages/pubsub/tsconfig.build.json
```diff
@@ -3,6 +3,5 @@
 	"compilerOptions": {
 		"allowJs": true
 	},
-	"exclude": ["src/**/*.js"],
```
elorzafe, line 6: dont you need this for the `paho-mqtt` vendor?

packages/pubsub/src/PubSub.ts
```diff
@@ -153,8 +155,10 @@ export class PubSubClass {
 			const subscriptions = observables.map(({ provider, observable }) =>
 				observable.subscribe({
 					start: console.error,
-					next: value => observer.next({ provider, value }),
-					error: error => observer.error({ provider, error }),
+					next: (value: Record<string, unknown>) =>
+						observer.next({ provider, value }),
+					error: (error: Record<string, unknown>) =>
```
elorzafe, line 160: I am not sure about this, maybe this to change as well 

https://github.com/aws-amplify/amplify-js/blob/main/packages/pubsub/src/Providers/MqttOverWSProvider.ts#L213

packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -1,11 +1,12 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import * as Paho from '../vendor/paho-mqtt';
+import { Client as PahoClient } from 'paho-mqtt';
 import { v4 as uuid } from 'uuid';
 import Observable, { ZenObservable } from 'zen-observable-ts';
```
elorzafe, line 6: Can we get rid of this `zen-observable-ts` dependency?

I remember having bundle size issues with this

elorzafe (commented):
Thanks @stocaaro take a look on the comments we can chat more offline if you want

packages/pubsub/tsconfig.build.json
stocaaro: You would think, but even with this, that js file is still included in the ts coverage calculation. It also has no impact on the build to remove it.

packages/pubsub/src/types/Provider.ts
stocaaro: In all cases, the observable emits objects of one type or another.  I had originally normalized these to `object` which Chris had me refactor as `Record<string, unknown>` as this is more idiomatic with TS since it will encourage the user to clarify the `unknown` values received back.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -69,11 +73,29 @@ const customDomainPath = '/realtime';
...
+		errors?: [{ errorType: string; errorCode: number }];
+	};
+};
+
 export interface AWSAppSyncRealTimeProviderOptions extends ProviderOptions {
 	appSyncGraphqlEndpoint?: string;
 	authenticationType?: GraphqlAuthModes;
 	query?: string;
-	variables?: object;
+	variables?: Record<string, unknown>;
```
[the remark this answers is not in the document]
stocaaro: Agreed that this needs review from the data team.

packages/pubsub/src/Providers/MqttOverWSProvider.ts
```diff
@@ -1,11 +1,12 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import * as Paho from '../vendor/paho-mqtt';
+import { Client as PahoClient } from 'paho-mqtt';
```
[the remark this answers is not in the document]
stocaaro: Correct. Types only. Dev dependency. I originally had inlined this interface and adopted this package approach for typing paho at Jim and Chris's recommendations

packages/pubsub/src/Providers/MqttOverWSProvider.ts
stocaaro: Took a look at removing `zen-observable-ts`, which should be replaceable with [@types/zen-observable](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/zen-observable). It didn't go well. VSCode is happy, but the build fails with errors like `zen observable _1.default is not a constructor`. There's got to be a work around, but I didn't find it in my 30 minute time box, so I'm not including any related changes in this update.

packages/pubsub/src/PubSub.ts
stocaaro: You're right. Also, several of these comments circle around the same topic. I find myself convinced that:
- `next` values can be `string`s as well as `Record`s/`object`s
- The category subscription returns `{provider: x, value: y}` instead of the `string`/`Record` itself, so that can be more clearly typed.
- The `publish` and subscribe `value` types should be represented by a shared type instead of being repeated everywhere.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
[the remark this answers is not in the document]
stocaaro: Regarding data team review, I've been watching the `api-graphql` category for issues. The subscription interface is translated to [Observable<any>](https://github.com/aws-amplify/amplify-js/blob/main/packages/api-graphql/src/GraphQLAPI.ts#L394) on output with strongly typed inputs, so I don't expect this set of type changes to cause any impact to api/datastore users.

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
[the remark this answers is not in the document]
elorzafe: This is used also for mock use case IIRC. That is a CLI feature that allows you to test the API without doing `amplify push`

I think it makes sense to not expose that unless is generated automatically by CLI

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -270,7 +303,7 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 		subscriptionId,
 	}: {
 		options: AWSAppSyncRealTimeProviderOptions;
-		observer: ZenObservable.SubscriptionObserver<any>;
+		observer: PubSubContentObserver;
```
elorzafe, line 306: ❤️

packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -370,9 +403,13 @@ export class AWSAppSyncRealTimeProvider extends AbstractPubSubProvider {
 	}
 
 	// Log logic for start subscription failures
-	private _logStartSubscriptionError(subscriptionId, observer, err) {
+	private _logStartSubscriptionError(
+		subscriptionId: string,
+		observer: ZenObservable.SubscriptionObserver<any>,
```
elorzafe, line 408: Can this also use an internal type?

packages/pubsub/src/Providers/MqttOverWSProvider.ts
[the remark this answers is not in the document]
elorzafe: Cool can we do  `import type` to make it explicit, I got a little scared about that import statement

packages/pubsub/src/types/PubSub.ts
```diff
@@ -1,5 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+import { ZenObservable } from 'zen-observable-ts';
```
elorzafe, line 3: Can this be imported as type, or import `SubscriptionObserver`?

elorzafe (dismissed):
I only have minor non blocking comments, I am good to merge this after checking usage numbers of the old AppSync provider

packages/pubsub/src/types/PubSub.ts
stocaaro: `SubscriptionObserver` isn't exported directly.

The d.ts file looks like this:
```ts
export declare namespace ZenObservable {
    interface SubscriptionObserver<T> {
        closed: boolean;
        next(value: T): void;
        error(errorValue: any): void;
        complete(): void;
    }
...
}
```

Am I misunderstanding what you're asking for here? What would I import to do this?

packages/pubsub/src/Providers/MqttOverWSProvider.ts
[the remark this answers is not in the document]
elorzafe: Yes I would use that

stocaaro:
On request, I have run [the integ tests](https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/18235/workflows/8533c67d-34dd-437d-9000-c3aa5724ed6d) against this change, which prompted working out some [Angular build issues](https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/18229/workflows/93fcdb1d-aec2-4e79-852e-b351e6c98fcd).

svidgen (approved):
It's hard to validate with my eyeballs, but from a DataStore/API perspective, it looks like most of the changes here should be non-impactful. If we've got passing integ tests, I think that's good for me.

elorzafe (approved):

packages/pubsub/src/types/PubSub.ts
iartemiev: I think Francisco is referring to: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

stocaaro: Have discussed with Francisco. We'll be looking at retooling our `zen-observable` usage for v6. I did some exploration here to try to excise `zen-observable-ts`, however the `@types/zen-observable` package is missing some of the features `zen-observable` patches onto `global`. Would be nice to have a ts first observable lib that is bundle-size efficient.

WolfWalter:
This PR changed the PubSub message format in an unexpected way:
https://github.com/aws-amplify/amplify-js/commit/5147c5ce555d042722e2888fc423430f517b91b7#diff-2a29f8a41f2b66733bb64a134e53acb5e67615e026667885a5686ce933d3869eL267-R300

`topicSymbol` is now an object key and the Symbol is not used anymore. This totally changed how to get the `topic` and the message object gets altered in an unexpected way.

Was that really intentional @stocaaro and @elorzafe?

PR #10968, opened 2023-02-12
Call errorCallback on error in _initializeUploadTask

Pickleboyonline:
#### Description of changes

Storage.put will not call the error callback when an error is thrown in the function `_initializeUploadTask` is thrown. This PR will emit an error in the catch block.

The documentation needs to change for page describing how to use Storage with an non-Amplify S3 bucket. The authed users should also have permissions to use the ListPartsCommand, AbortMultipartUploadCommand, and ListObjectsV2Command. ListPartsCommand allows the upload to resume; AbortMultipartUploadCommand allows the upload to cancel; and ListObjectsV2Command allows Storage.put to verify the remote file size but is nonblocking.

Many developers may be facing this issue but are unaware because the `errorCallback` is not called.

- Initiate upload (with resumable option on)
- Stop upload from completing
- Restart app
- Attempt same upload again but do not allow the Auth user access to the ListPartsCommand
- A 403 error is returned in the errorCallback

One could also replicate this with a network error when initiating an upload.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

@joebernard


AllanZhengYP (approved):
Looks good to me. We already emit error event in other cases of S3 exceptions. Emitting extra error event here should be a low risk.

joesilveira:
@AllanZhengYP When could we expect to see this merged with the main branch? This is extremely useful and I would argue that it is a necessity when using Storage.put with resumable:true and the 3 callbacks. Without this, the errorCallback is never actually called and thus the error object is never handed back to the user but rather only logged and so we can never actually display the error to the user. For instance, an error containing "access denied" is crucial in letting the user know such error

stocaaro (approved):

PR #10979, opened 2023-02-16
feat(notifications): Add getPermissionStatus API

cshfang:
#### Description of changes
This PR adds the `getPermissionStatus` API which is the companion API to `requestPermissions`.

Previously, the latter API directly returned either a granted or denied status. But this failed to enable use cases where developers may want to display rationale UIs before attempting to request permissions. The `getPermissionStatus` API will give them additional insight into the current permission state and enable them to better control their permission workflow.

Tested locally against React Native Android app on API 33

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/notifications/src/PushNotification/PushNotification.native.ts
+++ b/packages/notifications/src/PushNotification/PushNotification.native.ts
@@ -274,18 +274,18 @@ export default class PushNotification implements PushNotificationInterface {
 		return normalizeNativeMessage(await getLaunchNotification?.());
 	};
 
+	getPermissionStatus = async (): Promise<PushNotificationPermissionStatus> => {
+		const { getPermissionStatus } = this.nativeModule;
+		return normalizeNativePermissionStatus(await getPermissionStatus?.());
+	};
+
 	requestPermissions = async (
 		permissions: PushNotificationPermissions = {
 			alert: true,
 			badge: true,
 			sound: true,
 		}
-	): Promise<PushNotificationPermissionStatus> => {
-		const { requestPermissions } = this.nativeModule;
-		return normalizeNativePermissionStatus(
-			await requestPermissions?.(permissions)
-		);
-	};
+	): Promise<boolean> => this.nativeModule.requestPermissions?.(permissions);
 
 	/**
 	 * Background notifications on will start the app (as a headless JS instance running on a background service on
--- a/packages/notifications/src/PushNotification/PushNotification.native.ts
+++ b/packages/notifications/src/PushNotification/PushNotification.native.ts
@@ -132,8 +132,9 @@ export default class PushNotification implements PushNotificationInterface {
 							logger.error(err);
 						} finally {
 							// notify native module that handlers have completed their work (or timed out)
-							const { completeNotification } = this.nativeModule;
-							completeNotification?.(message.completionHandlerId);
+							this.nativeModule.completeNotification?.(
+								message.completionHandlerId
+							);
 						}
 					}
 				);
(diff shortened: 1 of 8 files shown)


helgabalashova (approved):

packages/notifications/src/PushNotification/PushNotification.native.ts
```diff
@@ -274,18 +274,18 @@ export default class PushNotification implements PushNotificationInterface {
 		return normalizeNativeMessage(await getLaunchNotification?.());
 	};
 
+	getPermissionStatus = async (): Promise<PushNotificationPermissionStatus> => {
+		const { getPermissionStatus } = this.nativeModule;
+		return normalizeNativePermissionStatus(await getPermissionStatus?.());
```
israx, line 279: what happens if `getPermissionStatus` is not assigned ?  I'm wondering if there's the need to add some assertions.

packages/notifications/src/PushNotification/types.ts
```diff
@@ -86,7 +87,8 @@ export interface PushNotificationPermissions extends Record<string, boolean> {
 export enum PushNotificationPermissionStatus {
```
israx, line 87: is this enum used internally only? If so, we could consider using a `const` enum , just for bundle size purposes

packages/notifications/src/PushNotification/utils.ts
```diff
@@ -18,8 +18,10 @@ export const normalizeNativePermissionStatus = (
 			return PushNotificationPermissionStatus.DENIED;
 		case 'Granted':
 			return PushNotificationPermissionStatus.GRANTED;
-		default:
-			return PushNotificationPermissionStatus.UNDETERMINED;
+		case 'NotRequested':
+			return PushNotificationPermissionStatus.NOT_REQUESTED;
+		case 'ShouldRequestWithRationale':
+			return PushNotificationPermissionStatus.SHOULD_REQUEST_WITH_RATIONALE;
 	}
```
israx, line 25: I see that the function `normalizeNativePermissionStatus` takes a `nativeStatus` parameter, is it possible to add a type to it ? So that way we can add a `default` case and assign a variable with the type `never`. This allows to make sure that whenever we add a new case we handle it on the switch statement. 

Something like:

```
default:
   const status:never = nativeStatus
```

jimblanc (approved):
Looks good, minor comments

packages/notifications/src/PushNotification/PushNotification.native.ts
```diff
@@ -274,18 +274,18 @@ export default class PushNotification implements PushNotificationInterface {
 		return normalizeNativeMessage(await getLaunchNotification?.());
 	};
 
+	getPermissionStatus = async (): Promise<PushNotificationPermissionStatus> => {
+		const { getPermissionStatus } = this.nativeModule;
```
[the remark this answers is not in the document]
cshfang: That's a good point. I tried some patterns out and I think just not destructuring it here will provide the most clarity without being excessively verbose.

packages/notifications/src/PushNotification/PushNotification.native.ts
cshfang: Nothing would happen. But the customer's app would have failed long before we got to that point. This is more of a general pattern for APIs that might be supported only in one platform. A couple of which is yet to be added with iOS.

packages/notifications/src/PushNotification/types.ts
cshfang: This should be exported. I'll do this in another PR to keep the intention of this PR clear.

packages/notifications/src/PushNotification/utils.ts
```diff
@@ -18,8 +18,10 @@ export const normalizeNativePermissionStatus = (
 			return PushNotificationPermissionStatus.DENIED;
 		case 'Granted':
 			return PushNotificationPermissionStatus.GRANTED;
-		default:
```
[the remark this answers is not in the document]
cshfang: `NOT_REQUESTED` is a state that's only applicable to Android. I thought about which one would make the most sense as a default, but they really are discreet statuses that should be guaranteed to describe a particular state.

packages/notifications/src/PushNotification/utils.ts
cshfang: Normally, I'd be all for stricter typing but, currently, without the codegen from RN new architecture, trying to type things coming from the native module while also trying to corral the platform differences is just not super practical. That exhaustive check is pretty cool. I've never seen it before but there should be lint options that we can enable that doesn't require an used variable 🤔

packages/notifications/src/PushNotification/PushNotification.native.ts
israx: got it

israx (approved):

Issue #10750, opened 2022-12-07
Can't remove hasOne or belongsTo relationship


hein-j:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

DataStore

### Amplify Categories

Not applicable

### Environment information

<details>

```
  System:
    OS: Linux 5.4 Amazon Linux 2
    CPU: (16) x64 Intel(R) Xeon(R) Platinum 8175M CPU @ 2.50GHz
    Memory: 55.50 GB / 61.37 GB
    Container: Yes
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 17.9.0 - ~/.nvm/versions/node/v17.9.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v17.9.0/bin/yarn
    npm: 8.5.5 - ~/.nvm/versions/node/v17.9.0/bin/npm
  Browsers:
    Firefox: 91.10.0esr
  npmPackages:
    @aws-amplify/codegen-ui: 2.5.6 => 2.5.6 
    @aws-amplify/codegen-ui-react: 2.5.6 => 2.5.6 
    @aws-amplify/codegen-ui-test-generator: 2.5.6 => 2.5.6 
    @aws-amplify/datastore: 4.0.5 => 4.0.5 
    @aws-amplify/ui-react: 4.1.0 => 4.1.0 
    @aws-amplify/ui-react-internal:  undefined ()
    @cypress/angular:  0.0.0-development 
    @cypress/mount-utils:  0.0.0-development 
    @cypress/react:  0.0.0-development 
    @cypress/react18:  0.0.0-development 
    @cypress/svelte:  0.0.0-development 
    @cypress/vue:  0.0.0-development 
    @cypress/vue2:  0.0.0-development 
    @testing-library/jest-dom: ^5.16.5 => 5.16.5 
    @testing-library/react: ^13.4.0 => 13.4.0 
    @testing-library/user-event: ^13.5.0 => 13.5.0 
[...]

PR #10983, opened 2023-02-16
fix(datastore): remove connections with model field as undefined

dpilch:
#### Description of changes

Allows removal of connection using the model field instead of the join fields.

```javascript
await DataStore.save(Parent.copyOf(parentWithChild, draft => {
  draft.child = undefined;
}));
```

Example:

Given the following schema:

```graphql
type Parent @model {
  id: ID!
  child: Child @hasOne
}

type Child @model {
  id: ID!
  parent: Parent @belongsTo
}
```

In order to remove a child from the parent the join field (`parentChildId`) must be set to undefined rather than the direct model field (`child`).

```javascript
const parent = new Parent({});
const child = new Child({});
const savedParent = await DataStore.save(parent);
const savedChild = await DataStore.save(child);

const parentWithChild = await DataStore.save(Parent.copyOf(savedParent, draft => {
  draft.child = savedChild;
}));
```

```javascript
await DataStore.save(Parent.copyOf(parentWithChild, draft => {
  draft.child = undefined; // does not remove the connection
}));
```

```javascript
await DataStore.save(Parent.copyOf(parentWithChild, draft => {
  draft.parentChildId = undefined; // removes the connection
}));
```

With the change `draft.child = undefined` will now also remove the connection.

#### Issue #, if available

Resolves https://github.com/aws-amplify/amplify-js/issues/10750

Tested with minimum repro above.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/__tests__/DataStore.ts
```diff
@@ -1733,6 +1733,41 @@ describe('Model behavior', () => {
...
+
+		const child = await DataStore.save(new HasOneChild({}));
+		const parent = await DataStore.save(
+			new HasOneParent({
+				child,
+			})
+		);
+
+		const parentWithoutChild = HasOneParent.copyOf(parent, draft => {
+			draft.child = undefined;
```
svidgen, line 1747: Do we already support `draft.child = null`? If not, can we?

dpilch: It doesn't pass type checking.

```
Type 'null' is not assignable to type 'LazyChild | undefined'.
```

One solution would be to change the generated types to:
```
readonly child: AsyncItem<Child | undefined | null>;
```

I'll see if there is something that can be done on the library side.

svidgen: Huh ... 🤔 ... That typing _might_ actually be incorrect. But, we should follow up on that independently. For this PR, can we define the behavior anyway? JS customers will be able to pass `null`, and I think we'll want it to have the same effect.

svidgen: FWIW, I actually thought we were mapping `Promise<Model | undefined>` to `Model | null | undefined` in `ModelInit<T>` or some util type downstream. It might be worth digging around to see if that mapping is incorrect.

dpilch: I made a change to deep writable to add `null` to the type union that is resolving the type error: https://github.com/aws-amplify/amplify-js/pull/10983/commits/0c866df55bc23c9b21cdabf07ba162641d9b1ed2

svidgen: Sounds good. I'll need a little time to review this one. The type changes feel like they warrant a bit more focus and manual in-IDE testing time.

packages/datastore/__tests__/DataStore.ts
```diff
@@ -1733,6 +1733,76 @@ describe('Model behavior', () => {
...
+				posts: [post],
+			})
+		);
+
+		const blogWithoutPosts = Blog.copyOf(blog, draft => {
+			draft.posts = null;
+		});
+
+		expect(await blogWithoutPosts.posts.toArray()).toEqual([]);
+	});
```
svidgen, line 1804: This use-case actually shouldn't work to begin with, unless we missed something! (Or if I'm reading it wrong.) The `posts` property might be getting stored on the instance here and then later removed. But, that shouldn't be persisting as a proper relationship.

This may actually signal the need to log a new issue to fix types and validation to prevent customer code from setting child collections like this.

Setting the relationship from the child's side is the supported approach:

```js
const parent = await DataStore.save(new Blog({...}));
const child = await DataStore.save(new Post({
  parent
  // ...
}));
```

And then to dissociate:

```js
const child = await DataStore.save(Post.copyOf(post, draft => 
  draft.parent = null
));
```

Can you dive and/or pair with me on this a bit to confirm?

dpilch: I somehow always seem to forget this. 🤦  I think a type and validation change would be a good idea. I'll create an issue.

packages/datastore/__tests__/DataStore.ts
```diff
@@ -1733,6 +1733,80 @@ describe('Model behavior', () => {
...
+		);
+
+		const childWithoutParent = CompositePKChild.copyOf(child, draft => {
+			draft.parent = undefined;
+		});
+
+		expect(await childWithoutParent.parent).toBeUndefined();
+	});
+
+	test('model field can be set to undefined to remove connection on child hasMany', async () => {
```
svidgen, line 1790:
```suggestion
	test('model field can be set to null to remove connection on child hasMany', async () => {
```

packages/datastore/__tests__/DataStore.ts
```diff
@@ -1733,6 +1733,80 @@ describe('Model behavior', () => {
...
+		);
+		const child = await DataStore.save(
+			new CompositePKChild({ childId: 'childId', content: 'content', parent })
+		);
+
+		const childWithoutParent = CompositePKChild.copyOf(child, draft => {
+			draft.parent = null;
+		});
+
+		expect(await childWithoutParent.parent).toBeUndefined();
```
svidgen, line 1807: Sorry I didn't think of all the feedbacks in one round. 😓 But, can we extend all of these tests a little to ensure parent nullification is saved correctly? I.e.,

1. Save the copied child without parent.
2. Retrieve the copied child.
3. Ensure the nullified parent is persisted correctly.

I'm also not clear on whether we need to confirm the round-trip to cloud storage successfully persists the removal there. Are we sure the right mutations are being generated? Have you tried it manually?

svidgen (commented):
Sorry I couldn't consolidate my feedback into fewer rounds on this. Slow brain days. 🙁

packages/datastore/__tests__/DataStore.ts
dpilch: I updated the tests to include assertions on the parent was well. I have tested manually using the original reproduction steps and it is persisting in cloud. Do you think we should add an E2E test for this as well?

svidgen: > Do you think we should add an E2E test for this as well?

Yes. It would also be nice to clamp the e2e behavior on this against our gql fake in the no-longer-accurately-named [`connectivityHandling` tests](https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/__tests__/connectivityHandling.test.ts).

Both of those are nonblocking IMO. Get the fix out sooner rather than later. But, one and/or both of those should be fast-followups, IMO, to guard against regression.

svidgen (approved):
Thanks for the iterations!

david-mcafee (approved):
LGTM! 🎉

PR #10996, opened 2023-02-18
chore(notifications): Export public types

cshfang:
#### Description of changes
This PR exports some Push Notification types intended for public use. It also updates an internal enum to be const.

`$ yarn test`
Tested locally that changes did not cause regression in event listeners

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/notifications/src/PushNotification/types.ts
```diff
@@ -98,11 +98,11 @@ export type OnPushNotificationMessageHandler = (
...
-	FOREGROUND_MESSAGE_RECEIVED = 'FOREGROUND_MESSAGE_RECEIVED',
-	LAUNCH_NOTIFICATION_OPENED = 'LAUNCH_NOTIFICATION_OPENED',
-	NOTIFICATION_OPENED = 'NOTIFICATION_OPENED',
-	TOKEN_RECEIVED = 'TOKEN_RECEIVED',
+	BACKGROUND_MESSAGE_RECEIVED,
+	FOREGROUND_MESSAGE_RECEIVED,
+	LAUNCH_NOTIFICATION_OPENED,
+	NOTIFICATION_OPENED,
+	TOKEN_RECEIVED,
 }
```
stocaaro, line 106: This seems like a really bad idea. Enums are numbered when they aren't given string equivalents. This can be okay when the type is internal only, but can cause lots of issues across an export boundary.

stocaaro (changes requested):

packages/notifications/src/PushNotification/types.ts
cshfang: They are exported only for internal use though. I specifically didn't include it in the index exports

stocaaro (approved):

Issue #7534, opened 2021-01-10
Getting records more than maxRecordsToSync (example: older records in case records are sorted)


mir1198yusuf:
**Is your feature request related to a problem? Please describe.**
Hello, considering below syncExpression

```
syncExpression(Events, (c) =>
  c.meetingId('eq', 'some-value').createdAt('gt', '2020-11-01')
)
```
Suppose i have a list displaying Events , by above expression only events created after 2020-11-01 will be stored in client-side database.

Now if user keeps scrolling up and suddenly wants to see an event records before 2020-11-01 or the very first oldest event, that data would not be available in client-side.

So how to inform Datastore to sync these older records with backend and display it, because `Datastore.query` returns only data from client-side .

Also `maxRecordsToSync` would not help as we dont know in advance the number of older records.

**Describe the solution you'd like**
Something similar to below:
when user scrolls to the end of list, call `Datastore.queryMore(Events, newLimit)`
where `newLimit` is the number of records to sync more from backend.
Since `Datastore.observe` listens to changes in backend database, a similar method `Datastore.observeLocal` to listen to changes in local database (caused by syncing more `newLimit` number of records).

`newLimit` will increase the `maxRecordsToSync` as   `maxRecordsToSync + newLimit` for either all tables or specified table `Events`.

**Describe alternatives you've considered**
I thought of a solution like when user scroll the end of list then execute `Datastore.stop`, change the value of `maxRecordsToSync` in `Datastore.config` which is usually in some starting file closer to`index.js` ( using some hooks in case of React) and then again `Datastore.start` and finally `Datastore.query`.
But this seems too complicated. I have not tried this though

iartemiev:
@mir1198yusuf you would need to pass a variable instead of a hardcoded value to the predicate in the syncExpression and have the value of the variable change at runtime. See the [Reevaluate expressions at runtime](https://docs.amplify.aws/lib/datastore/sync/q/platform/js#reevaluate-expressions-at-runtime) section of the docs.

In your case, it would look something like this:
```js
let fromDate = '2020-11-01';

DataStore.configure({
  syncExpressions: [
    syncExpression(Events, (c) =>
      c.meetingId('eq', 'some-value').createdAt('gt', fromDate)
    )
  ]
});

async function change meetingSync() {
  fromDate = '2019-11-01';
  await DataStore.stop();
  await DataStore.start();
}
```

mir1198yusuf:
[...]

PR #11000, opened 2023-02-20
feat(datastore): add support for AppSync RTF

iartemiev:
#### Description of changes

Add support for AppSync [Enhanced Subscription Filtering](https://docs.aws.amazon.com/appsync/latest/devguide/aws-appsync-real-time-enhanced-filtering.html) (aka Runtime Filtering / RTF)

Adds compatibility with dynamic auth on list fields, e.g. "list of owners", "list of groups".

Allows DataStore [selective sync](https://docs.amplify.aws/lib/datastore/sync/q/platform/js/#selectively-syncing-a-subset-of-your-data) predicates to be applied as a backend subscription filter, instead of receiving all authorized subscription messages and filtering them clientside as was done before this PR. 

RTF has several service constraints (listed [here](https://docs.aws.amazon.com/appsync/latest/devguide/extensions.html#extensions-exceptions)), therefore the current clientside filtering behavior will remain as a fallback in case the service rejects the provided filter expression.

In order to keep this change non-breaking, if DataStore receives an RTF-specific error message from the service at connect time, it will retry the subscription without a filter argument, falling back to filtering the subscription messages in the client. 

**Additional fix**
The [selective sync -> DDB Query](https://docs.amplify.aws/lib/datastore/sync/q/platform/js/#advanced-use-case---query-instead-of-scan) use case is currently broken in main, because the filter expression we send as a sync query param has a bunch of redundant `and` groups nested and the generated sync query request resolver isn't able to match the filter expression to a query expression. This PR flattens the resulting GQL filter to restore this use case ([see predicateToGraphqlFilter](https://github.com/aws-amplify/amplify-js/pull/11000/files#diff-51043f4ed19a784f212d1d628e3239e8406cb3c942d2dd3f0ea739f681521541R507-R517))

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/7534

* Manual testing
* Unit tests
* e2e test run - https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/18267/workflows/06be232a-5be6-40b4-86a6-cfbd41895e05

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [] Relevant documentation is changed or added (and PR referenced)


packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -478,6 +526,44 @@ class SubscriptionProcessor {
 															},
 														} = subscriptionError;
 
+														// Catch RTF errors
+														if (
+															message.includes(
```
dpilch, line 531: Is there anyway we can make this more robust incase the error messages change in the future?

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -660,6 +747,117 @@ class SubscriptionProcessor {
...
+
+		return true;
+	}
+
+	private warningMessage(
+		modelDefinition: SchemaModel,
+		predicatesGroup: PredicatesGroup<any> | undefined
+	): string | null {
+		const HEADER =
+			'Backend subscriptions filtering limit exceeded.\n' +
```
dpilch, line 791: This wouldn't always be accurate right? In the case that a a not group is used the filtering is not disable due to a limit (unless you are using limit here as in limitation)?

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -660,6 +747,117 @@ class SubscriptionProcessor {
...
+			);
+		}
+
+		if (totalFiltersFields > FILTER_FIELD_SERVICE_LIMIT) {
+			let warningContent =
+				`${HEADER}\n` +
+				`Your selective sync expression for ${modelDefinition.name} contains ${filterFields} different model fields.\n\n` +
+				`In order for serverside subscription filtering to succeed, you can apply a selective sync expression with at most ${
+					FILTER_FIELD_SERVICE_LIMIT - dynamicAuthModes
+				} different fields for this model.`;
```
dpilch, line 827:
```suggestion
				`Serverside subscription filtering supports a selective sync expression with a maximum of ${
					FILTER_FIELD_SERVICE_LIMIT - dynamicAuthModes
				} different fields for this model.`;
```

packages/datastore/tslint.json
```diff
@@ -5,7 +5,6 @@
 	"jsRules": {},
 	"rules": {
 		"prefer-const": true,
-		"max-line-length": [true, 120],
```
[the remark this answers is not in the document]
iartemiev: The line limit is automatically enforced by Prettier whenever possible, i.e. for code that can be broken up into multiple lines. But this lint rule also throws an error for strings that exceed 120 - like the new warnings that get emitted as part of this PR. Felt like an unnecessary constraint.

packages/datastore/src/sync/processors/subscription.ts
iartemiev: There's no way I'm aware of other than string matching. The error message comes from the service, so it's unlikely to change on a whim.

packages/datastore/src/sync/processors/subscription.ts
iartemiev: That sounds a lot better. Thanks!

undefobj:
Limits on clients have proven to be an anti-pattern in the past that we avoid. They should be set in the services and raised when necessary, either across all APIs or per-customer. Unless there's a safety reason you're trying to address here suggest removing any limits from clients and handling the service throttling/error message as appropriate. If customers bump up against these limits they can submit feedback to AppSync to have them raised.

iartemiev:
FYI - tried to use "hide as resolved" for Richard's comment above, because I've updated the PR with the change he requested, but GH is just showing it as "hidden".

svidgen (commented):
Might need another offline sync on this. I think I understood the balance we wanted to strike between client-side helpfulness + validation differently.

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -660,6 +733,137 @@ class SubscriptionProcessor {
...
+				`Your API was generated with an older version of the CLI that doesn't support backend subscription filtering.` +
+				'To enable backend subscription filtering: upgrade your Amplify CLI to the latest version and run `amplify push`';
+		} else if (message.includes('Filters exceed maximum attributes limit')) {
+			const selSyncFields = filterFields(predicatesGroup);
+			const selSyncFieldStr = [...selSyncFields].reduce((sum, cur, idx) => {
+				if (idx === 0) {
+					return cur;
+				}
+				return `${sum}, ${cur}`;
+			}, '');
```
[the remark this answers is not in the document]
iartemiev: Yep, good call. Updated this and the other 2 instances.

packages/datastore/src/sync/utils.ts
```diff
@@ -522,6 +579,177 @@ export function predicateToGraphQLFilter(
...
+ *
+ * @example returns "username"
+ * ```js
+ * { type: "and", predicates: [
+ * 		{ field: "username", operator: "beginsWith", operand: "a" },
+ * 		{ field: "username", operator: "contains", operand: "abc" },
+ * ] }
+ * ```
+ */
+export function repeatedFieldInGroup(
```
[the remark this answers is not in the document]
iartemiev: We can throw this into the catch as well, it'll return a unique error message. But my thought process for doing it ahead of time is that this is an inherent constraint in the RTF implementation and not something that can change on a case-by-case basis

packages/datastore/src/sync/utils.ts
```diff
@@ -522,6 +579,211 @@ export function predicateToGraphQLFilter(
...
+			);
+
+		case RTFError.MaxAttributes: {
+			let message = `Your selective sync expression for ${modelDefinition.name} contains ${selSyncFields.size} different model fields: ${selSyncFieldStr}.\n\n`;
+
+			if (dynamicAuthModeFields.size > 0) {
+				message +=
+					`Note: the number of fields you can use with selective sync is affected by @auth rules configured on the model.\n\n` +
+					`Dynamic auth modes, such as owner auth and dynamic group auth each utilize 1 field.\n` +
+					`You c	urrently have ${dynamicAuthModeFields.size} dynamic auth mode(s) configured on this model: ${dynamicAuthFieldsStr}.`;
```
svidgen, line 759:
```suggestion
					`You currently have ${dynamicAuthModeFields.size} dynamic auth mode(s) configured on this model: ${dynamicAuthFieldsStr}.`;
```

packages/datastore/src/sync/utils.ts
```diff
@@ -522,6 +579,211 @@ export function predicateToGraphQLFilter(
...
+			}
+			return message;
+		}
+
+		case RTFError.RepeatedFieldname:
+			return `Your selective sync expression for ${modelDefinition.name} contains multiple entries for ${repeatedField} in the same AND group.`;
+		case RTFError.NotGroup:
+			return (
+				`Your selective sync expression for ${modelDefinition.name} uses a \`not\` group. If you'd like to filter subscriptions in the backend, ` +
+				`please re-write your expression using \`ne\` or \`notContains\` operators.`
```
svidgen, line 782:
```suggestion
				`rewrite your expression using \`ne\` or \`notContains\` operators.`
```

packages/datastore/src/sync/utils.ts
```diff
@@ -522,6 +579,211 @@ export function predicateToGraphQLFilter(
...
+					`Note: the number of fields you can use with selective sync is affected by @auth rules configured on the model.\n\n` +
+					`Dynamic auth modes, such as owner auth and dynamic group auth factor in to the number of combinations you're using.\n` +
+					`You currently have ${dynamicAuthModeFields.size} dynamic auth mode(s) configured on this model: ${dynamicAuthFieldsStr}.`;
+			}
+			return message;
+		}
+
+		case RTFError.RepeatedFieldname:
+			return `Your selective sync expression for ${modelDefinition.name} contains multiple entries for ${repeatedField} in the same AND group.`;
+		case RTFError.NotGroup:
```
svidgen, line 779: If this is a forever rule, it might be nice down the road to automagically rewrite `not`'s out of queries for customers. (In cases where `not` doesn't turn into two conditions on the same field, I guess. 😅 )

packages/datastore/src/sync/utils.ts
```diff
@@ -522,6 +579,211 @@ export function predicateToGraphQLFilter(
...
+	const dynamicAuthModeFields = dynamicAuthFields(modelDefinition);
+	const dynamicAuthFieldsStr = [...dynamicAuthModeFields].join(', ');
+	const filterCombinations = countFilterCombinations(predicatesGroup);
+	const repeatedField = repeatedFieldInGroup(predicatesGroup);
+
+	switch (errorType) {
+		case RTFError.UnknownField:
+			return (
+				`Your API was generated with an older version of the CLI that doesn't support backend subscription filtering.` +
+				'To enable backend subscription filtering: upgrade your Amplify CLI to the latest version and run `amplify push`'
```
svidgen, line 749: Might be good to give them both commands they need to run.

```suggestion
				'To enable backend subscription filtering, upgrade your Amplify CLI to the latest version and push your app by running `amplify upgrade` followed by `amplify push`'.
```

packages/datastore/src/sync/utils.ts
```diff
@@ -522,6 +579,211 @@ export function predicateToGraphQLFilter(
...
+		} else if (rule.ownerField) {
+			fields.add(rule.ownerField);
+		}
+	}
+
+	return fields;
+}
+
+/**
+ *
```
svidgen, line 629: Any synopsis and/or examples to put here?

svidgen (changes requested):
Sorry, I guess I lied about being just about ready to approve. A few suggestions/requests. I think the only blocker for me is the typo.

packages/datastore/src/sync/utils.ts
iartemiev: 👍🏻 will add a backlog task for this

packages/datastore/src/sync/utils.ts
iartemiev: Added an example

svidgen (dismissed):

packages/datastore/src/sync/utils.ts
```diff
@@ -505,12 +543,31 @@ export function predicateToGraphQLFilter(
...
 
-			appendToFilter(gqlField);
+			children.push(gqlField);
 			return;
 		}
 
-		const child = predicateToGraphQLFilter(predicate, fieldsToOmit);
-		Object.keys(child).length > 0 && appendToFilter(child);
+		const child = predicateToGraphQLFilter(predicate, fieldsToOmit, false);
+		Object.keys(child).length > 0 && children.push(child);
```
manueliglesias, line 551: from https://eslint.org/docs/latest/rules/no-unused-expressions:
> An unused expression which has no effect on the state of the program indicates a logic error.

I know it is not a logic error here, and we used this pattern in other places, but it is clearer to just do the `if`

packages/datastore/src/sync/utils.ts
```diff
@@ -505,12 +543,31 @@ export function predicateToGraphQLFilter(
...
+	});
+
+	// flatten redundant list predicates
+	if (children.length === 1) {
+		const [child] = children;
+		if (
+			// any nested list node
+			(isList && !root) ||
+			// root list node where the only child is also a list node
+			(isList && root && ('and' in child || 'or' in child))
```
manueliglesias, line 561: question: do we have constants we can use here for `and` and `or`?

another question, what about double `not`? is that a condition we care about and need to flatten too in this code path?

packages/datastore/src/sync/utils.ts
```diff
@@ -505,12 +543,31 @@ export function predicateToGraphQLFilter(
...
+		) {
+			delete result[type];
+			Object.assign(result, child);
+			return result;
+		}
+	}
+
+	children.forEach(child => {
+		isList ? result[type].push(child) : (result[type] = child);
```
manueliglesias, line 570: Not blocking, see comment above about the`no-unused-expressions` eslint rule. I know we don't enforce that rule, I just think it is better / more readable style to not have unused expressions and be explicit with an if/else if there are side-effects involved. Ternaries re great for assignments though.

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -478,6 +518,33 @@ class SubscriptionProcessor {
...
+															this.catchRTFError(
+																message,
+																modelDefinition,
+																predicatesGroup
+															);
+
+														// Catch RTF errors
+														if (isRTFError) {
+															// Unsubscribe and clear subscription array for model/operation
+															subscriptions[modelDefinition.name][
```
manueliglesias, line 533: We are very deeply nested here, can we simplify this code or extract parts of it to help with readability?

manueliglesias (commented):
Looks great! I have some questions and a single change I would like to see regarding deeply nested code.

packages/datastore/src/sync/utils.ts
```diff
@@ -300,19 +304,39 @@ export function buildSubscriptionGraphQLOperation(
...
 
 	const opName = `on${transformerMutationType}${typeName}`;
+
 	let docArgs = '';
 	let opArgs = '';
 
-	if (isOwnerAuthorization) {
-		docArgs = `($${ownerField}: String!)`;
-		opArgs = `(${ownerField}: $${ownerField})`;
+	if (filterArg || isOwnerAuthorization) {
```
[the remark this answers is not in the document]
iartemiev: Agreed, that looks better. Updated with a minor change to the snippet.

packages/datastore/src/sync/utils.ts
iartemiev: > do we have constants we can use here for and and or?
We don't. They're hardcoded everywhere else as well. We can circle back with a refactor, but I felt that was out of scope for this PR.

> what about double `not`?
It wouldn't hurt to flatten it, but it's not crucial. Flattening redundant `and` groups is necessary for the selective sync -> index query use case, but flattening double `not`s would be more of an edge case stylistic improvement

packages/datastore/src/sync/processors/subscription.ts
iartemiev: It's at the same level of nesting as the rest of the code in that catch block (e.g., the multi-auth handling). I agree that the SubscriptionProcessor could use a refactor, but I think we should tackle that as a separate task.

svidgen: We have complexity issues across the board. IMHO, I think this PR is actually *less* complex to review by following suit here. A separate PR or two for refactors would be welcome. But, in the worst-case, we should discuss this + refactor opportunities in the biweekly thing -- get some stories on the board that will meaningfully improve the codebase.

svidgen (approved):

manueliglesias (approved):
LGTM 👍

PR #11009, opened 2023-02-21
fix(@aws-amplify/datastore): strictly define null vs undefined behavior on models

svidgen:
#### Resubmitted

Originally submitted as #10915, which caused failing integ tests, due to `owner: null` being submitted in mutations when the the `owner` field was made explicit. AppSync rejects `owner: null` as "Unauthorized."

This updated PR omits the owner field in mutations (or custom owner field per model `attributes.auth`) when the value would be `null` or `undefined`.

Added tests for `owner` handling are in [connectivityHandling.test.ts:102-168](https://github.com/aws-amplify/amplify-js/pull/10932/files#diff-68fad1b028a70ed30e99e616e38d45ec29a2993942226f28ff94cdb241bb3b25R102-R168), with owner field validation occurring in [the graphql fake in helper.ts](https://github.com/aws-amplify/amplify-js/pull/10932/files#diff-d80854c979fb9c08302352685e794dcd2cd3c7a4d29234d90b7757061d14ca5eR779-R781).

The fix is for `owner` handling is in `mutation.ts` on [lines 487-490](https://github.com/aws-amplify/amplify-js/pull/10932/files#diff-caf2490eda4f0fb658881cae74f9e9b5db5e70363b8d0a737bd7bdaca51e053aR487-R490) and [lines 526-529](https://github.com/aws-amplify/amplify-js/pull/10932/files#diff-caf2490eda4f0fb658881cae74f9e9b5db5e70363b8d0a737bd7bdaca51e053aR526-R529).

You can also see the bulk of the `owner` field fix and tests together by looking at the [2nd commit in the PR](https://github.com/aws-amplify/amplify-js/pull/10932/commits/bc756645a40317454cdfa6e2e94c4952a93efe09).

The 2nd attempt under #10932 revealed model instances failing validation during `copyOf` of models explicit timestamp fields. Timestamp fields are never actually required by AppSync, so it's safe from a runtime perspective to omit them even when marked as required in the MIPR. Even though this is a bug in runtime validation, it's an existing bug that we need to consider independently, especially as fixing the validation could actually be construed as a breaking change.

To address this, this PR treats all timestamp fields as optional.

#### Description of changes

Given:

```graphql
type Model @model {
  id: ID!
  value: String!
  otherValue: String
}
```

If we save an instance:

```js
    value: 'some value'
}));
```

We'll have this returned:

```json
{
  "value": "some value",
}
```

```json
{

  // LOOK!
  "otherValue": null, 

  "owner": "test",
  "_version": 1,
  "_deleted": null
}
```

2. Verified in sample apps.
3. Temporarily added integ testing to this branch where [all unit and integ tests pass](https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/18147/workflows/cf9f4de1-d4e7-4745-850b-fb7e50a0bab4).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/src/datastore/datastore.ts
```diff
@@ -169,6 +169,39 @@ const getModelPKFieldName = (
 	);
 };
 
+/**
+ * Determine what the managed timestamp field names are for the given model definition
+ * and return the mapping.
+ *
+ * All timestamp fields are in cluded in the mapping, regardless of whether the final field
```
david-mcafee, line 176: nit:
```suggestion
 * All timestamp fields are included in the mapping, regardless of whether the final field
```

david-mcafee (approved):
Besides a super minor nit on a comment, LGTM! 🎉 As always, really appreciate the in-depth level of testing, as well as the informative comments.

AllanZhengYP (dismissed):

dpilch (approved):

dpilch (dismissed):

david-mcafee (approved):
LGTM! 🎉

manueliglesias (approved):
LGTM 🎉

iartemiev (approved):

david-mcafee (approved):
LGTM!

Issue #11018, opened 2023-02-22
[DataStore] ObserveQuery unexpected behavior when Numeric Index for equality predicate


chrisbonifacio:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

DataStore

### Environment information

<details>

```
  System:
    OS: macOS 13.2
    CPU: (10) arm64 Apple M1 Max
    Memory: 12.36 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.13.0/bin/yarn
    npm: 8.3.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
    Watchman: 2022.12.12.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 110.0.5481.177
    Firefox: 101.0
    Safari: 16.3
  npmPackages:
    @ampproject/toolbox-optimizer:  undefined ()
    @aws-amplify/ui-react: ^4.3.5 => 4.3.5 
    @aws-amplify/ui-react-internal:  undefined ()
    @babel/core:  undefined ()
    @babel/runtime:  7.15.4 
    @edge-runtime/primitives:  2.0.0 
    @hapi/accept:  undefined ()
    @napi-rs/triples:  undefined ()
    @next/font: 13.1.3 => 13.1.3 
    @next/react-dev-overlay:  undefined ()
    @segment/ajv-human-errors:  undefined ()
    @types/node: 18.11.18 => 18.11.18 
    @types/react: 18.0.27 => 18.0.27 
    @types/react-dom: 18.0.10 => 18.0.10 
    @vercel/nft:  undefined ()
    acorn:  undefined ()
    amphtml-validator:  undefined ()
    anser:  undefined ()
    arg:  undefined ()
    assert:  undefined ()
    async-retry:  undefined ()
[...]

PR #11019, opened 2023-02-22
fix(@aws-amplify/datastore): fix for queries against numeric indexes in IDB

svidgen:
#### Description of changes

Queries against indexed fields in IDB were being queried as strings for `eq` matches due to a bad cast, causing missing results. This PR removes the cast.

Due to the removed cast, this also adds filters `null` and `undefined` conditions out of indexed queries, which would otherwise incorrectly be accepted as IDB key lookup values.

#### Issue #, if available

#11018

Red-greened tested.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/Predicate.ts
+++ b/packages/datastore/__tests__/Predicate.ts
@@ -2219,6 +2219,26 @@ describe('Predicates', () => {
 				matches: [{ name: 'tim' }, { name: 'sam' }],
 				mismatches: [{ name: 'al' }, { name: 'fran' }],
 			},
+			{
+				gql: {
+					and: [{ rating: { gt: 123 } }],
+				},
+				expectedRegeneration: {
+					and: [{ rating: { gt: 123 } }],
+				},
+				matches: [{ rating: 124 }, { rating: 125 }],
+				mismatches: [{ rating: 122 }, { rating: 123 }],
+			},
+			{
+				gql: {
+					and: [{ rating: { eq: 123 } }],
+				},
+				expectedRegeneration: {
+					and: [{ rating: { eq: 123 } }],
+				},
+				matches: [{ rating: 123 }],
+				mismatches: [{ rating: 122 }, { rating: 124 }],
+			},
 		];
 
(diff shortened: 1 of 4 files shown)


packages/datastore/__tests__/Predicate.ts
```diff
@@ -2219,6 +2219,26 @@ describe('Predicates', () => {
...
+			{
+				gql: {
+					and: [{ rating: { eq: 123 } }],
+				},
+				expectedRegeneration: {
+					and: [{ rating: { eq: 123 } }],
+				},
+				matches: [{ rating: 123 }],
+				mismatches: [{ rating: 122 }, { rating: 124 }],
+			},
```
svidgen, line 2241: FWIW, these additions were not helpful in diagnosing the problem. These passed as soon as I added them.

packages/datastore/__tests__/Predicate.ts
```diff
@@ -2270,11 +2290,16 @@ describe('Predicates', () => {
 				matches: [{ name: 'tim' }, { name: 'sam' }],
 				mismatches: [{ name: 'al' }, { name: 'fran' }],
 			},
+			{
+				predicate: p => p.rating.eq(123),
+				matches: [{ rating: 123 }],
+				mismatches: [{ rating: 122 }, { rating: 124 }],
+			},
```
svidgen, line 2297: Same here. There was no issue going from predicate builder to storage predicate AST.

packages/datastore/__tests__/commonAdapterTests.ts
```diff
@@ -243,6 +246,55 @@ export function addCommonQueryTests({
 			);
 			expect(results.length).toEqual(2);
 		});
+
```
svidgen, line 249: In the new tests that follow, the `string` test passed immediately. The two numeric tests showed the incorrect matching and were resolved by removing the cast.

dpilch (approved):

iartemiev (approved):

Issue #11042, opened 2023-03-03
Amplify DataStore - Ability to forward sync errors (e.g unauthorized) from AppSync to Hub events and errorHandler 


abhi7cr:
### Describe the feature you'd like to request

Unauthorized errors from AppSync are not forwarded to the appropriate Hub events and/or the `errorHandler`. There is a Hub event to handle inapplicable records (`nonApplicableDataReceived`) that forwards errors if the data is invalid (e.g. invalid type), but Unauthorized errors from AppSync are not forwarded. Currently, they are just logged as warnings to console.

### Describe the solution you'd like

It could be the same Hub event (`nonApplicableDataReceived`) or a new dedicated event, or the `errorHandler` where initial errors from AppSync (e.g Unauthorized) are captured.

### Describe alternatives you've considered

N/A

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #11062, opened 2023-03-08
Call error handler and Hub on unauthorized sync failure.

dpilch:
#### Description of changes

* Call error handler on unauthorized sync failure
* Call Hub on unauthorized sync failure

I used the same `nonApplicableDataReceived` as all the other errors. Thoughts on using a different event?

The old branching logic between unauthorized errors and other errors. seems to have been a mistake. It may have previously had a meaningful function in the past, but does not in the current version.

#### Issue #, if available

Resolves https://github.com/aws-amplify/amplify-js/issues/11042

Unit testing

Also added a unit test for error handler to be called on unauthorized subscription error. Mutation already has a test case to cover this.

I didn't add hub events for subscription or mutation because there currently are no other hub events in these processors.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/src/sync/processors/sync.ts
```diff
@@ -253,9 +240,9 @@ class SyncProcessor {
 						);
 					}
 
-					if (hasItems && otherErrors?.length) {
```
dpilch, line 256: `hasItems` doesn't seem to be a necessary check here.

packages/datastore/src/sync/processors/sync.ts
```diff
@@ -277,13 +264,19 @@ class SyncProcessor {
 						Hub.dispatch('datastore', {
 							event: 'nonApplicableDataReceived',
```
dpilch, line 265: Thoughts on using a separate event name for unauthorized errors?

svidgen (commented):
If I'm reading this right, `errorHandler` and the hub event are called *every time* there's an unauthorized error, even if the request is getting retried under a different auth mode. But, it might make more sense to do this only once after all auth modes are tried. If we go that route, I think that would occur somewhere around line 154 instead.

Also, can we chat with @manueliglesias and @iartemiev about this one? Per some of our discussion, I'm still not clear whether we need a doc with API review to circulate around for this change.

PR #11102, opened 2023-03-17
chore(notifications): Add unit tests

cshfang:
#### Description of changes
This PR adds the JS unit tests for the Push Notification feature
* This PR also enables `*.native` extension files which have RN dependency to be tested
* This PR also fixes minor issues found as a result of adding unit tests

`yarn test`
Tested locally with sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/notifications/package.json
+++ b/packages/notifications/package.json
@@ -13,7 +13,9 @@
 		"access": "public"
 	},
 	"scripts": {
-		"test": "npm run lint && jest -w 1 --coverage",
+		"test": "npm run lint && npm run test:default && npm run test:native",
+		"test:default": "jest -w 1 --coverage",
+		"test:native": "jest -w 1 --coverage --config=jest.native.config.js --coverageDirectory=coverage-native",
 		"test:watch": "tslint 'src/**/*.ts' && jest -w 1 --watch",
 		"test:size": "size-limit",
 		"build-with-test": "npm run clean && npm test && tsc && webpack",
@@ -69,49 +71,5 @@
 	],
 	"devDependencies": {
 		"@aws-amplify/rtn-push-notification": "0.0.1"
-	},
-	"jest": {
-		"globals": {
-			"ts-jest": {
-				"diagnostics": false,
-				"tsConfig": {
-					"lib": [
-						"es5",
-						"es2015",
-						"dom",
-						"esnext.asynciterable",
-						"es2017.object"
-					],
-					"allowJs": true
-				}
-			}
-		},
-		"transform": {
-			"^.+\\.(js|jsx|ts|tsx)$": "ts-jest"
-		},
-		"testRegex": "(/__tests__/.*|\\.(test|spec))\\.(tsx?|jsx?)$",
-		"moduleFileExtensions": [
-			"ts",
-			"tsx",
-			"js",
-			"json",
-			"jsx"
-		],
-		"testEnvironment": "jsdom",
-		"testURL": "http://localhost/",
-		"coverageThreshold": {
-			"global": {
-				"branches": 0,
-				"functions": 0,
-				"lines": 0,
-				"statements": 0
-			}
-		},
-		"coveragePathIgnorePatterns": [
-			"/node_modules/",
-			"dist",
-			"lib",
-			"lib-esm"
-		]
 	}
 }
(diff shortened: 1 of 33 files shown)


packages/notifications/src/PushNotification/Platform/index.ts
```diff
@@ -3,16 +3,17 @@
...
 const getOS = (): PlatformInterface['OS'] => {
+	if (userAgentIncludes('Android')) return 'android';
+	if (userAgentIncludes('like Mac')) return 'ios';
 	if (userAgentIncludes('Win')) return 'windows';
 	if (userAgentIncludes('Mac')) return 'macos';
-	if (userAgentIncludes('X11')) return 'unix';
 	if (userAgentIncludes('Linux')) return 'linux';
-	if (userAgentIncludes('Android')) return 'android';
-	if (userAgentIncludes('like Mac')) return 'ios';
+	if (userAgentIncludes('X11')) return 'unix';
```
stocaaro, line 16: Could this be a map instead of a sequence of conditionals?

Benefits are less / more direct code. Also, if we're running it a lot, might be faster, although for this number of elements its probably a wash.

packages/notifications/package.json
```diff
@@ -13,7 +13,9 @@
 		"access": "public"
 	},
 	"scripts": {
-		"test": "npm run lint && jest -w 1 --coverage",
+		"test": "npm run lint && npm run test:default && npm run test:native",
+		"test:default": "jest -w 1 --coverage",
+		"test:native": "jest -w 1 --coverage --config=jest.native.config.js --coverageDirectory=coverage-native",
 		"test:watch": "tslint 'src/**/*.ts' && jest -w 1 --watch",
 		"test:size": "size-limit",
```
jimblanc, line 20: Could you also double check the size limits defined in `size-limit` when using the notifications Class (just to double check, may be able to tweak them one way or the other)?

jimblanc (approved):

packages/notifications/src/PushNotification/Platform/index.ts
cshfang: This code would only execute once upon import anyway and, in writing my unit tests, I realized that order matters here so a map wouldn't work

packages/notifications/src/PushNotification/utils.ts
```diff
@@ -114,11 +115,12 @@ const getFcmOptions = ({
...
-	fcmOptions: {
-		channelId,
-		messageId,
-		senderId,
-		sendTime: new Date(sendTime),
-	},
-});
+}): Pick<PushNotificationMessage, 'fcmOptions'> => {
+	const fcmOptions = {
+		...(channelId && { channelId }),
```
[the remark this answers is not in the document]
cshfang: I generally don't like using `&&`  for this purpose and I do find this tricky to understand at first as well. I went back and forth on it a bit and initially did have it just output `undefined`. However I decided ultimately to do this for a few reasons imo
* The end result is consumed by the customer and I think giving them exactly what there *is* is cleaner
* The falsy check works the same in both cases: `!!message.fcmOptions?.channelId`
* This is not true, however, of the key itself - which would be there even if value is undefined otherwise

packages/notifications/package.json
cshfang: How granular do we need to be about the size limit? It is within a whole number at this time.

AllanZhengYP (approved):

packages/notifications/src/PushNotification/utils.ts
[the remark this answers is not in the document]
stocaaro: I've spent some time fiddling around with this. I can get to this result in other ways, but nothing inherently better / less confusing than this. Makes me wonder if the undefined avoidance is worth the added complexity, but not enough to further delay on this.

stocaaro (approved):

packages/notifications/src/PushNotification/Platform/index.ts
cshfang: Imo, the tests should cover it fairly well. This will likely need to be more thoroughly tested in the future if it ever gets used but, In the meantime, this code is almost just a placeholder that *should work*. It's not exposed externally, and is only scoped right now to a React Native-only feature.

As for a different data structure - I think anything else would actually just overcomplicate something that is very simple and straightforward.

PR #11108, opened 2023-03-20
fix(data): update SQLite utils to use 'instr' instead of 'like' when constructing WHERE conditions from predicates

david-mcafee:
#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/datastore-storage-adapter/__tests__/SQLiteUtils.test.ts
```diff
@@ -275,14 +276,69 @@ describe('SQLiteUtils tests', () => {
 			};
 
 			const expected = [
-				`WHERE ("firstName" = ? AND "lastName" LIKE ? AND "sortOrder" > ?)`,
-				['Bob', 'Sm%', 5],
+				`WHERE (\"firstName\" = ? AND instr(\"lastName\", ?) = 1 AND \"sortOrder\" > ?)`,
```
iartemiev, line 279: Nit: I don't think we need to escape the quote characters (`"`) here. The test should still match correctly without the backslash. Might as well keep these consistent with the other tests in this file and replace `\"` with `"`.

iartemiev (approved):
LGTM 👍🏻. Just a couple minor nits

dpilch (approved):
LGTM pending change to circleci config

iartemiev (approved):

dpilch (approved):

svidgen (approved):

PR #11121, opened 2023-03-23
feat: user agent enhancements: part1 core

erinleigh90:
#### Description of changes
User Agent Enhancements Part I for core:
- Adds type for CustomUserAgent and enums for Category, CategoryAction, and Framework
- Updates core Platform to build the user agent according to the new format `aws-amplify/5.0.14 auth/1 framework/1`
- Reverts Datastore to GraphQL calls that use userAgentSuffix
- Framework detection limited to ReactNative or None

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/__tests__/Platform-test.ts
+++ b/packages/core/__tests__/Platform-test.ts
@@ -1,5 +1,5 @@
-import { getAmplifyUserAgent } from '../src/Platform';
-import { Platform } from '../src/Platform';
+import { getAmplifyUserAgent, Platform } from '../src/Platform';
+import { Category, Framework } from '../src';
 
 describe('Platform test', () => {
 	describe('isReactNative test', () => {
@@ -10,12 +10,14 @@ describe('Platform test', () => {
 
 	describe('getAmplifyUserAgent test', () => {
 		test('without content', () => {
-			expect(getAmplifyUserAgent()).toBe(`${Platform.userAgent} {"f":"JS"}`);
+			expect(getAmplifyUserAgent()).toBe(
+				`${Platform.userAgent} (${Framework.None})`
+			);
 		});
 
 		test('with content', () => {
-			expect(getAmplifyUserAgent({ category: 'DataStore' })).toBe(
-				`${Platform.userAgent} {"f":"JS","c":"DataStore"}`
+			expect(getAmplifyUserAgent({ category: Category.DataStore })).toBe(
+				`${Platform.userAgent} (${Category.DataStore},${Framework.None})`
 			);
 		});
 	});
--- a/packages/core/__tests__/Platform-test.ts
+++ b/packages/core/__tests__/Platform-test.ts
@@ -15,7 +15,7 @@ describe('Platform test', () => {
 
 	describe('getAmplifyUserAgent test', () => {
 		test('without customUserAgent', () => {
-			expect(getAmplifyUserAgent()).toBe([
+			expect(getAmplifyUserAgent()).toStrictEqual([
 				['aws-amplify', version],
 				['framework', Framework.None],
 			]);
@@ -27,10 +27,10 @@ describe('Platform test', () => {
 					category: Category.DataStore,
 					action: CategoryAction.DataStoreJitteredRetry,
 				})
-			).toBe([
+			).toStrictEqual([
 				['aws-amplify', version],
 				['category', Category.DataStore],
-				[CategoryAction.DataStoreJitteredRetry],
+				['action', CategoryAction.DataStoreJitteredRetry],
 				['framework', Framework.None],
 			]);
 		});
--- a/packages/core/__tests__/Platform-test.ts
+++ b/packages/core/__tests__/Platform-test.ts
@@ -4,7 +4,7 @@ import {
 	Platform,
 } from '../src/Platform';
 import { version } from '../src/Platform/version';
-import { Category, DataStoreAction, Framework } from '../src/Platform/types';
+import { ApiAction, Category, Framework } from '../src/Platform/types';
 
 describe('Platform test', () => {
 	describe('isReactNative test', () => {
@@ -21,16 +21,16 @@ describe('Platform test', () => {
 			]);
 		});
 
-		/* TODO: Replace "DataStoreAction.None" action with actual action */
+		/* TODO: test with actual API action */
 		test('with customUserAgentDetails', () => {
 			expect(
 				getAmplifyUserAgent({
-					category: Category.DataStore,
-					action: DataStoreAction.None,
+					category: Category.API,
+					action: ApiAction.None,
 				})
 			).toStrictEqual([
 				['aws-amplify', version],
-				[Category.DataStore, DataStoreAction.None],
+				[Category.API, ApiAction.None],
 				['framework', Framework.None],
 			]);
 		});
@@ -46,11 +46,11 @@ describe('Platform test', () => {
 		test('with customUserAgentDetails', () => {
 			expect(
 				getAmplifyUserAgentString({
-					category: Category.DataStore,
-					action: DataStoreAction.None,
+					category: Category.API,
+					action: ApiAction.None,
 				})
 			).toBe(
-				`${Platform.userAgent} ${Category.DataStore}/${DataStoreAction.None} framework/${Framework.None}`
+				`${Platform.userAgent} ${Category.API}/${ApiAction.None} framework/${Framework.None}`
 			);
 		});
 	});
--- a/packages/core/__tests__/Platform-test.ts
+++ b/packages/core/__tests__/Platform-test.ts
@@ -4,7 +4,7 @@ import {
 	Platform,
 } from '../src/Platform';
 import { version } from '../src/Platform/version';
-import { Category, CategoryAction, Framework } from '../src/Platform/types';
+import { Category, Framework } from '../src/Platform/types';
 
 describe('Platform test', () => {
 	describe('isReactNative test', () => {
@@ -25,11 +25,10 @@ describe('Platform test', () => {
 			expect(
 				getAmplifyUserAgent({
 					category: Category.DataStore,
-					action: CategoryAction.DataStoreJitteredRetry,
 				})
 			).toStrictEqual([
 				['aws-amplify', version],
-				[Category.DataStore, CategoryAction.DataStoreJitteredRetry],
+				[Category.DataStore, undefined],
 				['framework', Framework.None],
 			]);
 		});
--- a/packages/core/__tests__/Platform-test.ts
+++ b/packages/core/__tests__/Platform-test.ts
@@ -1,5 +1,10 @@
-import { getAmplifyUserAgent, Platform } from '../src/Platform';
-import { Category, Framework } from '../src/Platform/types';
+import {
+	getAmplifyUserAgent,
+	getAmplifyUserAgentString,
+	Platform,
+} from '../src/Platform';
+import { version } from '../src/Platform/version';
+import { Category, CategoryAction, Framework } from '../src/Platform/types';
 
 describe('Platform test', () => {
 	describe('isReactNative test', () => {
@@ -9,15 +14,38 @@ describe('Platform test', () => {
 	});
 
 	describe('getAmplifyUserAgent test', () => {
-		test('without content', () => {
-			expect(getAmplifyUserAgent()).toBe(
-				`${Platform.userAgent} (${Framework.None})`
+		test('without customUserAgent', () => {
+			expect(getAmplifyUserAgent()).toBe([
+				['aws-amplify', version],
+				['framework', Framework.None],
+			]);
+		});
+
+		test('with customUserAgent', () => {
+			expect(
+				getAmplifyUserAgent({
+					category: Category.DataStore,
+					action: CategoryAction.DataStoreJitteredRetry,
+				})
+			).toBe([
+				['aws-amplify', version],
+				['category', Category.DataStore],
+				[CategoryAction.DataStoreJitteredRetry],
+				['framework', Framework.None],
+			]);
+		});
+	});
+
+	describe('getAmplifyUserAgentString test', () => {
+		test('without customUserAgent', () => {
+			expect(getAmplifyUserAgentString()).toBe(
+				`${Platform.userAgent} framework/${Framework.None}`
 			);
 		});
 
-		test('with content', () => {
-			expect(getAmplifyUserAgent({ category: Category.DataStore })).toBe(
-				`${Platform.userAgent} (${Category.DataStore},${Framework.None})`
+		test('with customUserAgent', () => {
+			expect(getAmplifyUserAgentString({ category: Category.DataStore })).toBe(
+				`${Platform.userAgent} category/${Category.DataStore} framework/${Framework.None}`
 			);
 		});
 	});
--- a/packages/core/__tests__/Platform-test.ts
+++ b/packages/core/__tests__/Platform-test.ts
@@ -10,12 +10,12 @@ describe('Platform test', () => {
 
 	describe('getAmplifyUserAgent test', () => {
 		test('without content', () => {
-			expect(getAmplifyUserAgent()).toBe(Platform.userAgent);
+			expect(getAmplifyUserAgent()).toBe(`${Platform.userAgent} {"f":"JS"}`);
 		});
 
 		test('with content', () => {
-			expect(getAmplifyUserAgent('/DataStore')).toBe(
-				`${Platform.userAgent}/DataStore`
+			expect(getAmplifyUserAgent({ category: 'DataStore' })).toBe(
+				`${Platform.userAgent} {"f":"JS","c":"DataStore"}`
 			);
 		});
 	});
--- a/packages/core/__tests__/Platform-test.ts
+++ b/packages/core/__tests__/Platform-test.ts
@@ -29,8 +29,7 @@ describe('Platform test', () => {
 				})
 			).toStrictEqual([
 				['aws-amplify', version],
-				['category', Category.DataStore],
-				['action', CategoryAction.DataStoreJitteredRetry],
+				[Category.DataStore, CategoryAction.DataStoreJitteredRetry],
 				['framework', Framework.None],
 			]);
 		});
@@ -45,7 +44,7 @@ describe('Platform test', () => {
 
 		test('with customUserAgent', () => {
 			expect(getAmplifyUserAgentString({ category: Category.DataStore })).toBe(
-				`${Platform.userAgent} category/${Category.DataStore} framework/${Framework.None}`
+				`${Platform.userAgent} ${Category.DataStore} framework/${Framework.None}`
 			);
 		});
 	});
--- a/packages/core/__tests__/Platform-test.ts
+++ b/packages/core/__tests__/Platform-test.ts
@@ -1,5 +1,5 @@
 import { getAmplifyUserAgent, Platform } from '../src/Platform';
-import { Category, Framework } from '../src';
+import { Category, Framework } from '../src/Platform/types';
 
 describe('Platform test', () => {
 	describe('isReactNative test', () => {
--- a/packages/core/__tests__/Platform-test.ts
+++ b/packages/core/__tests__/Platform-test.ts
@@ -4,7 +4,7 @@ import {
 	Platform,
 } from '../src/Platform';
 import { version } from '../src/Platform/version';
-import { Category, Framework } from '../src/Platform/types';
+import { Category, DataStoreAction, Framework } from '../src/Platform/types';
 
 describe('Platform test', () => {
 	describe('isReactNative test', () => {
@@ -14,36 +14,43 @@ describe('Platform test', () => {
 	});
 
 	describe('getAmplifyUserAgent test', () => {
-		test('without customUserAgent', () => {
+		test('without customUserAgentDetails', () => {
 			expect(getAmplifyUserAgent()).toStrictEqual([
 				['aws-amplify', version],
 				['framework', Framework.None],
 			]);
 		});
 
-		test('with customUserAgent', () => {
+		/* TODO: Replace "DataStoreAction.None" action with actual action */
+		test('with customUserAgentDetails', () => {
 			expect(
 				getAmplifyUserAgent({
 					category: Category.DataStore,
+					action: DataStoreAction.None,
 				})
 			).toStrictEqual([
 				['aws-amplify', version],
-				[Category.DataStore, undefined],
+				[Category.DataStore, DataStoreAction.None],
 				['framework', Framework.None],
 			]);
 		});
 	});
 
 	describe('getAmplifyUserAgentString test', () => {
-		test('without customUserAgent', () => {
+		test('without customUserAgentDetails', () => {
 			expect(getAmplifyUserAgentString()).toBe(
 				`${Platform.userAgent} framework/${Framework.None}`
 			);
 		});
 
-		test('with customUserAgent', () => {
-			expect(getAmplifyUserAgentString({ category: Category.DataStore })).toBe(
-				`${Platform.userAgent} ${Category.DataStore} framework/${Framework.None}`
+		test('with customUserAgentDetails', () => {
+			expect(
+				getAmplifyUserAgentString({
+					category: Category.DataStore,
+					action: DataStoreAction.None,
+				})
+			).toBe(
+				`${Platform.userAgent} ${Category.DataStore}/${DataStoreAction.None} framework/${Framework.None}`
 			);
 		});
 	});
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -51,7 +51,8 @@
 		"find": "^0.2.7",
 		"genversion": "^2.2.0",
 		"prepend-file": "^1.3.1",
-		"react-native": "^0.64.1"
+		"react-native": "^0.64.1",
+		"@aws-sdk/types": "3.6.1"
 	},
 	"files": [
 		"lib",
@@ -64,7 +65,6 @@
 		"@aws-sdk/client-cloudwatch-logs": "3.6.1",
 		"@aws-sdk/client-cognito-identity": "3.6.1",
 		"@aws-sdk/credential-provider-cognito-identity": "3.6.1",
-		"@aws-sdk/types": "3.6.1",
 		"@aws-sdk/util-hex-encoding": "3.6.1",
 		"tslib": "^1.8.0",
 		"universal-cookie": "^4.0.4",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -47,12 +47,12 @@
 	},
 	"homepage": "https://aws-amplify.github.io/",
 	"devDependencies": {
+		"@aws-sdk/types": "3.6.1",
 		"@react-native-async-storage/async-storage": "1.15.17",
 		"find": "^0.2.7",
 		"genversion": "^2.2.0",
 		"prepend-file": "^1.3.1",
-		"react-native": "^0.64.1",
-		"@aws-sdk/types": "3.6.1"
+		"react-native": "^0.64.1"
 	},
 	"files": [
 		"lib",
--- a/packages/core/src/Platform/detectFramework.ts
+++ b/packages/core/src/Platform/detectFramework.ts
@@ -4,23 +4,4 @@ interface ExtendedWindow extends Window {
 	__REACT_DEVTOOLS_GLOBAL_HOOK__?: any;
 }
 
-export const detectFramework = (isReactNative: Boolean): Framework => {
-	if (isReactNative) {
-		return Framework.ReactNative;
-	}
-
-	if (typeof document === 'undefined') {
-		return Framework.NodeJS;
-	}
-
-	const extendedWindow = window as ExtendedWindow;
-
-	if (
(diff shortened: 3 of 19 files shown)


packages/core/src/Platform/detectFramework.ts
```diff
@@ -0,0 +1,13 @@
+import { Framework } from '../constants';
+
+export const detectFramework = () => {
+	if (
+		Array.from(document.querySelectorAll('*')).some(e =>
+			e.hasOwnProperty('_reactRootContainer')
+		)
```
cshfang, line 7: I know that looking for specific things on the DOM isn't as reliable as this but I would treat this as a *fallback* to those other queries (similar to what you had before at some point in a previous PR) rather than the default. The concern about this is that it could, in theory, have to traverse through a huge document before finding a result (not to mention I think it'd probably first allocate the memory to essentially store all the selection results into an array before the iterations even begin).

You might even consider creating a function which takes in matchers that short circuit so you can pass in different matching criteria that go from quick but not always reliable to long but reliable
```ts
const isFramework = (matchers) => matchers.some(matcher => matcher())

export const detectFramework = () => {
  const reactMatchers = [
    () => !!window?.React,
    () => !!document.querySelector('[data-something-reacty]'),
    () => Array.from(document.querySelectorAll('*')).some(e =>
      e.hasOwnProperty('_reactRootContainer')
    );
  ]
  if (isFramework(reactMatchers)) { return Framework.React }
}
```

As you add more frameworks to detect, you may even find that some of the search logic can be reused (especially that last query) so you may be able to optimize further by combining them as well. Just something to consider.

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,30 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+import { CustomUserAgent } from '../types/types';
```
cshfang, line 3: I'm not sure why this folder is structured the way it is but we should really either:
1. Just get rid of the folder - leaving a `types.ts` on the `src` root
2. Get rid of `index.ts` and just rename the current `types.ts` to `index.ts`

Having an index whose sole purpose to to re-export another file seems pretty pointless.

Even if we don't want to make that change in this PR, I would suggest you update this to:
```suggestion
import { CustomUserAgent } from '../types';
```

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,30 @@
...
 
-export const getAmplifyUserAgent = (content?: string) => {
-	return `${Platform.userAgent}${content ? content : ''}`;
+export const getAmplifyUserAgent = (customUserAgent?: CustomUserAgent) => {
+	return `${Platform.userAgent} ${buildUserAgentDetails(customUserAgent)}`;
+};
+
+const buildUserAgentDetails = (customUserAgent?: CustomUserAgent) => {
+	const userAgentDetails = { framework: detectFramework(), ...customUserAgent };
+	return `(${Object.values(userAgentDetails).sort().toString()})`;
```
cshfang, line 29: I think the `toString()` can be omitted?

packages/core/src/index.ts
```diff
@@ -32,13 +32,16 @@ export { FacebookOAuth, GoogleOAuth } from './OAuthHelper';
 export { AppState, AsyncStorage, Linking } from './RNComponents';
 export { Credentials, CredentialsClass } from './Credentials';
 export { ServiceWorker } from './ServiceWorker';
-export { ICredentials } from './types';
+export { ICredentials, CustomUserAgent } from './types';
```
cshfang, line 35: Similar question to the enums below - does this type have any relevance outside of the user agent context?

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,30 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+import { CustomUserAgent } from '../types/types';
 import { version } from './version';
+import { detectFramework } from './detectFramework';
 
 const BASE_USER_AGENT = `aws-amplify/${version}`;
 
 export const Platform = {
```
cshfang, line 9: Should this be typed?

packages/core/src/types/types.ts
```diff
@@ -63,3 +64,15 @@ export interface CloudWatchDataTracker {
 	logEvents: InputLogEvent[];
 	verifiedLogGroup?: LogGroup;
 }
+
+export type Platform = {
```
cshfang, line 68: I may have missed it, but is this being used anywhere right now?

packages/core/src/types/types.ts
```diff
@@ -63,3 +64,15 @@ export interface CloudWatchDataTracker {
...
+	userAgent: string;
+	product: string;
+	navigator: Navigator | null;
+	isReactNative: Boolean;
+};
+export type CustomUserAgent = {
```
cshfang, line 74: Ditto here, is there a reason barring us from putting all these relevant things inside `'../Platform'`?
As it stands - it seems like to use all these things together I have to do something like
```ts
import { Category, CategoryAction, Framework } from '../constants';
import { Platform as PlatformType, CustomUserAgent } from '../types';
import { getAmplifyUserAgent, Platform } from '../Platform';
```

packages/datastore/__tests__/mutation.test.ts
```diff
@@ -17,8 +17,7 @@ import {
 } from '../src/types';
 import { createMutationInstanceFromModelOperation } from '../src/sync/utils';
 import { MutationEvent } from '../src/sync/';
-import { Constants } from '@aws-amplify/core';
-import { USER_AGENT_SUFFIX_DATASTORE } from '../src/util';
+import { Constants, Category, Framework } from '@aws-amplify/core';
```
cshfang, line 20: Just a thought/question but should these be exported from the root core? Can we have them be exported on a scoped path instead, especially keeping V6 propositions ahead?

packages/core/__tests__/Platform-test.ts
```diff
@@ -10,12 +10,14 @@ describe('Platform test', () => {
 
 	describe('getAmplifyUserAgent test', () => {
 		test('without content', () => {
```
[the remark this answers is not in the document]
erinleigh90: Right, the getAmplifyUserAgent used to take a param called content, now it takes customUserAgent. Will update on this iteration.

packages/core/src/types/types.ts
erinleigh90: I'm not sure what happened here, I think this was a holdover from someone else's unrelated change that has since disappeared. Will remove.

packages/core/__tests__/Platform-test.ts
```diff
@@ -1,5 +1,5 @@
-import { getAmplifyUserAgent } from '../src/Platform';
-import { Platform } from '../src/Platform';
+import { getAmplifyUserAgent, Platform } from '../src/Platform';
+import { Category, Framework } from '../src';
```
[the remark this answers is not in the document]
erinleigh90: I moved the Category and Framework definition to the Platform folder and updated the import to come from '../src/Platform/types', but I don't know if that truly resolves this item, since the core library still needs to export these unless we go with exporting on a scoped path. I'll discuss with the team.

packages/core/src/Platform/index.ts
erinleigh90: Yes, missed porting this change over to the clean feature branch, thank you!

packages/core/src/Platform/detectFramework.ts
jimblanc: We'll need to make sure this utility is safe on Node (SSR) and RN as well, `document` & `window` won't be available on Node for example. We could just have another set of matches that gets run first that could short circuit this if they're not available.

packages/datastore/__tests__/mutation.test.ts
jimblanc: In this case for v5, I'd probably prefer not changing our module exports to minimize risk (just until we can iron out the practicalities of doing so, e.g. Dustin's PR to `next`). Either way, the exports would still be part of our public API surface.

packages/core/src/Platform/detectFramework.ts
```diff
@@ -0,0 +1,26 @@
+import { Framework } from './types';
+
+interface ExtendedWindow extends Window {
+	__REACT_DEVTOOLS_GLOBAL_HOOK__?: any;
+}
+
+export const detectFramework = (isReactNative: Boolean): Framework => {
```
david-mcafee, line 7: Why do we pass `isReactNative` instead of detecting it here?

packages/core/src/Platform/types.ts
```diff
@@ -0,0 +1,32 @@
...
+	PubSub = 'pubsub',
+	PushNotification = 'pushnotification',
+	Storage = 'storage',
+}
+
+export enum CategoryAction {
+	AuthSignUp = '1',
+	AuthSignIn = '2',
+	APIQuery = '3',
+	DataStoreJitteredRetry = '4',
```
david-mcafee, line 24: Not blocking, just curious: looks like we aren't actually logging these actions anywhere, is that part2 / 3 of user agent enhancements? Would love to discuss offline what actions we are thinking of adding for API + DataStore!

david-mcafee (changes requested):
Left some feedback specifically around the framework detection :)

packages/core/src/Platform/types.ts
erinleigh90: Yeah, part 4. Will reduce the final project to just what is going in but need to make sure the format works with action.

packages/core/src/Platform/detectFramework.ts
erinleigh90: Good point 😁

packages/core/src/Platform/detectFramework.ts
```diff
@@ -0,0 +1,26 @@
...
+
+	if (typeof document === 'undefined') {
+		return Framework.NodeJS;
+	}
+
+	const extendedWindow = window as ExtendedWindow;
+
+	if (
+		extendedWindow.__REACT_DEVTOOLS_GLOBAL_HOOK__ &&
+		extendedWindow.__REACT_DEVTOOLS_GLOBAL_HOOK__.renderers.size > 0
```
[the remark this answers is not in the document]
erinleigh90: Interesting! I will definitely check on that, there are quite a few global variables related to react, we can probably just look for the global hook in general here. If you have apps that are setup for various react setups I would love to check them out for testing! 

Framework detection is Part III officially so this is the first run at React so we can make sure the format is coming through as we want it. I don't think this should block a merge into the feature branch.

packages/core/src/Platform/detectFramework.ts
erinleigh90: I guess the only benefit is we aren't pulling in framework here just for reactnative, detectFramework will return the framework result. It keeps platform/index clean of Framework.XXXX

david-mcafee: Hm, to me, it just seems odd that the concern of this function is to either 1) detect the framework, or 2) return the Framework name when passed a boolean. If I was debugging this, I would be confused, since I'd expect framework detection to happen in a single place. What do you think?

packages/core/src/Platform/types.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export enum Framework {
+	None = '0',
```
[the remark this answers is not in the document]
stocaaro: Maybe I'm wrong. The right hand values are still numbers, aren't they?

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,71 @@
...
+const buildUserAgentDetailsString = (
+	customUserAgent?: CustomUserAgent
+): string => {
+	const userAgentTuples = buildUserAgentTuples(customUserAgent);
+	let userAgentDetailsString = '';
+	for (let i = 0; i < userAgentTuples.length; i++) {
+		userAgentDetailsString += `${userAgentTuples[i][0]}${
+			userAgentTuples[i][1] ? `/${userAgentTuples[i][1]}` : ''
+		} `;
+	}
```
stocaaro, line 68: This feels like it could be code golfed into something a bit more concise without making it less readable. Maybe:
```ts
userAgentTuples.map(([agentKey, agentValue]) => `${agentKey}/${agentValue ?? ""}`).join(" ")
```

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,71 @@
...
+	return buildUserAgentTuples(customUserAgent);
+};
+
+export const getAmplifyUserAgentString = (
+	customUserAgent?: CustomUserAgent
+): string => {
+	return buildUserAgentDetailsString(customUserAgent);
+};
+
+const buildUserAgentTuples = (
```
stocaaro, line 36: Would it make more sense for the AWSUserAgent to know how to translate itself into tuples? If this object is publicly exposed, could we decorate it into an object that knows how to translate it to make the translation more of an "is a" vs the current "has a" type of action?

In usage, thinking that `customUserAgent.tuples` or `(new CustomerUserAgentTranslator(userAgent)).tuples` would be more concise and direct in this code, where internally, this could be making `this` refs to access its members.

stocaaro (commented):
Structure makes sense. Have a few nits and one piece of structural preference. Happy to clarify if my comments need it.

packages/core/src/Platform/detectFramework.ts
```diff
@@ -1,10 +1,23 @@
...
+export const detectFramework = (isReactNative: Boolean): Framework => {
+	if (isReactNative) {
+		return Framework.ReactNative;
+	}
+
+	if (typeof document === 'undefined') {
+		return Framework.NodeJS;
+	}
+
+	const extendedWindow = window as ExtendedWindow;
```
jimblanc, line 16: Along the lines of what we were discussing, we should make sure `window` exists as well before we try to use it.

packages/core/src/Platform/index.ts
```diff
@@ -3,11 +3,12 @@
 import { CustomUserAgent } from './types';
 import { version } from './version';
 import { detectFramework } from './detectFramework';
+import { UserAgent as AWSUserAgent } from '@aws-sdk/types';
```
jimblanc, line 6: Nit-pick unrelated to your change, but we might want to move `"@aws-sdk/types": "3.6.1",` to `devDependencies`. 

Also this type is fairly simple:

```
/**
 * @public
 *
 * A tuple that represents an API name and optional version
 * of a library built using the AWS SDK.
 */
export type UserAgentPair = [name: string, version?: string];

/**
 * @public
 *
 * User agent data that to be put into the request's user
 * agent.
 */
export type UserAgent = UserAgentPair[];
```

I wonder if we could just duplicate this locally in order to avoid another code dependency on the SDK? Might be more pertinent once we start on the custom clients, just something to consider though.

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,71 @@
...
+): string => {
+	return buildUserAgentDetailsString(customUserAgent);
+};
+
+const buildUserAgentTuples = (
+	customUserAgent?: CustomUserAgent
+): AWSUserAgent => {
+	const userAgentDetails = {
+		...customUserAgent,
+	};
```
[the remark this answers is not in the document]
cshfang: +1 I was trying to add this comment yesterday from my phone but decided to review more properly in front of a computer. To add to @stocaaro 's suggestion, I do believe all the properties we are accessing from `customUserAgent` are primitives as well.

I would take it a step further and directly destructure the properties you need to inspect/copy directly on the params to improve readability in the body of this function. E.g.
```
const buildUserAgentTuples = (
  { action, category, framework }: CustomUserAgent = {}
```

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,71 @@
...
+	const userAgentTuples: AWSUserAgent = [[BASE_USER_AGENT, version]];
+	if (userAgentDetails.category) {
+		userAgentTuples.push([userAgentDetails.category, userAgentDetails.action]);
```
cshfang, line 44: Does this imply `action` is guaranteed to be present just because `category` is? If so, we may want/need to convey this hard correlation via typing

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,71 @@
...
+	}
+
+	if (userAgentDetails.framework) {
+		userAgentTuples.push(['framework', userAgentDetails.framework]);
+	} else {
+		userAgentTuples.push([
+			'framework',
+			detectFramework(Platform.isReactNative),
+		]);
+	}
```
cshfang, line 54: I think this can be simplified to improve readability
```suggestion
	userAgentTuples.push([
	  'framework,
	  userAgentDetails.framework ?? detectFramework(Platform.isReactNative)
	]);
```

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,71 @@
...
 	}
 }
 
-export const getAmplifyUserAgent = (content?: string) => {
-	return `${Platform.userAgent}${content ? content : ''}`;
+export const getAmplifyUserAgent = (
+	customUserAgent?: CustomUserAgent
+): AWSUserAgent => {
+	return buildUserAgentTuples(customUserAgent);
+};
```
cshfang, line 28: Nit:
```suggestion
): AWSUserAgent => buildUserAgentTuples(customUserAgent);
```

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,71 @@
...
-			Platform.isReactNative = false;
-			break;
+	if (navigator.product === 'ReactNative') {
+		Platform.isReactNative = true;
 	}
 }
 
-export const getAmplifyUserAgent = (content?: string) => {
-	return `${Platform.userAgent}${content ? content : ''}`;
+export const getAmplifyUserAgent = (
```
cshfang, line 24: Similar line of thinking here - I don't think we have a need to extract the build logic from this function considering this function does nothing except to return the result of that build logic.

Suggestion:
*remove* `getAmplifyUserAgent`
*rename* `buildUserAgentTuples` => `getAmplifyUserAgent`
*update usage of* `buildUserAgentTuples` => `getAmplifyUserAgent`

packages/core/src/index.ts
```diff
@@ -35,7 +35,17 @@ export { ServiceWorker } from './ServiceWorker';
...
+export {
+	Platform,
+	getAmplifyUserAgent,
+	getAmplifyUserAgentString,
+} from './Platform';
+export {
+	CustomUserAgent,
+	Category,
+	CategoryAction,
+	Framework,
```
cshfang, line 47: Nit: I personally think there is value in alphabetizing imports for readability but this already looks much cleaner than before thank you!

packages/core/src/Platform/types.ts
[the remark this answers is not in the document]
erinleigh90: Right, for framework they are still numbers.

packages/core/src/Platform/detectFramework.ts
erinleigh90: To make this clear: we are currently detecting ReactNative in an almost deprecated way. I don't want to double down on that, but it's something we can use for now to have this change not break RN customers, until I get to Part III of development (detection of frameworks). I think my solution right now is to move detect framework changes to a new branch so we can merge part1, so it doesn't block all downstream work from getting reviewed.

I also want to bring up that reworking Platform was not part of my original scope, and while it's not a large file, it may be wise to look into Platform refactor as a follow-up to this project, since we are trying to get the user agent enhancements out as soon as possible.

packages/core/src/Platform/detectFramework.ts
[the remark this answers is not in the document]
david-mcafee: @cshfang - see my comment above, but I'm not sure that checking the renderers size is a valid way to detect React, as the value changes whether or not React Dev Tools is enabled (`1` when enabled, `0` when disabled).

We can possibly change the condition to `size >= 0`, but I haven't done enough digging into this to know if that's the best practice for detecting React, here.

packages/core/src/Platform/detectFramework.ts
erinleigh90: Note: none of this code will actually be shipped until Part III is complete.

packages/core/src/Platform/detectFramework.ts
david-mcafee, line 20:
```suggestion
/**
 * TODO: revisit React platform detection logic:
 * currently, the value of `renderers.size` is
 * affected by whether React Dev Tools is enabled.
 */
	if (
		extendedWindow.__REACT_DEVTOOLS_GLOBAL_HOOK__ &&
		extendedWindow.__REACT_DEVTOOLS_GLOBAL_HOOK__.renderers.size >= 0
```

packages/core/src/Platform/index.ts
erinleigh90: Nice! Way better.

packages/core/src/Platform/index.ts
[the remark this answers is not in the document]
erinleigh90: Oh yes, holdover from the previous iteration. It made sense then, not so much now.

Chris, thank you for that, makes a whole lot of sense.

packages/core/src/Platform/detectFramework.ts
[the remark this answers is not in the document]
david-mcafee: Also, as I mentioned before, I think it would be beneficial to add comments clarifying how things are working in this file, and some of the reasoning around why we're detecting platforms the way we are.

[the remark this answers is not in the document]
cshfang: Yeah I should've been clear that I was not commenting on whether this is the correct way to detect the framework, only that the logic can be simplified

packages/core/src/Platform/detectFramework.ts
erinleigh90: Resolving this for this PR Since I moved this code. Super helpful conversation for later reference. Thank you!

packages/core/src/Platform/detectFramework.ts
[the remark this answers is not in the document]
erinleigh90: Thank you for the feedback. Moving the code to a part 3 branch and resolving for now, will reference when I start moving on part 3.

packages/core/src/Platform/detectFramework.ts
```diff
@@ -0,0 +1,26 @@
+import { Framework } from './types';
+
+interface ExtendedWindow extends Window {
```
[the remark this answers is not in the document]
erinleigh90: Marking as resolved for now but will reference when I move to part 3

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,69 @@
...
+	if (userAgentDetails.category) {
+		/** TODO: add action as second element */
+		userAgentTuples.push([userAgentDetails.category, undefined]);
+	}
+
+	if (userAgentDetails.framework) {
+		userAgentTuples.push(['framework', userAgentDetails.framework]);
+	} else {
+		userAgentTuples.push(['framework', detectFramework()]);
+	}
```
cshfang, line 52: I just noticed a comment of mine got lost into the ether..

I think this can be made more concise without losing clarity:
```suggestion
	userAgentTuples.push(['framework', userAgentDetails.framework ? userAgentDetails.framework : detectFramework()]);
```

packages/core/src/Platform/detectFramework.ts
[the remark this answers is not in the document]
erinleigh90: Resolving for now, will address in Phase 3

packages/core/src/Platform/index.ts
erinleigh90: Going to go with Chris here.

packages/core/src/Platform/detectFramework.ts
erinleigh90: Will refer back to this in part 3, resolving for now as this code has been removed

packages/core/src/Platform/detectFramework.ts
erinleigh90: Resolving, moved code to part 3 branch.

stocaaro (approved):

AllanZhengYP (approved):
Looks good to me! 🚀

david-mcafee (approved):
LGTM!

packages/core/package.json
```diff
@@ -51,7 +51,8 @@
 		"find": "^0.2.7",
 		"genversion": "^2.2.0",
 		"prepend-file": "^1.3.1",
-		"react-native": "^0.64.1"
+		"react-native": "^0.64.1",
+		"@aws-sdk/types": "3.6.1"
```
jimblanc, line 55: Nit-pick: This should be at the top of the list (alphabetical)

packages/core/src/Platform/detectFramework.ts
```diff
@@ -0,0 +1,8 @@
+import { Framework } from './types';
```
jimblanc, line 1: This file needs the copyright header.

```suggestion
// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
// SPDX-License-Identifier: Apache-2.0
import { Framework } from './types';
```

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,54 @@
...
-			Platform.isReactNative = true;
-			break;
-		default:
-			Platform.userAgent = `${BASE_USER_AGENT} js`;
-			Platform.isReactNative = false;
-			break;
+
+// Rerun framework detection once when getAmplifyUserAgent is called if framework is None:
+// ReactNative framework must be detected initially, however
+// Other frameworks may not be detected in cases where DOM is not yet loaded
```
jimblanc, line 20: Nit: Since this is a function, we should use a proper doc-string format:

```suggestion
/**
 * Rerun framework detection once when getAmplifyUserAgent is called if framework is None.
 * ReactNative framework must be detected initially, however other frameworks may not be 
 * detected in cases where DOM is not yet loaded.
 */
```

packages/core/src/Platform/index.ts
```diff
@@ -1,30 +1,54 @@
...
+const rerunFrameworkDetection = () => {
+	if (Platform.framework === Framework.None && !frameworkHasBeenRerun) {
+		frameworkHasBeenRerun = true;
+		framework = detectFramework();
+		Platform.framework = framework;
```
jimblanc, line 25:
```suggestion
        framework = detectFramework();
		frameworkHasBeenRerun = true;
		Platform.framework = framework;
```

Nit: Run detection first then update states

jimblanc (changes requested):
Looks good! Some nits below

packages/core/src/Platform/index.ts
```diff
@@ -1,69 +1,54 @@
...
-		Platform.isReactNative = true;
-	}
-}
-
-export const getAmplifyUserAgent = (
-	customUserAgent?: CustomUserAgent
-): AWSUserAgent => {
-	return buildUserAgentTuples(customUserAgent);
+	framework,
+	isReactNative: framework === Framework.ReactNative,
```
cshfang, line 15: Is this just for backward compatibility? Can it be either removed (always rely on comparing framework) or extracted (separate convenience util for checking framework is RN) in V6?

packages/core/src/Platform/index.ts
```diff
@@ -1,69 +1,54 @@
...
-	if (userAgentDetails.category) {
+export const getAmplifyUserAgent = ({
+	category,
+	action,
+	framework,
+}: CustomUserAgentDetails = {}): AWSUserAgent => {
+	rerunFrameworkDetection();
+	const userAgent: AWSUserAgent = [[BASE_USER_AGENT, version]];
+	if (category) {
 		/** TODO: add action as second element */
```
cshfang, line 37: Is this TODO still valid?

packages/core/src/Platform/types.ts
```diff
@@ -16,9 +16,71 @@ export enum Category {
...
+}
+export enum PredictionsAction {
+	None = '0',
+}
+export enum PubSubAction {
+	None = '0',
+}
+export enum PushNotificationAction {
+	None = '0',
+}
```
cshfang, line 47: Will this get wrapped into Notifications or should we also have an `InAppMessagingAction` enum?

cshfang (approved):

packages/core/src/Platform/index.ts
erinleigh90: Yes, this is for backward compatibility, hopefully we can remove the need for this in V6 and move away from the use of navigator.product as well.

packages/core/src/Platform/index.ts
erinleigh90: Nope!

packages/core/src/Platform/types.ts
erinleigh90: Each category will have it's own action enum. If I missed any in this first pass they will get added as I work on categories.

cshfang (approved):

jimblanc (approved):

stocaaro (approved):

johnf:
@erinleigh90 There are lots of changes related to this in the 5.3.0 release.
Is there any documentation regarding what this is all about.
It would be amazing of the release notes could expand on larger changes like this.

PR #11131, opened 2023-03-23
chore(@aws-amplify/datastore): add simulated latency to interactions against gql fake

svidgen:
#### Description of changes

Adds the ability to introduce meaningful latency and jitter into test interactions against the graphql service fake. This will help with future fuzz testing and targeted tests intended to codify and/or fix expected behavior against slow connections.

Runtime fixes that this incorporates have been broken out into two other PR's:
1. #11132
2. #11133 

Those will be merged first prior to marking this one "ready for review".

#### Description of how you validated changes

No change to library runtime behavior. Only affects tests. Verified that latencies are used by increasing them and watching the affected test run times go up accordingly.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #11132, opened 2023-03-24
fix(@aws-amplify/datastore): mutation hub event drops during reconnect

svidgen:
#### Description of changes

[`outboxMutationProcessed`](https://docs.amplify.aws/lib/datastore/datastore-events/q/platform/js/#outboxmutationprocessed) Hub events were being dropped for mutations that were being raced against reconnect activities. A fringe case, but during reconnect, this could lead to customer code waiting for `outboxMutationProcessed` events that would never arrive.

This PR fixes the issue by properly removing the observer from the mutation processor during an unsubscribe or `stop()`. An `undefined` observer signals to the mutation processor not to immediately resume on mutation, but instead wait for an observer (the sync engine) to register itself and re-`start()` the mutation processor.

This PR also incidentally tamps down on `console.warn` noise in the scope of `connectivityHandler.test.ts`.

Un-skipped the test that originally caught the error. Made it pass.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/connectivityHandling.test.ts
+++ b/packages/datastore/__tests__/connectivityHandling.test.ts
@@ -538,6 +538,65 @@ describe('DataStore sync engine', () => {
 			) as any;
 			expect(cloudAnotherPost.title).toEqual('another title');
 		});
+
+		test('survives online -> offline -> update/online race', async () => {
+			const post = await DataStore.save(
+				new Post({
+					title: 'a title',
+				})
+			);
+
+			await waitForEmptyOutbox();
+			await simulateDisconnect();
+
+			const outboxEmpty = waitForEmptyOutbox();
+
+			const retrieved = await DataStore.query(Post, post.id);
+			await DataStore.save(
+				Post.copyOf(retrieved!, updated => (updated.title = 'new title'))
+			);
+
+			// NO PAUSE: Simulate reconnect IMMEDIATELY, causing a race
+			// between the save and the sync engine reconnection operations.
+
+			await simulateConnect();
+			await outboxEmpty;
+
+			const table = graphqlService.tables.get('Post')!;
+			expect(table.size).toEqual(1);
+
+			const cloudPost = table.get(JSON.stringify([post.id])) as any;
+			expect(cloudPost.title).toEqual('new title');
+		});
+
+		test('survives online -> offline -> delete/online race', async () => {
+			const post = await DataStore.save(
+				new Post({
+					title: 'a title',
+				})
+			);
+
+			await waitForEmptyOutbox();
+			await simulateDisconnect();
+
+			const outboxEmpty = waitForEmptyOutbox();
+
+			const retrieved = await DataStore.query(Post, post.id);
+			await DataStore.delete(retrieved!);
+
+			// NO PAUSE: Simulate reconnect IMMEDIATELY, causing a race
+			// between the save and the sync engine reconnection operations.
+
+			await simulateConnect();
+			await outboxEmpty;
+
+			const table = graphqlService.tables.get('Post')!;
+			expect(table.size).toEqual(1);
+
+			const cloudPost = table.get(JSON.stringify([post.id])) as any;
+			expect(cloudPost.title).toEqual('a title');
+			expect(cloudPost._deleted).toEqual(true);
+		});
 	});
 
 	describe('selective sync', () => {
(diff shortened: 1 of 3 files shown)


packages/datastore/src/sync/processors/mutation.ts
```diff
@@ -256,7 +267,7 @@ class MutationProcessor {
 						hasMore = (await this.outbox.peek(storage)) !== undefined;
 					});
 
-					this.observer.next!({
+					this.observer?.next?.({
```
david-mcafee, line 270: It's entirely possible I'm being dense, but according to the PR description: "An `undefined` observer signals to the mutation processor not to immediately resume on mutation, but instead wait for an observer (the sync engine) to register itself and re-start() the mutation processor`. Wouldn't this mean that resume doesn't happen if the observer is `undefined`? And if so, why is it possible that the observer is possibly `undefined`, here?

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -498,16 +503,7 @@ describe('DataStore sync engine', () => {
...
-		 * Existing bug. (Sort of.)
-		 *
-		 * Outbox mutations are processed, but the hub events are not sent, so
-		 * the test hangs and times out. :shrug:
-		 *
-		 * It is notable that the data is correct in this case, we just don't
-		 * receive all of the expected Hub events.
-		 */
-		test.skip('survives online -> offline -> save/online race', async () => {
+		test('survives online -> offline -> save/online race', async () => {
```
david-mcafee, line 506: Thoughts on also adding tests for `update` and `delete` immediately before reconnect?

david-mcafee (commented):
Added a few questions

manueliglesias (dismissed):
LGTM!

I only have one question (non-blocking)

packages/datastore/src/sync/processors/mutation.ts
svidgen: I wrestled with this little part of the diff for longer than I care to admit, actually. TypeScript isn't convinced that `this.observer` always exists here, or that `.next` exists on the observer. So, with that said, I'm actually not 100% sure PR eliminates *all* cases where a mutation event goes missing. It does provide an incremental improvement to one edge I identified while trying to add tests.

Maybe it makes more sense to revert to bangs to at least get errors logged if someone hits that edge. I think we can shake out more edges after https://github.com/aws-amplify/amplify-js/pull/11131, which will give us a little more control to stress the protocol.

packages/datastore/src/sync/processors/mutation.ts
```diff
@@ -130,6 +138,8 @@ class MutationProcessor {
 			}
 
 			return this.runningProcesses.addCleaner(async () => {
+				// The observer has unsubscribed and/or `stop()` has been called.
+				this.observer = undefined;
```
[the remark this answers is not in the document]
svidgen: We weren't calling it prior, so I assume it's inconsequential. But, for completeness, I may as well try adding it.

packages/datastore/__tests__/connectivityHandling.test.ts
svidgen: I don't think I have a compelling reason not to! I'll add them in.

manueliglesias (approved):
LGTM 👍

packages/datastore/src/sync/processors/mutation.ts
david-mcafee: Hm, though we check that the observer isn't `undefined` at the beginning of `resume` (i.e. the status of `isReady`), it's possible that between there and here, the observer can now become `undefined`. I think I'm just most concerned about silent failures, and think that error logs would be better. Thoughts?

dpilch (approved):

packages/datastore/src/sync/processors/mutation.ts
dpilch: I think TypeScript is just confused that `this.isReady` actually asserts that `this.observer` is defined. Even if you place `this.isReady` directly before `this.observer.next` you will still get the same error.

```
if (this.isReady()) {
  // still thinks that this.observer could be undefined
  this.observer.next ...
```

You could use assertion predicates to give TypeScript enough information to know it would be defined in this code path. But it looks like it isn't well supported on private class properties and would require some hacks.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

https://stackoverflow.com/questions/74264474/in-typescript-how-do-i-assert-the-value-of-a-class-variable-using-a-method

iartemiev (approved):

Issue #4112, opened 2019-10-01
Persist and expose AuthErrorType in AuthError. Also, allow AuthErrorTypes to be referenced.


clearwaterstream:
**Is your feature request related to a problem? Please describe.**

I need to check the error type for my flow. If I call `Auth.signup()`, for example, in the `catch(err)` block I need to do something like:

```javascript
import { Auth } from 'aws-amplify';

Auth.signUp(/*blah*/)
.catch(err => {
	if(err.errorType === AuthErrorTypes.emptyUsername) {
          // blah
	}
});
```

I think checking for errorType vs the actual error message (which can vary by locale) creates a cleaner and more readable code.

**Describe the solution you'd like**

The `type` is already being passed in to the constructor of `AuthError` class. It is simply used to resolve the error messages from `authErrorMessages` dictionary. So the `message` is available in the instance of AuthError class, the `type` however is lost once the constructor ran its course. 

```javascript
export class AuthError extends Error {
    public log: string;
    constructor(type: AuthErrorTypes) {
        const { message, log } = authErrorMessages[type];
        super(message);

        // Hack for making the custom error class work when transpiled to es5
        // TODO: Delete the following 2 lines after we change the build target to >= es2015
        this.constructor = AuthError;
        Object.setPrototypeOf(this, AuthError.prototype);

        this.name = 'AuthError';
        this.log = log || message;

        logger.error(this.log);
    }
}
```

**Can we add `errorType` public field to store the passed in `errorType` so that we can access it later?**

Also, how can I reference `AuthErrorTypes` in my code? I tried

```javascript
import { AuthErrorTypes } from 'aws-amplify';
```
and it did not work.

**Describe alternatives you've considered**

I tried  doing error message (vs type) comparison by trying to reference this class:

```javascript
export const authErrorMessages: AuthErrorMessages = { /* ... */
```

via

```javascript
import { authErrorMessages } from 'aws-amplify'
```

but that did not work. I'm running this in a React app btw, js.

**Additional context**
THANK YOU!!

P.S. The concept of "error type" or "error code" is probably used everywhere in Amplify framework, in one form another. I would consider adding "errorType" or "errorCode" to the base Error class. I think for multi-lingual apps having error code is helpful.

dohomi:
@haverchuck is there any workaround? Its quite common to have some error specific code and I don't understand why the AuthErrorTypes aren't re-exported?

jimblanc:
[...]

PR #11153, opened 2023-03-29
feat(auth): add sign up error handling

israx:
#### Description of changes

Changes added:
1. Validation errors
This errors are based on [swift](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Support/Constants/AuthPluginErrorConstants.swift#L167) validation constants.

It also address this [issue](https://github.com/aws-amplify/amplify-js/issues/4112)

2. Mapped service errors
Service errors are map into AuthError 

3. Service errors as enums. 
These enums are used on type doc annotation only and are exposed to customers for troubleshooting.

Changes not added:
1. Config errors
 config errors should be checked at one source of truth. The [singleton](https://github.com/aws-amplify/amplify-js/blob/poc/modular-auth/packages/core/src/Amplify.ts#L152) could be the place where that happens 
2. Unit tests
unit tests will be added after the first revision

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/auth/src/error/utils/assertServiceError.ts
```diff
@@ -0,0 +1,23 @@
...
+};
+
+
+export function assertServiceError(
+	error: unknown
+): asserts error is ServiceError {
+	if (
+		!error ||
+		!((error as ServiceError).name ||
+		!(error as ServiceError).message)
```
cshfang, line 15: Could we double check the logic here?

I am _guessing_ it was intended to be "if error does not contain either a name or message" so the second `!` might be throwing that off

packages/auth/src/error/utils/assertServiceError.ts
```diff
@@ -0,0 +1,23 @@
+import { AuthError } from "../AuthError";
+
+type ServiceError = {
+	name: string;
+	message: string;
+};
+
+
+export function assertServiceError(
```
cshfang, line 9: This also seems like it could be broadly applicable to Amplify as a whole, is there a reason we are scoping it to just Auth?

packages/auth/src/error/utils/assertValidationError.ts
```diff
@@ -0,0 +1,66 @@
+import { AuthError } from "../AuthError";
+import { AuthValidationErrorCode , AuthValidationErrorMap} from "../types/validation";
+
+const validationErrorMap:AuthValidationErrorMap ={
```
cshfang, line 4: Why does this need to be duplicated?

packages/auth/src/error/types/validation.ts
```diff
@@ -0,0 +1,66 @@
...
+	EmptyConfirmResetPasswordUsername = 'EmptyConfirmResetPasswordUsername',
+	EmptyConfirmResetPasswordNewPassword = 'EmptyConfirmResetPasswordNewPassword',
+	EmptyConfirmResetPasswordConfirmationCode = 'EmptyConfirmResetPasswordConfirmationCode',
+	EmptyResetPasswordUsername = 'EmptyResetPasswordUsername',
+}
+
+export type AuthValidationErrorMap = {
+	[AuthValidationErrorCode.EmptyChallengeResponse]: {
+		message: 'challengeResponse is required to confirmSignIn';
+		recovery: 'Make sure that a valid challenge response is passed for confirmSignIn';
```
cshfang, line 20: If we align this as `recoverySuggestion` - we
1. are clear that these are the same things
2. can take advantage of property shorthands

packages/auth/src/error/utils/assertValidationError.ts
```diff
@@ -0,0 +1,66 @@
...
+    [AuthValidationErrorCode.EmptySignUpUsername]: {
+        message: "Username is required to signUp",
+        recovery: "Make sure that a valid username is passed for signUp"
+    }
+}
+
+export function assertValidationError(assertion:boolean, name:  AuthValidationErrorCode): asserts assertion{
+
+    const message = validationErrorMap[name].message
+    const recoverySuggestion  = validationErrorMap[name].recovery
```
cshfang, line 58: I had a comment earlier about aligning `recovery` that would allow us to be more concise here imo
```suggestion
    const { message, recoverySuggestion } = validationErrorMap[name]
```

packages/auth/src/providers/cognito/apis/signUp.ts
```diff
@@ -31,56 +35,68 @@ import { signUpClient } from '../utils/clients/SignUpClient';
 export async function signUp(
 	signUpRequest: SignUpRequest<CognitoUserAttributeKey, CognitoSignUpOptions>
 ): Promise<AuthSignUpResult<AuthStandardAttributeKey | CustomAttribute>> {
+    const username = signUpRequest.username
+	const password = signUpRequest.password
```
cshfang, line 39: Seems like this can just be destructured as well
```suggestion
    const { username, password } = signUpRequest
```

packages/auth/src/providers/cognito/apis/signUp.ts
```diff
@@ -31,56 +35,68 @@ import { signUpClient } from '../utils/clients/SignUpClient';
...
+	assertValidationError(!!username, AuthValidationErrorCode.EmptySignUpUsername)
+	assertValidationError(!!password, AuthValidationErrorCode.EmptySignUpPassword)
 	// TODO: implement autoSignIn
 	let validationData: AttributeType[] | undefined;
-	const _config = Amplify.config;
+	const _config = Amplify.config
```
cshfang, line 44: Nit: Imo, we should avoid underscore prefixed variables with functional JS but that may be out of scope of this PR since it's meant to just add error handling

cshfang (commented):
Had some minor suggestions/comments. But a lot of files seem like they didn't run through prettier? Missing semi-colons, whitespaces are off etc.

packages/auth/src/error/types/validation.ts
```diff
@@ -0,0 +1,66 @@
...
+	[AuthValidationErrorCode.EmptyConfirmResetPasswordUsername]: {
+		message: 'username is required to confirmResetPassword';
+		recovery: 'Make sure that a valid username is passed for confirmResetPassword';
+	};
+	[AuthValidationErrorCode.EmptyConfirmSignUpCode]: {
+		message: 'code is required to confirmSignUp';
+		recovery: 'Make sure that a valid code is passed for confirmSignUp';
+	};
+	[AuthValidationErrorCode.EmptyConfirmSignUpUsername]: {
+		message: 'Username is required to confirmSignUp';
```
jimblanc, line 31:
```suggestion
		message: 'username is required to confirmSignUp';
```

Just to make it similar to all the other ones (same case), few more below

packages/auth/src/error/types/validation.ts
```diff
@@ -0,0 +1,66 @@
...
+	[AuthValidationErrorCode.EmptyConfirmResetPasswordConfirmationCode]: {
+		message: 'confirmationCode is required to confirmResetPassword';
+		recovery: 'Make sure that a valid confirmationCode is passed for confirmResetPassword';
+	};
+	[AuthValidationErrorCode.EmptyConfirmResetPasswordNewPassword]: {
+		message: 'newPassword is required to confirmResetPassword';
+		recovery: 'Make sure that a valid newPassword is passed for confirmResetPassword';
+	};
+	[AuthValidationErrorCode.EmptyResendSignUpCodeUsername]: {
+		message: 'Username is required to confirmSignUp';
```
jimblanc, line 43:
```suggestion
		message: 'username is required to confirmSignUp';
```

packages/auth/src/error/types/validation.ts
```diff
@@ -0,0 +1,66 @@
...
+		recovery: 'Make sure that a valid username is passed for confirmSignUp';
+	};
+	[AuthValidationErrorCode.EmptyResetPasswordUsername]: {
+		message: 'username is required to resetPassword';
+		recovery: 'Make sure that a valid username is passed for resetPassword';
+	};
+	[AuthValidationErrorCode.EmptySignInPassword]: {
+		message: 'Password is required to signIn';
```
jimblanc, line 51:
```suggestion
		message: 'password is required to signIn';
```

packages/auth/src/error/types/validation.ts
```diff
@@ -0,0 +1,66 @@
...
+		recovery: 'Make sure that a valid password is passed during sigIn';
+	};
+	[AuthValidationErrorCode.EmptySignInUsername]: {
+		message: 'Username is required to signIn';
```
jimblanc, line 55:
```suggestion
		message: 'username is required to signIn';
```

packages/auth/src/error/types/validation.ts
```diff
@@ -0,0 +1,66 @@
...
+		recovery: 'Make sure that a valid username is passed during sigIn';
+	};
+	[AuthValidationErrorCode.EmptySignUpPassword]: {
+		message: 'Password is required to signUp';
```
jimblanc, line 59:
```suggestion
		message: 'password is required to signUp';
```

packages/auth/src/error/types/validation.ts
```diff
@@ -0,0 +1,66 @@
...
+		recovery: 'Make sure that a valid password is passed for signUp';
+	};
+	[AuthValidationErrorCode.EmptySignUpUsername]: {
+		message: 'Username is required to signUp';
```
jimblanc, line 63:
```suggestion
		message: 'username is required to signUp';
```

packages/auth/src/error/utils/assertValidationError.ts
jimblanc, line 4: Is there a way to do this without the duplication in `AuthValidationErrorMap`? It may be difficult to maintain. Also why define the runtime error messages in the TS type at all?

packages/auth/src/error/types/validation.ts
```diff
@@ -0,0 +1,66 @@
...
+		recovery: 'Make sure that a valid username is passed for signUp';
+	};
+};
```
jimblanc, line 66:
```suggestion
};

```

packages/auth/src/error/utils/assertServiceError.ts
```diff
@@ -0,0 +1,23 @@
...
+		!((error as ServiceError).name ||
+		!(error as ServiceError).message)
+	) {
+		throw new AuthError({
+			name: 'UnknownError',
+			message: 'An unknown error has ocurred.',
+			underlyingException: error,
+		});
+	}
+}
```
jimblanc, line 23:
```suggestion
}

```

packages/auth/src/error/utils/assertValidationError.ts
```diff
@@ -0,0 +1,66 @@
...
+        recovery: "Make sure that a valid password is passed for signUp"
+    },
+    [AuthValidationErrorCode.EmptySignUpUsername]: {
+        message: "Username is required to signUp",
+        recovery: "Make sure that a valid username is passed for signUp"
+    }
+}
+
+export function assertValidationError(assertion:boolean, name:  AuthValidationErrorCode): asserts assertion{
+
```
jimblanc, line 56:
```suggestion
```

packages/auth/src/error/utils/assertValidationError.ts
```diff
@@ -0,0 +1,66 @@
...
+    const message = validationErrorMap[name].message
+    const recoverySuggestion  = validationErrorMap[name].recovery
+    if(!assertion){
+        throw new AuthError({name, message, recoverySuggestion})
+    }
+
```
jimblanc, line 62:
```suggestion
```

packages/auth/src/error/utils/assertValidationError.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+
```
jimblanc, line 65:
```suggestion
```

packages/auth/src/error/AuthError.ts
jimblanc, line 1: We'll need the copyright header on all these files:

```
```

Also should we call this directory `errors` instead of `error` (to match the other folders in `auth`, e.g. `utils`, `types`, `providers`, etc).

packages/auth/src/providers/cognito/apis/signUp.ts
```diff
@@ -20,6 +20,10 @@ import {
 	ValidationData,
 } from '..';
 import { signUpClient } from '../utils/clients/SignUpClient';
+import { assertServiceError } from '../../../error/utils/assertServiceError';
+import { AuthError } from '../../../error/AuthError';
+import { assertValidationError } from '../../../error/utils/assertValidationError';
+import { AuthValidationErrorCode } from '../../../error/types/validation';
 
 /**
  * Creates a user
```
jimblanc, line 29: Should we also document the failure modes here with `@throws`?

packages/auth/src/providers/cognito/apis/signUp.ts
```diff
@@ -31,56 +35,68 @@ import { signUpClient } from '../utils/clients/SignUpClient';
...
-			nextStep: {
-				signUpStep: AuthSignUpStep.DONE,
-			},
+	try {
+		const res: SignUpCommandOutput = await signUpClient({
+			Username: signUpRequest.username,
+			Password: signUpRequest.password,
+			UserAttributes: signUpRequest.options?.userAttributes.map(el => {
+				return { Name: el.userAttributeKey.toString(), Value: el.value };
+			}),
```
jimblanc, line 58:
```suggestion
			UserAttributes: signUpRequest.options?.userAttributes.map(el => ({ Name: el.userAttributeKey.toString(), Value: el.value });
			}),
```

packages/auth/src/providers/cognito/apis/signUp.ts
```diff
@@ -31,56 +35,68 @@ import { signUpClient } from '../utils/clients/SignUpClient';
...
+					},
+				},
+			};
+		}
+	} catch (error) {
+		assertServiceError(error)
+		throw new AuthError({name: error.name, message:error.message})
 	}
+
+
```
jimblanc, line 99:
```suggestion
```

packages/auth/src/providers/cognito/apis/signUp.ts
```diff
@@ -31,56 +35,68 @@ import { signUpClient } from '../utils/clients/SignUpClient';
...
+						attributeName: AttributeName
+							? (AttributeName as AuthStandardAttributeKey)
+							: undefined,
+					},
+				},
+			};
+		}
+	} catch (error) {
+		assertServiceError(error)
+		throw new AuthError({name: error.name, message:error.message})
```
jimblanc, line 96: If it's not an unknown service error, wouldn't that imply error is an `AuthError` that can be thrown directly? 

I wonder if it'd be a little cleaner to just check if the type of error is `AuthError`, if so re-throw it, otherwise throw an "unknown" AuthError.

packages/auth/src/error/utils/assertServiceError.ts
helgabalashova: I think it should be `&&` instead of `||` here

packages/auth/src/providers/cognito/apis/signUp.ts
```diff
@@ -31,56 +35,68 @@ import { signUpClient } from '../utils/clients/SignUpClient';
...
-
-	if (UserConfirmed === true) {
-		return {
-			isSignUpComplete: true,
-			nextStep: {
-				signUpStep: AuthSignUpStep.DONE,
-			},
+	try {
+		const res: SignUpCommandOutput = await signUpClient({
+			Username: signUpRequest.username,
```
helgabalashova, line 54: since you assigned username and password to variables, you can just use them here
```Username: username,```

File: packages/auth/src/providers/cognito/apis/signUp.ts [hunk 24]
```diff
@@ -31,56 +35,68 @@ import { signUpClient } from '../utils/clients/SignUpClient';
...
-				codeDeliveryDetails: {
-					deliveryMedium: DeliveryMedium
-						? (DeliveryMedium as DeliveryMedium)
-						: undefined,
-					destination: Destination ? (Destination as string) : undefined,
-					attributeName: AttributeName
-						? (AttributeName as AuthStandardAttributeKey)
-						: undefined,
+	
+		if (UserConfirmed === true) {
```
helgabalashova, line 70: ``` if (UserConfirmed) {```

packages/auth/src/error/AuthError.ts
```diff
@@ -0,0 +1,25 @@
+type AuthErrorConstructor = {
+	message: string;
+	underlyingException?: Error | unknown;
```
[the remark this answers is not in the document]
israx: they will pretty much use the `name` property that includes de actual service code

packages/auth/src/error/utils/assertValidationError.ts
israx: This is the value map that is used to to map validation errors into AuthError

packages/auth/src/error/AuthError.ts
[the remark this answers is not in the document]
israx: Yeah the idea of having and `underlyingError` was to catch service errors that didn't conform part of `{name:string, message:"string"}` syntax

packages/auth/src/error/types/validation.ts
```diff
@@ -0,0 +1,66 @@
...
+	EmptyConfirmSignUpCode = 'EmptyConfirmSignUpCode',
+	EmptyResendSignUpCodeUsername = 'EmptyresendSignUpCodeUsername',
+	EmptyChallengeResponse = 'EmptyChallengeResponse',
+	EmptyConfirmResetPasswordUsername = 'EmptyConfirmResetPasswordUsername',
+	EmptyConfirmResetPasswordNewPassword = 'EmptyConfirmResetPasswordNewPassword',
+	EmptyConfirmResetPasswordConfirmationCode = 'EmptyConfirmResetPasswordConfirmationCode',
+	EmptyResetPasswordUsername = 'EmptyResetPasswordUsername',
+}
+
+export type AuthValidationErrorMap = {
```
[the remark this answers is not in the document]
israx: yeah I do agree, recovery suggestion doesn't bring additional context comparing with `message`. The idea was to align the same message and recovery from swift but I see the point of not including it.

packages/auth/src/error/utils/assertValidationError.ts
israx: yeah I agree, there's no need to have a `AuthValidationErrorMap` and run time error messages can be moved to a `constants` module

packages/auth/src/error/AuthError.ts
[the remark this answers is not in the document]
cshfang: What is this property used for then?

packages/auth/src/error/AuthError.ts
```diff
@@ -0,0 +1,25 @@
...
+	recoverySuggestion?: string;
+	name: string;
+};
+
+export class AuthError extends Error {
```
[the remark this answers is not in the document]
cshfang: I'm wary that by keeping things simple we may accidentally lock ourselves into a pattern that is hard to walk back on. E.g. if Auth errors are already extending Error directly - how would an Amplify error even fit in in the future? If we have a clear vision on what errors should look like here, why do we think it's not applicable broadly to Amplify as a whole? If we do think it is applicable then why not just add it to core now?

packages/auth/src/error/utils/assertValidationError.ts
cshfang: Is there any way we can just type the `AuthValidationErrorMap` by keying off of this so that we don't have multiple sources of truth?

packages/auth/src/error/AuthError.ts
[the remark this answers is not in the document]
israx: This property is used to reference the error coming from the service 

```
 if(error.name === "any service error"){
   // customer logic here
 }

```

packages/auth/src/error/AuthError.ts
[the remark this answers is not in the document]
israx: I added an AmplifyError.  You're right, adding it now make sense, so the other categories can extend it eventually

packages/auth/src/error/utils/assertValidationError.ts
israx: I added a generic ErrorMap

packages/auth/src/errors/AuthError.ts
```diff
@@ -0,0 +1,23 @@
...
+	 *
+	 */
+	constructor({
+		message,
+		name,
+		recoverySuggestion,
+		underlyingError,
+	}: ErrorParams) {
+		super({ message, name, recoverySuggestion, underlyingError });
+	}
```
cshfang, line 22: Since this is just a naive extension of `AmplifyError` you can probably just do this for now and let the super handle everything else? Maybe even the documentation above wouldn't need to be repeated

```suggestion
	constructor(params: ErrorParams) {
		super(params);
	}
```

packages/auth/src/errors/types/models.ts
```diff
@@ -0,0 +1,18 @@
...
+	[name in ErrorCode]: {
+		message: string;
+		recoverySuggestion?: string;
+	};
+};
+
+export type ServiceError = {
+	name: string;
+	message: string;
+};
```
cshfang, line 18: Should these also move to become AmplifyError things? I think you are setting a very good pattern for how we can handle errors across the product and I think we'd benefit from it being generalized.

packages/auth/src/common/AuthErrorStrings.ts
```diff
@@ -1,17 +1,41 @@
...
-	EMPTY_CHALLENGE = 'Challenge response cannot be empty',
-	NO_USER_SESSION = 'Failed to get the session because the user is empty',
-	NETWORK_ERROR = 'Network Error',
-	DEVICE_CONFIG = 'Device tracking has not been configured in this User Pool',
-	AUTOSIGNIN_ERROR = 'Please use your credentials to sign in',
-}
+import { AuthErrorMap } from '../errors/types/models';
+import { AuthValidationErrorCode } from '../errors/types/validation';
+
+export const validationErrorMap: AuthErrorMap<AuthValidationErrorCode> = {
```
cshfang, line 4: ✨

packages/auth/src/errors/utils/assertServiceError.ts
```diff
@@ -0,0 +1,17 @@
+import { AuthError } from '../AuthError';
+import { ServiceError } from '../types/models';
+
+export function assertServiceError(
+	error: unknown
+): asserts error is ServiceError {
+	if (
+		!error ||
+		!((error as ServiceError).name || !(error as ServiceError).message)
+	) {
```
cshfang, line 10: I just want to confirm that we are asserting that if an error does not have *both* name and message then it will be "unknown"

Is there any risk here that we might swallow information on some error that only has one or the other?

cshfang (commented):
Still looks like we're missing the license headers on some of these files. Have a couple questions but mostly I just wanted to suggest that we continue to abstract AuthError and its related types as AmplifyError because I think this looks really promising as a way to be consistent with our error handling across the board.

jimblanc (commented):
Looks pretty good! Few nits below

packages/auth/src/providers/cognito/apis/signUp.ts
israx: clarified internally

packages/auth/src/errors/utils/assertServiceError.ts
israx: addressed comment and added unit tests as well

cshfang (commented):
Just a few more comments from me. Seems like maybe some of the refactoring may have created some duplication?

packages/auth/src/Errors.ts
```diff
@@ -0,0 +1,109 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { AuthErrorMessages, AuthErrorTypes } from './types';
+import { ConsoleLogger as Logger } from '@aws-amplify/core';
+import { AuthErrorStrings } from './common/AuthErrorStrings';
+
+const logger = new Logger('AuthError');
+
+export class AuthError extends Error {
```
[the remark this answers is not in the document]
israx: yeah that code is for reference only and to avoid the build from failing. The right AuthError is coming from `src/errors`

packages/auth/src/common/AuthErrorStrings.ts
```diff
@@ -39,3 +42,23 @@ export const validationErrorMap: AuthErrorMap<AuthValidationErrorCode> = {
 		message: 'username is required to signUp',
 	},
 };
+
+export const UnknownError = 'UnknownError';
+
+export enum AuthErrorStrings {
```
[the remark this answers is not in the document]
israx: yeah this is an old enum. I had to revert some changes in order the build to work.

packages/auth/__tests__/providers/cognito/signUp.test.ts
```diff
@@ -99,7 +99,7 @@ describe('SignUp API Error Path Cases:', () => {
 		}
 	});
 
-	test('SignUp API should expect an unknown error when the underlying error is not coming from the service', async () => {
+	test('SignUp API should expect an unknown error when underlying error is not comming from the service', async () => {
```
cshfang, line 102: Typo
```suggestion
	test('SignUp API should expect an unknown error when underlying error is not coming from the service', async () => {
```

packages/auth/src/errors/utils/assertServiceError.ts
```diff
@@ -14,7 +13,7 @@ export function assertServiceError(
 		!((error as ServiceError).name && (error as ServiceError).message)
 	) {
 		throw new AuthError({
-			name: UnknownError,
+			name: AmplifyErrorString.UNKNOWN,
```
cshfang, line 16: Nit: Now that I'm seeing it used here would it make sense to call it `AmplifyErrorName`?

cshfang (dismissed):
I know there were some comments from others still but changes lgtm thanks!

jimblanc (approved):
Approving this change. Change to bump up bundle size limits will come in a subsequent PR since Israel is unavailable.

cshfang (approved):

Issue #2144, opened 2018-11-18
Storage.get returns a presigned URL (rather than throws an error) for a non-exist key


YikSanChan:
**Describe the bug**
`Storage.get(non_exist_key)` returns a pre-signed URL rather than throwing an error.

**To Reproduce**
Same behavior as in https://github.com/aws-amplify/amplify-js/issues/218#issuecomment-431738070. I want to be able to pick/upload/preview an image, so I use `<S3Image picker imgKey="example.png" />`. However, a "broken paper" is rendered atop the picker.

1. Include `<S3Image picker imgKey="example.png" />` in `render()` of a react component.
2. Refresh the page.
3. From chrome dev tool console, `GET <a very long pre-signed url> 404 (Not Found)`
3. `cmd + p`, search `Storage.js` to debug in chrome dev tools, the very long pre-signed url is exactly the same URL generated at [Storage.ts#L127](https://github.com/aws-amplify/amplify-js/blob/4644b432/packages/storage/src/Storage.ts#L127).

**Expected behavior**
`Storage.get` should be able to return an error on such "failure", i.e., key doesn't exist.

**Desktop (please complete the following information):**
 - OS: macOS (v10.13.5)
 - Browser: Chrome (v70)

**Additional context**
The root cause is [s3.getSignedUrl](https://github.com/aws-amplify/amplify-js/blob/4644b432/packages/storage/src/Storage.ts#L127) returns a pre-signed URL even if the key doesn't exist in S3. (You can find https://github.com/aws/aws-sdk-js/issues/1529 for more context).

If there exists (and it does, e.g., https://stackoverflow.com/a/26950546/7550592) a way to check whether a pre-signed URL really points to an existing resource, let name if `checkExistence: string => boolean`, a few solutions I can think of:
- Inside `Storage.js`, after receiving pre-signed URL from `s3.getSignedUrl`, then `checkExistence`. If false, throw an error.
- Provide a `onError` props to `<S3Image />` and call `checkExistence` inside the `onError` props.

Which approach is preferred? I prefer the 1st solution so that app developers
- Don't need to dive into the `aws-amplify-react` etc source code to understand what actually happens.
- Don't need to include `aws-sdk-js` dependency in their app.
- Benefit from framework-level-consistent error handling.

***You can turn on the debug mode to provide more info for us by setting window.LOG_LEVEL = 'DEBUG'; in your app.***

powerful23:
[...]

Issue #1145, opened 2018-06-29
Storage.get('file') should throw an error if file does not exist


benevolentprof:
**Do you want to request a feature or report a bug?**
Feature

**What is the current behavior?**

Storage.get('file') resolves to a presigned URL, whether or not the file exists.

This is a problem because the URL results in a 404, which means an extra step is needed to check the validity of the URL. 

It's especially a problem in Angular, when using constructs such as the following in the ionic aws starter project.

account.ts, lines 42-45
```
  refreshAvatar() {
    Storage.get(this.userId + '/avatar')
      .then(url => this.avatarPhoto = (url as string));
  }
```

account.html, line 12
`    <div *ngIf="avatarPhoto" class="avatar" [style.background-image]="'url('+ avatarPhoto +')'">
`

**What is the expected behavior?**

Storage.get('file') should return an error that can be handled using `catch` or `if (error)`.

The [current documentation](https://aws.github.io/aws-amplify/api/classes/storageclass.html#get) doesn't specify what is returned on failure, so this change won't be a problem there. 

However, this would be a breaking change for code that relies on the current behaviour. One option is to simply accept the breaking change. Another option is to add a third parameter to the method, for the sake of discussion, called "safe," to request this new behaviour.

**Which versions of Amplify, and which browser / OS are affected by this issue? Did this work in previous versions?**

0.4.6

powerful23:
Related to #1135

iqbalt:
Dear,
Any update on this? Does this fix?

kkemple:
I'm currently working with someone who experiencing this issue as well. The problem is that they can't call download because the files are very large and that would be a terrible user experience, and doing `Storage.list()` seems like a bad workaround as well. As file number increases it could require looking through 100s or 1000s of items to see if it exists. Is there really no better way to tell if the resource actually exists using `Storage.get`?

russeg:
Return null or throw, but not url. Yes technically returning a 404 url is correct, but it does not make sense. The api should be practical and right now it is not.

ericclemmons:
It looks like the problem is that `Storage.get` isn't getting an object, but getting a signed URL to the (potentially missing) object:
> https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#getSignedUrlPromise-property

[...]

PR #11154, opened 2023-03-29
feat(storage): add validateObjectExistence option in Get API

kvramyasri7:
#### Description of changes
Adding validateObjectExistence flag in Get API to allow check existence of the Object.

#### Issue #, if available

#1145 
#2144

Tested with a sample app and added unit test cases

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -446,7 +448,24 @@ export class AWSS3Provider implements StorageProvider {
 				throw error;
 			}
 		}
-
+		if (validateObjectExistence === true) {
```
cshfang, line 451: Since it's typed and has even has a default value above, it should be safe to just do a truthy check
```suggestion
		if (validateObjectExistence) {
```

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -446,7 +448,24 @@ export class AWSS3Provider implements StorageProvider {
...
+			try {
+				const headCommand = new HeadObjectCommand(params);
```
cshfang, line 453: Is there any chance that the current role would lack the permissions to run this? i.e. `s3:ListBucket`

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
```diff
@@ -511,6 +511,48 @@ describe('StorageProvider test', () => {
 
 			curCredSpyOn.mockClear();
 		});
+
+		test('get Existed object with validateObjectExistence option', async () => {
```
cshfang, line 515: Super nit:
```suggestion
		test('get existing object with validateObjectExistence option', async () => {
```

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
```diff
@@ -511,6 +511,48 @@ describe('StorageProvider test', () => {
...
+			expect(await storage.get('key', { validateObjectExistence: true })).toBe(
+				'url'
+			);
+			expect(spyon.mock.calls[0][0].path).toEqual('/public/key');
+			expect(spyon.mock.calls[0][0].hostname).toEqual(
+				`${options.bucket}.s3.${options.region}.amazonaws.com`
+			);
+		});
+
+		test('get UnExisted object with validateObjectExistence option', async () => {
```
cshfang, line 538: Super nit:
```suggestion
		test('get non-existent object with validateObjectExistence option', async () => {
```

cshfang (commented):
Just a few minor suggestions

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -446,7 +448,24 @@ export class AWSS3Provider implements StorageProvider {
...
+			} catch (error) {
+				dispatchStorageEvent(
+					track,
+					'getSignedUrl',
+					{
+						method: 'get',
+						result: 'failed',
+					},
+					null,
+					`Key: ${key} Not Found`
```
cshfang, line 464: Suggest not having "key" and getting rid of the title casing in the message to align with the other dispatch events
```suggestion
					`${key} not found`
```

cshfang (approved):

packages/storage/src/providers/AWSS3Provider.ts
kvramyasri7: Thanks for the suggestion . Updated.

packages/storage/src/providers/AWSS3Provider.ts
kvramyasri7: Makes sense. Thanks.

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
```diff
@@ -511,6 +511,48 @@ describe('StorageProvider test', () => {
...
+			);
+		});
+
+		test('get non-existing object with validateObjectExistence option', async () => {
+			expect.assertions(1);
+			jest.spyOn(Credentials, 'get').mockImplementationOnce(() => {
+				return new Promise((res, rej) => {
+					res({});
+				});
+			});
```
jimblanc, line 544: Nit: Can this be pulled into a `beforeEach` function for both tests? In theory should be the same I believe (even though this one doesn't return the mocked credentials).

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -387,7 +389,7 @@ export class AWSS3Provider implements StorageProvider {
 			Bucket: bucket,
 			Key: final_key,
 		};
-
+		const headObjectCommand = new HeadObjectCommand(params);
```
jimblanc, line 392: You should initialize this on line 452 instead (inside of the `validateObjectExistence` check) since it's only used for that.

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -446,7 +448,23 @@ export class AWSS3Provider implements StorageProvider {
 				throw error;
 			}
 		}
-
+		if (validateObjectExistence) {
+			try {
+				await s3.send(headObjectCommand);
+			} catch (error) {
```
jimblanc, line 454: I'm wondering, do we need to be more precise with the error handling here? For example explicitly checking that the request failed with a 404 before dispatching the event that says `${key} not found`. We'd still want to throw the error regardless of failure type, but just to make sure our eventing is accurate.

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -446,7 +448,23 @@ export class AWSS3Provider implements StorageProvider {
...
+						method: 'get',
+						result: 'failed',
+					},
+					null,
+					`${key} not found`
+				);
+				throw error;
+			}
+		}
 		try {
```
jimblanc, line 468:
```suggestion
		}
		
		try {
```

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
```diff
@@ -511,6 +511,48 @@ describe('StorageProvider test', () => {
...
+			jest
+				.spyOn(S3Client.prototype, 'send')
+				.mockImplementationOnce(async params => {
+					throw 'err';
```
jimblanc, line 548: In line with some other feedback below, we should specifically simulate a 404 situation.

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
```diff
@@ -511,6 +511,48 @@ describe('StorageProvider test', () => {
...
+				});
+			try {
+				await storage.get('key', { validateObjectExistence: true });
+			} catch (e) {
+				expect(e).toBe('err');
```
jimblanc, line 553: We should also confirm that `dispatchStorageEvent` was invoked as expected.

jimblanc (commented):
Looks good, some feedback below.

Also it appears that we'll need to bump up the bundle size limit on storage to accommodate this change: https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/18675/workflows/a66e0bf2-7cdd-4786-baca-3fcc17afd29f/jobs/251512

You can find this configuration here: https://github.com/aws-amplify/amplify-js/blob/main/packages/storage/package.json#L63

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
kvramyasri7: Thanks for the suggestion. Updated.

packages/storage/src/providers/AWSS3Provider.ts
kvramyasri7: Makes sense. Updated.

packages/storage/src/providers/AWSS3Provider.ts
kvramyasri7: Sure. Updated.

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
kvramyasri7: Thanks for the suggestion.

packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
kvramyasri7: Thanks for the feedback. Updated,

packages/storage/src/providers/AWSS3Provider.ts
```diff
@@ -446,7 +448,26 @@ export class AWSS3Provider implements StorageProvider {
 				throw error;
 			}
 		}
-
+		if (validateObjectExistence) {
+			const headObjectCommand = new HeadObjectCommand(params);
+			try {
```
jimblanc, line 453: Nit (white-space makes it a little more readable)

```suggestion
			const headObjectCommand = new HeadObjectCommand(params);
			
			try {
```

packages/storage/src/providers/AWSS3Provider.ts
jimblanc, line 392: It seems like this line is still present in latest diff (duplicated with line 452).

```suggestion
```

jimblanc (commented):
LGTM! Some small comments below. We should also probably add an integration test for this behavior.

jimblanc (dismissed):
Looks good!

cshfang (approved):

packages/predictions/package.json
```diff
@@ -62,7 +62,7 @@
 			"name": "Predictions (all providers)",
 			"path": "./lib-esm/index.js",
 			"import": "{ Amplify, Predictions, AmazonAIPredictionsProvider }",
-			"limit": "133 kB"
+			"limit": "133.5 kB"
```
haverchuck, line 65: Why are we touching predictions?

haverchuck (commented):
Overall looks good - left one question

packages/predictions/package.json
jimblanc: I wonder if this is related to the fast-xml-parser `change` in main which did change those limits?

kvramyasri7: I believe there is no dependency from predictions and storage. Surprisingly, the build failed when merged with main. This prediction package.json change is inline with the fast-xml-parser change. Please let me know if we have any concerns. Thanks.

jimblanc: Hum it looks like the fast-xml-parser change was the one that bumped it up to 133 from 131.5. I'm still confused why this is increasing in your PR, let me see if I can debug.

kvramyasri7: FYI, I have updated the limit as the previous build is failed due to merging into main.

jimblanc: Okay I dug into this some more, it turns out one of the predictions providers does have a dependency on `Storage` for the purpose of importing source objects from S3 into Predictions, see: https://github.com/aws-amplify/amplify-js/blob/main/packages/predictions/src/Providers/AmazonAIIdentifyPredictionsProvider.ts#L81

Bumping this limit will be unavoidable in this PR unfortunately.

jimblanc (approved):

haverchuck (approved):

stocaaro (approved):

PR #11155, opened 2023-03-29
fix(notifications): Fix double registration issue

cshfang:
#### Description of changes

This PR fixes an issue where, upon a fresh install, the Android module will send off two token received events (from the service onNewToken and module launch). This triggers a race condition where both events are handled in parallel and ends up creating two separate endpoints.

Tested locally with sample app and verified only one endpoint is created on fresh install

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


stocaaro (approved):
Seems like the type of fix that deserves a testing story to ensure we don't fall back into the bad state across a refactor.

stocaaro (approved):

Issue #10864, opened 2023-01-12
[DataStore] Failed to execute 'index' error when deleting a record with `@hasMany` relationships


chrisbonifacio:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

DataStore

### Environment information

<details>

```
# Put output below this line


```

</details>

### Describe the bug

Getting an error from DataStore when trying to delete a record (with associated @hasMany related objects)
Model:

```graphql
type CutsheetRequirement @model @auth(rules: [{ allow: private, provider: userPools }, { allow: public, provider: apiKey }]) {
  cutsheetRequirementId: ID! @primaryKey
  name: String!
  ...
  CutsheetOptions: [CutsheetOption] @hasMany(indexName: "byCutsheetRequirement", fields: ["cutsheetRequirementId"])
  CutsheetImpacts: [CutsheetImpact] @hasMany(indexName: "byCutsheetRequirement", fields: ["cutsheetRequirementId"])
  ...
}

type CutsheetOption @model @auth(rules: [{ allow: private, provider: userPools }, { allow: public, provider: apiKey }]) {
  cutsheetOptionId: ID! @primaryKey
  name: String!
  cutsheetRequirementID: ID! @index(name: "byCutsheetRequirement", queryField: "listCutsheetOptionsByCutsheetRequirementId", sortKeyFields: ["cutsheetOptionId"])
  sort: Int!
  CutsheetImpacts: [CutsheetImpact] @hasMany(indexName: "byCutsheetOption", fields: ["cutsheetOptionId"])
  CutsheetInstructions: [CutsheetInstruction] @hasMany(indexName: "byCutsheetOption", fields: ["cutsheetOptionId"])
  updatedBy: String!
  updatedSource: String!
}

type CutsheetInstruction @model @auth(rules: [{ allow: private, provider: userPools }, { allow: public, provider: apiKey }]) {
  cutsheetInstructionId: ID! @primaryKey
  name: String!
[...]

PR #11160, opened 2023-03-30
fix(datastore): normalize deleteTraverse 1:M

iartemiev:
#### Description of changes

Before the change in this PR our business logical for performing cascading deletes only worked correctly when we could retrieve child relations using an index. This worked in most cases, but was not flexible enough, e.g. for the schema in the linked issue where the child index does not map back to the parent due to a different number of key fields in the index.

This PR leverages existing predicate logic in the storage adapters to attempt to use an index for retrieval when possible, but fall back to a scan when an appropriate index does not exist.

#### Issue #, if available

#10864 

* red/green unit test with schema from linked issue
* manual testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/src/util.ts
+++ b/packages/datastore/src/util.ts
@@ -1043,6 +1043,86 @@ export const getIndexFromAssociation = (
 };
 
 /**
+ * Retrieve child foreign key field names from parent's associatedWith value.
+ * Normalizes all possible permutations (legacy codegen, pre-cpk, bi-directional, uni-directional)
+ *
+ * @param associatedWith - parent model's `associatedWith` value.
+ * Can contain the actual FK field(s) or a model field (when relationship is bi-directional)
+ * @param modelDefinition - child model definition
+ *
+ * @returns foreign key fields on the child model used to form the relationship with the parent
+ *
+ */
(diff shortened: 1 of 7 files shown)


packages/datastore/src/util.ts
```diff
@@ -1042,6 +1042,86 @@ export const getIndexFromAssociation = (
...
+	});
+
+	const predicateGroup: PredicatesGroup<any> = {
+		type: 'and',
+		predicates: keyPredicates,
+	};
+
+	return predicateGroup;
+};
+
```
svidgen, line 1124: Should this live with the other predicate creation utils functions in [predicates/index.ts](https://github.com/aws-amplify/amplify-js/blob/29513118e3e92511e65a9670a830f3e1f4644b23/packages/datastore/src/predicates/index.ts)? Do any of the functions in there actually already help with this?

svidgen (approved):
Works for me!

I think there's still a lot of refactor potential here to look at in the future. I could be wrong, but I get the sense that this delete traversal stuff is still contains a lot of LOC that duplicate logic from elsewhere.

dpilch (approved):

PR #11179, opened 2023-04-03
Datastore/feat add per model sync page size

MixMasterT:
#### Description of changes
This is a DRAFT PR for implementing perModelSyncPageSize.

Currently, it is possible to set the syncPageSize value, but here is no way to set it to a different value for any one specific data model. This proposed solution would allow users to provide a perModelSyncPageSize option in their DataStore config to set different syncPageSize values for different data models.

The perModelSyncPageSize option would be completely optional, so most users could ignore it altogether.
However, this will be very helpful to developers who may have a heavy data model, since the default syncPageSize is 1000 and the max data volume allowed per request over appSync is 1mb.

So far the only validation has been to run the tests (with the new tests commented out), and verify that they still all pass.

NOTE: this is a draft PR, and the goal is to confirm the approach. If this approach is agreed upon, I will follow up with test implementation to verify the intended behavior.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/sync.test.ts
+++ b/packages/datastore/__tests__/sync.test.ts
@@ -299,6 +299,55 @@ describe('Sync', () => {
 		});
 	});
 
+	/*
+	 *  This is where new tests would be added for per-model syncPageSize
+	 *  Proposed test cases are added below with // comments
+	 */
+
+	// describe('syncPageSize', () => {
+	// 	test('should use default syncPageSize if not specified (null, undefined or zero)', async () => {
+	// 		const resolveResponse = {
+	// 			data: {
+	// 				syncPosts: {
+	// 					items: [
+	// 						{
+	// 							id: '1',
+	// 							title: 'Item 1',
+	// 						},
+	// 						{
+	// 							id: '2',
+	// 							title: 'Item 2',
+	// 						},
+	// 					],
+	// 				},
+	// 			},
+	// 		};
+
+	// 		const SyncProcessor = jitteredRetrySyncProcessorSetup({
+	// 			resolveResponse,
+	// 		});
+
+	// 		await SyncProcessor.jitteredRetry({
+	// 			query: defaultQuery,
+	// 			variables: defaultVariables,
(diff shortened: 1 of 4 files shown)


renebrandel:
Hi @MixMasterT  - thanks for starting this PR. Could you tell us a bit more about your use case that requires this level of granularity? 

>However, this will be very helpful to developers who may have a heavy data model, since the default syncPageSize is 1000 and the max data volume allowed per request over appSync is 1mb.

Would love it if you could expand more on this. For example with specific limitations that you might've ran into.

MixMasterT:
> Hi @MixMasterT - thanks for starting this PR. Could you tell us a bit more about your use case that requires this level of granularity?
> 
> > However, this will be very helpful to developers who may have a heavy data model, since the default syncPageSize is 1000 and the max data volume allowed per request over appSync is 1mb.
> 
> Would love it if you could expand more on this. For example with specific limitations that you might've ran into.

No problem.

Her is an example:
Let's say I'm creating an app for Chemistry research experiments. I have five data models (DynamoDB Tables): Users, Materials, Experiments, ResearchFacilities, and ResearchTeams.
Users, Materials, and ResearchFacilities are fairly standard, smallish data models (< 1kb per record). ResearchTeams are basically just groupings of Users, but let's say that we use a sync-expression to share Experiments between all Users in the same ResearchTeam, and that some Users may belong to multiple ResearchTeams.
Now, suppose that the Experiments Table is quite complex and includes a lot of detailed observation data, so a single Experiment may be over 100kb in data.
If we have a User who has more than 20 experiments shared with her/him, that may cause AppSync to fail with the "Transformation too large" error.
Currently, with only a single `syncPageSize` to set, we could solve this problem by setting that down to a very low number (like 5), but if we still want to share all of the Materials with all of the Users, this may cause loading to be painfully slow.
`perModelSyncPageSize` solves this problem by allowing the developers to specify a lower syncPageSize only for those models that are bulky.

Hopefully this description is helpful.

Please let me know if it makes sense to you. Thanks!

PR #11183, opened 2023-04-04
Adds AmplifyUserSessionProvider

elorzafe:
#### Description of changes

Adds `AmplifyUserSessionProvider` to Singleton 

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -94,7 +94,7 @@ const MAX_AUTOSIGNIN_POLLING_MS = 3 * 60 * 1000;
  * Provide authentication steps
  */
 export class AuthClass {
-	private _config: AuthOptions;
+	private _initialAuthOptions: AuthOptions;
 	private userPool: CognitoUserPool = null;
 	private user: any = null;
 	private _oAuthHandler: OAuth;
@@ -112,6 +112,7 @@ export class AuthClass {
 	 * @param {Object} config - Configuration of the Auth
 	 */
 	constructor(config: AuthOptions) {
+		this._initialAuthOptions = config;
 		this.configure(config);
 		this.currentCredentials = this.currentCredentials.bind(this);
 		this.currentUserCredentials = this.currentUserCredentials.bind(this);
@@ -133,6 +134,18 @@ export class AuthClass {
 		});
 	}
 
+	private get _config(): AuthOptions {
+		const config = Amplify.config;
+		const conf = Object.assign(
+			{},
+			this._initialAuthOptions,
+			parseAWSExports(config).Auth,
+			config
+		);
+
+		return conf;
+	}
+
 	public getModuleName() {
 		return 'Auth';
 	}
@@ -146,7 +159,7 @@ export class AuthClass {
 			parseAWSExports(config).Auth,
 			config
 		);
-		this._config = conf;
+		this._initialAuthOptions = conf;
 		const {
 			userPoolId,
 			userPoolWebClientId,
@@ -2780,5 +2793,3 @@ export class AuthClass {
 }
 
 export const Auth = new AuthClass(null);
-
-Amplify.register(Auth);
--- a/packages/core/src/Amplify.ts
+++ b/packages/core/src/Amplify.ts
@@ -1,4 +1,5 @@
 import { ConsoleLogger as LoggerClass } from './Logger';
+import { AmplifyUserSessionProvider } from './types';
 
 const logger = new LoggerClass('Amplify');
 
@@ -10,9 +11,7 @@ export class AmplifyClass {
 	// All modules (with `getModuleName()`) are stored here for dependency injection
 	private _modules = {};
 
-	// for backward compatibility to avoid breaking change
-	// if someone is using like Amplify.Auth
-	Auth = null;
+	private _Auth: AmplifyUserSessionProvider;
 	Analytics = null;
 	API = null;
 	Credentials = null;
@@ -37,6 +36,10 @@ export class AmplifyClass {
 		return Object.assign({}, this._config);
 	}
 
+	public get Auth() {
+		return this._Auth;
+	}
+
 	register(comp) {
 		logger.debug('component registered in amplify', comp);
 		this._components.push(comp);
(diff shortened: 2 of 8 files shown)


packages/core/src/Amplify.ts
```diff
@@ -56,8 +59,15 @@ export class AmplifyClass {
 		comp.configure(this._config);
 	}
 
-	configure(config?) {
-		if (!config) return this._config;
+	configure(
+		config,
+		libraryConfig?: { Auth: { sessionProvider: AmplifyUserSessionProvider } }
```
jimblanc, line 64: I wonder if we should merge the defaults instead of having a default parameter? E.g. if a customer specifies some other `libraryConfig`, they'd still get the default session provider.

packages/core/src/Amplify.ts
```diff
@@ -37,6 +36,10 @@ export class AmplifyClass {
 		return Object.assign({}, this._config);
 	}
 
+	public get Auth() {
+		return this._Auth;
+	}
```
jimblanc, line 41: Do we need a getter for this, can we just expose the `Auth` library config directly? Also is the plan to scope all library config behind the associated member?

packages/core/src/Amplify.ts
```diff
@@ -1,4 +1,5 @@
 import { ConsoleLogger as LoggerClass } from './Logger';
+import { AmplifyUserSessionProvider } from './types';
 
 const logger = new LoggerClass('Amplify');
```
jimblanc, line 5: Food for thought & the design, but I'm wondering do we even need `AmplifyClass` anymore? For example, we could have everything at module-scope in this module, which would behave as our singleton. This would have the advantage of allowing us to lock down the exports more tightly and further reducing our usage of classes. A downside is it would prevent initialization of multiple instances of Amplify if that ends up being a requirement.

packages/core/src/Amplify.ts
```diff
@@ -10,9 +11,7 @@ export class AmplifyClass {
 	// All modules (with `getModuleName()`) are stored here for dependency injection
 	private _modules = {};
 
-	// for backward compatibility to avoid breaking change
-	// if someone is using like Amplify.Auth
-	Auth = null;
+	private _Auth: AmplifyUserSessionProvider;
```
jimblanc, line 14: Hum this is a little bit confusing to me vs having the session provider named as such on the singleton. For example what if the `Auth` library config ends up getting expanded with another provider or configuration?

packages/core/src/types/types.ts
```diff
@@ -63,3 +64,37 @@ export interface CloudWatchDataTracker {
...
+
+export interface AmplifyUserSessionProvider {
+	getUserSession: (
+		options?: GetUserSessionOptions
+	) => Promise<AmplifyUserSession>;
+	listenUserSession(): Observable<AmplifyUserSession>;
+}
+
+// Other types just for reference
+
```
jimblanc, line 76:
```suggestion
```

(nit)

jimblanc (commented):
Looks good, some comments below!

packages/core/src/Amplify.ts
elorzafe: Is a good call, the thing there is the type that `Auth` is going to be. Maybe is more clear if I defined a type which will be the same as `AmplifyUserSessionProvider`?

packages/core/src/Amplify.ts
elorzafe: Yes, I will add a note to remove that, I was thinking on do this incrementally so still works

packages/auth/src/Auth.ts
```diff
@@ -94,7 +94,7 @@ const MAX_AUTOSIGNIN_POLLING_MS = 3 * 60 * 1000;
  * Provide authentication steps
  */
 export class AuthClass {
-	private _config: AuthOptions;
+	private _initialAuthOptions: AuthOptions;
```
haverchuck, line 97: Why are we touching the Auth class?

packages/core/src/types/types.ts
```diff
@@ -63,3 +64,37 @@ export interface CloudWatchDataTracker {
...
+export type GetUserSessionOptions = {
+	refresh?: boolean;
+};
+
+export type AmplifyUserSession = {
+	isLoggedIn: boolean;
```
haverchuck, line 82: isLoggedIn means what, exactly? JWTs are present but not expired? If so, how do we guarentee that we know they aren't expired when this is accessed? Does Identity Pool fedartion count as logged in?

Maybe instead of a bool this should be a function that could return an enum representing login status.

packages/core/src/types/types.ts
```diff
@@ -63,3 +64,37 @@ export interface CloudWatchDataTracker {
...
+	username?: string;
+	credentials?: ICredentials;
+	jwts?: JWTS;
+};
+
+export interface JWT {
```
haverchuck, line 88: Maybe this could a class with a decode method?

haverchuck (commented):
I personally think we should avoid putting JWTs and Credentials into the same type. I think this could simplify things when we support User Pools alone or Identity Pools alone - the singleton could simply throw a very simple exception when we attempt to access user pool or identity pool data when these resources aren't being used, instead of us having to supply logic to the singleton that would have to parse what was available.

packages/core/src/types/types.ts
elorzafe: Yeah, that is a good idea, I can add a TODO mentioning this will be discussed in a future PR

packages/core/src/types/types.ts
elorzafe: This is a good discussion.

maybe `isAuthenticated` rather than is `isLoggedIn` is more clear, I like the idea to validate if the tokens are expired or not. If they are but you are offline, you are in a limbo state because you can still refresh the session eventually when you are back online. We will iterate this more on the API review. and this is more like a starting point. Does it make sense to add TODO message that this will TBD when we do the API reviews?

packages/auth/src/Auth.ts
elorzafe: The thing is that category-classes will no longer be registered on the Singleton. And the `Auth` instance will be `sessionProvider` only.

packages/core/src/types/types.ts
haverchuck: Yeah, that makes sense.

elorzafe:
Close this in favor of https://github.com/aws-amplify/amplify-js/pull/11195/files

Issue #10955, opened 2023-02-09
Issue with DataStore.observeQuery change snapshots with  'or' predicate


alex-breen:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

DataStore

### Amplify Categories

Not applicable

### Environment information

<details>

```
# Put output below this line


```
  System:
    OS: Windows 10 10.0.22621
    CPU: (6) x64 Intel(R) Core(TM) i5-9600K CPU @ 3.70GHz
    Memory: 771.55 MB / 15.94 GB
  Binaries:
    Node: 14.17.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.10.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 109.0.5414.120
    Edge: Spartan (44.22621.1105.0), Chromium (109.0.1518.70)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @aws-amplify/ui-react: ^4.1.0 => 4.3.4
    @aws-amplify/ui-react-internal:  undefined ()
    @emotion/react: ^11.10.5 => 11.10.5
    @emotion/styled: ^11.10.5 => 11.10.5
    @mui/base: ^5.0.0-alpha.109 => 5.0.0-alpha.115
    @mui/icons-material: ^5.10.16 => 5.11.0
    @mui/material: ^5.10.17 => 5.11.6
    @tensorflow-models/universal-sentence-encoder: ^1.3.3 => 1.3.3
    @tensorflow/tfjs: ^4.2.0 => 4.2.0
    @testing-library/jest-dom: ^5.16.5 => 5.16.5
    @testing-library/react: ^13.4.0 => 13.4.0
    @testing-library/user-event: ^13.5.0 => 13.5.0
    aws-amplify: ^5.0.13 => 5.0.13
    compromise: ^14.8.0 => 14.8.1
    debounce: ^1.2.1 => 1.2.1
[...]

PR #11185, opened 2023-04-04
fix: query on null fields to not throw error

dpilch:
#### Description of changes

Some queries applied to models with null fields could throw a silent error. This resulted in `matches` being interpreted as false.

See https://github.com/aws-amplify/amplify-js/issues/10955 for an example.

I also discovered a potential issue with type coercion on some of the matching function. `lt`, `le`, `gt`, and `ge` will perform type coercion on the `null` values to be set to 0.

i.e. Given `foo.age = null` the predicate `foo.age.le(5)` would be `true`. I will need to test the behavior on the API category to see if this aligns. We would need to wait until v6 to make this change regardless, because some customers may be relying on this behavior. I'll add some test cases to assert the current behavior in a separate PR.

#### Issue #, if available

Resolves https://github.com/aws-amplify/amplify-js/issues/10955

Unit testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


svidgen (approved):

david-mcafee (approved):
LGTM!

PR #11207, opened 2023-04-07
feat(Auth): resetPassword API rewrite + tests

helgabalashova:
#### Description of changes
Auth resetPassword API rewrite

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/__tests__/providers/cognito/resetPassword.test.ts
```diff
@@ -0,0 +1,54 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { Amplify } from '@aws-amplify/core';
+import { ForgotPasswordCommandOutput } from '@aws-sdk/client-cognito-identity-provider';
+import { resetPassword } from '../../../src/providers/cognito';
+import * as resetPasswordClient from '../../../src/providers/cognito/utils/clients/ResetPasswordClient';
+import { authAPITestParams } from './testUtils/authApiTestParams';
+
+describe('ResetPassword API happy pathes', () => {
```
haverchuck, line 10: Nit: "pathes -> paths"

packages/auth/src/providers/cognito/apis/resetPassword.ts
```diff
@@ -0,0 +1,55 @@
...
+					deliveryMedium: codeDeliveryDetails?.DeliveryMedium as DeliveryMedium,
+					destination: codeDeliveryDetails?.Destination as string,
+					attributeName: codeDeliveryDetails?.AttributeName as AuthStandardAttributeKey,
+				},
+			},
+		};
+	} catch (error) {
+		assertServiceError(error);
+		throw new AuthError({ name: error.name, message: error.message });
+	}
```
jimblanc, line 54: I don't think we need to do it now, but looking and your & Israel's PRs it seems like this pattern would be repeated for all APIs. I wonder if it makes sense to push this down into the client itself (i.e. `UserPoolHttpClient`) and let the AuthError bubble up unless an API needed some unique error handling.

packages/auth/src/types/index.ts
```diff
@@ -17,13 +18,16 @@ export type { AuthStandardAttributeKey } from './models/AuthStandardAttributeKey
...
 export type { AuthServiceOptions } from './options/AuthServiceOptions';
 export type { AuthSignUpOptions } from './options/AuthSignUpOptions';
 
 // Requests
+export type { ResetPasswordRequest } from './requests/ResetPasswordRequest';
 export type { SignUpRequest } from './requests/SignUpRequest';
 
 // Results
 export type { AuthSignUpResult } from './results/AuthSignUpResult';
+export type { ResetPasswordResult } from './results/ResetPasswordResult';
```
jimblanc, line 33: Didn't catch this on the other PR, but I wonder if the type aliases are really required here since these are our types and the names are the same. 

```suggestion
export { AuthUserAttributeKey } from './models/AuthUserAttributeKey';
export { AuthUserAttribute } from './models/AuthUserAttribute';
export { GetAttributeKey } from './models/GetAttributeKey';
export { AuthNextResetPasswordStep } from './models/AuthNextResetPasswordStep';

// Options
export { AuthServiceOptions } from './options/AuthServiceOptions';
export { AuthSignUpOptions } from './options/AuthSignUpOptions';

// Requests
export { ResetPasswordRequest } from './requests/ResetPasswordRequest';
export { SignUpRequest } from './requests/SignUpRequest';

// Results
export { AuthSignUpResult } from './results/AuthSignUpResult';
export { ResetPasswordResult } from './results/ResetPasswordResult';
```

For some reason GitHub won't let me comment on the entire list, but the others above my suggestion can probably be changed as well.

jimblanc (approved):
Looks good! Some nits below

packages/auth/__tests__/providers/cognito/testUtils/authApiTestParams.ts
```diff
@@ -1,10 +1,38 @@
...
+		username: 'username',
+		options: {
+			serviceOptions: {
+				clientMetadata: { foo: 'bar' }
+			}
+		}
+	},
+	forgotPasswordCommandWithClientMetadata: {
+		Username: 'username',
+		ClientMetadata: {foo: 'bar'}
```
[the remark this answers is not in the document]
helgabalashova: command parameters are all CamelCase (we follow the same patter as aws-sdk)

jimblanc (dismissed):

packages/auth/src/providers/cognito/apis/resetPassword.ts
haverchuck: @jimblanc What you are saying is that instead of intercepting the missing username, we just let it go to the Cognito service and return the Cognito erorr?

I tend to agree with that. It seems nice to prevent a round trip, but trying to do 1-off validation handling like this could lead to inconsistencies.

haverchuck (changes requested):
Commented on Jim's comment.

packages/auth/src/providers/cognito/apis/resetPassword.ts
jimblanc: My suggestion was more about just moving this "common" error handling down a level or two, since this try-catch is essentially just checking for unknown service errors + wrapping any errors in an AuthError. The same logic exists in [signUp](https://github.com/aws-amplify/amplify-js/pull/11153/files#diff-8b8e0c0eaa2c2b2add12d12aae99d2583f07e8a42f9d585bd3b80aad86c0aad8R105) and I imagine it would exist in the other APIs as well by default unless we needed to do some special error handling for particular APIs.

Regarding the pre-request validation, I think I'm ambivalent. I like preventing the unneeded round trip (which might have benefits as far as avoiding cognito throttling, extra network calls, etc.), but keeping client-side validation in sync with the service is always tricky. Now that I think about it though, I wonder if we can just lean on TypeScript to enforce this type of basic validation?

haverchuck: Argh, sorry @jimblanc I think I misread your comment (or rather applied your comment to the wrong block of code). Moving the wrapping of the service error into an AuthError seems like something that could definitely go into the client abstraction layer.

@helgabalashova Could you make that change for both resetPassword and signUp, if you agree with this?

jimblanc (approved):
Looks good! Thank you for the refactor.

packages/auth/__tests__/providers/cognito/resetPassword.test.ts
```diff
@@ -0,0 +1,106 @@
...
+import { authAPITestParams } from './testUtils/authApiTestParams';
+
+describe('ResetPassword API happy path cases', () => {
+	let resetPasswordSpy;
+
+	beforeEach(() => {
+		resetPasswordSpy = jest.spyOn(resetPasswordClient, 'resetPasswordClient')
+			.mockImplementationOnce(async (params: resetPasswordClient.ResetPasswordClientInput) => {
+				return {
+					CodeDeliveryDetails: {
```
haverchuck, line 20: @helgabalashova @jimblanc How would we feel about peeling out re-used things like CodeDeliveryDetails objects into test utils for re-use?

haverchuck: We could make these changes to signUp test as well. If you think it's overkill, I'm fine with this as is.

haverchuck (approved):
Left a suggestion for the tests, but I'm fine with this PR as-is if you disagree.

PR #11262, opened 2023-04-17
fix(data): sync processor fixes

david-mcafee:
#### Description of changes

While working on consistency testing for DataStore, some of our unit tests were breaking when introducing network latency into the tests. [Related PR](https://github.com/aws-amplify/amplify-js/pull/11261).
If `runningProcesses` is not open when the sync processor starts, it means that the sync processor has been stopped (for example by calling `DataStore.clear()` upstream, as well as `DataStore.stop()`) and it has not yet finished terminating and/or waiting for its background processes to complete. In this case, we log a warning and resolve the promise. 
Additionally, it's possible that `retrievePage` will fail. If it does fail, we continue merging the rest of the data, and invoke the error handler for non-applicable data.
Added a try / catch around `retrievePage`, resolved promises were applicable, added a debug log, and added a call to the error handler.

[Successful run for all E2E tests](https://app.circleci.com/pipelines/github/aws-amplify/amplify-js/19070/workflows/ca30586e-f91c-4d58-af38-fbd18372a1b1)

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/datastore/src/sync/processors/sync.ts
```diff
@@ -357,23 +357,39 @@ class SyncProcessor {
...
 
 								do {
+									/**
+									 * It's possible that the running processes have been closed
+									 * (e.g. `DataStore.clear` has been called)
+									 */
 									if (!this.runningProcesses.isOpen) {
-										return;
+										logger.error(
+											'No open running processes when starting sync'
```
svidgen, line 366: I threw the word "error" around a bit more than I maybe should have during our call. A message here is *probably* only be useful for troubleshooting purposes by a DataStore engineer. That said, this can probably actually be a debug level log that says, something like:

> `The sync processor has been stopped. The pending retrievePage for model ${modelDefinition.name} will be skipped.`

packages/datastore/src/sync/processors/sync.ts
```diff
@@ -357,23 +357,39 @@ class SyncProcessor {
 								await Promise.all(promises);
 
 								do {
+									/**
+									 * It's possible that the running processes have been closed
```
svidgen, line 361: Maybe,

```
/**
 * If `runningProcesses` is not open, it means that the sync processor has been
 * stopped (for example by calling `DataStore.clear()` upstream) and has not yet
 * finished terminating and/or waiting for its background processes to complete.
 */

File: packages/datastore/src/sync/processors/sync.ts [hunk 3]
```diff
@@ -357,23 +357,39 @@ class SyncProcessor {
...
+										({ items, nextToken, startedAt } = await this.retrievePage(
+											modelDefinition,
+											lastSync,
+											nextToken,
+											limit,
+											filter,
+											onTerminate
+										));
+									} catch (error) {
+										logger.error('Error retrieving page:', error);
```
svidgen, line 391: For completeness, I think this should `return res()` or `return rej()` here as well.

Looking at the whole loop, I think `return res()` will happen organically with the introduction of the try-catch. But, I'd rather be very explicit.

That said, see my other comment about making sure we do the right thing here. If customers today expect DataStore to fail hard on a broken model, let's reject here and make sure that rejection is "consumed" properly. If they expect all other models to continue functioning, we should log the error and `res()` explicitly.

If I had to guess, today's behavior is a partial failure: My read on it is that DataStore will continue to operate on other models, but will be in a broken state that can't be successfully `stop()`'d or `clear()`'d. If I'm reading that correctly, I think "log the error and continue with other models" is probably the non-breaking change for customers. But, we should confirm that &mdash; maybe you could set up a test app and just add a non-existent model to the client to confirm how the client behaves *today*.

packages/datastore/src/sync/processors/sync.ts
```diff
@@ -357,23 +357,39 @@ class SyncProcessor {
...
-										nextToken,
-										limit,
-										filter,
-										onTerminate
-									));
+									/**
+									 * It's possible that retrievePage will error here, for instance,
+									 * during `authModeRetry` if there are no open running processes.
+									 * An example would be if a model does not exist in the schema.
+									 */
```
svidgen, line 380: Not "if there are no open running processes". I wouldn't necessarily even call out `authModeRetry` here. I think we more-so need to call out that:

1. `retrievePage` can fail.
2. If `retrievePage` does fail, we're going to raise/not-raise because Y.

I think we should

1. Test existing behavior against a real API to confirm today-behavior
2. Verify with @manueliglesias and/or spec what the expected behavior for a single failed model should be
3. Decide whether to `res()` or `rej()` based on that.
4. Update the comment accordingly.

dpilch (dismissed):

svidgen (changes requested):

packages/datastore/src/sync/processors/sync.ts
manueliglesias: > Verify with @manueliglesias and/or spec what the expected behavior for a single failed model should be

TL;DR:
DataStore should continue merging the rest of the data and make sure to invoke the error handler for non-applicable data..

---

Some excerpts from the spec:

> Data which is visible to client application code is referred to as “applicable data” by which we mean results from a GraphQL server which has been successfully synchronized down to client applications and they can perform DataStore operations over.

> All the data which conforms to the schema is applicable to clients to operate over (e.g. this is their “universe of data”). Any non-conforming data to the GraphQL schema “doesn’t exist” from the clients perspective.

>  When the system receives a request, it must eventually respond to that request. In this context it means we must either merge the applicable data or handle errors (or conflicts) to progress the system forward rather than being “stuck” in an error state.

> All errors resulting from non-applicable data will be passed into the errorHandler

svidgen: FWIW, errors that show here will not have an `items` collection. But, it sounds like the same principle applies.

packages/datastore/src/sync/processors/sync.ts
david-mcafee: Manuel's comment includes excerpts from the spec that address this. TLDR: DataStore should continue merging the rest of the data and make sure to invoke the error handler for non-applicable data. Tested this locally - other than receiving AppSync errors for `retrievePage`, this will continue to fail (for instance, in the event that the customer adds a model to their local schema but doesn't push that change to the backend), due to failures on establishing a subscription. Same applies to adding a non-existent field to a model, etc. In other words, if customers were previously depending on that failure state as a safeguard, it will continue to exist.

packages/datastore/src/sync/processors/sync.ts
david-mcafee: Updated

svidgen (approved):
Sounds like this does not address any customer-observed bugs and instead unblocks another PR you're prepping. So, I'm OK with letting this through without test changes; but if you *can* add a test that demonstrates what this fixes specifically, that would be ideal.

david-mcafee:
> Sounds like this does not address any customer-observed bugs and instead unblocks another PR you're prepping. So, I'm OK with letting this through without test changes; but if you _can_ add a test that demonstrates what this fixes specifically, that would be ideal.

This fix unblocks a PR that is specifically adding unit tests that will test these code paths. I played around a little with attempting to isolate only this change in testing, but think it probably just makes more sense to stick with the tests that are in the next PR.

dpilch (approved):

manueliglesias (approved):
LGTM 👍

PR #11281, opened 2023-04-20
feat(auth): resendSignUpCode API

Samaritan1011001:
#### Description of changes

Adds the re-written resendSignUpCode API with the necessary types and clients

Validated by running this against a react app

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/auth/src/providers/cognito/apis/resendSignUpCode.ts
```diff
@@ -0,0 +1,67 @@
...
+>(
+	resendRequest: ResendSignUpCodeRequest<PluginOptions>
+): Promise<AuthCodeDeliveryDetails<CognitoUserAttributeKey>> {
+	const username = resendRequest.username;
+	assertValidationError(
+		!!username,
+		AuthValidationErrorCode.EmptySignUpUsername
+	);
+	const config = Amplify.config;
+	try {
```
helgabalashova, line 45: no need to have try/catch block here since we moved this error handling to http client

packages/auth/src/providers/cognito/apis/resendSignUpCode.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @returns AuthCodeDeliveryDetails
+ * @throws service: {@link SignUpException } - Cognito service errors thrown during the sign-up process.
+ *
+ *
+ * TODO: add config errors
+ */
+
+// TODO(Samaritan1011001): Function type was changed, may need API change in doc
+
+export async function resendSignUpCode<
```
helgabalashova, line 34: function type is definitely different (look at how signUp and resetPassword are defined now). I think you can check with @israx on latest changes

packages/auth/src/providers/cognito/index.ts
```diff
@@ -3,9 +3,14 @@
...
 export { resetPassword } from './apis/resetPassword';
+export { resendSignUpCode } from './apis/resendSignUpCode';
 
 export { ClientMetadata } from './types/models/ClientMetadata';
-export { CognitoUserAttributeKey } from './types/models/CognitoUserAttributeKey';
+export { CognitoResendSignUpCodeOptions } from './types/options/CognitoResendSignUpCodeOptions';
+export { AuthPluginOptions } from './types/options/AuthPluginOptions';
+export { ResendSignUpCodeRequest } from '../../types/requests/ResendSignUpCodeRequest';
+export { AuthCodeDeliveryDetails } from './types/models/AuthCodeDeliveryDetails';
+export type { CognitoUserAttributeKey } from './types/models/CognitoUserAttributeKey';
```
helgabalashova, line 13: no need for type here. I fixed it in one of my PRs (make sure your branch is up to date)

helgabalashova (commented):
I left you a couple of comments. Also, I don't think you add new types to types index file

packages/auth/src/providers/cognito/index.ts
Samaritan1011001: Not sure what the fix is, can you point me to the changes you made and where I should get these types from?

Samaritan1011001: Removed a few but `CognitoUserAttributeKey` and `AuthPluginOptions` still needs to be exported here right? 
Can you point me to the fix you mention?

israx (changes requested):

packages/auth/src/providers/cognito/apis/resendSignUpCode.ts
```diff
@@ -0,0 +1,63 @@
...
+import { AuthValidationErrorCode } from '../../../errors/types/validation';
+import { SignUpException } from '../types/errors/service';
+import { resendSignUpConfirmationCodeClient } from '../utils/clients/resendSignUpCodeClient';
+
+/**
+ * Resend the confirmation code while signing up
+ *
+ * @param resendRequest - The resendRequest object
+ * @returns AuthCodeDeliveryDetails
+ * @throws service: {@link SignUpException } - Cognito service errors thrown during the sign-up process.
```
AllanZhengYP, line 26: More of a quesion of coding pattern: Should we use `@throws` in all the upcoming PRs? 

I usually don't use the `@throws` because it's very hard to have an exhausted list of errors in JS generally. Networking errors or runtime errors are likely to be thrown too. This is why TS models the errors as `unknown`. 
But it seems like we are going to have each line of code caught and re-wrapped by `SignUpException`.

Samaritan1011001: Thanks, I had to update the doc string. But still have the `@throws` so what would you recommend otherwise. I followe `signUp` API's lead on the patterns.

packages/auth/src/providers/cognito/apis/resendSignUpCode.ts
```diff
@@ -32,36 +32,32 @@ import { resendSignUpConfirmationCodeClient } from '../utils/clients/resendSignU
 // TODO(Samaritan1011001): Function type was changed, may need API change in doc
 
 export async function resendSignUpCode<
-	PluginOptions extends CognitoResendSignUpCodeOptions = AuthPluginOptions
+	ServiceOptions extends AuthServiceOptions
```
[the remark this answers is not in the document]
Samaritan1011001: Made this change but ServiceOptions are unused as I use `resendRequest.options?.serviceOptions?.clientMetadata`. 
If we are removing generics then I remove `ServiceOptions extends AuthServiceOptions` altogether?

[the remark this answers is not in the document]
israx: yeah and we are just passing `CognitoResendSignUpCodeOptions` as type argument directly in the request instead

packages/auth/src/providers/cognito/utils/clients/HttpClients.ts
```diff
@@ -14,9 +16,22 @@ import { assertServiceError } from '../../../../errors/utils/assertServiceError'
 
 const USER_AGENT = 'amplify test';
```
elorzafe, line 17: I would add a `TODO` here

packages/auth/src/providers/cognito/utils/clients/ResendSignUpCodeClient.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+	ResendConfirmationCodeCommandInput,
+	ResendConfirmationCodeCommandOutput,
+} from '@aws-sdk/client-cognito-identity-provider';
```
elorzafe, line 7: We should not use `@aws-sdk` this time

elorzafe (commented):
Is lookng very good @Samaritan1011001 

I think we need some tests unless I am missing something

Thanks!

packages/auth/src/providers/cognito/types/options/AuthPluginOptions.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Auth options generic to plugin.
+ */
+export type AuthPluginOptions = any;
```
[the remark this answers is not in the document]
israx: We should remove it, I think `AuthServiceOptions` is used instead

packages/auth/src/providers/cognito/utils/clients/ResendSignUpCodeClient.ts
israx: All the current implemented APIs on next are using the aws-sdk types for now. Probably just adding a TODO for replacing all the commands if that is the intention in the future.

israx: This wound't add any additional bundle size as these commands are just used as types

elorzafe: can we add the `type` keyword to help review it please

elorzafe (dismissed):
Looks good, a nit comment if you can add `type` keyword on the imported types.

Thanks @Samaritan1011001

packages/auth/src/providers/cognito/apis/resendSignUpCode.ts
```diff
@@ -0,0 +1,54 @@
...
+	const res: ResendConfirmationCodeCommandOutput =
+		await resendSignUpConfirmationCodeClient({
+			Username: username,
+			ClientMetadata:
+				resendRequest.options?.serviceOptions?.clientMetadata ??
+				config.clientMetadata,
+		});
+	const { DeliveryMedium, AttributeName, Destination } = {
+		...res.CodeDeliveryDetails,
+	};
```
AllanZhengYP, line 46: Maybe cleaner to use the desturcture, it will look like
```suggestion
	const { CodeDeliveryDetails: {
	  DeliveryMedium,
	  AttributeName,
	  Destination,
	 } }: ResendConfirmationCodeCommandOutput =
		await resendSignUpConfirmationCodeClient({
			Username: username,
			ClientMetadata:
				resendRequest.options?.serviceOptions?.clientMetadata ??
				config.clientMetadata,
		});
```

packages/auth/src/providers/cognito/index.ts
```diff
@@ -1,14 +1,21 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+// apis
```
AllanZhengYP, line 4: Very nit: I don't think `// apis` and `// models`, `// options` comment are really necessary. Grouping in lines and the import path is pretty self-explanatory

AllanZhengYP (commented):
This PR looks good to me, I only have 2 aesthetic feedbacks

packages/auth/src/providers/cognito/utils/clients/ResendSignUpCodeClient.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
```
israx, line 4: I think here we are missing the `type` syntax annotation

israx (approved):
Thanks Manoj. It is looking good. The only update left I think is to import all the sdk commands as `types`.

israx (approved):

elorzafe (approved):
Great work @Samaritan1011001 🎖️

AllanZhengYP (approved):
Looks good to me!

PR #11290, opened 2023-04-24
fix(datastore): add missing errorHandler errors for sync and mutation unauth responses 

svidgen:
#### Description of changes

Adds missing `errorHandler` events for "top level" unauthorized errors in AppSync resolvers (i.e., when resolver VTL calls `$util.unauthorized()`).

To support the testing for this, this PR adds:

1. Ability to hook into `errorHandler` using an `Observable` from the `getDatastore()` test factory.
2. Ability to inject a single `intercept` method into the graphql fake, so that individual responses can be faked in the context of a full sync. (Helpful if/when we clamp on what a single failed sync, mutation, or subscription means to the overall state of DataStore.)

#### Issue #, if available

#11042

Added protocol tests based on:

1. Forcing relevant resolvers to call `$util.unauthorize`.
3. Capturing responses.
4. Added simulated responses to protocol test.
5. Add assertion for `errorHandler` to trigger an `Unauthorized` error.
6. Made the tests pass.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/connectivityHandling.test.ts
+++ b/packages/datastore/__tests__/connectivityHandling.test.ts
@@ -1049,4 +1049,26 @@ describe('DataStore sync engine', () => {
 		`);
 		});
 	});
+
+	describe.only('error handling', () => {
+		test('unauthorized exception', async () => {
+			graphqlService.intercept = (request, next) => {
+				if (request.query.includes('syncLegacyJSONComments')) {
+					// e.g., not logged in.
+					// comes back as an HTTP 401, but also includes `errors` array.
+					throw {
+						errors: [
+							{
+								errorType: 'UnauthorizedException',
+								message: 'Valid authorization header not provided.',
+							},
+						],
+					};
+				} else {
+					return next();
+				}
+			};
+			await waitForDataStoreReady();
+		});
+	});
 });
--- a/packages/datastore/__tests__/connectivityHandling.test.ts
+++ b/packages/datastore/__tests__/connectivityHandling.test.ts
@@ -8,7 +8,7 @@ import {
 	waitForSyncQueriesReady,
 } from './helpers';
 import { Predicates } from '../src/predicates';
-import { syncExpression } from '../src/types';
+import { syncExpression, SyncError } from '../src/types';
 
 /**
  * Surfaces errors sooner and outputs them more clearly if/when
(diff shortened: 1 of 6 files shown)


packages/datastore/src/sync/processors/sync.ts
```diff
@@ -389,6 +414,11 @@ class SyncProcessor {
 											onTerminate
 										));
 									} catch (error) {
+										console.log(
+											'received error',
+											error,
+											getSyncErrorType(error)
+										);
```
svidgen, line 421: Whoops! This is troubleshooting cruft. Will remove.

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -26,10 +26,22 @@ async function waitForEmptyOutboxOrError(service) {
 	return await Promise.race([waitForEmptyOutbox(), pendingError]);
 }
 
+function waitForNextErrorHandlerError(
```
svidgen, line 29: Will rev with a docstring.

svidgen: Also generalizing the name. This doesn't need to be specific to the `errorHandler` wrapper.

packages/datastore/src/sync/processors/sync.ts
```diff
@@ -284,20 +285,44 @@ class SyncProcessor {
...
+							message: error.message,
+							model: modelDefinition.name,
+							operation: opName,
+							errorType: getSyncErrorType(error.errors[0]),
+							process: ProcessName.sync,
+							remoteModel: null!,
+							cause: error,
+						});
+						throw new NonRetryableError(error);
 					}
```
svidgen, line 326: I need focus here.

I'm not aware of a partial data scenario that comes with auth errors. My impression is that partial data with unauthorized records does not generate *errors*, but instead simply filters `items`. If I'm wrong, I'd like someone to how to produce that response.

dpilch (commented):
Is `errorHandler` not needed on the subscription processor?

svidgen:
> Is `errorHandler` not needed on the subscription processor?

Subscriptions look like they are already issuing error handler messages for the unauth case (observed in studio watching browser console). I can submit a separate PR to clamp on that at the same level.

I don't want to delay the fix for the tests on the not-broken part, especially since I haven't actually looked at how those errors look coming though the `graphql` call yet. I need to do some manual testing to ensure I'm clamping the right interaction there.

dpilch (dismissed):

dpilch (approved):

iartemiev (approved):

PR #11301, opened 2023-04-25
feat(core): Use lerna prepublishOnly version npm publish and use the …

stocaaro:
### Description of changes
The reported version in metrics is currently based on the `@aws-amplify/core` version when it would be more idiomatic to report on the `aws-amplify` version and the version is [trailing by one](https://www.npmjs.com/package/@aws-amplify/core/v/5.1.12?activeTab=code).

This change adds a `prepublishOnly` step to core is run by lerna between updating versioning and pushing the files to npm, then also updates the `genversion` command to source the version from `aws-amplify` instead of core.

This change also removes the `version.ts` file from core, adding it to `.gitignore` to make this a build artifact that isn't checked into git.

It is worth noting that this will now also differentiate preid releases, so a version like `5.1.4-unstable.0197e92.5`  will be published going forward. I have confirmed with product that this is a positive outcome, not a problem to fix.

### Why this change works
When running `lerna publish`, we want to rebuild `version.ts` in core after lerna versions all the packages, but before it publishes to NPM.

The [lerna publish lifecycle scripts](https://github.com/lerna/lerna/tree/main/libs/commands/publish) give us the opportunity to inject commands at different points in the lerna version and publish process. After versioning and immediately before npm release `prepublishOnly` is run if it exists on any releasable package. By adding `prepublishOnly` to `@aws-amplify/core` and having it run `build` for core, this ensures that the `version.ts` file is updated with the current `aws-amplify` package version.

### Description of how you validated changes
I have tested the `aws-amplify` based versioning and publishing to verdaccio to test that the build artifacts match the expected values.

### Checklist

- [x] PR description included
- [x] `yarn test` passes


elorzafe:
I have another idea, this should probably be generated but not versioned so lerna doesnt complain when this is generated and not commited. What do you think @stocaaro ?

stocaaro:
Per Francisco's comment, the `version.ts` is now a build artifact, published to npm and maintained locally after `yarn setup-dev` that won't be checked in to git.

cshfang (approved):

jimblanc (approved):
Looks good to me!

packages/core/package.json
```diff
@@ -23,11 +23,12 @@
 		"build:cjs:watch": "node ./build es5 --watch",
 		"build:esm:watch": "rimraf lib-esm && node ./build es6 --watch",
 		"build": "npm run clean && npm run generate-version && npm run build:esm && npm run build:cjs",
-		"generate-version": "genversion src/Platform/version.ts --es6 --semi",
+		"generate-version": "genversion src/Platform/version.ts --es6 --semi --source ../aws-amplify",
```
elorzafe, line 26: Why do you need that `--source...`?

stocaaro: This changes the version to be from `aws-amplify` instead of corresponding to `@aws-amplify/core`.

elorzafe (approved):
Looks good!

Thanks @stocaaro 🎖️

PR #11307, opened 2023-04-26
chore(clients): Make signing functions synchronous

cshfang:
#### Description of changes

In order to limit the impact of updating our existing Signer to use the updated signing functions, the underlying signing functions need to be synchronous calls as well. We will revisit this in the future in a breaking change.

Tested that signing functions now generate the same output without being `await`ed.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


AllanZhengYP (approved):
It makes sense, thank you for the fix!

PR #11335, opened 2023-05-04
feat(data): DataStore unit tests for rapid consecutive multi-client mutations; test refactors

david-mcafee:
#### Description of changes

- tests observed rapid single-field mutations with variable primary client connection latencies and immediate external client mutation latency
- refactor both single and multi-client consecutive update sections for easier debugging / better readability
- update fake GraphQL service's `graphqlServiceSettled` util to assert on external client updates
- update common assertion util to optionally check multiple fields
- refactor util params (common assertions / fake GraphQL service) for increased readability
- refactor how we track number of mutations (add counter to mutation helper)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/__tests__/helpers/util.ts
```diff
@@ -454,11 +454,13 @@ export async function waitForSyncQueriesReady(verbose = false) {
  * messages for a specific model.
  * @param fakeService - the fake GraphQL service
  * @param expectedNumberOfUpdates - the number of updates we expect to have been received for the model
+ * @param externalNumberOfUpdates - the number of external updates we expect to receive
```
david-mcafee, line 457: Though we don't make assertions specific to this field, I think it's important to separate this from the primary client updates in order to better understand what's happening in the tests.

svidgen: I see. This *sort of* asks the caller to put counts into implicitly named buckets. (Names visible *on hover*, anyway.)

I think I'd be able to reason about things better with an interface that lets me validate that a list of *particular* mutations has settled. E.g.,

```ts
const updates = [];

// we expect the service to see both of these, since they're made
// from different "clients", and we're explicitly waiting for outbox empty.
updates.push(await revPost(original));
updates.push(await externalPostUpdate(original, updates, 2));

await waitForEmptyOutbox();

// we expect these rapid updates to be merged in the outbox
await revPost(original);
updates.push(await revPost(original));

// do we need to make sure it leaves the outbox before we confirm the
// service is done with it?
await waitForEmptyOutbox();

// before we assert final state, wait for everything to settle.
await allSettled(service, updates);
```

But, this may be a good *future* refactor, if you also see the value in it.

david-mcafee: I had also considered this, but went with the simpler approach because I wasn't sure of the value in tracking specific requests, since each test is contained to a small subset of mutations. Perhaps let's discuss offline.

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -891,17 +894,62 @@ describe('DataStore sync engine', () => {
...
+							'\t\t\tid\n' +
+							'title\n' +
+							'blogId\n' +
+							'updatedAt\n' +
+							'createdAt\n' +
+							'_version\n' +
+							'_lastChangedAt\n' +
+							'_deleted\n' +
+							'\t\t}\n' +
+							'\t}',
```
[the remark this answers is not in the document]
david-mcafee: I'll try this! I just went with the pattern that already exists in other tests.

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -1441,6 +1516,839 @@ describe('DataStore sync engine', () => {
...
+						const subscription = await DataStore.observe(Post).subscribe(
+							({ opType, element }) => {
+								const response: SubscriptionLogTuple = [
+									element.title,
+									// @ts-ignore
+									element._version,
+								];
+								subscriptionLog.push(response);
+							}
+						);
```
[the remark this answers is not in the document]
david-mcafee: Sure, works for me!

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -1441,6 +1516,839 @@ describe('DataStore sync engine', () => {
...
+							numberOfUpdates,
+							waitOnOutbox: false,
+							pauseBeforeMutation: false,
+							externalClientMutation: async () =>
+								await externalPostUpdate(
+									original.id,
+									{ title: 'update from second client' },
+									1
+								),
+						});
```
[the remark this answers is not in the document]
david-mcafee: Discussed offline, will update.

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -954,19 +1003,25 @@ describe('DataStore sync engine', () => {
 					if (waitOnOutbox) {
 						await waitForEmptyOutbox();
 					}
+
+					// External update is in the middle of current client updates
+					if (externalClientMutation && number === 1)
+						await externalClientMutation();
```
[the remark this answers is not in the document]
david-mcafee: Discussed offline - will update. See other comment.

packages/datastore/__tests__/helpers/util.ts
david-mcafee: I updated `revPost` to increment a counter, which will be cleared in the `afterEach` block - will be helpful if we ever need to add additional mutations to these tests. Thanks for the recommendation, Jon!

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -1435,10 +1429,773 @@ describe('DataStore sync engine', () => {
...
+						//endregion
 
-					// Cleanup:
-					await subscription.unsubscribe();
+						await waitForEmptyOutbox();
+
+						await graphqlServiceSettled({
+							graphqlService,
+							expectedNumberOfUpdates: 0,
+							externalNumberOfUpdates: 1,
```
svidgen, line 1613: I don't understand how this is expected. The external update looks like it is rejected and never comes through as a subscription message.

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -1435,10 +1429,773 @@ describe('DataStore sync engine', () => {
...
+							['post title 1', undefined],
+							['post title 2', undefined],
+							['post title 2', 1],
+						]);
+
+						expectFinalRecordsToMatch({
+							postId: original.id,
+							version: 1,
+							title: 'post title 2',
+						});
```
svidgen, line 1628: Doesn't this disagree with this?

```
							expectedNumberOfUpdates: 0,
							externalNumberOfUpdates: 1,
```

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -1435,10 +1429,773 @@ describe('DataStore sync engine', () => {
...
```
svidgen, line 1675: Same as last test. These appear to disagree with each other. What am I missing?

Are you sure it isn't one update for "DataStore" updates (merged in outbox) + 0 external updates (rejected)?

(Are rejected updates counted as zero in the fake?)

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -1435,10 +1429,773 @@ describe('DataStore sync engine', () => {
...
+
+						await revPost(original.id, 'post title 2');
+						//endregion
+
+						await waitForEmptyOutbox();
+
+						await graphqlServiceSettled({
+							graphqlService,
+							expectedNumberOfUpdates,
+							externalNumberOfUpdates: 0,
```
svidgen, line 1848: I'm confused. Why is this 0, given we make a successful external update ~12 lines above?

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -1435,10 +1429,773 @@ describe('DataStore sync engine', () => {
...
+							modelName: 'Post',
+						});
+
+						/**
+						 * NOTE: Even though the primary client is updating `title`, the
+						 * second client's update to `blogId` reverts the primary client's
+						 * changes.
+						 */
```
svidgen, line 1856: My impression was that the external update takes effect while *the client under test*'s updates are still in flight. By the time the *client under test*'s updates reach the service, the `_version` has changed, so automerge chooses the existing server-side `title` over the proposed updated `title`. Can you confirm?

svidgen (commented):
I like the way these tests are shaping up! I just have a few questions, where either the assertion counts are mis-labeled or I'm misunderstanding something.

Thanks!

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -1435,10 +1429,773 @@ describe('DataStore sync engine', () => {
...
+
+						//region perform consecutive updates
+						await revPost(original.id, 'post title 0');
+
+						await revPost(original.id, 'post title 1');
+
+						await externalPostUpdate(
+							original.id,
+							{ title: 'update from second client' },
+							undefined
```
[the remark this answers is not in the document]
david-mcafee: Thinking about this more, this test shouldn't exist. For a single-client, testing updates while an initial save is pending makes sense. However, the second client would never have knowledge of this record, because the save is still pending. Will remove. Thanks for catching this!

packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -1435,10 +1429,773 @@ describe('DataStore sync engine', () => {
...
+						await pause(200);
+						await revPost(original.id, 'post title 0');
+
+						await pause(200);
+						await revPost(original.id, 'post title 1');
+
+						await externalPostUpdate(
+							original.id,
+							{ title: 'update from second client' },
+							undefined
```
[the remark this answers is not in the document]
david-mcafee: See above. Will remove.

packages/datastore/__tests__/connectivityHandling.test.ts
david-mcafee: No longer applies.

packages/datastore/__tests__/connectivityHandling.test.ts
david-mcafee: No longer applies

packages/datastore/__tests__/connectivityHandling.test.ts
david-mcafee: No longer applies (see previous comment)

packages/datastore/__tests__/connectivityHandling.test.ts
david-mcafee: I didn't originally have `externalNumberOfUpdates`, and was relying solely on `expectedNumberOfUpdates` when I was first writing these tests (and I forgot to flip this to 1). The corresponding single client tests have a value of `expectedNumberOfUpdates - 1` for the primary client param. Thanks for catching this!

svidgen (dismissed):

packages/datastore/__tests__/helpers/fakes/graphqlService.ts
```diff
@@ -508,15 +508,24 @@ export class FakeGraphQLService {
...
 	 * an external client, we want the response back ASAP in order to accurately test outbox merging consistently.
 	 */
-	public graphql(request: GraphQLRequest, ignoreLatency: boolean = false) {
-		return this.intercept(request, () => this.request(request, ignoreLatency));
+	public externalGraphql(request: GraphQLRequest, ignoreLatency = false) {
+		this.log('External request', {
+			request,
+		});
+		return this.graphql(request, ignoreLatency);
 	}
```
svidgen, line 529: Just planting a seed here: We might want try generating and tracking request ID's instead if we're struggling to make sense of the logging. Definitely a future PR thing, if-needed.

It might also be slick in the future to also output a structured event log or something that we can visualize outside the console.

But again, just seeding the ideas. Let's avoid further revs on this PR if possible.

svidgen (approved):

dpilch (dismissed):

packages/datastore/__tests__/helpers/fakes/graphqlService.ts
david-mcafee: I was actually playing around with request IDs, myself! I think that's a fantastic idea. Also, having a map in the service for debugging the sequencing of events is really, really helpful (especially because when Jest buffers the logs, it is not guaranteed to be in order). May add that back in with a comment (i.e. "use this instead of console.logs"), along with request ids, in the near future.

svidgen (approved):

dpilch (approved):

PR #11344, opened 2023-05-05
fix(datastore): don't set lastSync for a model after a sync error

MorFix:
#### Description of changes
When a model sync process failed because of GraphQL errors, 4 things happened that I thought were wrong:
1. It continued fetching the next pages, altough we already know this model has already failed (containing corrupt data)
2. It didn't mark the model sync status as failed anywhere, and as a result it updated its `lastSync` in storage, which led to the sync processor not trying to sync it again on the next refresh.
3. The sync engine raised the READY event even when a sync error occurred.
4. `Hub` didn't have a clear event that is symmetric to `modelSynced`, so I added `modelError`.

**My use case:**
I tried to fix all of the issues, because my use case is to alert the user when they have corrupt data in their local storage.
But after they refresh once, the model that failed to sync didn't try to sync again, and left the corrupt results in my local storage.

I tried making the changes very carefully because it's clear and obvious from the code that some behaviours are intentional, so please let me know if you think I'm changing the desired behavior.

I also have a project using `amplify-swift` and previously had one for android and both of them don't raise the `ready` event if a model fails to sync. Also they both have an event to catch the errored model.

Installing these changes in my project, and making sure failed models are syncing again when we refresh the page.
Also made sure it doesn't continue to the next page if the previous one has failed.

Will run unit tests soon.

#### Checklist
- [X] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/datastore/src/sync/index.ts
+++ b/packages/datastore/src/sync/index.ts
@@ -100,6 +100,8 @@ export enum ControlMessage {
 	SYNC_ENGINE_SUBSCRIPTIONS_ESTABLISHED = 'subscriptionsEstablished',
 	SYNC_ENGINE_SYNC_QUERIES_STARTED = 'syncQueriesStarted',
 	SYNC_ENGINE_SYNC_QUERIES_READY = 'syncQueriesReady',
+	SYNC_ENGINE_SYNC_QUERIES_FINSIHED_WITH_ERRORS = 'syncQueriesFinishedWithErrors',
+	SYNC_ENGINE_MODEL_ERROR = 'modelError',
 	SYNC_ENGINE_MODEL_SYNCED = 'modelSynced',
 	SYNC_ENGINE_OUTBOX_MUTATION_ENQUEUED = 'outboxMutationEnqueued',
 	SYNC_ENGINE_OUTBOX_MUTATION_PROCESSED = 'outboxMutationProcessed',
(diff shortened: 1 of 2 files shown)


packages/datastore/src/sync/index.ts
```diff
@@ -296,33 +301,43 @@ export class SyncEngine {
...
+																	resolve([]);
+																}
+
+																if (
+																	type ===
+																	ControlMessage.SYNC_ENGINE_SYNC_QUERIES_FINSIHED_WITH_ERRORS
+																) {
+																	// Here `data` will be an array of errors
+																	resolve(data);
+																}
```
MorFix, line 324: added this

MorFix:
If the changes are accepted, I suggest (to me) to add the new hub events to the docs

packages/datastore/src/sync/processors/sync.ts
```diff
@@ -436,21 +460,20 @@ class SyncProcessor {
 										 * with no items and allow the loop to continue organically. This ensures
 										 * all callbacks (subscription messages) happen as normal, so anything
 										 * waiting on them knows the model is as done as it can be.
+										 * However, we have also set the errors that caused this model to fail,
```
MorFix, line 463: !

cwomack:
Hey, @MorFix 👋.  Like you mentioned in the description of this PR, the behaviors are intentional and expected.  We'll close this PR, but would invite you to [create an issue](https://github.com/aws-amplify/amplify-js/issues/new/choose) rather on the `amplify-js` repo that details the desired changes that you're looking for.

Issue #8675, opened 2021-07-30
storage.put is not working when retention period is on for the s3 bucket 


rajnikumari022:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Storage

### Amplify Categories

storage

### Environment information

<details>

```
# Put output below this line
System:
    OS: macOS 10.15.7
    CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
    Memory: 26.22 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.18.2 - /usr/local/bin/node
    Yarn: 1.22.10 - ~/node_modules/.bin/yarn
    npm: 7.15.1 - ~/.npm-global/bin/npm
  Browsers:
    Chrome: 92.0.4515.107
    Edge: 92.0.902.62
    Firefox: 88.0
    Safari: 14.1.2
  npmPackages:
    @aws-amplify/ui-react: ^1.2.4 => 1.2.4 
    @material-ui/core: ^4.11.4 => 4.11.4 
    @material-ui/data-grid: ^4.0.0-alpha.33 => 4.0.0-alpha.33 
    @material-ui/icons: ^4.11.2 => 4.11.2 
    @material-ui/lab: ^4.0.0-alpha.58 => 4.0.0-alpha.58 
    @testing-library/jest-dom: ^5.11.4 => 5.14.1 
    @testing-library/react: ^11.1.0 => 11.2.7 
    @testing-library/user-event: ^12.1.10 => 12.8.3 
    aws-amplify: ^4.1.2 => 4.1.2 
    clsx: ^1.1.1 => 1.1.1 
    material-ui-dropzone: ^3.5.0 => 3.5.0 
    md5: ^2.3.0 => 2.3.0 
    moment: ^2.29.1 => 2.29.1 
    react: ^17.0.2 => 17.0.2 (17.0.1)
    react-dom: ^17.0.2 => 17.0.2 
[...]

PR #11345, opened 2023-05-05
fix(storage): add contentMd5 to put api

kvramyasri7:
#### Description of changes

This PR adds contentMd5 to put API which is mandatory for buckets when object lock is enabled.

#### Issue #, if available

#8675 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/storage/__tests__/providers/AWSS3Provider-unit-test.ts
```diff
@@ -692,6 +692,31 @@ describe('StorageProvider test', () => {
...
+					res({
+						identityId: 'id',
+					});
+				});
+			});
+			const spyon = jest.spyOn(S3Client.prototype, 'send');
+			expect(
+				await storage.put('key', 'object', {
+					level: 'private',
+					contentMd5: 'zYaSxCl62qjqjbLjF1pIoA==',
```
jimblanc, line 708: I know this is just a unit test, but it seems strange that the MD5 appears to be base64 encoded. Does the API require this? If so, is it something we should be doing for customers (with them just passing in the raw MD5)?

packages/storage/src/types/AWSS3Provider.ts
```diff
@@ -75,6 +75,7 @@ type _S3ProviderPutConfig = {
 	tagging?: PutObjectRequest['Tagging'];
 	useAccelerateEndpoint?: boolean;
 	resumable?: boolean;
+	contentMd5?: PutObjectRequest['ContentMD5'];
```
jimblanc, line 78: Nit: Would prefer to group this with the other `content` fields above

jimblanc (commented):
Change looks good, could you please add some more details to the PR on how it was tested & which Storage use-cases were exercised? (E.g. test bucket settings, multi-part upload, resumable, etc.)

stocaaro:
Good news on the BR review is that this only impacts the S3 provider. That bleeds through to the category through the generics default, but it should make the review more straightforward.

cshfang (commented):
The code looks fine to me. But is there any way for us to detect when this is *required* for customers? Just wondering if that's even something we'd want to consider.

kvramyasri7:
Closing this as this is already implemented with another approach.

PR #11413, opened 2023-05-26
fix(storage): make getProperties param optional

israx:
#### Description of changes

Release 5.2.3 brought a braking change by introducing a getProperties required field in the StorageProvider interface. The solution was to make this field optional

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/storage/src/Storage.ts
+++ b/packages/storage/src/Storage.ts
@@ -322,11 +322,11 @@ export class Storage {
 			throw new Error('No plugin found with providerName');
 		}
 		const cancelTokenSource = this.getCancellableTokenSource();
-		
+
 		if (typeof plugin.getProperties !== 'function') {
 			throw new Error(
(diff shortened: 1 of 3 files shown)


packages/storage/src/Storage.ts
```diff
@@ -322,6 +322,12 @@ export class Storage {
 			throw new Error('No plugin found with providerName');
 		}
 		const cancelTokenSource = this.getCancellableTokenSource();
+		
+		if (typeof plugin.getProperties !== 'function') {
+			throw new Error(
+				`.getProperties is not implemented on provider ${plugin.getProviderName()}`
+			) 
+		}
```
ioanabrooks, line 330: What would a test look like for this? We should probably include one here.

jimblanc (approved):

haverchuck (approved):

Issue #1550, opened 2018-08-31
Feature Request: Option to return object metadata with Storage.list()


lemmingworks:
It's often very useful to store extra information about the files stored on S3 in metadata. e.g. the row count in a csv file, a user readable name etc, but this info isn't available using Storage.list().

Storage.list() uses S3.listObjects() which doesn't return object metadata, so an extra S3.headObject() call would need to be made for each file returned by S3.listObjects() in order to get the metadata for each object, which is a performance hit, but it would be nice to have the option, e.g. to pass in a `metadata: true` flag in the options for Storage.list().

If this feature would be acceptable, then I would be happy to create a pull request for it.

mrichman:
I'm trying to do this as well.

theodorehadges:
Hi all, any update on this? Also could benefit from having object metadata included in `Storage.list()`.

sw-yx:
related - #6157

tiomun:
Its kind of absurd that list doesn't already return the files meta data and other parameters, or at least have an option to toggle the data on and off in the response.

Frankly this may end up being a deal breaker for us using amplify which sucks massively since were 6 months into development using amplify for our application already. But I never would have imagined that listing files would be the point of failure for us....

ashika01:
@tiomun this seems like a limitation even from s3 sdk side. There was FR -> https://github.com/aws/aws-sdk-js/issues/1181

I would suggest reaching out S3 directly if this is a feature for s3 to provide. A workaround in mentioned in the linked issue above. You can try to use that.

kvramyasri7:
Now we can access metadata of an object using `Storage.getProperties` method.

natikagg:
> Now we can access metadata of an object using `Storage.getProperties` method. Please do find the [Reference Doc](https://docs.amplify.aws/lib/storage/get-properties/q/platform/js/).

Not able to get the metadata back, it is always coming back as undefined

PR #11469, opened 2023-06-07
feat(storage): add getProperties API 

kvramyasri7:
#### Description of changes
This api enables users to retrieve properties of object from storage. Making getProperties API optional to avoid breaking changes to customers

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/6830
https://github.com/aws-amplify/amplify-js/issues/1550

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/storage/__tests__/Storage-unit-test.ts
```diff
@@ -614,6 +618,76 @@ describe('Storage', () => {
...
+			const customProvider = new TestCustomProvider();
+			const customProviderGetSpy = jest.spyOn(customProvider, 'getProperties');
+			storage.addPluggable(customProvider);
+			await storage.getProperties('key', {
+				provider: 'customProvider',
+			});
+			expect(customProviderGetSpy).toBeCalled();
+		});
+	});
+
```
israx, line 690: can we add an error test to check a custom provider not implementing the  `getProperties` API ?

packages/storage/src/Storage.ts
```diff
@@ -307,6 +309,30 @@ export class Storage {
...
+			pluggable => pluggable.getProviderName() === provider
+		);
+		if (plugin === undefined) {
+			logger.debug('No plugin found with providerName', provider);
+			throw new Error('No plugin found with providerName');
+		}
+		const cancelTokenSource = this.getCancellableTokenSource();
+		if (typeof plugin.getProperties !== 'function') {
+			return Promise.reject(
+				`.copy is not implemented on provider ${plugin.getProviderName()}`
```
israx, line 327: should be '.getProperties' instead of 'copy'

israx (commented):
Looks good. Thanks for putting this up Ramya. Just a few comments related to test cases and correcting a function name. 

I hope in the feature we can change all these types to have a more consistent and robust system that allows easily extensibility. Otherwise every time we add an API we would need to do sneak the code base and see nothing brakes .

packages/storage/__tests__/Storage-unit-test.ts
```diff
@@ -44,6 +44,10 @@ class TestCustomProvider implements StorageProvider {
 		return Promise.resolve({ newKey: 'get' });
 	}
 
+	getProperties(key: string, options?: CustomProviderConfig) {
```
ioanabrooks, line 47: It might be helpful to also have a test class that implements StorageProvider without providing this function.

packages/storage/src/Storage.ts
kvramyasri7: Thanks. Updated.

packages/storage/__tests__/Storage-unit-test.ts
kvramyasri7: I have just removed just removed `getProperties` from `CutomeTestProvider` .  Please let me know if that works.

packages/storage/__tests__/Storage-unit-test.ts
kvramyasri7: Updated it. Thanks for the suggestion.

jimblanc (approved):
LGTM!

packages/storage/src/types/Storage.ts
```diff
@@ -69,13 +71,22 @@ export type StorageCopyDestination = Omit<StorageCopyTarget, 'identityId'>;
...
+		| StorageProviderWithGetProperties,
 	U extends StorageProviderApi
 > = ReturnType<T['getProviderName']> extends 'AWSS3'
 	? LastParameter<AWSS3Provider[U]> // check if it has 'copy' function because 'copy' is optional
 	: T extends StorageProviderWithCopy
-	? LastParameter<T[U]> & { provider: ReturnType<T['getProviderName']> }
-	: LastParameter<T[Exclude<U, 'copy'>]> & {
+	? LastParameter<T['copy']> & {
+			provider: ReturnType<T['getProviderName']>;
+	  }
```
cshfang, line 84: What happens if the storage provider has both `copy` and `getProperties`?

packages/storage/src/types/Storage.ts
```diff
@@ -133,8 +152,10 @@ type PickProviderOutput<
 	? T['getProviderName'] extends 'AWSS3'
 		? DefaultOutput
 		: T extends StorageProviderWithCopy
-		? ReturnType<T[api]>
-		: ReturnType<T[Exclude<api, 'copy'>]>
+		? ReturnType<T['copy']>
```
cshfang, line 155: Same question here I think. Is the return type correct if the provider includes both properties?

packages/storage/src/types/Storage.ts
israx: now we are also testing with a custom provider that has both APIs `copy` and `getProperties`

israx (approved):
Thanks Ramya for putting this together. I hope we can get rid of the conditional type hell from most of this types.

cshfang (approved):
Lgtm thanks!

israx (approved):
lgtm

stocaaro (approved):

jimblanc (approved):

PR #11478, opened 2023-06-08
fix: React detection and Analytics batch upload

stocaaro:
#### Description of changes
React 18 makes complicated most types of detection. Using a common root element lookup wasn't working consistently. This adds a more robust search mechanism that works on the 3 react 18 apps I've tested.

Also, Analytics batch events where missing the user agent action, which this fixes.

Tested snippets manually, then by linking changes into a test app.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/analytics/src/Providers/EventBuffer.ts
+++ b/packages/analytics/src/Providers/EventBuffer.ts
@@ -1,4 +1,4 @@
-import { ConsoleLogger as Logger } from '@aws-amplify/core';
+import { AnalyticsAction, ConsoleLogger as Logger } from '@aws-amplify/core';
 import {
 	putEvents,
 	PutEventsInput,
@@ -6,6 +6,7 @@ import {
 } from '@aws-amplify/core/internals/aws-clients/pinpoint';
 import { EventBuffer, EventObject, EventMap } from '../types';
 import { isAppInForeground } from '../utils/AppUtils';
+import { getAnalyticsUserAgentString } from '../utils/UserAgent';
 
 const logger = new Logger('EventsBuffer');
 const RETRYABLE_CODES = [429, 500];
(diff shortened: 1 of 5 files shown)


erinleigh90 (approved):

packages/analytics/src/Providers/EventBuffer.ts
```diff
@@ -101,7 +102,11 @@ export default class EventsBuffer {
 		try {
 			const { credentials, region } = this._config;
 			const data: PutEventsOutput = await putEvents(
-				{ credentials, region },
+				{
+					credentials,
+					region,
+					userAgentValue: getAnalyticsUserAgentString(AnalyticsAction.Record),
```
cshfang, line 108: Is this meant to track how often `Record` is called on Analytics or how often events are batch sent to Pinpoint?

packages/core/src/Platform/detection/React.ts
```diff
@@ -6,7 +6,15 @@ import { documentExists, processExists, windowExists } from './helpers';
...
 
 export function reactWebDetect() {
-	return documentExists() && !!document.getElementById('react-root');
+	const elementKeyPrefixedWithReact = k => {
+		return k.startsWith('_react') || k.startsWith('__react');
+	};
+	const elementIsReactEnabled = e => {
+		return Object.keys(e).find(elementKeyPrefixedWithReact);
+	};
+	const allElementsWithId = () => Array.from(document.querySelectorAll('[id]'));
```
cshfang, line 15: Nit since this doesn't run often to my understanding but typically we'd want to extract these so they're not recreated every time `reactWebDetect()` is run.

Second nit would be to use more meaningful parameter names... it's easy enough to intuit that `k` is key or `e` is element but it's even easier when it's spelled out for you.

cshfang (commented):
Probably could have opened two tiny PRs since the changes are unrelated but did have one small question (even though I think I know the answer)

PR #11544, opened 2023-06-24
Adds new singleton initial version

elorzafe:
#### Description of changes

- Adds new Singleton initial implementation
- Adds tests for incremental config changes (Auth)
- Adds initial types for config
- Adds `AuthTokenOrchestrator`, `AuthTokenStore` interfaces and their implementations
- Adds `KeyValueStorageInterface` and `MemoryKeyValueStorage` implementation
- Adds `CredentialProvider` and `IdentityIdProvider` interfaces.
- Adds initial `fetchAuthSession` and `listenAuthSession` APIs
- Adds dependency of `rxjs` for Observables

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/core/src/types/types.ts
```diff
@@ -85,3 +85,54 @@ export type ServiceError = {
...
+};
+
+
+export type LibraryOptions = {
+	Auth?: {
+		tokenProvider?: TokenProvider;
+		refreshTokenClient?: TokenRefreshClient;
+		credentialsProvider?: CredentialsProvider;
+		storage?: AuthStorage;
+	} | null;
```
jimblanc, line 111:
```suggestion
	};
```

I don't think the `null` is needed?

jimblanc (commented):
Looks great! I left a bunch of documentation-related nit-picks & suggestions since `Amplify` is the first thing customers will interact with in the library

packages/core/src/singleton/index.ts
```diff
@@ -0,0 +1,165 @@
...
+	 * Obtain the backend resources the library is configured
+	 * @returns ResourceConfig
+	 */
+	export function getConfig(): ResourceConfig {
+		return JSON.parse(JSON.stringify(singletonResourcesConfig));
+	}
+
+	// TODO: Add more cross category functionality
+
+	export namespace Auth {
```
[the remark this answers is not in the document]
jimblanc: Really though I suppose it doesn't offer much as it would need to still be exposed on the public interface, but the organization would be easier to follow as we add more cross-category functionality (can namespaces be re-exported from another namespace?)

israx (commented):
My main concern is that there is not a clear separations of concerns for `AuthTokenStore` and `KeyValueStorageInterface`

In my opinion:

`AuthTokenStore`  -> the concrete implementation should manage service specific keys, including legacy key-value token migration.

`KeyValueStorageInterface` -> the concrete implementation should be service agnostic as this is what we are exposing to customers e.g `LocalStorage implements KeyValueStorageInterface`.

What I propose is to:

- expose singletonLibraryOptions to the `Amplify.getConfig` method 
- add a  required `authStorageStore` key to `singletonLibraryOptions` 
- expose a CognitoStorageStore to customers.

So then the token orchestrator can pull up the concrete `AuthTokenStore` from the singleton. That way there is no need to pass it to the internal methods of each orchestrator

```typescript
const DefaultAuthTokensOrchestrator: : AuthTokenOrchestrator = {
    authTokenStore: Amplify.getConfig().libraryOptions.Auth.authTokenStore,
    async getTokens(options:GetAuthTokenOptions, tokenRefresher: TokenRefresher){
    tokens = await this.authTokenStore.loadTokens() // This token store already has the 
   // concrete KeyValueStorage and handles token migration
    }
}

// customer DX
import {CognitoUserPoolTokenStore} from 'aws-amplify/auth/cognito'
const backendConfig = {};
const clientConfig ={
   Auth:{
      authTokenStore: CognitoUserPoolTokenStore
    }
}
Amplify.configure(backedConfig, clientConfig)

```
// The pro of this approach is that we won't have a circular dependency, the con is that customers would need to pass this token store

israx:
Another general comment that I have is why don't add classes if needed ? At the end JS nature is to be dynamic if we can combine OOP and functions we are getting the best things of both worlds. 

Also I think the prettier needs to be run. I see some pretty long lines of code.

packages/core/src/types/types.ts
```diff
@@ -85,3 +85,54 @@ export type ServiceError = {
...
+
+export type TokenRefreshClient = (metadata?: Record<string, string>) => Promise<AuthTokens>;
+
+export type CredentialsProvider = (options?: GetTokensOptions) => Promise<ICredentials>;
+
+export type GetTokensOptions = {
+	forceRefresh?: boolean;
+};
+
+export type AuthTokens = {
```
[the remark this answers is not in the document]
elorzafe: it will part of `metadata`

packages/core/src/singleton/Auth/index.ts
```diff
@@ -0,0 +1,252 @@
...
+}: {
+	expiresAt: number;
+	clockDrift: number;
+}): boolean {
+	// compare expiration against token exp and clock drift
+	const currentTime = Date.now();
+	return currentTime + clockDrift > expiresAt;
+}
+
+async function refreshTokens({
```
jimblanc, line 88: Should this be a member of the orchestrator itself?

packages/core/src/singleton/Auth/index.ts
```diff
@@ -0,0 +1,252 @@
...
+		await Amplify.Auth.clearTokens();
+		throw err;
+	}
+}
+
+export function decodeJWT(token: string): JWT {
+	const tokenSplitted = token.split('.');
+	if (tokenSplitted.length !== 3) {
+		throw new Error('Invalid token');
+	}
```
jimblanc, line 111: Might not be needed, but we could apply more strict validation to JWT validation, e.g.

`^([a-zA-Z0-9_=]+)\.([a-zA-Z0-9_=]+)\.([a-zA-Z0-9_\-\+\/=]*)`

packages/core/src/singleton/Auth/index.ts
```diff
@@ -0,0 +1,252 @@
...
+	AuthTokens,
+	GetAuthTokensOptions,
+	KeyValueStorageInterface,
+	TokenRefresher,
+} from '../../types';
+import { Amplify } from '../../Singleton';
+import { JWT, AuthStorageKeys } from './types';
+import { Buffer } from 'buffer'; // TODO: this needs to be a platform operation
+
+export class DefaultAuthTokensOrchestrator implements AuthTokenOrchestrator {
```
jimblanc, line 14: Nit pick, but this file is pretty long. Might be a good idea to break the classes into their own files.

packages/core/src/singleton/Auth/index.ts
```diff
@@ -0,0 +1,252 @@
...
+					Number.parseInt(
+						await this.keyValueStorage.getItem(authKeys.accessTokenExpAt)
+					) || 0,
+				metadata: JSON.parse(
+					(await this.keyValueStorage.getItem(authKeys.metadata)) || '{}'
+				),
+				clockDrift:
+					Number.parseInt(
+						await this.keyValueStorage.getItem(authKeys.clockDrift)
+					) || 0,
```
jimblanc, line 171: Style nit: Would prefer to construct these more complex ones in variables

packages/core/src/singleton/Auth/index.ts
```diff
@@ -0,0 +1,252 @@
...
+			name,
+			this.authConfig.userPoolWebClientId
+		);
+
+		// Not calling clear because it can remove data that is not managed by AuthTokenStore
+		await this.keyValueStorage.removeItem(authKeys.accessToken);
+		await this.keyValueStorage.removeItem(authKeys.idToken);
+		await this.keyValueStorage.removeItem(authKeys.accessTokenExpAt);
+		await this.keyValueStorage.removeItem(authKeys.clockDrift);
+		await this.keyValueStorage.removeItem(authKeys.metadata);
```
jimblanc, line 225: Nit: Promise.all

packages/core/src/singleton/Auth/index.ts
```diff
@@ -0,0 +1,252 @@
...
+				...acc,
+				[authKey]: `${prefix}.${identifier}.${authKey}`,
+			}),
+			{} as AuthKeys<keyof T & string>
+		);
+}
+
+type AuthKeys<AuthKey extends string> = {
+	[Key in AuthKey]: string;
+};
```
jimblanc, line 252: Should go to Auth/types

packages/core/src/singleton/index.ts
```diff
@@ -51,41 +87,62 @@ export namespace Amplify {
...
+			options?: GetAuthTokensOptions
+		): Promise<AuthSession> {
+			let tokens: AuthTokens;
+			let awsCreds: Credentials;
+			let awsCredsIdentityId: string;
+
+			try {
+				tokens = await tokenOrchestrator.getTokens({ options });
+			} catch (error) {
+				console.warn(error);
```
jimblanc, line 105: TODO: Better error handling

packages/core/src/types/types.ts
```diff
@@ -101,38 +102,89 @@ export type ResourceConfig = {
...
+	clearTokens(): Promise<void>;
+	setKeyValueStorage(keyValueStorage: KeyValueStorageInterface): void;
+}
+
+export type AuthSession = {
+	tokens?: AuthTokens;
+	awsCreds?: Credentials;
+	awsCredsIdentityId?: string;
+	authenticated: boolean;
+};
```
jimblanc, line 190: Nit, but should these go in Auth/types?

packages/core/__tests__/singleton/Amplify-test.ts
```diff
@@ -50,5 +51,53 @@ describe('Amplify config test', () => {
 });
 
 // TODO: add test get,set Tokens
```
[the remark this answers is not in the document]
elorzafe: correct

packages/core/src/singleton/index.ts
elorzafe: I should add a comment here, this has to be documented somewhere, but if you cannot `getTokens` it should still get guest credentials eventually. This is an important concern

packages/core/src/singleton/Auth/index.ts
```diff
@@ -29,44 +38,63 @@ function isTokenExpired({ expiresAt, metadata }: { expiresAt: number, metadata:
...
+    }
 }
 
-export const MyTokenStore: AuthTokenStore = {
+export const DefaultTokenStore: AuthTokenStore = {
     loadTokens: async function (keyValueStore: KeyValueStorageInterface): Promise<AuthTokens> {
-        const atString = await keyValueStore.getItem('accessToken');
-        const itString = await keyValueStore.getItem('idToken');
+        try {
+            const atString = await keyValueStore.getItem('accessToken');
```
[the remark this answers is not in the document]
elorzafe: Discussed offline and cleaned up

packages/core/src/singleton/Auth/index.ts
```diff
@@ -29,44 +38,63 @@ function isTokenExpired({ expiresAt, metadata }: { expiresAt: number, metadata:
...
+            toString: () => token,
+            payload: JSON.parse(Buffer.from(payloadString, 'base64').toString('utf8'))
+        };
+    } catch (err) {
+        throw new Error('Invalid token payload')
+    }
 }
 
-export const MyTokenStore: AuthTokenStore = {
+export const DefaultTokenStore: AuthTokenStore = {
```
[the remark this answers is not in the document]
elorzafe: Discussed offline and cleaned up

packages/core/src/singleton/Auth/index.ts
```diff
@@ -1,22 +1,31 @@
...
-        debugger;
-        if ((options && options.forceRefresh) ||
-            isTokenExpired({ expiresAt: tokens.idToken.payload.exp, metadata: tokens.metadata }) ||
-            isTokenExpired({ expiresAt: tokens.accessTokenExpAt, metadata: tokens.metadata })) {
-
-            // tokens = await refreshTokens(tokens);
+        let tokens: AuthTokens;
+        
+        try {
+            tokens = await tokenStore.loadTokens(keyValueStore);
```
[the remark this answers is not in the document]
elorzafe: Discussed offline and cleaned up

packages/core/src/types/types.ts
```diff
@@ -85,3 +85,54 @@ export type ServiceError = {
 	name: string;
 	message: string;
 };
+
+export type ResourceConfig = {
```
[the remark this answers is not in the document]
elorzafe: Yeah, this will be composed from each category configuration, the same that it will be used on each category but defined on core

packages/core/src/singleton/index.ts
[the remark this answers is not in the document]
elorzafe: I move it to classes just to keep variables nested inside of the class

packages/core/src/singleton/index.ts
```diff
@@ -0,0 +1,165 @@
...
+			'Configure'
+		);
+	}
+
+	/**
+	 * Obtain the backend resources the library is configured
+	 * @returns ResourceConfig
+	 */
+	export function getConfig(): ResourceConfig {
+		return JSON.parse(JSON.stringify(singletonResourcesConfig));
```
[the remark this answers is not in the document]
elorzafe: I dont remember having issue on `parse` and `stringify`.

packages/core/src/singleton/index.ts
```diff
@@ -0,0 +1,165 @@
...
+		 */
+		export function getTokens(): Promise<AuthTokens> {
+			// TODO: Read tokens from Auth Storage
+
+			return;
+		}
+	}
+}
+
+function incrementResourceConfig(existingConfig: ResourceConfig, newConfig: ResourceConfig): ResourceConfig {
```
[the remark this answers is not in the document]
elorzafe: We will review how deep this will merge nested config on API review

packages/core/src/singleton/Auth/TokenStore.ts
```diff
@@ -0,0 +1,166 @@
...
+				accessTokenExpAt,
+				metadata,
+				clockDrift,
+			};
+		} catch (err) {
+			throw new Error('No valid tokens');
+		}
+	}
+	async storeTokens(tokens: AuthTokens): Promise<void> {
+		if (this.authConfig === undefined) {
```
israx, line 75: I see this logic is repetitive every time we want to assert, we can create a generic assertion guard  outside the class that will throw AmplifyError. E.g

```typescript
function asserts<T>(assertion:T, errorParams:ErrorParams):asserts assertion {
    if(!assertion) throw new AmplifyError(errorParams);
}
```
and then we can use it like

```typescript
asserts(
        config, 
       {name:'AuthConfigException', 
        message:'Auth not configure',
        recoverySuggestion:'Add Amplify.configure({ Auth: {...} }) to your App'
        })
```

packages/core/src/singleton/Auth/TokenStore.ts
```diff
@@ -0,0 +1,166 @@
...
+			throw new AmplifyError({
+				message: 'Auth not configured',
+				name: 'Auth not configure',
+				recoverySuggestion:
+					'Add Amplify.configure({ Auth: {...} }) to your App',
+			});
+		}
+
+		if (tokens === undefined) {
+			throw new AmplifyError({
```
israx, line 85: This logic repeats as well. Suggested approach above could help

packages/core/src/singleton/Auth/TokenStore.ts
```diff
@@ -0,0 +1,166 @@
...
+		this.keyValueStorage.setItem(
+			authKeys.metadata,
+			JSON.stringify(tokens.metadata)
+		);
+
+		this.keyValueStorage.setItem(authKeys.clockDrift, `${tokens.clockDrift}`);
+	}
+
+	async clearTokens(): Promise<void> {
+		if (this.authConfig === undefined) {
```
israx, line 121: This logic repeats as well. Suggested approach above could help

packages/core/src/singleton/Auth/TokenStore.ts
```diff
@@ -0,0 +1,166 @@
...
+			name,
+			this.authConfig.userPoolWebClientId
+		);
+
+		this.keyValueStorage.setItem(
+			authKeys.accessToken,
+			tokens.accessToken.toString()
+		);
+
+		if (!!tokens.idToken) {
```
israx, line 103: what do we do if tokens are undefined ?

packages/core/src/singleton/Auth/types.ts
```diff
@@ -0,0 +1,145 @@
...
+	storeTokens(tokens: AuthTokens): Promise<void>;
+	clearTokens(): Promise<void>;
+	setKeyValueStorage(keyValueStorage: KeyValueStorageInterface): void;
+}
+
+export type AuthSession = {
+	tokens?: AuthTokens;
+	awsCreds?: Credentials;
+	awsCredsIdentityId?: string;
+	authenticated: boolean;
```
israx, line 53: I think the other platforms have it as `isSignedIn`, wondering it there's any consideration to align

packages/core/src/singleton/Auth/TokenOrchestrator.ts
```diff
@@ -0,0 +1,89 @@
...
+			throw new Error('No session');
+		}
+
+		return { ...tokens };
+	}
+
+	private async refreshTokens({
+		tokens,
+		tokenRefresher,
+		authConfig,
```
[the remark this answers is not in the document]
jimblanc: +1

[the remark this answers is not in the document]
elorzafe: good catch, fixing that now

packages/core/src/singleton/Auth/TokenStore.ts
```diff
@@ -0,0 +1,166 @@
...
+		return;
+	}
+
+	async loadTokens(): Promise<AuthTokens> {
+		if (this.authConfig === undefined) {
+			throw new AmplifyError({
+				message: 'Auth not configured',
+				name: 'Auth not configure',
+				recoverySuggestion:
+					'Add Amplify.configure({ Auth: {...} }) to your App',
```
jimblanc, line 32: Nit: 

```suggestion
					'Make sure to call Amplify.configure in your app',
```

packages/core/src/singleton/Auth/TokenOrchestrator.ts
```diff
@@ -0,0 +1,89 @@
...
+			if (options?.forceRefresh || idTokenExpired || accessTokenExpired) {
+				tokens = await this.refreshTokens({
+					tokens,
+					tokenRefresher: this.tokenRefresher,
+					authConfig: this.authConfig,
+				});
+			}
+		} catch (err) {
+			// TODO: review token handling mechanism, including exponential retry, offline, etc
+			throw new Error('No session');
```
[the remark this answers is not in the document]
elorzafe: Error handling here is different because it might be because there are not tokens which is not an error necessarily. I would like to tweak that after this

packages/core/src/singleton/Auth/TokenStore.ts
jimblanc: Also we might want to extract out the `AmplifyError` for not configured into a helper function somewhere, it'll get used a lot and would be easier to keep messaging aligned.

packages/core/src/singleton/Auth/index.ts
```diff
@@ -1,252 +1,198 @@
-import type {
+import { Buffer } from 'buffer'; // TODO: this needs to be a platform operation
```
jimblanc, line 1: Just a nit on TODOs, I've been standardizing on `TODO(v6):` so it'll be easier to find these since we have TODOs scattered everywhere in the library currently, but if you're tracking it somewhere no worries.

jimblanc (commented):
Looking good! Few changes & nits below but we're getting there :)

packages/core/src/singleton/Auth/TokenOrchestrator.ts
```diff
@@ -0,0 +1,89 @@
...
+		options,
+	}: {
+		options?: FetchAuthSessionOptions;
+	}): Promise<AuthTokens> {
+		// TODO: how to handle if there are not tokens on tokenManager
+		let tokens: AuthTokens;
+
+		try {
+			tokens = await this.tokenStore.loadTokens();
+
```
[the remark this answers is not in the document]
elorzafe: Good catch!

packages/core/src/singleton/Auth/TokenStore.ts
elorzafe: `idToken` is optional on the OIDC spec. so is fine to skip

packages/core/src/singleton/Auth/TokenStore.ts
```diff
@@ -0,0 +1,166 @@
...
+
+			return {
+				accessToken,
+				idToken,
+				accessTokenExpAt,
+				metadata,
+				clockDrift,
+			};
+		} catch (err) {
+			throw new Error('No valid tokens');
```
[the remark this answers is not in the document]
elorzafe: I will add a TODO here as well, needs to be aligned with other platforms the behavior for partial results

packages/core/src/singleton/Auth/TokenStore.ts
```diff
@@ -0,0 +1,147 @@
...
+			};
+		} catch (err) {
+			// TODO(v6): validate partial results with mobile implementation
+			throw new Error('No valid tokens');
+		}
+	}
+	async storeTokens(tokens: AuthTokens): Promise<void> {
+		assertTokenProviderConfig(this.authConfig);
+
+		if (tokens === undefined) {
```
israx, line 72: we can use the asserts function here

packages/core/src/singleton/Auth/TokenStore.ts
```diff
@@ -0,0 +1,147 @@
...
+			throw new AmplifyError({
+				message: 'Invalid tokens',
+				name: 'Invalid tokens',
```
israx, line 75: name: InvalidAuthTokens

israx (approved):
No blocking but for a next PR we could keep using the `asserts` function where it make sense and the `name` param on the AmplifyError should have a format of `<errorname>Exception` to be consistent with Auth

israx (dismissed):

Samaritan1011001 (approved):

packages/core/src/singleton/Auth/index.ts
```diff
@@ -0,0 +1,218 @@
...
+		try {
+			tokens = await this.tokenOrchestrator.getTokens({ options });
+		} catch (error) {
+			// TODO(v6): validate error depending on conditions it can proceed or throw
+		}
+
+		try {
+			if (this.authOptions.identityIdProvider) {
+				awsCredsIdentityId = await this.authOptions.identityIdProvider({
+					tokens,
```
Samaritan1011001, line 114: If there are no tokens (user hasn't logged in) will this be undefined? 
With this code, it won't be because `this.tokenOrchestrator.getTokens` return type `AuthTokens`

elorzafe: if it throws `tokens` will be `undefined`

jimblanc (approved):
Looks good 🔥

PR #11616, opened 2023-07-12
Unblock tagged release publish to npm

manueliglesias:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -67,7 +67,7 @@
 			"name": "API (top-level class)",
 			"path": "./lib-esm/index.js",
 			"import": "{ Amplify, API }",
-			"limit": "89.79 kB"
+			"limit": "95 kB"
 		}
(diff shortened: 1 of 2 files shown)


packages/api/package.json
```diff
@@ -67,7 +67,7 @@
 			"name": "API (top-level class)",
 			"path": "./lib-esm/index.js",
 			"import": "{ Amplify, API }",
-			"limit": "89 kB"
+			"limit": "89.79 kB"
```
svidgen, line 70: I think we should just give ourselves a few kb of margin while we work. Pick a value that'll signal us if we do something downright silly. But otherwise we can backfill and set a target value later, IMHO.

```suggestion
			"limit": "95 kB"
```

svidgen (approved):

PR #11632, opened 2023-07-17
feat: custom user agent Geo changes for UI handoff

erinleigh90:
#### Description of changes
- Creates and exports InternalGeo class from /internals scope with added customUserAgentDetails parameter on public API's.
- Sends received customUserAgentDetails through to geo service calls

NOTES
- To see the changes to InternalGeo after moving the changes over from Geo, see [this comparison](https://github.com/aws-amplify/amplify-js/compare/ca34154e77e16ccf9c36036207259335a2d5bdd6...f590271222def423d1ee98b5b588666fb489277c#diff-909ce785ee7f752dd5c05661be527fa1cc4a3a7caa74aa9dd426a4a09dd141ac)

yarn tests pass

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -151,7 +152,8 @@ export class AmazonLocationServiceProvider implements GeoProvider {
 	 */
 	public async searchByText(
 		text: string,
-		options?: SearchByTextOptions
+		options?: SearchByTextOptions,
+		customUserAgentDetails?: CustomUserAgentDetails
```
stocaaro, line 156: Having a prior art debate in my head. Could do a bit of searching on my own, but instead I'm asking the question...

In this case (and probably storage), we're changing the provider interface itself, which will be exposed to customers.  We need to do this because we want external callers such as UI to add details which will enhance the user agent. I can't think of a way of doing this that doesn't stem the category interface normal/internal (as done here), but then revising all provider interfaces to offer user agent details as an optional input (again, as done here), which get passed along into the client).

Questions:
- Have we rolled out changes that look like this already?
- Have we reviewed the public interface changes where our provider interfaces are growing by this optional input?

Assuming the answer to both these questions is yes, without qualification, then I'm good with this change.

jimblanc: Wondering about this as well. Technically I don't think we advertise that customers can create a Geo provider like we do for Storage & Analytics, but `GeoProvider` is a public surface.

packages/geo/src/internals/InternalGeo.ts
```diff
@@ -0,0 +1,380 @@
...
+	public getDefaultMap(provider = DEFAULT_PROVIDER): MapStyle {
+		const prov = this.getPluggable(provider);
+
+		return prov.getDefaultMap();
+	}
+
+	/**
+	 * Search by text input with optional parameters
+	 * @param  {string} text - The text string that is to be searched for
+	 * @param  {SearchByTextOptions} options? - Optional parameters to the search
```
jimblanc, line 147: Non-blocking nit since it's internal, but we should probably document the new params in doc-strings

jimblanc (approved):
Nits & clarifications below, nothing blocking

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
erinleigh90: We had a conversation with Tim, Francisco, Aaron and I, and the decision was that it is not a concern since it doesn't affect the customer Geo.action api's. It would not be worth the hours/delay/extra lines of code to make them more private.

packages/geo/src/internals/InternalGeo.ts
erinleigh90: Oh right! I forgot to go through and do that in this branch. On it.

jimblanc (dismissed):

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
erinleigh90, line 1: Go back through and update JSDocs

stocaaro (approved):
This makes sense to me. Would be interested in getting any insight from @thaddmt on this work.

stocaaro (approved):

thaddmt (approved):
LGTM, just had two small q's

jimblanc (approved):

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -147,11 +148,13 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
 	 * @param  {string} text - The text string that is to be searched for
 	 * @param  {SearchByTextOptions} options? - Optional parameters to the search
+	 * @param  {CustomUserAgentDetails} customUserAgentDetails - Optional parameter to send user agent details
 	 * @returns {Promise<Place[]>} - Promise resolves to a list of Places that match search parameters
 	 */
 	public async searchByText(
 		text: string,
-		options?: SearchByTextOptions
+		options?: SearchByTextOptions,
+		customUserAgentDetails?: CustomUserAgentDetails
```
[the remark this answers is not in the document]
erinleigh90: We do have BR sign-off here, thanks for checking!

packages/geo/src/Providers/AmazonLocationServiceProvider.ts
```diff
@@ -181,7 +184,7 @@ export class AmazonLocationServiceProvider implements GeoProvider {
 		const client = new LocationClient({
 			credentials: this._config.credentials,
 			region: this._config.region,
-			customUserAgent: getAmplifyUserAgentObject(),
+			customUserAgent: getAmplifyUserAgentObject(customUserAgentDetails),
```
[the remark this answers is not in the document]
erinleigh90: Interesting point. I tried to explain via the CustomUserAgent typings what is expected. I don't think we need to sanitize since sdk will do that for us, but it may be worth noting in the parameter that we are expecting from geo/ui.

PR #11638, opened 2023-07-17
chore: Migrate CI/CD to GitHub Actions

stocaaro:
#### Description of changes

This is the current state of the Github Actions CI/CD migration. It models the following automated processes:
- PRs run
  - unit tests
  - bundle size 
- Merging to main runs:
  - unit tests
  - bundle size
  - integ/e2e tests
  - When all pass
    - Release to `unstable` preid
- Merging to configured preid branch runs:
  - unit tests
  - bundle size
  - integ/e2e tests
  - When all pass
    - Release to specified preid
- Merging to release branch runs:
  - unit tests
  - bundle size
  - integ/e2e tests
  - When all pass
    - Release to `latest`
    - Updates `required-release` tag if needed
    - Commits new versions and docs updates
    - Merged updates back to `main` branch 

Patterns follows:
- Triggered workflows prefixed by the event type that triggers them.
- Callable workflow used to encapsulate reusable operations that run on isolated Github runners.
- Actions used to encapsulate code called across multiple workflows.
- All internal actions/workflows and actions referenced by relative path action and workflow.
- Use caching to reduce runtime of processes
- Prefer to test by execution - including NPM push, which is protected using verdaccio

Examples executions (for execution speed, I only ran a single RN test and ran all other tests on a single browser):
- Change [builds to main](https://github.com/stocaaro/amplify-js/actions/runs/5578141265) releasing to `unstable`
- [Release PR](https://github.com/stocaaro/amplify-js/pull/42) to `release`.
- Merge to `release` [builds/tests/releases](https://github.com/stocaaro/amplify-js/actions/runs/5578685223) to `latest` and post-review process commits versions/docs and merges back to `main`.

#### Common actions
- How do I run integ tests for my branch?
  - Add your branch to the [push-integ-test](https://github.com/aws-amplify/amplify-js/pull/11638/files#diff-94c910063dcaee60cade5e488df24cdc080d49b710a3d64b83b09ee7e6fae48aR3-R7) workflow and push the change to your branch.
- How do I release my changes to a preid?
  - If you've followed the pattern `feat/{preid}/main`, then just add your branch to the [push-preid-release](https://github.com/aws-amplify/amplify-js/pull/11638/files#diff-ccab49738f6826c9d44cf149ef14092b39c7d600a9b79c6ff4712e73cafec236R7) workflow
  - If your branch name follows a different pattern, then [add it to the workflow](https://github.com/aws-amplify/amplify-js/pull/11638/files#diff-ccab49738f6826c9d44cf149ef14092b39c7d600a9b79c6ff4712e73cafec236R7) and also update the [preid passed to the npm publish workflow](https://github.com/aws-amplify/amplify-js/pull/11638/files#diff-ccab49738f6826c9d44cf149ef14092b39c7d600a9b79c6ff4712e73cafec236R32)

By executing workflows, either trimmed to get to the desired condition promptly, but always rolling the whole process together and running end to end.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


.github/workflows/callable-e2e-tests.yml
```diff
@@ -0,0 +1,111 @@
...
+        type: number
+      retry_count:
+        required: true
+        type: number
+
+env:
+  NPM_REGISTRY: http://localhost:4873/
+  NPM_USER: circleci
+  NPM_PASS: circleci
+  NPM_EMAIL: circleci@amplify.js
```
stocaaro, line 41: Doesn't really matter, but we can change away from circleci centric name/pass for verdaccio.

.github/workflows/callable-npm-publish-preid.yml
```diff
@@ -0,0 +1,46 @@
...
+      # <remove>
+      - name: Load Verdaccio with AmplifyJs
+        uses: ./amplify-js/.github/actions/load-verdaccio-with-amplify-js
+      ### </remove>
+
+      - name: Run npm publish
+        uses: ./amplify-js/.github/actions/npm-publish
+        with:
+          github_email: aws@amazon.com
+          github_name: aws-amplify-bot
```
stocaaro, line 43: These are the values used by circle. For circle they are configured as variables on the package / environment.  We should mirror that here.

.github/actions/npm-publish/action.yml
```diff
@@ -0,0 +1,41 @@
...
+  using: 'composite'
+  steps:
+    - name: Run npm publish
+      working-directory: ./amplify-js
+      shell: bash
+      run: |
+        echo "//registry.npmjs.org/:_authToken=${{ inputs.npm_token }}" >> ~/.npmrc
+        touch {{ inputs.bash_env }}
+        echo "export PREID_HASH_SUFFIX=$(echo ${{ github.sha }} | cut -c -7 | sed 's/^/\./')" >> "{{ inputs.bash_env }}"
+        echo "export PREID_PREFIX=${{ inputs.preid || '' }}" >> "{{ inputs.bash_env }}"
```
wlee221, line 35: We should sanitize the inputs (`github.sha` and `inputs.preid`) since `npm-publish` is technically a standalone action and should not trust its inputs.

.github/actions/setup-samples-staging/action.yml
```diff
@@ -0,0 +1,33 @@
+name: Setup Amplify Samples Staging
+description: Checks out Amplify Samples Staging and builds the package with caching
+
+inputs:
+  GH_ACTIONS_STAGING_READ:
```
wlee221, line 5: Non-blocking nit, but I think the name could be a little better to indicate it's a token, e.g. `GITHUB_TOKEN_STAGING_READ` or even `GITHUB_TOKEN` since really this workflow only cares about one `GITHUB_TOKEN`. The caller can do the differentiation.

.github/actions/setup-samples-staging/action.yml
```diff
@@ -0,0 +1,33 @@
...
+runs:
+  using: 'composite'
+  steps:
+    - name: Create cache
+      uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1 https://github.com/actions/cache/commit/88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8
+      id: cache-samples-staging-build
+      with:
+        key: aws-amplify-js-samples-staging-build-${{ github.sha }}
+        path: |
+          ./amplify-js-samples-staging
```
wlee221, line 18: Is this a safe thing to cache?

> We recommend that you don't store any sensitive information in the cache. 
> - https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#about-caching-workflow-dependencies

And I believe `./amplify-js-samples-staging` has sensitive stuffs.

.github/workflows/callable-bundle-size-tests.yml
```diff
@@ -0,0 +1,21 @@
+name: 'Unit Tests'
```
wlee221, line 1: Non-blocking nit: I personally think adding single quotes around the name is extraneous, but this is obviously up to your preference.

.github/workflows/callable-e2e.yml
```diff
@@ -0,0 +1,40 @@
+name: 'E2E Tests'
```
wlee221, line 1: Can we change the file name and the workflow name (L1 here) to differentiate it from `callable-e2e-tests` above?

.github/actions/npm-publish/action.yml
```diff
@@ -0,0 +1,41 @@
...
+runs:
+  using: 'composite'
+  steps:
+    - name: Run npm publish
+      working-directory: ./amplify-js
+      shell: bash
+      run: |
+        echo "//registry.npmjs.org/:_authToken=${{ inputs.npm_token }}" >> ~/.npmrc
+        touch {{ inputs.bash_env }}
+        echo "export PREID_HASH_SUFFIX=$(echo ${{ github.sha }} | cut -c -7 | sed 's/^/\./')" >> "{{ inputs.bash_env }}"
```
wlee221, line 34: Instead of inlining `github.sha` directly in the script, can we use an intermediate environment secret for security reasons? Same with `inputs.preid`, and other inputs across the workflows.

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable

.github/workflows/callable-codeql.yml
```diff
@@ -2,8 +2,7 @@
 name: 'CodeQL'
 
 on:
-  push:
-  pull_request:
+  workflow_call:
```
wlee221, line 5: Nit: think lines 4-5 can be just 

```
on: workflow_call
```

if you're interested in shorter syntax.

.github/workflows/callable-e2e-runner.yml
```diff
@@ -0,0 +1,75 @@
...
+on:
+  workflow_call:
+
+jobs:
+  e2e-prep:
+    name: Get required configs to run e2e tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout AmplifyJs
+        uses: actions/checkout@24cb9080177205b6e8c946b17badbe402adc938f
```
wlee221, line 12: Just for future readers, can we annotate which version of checkout this is? Same with other workflows.

.github/workflows/callable-e2e-tests-headless.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: |
+          yarn link amazon-cognito-identity-js @aws-amplify/core @aws-amplify/cache @aws-amplify/auth
+        working-directory: amplify-js-samples-staging
+
+      - name: Run cypress tests for ${{ inputs.test_name }}
+        run: |
+          ../amplify-js/.circleci/retry-yarn-script.sh -s \
+          "cypress run \          
+          --headless \
+          --spec ./cypress/integration/${{ inputs.category }}/${{ inputs.spec }}.spec.js \
```
wlee221, line 58: Let's use environment variables and sanitization if necessary. Same for other inputs/context that we use in other workflows.

.github/workflows/callable-get-package-list.yml
```diff
@@ -0,0 +1,42 @@
+name: 'Capture Package Names List to Outputs'
```
wlee221, line 1: Non-blocking nit: think this makes more sense to be a composite action than a reusable workflow.

.github/workflows/callable-e2e-tests-headless.yml
```diff
@@ -0,0 +1,72 @@
...
+          --config baseUrl=https://aws.amazon.com/" \
+          -n ${{ inputs.retry_count }}
+        working-directory: amplify-js-samples-staging
+        shell: bash
+
+      - name: upload artifact
```
wlee221, line 64: nit (based on your other code):

```suggestion
      - name: Upload artifact
```

.github/workflows/callable-e2e-tests.yml
ashwinkumar6: Changing to

```
  NPM_REGISTRY: http://localhost:4873/
  NPM_USER: verdaccio
  NPM_PASS: verdaccio
  NPM_EMAIL: verdaccio@amplify.js
```

.github/workflows/push-integ-test.yml
```diff
@@ -0,0 +1,12 @@
+on:
+  push:
+    branches:
+      - feat/github-actions/main
+      - feat/github-actions/test-stocaaro
+      - feat/github-actions/test-ashwinkumar6
+      - feat/github-actions/test-yingku
```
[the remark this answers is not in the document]
ashwinkumar6: Yep, this is just for the current development purpose. It will get cleaned up

.github/workflows/callable-e2e-tests-headless.yml
ashwinkumar6: thank you

.github/workflows/callable-npm-publish-preid.yml
ashwinkumar6: Updated to
```
git config --global user.email $GITHUB_EMAIL
git config --global user.name $GITHUB_USER
```

.github/actions/load-verdaccio-with-amplify-js/action.yml
```diff
@@ -0,0 +1,33 @@
...
+        sleep 1
+        yarn config set registry $NPM_REGISTRY
+        npm set registry $NPM_REGISTRY
+        git config --global user.email $NPM_EMAIL
+        git config --global user.name $NPM_USER
+        git status
+        git --no-pager diff
+        .circleci/retry-yarn-script.sh -s publish:verdaccio -n 5 -r true
+        yarn info aws-amplify@unstable description
+        npm info aws-amplify@unstable version
```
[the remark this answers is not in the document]
ashwinkumar6: makes sense, splitting this into 3 steps
- Install and run npm-cli-login
- Configure registry and git
- Publish to verdaccio

.github/workflows/callable-npm-publish-release.yml
```diff
@@ -0,0 +1,65 @@
...
+
+      - name: Run npm publish
+        uses: ./amplify-js/.github/actions/npm-publish
+        with:
+          github_email: aws@amazon.com
+          github_name: aws-amplify-bot
+          target: release
+          npm_token: ${{ secrets.NPM_TOKEN }}
+
+      - name: 'Update roll-back infrastructure'
```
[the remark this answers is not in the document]
ashwinkumar6: makes sense, removing '' from `name`
Adding '' to `name` only if absolutely necessary

.github/workflows/push-preid-release.yml
```diff
@@ -0,0 +1,32 @@
...
+    outputs:
+      preid: ${{ steps.output_preid.outputs.preid }}
+
+  preid-release:
+    needs:
+      - e2e
+      - parse-preid
+    secrets: inherit
+    uses: ./.github/workflows/callable-npm-publish-preid.yml
+    # The preid should be detected from th branch name recommending feat/PREID/whatvever as branch naming pattern
```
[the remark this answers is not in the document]
ashwinkumar6: Thank you

.github/workflows/callable-bundle-size-tests.yml
ashwinkumar6: Makes sense, removing '' from `name`

.github/workflows/callable-e2e-runner.yml
ashwinkumar6: Adding version annotation in all places to keep it uniform

.github/actions/node-and-build/action.yml
```diff
@@ -0,0 +1,46 @@
...
+  steps:
+    - name: Setup Node.js 16
+      uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0 https://github.com/actions/setup-node/commit/64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c
+      with:
+        node-version: 16
+      env:
+        SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
+    - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
+      id: cache-build-artifacts
+      with:
```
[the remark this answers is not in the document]
ashwinkumar6: Thanks for pointing it out.
Added the following to all the places where we're using `actions/cache`
```
        env:
          SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
```

.github/workflows/callable-e2e-tests.yml
```diff
@@ -0,0 +1,111 @@
...
+          ${{ inputs.spec }} \
+          ${{ matrix.browser }} \
+          prod \
+          ${{ inputs.amplifyjs_dir == true && env.AMPLIFY_DIR || ''}}" \
+          -n ${{ inputs.retry_count }}
+        working-directory: amplify-js-samples-staging
+        shell: bash
+      - name: upload artifact
+        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2 https://github.com/actions/upload-artifact/commit/0b7f8abb1508181956e8e162db84b466c27e18ce
+        if: always()
```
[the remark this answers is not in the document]
ashwinkumar6: this is mirroring the same behavior as CCI.
But yes, i agree it's better if we upload artifact only failure runs
Updated

.github/workflows/callable-e2e-tests.yml
```diff
@@ -0,0 +1,111 @@
...
+        with:
+          name: e2e-test-artifacts
+          if-no-files-found: ignore
```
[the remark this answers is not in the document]
ashwinkumar6: good point, added a retention period of 2weeks
`          retention-days: 14
`

.github/workflows/callable-get-package-list.yml
stocaaro: Would be interested in catching up with you about this one. In this case, we are retrieving the package list so that we can run unit tests for all packages in parallel on separate runners.

By retrieving the package list here, in a runner, we can immediately put it through a matrix to fan out to one runner per catagory.  If I move this into an action, then I still need a workflow to call it that can be in the `needs` list of the  unit test matrix workflow, right?

Maybe there's a clever way to get around this that I'm not coming up with. Happy to consider options, but I'm not clear on what I get by wrapping this behavior in an action which will then need to be called in a workflow.

.github/actions/setup-samples-staging/action.yml
ashwinkumar6: makes sense
renaming `GH_ACTIONS_STAGING_READ`  to `GITHUB_TOKEN_STAGING_READ`

.github/actions/setup-samples-staging/action.yml
ashwinkumar6: good point,
we have made a change to only cache the node_modules

.github/workflows/callable-get-package-list.yml
wlee221: That's very fair, thanks for the context. That makes sense, and I don't feel strongly anymore.

One idea that you can consider is moving `callable-get-package-list.yml` logic inside `callable-unit-tests`, since it looks like from usage that `get-package-list` is only used for unit tests. I think that might make it nicer since we don't have to call `callable-get-package-list.yml` every time `callable-unit-tests` is called (4 places currently). I think this might be a better abstraction for `unit-tests` too, since it'll be a bit smarter now to do package filtering itself. In general I think actions should be dumber and workflows should be smarter.

Just for context, the reason I thought this doesn't need to be its own workflow is because (1) it just runs one task without parallization, and (2) it doesn't need its own runner. But again, don't feel strongly and can be convinced!

stocaaro: You where right that moving this into the callable-unit-tests and moving the fanout matrix into that file cleans a lot of things up. I have made that change.

I still don't know how to get away from having a single runner make a packages list (or pull it from cache) and store that for use by the strategy/matrix to fan out the unit tests.  Pretty happy with the updated factoring from your recommendations. Interested in any further thoughts you have.

.github/workflows/callable-codeql.yml
stocaaro: Have gone through and made this update in a number of places.

.github/workflows/callable-npm-publish-preid.yml
stocaaro: Fixed an issue using vars and tested this morning.

.github/actions/setup-samples-staging/action.yml
```diff
@@ -0,0 +1,33 @@
...
+          /home/runner/.cache/Cypress
+    - name: Checkout staging repo
+      if: steps.cache-samples-staging-build.outputs.cache-hit != 'true'
+      uses: actions/checkout@24cb9080177205b6e8c946b17badbe402adc938f
+      with:
+        repository: stocaaro/amplify-js-samples-staging
```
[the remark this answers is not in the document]
stocaaro: Updated slightly to use the local to amplify-js staging repo. This no longer needs changes to merge.

.github/workflows/callable-get-package-list.yml
wlee221: Thanks! Yeah, I can't think of a way to avoid running `get-package-list` in a separate runner. Usually, I would try to combine these one-off set up to a single job, but this obviously isn't possible since this is the only setup task we need. 

I'm good with where we are, thanks!

.github/workflows/push-preid-release.yml
```diff
@@ -0,0 +1,32 @@
...
+  e2e:
+    secrets: inherit
+    uses: ./.github/workflows/callable-e2e.yml
+  parse-preid:
+    name: 'Parse preid from branch'
+    runs-on: ubuntu-latest
+    env:
+      BRANCH: ${{ github.ref_name }}
+    steps:
+      - id: output_preid
```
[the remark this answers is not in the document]
stocaaro: I looked for a validation procedure to apply on the env variables, but there are no GH functions to perform regex matching or similar (closest is `startsWith`).  Below, I added character filtering against the env variable `echo`.

.github/workflows/push-preid-release.yml
```diff
@@ -0,0 +1,32 @@
+name: Push - release for feature preid
+
+on:
+  push:
+    branches:
+      # Change this to your branch name where "example-preid" corresponds to the preid you want your changes released on
+      - feat/example-preid/main
```
stocaaro, line 7: Got feedback that this workflow should be proofed against tags/preids that are used elsewhere such as `latest`, `unstable` and `next`.

.github/workflows/push-latest-release.yml
```diff
@@ -0,0 +1,18 @@
+name: Push - release from release to latest
+
+on:
+  push:
+    branches:
```
stocaaro, line 5: Each of these push release workflows needs concurrency protection.

.github/workflows/callable-e2e-test.yml
```diff
@@ -0,0 +1,123 @@
...
+          prod \
+          $E2E_AMPLIFY_JS_DIR" \
+          -n $E2E_RETRY_COUNT
+      - name: Upload artifact
+        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2 https://github.com/actions/upload-artifact/commit/0b7f8abb1508181956e8e162db84b466c27e18ce
+        if: failure()
+        with:
+          name: ${{ inputs.test_name }}
+          if-no-files-found: ignore
+          path: |
```
ragingsquirrel3, line 120: Do you have an example job with these outputs? I see https://github.com/stocaaro/amplify-js/actions/runs/5613369597 but not sure this matches based on paths here.

.github/workflows/callable-get-package-list.yml
```diff
@@ -0,0 +1,41 @@
+name: Capture Package Names List to Outputs
```
ragingsquirrel3, line 1: question: What's the package list for? Looks like prereq to unit tests but I don't understand what/why? might suggest adding comment/description with this answer

.github/workflows/callable-npm-publish-preid.yml
```diff
@@ -0,0 +1,40 @@
...
+on:
+  workflow_call:
+    inputs:
+      preid:
+        required: false
+        type: string
+
+# TODO: configure these env variables
+env:
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ragingsquirrel3, line 12: question: Is this another GH token manually added to secrets for publish permission?

.github/workflows/callable-npm-publish-release.yml
```diff
@@ -0,0 +1,60 @@
...
+jobs:
+  deploy:
+    name: Publish to Amplify Package
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@24cb9080177205b6e8c946b17badbe402adc938f # v3.4.0 https://github.com/actions/checkout/commit/24cb9080177205b6e8c946b17badbe402adc938f
+        with:
+          path: amplify-js
+          token: ${{ secrets.GH_TOKEN_AMPLIFY_JS_WRITE }}
```
ragingsquirrel3, line 14: question: Is this what allows `git push ..` below (not the token for the workflow itself)?

.github/workflows/callable-release-verification.yml
```diff
@@ -0,0 +1,30 @@
...
+  prebuild-ubuntu:
+    uses: ./.github/workflows/callable-prebuild-amplify-js.yml
+    with:
+      runs_on: ubuntu-latest
+  prebuild-macos:
+    uses: ./.github/workflows/callable-prebuild-amplify-js.yml
+    with:
+      runs_on: macos-latest
+  prebuild-samples-staging:
+    secrets: inherit
```
ragingsquirrel3, line 15: nit/callout: I didn't know about this syntax but seems like it could lead to some unintended leaking of secrets to workflow code that doesn't need them. I'm sure best judgement is used given it saves a lot of code.

.github/workflows/codeql.yml
```diff
@@ -1,9 +1,13 @@
...
-name: 'CodeQL'
+name: CodeQL
 
 on:
   push:
+    branches: ['*']
   pull_request:
+    branches: ['main']
+  schedule:
+    - cron: '28 21 * * 2'
```
ragingsquirrel3, line 10: nit: I like to always have comment on these saying what it means bc I don't speak cron.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,39 @@
+name: PR Validation
+
+concurrency:
+  group: pr-${{ github.event.pull_request.number }}-${{ github.event.sender.login }} # This is to make the group name unique for a PR.
+  cancel-in-progress: true
+
+on: pull_request
+
```
ragingsquirrel3, line 8: suggestion: Depending on repo settings for a) if forked repo PRs can run workflows b) default permissions, I might suggest explicitly limiting permissions for PR workflow. In amplify-flutter we allow only "read-all" [example](https://github.com/aws-amplify/amplify-flutter/blob/main/.github/workflows/amplify_auth_cognito.yaml#L55) as an extra mitigation.

.github/workflows/push-main-release.yml
```diff
@@ -0,0 +1,18 @@
+name: Push - release from main to unstable
```
ragingsquirrel3, line 1: nit: dk if possible to combine this with push-latest-release and use the branch name as a variable, but ... also sometimes duplicating code is easier.

.github/workflows/callable-e2e-test.yml
stocaaro: The job you link matches. The original version used a shared name across the board and I split it based on test name to make it more navigable.

.github/workflows/callable-get-package-list.yml
stocaaro: The unit tests run in ~8 minutes locally on my Mac by running in parallel. My first test runs in gh took ~20 minutes. To improve the speed, I have split the runs per category across different runners. I was worried about keeping the category list hardcoded, so this allows me to derive the list of categories and use that as a matrix input to ensure that as we add/remove categories they will automatically get tested.

.github/workflows/callable-npm-publish-release.yml
stocaaro: The GH token permitted `git push`, but didn't have permission to merge to the `main` branch. I tested with different permission settings, but was only able to get the push to main working using this token approach.  Would be happy to adopt a better pattern if you've got one.

.github/workflows/callable-release-verification.yml
stocaaro: I've gone back and forth on this. I agree that limiting which secrets are shared down to lower layers observes "least privilege" better, however in this case the risk mitigation is that all workflow executions from external users require team approval, so its not clear to me that the added code/convenience achieves an improved security result.

Since I can make a case for and against, I thought it would be good to leave it this way to get input from AppSec. I've got work staged migrating it, so I'm ready if this comes through as a finding.

.github/workflows/codeql.yml
stocaaro: Good call

.github/workflows/pr.yml
stocaaro: I can get behind this. Making the change

.github/workflows/push-main-release.yml
stocaaro: I get what you mean. This is closer the the pre-id workflow.  For clarity, I have the different entry point workflows separated with duplicate code pushed down into callable workflows where the small amount of redundant calling adds clear entry points for users to trace forward from.

If there where problems with overlapping runs because of complex selection logic or some other driver that made a single workflow for multiple purposes better, I could get behind that. In this case, I like how simple each of these entry points are and am intentionally tuning for that.

.github/workflows/callable-e2e-test-detox.yml
```diff
@@ -0,0 +1,77 @@
...
+          $GITHUB_WORKSPACE/amplify-js/scripts/retry-yarn-script.sh -s 'install --frozen-lockfile --non-interactive' -n 3
+        shell: bash
+      - name: Install CocoaPods
+        run: |
+          cd ios && pod install
+        working-directory: ${{ inputs.working_directory }}
+        shell: bash
+      - name: Start iOS simulator (background)
+        run: |
+          xcrun simctl boot "iPhone 11" &
```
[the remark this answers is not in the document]
stocaaro: There are for sure some oddities with the execution of these. I've noticed long / inconsistent detox builds and I'm not sure if thats related the the simulator booting or not. If it weren't working in most cases, this is one of a few places I could see us investing more time in tuning.

.github/workflows/callable-bundle-size-tests.yml
```diff
@@ -0,0 +1,18 @@
+name: Unit Tests
```
jimblanc, line 1: Is this name accurate?

.github/workflows/callable-npm-publish-release.yml
```diff
@@ -0,0 +1,55 @@
+name: Publish package to npm and create release on github
```
jimblanc, line 1: Might be missing it, but does this actually create the release on github?

.github/workflows/callable-prebuild-samples-staging.yml
```diff
@@ -0,0 +1,20 @@
+name: Prebuild AmplifyJs
```
jimblanc, line 1:
```suggestion
name: Prebuild Amplify Integration Test Package
```

(or Samples Staging)

jimblanc (dismissed):
Great work, some nits below but looks good 🔥 

The yarn.lock file! 🤩

.github/workflows/callable-prebuild-samples-staging.yml
stocaaro: On it.

.github/workflows/push-next-release.yml
```diff
@@ -17,7 +17,7 @@ jobs:
   e2e:
     secrets: inherit
     uses: ./.github/workflows/callable-release-verification.yml
-  unstable-release:
+  next-release:
```
jimblanc, line 20: Good catch

jimblanc (approved):

PR #11647, opened 2023-07-18
feat(auth): add fetchMFAPreference API

israx:
#### Description of changes

fetchMFAPreference API is used to get the user's MFA settings such as preferred MFA or enabled MFA options. This API is meant to be a cognito specific and mainly used after the user is authenticated

**TODO List:**
- Get accessToken from auth token provider once implementation is complete

#### Issue #, if available

Created a sample app where I was able to get the expected MFA settings for the user

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/src/providers/cognito/apis/fetchMFAPreference.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { FetchMFAPreferenceRequest } from '../types/requests';
+import { getUserClient } from '../utils/clients/GetUserClient';
+import { getMFAType, getMFATypes } from '../utils/signInHelpers';
+import { GetUserException } from '../types/errors';
+
+/**
+ * Fetches the preferred MFA setting and enabled MFA settings for the user.
+ * @throws  -{@link GetUserException} : error thrown when the service fails to fetch MFA preference
+ * and settings.
```
jimblanc, line 12:
```suggestion
 *
 * @throws {@link GetUserException}
 * Thrown when the service fails to fetch MFA preference and settings.
 *
```

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -552,22 +554,28 @@ export function mapMfaType(mfa: string): CognitoMFAType {
...
 }
 
-export function parseMFATypes(mfa?: string): AllowedMFATypes {
+export function getMFAType(type?: string): MFAType | undefined {
+	if (type === 'SMS_MFA') return 'SMS';
+	if (type === 'SOFTWARE_TOKEN_MFA') return 'TOTP';
+}
+
+export function getMFATypes(types?: string[]): MFAType[] {
+	if (!types) return [];
```
jimblanc, line 563: Nit: For consistency with `getMFAType`, I wonder if we should return `undefined`?

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -552,22 +554,28 @@ export function mapMfaType(mfa: string): CognitoMFAType {
...
+	return types.map(getMFAType).filter(Boolean) as MFAType[];
```
jimblanc, line 564: Might not be required, but should we log something if an invalid/unknown MFA type is encountered? (maybe in `getMFAType`)

jimblanc (approved):
Lookin good, couple nits & questions below.

packages/auth/src/providers/cognito/apis/fetchMFAPreference.ts
israx: type doc weird intellisense as discussed on a previous PR

packages/auth/src/providers/cognito/utils/signInHelpers.ts
israx: good catch, addressing it

packages/auth/src/providers/cognito/utils/signInHelpers.ts
israx: yeah I added a TODO  inside `getMFAType`  to log any unknown MFA type

jimblanc (dismissed):

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -351,8 +351,8 @@ export async function getSignInResult(params: {
 			if (!isMFATypeEnabled(challengeParameters, 'TOTP'))
 				throw new AuthError({
 					name: AuthErrorCodes.SignInException,
-					message: `Cannot initiate MFA setup from available types: ${parseMFATypes(
-						challengeParameters.MFAS_CAN_SETUP
+					message: `Cannot initiate MFA setup from available types: ${getMFATypes(
+						parseMFATypes(challengeParameters.MFAS_CAN_SETUP)
```
elorzafe, line 355: This seems like by design you cannot have that flow, if the client has a challenge for setup totp is because TOTP is an MFA method. I think this is an edge case that should never happen, if it happens something is very bad server side

packages/auth/src/providers/cognito/utils/clients/GetUserClient.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import type {
+	GetUserCommandInput,
+	GetUserCommandOutput,
+} from '@aws-sdk/client-cognito-identity-provider';
```
elorzafe, line 7: I would add a TODO here for remove this to the internal type

elorzafe (approved):
Just add a TODO for the sdk types

packages/auth/src/providers/cognito/utils/signInHelpers.ts
israx: This should never happen unless cognito supports a 3red MFA option. The actual challengeName you get is `MFA_SETUP`, which means that is used to set up an MFA option, as now, TOTP is the only option that can be setup, in the feature could be more alternatives

jimblanc (approved):

packages/auth/src/providers/cognito/utils/clients/GetUserClient.ts
israx: I added a TODO to vend AWS SDK types

jimblanc (approved):

elorzafe (approved):
Looks good!

Thanks @israx

PR #11660, opened 2023-07-20
feat(auth): credentials provider

Samaritan1011001:
#### Description of changes

WIP: Adds the following,
- Credentials provider interface
- Cognito credentials provider 
- IdentityId manager
- Cognito Guest credentials
- Cognito Credentials for OIDC tokens

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/core/src/singleton/Auth/types.ts
```diff
@@ -36,6 +36,12 @@ export const AuthStorageKeys = {
 	oidcProvider: 'oidcProvider',
 	clockDrift: 'clockDrift',
 	metadata: 'metadata',
+	accessKeyId: 'accessKeyId',
+	secretAccessKey: 'secretAccessKey',
+	sessionToken: 'sessionToken',
+	expiration: 'expiration',
```
elorzafe, line 42: I would make this as a separate type, this is coupled with the function `getAuthStorageKeys`. Maybe we could rename this to `AuthTokenStorageKeys`

packages/core/src/singleton/Auth/types.ts
```diff
@@ -36,6 +36,12 @@ export const AuthStorageKeys = {
...
+	identityId: 'identityId',
```
elorzafe, line 43: and maybe keep the `identityId` here, and this is only for guest identity

packages/core/src/singleton/Auth/IdentityIdStore.ts
```diff
@@ -0,0 +1,127 @@
...
+) {
+	const keys = Object.values({ ...authKeys });
+	return (prefix: string, identifier: string) =>
+		keys.reduce(
+			(acc, authKey) => ({
+				...acc,
+				[authKey]: `${prefix}.${identifier}.${authKey}`,
+			}),
+			{} as AuthKeys<keyof T & string>
+		);
```
elorzafe, line 124: I would use another type for `identityId` stuff

packages/core/src/singleton/Auth/CredentialStore.ts
```diff
@@ -0,0 +1,162 @@
...
+	const keys = Object.values({ ...authKeys });
+	return (prefix: string, identifier: string) =>
+		keys.reduce(
+			(acc, authKey) => ({
+				...acc,
+				[authKey]: `${prefix}.${identifier}.${authKey}`,
+			}),
+			{} as AuthKeys<keyof T & string>
+		);
+}
```
elorzafe, line 162: Similar to the other one this function could be rename because is the same logic.

packages/core/src/singleton/Auth/CredentialStore.ts
```diff
@@ -0,0 +1,162 @@
...
+	AuthKeys,
+	AuthStorageKeys,
+} from './types';
+import { KeyValueStorageInterface } from '../../types';
+import { asserts } from '../../Util/errors/AssertError';
+import { Credentials } from '@aws-sdk/types';
+import { assertCredentialsProviderConfig } from './utils';
+
+export class DefaultCredentialStore implements AuthCredentialStore {
+	keyValueStorage: KeyValueStorageInterface;
```
elorzafe, line 16: I think this is a little over engineered, you can define the properties that you want to persist on the instance instead of using `KeyValueStorageInterface`

packages/auth/src/providers/cognito/utils/clients/HttpClients.ts
```diff
@@ -65,20 +75,34 @@ export type ClientOperations =
...
+
+export type ServiceTypes = 'identity' | 'idp';
+
+type Services =
+	| 'AWSCognitoIdentityProviderService'
+	| 'AWSCognitoIdentityService';
 
 export class UserPoolHttpClient {
 	private _endpoint: string;
+	private _service: Services = 'AWSCognitoIdentityProviderService';
```
elorzafe, line 90: I would not use the UserPoolHttpClient for IdentityPools, I think is better to have another client.

packages/auth/src/providers/cognito/apis/credentialsProvider.ts
```diff
@@ -0,0 +1,256 @@
...
+		return isExp;
+	}
+
+	private _isPastTTL(): boolean {
+		return this._nextCredentialsRefresh <= Date.now();
+	}
+}
+
+export function formLoginsMap(idToken: string, oidcProvider: string) {
+	// TODO(V6): Update with Amplify.config values
```
elorzafe, line 230: You can use `Amplify.getConfig().Auth`

elorzafe: I didnt consider the `region` on the config, I think is redundant because is on the `userPoolId`

elorzafe (changes requested):
Thanks @Samaritan1011001 you are getting close, we can discuss some of the changes offline

Can you add default providers to the singleton https://github.com/aws-amplify/amplify-js/blob/next/packages/core/src/singleton/index.ts#L23

packages/auth/src/providers/cognito/apis/IdentityIdProvider.ts
```diff
@@ -0,0 +1,103 @@
...
+	authConfig,
+}: {
+	tokens?: AuthTokens;
+	authConfig?: AuthConfig;
+}): Promise<string> {
+	if (authConfig) defaultIdentityIdStore.setAuthConfig(authConfig);
+	let identityId = await defaultIdentityIdStore.loadIdentityId();
+
+	if (tokens) {
+		// Tokens are avaliable so retrun primary identityId
```
israx, line 28:
```suggestion
// Tokens are available so return primary identityId
```

packages/auth/src/providers/cognito/utils/clients/HttpClients.ts
```diff
@@ -59,6 +67,8 @@ export type ClientOutputs =
 	| ConfirmSignUpCommandOutput
 	| VerifySoftwareTokenCommandOutput
 	| AssociateSoftwareTokenCommandOutput
+	| GetIdCommandOutput
+	| GetCredentialsForIdentityCommandOutput
```
israx, line 71: the `UserPoolHttpClient` is only taking user pool commands, thus  shouldn't be mixed identity pool commands.

packages/auth/src/providers/cognito/utils/clients/HttpClients.ts
```diff
@@ -73,20 +83,27 @@ export type ClientOperations =
 	| 'ResendConfirmationCode'
 	| 'VerifySoftwareToken'
 	| 'AssociateSoftwareToken'
+	| 'GetId'
+	| 'GetCredentialsForIdentity'
```
israx, line 87: same as above

packages/auth/src/providers/cognito/utils/clients/IdentityIdForPoolIdClient.ts
```diff
@@ -0,0 +1,25 @@
...
+): Promise<GetIdCommandOutput> {
+	const authConfig = AmplifyV6.getConfig().Auth;
+	const client = new IdentityPoolHttpClient(authConfig);
+	const result: GetIdCommandOutput = await client.send<GetIdCommandOutput>(
+		'GetId',
+		{
+			...params,
+		}
+	);
+	return result;
```
israx, line 24:
```suggestion
return client.send<GetIdCommandOutput>(
		'GetId',
		{
			...params,
		}
	);

```

packages/auth/src/providers/cognito/utils/types.ts
```diff
@@ -9,3 +13,12 @@ export function isTypeUserPoolConfig(
...
 	return false;
 }
+
+export function isTypeIdentityPoolConfig(
+	authConfig?: AuthConfig
+): authConfig is IdentityPoolConfig {
+	if (authConfig && authConfig.identityPoolId) {
+		return true;
+	}
+	return false;
```
israx, line 23: we can simplify this 
``` suggestion
return authConfig && authConfig.identityPoolId
```

packages/core/src/singleton/Auth/IdentityIdStore.ts
```diff
@@ -0,0 +1,127 @@
...
+			}
+		} catch (err) {
+			// TODO(v6): validate partial results with mobile implementation
+			throw new Error(`Error loading identityId from storage: ${err}`);
+		}
+	}
+	async storeIdentityId(identity: Identity): Promise<void> {
+		assertIdentityIdProviderConfig(this.authConfig);
+		asserts(!(identity === undefined), {
+			message: 'Invalid Identity',
```
israx, line 72: what does `Identity` mean here. Will a customer be able to troubleshoot with this message? Probably adding more context

packages/core/src/singleton/Auth/IdentityIdStore.ts
```diff
@@ -0,0 +1,127 @@
...
+			this.inMemoryStorage.setItem(authKeys.identityId, identity.id);
+		} else {
+			this.localStorage.setItem(authKeys.identityId, identity.id);
+		}
+	}
+
+	async clearIdentityId(): Promise<void> {
+		assertIdentityIdProviderConfig(this.authConfig);
+
+		const name = 'Cognito'; // TODO(v6): update after API review for Amplify.configure
```
israx, line 92: I see `name` is repeated in some methods, probably adding it as a data member in the class

israx (changes requested):
Thanks for putting all this together Manoj. 

My main concern is that I think we are not using the custom identity client or am I missing something ? 

Also we should try to be consistent with errors. If the error is coming from the `auth` category, it should be `AuthError`

elorzafe (changes requested):

packages/auth/src/providers/cognito/apis/IdentityIdProvider.ts
```diff
@@ -0,0 +1,103 @@
...
+			//TODO(V6): clear guest id in local storage
+		}
+	} else {
+		// Tokens are avaliable so return guest identityId
+		if (identityId && identityId.type === 'guest') {
+			return identityId.id;
+		} else {
+			identityId = {
+				id: await generateIdentityId({}, authConfig),
+				type: 'guest',
```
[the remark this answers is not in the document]
Samaritan1011001: Didn't quite understand this suggestion

packages/auth/src/providers/cognito/apis/credentialsProvider.ts
```diff
@@ -0,0 +1,242 @@
...
+	const amplifyConfig = {
+		region: 'us-east-2',
+		userPoolId: 'us-east-2_Q4ii7edTI',
+	};
```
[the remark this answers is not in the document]
Samaritan1011001: Oh shoot, good catch!

packages/auth/src/providers/cognito/utils/clients/HttpClients.ts
Samaritan1011001: These are common client inputs and outputs that userPoolClient and identityPoolClient uses. Do we want to rename this to UserPoolClientInputs and create a new IdentityPoolClientInputs?

packages/core/src/singleton/Auth/IdentityIdStore.ts
Samaritan1011001: This is only internal usage and is not exported to customers. But I agree, will update the message.

packages/core/src/singleton/Auth/IdentityIdStore.ts
Samaritan1011001: That will be updated as the TODO says

packages/core/src/singleton/index.ts
```diff
@@ -1,4 +1,8 @@
 import { AuthClass } from './Auth';
+import {
+	cognitoCredentialsProvider,
+	getIdentityId,
+} from '@aws-amplify/auth/cognito';
```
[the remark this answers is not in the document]
Samaritan1011001: Discussed offline and this is longer needed to be done.

packages/auth/__tests__/providers/cognito/credentialsProvider.test.ts
```diff
@@ -3,44 +3,48 @@
 
 import { cognitoCredentialsProvider } from '../../../src/providers/cognito';
 import { authAPITestParams } from './testUtils/authApiTestParams';
-
-import * as client from '../../../src/providers/cognito/utils/clients/CredentialsForIdentityIdClient';
-import { GetCredentialsForIdentityCommandOutput } from '@aws-sdk/client-cognito-identity';
+import {
+	GetCredentialsForIdentityCommandOutput,
```
israx, line 7: can we import this types from core instead ?

israx (commented):
Credential provider changes look good to me. I just have 2 general concerns:

1. We should not import types from the AWS SDK at this point, I think all types are already vended.

2. We can change the location of `IdenityIdProvider.ts` and `credentialsProvider.ts` to their own directory inside `cognito` and then export them from `index.ts` . Something similar how the TokenProvider is doing it.

packages/auth/src/providers/cognito/apis/IdentityIdProvider.ts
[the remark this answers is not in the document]
israx: yeah that was just to avoid passing the value guest, I think at that point type is already 'guest' based on the assertion above

packages/core/src/index.ts
```diff
@@ -114,19 +114,36 @@ export { InMemoryCache } from './Cache/InMemoryCache';
...
+	GetCredentialsOptions,
 } from './singleton/types';
 
+//AWSClients exports
+export {
+	getCredentialsForIdentity,
+	getId,
+	GetCredentialsForIdentityInput,
+	GetCredentialsForIdentityOutput,
+} from './AwsClients/CognitoIdentity';
```
elorzafe, line 145: I think need to be exported from `internals`

packages/core/src/singleton/Auth/types.ts
```diff
@@ -74,11 +73,13 @@ export type AuthConfig =
...
 	| UserPoolConfigAndIdentityPoolConfig;
 
-type IdentityPoolConfig = {
+export type IdentityPoolConfig = {
 	identityPoolId: string;
 	userPoolWebClientId?: never;
 	userPoolId?: never;
 	clientMetadata?: never;
+	isMandatorySignInEnabled?: never;
+	oidcProvider?: never;
```
elorzafe, line 82: it seems the other are incorrect

elorzafe: why `oidcProvider`?

packages/core/src/singleton/Auth/utils/index.ts
```diff
@@ -14,7 +14,7 @@ export function assertTokenProviderConfig(
 	});
 }
 
-export function assertCredentialsProviderConfig(authConfig: AuthConfig) {
+export function assertIdentityPoolIdInConfig(authConfig: AuthConfig) {
```
elorzafe, line 17: you can add the same as assertTokenProviderConfig that, `asserts authConfig is IdentityPoolConfig`

that way the code after that can safely use the config with the assertion guard

packages/auth/src/providers/cognito/utils/clients/HttpClients.ts
```diff
@@ -79,14 +79,19 @@ export type ClientOperations =
...
 		'Cache-Control': 'no-store',
 	};
 
 	constructor(authConfig?: AuthConfig) {
-		if (authConfig && isTypeUserPoolConfig(authConfig)) {
+		if (
+			authConfig &&
+			isTypeUserPoolConfig(authConfig) &&
+			authConfig.userPoolId
+		) {
```
elorzafe, line 94: why you need to change this?

packages/auth/src/providers/cognito/credentialsProvider/IdentityIdStore.ts
```diff
@@ -0,0 +1,137 @@
...
+	const keys = Object.values({ ...authKeys });
+	return (prefix: string, identifier: string) =>
+		keys.reduce(
+			(acc, authKey) => ({
+				...acc,
+				[authKey]: `${prefix}.${identifier}.${authKey}`,
+			}),
+			{} as AuthKeys<keyof T & string>
+		);
+}
```
elorzafe, line 137: I would reuse the other for TokenStore, create a util that can be share between those.

packages/auth/src/providers/cognito/credentialsProvider/credentialsProvider.ts
```diff
@@ -0,0 +1,280 @@
...
+		}
+	}
+
+	private async getGuestCredentials(
+		identityId: string,
+		authConfig: UserPoolConfigAndIdentityPoolConfig
+	): Promise<AWSCredentialsAndIdentityId> {
+		if (
+			this._credentialsAndIdentityId &&
+			!this._isExpired(this._credentialsAndIdentityId.credentials) &&
```
elorzafe, line 76: I would not check this because if the clock is skewed it will refresh all the time, I would trus isPastTTL

packages/auth/src/providers/cognito/credentialsProvider/credentialsProvider.ts
```diff
@@ -0,0 +1,280 @@
...
+			https://github.com/aws/aws-sdk-js-v3/blob/v1.0.0-beta.1/packages/types/src/credentials.ts#L26
+		*/
+		const { expiration } = credentials;
+		// TODO(V6): when  there is no expiration should we consider it not expired?
+		if (!expiration) return false;
+		const expDate = new Date(Number.parseInt(expiration.toString()) * 1000);
+		const isExp = expDate.getTime() <= ts;
+		logger.debug('are the credentials expired?', isExp);
+		return isExp;
+	}
```
elorzafe, line 234: This will not be accurate if clock is skewed

elorzafe (approved):
Looks good! Thanks @Samaritan1011001 🎖️

jimblanc (commented):
Looks good! Just need to resolve merge conflicts

elorzafe (approved):
Looks good!

israx (approved):
lgtm

jimblanc (approved):

israx (approved):
LGTM!

jimblanc (approved):

elorzafe (approved):

PR #11676, opened 2023-07-25
feat(core): add storage mechanisms

israx:
#### Description of changes

Adds the following storage mechanisms:

LocalStorage
InMemoryStorage
SessionStorage
CookieStorage
Async Storage

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/core/src/StorageHelper/reactnative.ts
```diff
@@ -100,3 +101,78 @@ export class StorageHelper {
 		return this.storageWindow;
 	}
 }
+
+class AsyncStorageClass implements KeyValueStorageInterface {
```
elorzafe, line 105: I would add a `TODO` that this needs to be included on `react-native` Amplify package

packages/core/src/StorageHelper/index.ts
```diff
@@ -99,3 +104,202 @@ export class MemoryKeyValueStorage implements KeyValueStorageInterface {
...
+	 * This is used to clear the storage
+	 * @returns {string} nothing
+	 */
+	async clear(): Promise<void> {
+		if (!this.storage) throw PlatformNotSupportedError;
+		this.storage.clear();
+	}
+}
+
+export const SessionStorage = new SessionStorageClass();
```
elorzafe, line 305: What if this is an abstract class, the logic is the same except the constructor, what do you think?

packages/core/src/StorageHelper/index.ts
```diff
@@ -1,7 +1,12 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-
-import { KeyValueStorageInterface } from '../types';
+import * as Cookies from 'js-cookie';
```
elorzafe, line 3: Would this be tree shakable? maybe we need to add this to a different module?

packages/core/package.json
```diff
@@ -72,7 +72,8 @@
 		"tslib": "^2.5.0",
 		"universal-cookie": "^4.0.4",
 		"zen-observable-ts": "0.8.19",
-		"rxjs": "^7.8.1"
+		"rxjs": "^7.8.1",
+		"js-cookie": "3.0.5"
```
elorzafe, line 76: Do we need to add another dependency? could be possible to not have this?

elorzafe (changes requested):
Biggest concern is adding another dependency and maybe split the helpers in separate files just in case there are issue with tree shaking.

packages/core/package.json
israx: Yeah I think we discussed this in the design already. Pretty much `js-cookie` is to handle the overwhelming cookies operations. The package is pretty light weight and well supported: https://www.npmjs.com/package/js-cookie

packages/core/src/StorageHelper/index.ts
israx: I don't think an abstract class will be the right way. It seems to me that it is more like a hacky thing. I mean, yeah abstract classes can't be instantiated but they are mainly used for OOP scenarios. The other solution is just to have a class with static members.

elorzafe (dismissed):
Thanks @israx 

Good work 🏅

packages/core/package.json
jimblanc: Looks good but we might want to loosen up the dependency in-case customers have this in their graph as well

```suggestion
		"js-cookie": "^3.0.5"
```

packages/core/src/index.ts
```diff
@@ -43,7 +43,14 @@ export {
...
 } from './types';
-export { StorageHelper, MemoryStorage } from './StorageHelper';
+export {
+	StorageHelper,
+	MemoryStorage,
+	LocalStorage,
+	CookieStorage,
+	SessionStorage,
+	MemoryKeyValueStorage,
+} from './StorageHelper';
```
jimblanc, line 53: Do we need to export these from the umbrella package as well so customers can customize them?

Here: https://github.com/aws-amplify/amplify-js/blob/next/packages/aws-amplify/src/utils/index.ts (`aws-amplify/utils` route)

jimblanc (commented):
Looks good! Some small comments below

jimblanc (approved):
Looks good!

elorzafe (approved):

PR #11680, opened 2023-07-25
Integrate V6 singleton on Auth APIs

elorzafe:
#### Description of changes

Adds `AmplifyV6` singleton on new Auth category APIs

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -571,3 +581,47 @@ export function isMFATypeEnabled(
 
 	return isMFAparseMFATypes;
 }
+
+export async function cacheTokens(
```
israx, line 585: I'm wondering if this code should live in core. Seems to be tight

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -260,8 +267,11 @@ export async function handleCustomSRPAuthFlow(
 	password: string,
 	clientMetadata: ClientMetadata | undefined
 ) {
-	const userPoolId = Amplify.config['aws_user_pools_id'];
-	const userPoolName = userPoolId.split('_')[1];
+	const config = AmplifyV6.getConfig().Auth;
+	assertTokenProviderConfig(config);
+
+	const userPoolId = config?.userPoolId;
```
israx, line 273: I think at this point we know that the userPoolId will not be null.

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -260,8 +267,11 @@ export async function handleCustomSRPAuthFlow(
 	password: string,
 	clientMetadata: ClientMetadata | undefined
 ) {
-	const userPoolId = Amplify.config['aws_user_pools_id'];
-	const userPoolName = userPoolId.split('_')[1];
+	const config = AmplifyV6.getConfig().Auth;
```
israx, line 270: I think the `getConfig` function should assert the config. That way we have it as one source of truth

israx (commented):
PR looks good. Thanks Francisco for putting this together. 
I'm just wonder if the `Amplify.getConfig()` should assert for the configuration

packages/auth/src/providers/cognito/apis/confirmResetPassword.ts
```diff
@@ -26,13 +26,13 @@ export async function confirmResetPassword(
 		!!code,
 		AuthValidationErrorCode.EmptyConfirmResetPasswordConfirmationCode
 	);
-	const config = Amplify.config;
+	const authConfig = AmplifyV6.getConfig().Auth;
```
jimblanc, line 29: Unrelated to this PR but just wanted to call it out. Seeing `getConfig()` in practice makes me think we need some error handling if the Library hasn't been configured yet 🤔 Maybe just throw an exception with that standard "Not configured" error & recovery suggestion.

Also maybe some memoization on `getConfig` to avoid the expensive JSON parsing (TODO)

packages/auth/src/providers/cognito/apis/signInWithCustomAuth.ts
```diff
@@ -44,7 +45,8 @@ export async function signInWithCustomAuth(
 ): Promise<AuthSignInResult> {
 	const { username, password, options } = signInRequest;
 	const metadata =
-		options?.serviceOptions?.clientMetadata || Amplify.config.clientMetadata;
+		options?.serviceOptions?.clientMetadata ||
+		AmplifyV6.getConfig().Auth?.clientMetadata;
```
jimblanc, line 49: Could we standardize on `const authConfig = AmplifyV6.getConfig().Auth;` like the above APIs? (easier to read). Ditto for other APIs in this PR.

packages/auth/src/providers/cognito/apis/confirmSignUp.ts
```diff
@@ -32,8 +32,7 @@ export async function confirmSignUp(
 ): Promise<AuthSignUpResult<AuthStandardAttributeKey | CustomAttribute>> {
 	const { username, confirmationCode, options } = confirmSignUpRequest;
 
-	// TODO: replace by singleton implementation.
-	const config = Amplify.config;
+	const authConfig = AmplifyV6.getConfig().Auth;
```
jimblanc, line 35: Again probably out of scope for this PR, but if Auth hasn't been configured should we throw an error? Maybe we can have a little helper function to do this across APIs (with a similar pattern for other categories when we get there).

```
export const getAuthConfig(amplifySingleton) {
  const authConfig = AmplifyV6.getConfig().Auth;
  
  if (!authConfig) {
    throw new AmplifyError(...);
  }
  
  return authConfig;
}
```

packages/auth/src/providers/cognito/utils/clients/HttpClients.ts
```diff
@@ -73,8 +75,13 @@ export class UserPoolHttpClient {
 		'Cache-Control': 'no-store',
 	};
 
-	constructor(region: string) {
-		this._endpoint = `https://cognito-idp.${region}.amazonaws.com/`;
+	constructor(authConfig?: AuthConfig) {
```
jimblanc, line 78: IMO it seems better to keep `HttpClient` agnostic of our config and just pass in what it needs? Is there a good reason for it to require the whole config?

packages/auth/src/providers/cognito/utils/signInHelpers.ts
elorzafe: Is kind of specific to Cognito, maybe rename to `cacheCognitoTokens` is more accurate

packages/auth/src/providers/cognito/utils/signInHelpers.ts
elorzafe: Do you know how to do that in TypeScript, basically the assertion means is not null and has the `userPoolId` and `userPoolWebClientId` but that needs to be somehow expressed on the contract of the assertion, or with the type guard? any suggestions?

packages/auth/src/providers/cognito/utils/signInHelpers.ts
elorzafe: `getConfig` can end up be more flexible than what it is on this PR (depending on `Amplify.configure` review. the assertion needs to be made on runtime.

packages/auth/src/providers/cognito/apis/confirmResetPassword.ts
elorzafe: I see, is ok if I add a TODO?

packages/auth/src/providers/cognito/apis/signInWithCustomAuth.ts
elorzafe: sounds good, i will do that

packages/auth/src/providers/cognito/apis/confirmSignUp.ts
elorzafe: I like that idea, the thing is that in the end if we support different service implementation, we cannot enforce it on the Singleton, it should be on the implementation that is consuming the config

packages/auth/src/providers/cognito/utils/clients/HttpClients.ts
elorzafe: is a good question, for the custom endpoint it might need it, is not supported on this PR but eventually yes.

packages/auth/src/providers/cognito/utils/signInHelpers.ts
elorzafe: I think you are right, token refresh will need this as well, I will move it to core, thanks!

packages/auth/src/providers/cognito/apis/confirmResetPassword.ts
jimblanc: Okay with me!

packages/auth/src/providers/cognito/apis/confirmSignUp.ts
jimblanc: The example code I posted would live with the implementation, it would just get called in each API that requires the config.

jimblanc (approved):

israx (approved):
LGTM

PR #11701, opened 2023-07-31
Add refresh token for Cognito UserPools

elorzafe:
#### Description of changes

- Adds refresh token on `@aws-amplify/auth` 
- Upgrades TS version on `@aws-amplify/auth` to work with latest transfer client (from core)
- Missing unit tests to validate different cases (need to add some additional logic as well)

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/auth/src/providers/cognito/utils/refreshTokens.ts
```diff
@@ -0,0 +1,57 @@
...
+	const body = JSON.stringify(input);
+	return buildHttpRpcRequest(endpoint, headers, body);
+};
+
+const refreshTokenDeserializer = async (
+	response: HttpResponse
+): Promise<InitiateAuthCommandOutput> => {
+	if (response.statusCode >= 300) {
+		const error = await parseJsonError(response);
+		throw error;
```
israx, line 38: I think here we should map the service error into AuthError. We can use the `assertsServiceError` helper function that is already implemented

israx:
``` suggestion
		assertServiceError(error);
		throw new AuthError({ name: error.name, message: error.message });
```

packages/auth/src/providers/cognito/utils/refreshTokens.ts
```diff
@@ -0,0 +1,57 @@
...
+			AuthenticationResult: body.AuthenticationResult,
+			ChallengeParameters: body.ChallengeParameters,
+		};
+	}
+};
+
+/**
+ * @internal
+ */
+export const refreshToken = composeServiceApi(
```
israx, line 52: I think this should be part of the `clients` directory

israx (commented):
PR looks good. We should create a task to change the current clients by this approach.

PR #11715, opened 2023-08-03
feat(api): add alpha types package; custom id; custom sel set support

iartemiev:
#### Description of changes

Increased bundle size is from upgrading `tslib@1.x` -> `2.6`

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-v6-types-package/index.ts
```diff
@@ -0,0 +1,77 @@
...
+					: never;
+			} & {
+				[k in Paths[number] as k extends `${string}.${string}`
+					? never
+					: k]: M[k];
+			}
+	  >;
+
+type ModelIdentifier<Model extends Record<any, any>> = Prettify<
+	Record<Model['identifier'] & string, string>
```
manueliglesias, line 51: Identifier fields can be other types than string (most of the time `number`), is is possible to adjust this type to account for that?

packages/api/package.json
```diff
@@ -26,16 +26,16 @@
...
 		"test:size": "size-limit",
 		"build-with-test": "npm test && npm run build",
-		"build:cjs": "node ./build es5 && webpack && webpack --config ./webpack.config.dev.js",
-		"build:esm": "node ./build es6",
-		"build:cjs:watch": "node ./build es5 --watch",
-		"build:esm:watch": "node ./build es6 --watch",
+		"build:cjs": "rimraf lib && tsc -p ./tsconfig.build.json -m commonjs --outDir lib && webpack && webpack --config ./webpack.config.dev.js",
+		"build:esm": "rimraf lib-esm && tsc -p ./tsconfig.build.json -m esnext --outDir lib-esm",
+		"build:cjs:watch": "rimraf lib && tsc -m commonjs --outDir lib --watch",
+		"build:esm:watch": "rimraf lib-esm && tsc -m esnext --outDir lib-esm --watch",
```
manueliglesias, line 32: For the sake of consistency, can we keep using the build script? (and made changes to it if needed?)

packages/api/src/API.ts
```diff
@@ -83,6 +84,11 @@ export class APIClass extends InternalAPIClass {
 						// flatten response
 						if (res.data !== undefined) {
 							const [key] = Object.keys(res.data);
+
+							if (res.data[key].items) {
+								return res.data[key].items;
+							}
+
 							return res.data[key];
 						}
```
manueliglesias, line 93: It would be better to use the information coming from the model instrospection

manueliglesias (approved):
Non blocking comments only

LGTM!

packages/api/package.json
iartemiev: Nope, we can't use the build script if we want to use a newer version of TS. Have to use this approach but it's aligned with what we do in Core for the same purpose

packages/amplify-v6-types-package/index.ts
iartemiev: Yes, I'm going to circle back and improve this type in the next PR

packages/api/src/API.ts
iartemiev: For this one, we can probably just narrow it based on the operation, e.g. `list` => flatten items. I'm not sure model intro will give us the necessary metadata

PR #11727, opened 2023-08-04
feat(storage): implement getProperties functional and util functions 

kvramyasri7:
#### Description of changes

- Add `getProperties` functional implementation
- Add `resolveStorageConfig`, `getKeyWithPrefix`, `resolveCredentials` common util functions

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,59 @@
...
+		await AmplifyV6.Auth.fetchAuthSession();
+	assertValidationError(!!awsCreds, StorageValidationErrorCode.NoCredentials);
+	const { bucket, defaultAccessLevel } = AmplifyV6.getConfig().Storage;
+	// TODO: assert bucket and region;
+	const { key, options: { level = defaultAccessLevel } = {} } = req;
+	const { prefixResolver = defaultPrefixResolver } =
+		AmplifyV6.libraryOptions?.Storage ?? {};
+	try {
+		assertValidationError(!!key, StorageValidationErrorCode.NoKey);
+		const final_key =
```
AllanZhengYP, line 35: nit:
```suggestion
		const finalKey =
```

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,59 @@
...
+			prefixResolver({
+				level,
+				identityId: awsCredsIdentityId,
+			}) + key;
+		const response = await headObject(options, {
+			Bucket: bucket,
+			Key: final_key,
```
AllanZhengYP, line 42:
```suggestion
			Key: finalKey,
```

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,59 @@
...
+		});
+		const getPropertiesResponse: S3GetPropertiesResult = {
+			key: final_key,
```
AllanZhengYP, line 45:
```suggestion
			key: finalKey,
```

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,59 @@
...
+			eTag: response.ETag,
+			lastModified: response.LastModified,
+			metadata: response.Metadata,
+		};
+		return getPropertiesResponse;
+	} catch {
+		throw new StorageError({
+			name: 'KeyNotFound',
+			message: 'Error retrieving the key',
+		});
```
AllanZhengYP, line 57: This is a service error, you should not use the validation error pattern and return pre-defined name and message. Here you should check if this is error from service endpoint with friendly name and message already, if so throw directly; If not(networking error for instance), throw StorageError with underlying error.

kvramyasri7: can we just throw an error and will that work @AllanZhengYP ?

packages/storage/src/errors/types/validation.ts
```diff
@@ -16,4 +17,7 @@ export const validationErrorMap: AmplifyErrorMap<StorageValidationErrorCode> = {
 		message:
 			'Missing identity ID when accessing objects in protected or private access level',
 	},
+	[StorageValidationErrorCode.NoKey]: {
+		message: 'Missing key',
```
israx, line 21: Is this message enough to let customers know how to solve the validation error ? probably adding more context here.

packages/storage/__tests__/providers/s3/getProperties.test.ts
```diff
@@ -0,0 +1,46 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { getProperties } from '../../../src/providers/s3';
+
+jest.mock('../../../src/AwsClients/S3');
+const headObject = jest.fn();
+
+test.skip('getProperties API happy path case', async () => {
```
israx, line 9: happy and error test cases should be inside a describe block each.

packages/storage/__tests__/providers/s3/getProperties.test.ts
```diff
@@ -0,0 +1,46 @@
...
+		key: 'key',
+		contentLength: '100',
+		contentType: 'text/plain',
+		eTag: 'etag',
+		lastModified: 'last-modified',
+		metadata,
+	});
+});
+
+test.skip('getProperties API should return a not found error', async () => {
```
israx, line 32: should be part of the error cases describe block

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,56 @@
...
+import { StorageError } from '../../../errors/StorageError';
+import { NETWORK_ERROR_CODE } from '../../../AwsClients/S3/runtime/constants';
+
+/**
+ * Get Properties of the object
+ *
+ * @param {StorageOperationRequest} req
+ * @return {Promise<S3GetPropertiesResult>}
+ */
+export const getProperties = async function (
```
israx, line 21: what errors do this API throw? We should document all possible errors including service and validation errors

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,56 @@
...
+		});
+		const getPropertiesResponse: S3GetPropertiesResult = {
+			key: finalKey,
+			contentType: response.ContentType,
+			contentLength: response.ContentLength,
+			eTag: response.ETag,
+			lastModified: response.LastModified,
+			metadata: response.Metadata,
+		};
+		return getPropertiesResponse;
```
israx, line 52:
```suggestion
    return {
			   key: finalKey,
		   	 contentType: response.ContentType,
			contentLength: response.ContentLength,
			eTag: response.ETag,
			lastModified: response.LastModified,
			metadata: response.Metadata,
		}
```

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,56 @@
...
+	const { awsCreds, awsCredsIdentityId } =
+		await AmplifyV6.Auth.fetchAuthSession();
+	assertValidationError(!!awsCreds, StorageValidationErrorCode.NoCredentials);
+	const { bucket, defaultAccessLevel } = AmplifyV6.getConfig().Storage;
+	// TODO: assert bucket and region;
+	const { key, options: { level = defaultAccessLevel } = {} } = req;
+	const { prefixResolver = defaultPrefixResolver } =
+		AmplifyV6.libraryOptions?.Storage ?? {};
+	assertValidationError(!!key, StorageValidationErrorCode.NoKey);
+	try {
```
israx, line 34: we don't really need a try and catch clause. If it fails, the error will be caught by the customers code

israx: we should use `try and catch` if we depend on the result for a subsequent call to a service.

israx (commented):
We should consider testing more error cases.

1. Testing validation errors
2. Testing service errors

For each test we should check if the error was correctly mapped for an `StorageError`

packages/storage/src/providers/s3/apis/getProperties.ts
israx: how are we mapping service errors ? We need to map the service errors to be consistent on the way errors are thrown

kvramyasri7: Thanks for the feedback. Looking at auth Error handling for reference.

packages/storage/src/providers/s3/apis/getProperties.ts
kvramyasri7: yeah but for the hubEvents do you think we need this `try` &  `catch` blocks🤔?

packages/storage/src/providers/s3/apis/getProperties.ts
kvramyasri7: Updated thanks! took a reference from Auth✌️

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,53 @@
...
+import { StorageValidationErrorCode } from '../../../errors/types/validation';
+import { prefixResolver as defaultPrefixResolver } from '../../../utils/prefixResolver';
+import { S3GetPropertiesResult } from '../types/results';
+
+/**
+ * Get Properties of the object
+ *
+ * @param {StorageOperationRequest} The request object
+ * @return {Promise<S3GetPropertiesResult>} Properties of the object
+ * @throws service: {@link NotFoundException} - thrown when there is no given object in bucket
```
israx, line 18: the `NotFoundException` is too specific. The service will throw different errors not only one.

packages/storage/src/errors/utils/assertServiceError.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { StorageError } from '../StorageError';
+import { AmplifyErrorString, ServiceError } from '@aws-amplify/core';
+
+export function assertServiceError(
```
israx, line 7: how is this being used?

packages/storage/src/providers/s3/apis/getProperties.ts
kvramyasri7: Got it thanks updated.

packages/storage/src/types/params.ts
```diff
@@ -8,16 +8,16 @@ export type StorageOptions =
...
 	key: string;
 	options?: Options;
 };
 
-export type StorageDownloadDataParameter<Options extends StorageOptions> =
-	StorageOperationParameter<Options>;
+export type StorageDownloadDataRequest<Options extends StorageOptions> =
+	StorageOperationRequest<Options>;
 
 export type StorageDownloadFileParameter<Options extends StorageOptions> =
```
ashwinkumar6, line 19: should we rename these `StorageDownloadFileParameter` as well ?
or i'm wondering if only makes sense to update those related to getProperties 

cause we have also updated `StorageDownloadDataParameter`

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,61 @@
...
+	const { bucket, defaultAccessLevel } = AmplifyV6.getConfig().Storage;
+	assertValidationError(!!bucket, StorageValidationErrorCode.NoBucket);
+	const { prefixResolver = defaultPrefixResolver } =
+		AmplifyV6.libraryOptions?.Storage ?? {};
+	const { key, options: { level = defaultAccessLevel } = {} } = req;
+	assertValidationError(!!key, StorageValidationErrorCode.NoKey);
+	const finalKey =
+		prefixResolver({
+			level,
+			identityId: awsCredsIdentityId,
```
ashwinkumar6, line 41: Should we move lines 28-40 to a separate utils function. These might be common among different APIs 
(thanks for feedback Allan)

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,61 @@
...
+		}) + key;
+	const params: HeadObjectInput = {
+		Bucket: bucket,
+		Key: finalKey,
+	};
+	try {
+		const response = await headObject(options, params);
```
ashwinkumar6, line 48: Can we manually create an `options` object and then pass it here so it's more evident as to what we're passing

ashwinkumar6: and that can be typed as well

kvramyasri7: do we need to infer types manually? @ashwinkumar6

packages/storage/src/types/params.ts
kvramyasri7: to avoid confusion I am just updating `getProperties` params. But, we can do that as well.

packages/storage/src/providers/s3/apis/getProperties.ts
ashwinkumar6: we could do it manually like what we have done for response

packages/storage/src/providers/s3/apis/getProperties.ts
kvramyasri7: will add a TODO here and will replace.

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,61 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-// TODO: pending implementation
-export declare const getProperties: (params: any) => Promise<any>;
+import { HeadObjectInput, headObject } from '../../../AwsClients/S3';
+import { StorageOperationRequest } from '../../../types';
+import { AmplifyV6 } from '@aws-amplify/core';
+import { StorageOptions } from '../../../types/params';
```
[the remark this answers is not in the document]
kvramyasri7: That's a good catch. Will make it consistent. Thanks.

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,48 @@
...
+ * - S3 Service errors thrown while getting properties
+ * @throws validation: {@link StorageValidationErrorCode } - Validation errors thrown
+ *
+ */
+export const getProperties = async function (
+	req: StorageOperationRequest<StorageOptions>
+): Promise<S3GetPropertiesResult> {
+	const { awsCredsIdentityId, awsCreds, defaultAccessLevel, bucket, region } =
+		await getStorageConfig();
+	const { key, options: { accessLevel = defaultAccessLevel } = {} } = req;
```
ashwinkumar6, line 26:
```suggestion
  // TODO should we use targetIdentityId from req as well 
	const { key, options: { accessLevel = defaultAccessLevel } = {} } = req;
```

ashwinkumar6 (approved):

packages/storage/src/AwsClients/S3/headObject.ts
```diff
@@ -49,7 +38,8 @@ const headObjectSerializer = async (
 	input: HeadObjectInput,
 	endpoint: Endpoint
 ): Promise<HttpRequest> => {
-	const headers = await serializeObjectSsecOptionsToHeaders(input);
+	// TODO need to check with BR for headers
+	const headers = {};
```
AllanZhengYP, line 42: This can be in-line on line 47

AllanZhengYP (commented):
It's getting closer, thank you @kvramyasri7

packages/storage/src/types/results.ts
```diff
@@ -19,3 +19,7 @@ export type StorageGetUrlResult = {
 export type StorageUploadResult = {
 	key: string;
 };
+
+export type StorageGetPropertiesResult = {
+	contentType?: string;
+};
```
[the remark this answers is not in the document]
kvramyasri7: So the result type of `S3GetPropertiesResult` is `StorageGetPropertiesResult & S3Item`,  basically `S3Item` does not have contentType.

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,49 @@
...
+ */
+export const getProperties = async function (
+	req: StorageOperationRequest<StorageOptions>
+): Promise<S3GetPropertiesResult> {
+	const { awsCredsIdentityId, awsCreds, defaultAccessLevel, bucket, region } =
+		await getStorageConfig();
+	// TODO should we use targetIdentityId from req as well
+	const { key, options: { accessLevel = defaultAccessLevel } = {} } = req;
+	assertValidationError(!!key, StorageValidationErrorCode.NoKey);
+	const finalKey = getFinalKey(accessLevel, awsCredsIdentityId, key);
```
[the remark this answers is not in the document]
kvramyasri7: Updated. Thanks!

packages/storage/src/providers/s3/types/results.ts
```diff
@@ -6,8 +6,13 @@ import {
 	StorageGetUrlResult,
 	StorageUploadResult,
 } from '../../../types';
+import { StorageGetPropertiesResult } from '../../../types/results';
 
 type S3ObjectInformation = {
```
AllanZhengYP, line 11: Rename to `S3Item` and extend it from `StorageItem` which excludes `versionId` and `contentType`

AllanZhengYP (commented):
Almost there. I only have a couple more comments

packages/storage/src/AwsClients/S3/headObject.ts
```diff
@@ -74,10 +72,8 @@ const headObjectDeserializer = async (
 				LastModified: ['last-modified', deserializeTimestamp],
 				VersionId: 'x-amz-version-id',
 			}),
-			Metadata: deserializeMetadata(response.headers),
```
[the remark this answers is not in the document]
kvramyasri7: Gotcha updating. 🚶‍♀️

packages/storage/src/providers/s3/types/results.ts
```diff
@@ -26,15 +34,21 @@ type S3ObjectInformation = {
 	 * @see https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata
 	 */
 	metadata?: Record<string, string>;
+	//TODO add versionId
```
HuiSF, line 37: Is this TODO message meant for somewhere else? 😅

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,50 @@
...
+
+	const { key, options: { accessLevel = defaultAccessLevel } = {} } = req;
+	assertValidationError(!!key, StorageValidationErrorCode.NoKey);
+	const finalKey = await getKeyWithPrefix(accessLevel!, identityId!, key);
+	const headObjectOptions = {
+		accessLevel,
+		targetIdentityId: identityId,
+		region,
+		credentials: credentials,
+	};
```
HuiSF, line 35: If you don't want to use object literal inline calling `headObject`, please type it, `accessLevel` and `targetIdentityId` are not properties of this parameter.

But I recommend just write the object literal inline, so you don't need to import the type.

packages/storage/src/providers/s3/utils/resolveStorageConfig.ts
```diff
@@ -0,0 +1,29 @@
...
+		StorageValidationErrorCode.NoCredentials
+	);
+	assertValidationError(
+		!!credentials,
+		StorageValidationErrorCode.NoCredentials
+	);
+	const { bucket, region } = AmplifyV6.getConfig()?.Storage ?? {};
+	assertValidationError(!!bucket, StorageValidationErrorCode.NoBucket);
+	assertValidationError(!!region, StorageValidationErrorCode.NoRegion);
+	const { defaultAccessLevel } = AmplifyV6.libraryOptions?.Storage ?? {};
```
HuiSF, line 21: Can destruct `defaultAccessLevel` at line 18.

HuiSF (changes requested):

packages/storage/src/providers/s3/utils/resolveStorageConfig.ts
kvramyasri7: `defaultAccessLevel` is not available on `getConfig` method. It's available on `libraryOptions `.

packages/storage/src/providers/s3/utils/resolveStorageConfig.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { AmplifyV6 } from '@aws-amplify/core';
+import { assertValidationError } from '../../../errors/utils/assertValidationError';
+import { StorageValidationErrorCode } from '../../../errors/types/validation';
+
+export async function resolveStorageConfig() {
+	const { identityId, credentials } = await AmplifyV6.Auth.fetchAuthSession();
```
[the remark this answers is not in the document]
Samaritan1011001: That's right. It will be exported from `aws-amplify/auth`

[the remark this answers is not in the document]
kvramyasri7: Right now in next branch it is causing build_failure. can I add a todo for that and change later on once fetchAuthSession is merged. 🤔

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,50 @@
...
+ * @return {Promise<S3GetPropertiesResult>} Properties of the object
+ * @throws service: {@link GetPropertiesException}
+ * - S3 Service errors thrown while getting properties
+ * @throws validation: {@link StorageValidationErrorCode } - Validation errors thrown
+ *
+ */
+export const getProperties = async function (
+	req: StorageOperationRequest<StorageOptions>
+): Promise<S3GetPropertiesResult> {
+	const { identityId, credentials, defaultAccessLevel, bucket, region } =
```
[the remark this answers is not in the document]
kvramyasri7: Thanks updated!

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,50 @@
...
+		await resolveStorageConfig();
+
+	const { key, options: { accessLevel = defaultAccessLevel } = {} } = req;
+	assertValidationError(!!key, StorageValidationErrorCode.NoKey);
+	const finalKey = await getKeyWithPrefix(accessLevel!, identityId!, key);
```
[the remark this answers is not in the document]
kvramyasri7: Thanks for the suggestion. Updated.

packages/storage/src/AwsClients/S3/headObject.ts
```diff
@@ -64,7 +54,7 @@ const headObjectDeserializer = async (
 ): Promise<HeadObjectOutput> => {
 	if (response.statusCode >= 300) {
 		const error = await parseXmlError(response);
-		throw error;
+		assertServiceError(error);
```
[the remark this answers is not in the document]
AllanZhengYP: Sorry the suggestion does not mean to be exact. Can you build static function like @HuiSF suggestion to build the service error?

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,61 @@
...
+		accessLevel,
+		targetIdentityId: identityId,
+		region,
+		credentials,
+	};
+	const headObjectParams: HeadObjectInput = {
+		Bucket: bucket,
+		Key: finalKey,
+	};
+	const response = await headObject(headObjectOptions, headObjectParams);
```
AllanZhengYP, line 51: I just realize we need to throw an error with recoverSuggestion here. Suggesting customer needs to create either the bucket or the object because they do not exist

packages/storage/src/AwsClients/S3/headObject.ts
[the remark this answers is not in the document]
kvramyasri7: if (response.statusCode >= 300)  {
```suggestion
		const error = await parseXmlError(response);
		StorageError.fromXmlError(error);
		assertServiceError(error);
```
	}

Will this work?

packages/storage/src/providers/s3/apis/getProperties.ts
kvramyasri7: we are doing it in while getting it from `resolveStorageConfig`. Do we need to it here again?

packages/storage/src/providers/s3/utils/resolveCredentials.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { AmplifyV6 } from '@aws-amplify/core';
+import { assertValidationError } from '../../../errors/utils/assertValidationError';
+import { StorageValidationErrorCode } from '../../../errors/types/validation';
+
+export async function resolveCredentials() {
+	// TODO import fetchAuthSession from
```
kvramyasri7, line 9:
```suggestion
	// TODO import fetchAuthSession directly from `aws-amplify`
```

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,61 @@
...
+	const { defaultAccessLevel, bucket, region } = resolveStorageConfig();
+	const { identityId, credentials } = await resolveCredentials();
+	const {
+		key,
+		options: { accessLevel },
+	} = req;
+	let targetIdentityId = identityId;
+	if (req?.options?.accessLevel === 'protected') {
+		targetIdentityId = req.options?.targetIdentityId;
+	}
```
AllanZhengYP, line 34: Have we asserted the `req.options?.targetIdentityId` is always set when accessLevel is `protected`? If not let's add it. Also I prefer conditional assignment here: `const targetIdentityId = [condition] ? [val1] : [val2]`

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,61 @@
...
+		accessLevel: accessLevel,
+		targetIdentityId: targetIdentityId,
+		region,
+		credentials,
+	};
+	const headObjectParams: HeadObjectInput = {
+		Bucket: bucket,
+		Key: finalKey,
+	};
+	const response = await headObject(headObjectOptions, headObjectParams);
```
AllanZhengYP, line 51: These can be in-lined too.

kvramyasri7: I have to do that because of the type and using inline object lteral.

packages/storage/src/providers/s3/apis/getProperties.ts
kvramyasri7: Thanks for this suggestion. Updated.

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,64 @@
...
+/**
+ * Get Properties of the object
+ *
+ * @param {StorageOperationRequest} The request object
+ * @return {Promise<S3GetPropertiesResult>} Properties of the object
+ * @throws service: {@link GetPropertiesException}
+ * - S3 Service errors thrown while getting properties
+ * @throws validation: {@link StorageValidationErrorCode } - Validation errors thrown
+ *
+ */
```
HuiSF, line 24: Great ideas to add throws here, thanks.
One suggestion on the formate:

Probably need to verify the format of the added comment docs. E.g. the doc of `getProperties` looks like this currently:

image

```suggestion
/**
 * Gets the properties of a file. The properties include S3 system metadata and
 * the user metadata that was provided when uploading the file.
 *
 * @param {StorageOperationRequest} req The request to make an API call.
 * @returns {Promise<S3GetPropertiesResult>} A promise that resolves the properties.
 * @throws A {@link GetPropertiesException} when the underlying S3 service returned error.
 * @throws A {@link StorageValidationErrorCode} when API call parameters are invalid.
 */
```

image

Please check the accuracy other the inline-docs else where, too.

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,64 @@
...
+): Promise<S3GetPropertiesResult> {
+	const { defaultAccessLevel, bucket, region } = resolveStorageConfig();
+	const { identityId, credentials } = await resolveCredentials();
+	const {
+		key,
+		options: { accessLevel },
+	} = req;
+	let targetIdentityId;
+	if (req?.options?.accessLevel === 'protected') {
+		targetIdentityId = identityId ?? req.options?.targetIdentityId;
```
HuiSF, line 36: I think when `accessLevel === 'protected'` the `targetIdentityId` from the `req` has a higher priority.

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -1,5 +1,64 @@
...
+	}
+	assertValidationError(!!key, StorageValidationErrorCode.NoKey);
+	const finalKey = getKeyWithPrefix(
+		accessLevel ?? defaultAccessLevel,
+		targetIdentityId ?? identityId,
+		key
+	);
+	const headObjectOptions = {
+		accessLevel: accessLevel,
+		targetIdentityId: targetIdentityId,
```
HuiSF, line 46: Still these two properties are not defined in the interface of the first param of `headObject` function.

packages/storage/src/providers/s3/utils/resolveCredentials.ts
```diff
@@ -0,0 +1,19 @@
...
+	const { identityId, credentials } = await AmplifyV6.Auth.fetchAuthSession();
+	assertValidationError(
+		!!credentials,
+		StorageValidationErrorCode.NoCredentials
+	);
+	return {
+		identityId,
+		credentials,
+	};
+}
```
HuiSF, line 19: Um I have a question about the approach of using `assertValidationError` here - I understood this is useful when assert non-undefined values in pure JS. 

But considering the TS, both `identityId` and `credentials` can be `undefined` looking at the type `AuthSession`. using `assertValidationError` doesn't narrow `string | undefined` to `string`.

The return type of `resolveCredentials` is going to be 
```typescript
{
  identityId?: string,
  credentials?: AWSCredentials,
}
```

I think I heard we are going to enable `strict` and `alwaysStrict` in TS? If so this will require quite a lot of manual changes. Take the above return type, you need to check `undefined` again in the consumer of `resolveCredentials`.

Gut feeling that we should use `if (!identityId) throw` inline check to make it easier for refactoring later. cc @AllanZhengYP

packages/storage/src/providers/s3/apis/getProperties.ts
kvramyasri7: sorry. changed it in recent commit.

packages/storage/src/providers/s3/apis/getProperties.ts
kvramyasri7: Thanks for the suggestion. Updated.

packages/storage/src/providers/s3/utils/resolveCredentials.ts
AllanZhengYP: >But considering the TS, both identityId and credentials can be undefined looking at the type AuthSession. using assertValidationError doesn't narrow string | undefined to string.

I don't think this is correct, the assertion function does narrow the types, [see playground](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXygGdCQYMAKANygmRAC540BrVHAd1QEpGiSzC8arRABuAFDjQkWAjB5CGAo1TIAtgCNSAHzShEWVCGASo8APTn4qzaXjamqfYePj5qRfBwaAVvAC88ADeUAC+psSkFFBcphZWNlowEu6eUL4ATAHBYeJAA)

HuiSF: Ahh nice, I didn't look the `assertValidationError` closely. 👍
Please ignore the comment regarding the strict mode. But I think we need to assert `identityId`, too.

kvramyasri7: Yeah. Thanks updated.

packages/storage/src/errors/StorageError.ts
```diff
@@ -1,6 +1,17 @@
...
 
 export class StorageError extends AmplifyError {
+	static formServiceError(error: Error, statusCode: number): ServiceError {
+		const storageError = new StorageError({
+			name: error.name,
+			message: error.message,
+		});
+		if (statusCode === 404) {
+			storageError.recoverySuggestion =
+				'Please add the object with this key to the bucket as the key is not found.';
```
HuiSF, line 11: nit: This `recoverSuggestion` may not cover the case that API caller provided a wrong combination of accessLevel, key and targetIdentityId.

HuiSF (approved):

israx (approved):
LGTM. Seems that some unit tests are failing.

packages/storage/src/AwsClients/S3/headObject.ts
```diff
@@ -49,12 +48,11 @@ const headObjectSerializer = async (
 	input: HeadObjectInput,
 	endpoint: Endpoint
 ): Promise<HttpRequest> => {
-	const headers = await serializeObjectSsecOptionsToHeaders(input);
 	const url = new URL(endpoint.url.toString());
-	url.pathname = serializePathnameObjectKey(url, input.Key);
+	url.pathname = serializePathnameObjectKey(url, input.Key!);
```
AllanZhengYP, line 52: Why is this change needed?

packages/storage/src/AwsClients/S3/headObject.ts
```diff
@@ -64,7 +62,7 @@ const headObjectDeserializer = async (
 ): Promise<HeadObjectOutput> => {
 	if (response.statusCode >= 300) {
 		const error = await parseXmlError(response);
-		throw error;
+		throw StorageError.formServiceError(error, response.statusCode);
```
AllanZhengYP, line 65: Nit: from previous review:
```suggestion
		throw StorageError.fromServiceError(error, response.statusCode);
```

packages/storage/src/AwsClients/S3/headObject.ts
```diff
@@ -77,7 +75,6 @@ const headObjectDeserializer = async (
 			Metadata: deserializeMetadata(response.headers),
 		};
 		return {
-			$metadata: parseMetadata(response),
```
AllanZhengYP, line 80: Let's keep this still, we might need it for logger later.

packages/storage/__tests__/AwsClients/S3/cases/headObject.ts
```diff
@@ -48,7 +48,7 @@ const headObjectHappyCase: ApiFunctionalTestCase<typeof headObject> = [
 		body: '',
 	},
 	{
-		$metadata: expect.objectContaining(expectedMetadata),
+		//$metadata: expect.objectContaining(expectedMetadata),
```
AllanZhengYP, line 51: Let's keep this for now, see the comments below:

packages/storage/src/providers/s3/types/results.ts
```diff
@@ -7,15 +7,23 @@ import {
 	StorageUploadResult,
 } from '../../../types';
 
-type S3ObjectInformation = {
+type S3Item = {
```
AllanZhengYP, line 10: Not a blocking issue. In the future this needs to be extended from `StorageItem`.

packages/storage/src/providers/s3/types/results.ts
```diff
@@ -26,15 +34,20 @@ type S3ObjectInformation = {
...
 	metadata?: Record<string, string>;
+	/**
+	 * VersionId used to reference a specific version of the object.
+	 */
+	versionId?: string;
 };
 
-export type S3DownloadDataResult = StorageDownloadDataResult &
-	S3ObjectInformation;
+export type S3DownloadDataResult = StorageDownloadDataResult & S3Item;
```
AllanZhengYP, line 43: I'd prefer to use generics like below:

```suggestion
export type S3DownloadDataResult = StorageDownloadDataResult<S3Item>;
```

packages/storage/src/AwsClients/S3/headObject.ts
kvramyasri7: Reverted back. Thanks.

packages/storage/src/providers/s3/types/results.ts
```diff
@@ -4,37 +4,29 @@
...
-export type S3DownloadFileResult = S3ObjectInformation;
+export type S3DownloadFileResult = S3Item;
 
 export type S3GetUrlResult = StorageGetUrlResult;
 
 export type S3UploadDataResult = StorageUploadResult;
 
 export type S3UploadFileResult = StorageUploadResult;
+
+export type S3GetPropertiesResult = StorageItem<S3Item>;
```
AllanZhengYP, line 32: nit: I think this should just S3Item:
```suggestion
export type S3GetPropertiesResult = S3Item;
```

packages/storage/src/types/results.ts
```diff
@@ -3,10 +3,35 @@
 
 import { Headers } from '@aws-amplify/core/internals/aws-client-utils';
 
+export type StorageItem<T = {}> = T & {
```
AllanZhengYP, line 6:
```suggestion
export type StorageItem = {
```

packages/storage/src/types/results.ts
```diff
@@ -3,10 +3,35 @@
...
+	metadata?: Record<string, string>;
+};
+
 // TODO: replace with ResponsePayloadMixin from core
 type Payload = Pick<Body, 'blob' | 'json' | 'text'>;
 
-export type StorageDownloadDataResult = {
+export type StorageDownloadDataResult<T = {}> = T & {
 	body: Payload;
 };
```
AllanZhengYP, line 36:
```suggestion
export type StorageDownloadDataResult<T extends StorageItem> = T & {
	body: Payload;
};
```

packages/storage/src/providers/s3/types/results.ts
```diff
@@ -4,37 +4,29 @@
 import {
 	StorageDownloadDataResult,
 	StorageGetUrlResult,
+	StorageItem,
 	StorageUploadResult,
 } from '../../../types';
 
-type S3ObjectInformation = {
+type S3Item = {
```
AllanZhengYP, line 11: This should be some thing like:

```
export type S3Item = StorageItem & {
  versionId?: string;
  contentType?: string;
}
```
or 
```
export interface S3Item extends StorageItem {
  versionId?: string;
  contentType?: string;
}
```

AllanZhengYP (approved):
Looks good to me! 🚀🚀🚀
Thank you so much for putting this together! @kvramyasri7

israx (approved):

packages/storage/src/providers/s3/utils/resolveStorageConfig.ts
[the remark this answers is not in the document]
AllanZhengYP: We should use the `fetchAuthSession` from the core package instead. In one from umbrella package is suppose to be public-facing. Using it internally will alter our dependency structure of the Storage scoped package.

@HuiSF @Samaritan1011001

PR #11757, opened 2023-08-08
chore: Update `amazon-cognito-identity-js` to use Webpack 5

jimblanc:
#### Description of changes

This change upgrades `amazon-cognito-identity-js` to use Webpack 5 to resolve a lingering `glob-parent` alert. With this change the entire library will be aligned on Webpack 5 for UMD artifact generation. Also updated the clean script to remove `dist` and removed some unused `babel` imports in `pushnotification`.

**On the `buffer` dependencies**
Without these dependencies we observed a bundle size regression as Buffer 5.7.1 was being preferred for categories that directly or transitively use `buffer`. It appears that when a sub-category has a dependency on Webpack v4, the package manager was correctly hoisting buffer v4.9.2. However with the removal of Webpack v4, the resolution logic was preferring v5.7.1 (from `react-native-url-polyfill` in `core`) instead resulting in a duplicated dependency in the bundle. The fix here was to add an explicit dependency on `buffer` for all packages that are directly using it to coerce the previous behavior (i.e. buffer v4.9.2).

Artifacts produced:
```
-rw-r--r--   1 jablanch  staff  352929 Aug  8 14:12 amazon-cognito-identity.js
-rw-r--r--   1 jablanch  staff  123193 Aug  8 14:12 amazon-cognito-identity.min.js
-rw-r--r--   1 jablanch  staff   29987 Aug  8 14:12 amazon-cognito-identity.min.js.gz
```

Verified `glob-parent` is at an appropriate version.

Smoke tested in a vanilla HTML app using the signUp & authenticateUser APIs.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amazon-cognito-identity-js/package.json
+++ b/packages/amazon-cognito-identity-js/package.json
@@ -36,7 +36,7 @@
     "reactnative"
   ],
   "scripts": {
-    "clean": "rimraf lib es",
+    "clean": "rimraf lib es dist",
     "build:cjs": "cross-env BABEL_ENV=commonjs babel src --out-dir lib",
     "build:cjs:watch": "cross-env BABEL_ENV=commonjs babel src --out-dir lib --watch",
     "build:esm": "cross-env BABEL_ENV=es babel src --out-dir es",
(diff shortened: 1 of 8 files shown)


packages/amazon-cognito-identity-js/package.json
```diff
@@ -36,7 +36,7 @@
     "reactnative"
   ],
   "scripts": {
-    "clean": "rimraf lib es",
+    "clean": "rimraf lib es dist",
```
cshfang, line 39: 👍  good catch

cshfang (approved):
:shipit:

ashwinkumar6 (dismissed):

HuiSF (dismissed):

cshfang (dismissed):

packages/amazon-cognito-identity-js/webpack.config.js
```diff
@@ -23,35 +26,43 @@ var config = {
...
+		new webpack.BannerPlugin({ banner, raw: true }),
+		new CompressionPlugin({
+			include: /\.min\.js$/,
+		})
 	],
 	optimization: {
 		minimizer: [
 			new TerserPlugin({
-				sourceMap: true,
+				extractComments: false,
```
elorzafe, line 42: Why not true?

jimblanc: This was done to preserve the current artifact structure where license comments are embedded in the JS directly (vs being extracted into a `.LICENSE.js` file). Other comments are still extracted.

cshfang (approved):

elorzafe (approved):

svidgen (approved):

PR #11766, opened 2023-08-09
chore: Setup analytics export paths

jimblanc:
#### Description of changes

This change sets up the initial Analytics export paths to prepare for functional APIs.

```
aws-amplify/analytics
aws-amplify/analytics/pinpoint
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/analytics/src/Analytics.ts
```diff
@@ -7,7 +7,7 @@ import {
 	Hub,
 	parseAWSExports,
 } from '@aws-amplify/core';
-import { AWSPinpointProvider } from './Providers/AWSPinpointProvider';
+import { AWSPinpointProvider } from './providers/AWSPinpointProvider';
```
elorzafe, line 10: What happen with the `Providers` directory?

elorzafe (approved):
Looks good, only thing is the `Providers` vs `providers` directory. It seems that could be problematic depending on the platform you are compiling?

packages/analytics/src/Analytics.ts
jimblanc: I just renamed it to be lower case to align with the other categories & folders in this package. No impact on the public surface.

jimblanc:
> Looks good, only thing is the `Providers` vs `providers` directory. It seems that could be problematic depending on the platform you are compiling?

Ended up reverting this, kept running into build issues even when all imports were updated to use the lower-case directory.

elorzafe (approved):
Looks good!

elorzafe (approved):

Samaritan1011001 (approved):
lgtm

PR #11781, opened 2023-08-10
chore: Remove GA providers from Analytics

jimblanc:
#### Description of changes

Removed non-Dev Preview providers from the Analytics export surface and updates AWS SDK dependencies. Updating these AWS SDK dependencies will require that the `react-native-get-random-values` polyfill be installed for RN customers, which is already planned for v6 (and required by the UUID upgrade as well).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/analytics/package.json
```diff
@@ -48,10 +48,10 @@
...
 	],
 	"dependencies": {
-		"@aws-sdk/client-firehose": "3.6.1",
-		"@aws-sdk/client-kinesis": "3.6.1",
-		"@aws-sdk/client-personalize-events": "3.6.1",
-		"@aws-sdk/util-utf8-browser": "3.6.1",
+		"@aws-sdk/client-firehose": "3.388.0",
+		"@aws-sdk/client-kinesis": "3.388.0",
+		"@aws-sdk/client-personalize-events": "3.388.0",
+		"@aws-sdk/util-utf8-browser": "3.259.0",
```
jimblanc, line 54: We need to retain these deps for now to avoid having to remove the unused providers for dev preview, unused providers are excluded from the public package interface.

cshfang: Not that I'm complaining but why upgrade them at all? On the complete and opposite hand, would it be possible to use caret syntax for these?

cshfang (approved):
Have questions but non blocking imo

packages/analytics/src/index.ts
```diff
@@ -1,11 +1,18 @@
...
+// TODO(v6) Remove once all default provider functional APIs available
 export { Analytics } from './Analytics';
 export { AnalyticsProvider } from './types';
 export {
 	AWSPinpointProvider,
+} from './Providers';
+
+// TODO(v6) Refactor as additional Analytics providers come online
+/*
+export {
```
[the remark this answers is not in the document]
jimblanc: Yeah good question, we won't be exporting non-default providers from root index. I originally kept them here as a reference when we make the other providers functional, but I can remove in a future PR. There's another PR where I set up the default provider exports: https://github.com/aws-amplify/amplify-js/pull/11766

packages/analytics/package.json
jimblanc: Big reason is to avoid installing old/unneeded dependencies in the v6 dev preview (e.g. RN). Thought about using caret syntax, but we'll need to go through and do that everywhere ahead of v6 to make sure we're getting optimal flattening. Also I think there's a discussion to be had around risk of breaking changes since they sometimes release breaking changes in minor versions (e.g. when they removed RN polyfills).

Samaritan1011001 (approved):
After confirming offline on the dx to install the pollyfill and confirming that commenting out other providers helps further dev, this lgtm

PR #11783, opened 2023-08-10
chore: Upgrade prettier and config

cshfang:
### Update Prettier column width (`printWidth`) rule to be consistent with `tslint`
Previously, `tslint` was configured to allow `120` maximum column width but Prettier was not and so, by default, would rewrite using `80` as the width. Based on a team decision, the goal of this PR is ultimately to align this on `120`. In an effort to shore up our linting, several tools were also upgraded.

#### Upgrade [Prettier](https://github.com/prettier/prettier) to latest (3.0.1)
Validated the previous version (2.88.0) results in the same output with as 3.0.1 when trailingComma is configured to use `es5` as before.

#### Replace [pretty-quick](https://github.com/azz/pretty-quick) with [lint-staged](https://github.com/okonet/lint-staged)
The `pretty-quick` package is no longer actively maintained (last activity 2 years ago) and does not work with Prettier 3.0. `lint-staged` implements a more generic approach and is not coupled to prettier, allowing more freedom to add more pre-commit linting to be applied to modified files should the need arise.

#### Upgrade [husky](https://github.com/typicode/husky) to latest (8.0.3)
Husky changed model to be more idiomatic to git hooks. It now vends tooling to create hooks rather than auto-installing. This change should be opaque to development flow. See https://blog.typicode.com/husky-git-hooks-autoinstall/ for their rationale.

Commits in this PR are granular on purpose. These changes were performed in discreet steps to aid reviewer in understanding the approach:
1. Upgrade Prettier to version 3
2. Without updating the configuration to `printWidth: 120`, run prettier against the code base to:
    1. Validate against a smaller set of changes that non-width changes are cosmetic
    2. Align the code base on a standard formatting
3. Run prettier against a even smaller subset of files where the changes appear to be the entire file due to aligning on `LF` instead of `CRLF` carriage returns
4. Update the pre-commit hook with new and upgraded toolset
    1. This was validated by adding formatting errors to several random files and verifying that they are automatically formatted by prettier upon commit
5. Update the configuration to `printWidth: 120` and run prettier against the code base
6. Update ts(es)-lint configuration to expect space before parentheses on anonymous functions. There is not a way to configure this in Prettier (see: https://prettier.io/docs/en/option-philosophy.html)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


HuiSF (approved):

stocaaro (approved):
Wondering if this is the right moment for this change. Seeing it out there, this type of process improvement seems awesome and I want it to happen. With so many PRs out there and divergence between main/next, should this be the first thing we introduce after GA? Could we apply it to next to ensure we don't miss anything in the merge back to main?

PR #11795, opened 2023-08-14
chore: Deprecate i18n & make work with v6

jimblanc:
#### Description of changes

This change deprecates the i18n utility and makes it compatible with the v6 singleton. The utility will check for a default language set in the configuration upon first use, otherwise will use the browser language or whatever the customer has configured via `setLanguage`.

Local testing.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


israx (approved):
LGTM

packages/core/src/I18n/index.ts
```diff
@@ -104,11 +95,12 @@ export class I18n {
 
 	public static checkConfig() {
 		if (!_i18n) {
```
cshfang, line 97: Nit: Is this even possible anymore?

cshfang (approved):

packages/core/src/I18n/index.ts
jimblanc: Yep this member still gets set when a new instance is constructed

PR #11797, opened 2023-08-14
Update Hub to use Observables

elorzafe:
#### Description of changes

Using Observables instead of callbacks on `Hub`

#### Checklist

- [x] PR description included
- [ x `yarn test` passes

### DIFF
--- a/packages/core/src/Hub.ts
+++ b/packages/core/src/Hub.ts
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { Observable, Observer } from 'rxjs';
 import { ConsoleLogger as Logger } from './Logger';
 
 const logger = new Logger('Hub');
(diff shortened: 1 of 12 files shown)


AllanZhengYP (approved):

packages/core/src/Hub.ts
```diff
@@ -140,90 +100,27 @@ export class HubClass {
...
-				pattern: channel,
-				callback: cb,
-			});
-		} else {
-			let holder = this.listeners[channel];
-
-			if (!holder) {
-				holder = [];
-				this.listeners[channel] = holder;
+	listen(channel: string): Observable<HubCapsule> {
```
kvramyasri7, line 106: One question: Are we not supporting `RegExp` for `channel` anymore?🤔?

elorzafe: Good question, I think the regex is more useful when doing `filter`

For example: 

`Hub.listen('auth').pipe(filter(({ payload }) => payload.match(/user/))).subscribe(...)`

packages/core/src/Hub.ts
kvramyasri7, line 1: nit : can we remove `IPattern`, `IListener` intefaces if we are not using elsewhere

stocaaro (commented):
Little bit unsure on where we landed with this one. Are we moving to Observables or keeping callbacks in V6?

stocaaro:

PR #11806, opened 2023-08-16
feature: Integrate Cache utilities with v6.

jimblanc:
#### Description of changes

This change integrates the Cache utilities with the new Singleton and makes them operable with v6. There is no change to the public interface (beyond import path), customers will still be able to:
- Configure the Cache globally
- Create new instances of a Cache with a custom configuration
- Re-configure an existing instance of a Cache

Also made some TypeScript improvements where appropriate.

This most important piece of this PR is the `cacheConfig` function, which contains the logic for resolving the configuration and applying any global or default config values: https://github.com/aws-amplify/amplify-js/pull/11806/files#diff-1d53930c2703b934a6d9cfb22b792f3a82befe22e186250f98f4ac3da7bae5e6R134

Manual testing & unit tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/core/src/Cache/StorageCache.ts
```diff
@@ -13,83 +13,66 @@ const logger = new Logger('StorageCache');
...
+
+		this.sanitizeConfig();
 	}
 
 	public getModuleName() {
 		return 'Cache';
 	}
 
-	private checkConfig(): void {
-		// check configuration
```
jimblanc, line 34: I felt that some of these basic runtime validations were unnecessary with TypeScript, can add them back if folks think they have value.

stocaaro: Yeah. Seems redundant to ts behavior. Between writing tests to cover this behavior and removing, I would vote for removal for sure.

stocaaro (dismissed):
Seems undifferentiated from other options. I wonder a bit that we're maintaining our own LRU variant. Do our customers use this? Does our internal use justify its existence?

stocaaro (approved):

israx (approved):
LGTM

packages/core/src/Cache/Utils/CacheUtils.ts
```diff
@@ -88,3 +88,5 @@ export class CacheObject {
 		delete store[key];
 	}
 }
+
+export const getCurrSizeKey = (keyPrefix: string) => keyPrefix + 'CurSize';
```
cshfang, line 92: Nit: Think this may have been a typo: Cur

May prefer string interp

PR #11816, opened 2023-08-17
Add typing to Hub apis

kvramyasri7:
#### Description of changes

1. Adding typing to Hub apis `dispatch`, `listen`, `_remove`
2. Remove [packages/core/src/Hub/Hub.ts](https://github.com/aws-amplify/amplify-js/pull/11816/files#diff-b1f25be734e93c3f5fc95757a00c9bef2ac8a8fcb13bdf55a839d13b2800a528) and place Hub logic in [packages/core/src/Hub/index.ts ](https://github.com/aws-amplify/amplify-js/pull/11816/files#diff-846386190c9957b3cb7a75794326b92f6714bdfa4caf381dbaebdd55d27cf3f6) 
4. Deprecate `remove` api 
5. Deprecate `regEx` patterns for dev-preview and remove corresponding unit test.
6. Add [AuthHubTypes](https://github.com/aws-amplify/amplify-js/pull/11816/files#diff-943ce5c26d9aa0b52a434573112326686f3dd6760f0f1e6bf14e5a10831e063f) and add correct events
7. Change datastore events to accommodate the api typing.

#### Checklist

- [X] PR description included
- [ ] `yarn test` passes


packages/core/src/Hub/index.ts
```diff
@@ -1,50 +1,35 @@
...
-	message?: string;
-};
-
-export type HubCallback = (capsule: HubCapsule) => void;
-
-export type LegacyCallback = { onHubCapsule: HubCallback };
-
-function isLegacyCallback(callback: any): callback is LegacyCallback {
-	return (<LegacyCallback>callback).onHubCapsule !== undefined;
+function isLegacyCallback<
```
israx, line 20: I think this is going to be deprecated

packages/core/src/Hub/types/HubTypes.ts
```diff
@@ -0,0 +1,94 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { NotificationsHubEventData } from './AnalyticsTypes';
+import { AuthHubEventData } from './AuthTypes';
+
+export const AMPLIFY_SYMBOL = (
```
israx, line 7: This is not a type but a value

israx (commented):
we can potentially remove this method now

packages/core/src/Hub/index.ts
```diff
@@ -1,50 +1,35 @@
...
+	return (
+		(<LegacyCallback<Channel, EventData>>callback).onHubCapsule !== undefined
+	);
 }
 
 export class HubClass {
 	name: string;
 	private listeners: IListener[] = [];
-	private patterns: IPattern[] = [];
+	private patterns: IPattern<any>[] = [];
```
israx, line 32: can we add a type here?

packages/core/src/Hub/types/HubTypes.ts
kvramyasri7: sorry, missed it, will updated this. Can we move this to `index.ts` or do we need to create a new file for this.

packages/core/src/Hub/index.ts
israx: I think this is to handle RegEx cases

packages/core/src/Hub/index.ts
```diff
@@ -68,32 +53,43 @@ export class HubClass {
...
+				];
 			}
-			this.listeners[channel] = [
-				...holder.filter(({ callback }) => callback !== listener),
-			];
 		}
 	}
 
 	/**
 	 * @deprecated Instead of calling Hub.remove, call the result of Hub.listen.
```
israx, line 87: we can potentially remove this

packages/core/src/Hub/index.ts
```diff
@@ -106,13 +102,17 @@ export class HubClass {
...
+		Channel extends ChannelMap['channel'] = ChannelMap['channel']
+	>(
+		channel: Channel,
+		payload: HubPayload<EventData>,
+		source = '',
 		ampSymbol?: Symbol
-	) {
-		if (this.protectedChannels.indexOf(channel) > -1) {
+	): void {
+		if (this.protectedChannels.indexOf(channel as string) > -1) {
```
israx, line 115:
```suggestion
    if (typeof channel === 'string' && this.protectedChannels.indexOf(channel) > -1) {
```

packages/core/src/Hub/index.ts
```diff
@@ -145,13 +145,16 @@ export class HubClass {
...
+		ChannelMap extends AmplifyChannelMap,
+		Channel extends ChannelMap['channel'] = ChannelMap['channel']
+	>(
+		channel: Channel,
+		callback: GetHubCallBack<Channel, ChannelMap['eventData']>,
+		listenerName?: string
+	): () => void {
+		let cb: GetHubCallBack<Channel, ChannelMap['eventData']>;
+		// Check for legacy onHubCapsule callback for backwards compatibility
 		if (isLegacyCallback(callback)) {
```
israx, line 158: We can remove this now

packages/core/src/Hub/index.ts
```diff
@@ -169,11 +172,11 @@ export class HubClass {
 				callback: cb,
 			});
 		} else {
-			let holder = this.listeners[channel];
+			let holder = this.listeners[channel as string];
```
israx, line 175: I'm wonder if we should check here if the channel is a `string` or `RegEx` instance

packages/core/src/Hub/index.ts
```diff
@@ -169,11 +172,11 @@ export class HubClass {
...
 
 			if (!holder) {
 				holder = [];
-				this.listeners[channel] = holder;
+				this.listeners[channel as string] = holder;
```
israx, line 179: ditto

packages/core/src/Hub/index.ts
```diff
@@ -187,9 +190,12 @@ export class HubClass {
...
 	}
 
-	private _toListeners(capsule: HubCapsule) {
+	private _toListeners<
+		Channel extends string | RegExp,
+		EventDataMap extends AmplifyEventDataMap
+	>(capsule: HubCapsule<Channel, EventDataMap>) {
 		const { channel, payload } = capsule;
-		const holder = this.listeners[channel];
+		const holder = this.listeners[channel as string];
```
israx, line 198: ditto

packages/core/src/Hub/index.ts
```diff
@@ -106,13 +102,17 @@ export class HubClass {
...
-		channel: string,
-		payload: HubPayload,
-		source: string = '',
+	dispatch<
+		EventData extends AmplifyEventDataMap,
+		ChannelMap extends AmplifyChannelMap,
+		Channel extends ChannelMap['channel'] = ChannelMap['channel']
+	>(
+		channel: Channel,
+		payload: HubPayload<EventData>,
```
israx, line 111: I think we are missing an important [type util](https://stackblitz.com/edit/rfc-typescript-v6?file=core%2Findex.ts%3AL113) here.

israx (commented):
Thanks @kvramyasri7  for putting this together. I think we are missing a type in the dispatch API

packages/core/src/Hub/index.ts
kvramyasri7: Removed it in recent commit. Thanks for suggestion

packages/core/src/Hub/index.ts
kvramyasri7: string is a type we cannot check for the instance of it and should we change it to `String` ?

packages/core/src/Hub/index.ts
```diff
@@ -0,0 +1,202 @@
...
+export const AMPLIFY_SYMBOL = (
+	typeof Symbol !== 'undefined' && typeof Symbol.for === 'function'
+		? Symbol.for('amplify_default')
+		: '@@amplify_default'
+) as Symbol;
+
+const logger = new Logger('Hub');
+
+export class HubClass<
+	EventData extends AmplifyEventDataMap = AmplifyEventDataMap
```
israx, line 26: We shouldn't add generics on top of the HubClass. How is it used ?

packages/core/src/Hub/index.ts
```diff
@@ -0,0 +1,202 @@
...
+			| HubCallback<Channel, AmplifyEventData[Channel]>
+			| HubCallback<string, EventData>,
+		listenerName?: string
+	): StopListenerCallback {
+		let cb:
+			| HubCallback<Channel, AmplifyEventData[Channel]>
+			| HubCallback<string, EventData>;
+
+		if (typeof callback !== 'function') {
+			throw new Error('No callback supplied to Hub');
```
israx, line 159: I think now we can use `AmplifyError`

packages/core/__tests__/Hub-test.ts
```diff
@@ -107,175 +62,4 @@ describe('Hub', () => {
...
-		);
-
-		expect(listener).not.toHaveBeenCalled();
-		expect(loggerSpy).toHaveBeenCalledWith(
-			'WARN',
-			'Cannot perform pattern matching without a message key'
-		);
-	});
-
-	test('Remove listener', () => {
```
stocaaro, line 246: Will there be a corrolary of remove listener in the changed implementation?

packages/core/src/Hub/types/AuthTypes.ts
```diff
@@ -0,0 +1,35 @@
...
+			data: {
+				url: string | undefined;
+			};
+	  }
+	| { event: 'customOAuthState'; data: string }
+	| { event: 'cognitoHostedUI_failure'; data: Error }
+	| { event: 'customState_failure'; data: Error }
+	| { event: 'forgotPasswordSubmit'; data: any }
+	| { event: 'forgotPasswordSubmit_failure'; data: Error }
+	| { event: 'autoSignIn_failure'; data: null };
```
israx, line 35: I think this events need to be updated

packages/core/src/Hub/index.ts
```diff
@@ -0,0 +1,202 @@
...
+	 */
+	dispatch<Channel extends AmplifyChannel>(
+		channel: Channel,
+		payload: HubPayload<AmplifyEventData[Channel]>,
+		source?: string,
+		ampSymbol?: Symbol
+	): void;
+
+	dispatch(
+		channel: string & {},
```
stocaaro, line 82: `string & {}`? I was expecting this to just be `string`. What do we get from combining with object? Why isn't the object behavior included in the combined function interface down on 91?

packages/core/src/Hub/index.ts
```diff
@@ -0,0 +1,202 @@
...
+
+	dispatch<Channel extends AmplifyChannel>(
+		channel: Channel | string,
+		payload: HubPayload<AmplifyEventData[Channel]> | HubPayload<EventData>,
+		source?: string,
+		ampSymbol?: Symbol
+	): void {
+		if (
+			typeof channel === 'string' &&
+			this.protectedChannels.indexOf(channel) > -1
```
stocaaro, line 99: Aren't AmplifyChannel's also strings? Wondering if the error here will fire for every internal call.

packages/core/src/Hub/index.ts
```diff
@@ -0,0 +1,202 @@
...
+		}
+		let holder = this.listeners[channel as unknown as number];
+
+		if (!holder) {
+			holder = [];
+			this.listeners[channel as unknown as number] =
+				holder as unknown as IListener<string>;
+		}
+
+		(holder as any).push({
```
stocaaro, line 175: Whats up with this `as any`? This seems concerning since `as` is usually a bit of a "lets pretend" type idiom.  Maybe the same can be achieved with guard clauses or type narrowing?

packages/core/src/Hub/types/AuthTypes.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// TODO Need to update types of data
```
stocaaro, line 4: Thankyou for this line. Was going to comment on the `any`'s. Glad to see that clarification is planned for a future change.

packages/core/src/Hub/index.ts
```diff
@@ -0,0 +1,202 @@
...
+		return () => {
+			this._remove(channel, cb as any);
+		};
+	}
+
+	private _toListeners<Channel extends string>(
+		capsule: HubCapsule<Channel, EventData>
+	) {
+		const { channel, payload } = capsule;
+		const holder = this.listeners[channel as unknown as number];
```
stocaaro, line 190: Channel becomes a number to index listeners? This looks funny. `.listeners` is an array, but channel is a string, right? Are we using the right underlying object? Should this be a different structure, maybe a map?  Not sure I'm thinking through the full problem, but it looks funny.

stocaaro (changes requested):
`as` is probably an indication that we're cutting corners. Can we review the `as` usage and identify alternative strategies for these implementation patterns?

packages/core/src/Hub/index.ts
israx: The current implementation is accessing array elements with an string instead of number at run time (didn't know it was supported) but TS only admits numbers as index

packages/core/__tests__/Hub-test.ts
kvramyasri7: Added this test back. Thanks for mentioning.

packages/core/src/Hub/index.ts
kvramyasri7: We have updated this one to of only type string please take a look at it.

packages/core/src/Hub/index.ts
```diff
@@ -0,0 +1,205 @@
...
+	>(
+		channel: Channel,
+		callback: HubCallback<Channel, EventData>,
+		listenerName: string = 'noname'
+	): StopListenerCallback {
+		let cb: HubCallback<string, EventDataMap>;
+		if (typeof callback !== 'function') {
+			throw new AmplifyError({
+				name: NO_HUBCALLBACK_PROVIDED_EXCEPTION,
+				message: 'Service Worker not available',
```
israx, line 161: I think `message` is not accurate.

israx (dismissed):
Looks good. Only improving the error message and also add more description on the PR to highlight all the changes

packages/core/src/Hub/index.ts
kvramyasri7: Thanks updtaed.

packages/core/src/Hub/index.ts
stocaaro: Got it. So this is funky and we're not planning to improve upon it?

israx: I think we can do a better job refactoring this code so each channel is mapped to a number instead. We would need to modify some run time behavior in order to do it. I think if it is not an overwhelming job we can address it on this PR. Otherwise we can add a `TODO` for a next PR

packages/core/src/Hub/index.ts
israx: It won't fire as long as each category has added an `ampSymbol` while using `dispatch`

packages/core/src/Hub/index.ts
kvramyasri7: I am adding `TODO` for this change as it involves change in run time behavior too and need to test in sample app. Will address this in a separate PR and added more description to this PR. What do you think @israx @stocaaro ?

israx: Sounds good to me

stocaaro (approved):
Are we introducing any new behavior here that needs to be unit tested? I don't think we are, but with my approval want to ask the question.

israx (approved):
Awesome. Thanks @kvramyasri7

kvramyasri7:
> 
We are not introducing anything new in this PR. But please free to comment if I am missing anything. Thanks.

packages/core/src/Hub/index.ts
kvramyasri7: Thanks for explanation people. Forgot resolve this argument. Doing it right away.

packages/core/src/Hub/index.ts
kvramyasri7: Thanks for letting ME know. I will take a look at it. But not sure why typescript accepted `as any` is acceptable. May be TS compiler is okay with pretending. Just kidding 😅. For sure we need to add guard clauses/type narrowing so we only allow few `types`.

packages/core/src/Hub/index.ts
kvramyasri7: yes, looks good to me too

Issue #11846, opened 2023-08-21
Authentication/redirect loops with federatedSignIn after upgrade to 5.3.8


mat-mik:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication

### Amplify Categories

auth

### Environment information

<details>

```
# Put output below this line
System:
    OS: macOS 13.4.1
    CPU: (8) arm64 Apple M1
    Memory: 39.38 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.14.0 - ~/.nvm/versions/node/v18.14.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.14.0/bin/yarn
    npm: 9.3.1 - ~/.nvm/versions/node/v18.14.0/bin/npm
    Watchman: 2023.07.24.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 115.0.5790.170
    Edge: 115.0.1901.203
    Safari: 16.5.2
  npmPackages:
    @testing-library/jest-dom: 5.17.0 => 5.17.0 
    @testing-library/react: 13.4.0 => 13.4.0 
    @testing-library/user-event: 13.5.0 => 13.5.0 
    aws-amplify: 5.3.8 => 5.3.8 
    react: 18.2.0 => 18.2.0 
    react-dom: 18.2.0 => 18.2.0 
    react-scripts: 5.0.1 => 5.0.1 
  npmGlobalPackages:
    corepack: 0.15.3
    npm: 9.3.1
    yarn: 1.22.19

```

</details>

### Describe the bug

[...]

PR #11847, opened 2023-08-21
fix: avoid federated signin redirect loop

erinleigh90:
#### Description of changes

Made sure urlListener is registered on Auth instance so the oAuthFlowInProgress refers to the Auth instance that customers are using.
- NOTE: this will be an issue for UI team when they start using InternalAuth, so a more robust solution will need to be found going forward.

#### Issue #, if available
https://github.com/aws-amplify/amplify-js/issues/11846

Reproduced issue in sample app, validated changes using yarn linking.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


elorzafe (approved):

stocaaro (approved):

kvramyasri7 (approved):
Thanks for the Quick Fix Erin

jimblanc (approved):

PR #11860, opened 2023-08-23
feature: Add functional `record` API

jimblanc:
#### Description of changes

This change adds the functional Pinpoint `record` APIs to core and Analytics. `EventBuffer` and related utilities will be migrated in a different PR. Performed some additional cleanup in the analytics category.

Tested in React sample with category & provider level APIs.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/analytics/src/providers/pinpoint/apis/record.ts
```diff
@@ -0,0 +1,51 @@
...
+				credentials,
+				event,
+				identityId,
+				region,
+				userAgentValue: getAnalyticsUserAgentString(AnalyticsAction.Record),
+			});
+		})
+		.catch(e => {
+			// An error occured while fetching credentials or persisting the event to the buffer
+			logger.warn('Failed to record event.', e);
```
HuiSF, line 49: Sanity check: the error caught here won't include any sensitive info, correct?

HuiSF (commented):
The PR seems to include unintended commits (from the other PR of Chris's), maybe need a rebase?

HuiSF (commented):
The PR seems to include unintended commits (from the other PR of Chris's), maybe need a rebase?

jimblanc:
> The PR seems to include unintended commits (from the other PR of Chris's), maybe need a rebase?

This is expected due to how our branches were prepared originally, it'll sort itself out on squash and merge

packages/core/src/providers/pinpoint/apis/record.ts
```diff
@@ -0,0 +1,57 @@
...
+			identityId,
+			region,
+			userAgentValue,
+		});
+
+		endpointId = await getEndpointId(appId, category);
+	}
+
+	if (!endpointId) {
+		throw new Error('Endpoint was not created.');
```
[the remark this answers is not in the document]
jimblanc: Not that I can think of 🤔 Any service error from `updateEndpoint` would get bubbled up before this point, this would basically indicate that there was an issue retrieving the endpoint ID from the cache.

packages/core/src/providers/pinpoint/types/pinpoint.ts
```diff
@@ -34,3 +45,11 @@ export type PinpointUpdateEndpointParameters = PinpointServiceOptions & {
...
 	userAgentValue?: string;
 };
+
+export type PinpointRecordParameters = {
+	event: PinpointAnalyticsEvent;
+	sendImmediately?: boolean;
+} & Omit<
+	PinpointUpdateEndpointParameters,
+	'channelType' | 'userId' | 'userProfile'
+>;
```
cshfang, line 55: I think it would be cleaner for both API parameters to extend/join a base type instead of coupling them with omit. Also, right now UpdateEndpoint params are also now joining service options after the latest round of changes which include optOut and address - which I don't think you need here

packages/analytics/__tests__/providers/pinpoint/apis/testUtils/data.ts
```diff
@@ -0,0 +1,30 @@
...
+export const region = 'region';
+export const accessKeyId = 'access-key-id';
+export const secretAccessKey = 'secret-access-key';
+export const identityId = 'identity-id';
+
+export const event = {
+	name: 'event',
+	properties: {
+		property: 'property-value',
+	},
```
cshfang, line 11: Should this be attributes?

packages/core/__tests__/providers/pinpoint/testUtils/data.ts
```diff
@@ -16,6 +16,15 @@ export const clientDemographic = {
 	platform: 'platform',
 };
 export const endpointId = 'endpoint-id';
+export const event = {
+	name: 'event',
+	properties: {
+		property: 'property-value',
+	},
```
cshfang, line 23: Should this be attributes?

packages/analytics/__tests__/providers/pinpoint/apis/record.test.ts
```diff
@@ -0,0 +1,101 @@
...
+		expect.assertions(1);
+	});
+
+	it('throws a validation error when event does not specify a name', () => {
+		const mockParams = {
+			event: {},
+		};
+
+		try {
+			record(mockParams as PinpointRecordParameters);
```
cshfang, line 94: I think this should be asserted as RecordParameters since it's being called against the Analytics api. Then it probably doesn't need to even be imported

packages/core/__tests__/providers/pinpoint/apis/record.test.ts
```diff
@@ -0,0 +1,132 @@
...
+		await record({
+			appId,
+			category,
+			credentials,
+			event,
+			identityId,
+			region,
+		});
+
+		expect(mockUpdateEndpoint).toBeCalledTimes(0);
```
cshfang, line 48: Nit: Idiomatic to use `.not.toBeCalled()`

packages/core/__tests__/providers/pinpoint/apis/record.test.ts
```diff
@@ -0,0 +1,132 @@
...
+		await record({
+			appId,
+			category,
+			credentials,
+			event,
+			identityId,
+			region,
+		});
+
+		expect(mockClientPutEvents).toBeCalledTimes(0);
```
cshfang, line 84: Nit: Ditto

packages/core/src/providers/pinpoint/apis/record.ts
```diff
@@ -0,0 +1,61 @@
...
+	userAgentValue,
+}: PinpointRecordParameters): Promise<void> => {
+	const timestampISOString = new Date().toISOString();
+	let endpointId = await getEndpointId(appId, category);
+
+	// TODO Prepare event buffer if required
+
+	// Generate endpoint if required
+	if (!endpointId) {
+		await updateEndpoint({
```
cshfang, line 31: Nit: Would add a comment here about why we're "generating an endpoint" by calling this. It's not clear that `updateEndpoint` even performs such a function at a glance.

packages/core/src/providers/pinpoint/apis/record.ts
```diff
@@ -0,0 +1,61 @@
...
+
+		endpointId = await getEndpointId(appId, category);
+	}
+
+	if (!endpointId) {
+		throw new AmplifyError({
+			name: 'ENDPOINT_NOT_CREATED',
+			message: 'Endpoint was not created.'
+		});
+	}
```
cshfang, line 48: Should we move this error into the endpoint utils? I think it could be a little misleading to have it here given that endpoint "creation" is just UUID generation. So either updateEndpoint would have failed by now or something went wrong either caching or retrieving the endpoint id

jimblanc: IMO it makes sense here, but is also extremely unlikely to occur 🤔 `updateEndpoint` will create an endpoint if it doesn't exists, and `getEndpointId` would return the new ID after it was created (otherwise undefined). Alternatively, I could just assert that `getEndpointId` will not be undefined, but I like having this as a safety I think.

packages/core/src/providers/pinpoint/types/pinpoint.ts
```diff
@@ -34,22 +34,23 @@ export type PinpointAnalyticsEvent = {
 	metrics?: Record<string, number>;
 };
 
-export type PinpointUpdateEndpointParameters = PinpointServiceOptions & {
+// Common type that is required for operations that may trigger an endpoint update
+type PinpointCommonEndpointOptions = {
```
cshfang, line 38: Nit: To separate the "options" concept from the wrapping parameters
```suggestion
type PinpointCommonParameters = {
```

packages/core/src/providers/pinpoint/types/pinpoint.ts
```diff
@@ -34,22 +34,23 @@ export type PinpointAnalyticsEvent = {
...
 	appId: string;
 	category: SupportedCategory;
-	channelType?: SupportedChannelType;
 	credentials: AuthSession['credentials'];
 	identityId?: AuthSession['identityId'];
 	region: string;
+	userAgentValue?: string;
+};
+
+export type PinpointUpdateEndpointParameters = PinpointCommonEndpointOptions & PinpointServiceOptions & {
```
cshfang, line 47:
```suggestion
export type PinpointUpdateEndpointParameters = PinpointCommonParameters & PinpointServiceOptions & {
```

packages/core/src/providers/pinpoint/types/pinpoint.ts
```diff
@@ -34,22 +34,23 @@ export type PinpointAnalyticsEvent = {
...
+	channelType?: SupportedChannelType;
 	userId?: string;
 	userProfile?: UserProfile;
-	userAgentValue?: string;
 };
 
-export type PinpointRecordParameters = {
+export type PinpointRecordParameters = PinpointCommonEndpointOptions & {
```
cshfang, line 53:
```suggestion
export type PinpointRecordParameters = PinpointCommonParameters & {
```

cshfang (approved):
:shipit:

PR #11876, opened 2023-08-24
fix(core): remove type assertion for listeners

kvramyasri7:
#### Description of changes

Fixed `unknown` type assertions for Hub apis.
Changed `listener` type in Hub class.

Tested with a sample app

#### Checklist

- [X] PR description included
- [X] `yarn test` passes

### DIFF
--- a/packages/core/src/Hub/index.ts
+++ b/packages/core/src/Hub/index.ts
@@ -185,7 +185,7 @@ export class HubClass {
 		capsule: HubCapsule<Channel, EventDataMap | AmplifyEventData[Channel]>
 	) {
 		const { channel, payload } = capsule;
-		const holder = this.listeners[channel as unknown as number];
+		const holder = this.listeners[channel];
 		if (holder) {
 			holder.forEach(listener => {
 				logger.debug(`Dispatching to ${channel} with `, payload);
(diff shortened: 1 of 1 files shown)


packages/core/src/Hub/index.ts
```diff
@@ -190,7 +185,6 @@ export class HubClass {
 		capsule: HubCapsule<Channel, EventDataMap | AmplifyEventData[Channel]>
 	) {
 		const { channel, payload } = capsule;
-		// TODO[kvramya] need to update mapping of channel to remove type assertion.
 		const holder = this.listeners[channel as unknown as number];
```
erinleigh90, line 188: This also should just be channel

packages/core/src/Hub/index.ts
```diff
@@ -25,7 +25,7 @@ const logger = new Logger('Hub');
 
 export class HubClass {
 	name: string;
-	private listeners: IListener[] = [];
+	private listeners: { [key: string]: IListener } = {};
```
stocaaro, line 28: Can we transition this to [a map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instead of using objects?

stocaaro (approved):

erinleigh90 (approved):

PR #11877, opened 2023-08-24
feat(auth): make a subset of auth apis runnable with the server context

HuiSF:
### `@amplify/core`
* Removed `**/package.json` from the `files` field in the `package.json`
* Now export the low level `fetchAuthSession` from the `/internals` subpath

### `@amplify/auth`
Made the following APIs to be callable with Amplify server context.
* `getCurrentUser`
* `fetchUserAttributes`
* Updated relevant unit tests

* Unit tests
* Making API call in Next.js server.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/auth/src/providers/cognito/apis/server/fetchUserAttributes.ts
+++ b/packages/auth/src/providers/cognito/apis/server/fetchUserAttributes.ts
@@ -1,0 +1,18 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+	AmplifyServer,
+	getAmplifyServerContext,
+} from '@aws-amplify/core/internals/adapter-core';
+import { AuthUserAttribute } from '../../../../types';
+import { CognitoUserAttributeKey } from '../../types';
+import { fetchUserAttributes as fetchUserAttributesInternal } from '../internal/fetchUserAttributes';
+
+export const fetchUserAttributes = (
+	contextSpec: AmplifyServer.ContextSpec
+): Promise<AuthUserAttribute<CognitoUserAttributeKey>> => {
+	return fetchUserAttributesInternal(
+		getAmplifyServerContext(contextSpec).amplify
+	);
+};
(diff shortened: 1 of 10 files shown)


stocaaro (approved):
How do we unit test this type of behavior for the server?

packages/auth/src/providers/cognito/apis/server/fetchUserAttributes.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import {
+	AmplifyServer,
+	getAmplifyServerContext,
+} from '@aws-amplify/core/lib-esm/adapterCore';
+import { AuthUserAttribute } from '../../../../types';
+import { CognitoUserAttributeKey } from '../../types';
+import { fetchUserAttributes as fetchUserAttributesInternal } from '../internal/fetchUserAttributes';
+
+export const fetchUserAttributes = (
```
israx, line 12: should we add doc strings?

HuiSF: Yup, I will add the docs to all server side API as a whole in a separate PR.

HuiSF:
> How do we unit test this type of behavior for the server?

The only difference between server and client is where the Amplify instance comes from. The existing unit tests has been verifying the common share internal logic, so an extra set of unit tests probably unnecessary.

I will add unit tests to all server side APIs to verify `getAmplifyServerContext` function at least in a separate PR.

israx (dismissed):

stocaaro (dismissed):

packages/auth/package.json
```diff
@@ -88,7 +88,7 @@
 		"@aws-amplify/core": "^6.0.0"
 	},
 	"devDependencies": {
-		"@aws-amplify/core": "6.0.0",
+		"@aws-amplify/core": "^6.0.0",
```
jimblanc, line 91:
```suggestion
		"@aws-amplify/core": "6.0.0",
```

The dev dep shouldn't be pinned, it's for our internal dev and lerna will handle it

packages/core/internals/package.json
```diff
@@ -0,0 +1,8 @@
+{
+	"name": "@aws-amplify/core/internals",
```
jimblanc, line 2: Do we need a new path for this? Could we possible re-use `@aws-amplify/core/internals/utils`?

jimblanc (changes requested):
Looks good, one blocking comment on version range

jimblanc (approved):

israx (approved):

PR #11880, opened 2023-08-24
v6 feat(auth) sign out

elorzafe:
#### Description of changes

`signOut` API
- [x] clear local tokens
- [x] token revocation
- [x] logout endpoint for OAuth
- [x] global sign out
- [x] unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/src/providers/cognito/apis/signOut.ts
```diff
@@ -0,0 +1,169 @@
...
+		return clientSignOut();
+	}
+}
+
+async function clientSignOut() {
+	const authConfig = AmplifyV6.getConfig().Auth;
+
+	try {
+		const { refreshToken, accessToken } =
+			await tokenOrchestrator.tokenStore.loadTokens();
```
Samaritan1011001, line 42: Should we be using fetchAuthSession here? and in line 73?

packages/auth/src/providers/cognito/apis/signOut.ts
```diff
@@ -0,0 +1,169 @@
...
+			oAuthSignOutRedirect(authConfig);
+		}
+	} catch (err) {
+		throw new AuthError({
+			message: 'SignOut error',
+			name: 'SignOutError',
+			underlyingError: err,
+		});
+	} finally {
+		tokenOrchestrator.clearTokens();
```
Samaritan1011001, line 64: We should also call `cognitoCredentialsProvider.clear()` and `IdentityIdProvider.clear()`

packages/auth/src/providers/cognito/apis/signOut.ts
```diff
@@ -0,0 +1,169 @@
...
+	// 	'oAuthSignOut',
+	// 	{ oAuth: 'signOut' },
+	// 	`Signing out from ${oAuthLogoutEndpoint}`
+	// );
+	// logger.debug(`Signing out from ${oAuthLogoutEndpoint}`);
+
+	window.open(oAuthLogoutEndpoint, SELF);
+}
+
+async function revokeTokens({
```
erinleigh90, line 137: What is the benefit of this layer of abstraction vs calling revokeToken directly?

packages/auth/src/providers/cognito/apis/signOut.ts
```diff
@@ -0,0 +1,169 @@
...
+			}
+		);
+
+		tokenOrchestrator.clearTokens();
+
+		await handleOAuthSignOut(authConfig);
+		const oauthStore = new DefaultOAuthStore(LocalStorage);
+		oauthStore.setAuthConfig(authConfig);
+		if (authConfig.oauth && oauthStore.loadOAuthSignIn()) {
+			oAuthSignOutRedirect(authConfig);
```
erinleigh90, line 89: Isn't this already being called through handleOAuthSignOut? Does it need to be called twice here?

erinleigh90: I'm wondering if all of this code from 85-89 can be moved to handleOAuthSignOut since it is all repeated above in clientSignOut as well

erinleigh90 (changes requested):
Had a couple of questions but otherwise lgtm

packages/auth/src/providers/cognito/apis/signOut.ts
```diff
@@ -0,0 +1,169 @@
...
+		}
+	} catch (err) {
+		throw new AuthError({
+			message: 'SignOut error',
+			name: 'SignOutError',
+			underlyingError: err,
+		});
+	} finally {
+		tokenOrchestrator.clearTokens();
```
Samaritan1011001, line 98: Same here: We should also call `cognitoCredentialsProvider.clear()` and `IdentityIdProvider.clear()`

packages/auth/src/providers/cognito/apis/signOut.ts
elorzafe: `fetchAuthSession` doesnt return `refreshToken`

packages/auth/src/providers/cognito/apis/signOut.ts
elorzafe: But how can be called if is decoupled?, when `fetchAuthSession` is called, it will not have tokens unless I am missing something.

packages/auth/src/providers/cognito/apis/signOut.ts
elorzafe: yes! good catch!

Thank you!

packages/auth/src/providers/cognito/apis/signOut.ts
elorzafe: Good catch, is doing the same lol

packages/auth/src/types/results.ts
```diff
@@ -19,6 +19,8 @@ export type AuthSignInResult<
 	nextStep: AuthNextSignInStep<UserAttributeKey>;
 };
 
+export type AuthSignOutResult = void;
```
israx, line 22: The result doesn't conform to the stablished API definition

packages/auth/src/providers/cognito/apis/signOut.ts
```diff
@@ -0,0 +1,147 @@
...
+		await globalSignOutClient(
+			{
+				region: getRegion(authConfig.userPoolId),
+			},
+			{
+				AccessToken: accessToken.toString(),
+			}
+		);
+
+		await handleOAuthSignOut(authConfig);
```
israx, line 80: this doesn't resolve to a promise, so we don't really need to await

israx (commented):
My biggest concern is that we are not confirming and aligning to the proposed API definition result

packages/auth/src/providers/cognito/apis/signOut.ts
```diff
@@ -0,0 +1,147 @@
...
+				{
+					ClientId: authConfig.userPoolWebClientId,
+					Token: refreshToken,
+				}
+			);
+		}
+
+		await handleOAuthSignOut(authConfig);
+	} catch (err) {
+		throwSignOutError(err);
```
israx, line 60: I'm not sure we need this util because the underlying client already maps to an `AuthError`

israx: Although I think based on the stablished API definition we would need to **return** the error instead

erinleigh90 (dismissed):
LGTM!

packages/auth/src/providers/cognito/apis/signOut.ts
elorzafe: From our conversation for dev preview it will not throw, we can chat more for GA launch if we can add an E2E story that customers can retry easily to make sure the tokens were revoked or the global signout was done

packages/auth/src/types/results.ts
elorzafe: Is ok if we modify that for GA launch, is adding and not removing.

israx (dismissed):
I'm still hesitant to click the `approve` button as we are not conforming with native. I see the point where exposing `service` errors in the result might not be useful for customers and that there's not a clear scape hatch. But I think the DX is better than just swallowing service errors.

For now I'm okay with this API signature as changing the API result shouldn't be breaking, and considering that we need a discussion.

jimblanc (approved):

packages/auth/src/providers/cognito/apis/signOut.ts
```diff
@@ -0,0 +1,150 @@
...
+
+function oAuthSignOutRedirect(authConfig: AuthConfig) {
+	assertOAuthConfig(authConfig);
+	let oAuthLogoutEndpoint = 'https://' + authConfig.oauth.domain + '/logout?';
+
+	const client_id = authConfig.userPoolWebClientId;
+
+	const signout_uri = authConfig.oauth.redirectSignOut;
+
+	oAuthLogoutEndpoint += Object.entries({
```
jimblanc, line 122: Non-blocking issue, but would prefer if we just used URL to build this

israx (dismissed):

jimblanc (approved):

Samaritan1011001 (approved):

PR #11926, opened 2023-08-29
fix: avoid duplicate instances in Auth/InternalAuth

erinleigh90:
#### Description of changes
Updated AuthClass to not extend InternalAuth, but to call InternalAuth methods. This accomplishes the following:
1. Configure is only being called on a single instance (InternalAuth) through Auth, avoiding duplicate listeners, etc
2. InternalAuth instance is the keeper of all private properties and methods: "this" will always refer to InternalAuth

Because withSSRContext needs to create fresh instances of Auth and InternalAuth, I altered the constructor of AuthClass to accept an instance of InternalAuth that will be used. 

We could potentially use the form of dependency injection used elsewhere (via Amplify._modules), however there is a risk that Auth will not function or be configured without InternalAuth defined first, and this may introduce problems related to race conditions client-side.

- [x] unit tests pass
- [ ] integ tests pass
- [ ] test in sample app with yarn linking

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -23,19 +23,38 @@ import {
 	MFAOption,
 	CognitoUserSession,
 	CognitoUserAttribute,
+	NodeCallback,
 } from 'amazon-cognito-identity-js';
 
 import { AuthError } from './Errors';
 import { IAuthDevice } from './types/Auth';
-import { InternalAuthClass } from './internals/InternalAuth';
+import { InternalAuth, InternalAuthClass } from './internals/InternalAuth';
 
 /**
  * Provide authentication steps
  */
-export class AuthClass extends InternalAuthClass {
+export class AuthClass {
+	private InternalAuth: InternalAuthClass;
+
+	/**
+	 * Initialize Auth with AWS configurations
+	 * @param {Object} config - Configuration of the Auth
+	 */
+	constructor(InteralAuthInstance: InternalAuthClass) {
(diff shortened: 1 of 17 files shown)


packages/auth/src/Auth.ts
```diff
@@ -23,20 +23,39 @@ import {
...
+	private InternalAuth: InternalAuthClass;
+
+	/**
+	 * Initialize Auth with AWS configurations
+	 * @param {Object} config - Configuration of the Auth
+	 */
+	constructor(InteralAuthInstance: InternalAuthClass) {
+		this.InternalAuth = InteralAuthInstance;
+	}
+
```
erinleigh90, line 46: Setting InternalAuth to a specific instance to allow withSSRContext to create fresh instances of both Auth and InternalAuth

We could potentially use the form of dependency injection used elsewhere (via Amplify._modules), however there is a risk that Auth will not function or be configured without InternalAuth defined first, and this may introduce problems related to race conditions client-side.

packages/auth/src/Auth.ts
```diff
@@ -23,20 +23,39 @@ import {
...
 	public getModuleName() {
 		return 'Auth';
 	}
 
+	configure(config?) {
+		return this.InternalAuth.configure(config);
+	}
```
erinleigh90, line 53: Auth.configure now calls configure directly on the instance of InternalAuth, which is acting as the "singleton"-type instance.

PR #11933, opened 2023-08-30
fix(auth): rename assertTokenProviderConfig to assertUserPoolClientIdInConfig

Samaritan1011001:
#### Description of changes

Update the assertion method name and reworded the error message to appropriate message. 
Discovery: Not having a `UserPoolClientId` in config threw Auth Token Provider not found exception which is not correct.

yarn test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/core/src/singleton/Auth/utils/index.ts
```diff
@@ -26,14 +26,12 @@ export function assertTokenProviderConfig(
...
-			!!cognitoConfig.userPoolClientId && !!cognitoConfig.userPoolClientId;
+		assertionValid = !!cognitoConfig.userPoolClientId;
 	}
-
 	return asserts(assertionValid, {
-		name: 'AuthTokenConfigException',
-		message: 'Auth Token Provider not configured',
-		recoverySuggestion: 'Make sure to call Amplify.configure in your app',
+		name: 'UserPoolClientIdConfigException',
+		message: 'UserPoolClientId not present in Config',
```
HuiSF, line 33: nit: prefer that messages end with a period.

HuiSF (approved):

packages/core/src/singleton/Auth/utils/index.ts
```diff
@@ -26,14 +26,12 @@ export function assertTokenProviderConfig(
 	if (!cognitoConfig) {
 		assertionValid = false;
 	} else {
-		assertionValid =
-			!!cognitoConfig.userPoolClientId && !!cognitoConfig.userPoolClientId;
+		assertionValid = !!cognitoConfig.userPoolClientId;
```
israx, line 29: we also need to check if the `cognitoConfig.userPoolId` is not null.  These are required to make API calls and to get the `region` . I think the previous change intended doing it but had a duplication on `userPoolClientId`

israx (changes requested):
Thanks for putting this together. The name is much more clear now. Although the purpose of the assertion function is to check the `userPoolId` and `userPoolClientId` are not falsy.

packages/core/src/singleton/Auth/utils/index.ts
Samaritan1011001: Ah, makes sense! I thought it was a typo but couldn't figure out. I'll update it.

packages/core/src/singleton/Auth/utils/index.ts
```diff
@@ -27,13 +27,12 @@ export function assertTokenProviderConfig(
...
 	}
-
 	return asserts(assertionValid, {
-		name: 'AuthTokenConfigException',
-		message: 'Auth Token Provider not configured',
-		recoverySuggestion: 'Make sure to call Amplify.configure in your app',
+		name: 'UserPoolClientIdConfigException',
+		message: 'UserPoolClientId not present in Config.',
+		recoverySuggestion: 'Make sure to provide UserPoolClientId in the config.',
 	});
```
israx, line 36:
```suggestion
return asserts(assertionValid, {
		name: 'UserPoolConfigException',
		message: 'UserPoolClientId or userPoolId are not present in Config.',
		recoverySuggestion: 'Make sure to provide UserPoolClientId and UserPoolId in the config.',
	});
```
Wondering if we should improve this message as it might be miss leading if customers are passing the `clientId` but not the `userpoolId`

stocaaro:

PR #11949, opened 2023-08-31
chore: Enable strict mode in `aws-amplify`

jimblanc:
#### Description of changes

This change enabled strict mode in `aws-amplify`.

Local testing.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/aws-amplify/src/initSingleton.ts
+++ b/packages/aws-amplify/src/initSingleton.ts
@@ -33,6 +33,8 @@ export const DefaultAmplify = {
 			);
 
 			Amplify.configure(resourceConfig, defaultLibraryOptions);
+		} else {
+			Amplify.configure(resourceConfig, libraryOptions);
 		}
 	},
 	getConfig(): ResourcesConfig {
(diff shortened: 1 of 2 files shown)


packages/aws-amplify/src/initSingleton.ts
```diff
@@ -14,30 +14,26 @@ import {
...
 export const DefaultAmplify = {
 	configure(resourceConfig: ResourcesConfig, libraryOptions?: LibraryOptions) {
-		CognitoUserPoolsTokenProvider.setAuthConfig(resourceConfig.Auth);
-		const defaultLibraryOptions: LibraryOptions = {
-			Auth: {
-				tokenProvider: CognitoUserPoolsTokenProvider,
-				credentialsProvider: cognitoCredentialsProvider,
-			},
-		};
+		if (resourceConfig.Auth && !libraryOptions) {
```
israx, line 17: I see with this change the code bellow will only be executed if there are not `libraryOptions`, whereas the previous code will assign `libraryOptions` to the config if they are passed or not.

israx (commented):
Looks pretty good. Just a question about whether the intention of this change is to only set `libraryOptions` within the library

packages/aws-amplify/src/initSingleton.ts
jimblanc: Good catch, added a fallback case

israx (approved):
lgtm!

stocaaro (approved):

PR #11958, opened 2023-09-01
fix(auth): associate tokens with credentials

Samaritan1011001:
#### Description of changes

Change in idToken will now result in updating the credentials. 
This may happen if the tokens are refreshed while the web app is still in session resulting in credentials in-memory going out of sync with tokens, this fix will make sure credentials stored in-memory is linked to a particular idToken and is updated if that changes.

- log user in
- fetchAuthSession once and check the creds returned
- update the idToken in local storage 
- fetchAuthSession again and see that new creds are returned (could not create a new idToken JWT so made sure with console logs)
- Unit test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


elorzafe (dismissed):

packages/auth/src/providers/cognito/credentialsProvider/credentialsProvider.ts
```diff
@@ -241,6 +240,14 @@ export class CognitoAWSCredentialsAndIdentityIdProvider
 			? true
 			: this._nextCredentialsRefresh <= Date.now();
 	}
+	private hasTokenChanged(tokens?: AuthTokens): boolean {
+		return (
+			!!tokens &&
```
HuiSF, line 245: nit: probably don't need `!!` here?

packages/auth/src/providers/cognito/credentialsProvider/credentialsProvider.ts
```diff
@@ -106,7 +105,7 @@ export class CognitoAWSCredentialsAndIdentityIdProvider
 			this._credentialsAndIdentityId.isAuthenticatedCreds === false
 		) {
 			logger.info(
-				'returning stored credentials as they neither past TTL nor expired'
+				'returning stored credentials as they neither past TTL nor expired.'
```
HuiSF, line 108: 👍

packages/auth/src/providers/cognito/credentialsProvider/credentialsProvider.ts
```diff
@@ -83,16 +83,15 @@ export class CognitoAWSCredentialsAndIdentityIdProvider
...
 			this.clearCredentials();
 		}
-
 		if (!isAuthenticated) {
 			return this.getGuestCredentials(identityId, authConfig.Cognito);
 		} else {
 			assertIdTokenInAuthTokens(tokens);
 			// Tokens will always be present if getCredentialsOptions.authenticated is true as dictated by the type
-			return this.credsForOIDCTokens(authConfig.Cognito, tokens, identityId);
+			return this.credsForOIDCTokens(authConfig.Cognito, tokens!, identityId);
```
HuiSF, line 94: Isn't `assertIdTokenInAuthTokens` narrowing down the types of `tokens` getting rid of `undefined`?

HuiSF (approved):

packages/auth/src/providers/cognito/credentialsProvider/credentialsProvider.ts
Samaritan1011001: We want it to be true only if tokens exist so if it's undefined we gotta return false

packages/auth/__tests__/providers/cognito/credentialsProvider.test.ts
```diff
@@ -219,6 +219,45 @@ describe('Primary Credentials', () => {
...
+				},
+				{
+					IdentityId: 'identity-id-test',
+					Logins: {
+						'cognito-idp.us-east-2.amazonaws.com/us-east-2_Q4ii7edTI':
+							'eyJraWQiOiIyd1FTbElUQ2N0bWVMdTYwY3hzRFJPOW9DXC93eDZDdVMzT2lQbHRJRldYVT0iLCJhbGciOiJSUzI1NiJ9.eyJzdWIiOiIzOGEwODU1Ny1hMTFkLTQzYjEtYjc5Yi03ZTNjNDE2YWUzYzciLCJlbWFpbF92ZXJpZmllZCI6dHJ1ZSwiaXNzIjoiaHR0cHM6XC9cL2NvZ25pdG8taWRwLnVzLWVhc3QtMi5hbWF6b25hd3MuY29tXC91cy1lYXN0LTJfUTRpaTdlZFRJIiwiY29nbml0bzp1c2VybmFtZSI6InRlc3QyIiwib3JpZ2luX2p0aSI6ImRiM2QxOGE1LTViZTAtNDVmOS05Y2RjLTI3OWQyMmJmNzgxZCIsImF1ZCI6IjZ1bG1sYTc0Y245cDlhdmEwZmcxYnV1cjhxIiwiZXZlbnRfaWQiOiJhZjRjMmM5NC04ZTY0LTRkYWYtYjc5ZS02NTE0NTEyMjE3OTAiLCJ0b2tlbl91c2UiOiJpZCIsImF1dGhfdGltZSI6MTY5MDkzMjM0MCwiZXhwIjoxNjkwOTM1OTQwLCJpYXQiOjE2OTA5MzIzNDAsImp0aSI6ImVhM2JmNmNlLWEyZWUtNGJiMC05MjdkLWNjMzRjYzRhMWVjMiIsImVtYWlsIjoiYW16bm1hbm9qQGdtYWlsLmNvbSJ9.i71wkSBPZt8BlBFMZPILJ6RsfDaJx0xqriD9y6ly3LnNB2vNAIOZqPLcCKEi8u0obyoFIK_EY7jKVRva5wbDDcHGt5YrnjT3SsWc1FGVUhrPW6IzEwbfYkUsbVGYjfO1hqTMW7q3FHvJ4yFjLDIUHQe-1_NogYeuhjrNxEupOPmE5-52N4dRriZ0DlHD4fe7gqL8B6AJXr5np1XaxZySU4KpdePwIp1Nb2fkolMEGHvOANHqWdBe5I0vRhAh0MDJ6IxvEr65tnaJNgVQuQaZFR4kQlpjemvB7kaVQ-SpH-tV_zXzqpwr_OEH6dgGMcxIsFrBFC8AGQnGXlSsS-5ThQ',
```
jimblanc, line 236: Can we make these tokens more mockish to avoid security alerts on the repo (due to secret scanning)?

packages/auth/src/providers/cognito/credentialsProvider/credentialsProvider.ts
```diff
@@ -34,6 +34,7 @@ export class CognitoAWSCredentialsAndIdentityIdProvider
 
 	private _credentialsAndIdentityId?: AWSCredentialsAndIdentityId & {
 		isAuthenticatedCreds: boolean;
+		idTokenAssociated?: string;
```
jimblanc, line 37: Nit (idTokenAssociated reads like it should be a bool)

```suggestion
		associatedIdToken?: string;
```

jimblanc (approved):

packages/auth/src/providers/cognito/credentialsProvider/credentialsProvider.ts
Samaritan1011001: It actually does, my vscode was being wonky with the imports. Don't need the ! assertion on it 👍

packages/auth/__tests__/providers/cognito/credentialsProvider.test.ts
Samaritan1011001: Good idea. This type of token string is used in many other tests too so would be better to do it in a separate PR together. One minor hurdle is that decodeJWT expects a valid token string.

jimblanc (approved):

PR #11968, opened 2023-09-01
chore: Rebuild yarn.lock - release pr to main - one generic PAT and an env scoped write PAT

stocaaro:
#### Description of changes
Two big changes:
- A scheduled job to post a PR to update the `yarn.lock` file
- Revise the release process to use a PR to merge from `release` back to `main`

Smaller changes:
- Add owner protection for canaries and yarn.lock (keep these from running on forks)
- Migrate to using a single Personal Access Token (PR and Staging only, eliminate the write token)
- Cache `yarn` separately from `build` files so we can go back to invalidating the cache less frequently

Tested by running on my fork.
- [Release](https://github.com/stocaaro/amplify-js/actions/runs/6053723498) ***Without e2e tests***
- [Post release merge to main workflow](https://github.com/stocaaro/amplify-js/actions/runs/6053726955)
- [PR created](https://github.com/stocaaro/amplify-js/pull/80)
- [Triggers validation](https://github.com/stocaaro/amplify-js/actions/runs/6053737196)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.github/actions/node-and-build/action.yml
+++ b/.github/actions/node-and-build/action.yml
@@ -14,10 +14,22 @@ runs:
       env:
         SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
     - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
-      id: cache-build-artifacts
+      id: cache-yarn-artifacts
       with:
         path: |
           **/node_modules
+        key: ${{ runner.os }}-yarn-artifacts-${{ hashFiles('./amplify-js/yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-yarn-artifacts-
+    - name: Install
+      if: steps.cache-yarn-artifacts.outputs.cache-hit != 'true'
+      run: yarn
+      shell: bash
+      working-directory: ./amplify-js
+    - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
+      id: cache-build-artifacts
+      with:
+        path: |
           **/dist
           **/lib
           **/lib-esm/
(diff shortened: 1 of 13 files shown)


.github/actions/node-and-build/action.yml
```diff
@@ -14,30 +14,36 @@ runs:
...
+        path: |
           **/dist
           **/lib
           **/lib-esm/
           **/es/
           **/esm/
           **/cjs/
           **/packages/core/src/Platform/version.ts
-        key: ${{ runner.os }}-build-artifacts-${{ github.sha }}
+        key: ${{ runner.os }}-build-artifacts-${{ hashFiles('./amplify-js/packages/*/src/**', './amplify-js/packages/*/*.json', './amplify-js/packages/*/*.js') }}
```
ashwinkumar6, line 40: This looks good 💯
thank you

.github/workflows/on-schedule-canary-test.yml
```diff
@@ -6,5 +6,6 @@ on:
 
 jobs:
   canaries:
+    if: github.repository == 'aws-amplify/amplify-js'
```
ashwinkumar6, line 9: is this to prevent canaries from running on forked branched 💯

ashwinkumar6 (dismissed):
LGTM
thanks a lot Aaron

jimblanc (commented):
Looks good! Few comments & questions below. Biggest concern is I'm not seeing where the `required-release` tag gets set

.github/actions/node-and-build/action.yml
stocaaro: < 15s. Looking at the example jobs from the overview, the cache is calculated and downloaded in under 15s. 

There aren't timestamps in the log, but breaking down the steps:

The download rate is ~100Mb/s and there are ~300Mb of cached files to download, so this could take as little as 3s of the total 15s.

If 100% of the remaining time is calculating the hash, then this would be taking 12s. I would bet that it's closer to 8-10s. We could do some testing to get a more accurate estimate. Even at 10s, I would propose caching on the correct changes is saving us substantially more time than the 5m+ rebuild. More than happy to use a different hash set if there's a more precise pattern we might follow to speed up our builds.

.github/workflows/on-schedule-update-yarn-lock.yml
```diff
@@ -0,0 +1,49 @@
...
+          BRANCH: update-yarnlock/${{ steps.current-date.outputs.date }}
+        run: |
+          git fetch origin
+          git checkout -b $BRANCH
+          rm yarn.lock
+          yarn || true
+          git add yarn.lock
+          git commit -am "chore: update yarn.lock"
+          git push origin HEAD
+      - name: Create Pull Request
```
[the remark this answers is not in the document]
stocaaro: I had thought about this, before I spent too much time struggling to get PR verifications running. I'll take a look at adding slack notification to this work today. Should be straightforward.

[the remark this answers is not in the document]
jimblanc: Sounds good, could always come as a follow up PR

jimblanc (dismissed):

Issue #11945, opened 2023-08-30
using sameSite: "lax" with cookies leads to authentication error - special characters in username are encoded in lax vs not from strict


asp3:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Next.js

### Amplify APIs

Authentication

### Amplify Categories

auth

### Environment information

latest amplify version

### Describe the bug

For our use case, we needed to make sure that users were authenticated for requests where they clicked on a link they were sent, which means we had to set sameSite to `lax`.

// awsConfig
```
    cookieStorage: {
        sameSite: "lax",
    },
```

After this, however, all server side requests for Auth.currentSession threw an error, with the message:

```
- error ../node_modules/amazon-cognito-identity-js/lib/CognitoUser.js (1150:15) @ CognitoUser.getSession
- error Error: Local storage is missing an ID Token, Please authenticate
    at new Promise (<anonymous>)
    at new Promise (<anonymous>)
    at new Promise (<anonymous>)
```

I verified that the cookies were available, both from inspect element, and right before the `withSSRContext` call was made.

```
export const createSSRContext = cache(() => {
    // Construct a req object & prepare an SSR enabled version of Amplify
    const req = {
        headers: {
            cookie: headers().get("cookie"),
        },
    };

    const val = new Cookies(req.headers.cookie);
    console.log(val);

    return withSSRContext({ req });
});
```

Here, val has the proper values, which means that the UniversalStorage object should also have them.

I did some investigation into the bug, and I realized that on our deployed site (where `sameSite: "strict"`), the keys are stored as:

`CognitoIdentityServiceProvider.{id}.{email}@-@{site}.idToken`

however, when sameSite: "lax", it becomes

`CognitoIdentityServiceProvider.{id}.{email}%40-%40{site}.idToken`

And therefore, the fetch for the cookies fails!

Anything we can do to resolve this? Its currently stopping users from clicking on links to our site, which is a pretty big source of traffic.

[...]

PR #11984, opened 2023-09-06
fix(core): decode encoded cookie name while creating universal storage

HuiSF:
#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/11945

* Decode the cookie header before using the value to create a cookie store

**Context**

When `Amplify.configure({ ssr: true })`, the Cognito auth tokens will be stored in cookie store. This is handled by the package `amazon-cognito-identity-js` using `js-cookie` under the hood.

Per RFC 6265 [section 4.1.1](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1) and RFC 2616 [section 2.2](https://datatracker.ietf.org/doc/html/rfc2616#section-2.2), a cookie name must be a `token` that doesn't contain any separators. 

In the situation described in the linked issue, the cookie name contains `'@'` as a part of user email address, hence, `js-cookie` encodes it following above the aforementioned standard. However, the `UniversalStorage` creates a cookie store instance using the cookie header value directly, without decoding it, this further caused the issue.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/__tests__/UniversalStorage-browser-test.ts
+++ b/packages/core/__tests__/UniversalStorage-browser-test.ts
@@ -44,6 +44,23 @@ describe(UniversalStorage.name, () => {
 				const universalStorage = new UniversalStorage();
 				expect(universalStorage.store).toMatchObject({ bar: 'barz' });
 			});
+
+			test('decodes encoded cookie name', () => {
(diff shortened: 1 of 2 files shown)


jimblanc (approved):
LGTM

cwomack (approved):
LGTM!

PR #11996, opened 2023-09-08
Geo v6 Migration

thaddmt:
#### Description of changes

- Updates types to match other config conventions (converted casing, renamed some keys, etc)
- Switched to using v6
  - updated all references to Auth to use `fetchAuthSession`
  - Removed `configure` from Geo provider, switched to calling `getConfig` to refresh the config whenever necessary
- Moved size limit checks to aws-amplify base package
- Tested with local sample apps and with `maplibre-gl-js-amplify`. There will be another PR for updating maplibre-gl-js-amplify as well since these two libraries are very closely related

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/geo/src/providers/location-service/AmazonLocationServiceProvider.ts
```diff
@@ -448,40 +438,42 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
-						GeofenceId,
-					} = error;
-					results.errors.push({
-						error: {
-							code: Code,
-							message: Message,
-						},
-						geofenceId: GeofenceId,
+				if (response.Errors) {
+					response.Errors.forEach(error => {
```
stocaaro, line 465: Could probably get away with optional chaining here `response.Errors?.forEach`

stocaaro (approved):
Couple nits. Nothing blocking.

packages/geo/package.json
```diff
@@ -20,16 +19,26 @@
...
 		"build": "yarn clean && yarn build:esm && yarn run build:cjs",
 		"clean": "npm run clean:size && rimraf lib-esm lib dist",
 		"clean:size": "rimraf dual-publish-tmp tmp*",
 		"format": "echo \"Not implemented\"",
 		"lint": "tslint '{__tests__,src}/**/*.ts' && npm run ts-coverage",
-		"ts-coverage": "typescript-coverage-report -p ./tsconfig.build.json -t 86.56"
+		"ts-coverage": "typescript-coverage-report -p ./tsconfig.json -t 84.00"
+	},
+	"typesVersions": {
+		">=3.8": {
```
jimblanc, line 34: We recently made a change to set the floor at 4.0 (will be updated in other categories)

```suggestion
		">=4.0": {
```

packages/geo/package.json
```diff
@@ -56,7 +65,11 @@
 		"@aws-amplify/core": "^6.0.0"
 	},
 	"devDependencies": {
-		"@aws-amplify/core": "6.0.0"
+		"@aws-amplify/core": "6.0.0",
+		"typescript": "5.0.2"
+	},
+	"resolutions": {
```
jimblanc, line 71: Why do we need this resolution? We should try to use the `^2.5.0` in the deps to improve flattening across the lib. If not possible for some reason, I would just update the dependency for Geo instead of using a resolution.

```suggestion
	"resolutions": {
```

packages/geo/src/providers/location-service/AmazonLocationServiceProvider.ts
```diff
@@ -718,19 +702,32 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
 		}
 	}
 
+	private _refreshConfig() {
+		this._config = Amplify.getConfig().Geo?.LocationService;
+		if (!this._config) {
+			const errorString =
+				"No Geo configuration found in amplify config, run 'amplify add geo' to create one and run `amplify push` after";
+			logger.debug(errorString);
+			throw new Error(errorString);
```
jimblanc, line 725: Might not be in scope, but in other categories we standardized on `<category>Error` which extends from Amplify error. Might be good to re-use that pattern, e.g. for auth: https://github.com/aws-amplify/amplify-js/blob/next/release/packages/auth/src/errors/AuthError.ts

```suggestion
			throw new Error(errorString);
```

packages/geo/src/providers/location-service/AmazonLocationServiceProvider.ts
```diff
@@ -718,19 +702,32 @@ export class AmazonLocationServiceProvider implements GeoProvider {
 	 */
 	private async _ensureCredentials(): Promise<boolean> {
```
jimblanc, line 703: Might be missing something, but where in the life-cycle does this called to ensure that credentials are fresh? How does it update the credentials in clients that have been created (or re-created)?

packages/geo/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": "../tsconfig.base.json",
+	"compilerOptions": {
+		"strict": true,
```
jimblanc, line 4: ❤️

packages/geo/tsconfig.json
```diff
@@ -0,0 +1,11 @@
...
+		"importHelpers": true,
+		"typeRoots": ["./node_modules/@types", "../../node_modules/@types"],
+		"types": ["node", "jest"],
+		"skipLibCheck": true
```
jimblanc, line 8: Is this needed? If so, which dep is giving issues?

jimblanc (commented):
Looks good!! Left some comments & questions

packages/geo/package.json
thaddmt: I don't recall the reason now, will look into it

packages/geo/tsconfig.json
thaddmt: There were some issues with the smithy client as I recall. 

Various issues like

```../../node_modules/@smithy/smithy-client/dist-types/extensions/defaultExtensionConfiguration.d.ts:31:51 - error TS2694: Namespace '"/Users/thadtrin/workspace/amplify-js/node_modules/@smithy/types/dist-types/index"' has no exported member 'ChecksumAlgorithm'.

31     checksumAlgorithms(): import("@smithy/types").ChecksumAlgorithm[];
                                                     ~~~~~~~~~~~~~~~~~

../../node_modules/@smithy/smithy-client/dist-types/extensions/retry.d.ts:1:35 - error TS2724: '"@smithy/types"' has no exported member named 'RetryStrategyConfiguration'. Did you mean 'RetryStrategyOptions'?

1 import { Provider, RetryStrategy, RetryStrategyConfiguration, RetryStrategyV2 } from "@smithy/types";
```

packages/geo/src/providers/location-service/AmazonLocationServiceProvider.ts
thaddmt: Each of the wrapped location API's call this helper method first to ensure that credentials are fresh

Example:
```
	public async searchByText(
		text: string,
		options?: SearchByTextOptions
	): Promise<Place[]> {
		const credentialsOK = await this._ensureCredentials();
		if (!credentialsOK) {
		}
```

packages/geo/src/providers/location-service/AmazonLocationServiceProvider.ts
thaddmt: Would you be alright doing this as a follow up to this PR? Just don't want to add scope to this PR as it's been open for a while

stocaaro (dismissed):

packages/geo/src/providers/location-service/AmazonLocationServiceProvider.ts
jimblanc: Yeah sounds good

packages/geo/tsconfig.json
jimblanc: Oh hum, we should create a task to dig into this further. We have a goal to not require customers use `skipLibCheck` when using v6. It might be because we had the `smithy/types` version pinned on `next/main`, a PR is being drafted to remove this. Once that's merged could you retry to see if it improves?

thaddmt: resolved after merging

jimblanc (approved):
Looks good!

calebpollman (approved):
LGTM 🚢

PR #12012, opened 2023-09-12
fix(core): allow credentials to be called without token provider and no guest mode

elorzafe:
#### Description of changes

Disable check for `allowGuestAccess: true` in order to call `credentialsProvider` when no tokens are present

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/src/singleton/Auth/index.ts
+++ b/packages/core/src/singleton/Auth/index.ts
@@ -8,8 +8,6 @@ import {
 	FetchAuthSessionOptions,
 	LibraryAuthOptions,
 } from './types';
-import { asserts } from '../../Util/errors/AssertError';
-import { AUTH_CONFING_EXCEPTION } from '../../Util/Constants';
 
 export function isTokenExpired({
 	expiresAt,
@@ -53,13 +51,6 @@ export class AuthClass {
 		let credentialsAndIdentityId: AWSCredentialsAndIdentityId | undefined;
 		let userSub: string | undefined;
 
-		asserts(!!this.authConfig, {
-			name: AUTH_CONFING_EXCEPTION,
-			message: 'AuthConfig is required',
-			recoverySuggestion:
-				'call Amplify.configure in your app with a valid AuthConfig',
-		});
-
 		// Get tokens will throw if session cannot be refreshed (network or service error) or return null if not available
 		tokens =
 			(await this.authOptions?.tokenProvider?.getTokens(options)) ?? undefined;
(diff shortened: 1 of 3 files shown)


packages/core/src/singleton/Auth/index.ts
```diff
@@ -77,7 +68,7 @@ export class AuthClass {
 						forceRefresh: options.forceRefresh,
 					}
 				);
-		} else if (this.authConfig.Cognito.allowGuestAccess) {
+		} else {
```
Samaritan1011001, line 71: The allowGuestAccess is still being [used](https://github.com/Samaritan1011001/amplify-js/blob/e561195255d4eb0e92bb1f2a4d5c0fbd31b9c28b/packages/auth/src/providers/cognito/credentialsProvider/credentialsProvider.ts#L65) within the CognitoCredentialsProvider 

Are we removing this flag from the config? I would be concerned if so.

packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
```diff
@@ -40,7 +40,7 @@ export function signInWithRedirect(
 	const authConfig = Amplify.getConfig().Auth?.Cognito;
 	assertTokenProviderConfig(authConfig);
 	assertOAuthConfig(authConfig);
-
+	store.setAuthConfig(authConfig);
```
jimblanc, line 43: Readability nit: Would prefer to move `store` declaration ahead of `signInWithRedirect` to make this easier to ready

packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
```diff
@@ -347,45 +347,49 @@ function validateState(state?: string | null): asserts state {
 	}
 }
 
+async function parseRedirectURL() {
```
jimblanc, line 350: Should this change be included here? We'll have merge conflicts with your other PR. PR to pull that change into the correct branch: https://github.com/aws-amplify/amplify-js/pull/12017

elorzafe: I will revert that change

packages/core/src/singleton/Auth/index.ts
elorzafe: Is validated on the provider, `fetchAuthSession` should ignore that setting

Samaritan1011001 (approved):

jimblanc (approved):

PR #12021, opened 2023-09-12
chore(auth): general clean up for credentialsProvider and also rename Credentials from sdk

Samaritan1011001:
#### Description of changes

- General clean up for the CognitoCredentialsProvider
- IdentityIdStore's loadIdentityId now returns Identity or null 
- Rename Credentials to AWSCredentialsIdentity as a BB item

Works as expected for the following scenarios on a React app,
- guest user fetchAuthSession
- guest user -> sign In -> fetchAuthSession -> sign out -> fetchAuthSession
- sign in -> fetchAuthSession -> sign out

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/core/src/clients/types/aws.ts
```diff
@@ -5,7 +5,10 @@ import { MetadataBearer } from '@aws-sdk/types';
 import { Endpoint } from './core';
 import { HttpResponse } from './http';
 
-export type { Credentials, MetadataBearer } from '@aws-sdk/types';
+export type {
+	AwsCredentialIdentity as Credentials,
+	MetadataBearer,
+} from '@aws-sdk/types';
```
elorzafe, line 11: are we doing the thing that Allan did on v5 to generate those and have `@aws-sdk/types` as dev dependency

elorzafe (approved):
Looks good!
qq: what about generating the types instead of making it a dependency

packages/core/src/clients/types/aws.ts
israx: +1

israx (approved):
I think we should add a todo to auto generate the types from the SDK. Other than that, LGTM.

packages/core/src/clients/types/aws.ts
Samaritan1011001: Okay, we can create a follow up PR for this. Thanks

PR #12025, opened 2023-09-13
chore(core): remove no longer used isNode check 

HuiSF:
#### Description of changes

* removed no longer used code that depends on the removed `browserOrNode`
* removed `browserOrNode`
* added `isBrowser()`
* refactored usage of `browserOrNode`
* removed branches in DataStore, PubSub when `isNode: true`, these categories won't be available in SSR context. (with V6 all APIs that can run within SSR context would be exported from `/server` subpaths)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/core/src/ServiceWorker/ServiceWorker.ts
elorzafe, line 1: Do we need this module for V6?

packages/core/src/Util/JS.ts
```diff
@@ -170,19 +170,8 @@ export const isWebWorker = () => {
...
-		process.versions != null &&
-		process.versions.node != null;
-
-	return {
-		isBrowser,
-		isNode,
-	};
-};
+export const isBrowser = () =>
+	typeof window !== 'undefined' && typeof window.document !== 'undefined';
```
elorzafe, line 174: wont this polyfill something on Next.js?

packages/core/src/ServiceWorker/ServiceWorker.ts
HuiSF: Synced up with @jimblanc that we need to keep this around till GA.

packages/core/src/Util/JS.ts
HuiSF: So far in v6 impl. I haven't seen anything that relies on this check to be working in Next.js.

And you can see the relevant refactoring in this PR, all the code paths are not invoked in the current V6.

Mostly noticeable changes reside in the DataStore, which is not in the plan to be running in an environment other than browser in the future.

packages/core/src/ServiceWorker/ServiceWorker.ts
jimblanc: Yeah currently we don't have sign off on deprecating this utility in v6. It's not exported for Dev Preview so it's probably okay to leave it for now until we make a decision in GA.

packages/core/src/Util/Reachability.ts
```diff
@@ -13,10 +13,6 @@ export default class ReachabilityNavigator implements Reachability {
 	> = [];
 
 	networkMonitor(netInfo?: any): Observable<NetworkStatus> {
-		if (browserOrNode().isNode) {
-			return Observable.from([{ online: true }]);
-		}
```
jimblanc, line 18: As a sanity check, do we need to retain this for Node? We could update the check to !isBrowser && !isWebWorker. Might not be needed if it was just required for Datastore though

jimblanc (approved):

stocaaro (approved):

packages/core/src/Util/Reachability.ts
HuiSF: If I understood correctly, this logic basically set online: true when the code is running on Node (assuming running on SSR server that's always online)

Since DataStore won't be supported running in SSR context in the future, I think this can be safely removed.

PR #12031, opened 2023-09-13
feat(data): API V6 migration; 'core' util and config mapping updates; add API 'generateClient' and types; migrate subscription provider; REST client updates

david-mcafee:
#### Description of changes

This PR includes the following changes:
- Migration of `api-v6` work (with some simplification)
- Updates to configuration of `api`, `api-graphql`, and `api-rest`
- Updates to `parseAWSExports` to include config mapping for the API category (including multiple auth modes)
- Type updates
- Updates to `api`, `api-graphql`, and `api-rest` to use `fetchAuthSession` and use the new Amplify singleton
- Internal API category code has been trimmed down to exclude logic that is not being used by `generateClient`
- Addition of `core` singleton types for the API category
- Migration from PubSub to AWSAppSyncRealTimeProvider (with updates)
- REST API updates
- Resource config + library options types for API
- Dependency resolutions have been added to unblock regeneration of `yarn.lock` file

#### Checklist

- [x] PR description included
- [x] `yarn test` passes **(NOTE: Some tests have been disabled. Updating tests to work with V6 will be a fast-follow)**


.github/workflows/push-preid-release.yml
```diff
@@ -10,6 +10,7 @@ on:
     branches:
       # Change this to your branch name where "example-preid" corresponds to the preid you want your changes released on
       - feat/example-preid-branch/main
+      - next-api-v6
```
svidgen, line 13: Is this right? Does this need to match the branch name?

elorzafe, line 13: do we need this?

lerna.json
```diff
@@ -7,7 +7,11 @@
 		"packages/analytics",
 		"packages/storage",
 		"packages/aws-amplify",
-		"packages/adapter-nextjs"
+		"packages/adapter-nextjs",
+		"packages/api",
+		"packages/api-rest",
+		"packages/api-graphql",
+		"packages/amplify-v6-types-package"
```
elorzafe, line 14: I assume this name is approved?

packages/amplify-v6-types-package/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@aws-amplify/types-package-alpha",
```
elorzafe, line 2: This needs to be reviewed before we publish

packages/api-graphql/package.json
```diff
@@ -1,118 +1,118 @@
...
-			}
-		},
-		"coveragePathIgnorePatterns": [
-			"/node_modules/",
-			"dist",
-			"lib",
-			"lib-esm"
-		]
-	}
+  "name": "@aws-amplify/api-graphql",
```
elorzafe, line 2: I am not sure this is the correct spacing, can you check this?

packages/datastore/package.json
```diff
@@ -48,7 +48,7 @@
 		"ssr"
 	],
 	"dependencies": {
-		"@aws-amplify/api": "6.0.0",
+		"@aws-amplify/api": "5.4.2",
```
elorzafe, line 51: I would revert this change

svidgen (dismissed):

packages/core/src/singleton/API/types.ts
```diff
@@ -0,0 +1,30 @@
...
+	};
+	customHeaders: Function;
+};
+
+export type APIConfig = {
+	AppSync?: {
+		defaultAuthMode?: GraphQLAuthMode;
+		region?: string;
+		endpoint?: string;
+		apiKey?: string;
```
elorzafe, line 20: this needs to be removed and use the internal one

elorzafe: on `defaultAuthMode`

lerna.json
david-mcafee: cc @svidgen

packages/amplify-v6-types-package/package.json
david-mcafee: cc @svidgen

packages/api-graphql/package.json
david-mcafee: This was formatted with `Prettier`

packages/core/src/singleton/API/types.ts
david-mcafee: Done

.github/workflows/push-preid-release.yml
david-mcafee: Nope, will remove

lerna.json
david-mcafee: Removed

packages/amplify-v6-types-package/package.json
david-mcafee: This has been removed

.github/workflows/push-preid-release.yml
david-mcafee: Updated, but we actually don't need this. Has been removed.

svidgen (dismissed):

svidgen (approved):

svidgen (approved):

svidgen (approved):

svidgen (dismissed):

elorzafe (approved):

elorzafe (approved):

packages/api-graphql/src/Providers/constants.ts
```diff
@@ -0,0 +1,110 @@
...
+// SPDX-License-Identifier: Apache-2.0
+export { AMPLIFY_SYMBOL } from '@aws-amplify/core/internals/utils';
+
+export const MAX_DELAY_MS = 5000;
+
+export const NON_RETRYABLE_CODES = [400, 401, 403];
+
+export const CONNECTION_STATE_CHANGE = 'ConnectionStateChange';
+
+export enum MESSAGE_TYPES {
```
stocaaro, line 11: I think other packages are moving from enums to string literals, especially in public interfaces.  Should this happen here as well?  Even if the answer is "yes" doesn't need to be in this change.

packages/api-graphql/src/internals/v6.ts
```diff
@@ -0,0 +1,106 @@
...
+	options: GraphQLOptionsV6<FALLBACK_TYPES, TYPED_GQL_STRING>,
+	additionalHeaders?: { [key: string]: string }
+): GraphQLResponseV6<FALLBACK_TYPES, TYPED_GQL_STRING> {
+	/**
+	 * The correctness of these typings depends on correct string branding or overrides.
+	 * Neither of these can actually be validated at runtime. Hence, we don't perform
+	 * any validation or type-guarding here.
+	 */
+	const result = GraphQLAPI.graphql(options, additionalHeaders);
+	return result as any;
```
stocaaro, line 103: Return as `any`?

packages/api/__tests__/API.test.ts
```diff
@@ -1,166 +1,170 @@
...
+// 				.spyOn(RestAPIClass.prototype, 'hasCancelToken')
+// 				.mockImplementation(() => false);
+// 			const api = new API(null);
+// 			const request = Promise.resolve();
+// 			expect(api.cancel(request)).toBe(true);
+// 			expect(graphQLAPICancelSpy).toHaveBeenCalled();
+// 		});
+// 	});
+// });
+// TODO(v6): add tests
```
stocaaro, line 167: Should this happen before this work merges to main?

david-mcafee: @stocaaro unit tests are a fast-follow. It's expected that there are no tests for this release.

elorzafe: we are going to refactor `api-rest` to replace axios, once is that done we will turn the tests on

packages/api-graphql/src/Providers/constants.ts
elorzafe: lets add a TODO, I am working on DataStore changes that now I can address this change as well

packages/api-graphql/src/internals/v6.ts
svidgen: I'm not actually sure what we can do better here. We could return `as unknown as ...`, but that's no less a lie and no more safe, AFAIK.

Unless I'm missing something, there's isn't a meaningful type guard we can add either.

In the case of generated queries, we'll be adding branding to the query strings. In the case of customer provided queries, they'll be able to add their types as generic params. But, we can't infer full types from either of those at runtime, so I'm not sure there's any additional safety or precision we can add here.

I may help to know that the new function/interface essentially exists for the sole purpose of extending our (and the customer's) ability to "instruct" TypeScript about what type hints and auto-completions it should give in a case where the function arguments don't contain enough structure to either infer or validate those types.

I tried to capture this more concisely in the comment above. But, maybe it needs more? Or, am I missing something else?

stocaaro: I'll be honest. I didn't dig too deep on this. Do we really not know anything about the return here? Is `GraphQLResponseV6<FALLBACK_TYPES, TYPED_GQL_STRING>` more refined than `any`. Anything from this that we can include in the typing?  When our user receives the content from this version, is it an `any` or is any a compatability hack that we're uring the return, where the type is cast to the graphql return type?

svidgen (approved):

packages/api-graphql/src/internals/v6.ts
svidgen: Just a hack to bridge the gap between the internal interface and the customer facing one.

I could play with some more precise type lies as a follow-up. There might be opportunity to map into the underlying function. At the time, it seemed better to gain confidence with tests.

stocaaro (approved):

packages/api-graphql/src/internals/v6.ts
stocaaro: Makes sense. Probably a bit of lets pretend. At least there are types on the surface

PR #12033, opened 2023-09-13
chore(auth): add authenticated client side validation

israx:
#### Description of changes

This PR adds client side validation for authenticated users that try to sign-in again.

**Other changes:**
Improve error messaging for users that call APIs that require the user to be authenticated.

tested change in a sample app, where I was able to get an error when trying to call sign-in while authenticated.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


elorzafe (approved):
Looks good

Thanks @israx 🏅

elorzafe (changes requested):
Sorry @israx 

can we add also this check for `signInWithRedirect`

Thanks!

elorzafe (dismissed):

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -630,3 +633,19 @@ export function isMFATypeEnabled(
...
+	let authUser: AuthUser | undefined;
+	try {
+		authUser = await getCurrentUser();
+	} catch (error) {}
+
+	if (authUser && authUser.userId && authUser.username) {
+		throw new AuthError({
+			name: USER_ALREADY_AUTHENTICATED_EXCEPTION,
+			message:
+				'There is already a signed in user because auth tokens were found.',
```
jimblanc, line 647:
```suggestion
				'There is already a signed in user.',
```

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -630,3 +633,19 @@ export function isMFATypeEnabled(
...
+			recoverySuggestion: 'Please call signOut before calling signIn again.',
```
jimblanc, line 648: Just to match tone of the other recovery suggestions elsewhere in the lib.

```suggestion
			recoverySuggestion: 'Call signOut before calling signIn again',
```

packages/auth/src/providers/cognito/utils/types.ts
```diff
@@ -30,8 +31,10 @@ export function assertAuthTokens(
 ): asserts tokens is AuthTokens {
 	if (!tokens || !tokens.accessToken) {
 		throw new AuthError({
-			name: 'Invalid Auth Tokens',
-			message: 'No Auth Tokens were found',
+			name: USER_UNAUTHENTICATED_EXCEPTION,
+			message: 'User needs to be authenticated to call this API',
+			recoverySuggestion:
+				'Please try to sign in first before calling this API again',
```
jimblanc, line 37:
```suggestion
				'Sign in before calling this API again',
```

packages/auth/src/providers/cognito/utils/types.ts
```diff
@@ -41,8 +44,10 @@ export function assertIdTokenInAuthTokens(
 ): asserts tokens is AuthTokens {
 	if (!tokens || !tokens.idToken) {
 		throw new AuthError({
-			name: 'IdToken not present in Auth Tokens',
-			message: 'No IdToken in Auth Tokens',
+			name: USER_UNAUTHENTICATED_EXCEPTION,
+			message: 'User needs to be authenticated to call this API',
+			recoverySuggestion:
+				'Please try to sign in first before calling this API again',
```
jimblanc, line 50:
```suggestion
				'Sign in before calling this API again',
```

packages/auth/src/providers/cognito/utils/types.ts
```diff
@@ -52,8 +57,10 @@ export function assertAuthTokensWithRefreshToken(
 ): asserts tokens is CognitoAuthTokens & { refreshToken: string } {
 	if (!tokens || !tokens.accessToken || !tokens.refreshToken) {
 		throw new AuthError({
-			name: 'Invalid Cognito Auth Tokens',
-			message: 'No Cognito Auth Tokens were found',
+			name: USER_UNAUTHENTICATED_EXCEPTION,
+			message: 'User needs to be authenticated to call this API',
+			recoverySuggestion:
+				'Please try to sign in first before calling this API again',
```
jimblanc, line 63:
```suggestion
				'Sign in before calling this API again',
```

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -630,3 +633,19 @@ export function isMFATypeEnabled(
 	if (!mfaTypes) return false;
 	return mfaTypes.includes(mfaType);
 }
+
+export async function isUserAuthenticated() {
```
jimblanc, line 637: Nit pick but the current name makes it sound like this behaves like a predicate (i.e. returns a bool to know if user is signed in).

```suggestion
export async function assertUserNotAuthenticated() {
```

jimblanc (commented):
Looks good, some nits

packages/auth/src/providers/cognito/utils/signInHelpers.ts
israx: Yeah agree. Sounds more like a predicate. I changed it already. Thanks

jimblanc (approved):

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -630,3 +633,19 @@ export function isMFATypeEnabled(
 	if (!mfaTypes) return false;
 	return mfaTypes.includes(mfaType);
 }
+
+export async function assertUserNotAuthenticated() {
+	let authUser: AuthUser | undefined;
+	try {
+		authUser = await getCurrentUser();
+	} catch (error) {}
```
stocaaro, line 641: Should we do something with this catch? Maybe log?

stocaaro (approved):

packages/auth/src/providers/cognito/utils/signInHelpers.ts
israx: good point. If `getCurrentUser` throws it's because no tokens were found. So in the context of sign-in that is expected always. So logging something doesn't really help here.

TeoChi-CF:
To bad as this change made no longer possible to revalidate user password :(

PR #12041, opened 2023-09-14
feat(adapter-nextjs): let withAmplify read the aws-exports content

HuiSF:
#### Description of changes

* `withAmplify` now reads the content of `aws-exports.js` instead of relying on letting developers to import and pass the object

- Unit tests
- Local testing with sample Next.js app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/adapter-nextjs/src/withAmplify.ts
+++ b/packages/adapter-nextjs/src/withAmplify.ts
@@ -1,8 +1,14 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { ResourcesConfig, parseAmplifyConfig } from 'aws-amplify';
+import * as path from 'path';
+import * as fs from 'fs';
+import { parseAmplifyConfig } from 'aws-amplify';
 import { NextConfig } from 'next';
+import { AmplifyServerContextError } from '@aws-amplify/core/internals/adapter-core';
+
+const configTextRegex = /const awsmobile = ({.*});/s;
+const nextCommands = ['build', 'dev', 'export', 'start'];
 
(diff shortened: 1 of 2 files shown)


packages/adapter-nextjs/src/withAmplify.ts
```diff
@@ -44,3 +46,36 @@ export const withAmplify = (
 
 	return nextConfig;
 };
+
+const getAmplifyConfig = (configPath: string) => {
+	try {
+		const projectRoot = process.cwd();
+		const resolvedPath = path.resolve(projectRoot, getDirArg(), configPath);
+		const fileContent = fs.readFileSync(resolvedPath, 'utf8');
+		const match = configTextRegex.exec(fileContent);
```
AllanZhengYP, line 55: Do we always presume the `aws-exports.js` to be ESM? We should be able to use async `import` no matter `withAmplify` is called from ESM or CommonJS right? 

The downside is the `withAmplify` is now async too. Customers have to be on Next.js > 12.0.10, [next doc](https://nextjs.org/docs/messages/promise-in-next-config#possible-ways-to-fix-it)

AllanZhengYP: If `aws-exports.js` is in CommonJS, depending on whether `withAmplify` is run in ESM or CommonJS, we can import it with either [`module.createRequire`](https://nodejs.org/api/module.html#modulecreaterequirefilename) or plain `require()`

AllanZhengYP: Regarding Next.js usage over 12.0.10, it's probably not that bad? 

Screenshot 2023-09-14 at 12 50 47 PM

HuiSF:
No go. 🫡

PR #12052, opened 2023-09-14
chore(auth): fix unnecessary network calls in getCurrentUser API

israx:
#### Description of changes

This PR fixes the `getCurrentUser` API making unnecessary network calls. It removes the usage of `fetchAuthSession`, instead it calls the `getTokens` method from the internal `Auth` class.

**Other changes**
 - remove the ability to force refresh tokens as this behavior can be used by `fetchAuthSession` instead

tested API in a sample App

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/src/providers/cognito/apis/internal/getCurrentUser.ts
```diff
@@ -4,20 +4,22 @@
...
 import { assertAuthTokens } from '../../utils/types';
+import { CognitoUserPoolsTokenProvider } from '../../tokenProvider';
 
 export const getCurrentUser = async (
 	amplify: AmplifyClassV6,
 	getCurrentUserRequest?: GetCurrentUserRequest
 ): Promise<AuthUser> => {
 	const authConfig = amplify.getConfig().Auth?.Cognito;
 	assertTokenProviderConfig(authConfig);
-	const { tokens } = await fetchAuthSession(amplify, {
```
jimblanc, line 18: Isn't this more of a bug in `fetchAuthSession`? I.e. it should only make a network call if needs to refresh tokens. Would it make more sense to fix it there to avoid using the token providers directly? Thinking about a case where a customer doesn't have cognito providers configured, but calls the cognito API, might be other edge cases

israx: I think this API is mainly focused on the usage of userPools. Checked with the other platforms and they rely on tokens mainly. About the edge case, I think a validation error will be thrown, at the end this API is meant to be a Cognito implementation.

packages/auth/src/providers/cognito/apis/internal/getCurrentUser.ts
```diff
@@ -4,20 +4,22 @@
...
-		forceRefresh: getCurrentUserRequest?.recache ?? false,
+
+	const tokens = await CognitoUserPoolsTokenProvider.getTokens({
+		forceRefresh: getCurrentUserRequest?.recache,
```
elorzafe, line 20: What is this `recache` thing?

israx: That was the BR naming suggestion in the API reviews to by pass cache

packages/auth/src/providers/cognito/apis/internal/getCurrentUser.ts
jimblanc: Got it, that's fair. But isn't it also true that `fetchAuthSession` shouldn't make network calls unless it needs to refresh?

packages/auth/src/providers/cognito/apis/internal/getCurrentUser.ts
```diff
@@ -4,20 +4,22 @@
...
 export const getCurrentUser = async (
 	amplify: AmplifyClassV6,
 	getCurrentUserRequest?: GetCurrentUserRequest
 ): Promise<AuthUser> => {
 	const authConfig = amplify.getConfig().Auth?.Cognito;
 	assertTokenProviderConfig(authConfig);
-	const { tokens } = await fetchAuthSession(amplify, {
-		forceRefresh: getCurrentUserRequest?.recache ?? false,
+
+	const tokens = await CognitoUserPoolsTokenProvider.getTokens({
```
HuiSF, line 19: This is going to break SSR use case as this change uses the client singleton,

HuiSF (changes requested):

AllanZhengYP (approved):
The change looks good to me! We need to fix the token storage issue with `fetchAuthSession` separately.

packages/auth/src/providers/cognito/apis/internal/getCurrentUser.ts
israx: is there away to not have fetchAuthSession for the SSR case ?

packages/auth/src/providers/cognito/apis/internal/getCurrentUser.ts
israx: yeah I mean I just wanted to decoupled credentials from this API.

packages/auth/src/providers/cognito/apis/internal/getCurrentUser.ts
israx: removing this flag after discussing async

packages/auth/src/providers/cognito/apis/internal/getCurrentUser.ts
israx: just pushed a solution based on your recommendation. Now tokens are retrieved from the internal Auth class

packages/auth/src/providers/cognito/apis/getCurrentUser.ts
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { Amplify } from '@aws-amplify/core';
-import { AuthUser, GetCurrentUserRequest } from '../../../types';
```
HuiSF, line 5: Probably can remove `GetCurrentUserRequest` from types now.

HuiSF (approved):

stocaaro (approved):

stocaaro (approved):

PR #12095, opened 2023-09-20
chore: Refactor AppSyncRealtimeProvider tests to run in api-graphql c…

stocaaro:
#### Description of changes
This change moves the `__tests__/AWSAppSyncRealTimeProvider.test.ts` file from `pubsub` into `api-graphql`.

From here, the minimum changes possible to make the tests work in the refactored environment where made.

This is step 1 in a multipart piece of work to complete the AppSyncRealtimeProvider migration:
- ~~Get AppSyncRealtimeProvider under test~~
- Move shared (now duplicated) code into `core`
- Migrate from `zen-observable` to `RxJS`

Open question:
- The `AppSyncRealtimeProvider` migrated version has removed a bunch of auth validation that was included in the prior implementation, was this expedient or is the current version preferred? Eg. Is there a work item to re-add validation and reintroduce validation tests?

Running the tests over and over.  I confirmed the packed builds and the tests run before  merging the work.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -10,17 +10,61 @@ jest.mock('@aws-amplify/core', () => ({
...
 
 import Observable from 'zen-observable-ts';
-import { Reachability, Credentials, Logger, Signer, Cache } from '@aws-amplify/core';
-import { Auth } from '@aws-amplify/auth';
-
+import { Reachability } from '@aws-amplify/core/internals/utils';
+import { fetchAuthSession, Cache } from '@aws-amplify/core';
+import { ConsoleLogger as Logger } from '@aws-amplify/core/internals/utils';
+import * as Auth from '@aws-amplify/auth';
+import { GraphQLAuthMode } from '@aws-amplify/core/lib-esm/singleton/API/types';
```
elorzafe, line 17: nit: not a big fan of Import `lib-esm` if you try to run it before build it will fail

packages/api-graphql/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -10,17 +10,61 @@ jest.mock('@aws-amplify/core', () => ({
...
+import {
+	delay,
+	FakeWebSocketInterface,
+	replaceConstant,
+} from '../__tests__/helpers';
+import { ConnectionState as CS } from '../src/types/PubSub';
 
 import { AWSAppSyncRealTimeProvider } from '../src/Providers/AWSAppSyncRealTimeProvider';
-import { loggers } from 'winston';
+import { AuthSession } from '@aws-amplify/core/lib-esm/singleton/Auth/types';
```
elorzafe, line 29: same as the above

packages/api-graphql/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -942,235 +970,104 @@ describe('AWSAppSyncRealTimeProvider', () => {
...
 						expect(loggerSpy).toBeCalledWith(
 							'DEBUG',
-							'Authenticating with OPENID_CONNECT'
+							'Authenticating with {"type":"jwt","token":"id"}'
 						);
 					});
 
-					test('authenticating with AMAZON_COGNITO_USER_POOLS', async () => {
-						expect.assertions(1);
+					// test('authenticating with AMAZON_COGNITO_USER_POOLS', async () => {
```
elorzafe, line 1040: what happened with this one, it was very close

packages/api-graphql/__tests__/helpers.ts
```diff
@@ -0,0 +1,379 @@
+import { Hub } from '@aws-amplify/core';
+import Observable from 'zen-observable-ts';
```
elorzafe, line 2: David already changed this to be `rxjs` on `api-graphql`

packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -932,7 +934,7 @@ export class AWSAppSyncRealTimeProvider {
 		const session = await fetchAuthSession();
 
 		return {
-			Authorization: session.tokens.accessToken.toString(),
+			Authorization: session?.tokens?.accessToken?.toString(),
```
elorzafe, line 937: can we add a TODO here, after is finalized the config type, customers could customize using `id` or `access` token

packages/api-graphql/__tests__/helpers.ts
stocaaro: Yep. Was planning to change this to RxJS as a subsequent change. Working on that now with the guidance not to move shared code to core.

packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts
stocaaro: I don't have enough context to make a comment here that I would understand later. Can you either propose the comment you want or provide more detail so that I can accomplish what you're after?

packages/api-graphql/__tests__/AWSAppSyncRealTimeProvider.test.ts
stocaaro: I wasn't sure what to do with user pools. It looks like when you ported it across, you removed the `_awsRealTimeCUPHeader` method call from the handler mapping and didn't include a user pool oriented key in `GraphQLAuthModeKeys` over on `core`. I don't understand enough about whats changing in `auth` to know if this was intentional or incidental, can you give me more context?  I suppose I could chase this down, but I would need to extend timelines to make room to get up to speed on whats happening around this on the `auth` side.

svidgen (commented):
It's unclear to me whether the removed tests indicate *intentionally* removed behavior. I think we need @elorzafe to weigh in and either provide followup tasks to re-add these or guidance on putting them back in.

My current read is that we've removed signals for customers to help diagnose failing subscriptions, and we need to put them back in. But, I'm happy to hit the approve button if @elorzafe can either assure me I've misunderstood or show me tasks to add this back in pre-launch.

packages/api-graphql/__tests__/AWSAppSyncRealTimeProvider.test.ts
elorzafe: Yeah, because the library on v6 will be reading the tokens from Auth provider (could be any OIDC provider) you only need one

packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts
elorzafe: I think is ok to not increase scope and we can add that in the future

packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -980,7 +982,7 @@ export class AWSAppSyncRealTimeProvider {
 			},
 			{
 				credentials: creds,
-				signingRegion: endpointInfo.region,
+				signingRegion: endpointInfo.region ?? '',
```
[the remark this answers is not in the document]
stocaaro: Yeah. I think my ts got overly strict for a moment there. I am able to remove this without any ts errors now because these types aren't optional.

stocaaro:
@svidgen On your broader theme of intentionally removed behavior, from chat, we discovered:
- The User Pools behavior has been merged with the JWT behavior and we no longer need that code / test
- The auth validation input checking is no longer needed with the new auth implementation

svidgen (dismissed):

elorzafe (approved):
Looks good

Thanks @stocaaro 🏅

svidgen (approved):

PR #12102, opened 2023-09-21
chore(core,auth): modularize oauth helpers

HuiSF:
### `/packages/core`
1. added `convert` utils to the `core` package, currently `base64Encoder` and `base64Decoder`
    * The `.native` implementation uses `base-64` package
2. added `getGlobal` util function to the `core` for resolving `crypto`, `atob` and `btoa` from the global

NOTE: the `base-64` and `react-native-get-random-values` added as dev dependencies will be moved into the upcoming `@aws-amplify/react-native` package.

### `/packages/auth`
1. refactored to use the newly added `base64Encoder` and `base64Decoder`
2. refactored the utils functions used in the oauth flow, and make the functions are usable in react-native

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


israx (dismissed):
SRP changes look fine to me. Thanks @HuiSF . It's good to not use the '@smithy/util-base64' library. Wondering if in the future we can orchestrate our own helpers

packages/auth/src/providers/cognito/utils/oauth/generateCodeVerifier.ts
```diff
@@ -0,0 +1,66 @@
...
+	method: 'S256';
+	toCodeChallenge: () => string;
+} => {
+	const randomBytes = new Uint8Array(length);
+	getCrypto().getRandomValues(randomBytes);
+
+	let value = '';
+	let codeChallenge: string | undefined;
+
+	for (let i = 0; i < randomBytes.length; i++) {
```
cshfang, line 34: Nit: Think we can use `for..of`?

packages/auth/src/providers/cognito/utils/oauth/generateCodeVerifier.ts
```diff
@@ -0,0 +1,66 @@
...
+ * `toCodeChallenge` to generate the code challenge from the `codeVerifier`
+ * following the spec of [RFC 7636](https://datatracker.ietf.org/doc/html/rfc7636#section-4.2).
+ */
+export const generateCodeVerifier = (
+	length: number
+): {
+	value: string;
+	method: 'S256';
+	toCodeChallenge: () => string;
+} => {
```
cshfang, line 27: Nit: Use the `CodeVerifier` type?

packages/auth/__tests__/providers/cognito/utils/oauth/oauth.test.ts
```diff
@@ -0,0 +1,133 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
cshfang, line 1: Nit: Rename as `index.test.ts` to align with actual file under test

packages/core/src/utils/globalHelpers.ts
```diff
@@ -0,0 +1,54 @@
...
+		return crypto;
+	}
+
+	throw new AmplifyError({
+		name: 'MissingPolyfill',
+		message: 'Cannot resolve the `crypto` function from the environment.',
+	});
+};
+
+export const getBtoA = () => {
```
cshfang, line 22: Nit: Imo, since the actual function is called btoa, this should align: `getBtoa`. Ditto for `atob`

packages/core/src/utils/convert/base64/base64Encoder.native.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
cshfang, line 1: Instead of doing this, could we have a `globalHelper.native.ts` that returns `encode/decode` from `getBtoa/getAtob`?

Concern is that someone may want to use those helpers directly but find that it doesn't work in RN

packages/auth/src/providers/cognito/utils/oauth/index.native.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// import the polyfill for `crypto.getRandomValues` for react-native
+// so the subsequent call to `generateCodeVerifier` can use it
+import 'react-native-get-random-values';
```
cshfang, line 6: Could this also move to a proposed `globalHelpers.native.ts`?

packages/auth/src/providers/cognito/utils/oauth/generateCodeVerifier.ts
HuiSF: The return type is not expected to be reused, inline type has a better visibility in IDE hovering the function.🤔

cshfang: What is the `CodeVerifier` type used for?

packages/core/__tests__/utils/covert/base64Decoder.test.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
cshfang, line 1: Parent directory typo `covert` -> `convert`

cshfang (approved):

cshfang (dismissed):

cshfang (approved):

israx (approved):

Issue #2087, opened 2018-11-12
Auth.confirmSignIn throw exception "Invalid device key given." 


ruiyang:
**Describe the bug**

**To Reproduce**
1. Create a cognito pool
2. Create a user with email and phone number verified
3. Enable SMS 2FA
4. Call Auth.signIn. (using the USER_PASSWORD_AUTH).
5. The response will have challengeName="SMS_MFA"
6. Call Auth.confirmSignIn, the promise got rejected with error "Invalid device key given."

There are two calls in confirmSignIn:
1. Confirmed the MFA code - Success
2. Confirm the device - Failed.
The user actually successfully logged in. Because Cognito wants to remember the user 's device, so triggered the second request, which failed. 

**Expected behavior**
Auth.confirmSignIn should run successfully.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [Mac]

**Smartphone (please complete the following information):**

**Additional context**

***You can turn on the debug mode to provide more info for us by setting window.LOG_LEVEL = 'DEBUG'; in your app.***

Curl copy of the failed request from Chrome:
curl 'https://cognito-idp.ap-southeast-2.amazonaws.com/' -H 'Referer: http://localhost:3000/public/login' -H 'Origin: http://localhost:3000' -H 'X-Amz-Target: AWSCognitoIdentityProviderService.ConfirmDevice' -H 'X-Amz-User-Agent: aws-amplify/0.1.x js' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.77 Safari/537.36' -H 'Content-Type: application/x-amz-json-1.1' --data-binary '{"DeviceKey":"...","AccessToken":"...","DeviceSecretVerifierConfig":{"Salt":"...","PasswordVerifier":"..."},"DeviceName":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.77 Safari/537.36"}' --compressed

jonixj:
I get the same issue when using phone_number as username. When I use the GUUID as username I do not get this error

powerful23:
@gbenattar how did you get this error? I tried to follows @ruiyang 's steps but couldn't reproduce the error.

powerful23:
[...]

PR #12107, opened 2023-09-22
feat(auth): add ConfirmDevice API

israx:
#### Description of changes

This API is used to kick off the device management flow. 

`NewDeviceMetadata` will be included on the response of `InitiateAuth` or `RespondToAuthChallenge` if `Device Tracking` is enabled on the User Pool. Then, the library should be able to call `ConfirmDevice` on every `NewDeviceMetadata` and being able to store it in the storage mechanism. 

Validate `ConfirmDevice` is called in a sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -648,3 +657,68 @@ export async function assertUserNotAuthenticated() {
...
+					resolve(undefined);
+					return;
+				}
+
+				const deviceSecretVerifierConfig = {
+					Salt: toBase64(fromHex(authenticationHelper.getSaltToHashDevices())),
+					PasswordVerifier: toBase64(
+						fromHex(authenticationHelper.getSaltToHashDevices())
+					),
+				};
```
ashwinkumar6, line 698:
```suggestion
				const saltToHash = toBase64(fromHex(authenticationHelper.getSaltToHashDevices()));
				const deviceSecretVerifierConfig = {
					Salt: saltToHash,
					PasswordVerifier: saltToHash
				};
```

ashwinkumar6 (dismissed):
Thank you

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -648,3 +657,68 @@ export async function assertUserNotAuthenticated() {
...
+ * This function is used to kick off the device management flow.
+ *
+ * If an error is thrown while generating a hash device or calling the `ConfirmDevice`
+ * client, then this API will ignore the error and return undefined. Otherwise the authentication
+ * flow will not complete and the user won't be able to be signed in.
+ *
+ * @returns DeviceMetadata | undefined
+ */
+export async function getNewDeviceMetatada(
+	amplify: AmplifyClassV6,
```
elorzafe, line 671: what is this?

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -648,3 +657,68 @@ export async function assertUserNotAuthenticated() {
...
+					);
+
+					resolve({
+						deviceKey,
+						deviceGroupKey,
+						randomPassword,
+					});
+				} catch (error) {
+					// TODO: log error here
+					resolve(undefined);
```
elorzafe, line 719: why not reject?

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -648,3 +657,68 @@ export async function assertUserNotAuthenticated() {
...
+	const deviceGroupKey = newDeviceMetadata?.DeviceGroupKey;
+
+	return new Promise((resolve, _) => {
+		authenticationHelper.generateHashDevice(
+			deviceGroupKey ?? '',
+			deviceKey ?? '',
+			async (errGenHash: unknown) => {
+				if (errGenHash) {
+					// TODO: log error here
+					resolve(undefined);
```
elorzafe, line 689: why not reject?

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
```diff
@@ -140,7 +146,6 @@ export class DefaultTokenStore implements AuthTokenStore {
 			this.getKeyValueStorage().removeItem(authKeys.idToken),
 			this.getKeyValueStorage().removeItem(authKeys.clockDrift),
 			this.getKeyValueStorage().removeItem(authKeys.refreshToken),
-			this.getKeyValueStorage().removeItem(authKeys.NewDeviceMetadata),
```
elorzafe, line 143: Is not deleting deviceMetadata to avoid  MFA right?

packages/auth/src/providers/cognito/apis/signInWithUserPassword.ts
```diff
@@ -72,7 +73,14 @@ export async function signInWithUserPassword(
 			challengeName: ChallengeName as ChallengeName,
 		});
 		if (AuthenticationResult) {
-			await cacheCognitoTokens(AuthenticationResult);
+			await cacheCognitoTokens({
+				...AuthenticationResult,
+				NewDeviceMetadata: await getNewDeviceMetatada(
```
elorzafe, line 78: Is this always needed? My understand is that is only when remember device is in place?

elorzafe (commented):
Looks very good overall, I am not sure if this is needed to be called always?

packages/auth/src/providers/cognito/apis/signInWithUserPassword.ts
israx: It's not always needed. This function will not run if there's no `NewDeviceMetadata` in the `AuthenticationResult`.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
israx: yes that is correct.

packages/auth/src/providers/cognito/utils/signInHelpers.ts
israx: because otherwise the user will not be authenticated. 

Here's a closed [issue](https://github.com/aws-amplify/amplify-js/issues/2087) that still is relevant . Confirmed that `swift` does the same. They swallowed the error and authenticate the user.

packages/auth/src/providers/cognito/utils/signInHelpers.ts
israx: because otherwise the user will not be authenticated.

Here's a closed https://github.com/aws-amplify/amplify-js/issues/2087 that still is relevant . Confirmed that swift does the same. They swallowed the error and authenticate the user.

packages/auth/src/providers/cognito/utils/signInHelpers.ts
israx: That was to account for SSR cases but I'm passing the `userPoolId` directly, which can be extracted from the config

elorzafe (approved):
Thanks @israx 

great work 💪

ashwinkumar6 (approved):
thank you @israx !!!

PR #12139, opened 2023-09-27
feat(react-native): setup package + big int native module

HuiSF:
- Includes an example app for the ease of developing and testing the native module

#### Description of changes

* Set up the package `@aws-amplify/react-native`
* Added the native module `AmplifyRTNCore`
* Rewrote the `computeModPow` and `computeS` in Swift and Kotlin
* Added an example app in `packages/react-native` for developing and testing the native modules

**NOTE:** peer dependencies list update to follow

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/react-native/android/build.gradle
```diff
@@ -0,0 +1,108 @@
...
+  if (major == 7 && minor >= 3) {
+    return true
+  }
+
+  return major >= 8
+}
+
+android {
+  if (supportsNamespace()) {
+    namespace "com.awsamplify.reactnative"
```
cshfang, line 54: What is this for? Maybe it's for the codegen stuff below - do we need all that if this is not yet a turbomodule?

Should it just align anyway with the package name in the manifest?

packages/react-native/android/build.gradle
```diff
@@ -0,0 +1,108 @@
...
+
+    sourceSets {
+      main {
+        manifest.srcFile "src/main/AndroidManifestNew.xml"
```
cshfang, line 58: Are we supporting multiple manifests at this time?

packages/react-native/example/metro.config.js
```diff
@@ -0,0 +1,39 @@
+const { getDefaultConfig, mergeConfig } = require('@react-native/metro-config');
+const path = require('path');
+const escape = require('escape-string-regexp');
+const exclusionList = require('metro-config/src/defaults/exclusionList');
+const pak = require('../package.json');
+
+const root = path.resolve(__dirname, '..');
+const modules = Object.keys({ ...pak.peerDependencies });
+
```
cshfang, line 9: Is this config what would be expected of a react native app? I.e. are there opinionated settings that were generated that might mislead our development against this example app?

packages/react-native/example/package.json
```diff
@@ -0,0 +1,37 @@
...
+	"scripts": {
+		"android": "react-native run-android",
+		"ios": "react-native run-ios",
+		"start": "react-native start",
+		"build:android": "cd android && ./gradlew assembleDebug --no-daemon --console=plain -PreactNativeArchitectures=arm64-v8a",
+		"build:ios": "cd ios && xcodebuild -workspace AmplifyRTNCoreExample.xcworkspace -scheme AmplifyRTNCoreExample -configuration Debug -sdk iphonesimulator CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++ GCC_OPTIMIZATION_LEVEL=0 GCC_PRECOMPILE_PREFIX_HEADER=YES ASSETCATALOG_COMPILER_OPTIMIZATION=time DEBUG_INFORMATION_FORMAT=dwarf COMPILER_INDEX_STORE_ENABLE=NO"
+	},
+	"dependencies": {
+		"react": "18.2.0",
+		"react-native": "0.72.5"
```
cshfang, line 14: Thoughts on pinning these to our lowest supported?

packages/react-native/package.json
```diff
@@ -0,0 +1,53 @@
+{
+	"name": "@aws-amplify/react-native",
+	"version": "1.0.0",
+	"description": "React Native module polyfill for aws-amplify",
```
cshfang, line 4: Nit:
```suggestion
	"description": "React Native core module for aws-amplify",
```

cshfang (dismissed):

packages/react-native/example/metro.config.js
HuiSF: This config is meant for able to use the resource from the parent directory of the example app, without workspace linking.

packages/react-native/android/build.gradle
HuiSF: I think this is generated while creating the lib the value should be interfered from the package name, let me double check.

HuiSF: This is a feature provided by gradle that enabling different source sets for different namespace, e.g. release vs testing. We don't need this though for this module. Will remove it.

packages/react-native/example/metro.config.js
HuiSF: To clarify this is not something required to use the native modules.

packages/react-native/android/build.gradle
```diff
@@ -0,0 +1,85 @@
...
+  //noinspection GradleDynamicVersion
+  implementation "com.facebook.react:react-native:+"
+  implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+}
+
+if (isNewArchitectureEnabled()) {
+  react {
+    jsRootDir = file("../src/")
+    libraryName = "AmplifyRTNCore"
+    codegenJavaPackageName = "com.awsamplify.reactnative"
```
elorzafe, line 83: should the package name be `aws.amplify.reactnative`?

packages/react-native/android/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+	package="com.amazonaws.amplify.rtncore">
```
elorzafe, line 3: should this be consistent to https://github.com/aws-amplify/amplify-js/pull/12139/files#diff-d8ba60726cba14a5871dd1f12ea100bf1ce26b543aa42866c49cc82217cc1fa7R83 ?

packages/react-native/package.json
```diff
@@ -0,0 +1,56 @@
...
+		"./lib/index": "./lib-esm/index.js"
+	},
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/aws-amplify/amplify-js.git"
+	},
+	"author": "Amazon Web Services",
+	"license": "Apache-2.0",
+	"bugs": {
+		"url": "https://github.com/aws/aws-amplify/issues"
```
elorzafe, line 44: We can update that to `https://github.com/aws-amplify/amplify-js/`

packages/react-native/example/metro.config.js
cshfang: My concern is that if we develop against an app with non-standard configurations - we are magicking away some things that maybe are not there in a default RN app. But this is not really a blocking concern tbh.

cshfang (approved):

packages/react-native/android/build.gradle
HuiSF: Ahh I should clean this up. 👍

packages/react-native/example/metro.config.js
HuiSF: This metro config here doesn't change how the native module gets linked in a default RN app. It just tells the bundler where to look up the other dependencies (e.g. the dependencies of `react-native` package) from the yarn workspace, so that the example app can run with the workspace hoisted dependencies

packages/react-native/android/src/main/AndroidManifest.xml
HuiSF: This is the correct one, fixed the other place 🙏

cshfang (approved):

jimblanc (approved):

PR #12140, opened 2023-09-27
feat(auth): handles DEVICE_SRP_AUTH challengeName 

israx:
#### Description of changes

This PR handles the DEVICE_SRP_AUTH challengeName that is returned when a DEVICE is remembered in the Cognito User Pool. This change will allow to bypass MFA if this option was previously enabled in the user pool.

Tested the following steps in a sample app:

1. Enabled `Always remember` on the `Device Tracking` section of the Cognito UserPool console.
2. Enabled `Require MFA`
3. Login using any `authFlow` (SRP, CUSTOM, etc)
4. Handle MFA 
5. Log out
6. Login again.
7. User should be authenticated without having to address MFA.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


ashwinkumar6 (approved):
Thanks for adding this in !!

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -359,15 +411,112 @@ export async function handleCustomSRPAuthFlow(
 	);
 }
 
+async function handleDeviceSRPAuth(
```
HuiSF, line 414: nit: I understand this is an internal function, but it may be worth to use named parameters?

packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -359,15 +411,112 @@ export async function handleCustomSRPAuthFlow(
...
+	session: string | undefined,
+	authenticationHelper: AuthenticationHelper,
+	{ userPoolId, userPoolClientId }: CognitoUserPoolConfig,
+	tokenOrchestrator?: AuthTokenOrchestrator
+): Promise<RespondToAuthChallengeCommandOutput> {
+	const deviceMetadata = await tokenOrchestrator?.getDeviceMetadata();
+	assertDeviceMetadata(deviceMetadata);
+
+	const serverBValue = new (BigInteger as any)(challengeParameters?.SRP_B, 16);
+	const salt = new (BigInteger as any)(challengeParameters?.SALT, 16);
```
HuiSF, line 470: We can probally add an interface to make the ts happy?

In `BigInteger.ts`

```typescript
interface BigInteger {
	new (a?: any, b?: any): BigInteger;
}
```

and 

```typescript
export default BigInteger as BigInteger;
```

HuiSF (approved):

ashwinkumar6 (approved):

PR #12144, opened 2023-09-27
feat: Update signInWithRedirect to work with RN

cshfang:
#### Description of changes

This PR updates the `signInWithRedirect` API to use `@aws-amplify/rtn-web-browser` to open the web login in either an ephemeral `ASWebAuthenticationSession` or Custom Tab to ensure a higher security bar

* Tested with an React Native app on both iOS and Android
* `yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
```diff
@@ -105,8 +105,18 @@ function oauthSignIn({
 		.join('&');
 
 	// TODO(v6): use URL object instead
-	const URL = `https://${oauthConfig.domain}/oauth2/authorize?${queryString}`;
-	window.open(URL, SELF);
+	const oAuthUrl = `https://${domain}/oauth2/authorize?${queryString}`;
+	const { type, url } = (await openAuthSession(oAuthUrl, redirectSignIn)) ?? {};
```
HuiSF, line 109: I think we should make the `openAuthSession` function to have the same interface on both Web and native, so the logic here doesn't reply on one of the variant returning a `undefined`.  Wdyt?

HuiSF (dismissed):

packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
cshfang: I think it is a fair point. The result type can't really happen but at least both can be Promisified even though the web version really doesn't need to be. Let me follow this up separately in a PR for web/expo support

jimblanc (dismissed):
Looks good!

packages/core/__tests__/utils/globalHelpers/globalHelpers.native.test.ts
```diff
@@ -31,8 +31,6 @@ describe('getGlobal (native)', () => {
 		});
 	});
 
-	afterAll;
```
HuiSF, line 34: After all 🤧

HuiSF (approved):

jimblanc (approved):

PR #12157, opened 2023-09-29
feat(react-native): module loaders and relevant refactoring

HuiSF:
#### Description of changes

* Moved native packages out from the dev dependencies for `packages/core` and `packages/datastore` added `packages/react-native` as dev dependencies
* Added module loaders to the `packages/react-native` to support usage from other packages
* Added relevant native implementation to the SRP sign in flow so it works with react-native

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


israx (approved):
SRP refactor changes look good to me!

packages/react-native/package.json
```diff
@@ -23,12 +23,23 @@
...
 	},
+	"dependencies": {
+		"base-64": "1.0.0",
+		"buffer": "6.0.3"
+	},
 	"peerDependencies": {
-		"react-native": ">=0.70"
+		"react-native": ">=0.70",
+		"react-native-get-random-values": ">=1.9.0",
+		"react-native-url-polyfill": ">=2.0.0"
```
cshfang, line 33: Should this be a dep instead?

packages/react-native/src/moduleLoaders/loadAsyncStorage.ts
```diff
@@ -0,0 +1,24 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
```
cshfang, line 3: I think we should add comments to all the loaders just so people don't ask the same questions:
1. Why not extract as a common function and use dynamic string for require
2. Why not extract error as function and import

cshfang (approved):

jimblanc (approved):
Looks good, nice work

jimblanc (approved):

PR #12159, opened 2023-09-29
feat(InApp): functional identifyUser API

Samaritan1011001:
#### Description of changes

Functional identifyUser for InAppMessaging. Also removes and moves some types around for a better folder structure. 

Monitored the network request

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/notifications/src/index.ts
jimblanc, line 1: Would you mind setting up an export test in `aws-amplify` -> `exports.test.ts` so we can track the export surface as the category is fleshed out?

jimblanc (commented):
Looking good! Couple nits

packages/notifications/src/inAppMessaging/index.ts
```diff
@@ -2,16 +2,3 @@
...
-	InAppMessageAction,
-	InAppMessageButton,
-	InAppMessageContent,
-	InAppMessageImage,
-	InAppMessageInteractionEvent,
-	InAppMessageLayout,
-	InAppMessageStyle,
-	InAppMessageTextAlign,
-	InAppMessagingConfig,
-	InAppMessagingEvent,
```
cshfang, line 16: This is tricky. I don't think we can remove all of these. Most of these types are exposed to help customers build custom in-app UIs

packages/notifications/src/inAppMessaging/providers/pinpoint/apis/identifyUser.ts
```diff
@@ -1,11 +1,79 @@
...
+ *
+ * @example
+ * ```ts
+ * // Identify a user with Pinpoint
+ * await identifyUser({
+ *     userId,
+ *     userProfile: {
+ *         email: [userEmail]
+ *         customProperties: {
+ *             phoneNumber: ['555-555-5555'],
```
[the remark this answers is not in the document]
cshfang: CustomProperties are in an array I think. Email was a miss

packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/identifyUser.test.ts
```diff
@@ -1,6 +1,86 @@
...
+	CATEGORY,
+} from '../../../../../src/inAppMessaging/providers/pinpoint/utils';
+import { updateEndpoint } from '@aws-amplify/core/internals/providers/pinpoint';
+
+import { IdentifyUserInput } from '../../../../../src/inAppMessaging/providers/pinpoint/types';
+
+jest.mock('@aws-amplify/core/internals/providers/pinpoint');
+jest.mock('../../../../../src/inAppMessaging/providers/pinpoint/utils');
+
+describe('Analytics Pinpoint Provider API: identifyUser', () => {
```
cshfang, line 18: Nit: In app

packages/notifications/src/inAppMessaging/types/config.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { AWSPinpointProviderConfig } from '../../common/AWSPinpointProviderCommon/types';
+
+export interface InAppMessagingConfig {
+	listenForAnalyticsEvents?: boolean;
+	AWSPinpoint?: AWSPinpointProviderConfig;
```
cshfang, line 7: I think we're supposed to be removing the `AWS` prefix right?

packages/notifications/src/inAppMessaging/types/config.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { AWSPinpointProviderConfig } from '../../common/AWSPinpointProviderCommon/types';
```
cshfang, line 3: Same as below, I think we've removed the `AWS` prefix from types (see core)

packages/notifications/src/inAppMessaging/types/event.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export enum InAppMessageInteractionEvent {
```
cshfang, line 4: Are we switching these over to string union? I.e.
```
export type InAppMessageInteractionEvent = 'MessageReceived' | 'MessageDisplayed'
```

packages/notifications/src/inAppMessaging/index.ts
Samaritan1011001: Yeah, I'll add them back in its respective PR. Wanted to clean up a bit and keep only types for APIs done so far and the types used in the common folder for now.

packages/notifications/src/inAppMessaging/providers/pinpoint/apis/identifyUser.ts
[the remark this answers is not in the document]
Samaritan1011001: This format is correct. customProperties' type is `Record<string, string[]>` So it takes a Record with the value for a key being an array.

jimblanc (approved):
LGTM

packages/core/src/Platform/types.ts
```diff
@@ -96,7 +96,8 @@ export enum GeoAction {
 	None = '0',
 }
 export enum InAppMessagingAction {
-	None = '0',
+	SyncMessages = '1',
+	UpdateEndpoint = '2',
```
cshfang, line 100: I think this and the AnalyticsAction should both be `IdentifyUser`, not `UpdateEndpoint`

packages/notifications/src/common/AWSPinpointProviderCommon/types.ts
```diff
@@ -3,12 +3,12 @@
...
 import { UserInfo } from '../../types';
 
-export interface AWSPinpointProviderConfig {
+export interface PinpointProviderConfig {
 	appId: string;
 	region: string;
 }
 
-export interface AWSPinpointUserInfo extends UserInfo {
+export interface PinpointUserInfo extends UserInfo {
```
cshfang, line 11: Sorry didn't notice this earlier but I think this should be using the same pattern as Analytics, no?
1. Using `UserProfile` instead of `UserInfo`
2. Using `serviceOptions` instead of extending the base type?

It's ok too if the intent is to do this incrementally

packages/core/src/Platform/types.ts
jimblanc: Ah good catch

Samaritan1011001: Thought it was the underlying operation but that would mean SyncMessages is getInAppMessages but that doesn't make sense. So yeah, will update this and Analytics to IdentifyUser

packages/notifications/src/common/AWSPinpointProviderCommon/types.ts
Samaritan1011001: Good catch. This type isn't used in identifyUser as of now so I'll update incrementally. 
But I wasn't using the address and optOut from Pinpoint options so updated that. Updated the test too.

cshfang (approved):

cshfang (approved):

jimblanc (approved):

PR #12226, opened 2023-10-06
fix(datastore): unexpected nulls in creates on undefined fields

svidgen:
#### Description of changes

Fixes an issue where DataStore would send explicit `{field: null}` values in graphql `create` operations when models were initialized with explicit `{field: undefined}`. E.g. in a model where `description` is optional, a DataStore save like this,

```js
await DataStore.save(new Model({
  title: 'some title',
  description: undefined
}));
```

Would result in a graphql create like this:

```json
{
  id: "123...",
  title: "some title",
  description: null
}
```

AppSync then rejects this operation.

This PR updates DataStore to ignore `undefined` fields in updates, regardless of whether they're implicit (omitted from the `new Model()` initialization) or explicit as above. The new graphql operation will attempt to save this:

```json
{
  id: "123...",
  title: "some title"
}
```

1. Repro'd manually
1. Added a test that replicates the manual repro
1. Made the test pass
1. Validated manually (pending)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/connectivityHandling.test.ts
+++ b/packages/datastore/__tests__/connectivityHandling.test.ts
@@ -176,9 +176,12 @@ describe('DataStore sync engine', () => {
 			}
 		});
 
+		// TODO: Need to revisit this and determine why it's necessary. In general, AppSync does
+		// NOT appear to accept `null` fields on create. Need to understand in what cases it accepts
+		// them and ensure DataStore field omission logic follows suit.
 		test('includes explicit null fields from mutation on create', async () => {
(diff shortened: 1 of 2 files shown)


iartemiev (approved):

stocaaro (approved):

david-mcafee (approved):

svidgen:
Turning this back into a draft for now to prevent merging. I think we need to explore whether these `null` values are OK (and correct) to send for other fields in case the explicit nulls are being caught and used in custom handlers.

PR #12264, opened 2023-10-11
chore: Clean up Cache to use share code and use native extension.

cshfang:
#### Description of changes

This PR updates the `Cache` util:
1. No longer using custom `reactnative.ts` file as entry point
2. Aligned on exporting a singular Cache regardless of platform
3. Removes unused implementations (E.g. in memory cache)
4. Move common functionality into super class instead of duplicating logic

* `yarn test`
* Tested Analytics (which uses Cache) against web, iOS and Android

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


5d (dismissed):

packages/core/src/Cache/StorageCacheCommon.ts
```diff
@@ -0,0 +1,581 @@
...
+	 */
+	constructor({
+		config,
+		keyValueStorage,
+	}: {
+		config?: CacheConfig;
+		keyValueStorage: KeyValueStorageInterface;
+	}) {
+		// const globalCacheConfig = Amplify.getConfig().Cache ?? {};
+		const globalCacheConfig = {};
```
jimblanc, line 34: Could you add this back & re-enable the cache config in the singleton types? Logic should work the same

jimblanc (commented):
Thank you!!

packages/core/src/Cache/StorageCacheCommon.ts
cshfang: Sure 👍

jimblanc (approved):

5d (approved):

jimblanc (approved):

5d (approved):

PR #11678, opened 2023-07-25
feat(api): add basic model types

iartemiev:
#### Description of changes

Simplified version of https://github.com/aws-amplify/amplify-js/pull/11671 without template literal types.

* Adds new V6 client types.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #12275, opened 2023-10-12
feat: add `generateClient().models` property

svidgen:
#### Description of changes

Copied existing implementation changing as little as possible at this stage:

1. https://github.com/aws-amplify/amplify-js/pull/11678
2. https://github.com/aws-amplify/amplify-js/pull/11715
3. https://github.com/aws-amplify/amplify-js/commit/fd73ef77567916a17aa2f03da2c225c69f12aade
4. https://github.com/aws-amplify/amplify-js/commit/e69c25f65dc88242229ce67614ca837d48c52cbf
5. https://github.com/aws-amplify/amplify-js/commit/a568d17235b1dd91cee6465b4ba375cc79a27301
6. https://github.com/aws-amplify/amplify-js/commit/7cf9eb287ada0585bb81399ae72192c91fecb69c

I'm unable to quickly find the relevant PR's for all of this. The copied-in code appears to be largely "POC" code. I'd like to review and refactor independently of this PR, which is intended to "rebase" those existing changes onto the v6 lineage (with a little added testing for my own confidence).

Added failing tests. Made them pass.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/src/internals/generateModelsProperty.ts
```diff
@@ -0,0 +1,124 @@
+import { ModelTypes } from '@aws-amplify/amplify-api-next-types-alpha';
```
svidgen, line 1: Copy-paste. Changes I made will be noted below.

Intending to address as many `any` types as I can in followups over the next couple sprints.

packages/api-graphql/src/internals/generateModelsProperty.ts
```diff
@@ -0,0 +1,124 @@
...
+	client: V6Client,
+	params: ClientGenerationParams
+): ModelTypes<T> {
+	const models = {} as any;
+	const config = params.amplify.getConfig();
+
+	const modelIntrospection = config.API?.GraphQL?.modelIntrospection;
+	if (!modelIntrospection) {
+		return {} as any;
+	}
```
svidgen, line 22: Was previously a simple dereference, assuming `modelInstrospection` as a top-level config value. This is no longer the case. And, if `modelInstrospection` is missing, there's nothing to do.

packages/api-graphql/src/internals/generateModelsProperty.ts
```diff
@@ -0,0 +1,124 @@
...
+							args
+						);
+						const variables = buildGraphQLVariables(
+							model,
+							'LIST',
+							args,
+							modelIntrospection
+						);
+
+						const res = (await client.graphql({
```
svidgen, line 50: From `this.graphql` to `client.graphql`, where `client` is the result of `generateClient()`.

packages/api-graphql/src/internals/generateModelsProperty.ts
```diff
@@ -0,0 +1,124 @@
...
+							operation
+						);
+						const variables = buildGraphQLVariables(
+							model,
+							operation,
+							arg,
+							modelIntrospection
+						);
+
+						const res = (await client.graphql({
```
svidgen, line 96: From `this.graphql` to `client.graphql`, where `client` is the result of `generateClient()`.

packages/api-graphql/src/internals/generateModelsProperty.ts
```diff
@@ -0,0 +1,124 @@
...
+	buildGraphQLVariables,
+	graphQLOperationsInfo,
+	ModelOperation,
+} from './APIClient';
+import { ClientGenerationParams } from './types';
+import { V6Client } from '../types';
+
+export function generateModelsProperty<T extends Record<any, any> = never>(
+	client: V6Client,
+	params: ClientGenerationParams
```
svidgen, line 14: Params changed, including addition of `client` param, which is the result of `generateClient()` that is bound to a specific `Amplify` scope.

packages/api-graphql/src/internals/generateClient.ts
```diff
@@ -17,11 +15,18 @@ type ClientGenerationParams = {
  * @param params
  * @returns
  */
-export function generateClient(params: ClientGenerationParams): V6Client {
-	return {
+export function generateClient<T extends Record<any, any> = never>(
```
svidgen, line 18: The types here are goofy as all hell.

It's on my mental list to figure out what the devil is going on here. But, the type system neither likes it when I omit `models` nor when I *include* `models`, even when I include models `as ModelType<T>`, which is what `.models` should be.

I don't feel like it's worth the time dig on this quite yet. But, *soon* ...

packages/core/__tests__/singleton/Singleton.test.ts
```diff
@@ -83,7 +160,18 @@ describe('Amplify.configure() and Amplify.getConfig()', () => {
...
+			API: {
+				GraphQL: {
+					apiKey: 'some-key',
+					defaultAuthMode: 'userPool',
+					endpoint: 'https://some.domain.com/graphql',
+					region: 'us-west-1',
+					modelIntrospection,
+				},
+			},
+		});
```
svidgen, line 174: I didn't realize going into this that `Amplify.configure()` did partial config changes the way it does, but it makes sense.

My concern with these config tests *in general* is the temporal coupling. So, just a note for the JS team, I guess: There's temporal coupling in these tests due to shared state, and I'm not sure if this is intended. But, it's a little unexpected and has the potential to be a minor PITA in the future.

packages/api-graphql/package.json
```diff
@@ -52,9 +54,10 @@
     "@aws-amplify/api-rest": "4.0.0",
     "@aws-amplify/auth": "6.0.0",
     "@aws-amplify/core": "6.0.0",
+    "@aws-amplify/amplify-api-next-types-alpha": "^0.0.5",
```
iartemiev, line 57: I think this can be a dev dep as well, it's a pure types package

packages/api-graphql/src/internals/generateClient.ts
iartemiev: `ModelType<T>` is the schema builder type, not the client-facing return type of `ClientSchema`. There is no `models` property in the client types. 
It's essentially a flattened POJO of return types. I think `Record<string, unknown>` should do it.

Here's an example client schema type:
```
type Schema = {
    Post: {
        title: string;
        tags: () => Promise<({
            id?: string | undefined;
            post: () => Promise<({
                title: string;
                description: string | null;
                tags: () => Promise<({
                    ...
                } | null | undefined)[] | null | undefined>;
                readonly id: string;
                readonly createdAt: string;
                readonly updatedAt: string;
            } | null | undefined)[] | null | undefined>;
            tag: () => Promise<({
                label: string;
                post: ...;
                readonly id: string;
                readonly createdAt: string;
                readonly updatedAt: string;
            } | null | undefined)[] | null | undefined>;
        } | null | undefined)[] | null | undefined>;
        readonly id: string;
        readonly createdAt: string;
        readonly updatedAt: string;
        description?: string | ... 1 more ... | undefined;
    };
    [__modelMeta__]: {
        ...;
    };
}
```

packages/api-graphql/src/internals/generateModelsProperty.ts
```diff
@@ -0,0 +1,126 @@
...
+	const models = {} as any;
+	const config = params.amplify.getConfig();
+
+	const modelIntrospection = config.API?.GraphQL?.modelIntrospection;
+	if (!modelIntrospection) {
+		return {} as any;
+	}
+
+	// TODO: refactor this to use separate methods for each CRUDL.
+	// Doesn't make sense to gen the methods dynamically given the different args and return values
```
iartemiev, line 27: Despite what I said in this comment, this actually will need to be dynamic when we support custom queries/mutations. I.e., those queries and mutations will be additional methods we expose on the model

packages/api-graphql/src/internals/generateClient.ts
iartemiev: However, `ModelTypes` (plural) IS the type name for the return value of `client.models`. 
We should probably rename it to avoid confusion. 

`generateClient` should accept a type param `Record<string, unknown>` but it returns:
```ts
{
  graphql: typeof v6graphql,
  models: ModelTypes<T>
}
```

iartemiev (approved):

packages/api-graphql/src/internals/generateClient.ts
svidgen: I meant `ModelTypes<T>` (plural), from here:

https://github.com/aws-amplify/amplify-js/blob/api-v6-models/packages/api/src/API.ts#L195

PR #12280, opened 2023-10-12
fix(adapter-nextjs): align with js-cookie on the client handling double encoded cookie names

HuiSF:
#### Description of changes

As of today, `signIn` is handled by the client side API, which uses js-cookie to operate the underlying cookie storage in a SSR use case (i.e. `Amplify.configure(config, {ssr: true})`).

According to [js-cookie document](https://github.com/js-cookie/js-cookie#encoding), it encodes cookie name.

In Amplify, we encode cookie name in the case of using `username` as a part of the cookie names, where the `username` can be an email address, for example, cookie name `test@email.com` will be encoded as `test%40email.com`.

When store the item into cookie store via js-cookie, the cookie name gets double encoded as `test%2540email.com`. 

In adapte-nextjs, we need to handle this ensure the sever side functionality can extra tokens from cookies. 

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/adapter-nextjs/tsconfig.json
+++ b/packages/adapter-nextjs/tsconfig.json
@@ -9,7 +9,7 @@
 		"sourceMap": true,
 		"module": "commonjs",
 		"moduleResolution": "node",
-		"allowJs": false,
+		"allowJs": true,
 		"declaration": true,
 		"typeRoots": ["./node_modules/@types", "../../node_modules/@types"],
 		"types": ["node"],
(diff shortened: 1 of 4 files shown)


packages/adapter-nextjs/tsconfig.json
```diff
@@ -9,7 +9,7 @@
 		"sourceMap": true,
 		"module": "commonjs",
 		"moduleResolution": "node",
-		"allowJs": false,
+		"allowJs": true,
```
HuiSF, line 12: ts-jest spitting tons of warning when setting it false as it transpiles everything into js I presume

cshfang (approved):

jimblanc (approved):

PR #12283, opened 2023-10-12
fix(storage): replace BufferSource with subtypes for types to get picked up

kvramyasri7:
#### Description of changes

- replace BufferSource with subtypes for types to get picked up
Before

image
image

After

image
image

- remove file type from `StorageUploadDataPayload`. This change is missed  sent previous PR to `next/release`

Build library with verdaccio, install in RN sample app and see the types

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/storage/src/types/inputs.ts
```diff
@@ -49,4 +49,4 @@ export type StorageCopyInput<
 /**
  * The data payload type for upload operation.
  */
-export type StorageUploadDataPayload = Blob | BufferSource | string | File;
+export type StorageUploadDataPayload = Blob | BufferSource | string;
```
AllanZhengYP, line 52: I think I found a more organic solution to make `File` type available according to the runtime:

```suggestion
export type StorageUploadDataPayload =
	| Blob
	| ArrayBufferView
	| ArrayBuffer
	| string
	| OptionalFile;

/**
 * File type is not available in environments like React Native. This type
 * returns File type only when it's defined in global.
 */
type OptionalFile = unknown extends File ? never : File;
```

The `BufferSource` is replaced by `ArrayBufferView | ArrayBuffer` explicitly and they are equivalent.

AllanZhengYP: In browser:
Screenshot 2023-10-12 at 3 28 11 PM
In RN:
Screenshot 2023-10-12 at 3 28 19 PM

kvramyasri7: I think for removing `File` we are creating new api this change is to support `uploadFile` api. What do you think ?

kvramyasri7: Thanks for the suggestion I will change the types and test in sample app.

AllanZhengYP (approved):
It looks good to me!

israx (approved):
lgtm

HuiSF (approved):

PR #12284, opened 2023-10-12
fix(auth): returns refreshToken from login 

israx:
#### Description of changes

The `refreshToken` originated from login was cleared after `refreshAuthTokens` was triggered, and any subsequent call would end up in an `AuthError`.

This PR returns the original `refreshToken`, that was issued during sign-in, when calling `refreshAuthTokens`.

- added a unit test to check the same refreshToken is return
- test in a sample app that multiple calls to `fetchAuthSession({forceRefresh:true})` would not result in an AuthError

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/auth/src/providers/cognito/utils/refreshAuthTokens.ts
+++ b/packages/auth/src/providers/cognito/utils/refreshAuthTokens.ts
@@ -63,13 +63,12 @@ export const refreshAuthTokens: TokenRefresher = async ({
 		});
 	}
 	const clockDrift = iat * 1000 - new Date().getTime();
-	const refreshToken = AuthenticationResult?.RefreshToken ?? refreshTokenString;
 
 	return {
 		accessToken,
 		idToken,
 		clockDrift,
-		refreshToken,
+		refreshToken: refreshTokenString,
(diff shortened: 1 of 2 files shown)


packages/auth/src/providers/cognito/utils/refreshAuthTokens.ts
```diff
@@ -63,7 +63,7 @@ export const refreshAuthTokens: TokenRefresher = async ({
 		});
 	}
 	const clockDrift = iat * 1000 - new Date().getTime();
-	const refreshToken = AuthenticationResult?.RefreshToken;
+	const refreshToken = AuthenticationResult?.RefreshToken ?? refreshTokenString;
```
israx, line 66: cognito will not return a refreshToken when refreshing id and access tokens.

elorzafe: maybe this could be `refreshTokenString` instead

elorzafe (approved):

HuiSF (approved):

PR #12285, opened 2023-10-12
feat(interactions): adapt v6 api design

5d:
#### Description of changes

- export new public packages for Interactions category
  - `@aws-amplify/interactions/lex-v1`
  - `@aws-amplify/interactions/lex-v2` == (`@aws-amplify/interactions`)
  
- packages will export `Interactions` object with 2 public APIs:
  - `send(botName: string, message: string | InteractionsMessage): Promise<InteractionsResponse>`
  - `onComplete(botname: string, callback: CompletionCallback): void`

tested with local react App

```typescript
import './App.css';
import { Interactions as InteractionsLexV1 } from '@aws-amplify/interactions/lex-v1'
import { Interactions } from '@aws-amplify/interactions';
import { Amplify } from 'aws-amplify';
import awsExports from './aws-exports'

// lexV1 bots added used CLI
Amplify.configure(awsExports)
// lexV2 bots need to be manually configured
// optionally lexV1 bots can also be configured manually 
Amplify.configure({
  Interactions: {
    LexV2: { // V2 bot
      BookTrip: {
        aliasId: '<botAliasId>',
        botId: '<botId>',
        localeId: 'en_US',
        region: 'us-east-1'
      }
    },

    LexV1: { // V1 bot
      'BookTrip_dev': {
        alias: '$LATEST',
        region: 'us-east-1'
      }
    }
  }
})

function App() {
  return (
    <div className="App">
      <button onClick={() => InteractionsLexV1.send('BookTrip_dev', `book a car`)}>Send LexV1 Message</button>
      <button onClick={() => Interactions.send('BookTrip', 'This is a test')}>Send LexV2 Message</button>
    </div>
  );
}

export default App;
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/interactions/src/types/index.ts
```diff
@@ -1,7 +1,4 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 export * from './Interactions';
```
ashwinkumar6, line 3: should we change this to named exports in places we `export *`
can also export only api, input and output types

packages/core/src/parseAWSExports.ts
```diff
@@ -84,6 +86,12 @@ export const parseAWSExports = (
 		}
 	}
 
+	if (Array.isArray(aws_bots_config)) {
```
ashwinkumar6, line 89:
```suggestion
  // Interactions
	if (Array.isArray(aws_bots_config)) {
```

packages/interactions/src/lex-v1/apis/onComplete.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { CompletionCallback } from '../../types';
+import { resolveConfig } from '../utils';
+import { lexProvider } from '../AWSLexProvider';
+
+export const onComplete = (
```
ashwinkumar6, line 8: Should we update all APIs to named params

packages/interactions/src/lex-v1/apis/sendMessage.ts
```diff
@@ -0,0 +1,17 @@
...
+import { resolveConfig } from '../utils';
+import { lexProvider } from '../AWSLexProvider';
+
+export const sendMessage = (
+	botname: string,
+	message: string | InteractionsMessage
+): Promise<InteractionsResponse> => {
+	const botConfig = resolveConfig(botname);
+	if (!botConfig) {
+		return Promise.reject(`Bot ${botname} does not exist.`);
```
ashwinkumar6, line 14: For all APIs, should we add separate error handling ([reference](https://github.com/aws-amplify/amplify-js/blob/550cc9dc27c0909bf73618d94b1b5264aa02404c/packages/storage/src/errors/utils/assertValidationError.ts#L10), [usage](https://github.com/aws-amplify/amplify-js/blob/550cc9dc27c0909bf73618d94b1b5264aa02404c/packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts#L46-L50))

packages/interactions/src/lex-v1/AWSLexProvider.ts
```diff
@@ -0,0 +1,165 @@
...
+		// check if credentials are present
+		let session;
+		try {
+			session = await fetchAuthSession();
+		} catch (error) {
+			return Promise.reject('No credentials');
+		}
+
+		const { name, region, alias } = botConfig;
+		const client = new LexRuntimeServiceClient({
```
ashwinkumar6, line 83: Should we try to maintain a single instance

packages/interactions/src/types/index.ts
5d: will update the exports.

packages/core/src/parseAWSExports.ts
5d: good catch!

packages/interactions/src/lex-v1/apis/onComplete.ts
5d: The current implementation is exporting the `sendMessage` and `onCompelete` with thier existing function signature in `InteractionsProvider` interface. I think the migration for user is remove the provider instance and import the function directly without anyother changes. Changing the API signature would involve more changes.

packages/interactions/src/lex-v1/AWSLexProvider.ts
5d: The current implementation of `sendMessage` and `onComplete` is more like a static function (although it uses the callbacks). We want to keep this logic as much as possible to not involve pentest or security reviews.

ashwinkumar6: got it,
thanks makes sense

packages/core/src/singleton/Interactions/types.ts
```diff
@@ -0,0 +1,25 @@
...
+type InteractionsLexV1Config = {
+	LexV1?: Record<string, LexV1BotConfig>;
+};
+
+type InteractionsLexV2Config = {
+	LexV2?: Record<string, LexV2BotConfig>;
+};
+
+export type InteractionsConfig = InteractionsLexV1Config &
+	InteractionsLexV2Config;
```
ashwinkumar6, line 25: I think `InteractionsConfig` would also allow `{}`. How about something like this. 
```suggestion
type InteractionsLexV1Config = {
	LexV1: Record<string, LexV1BotConfig>;
};

type InteractionsLexV2Config = {
	LexV2: Record<string, LexV2BotConfig>;
};

export type InteractionsConfig = 
  | InteractionsLexV1Config
	| InteractionsLexV2Config
	| InteractionsLexV1Config & InteractionsLexV2Config
```

packages/core/src/parseAWSExports.ts
```diff
@@ -84,6 +86,13 @@ export const parseAWSExports = (
 		}
 	}
 
+	// Interactions
+	if (Array.isArray(aws_bots_config)) {
+		amplifyConfig.Interactions = {
+			LexV1: Object.fromEntries(aws_bots_config.map(bot => [bot.name, bot])),
```
ashwinkumar6, line 92: QQ
Would the resultant object also have the "name" attribute.

packages/interactions/__tests__/lex-v1/apis/onComplete.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+	afterEach(() => {
+		mockLexProvider.mockReset();
+		mockResolveBotConfig.mockReset();
+	});
+
+	it('invokes provider onComplete API', () => {
+		const message = uuid();
+		const mockCallback = jest.fn();
+		onComplete(v1BotConfig.name, mockCallback);
```
ashwinkumar6, line 32: should we also check if the attached mockCallback is fired upon task completion

packages/interactions/__tests__/lex-v1/apis/sendMessage.test.ts
```diff
@@ -0,0 +1,42 @@
...
+	});
+
+	afterEach(() => {
+		mockLexProvider.mockReset();
+		mockResolveBotConfig.mockReset();
+	});
+
+	it('invokes provider sendMessage API', async () => {
+		const message = uuid();
+		await sendMessage(v1BotConfig.name, message);
```
ashwinkumar6, line 31: I think there are three ways the user can interact with the bot
1. string (text)
2. object (text)
3. object (voice)


Should we add unit tests for these usecases for lexV1 and lexV2

packages/interactions/__tests__/lex-v1/apis/onComplete.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import { resolveBotConfig } from '../../../src/lex-v1/utils';
+import { InteractionsError } from '../../../src/errors/InteractionsError';
+
+jest.mock('../../../src/lex-v1/AWSLexProvider');
+jest.mock('../../../src/lex-v1/utils');
+
+describe('Interactions LexV1 API: onComplete', () => {
+	const v1BotConfig = generateRandomLexV1Config();
+
+	const mockLexProvider = lexProvider.onComplete as jest.Mock;
```
ashwinkumar6, line 17: I think this mocks the provider class's onComplete method. Should we mock the underlying client that the provider class calls ([reference](https://github.com/aws-amplify/amplify-js/blob/7762f1a7076e622ec354c24539a3b57ce3ec4290/packages/interactions/__tests__/providers/AWSLexV2Provider-unit-test.ts#L195-L210)). That way our unit tests would cover the provider class's implementation as well.

packages/interactions/__tests__/lex-v1/apis/onComplete.test.ts
```diff
@@ -0,0 +1,43 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { v4 as uuid } from 'uuid';
+import { lexProvider } from '../../../src/lex-v1/AWSLexProvider';
+import { onComplete } from '../../../src/lex-v1/apis';
```
ashwinkumar6, line 6: I think the DX customers will be following is 
```
// V1
import { Interactions } from '@aws-amplify/interactions';
Interactions.send('BookTrip', 'This is a test')

// V2
import { Interactions as InteractionsLexV1 } from '@aws-amplify/interactions/lex-v1'
InteractionsLexV1.send('BookTrip_dev', `book a car`)
```
Would it add value if unit tests also follow this pattern

packages/interactions/__tests__/lex-v1/utils/resolveBotConfig.test.ts
```diff
@@ -0,0 +1,78 @@
...
+	});
+
+	it('return first match bot', () => {
+		const v1BotConfigs = [...Array(5)].map(generateRandomLexV1Config);
+		const v2BotConfigs = [...Array(5)].map(generateRandomLexV2Config);
+		getConfigSpy.mockReturnValue({
+			Interactions: {
+				LexV1: {
+					...Object.fromEntries(v1BotConfigs.map(bot => [bot.name, bot])),
+					[v1BotConfigs[3].name]: { ...v1BotConfigs[3], alias: 'test' },
```
ashwinkumar6, line 68: QQ
Are we testing multiple bots with the same name here. Since it's a dictionary can we have mutiple bots with the same name.

packages/interactions/src/lex-v2/apis/sendMessage.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { InteractionsMessage, InteractionsResponse } from '../../types';
+import { lexProvider } from '../AWSLexV2Provider';
+import { resolveBotConfig } from '../utils';
+import {
+	assertValidationError,
+	InteractionsValidationErrorCode,
+} from '../../errors';
+
+export const sendMessage = async (
```
ashwinkumar6, line 12: Since the DX customers will be using is `Interactions.send`, should we rename these to send in V1 and V2

packages/interactions/src/utils/index.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export * from './utils';
```
ashwinkumar6, line 4: should we use named exports here as well

ashwinkumar6:
I think for umbrella packages, we add providers in the following path `packages/<category>/src/providers/<provider_name>`. Should we follow something similar.

packages/interactions/src/lex-v2/apis/onComplete.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ashwinkumar6, line 1: I think for umbrella packages, we add providers in the following path `packages/<category>/src/providers/<provider_name>`. Should we follow something similar.

packages/interactions/src/index.ts
```diff
@@ -1,8 +1,5 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-export { Interactions } from './Interactions';
 export * from './types';
```
ashwinkumar6, line 4: I think this might expose all the types. We can use named export and only expose requied input/output types

ashwinkumar6:
Trying to understand how we're exposing these APIs under the mentioned DX in the PR desc. 
Are we actually exposing functional APIs, but as an object as shown below. When users import it, it creates an impress that a class is imported based on the DX. 
This is really interesting 
```
import { sendMessage, onComplete } from './apis';
interface IInteractions {
	send(
		botName: string,
		message: string | InteractionsMessage
	): Promise<InteractionsResponse>;
	onComplete(botName: string, callback: CompletionCallback): void;
}

export const Interactions: IInteractions = {
	send: sendMessage,
	onComplete,
};
```

ashwinkumar6:
Thanks a lot for incorporating all the feedback @5d. This looks great !!!!!!

packages/interactions/src/lex-v1/index.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ashwinkumar6, line 1: Trying to understand how we're exposing these APIs under the mentioned DX in the PR desc. 
Are we actually exposing functional APIs, but via an object as shown below. When users import it, it creates an impress that a class is imported based on the DX.
This is really interesting !!!!
```
import { sendMessage, onComplete } from './apis';
interface IInteractions {
	send(
		botName: string,
		message: string | InteractionsMessage
	): Promise<InteractionsResponse>;
	onComplete(botName: string, callback: CompletionCallback): void;
}

export const Interactions: IInteractions = {
	send: sendMessage,
	onComplete,
};
```

ashwinkumar6:
Thanks a lot for incorporating all the feedback @5d. This looks great !!!!!!

packages/core/src/singleton/Interactions/types.ts
5d: For typescript intersection type, my understanding is 
```
{A?} & {B?} ≅ ({} | {A}) & ({} | {B}) ≅ {} | {A} | {B} | {A, B} 
```
it already contains all the cases you mentioned above.

packages/core/src/parseAWSExports.ts
5d: yes, it will contains the name attribute.

packages/interactions/__tests__/lex-v1/apis/onComplete.test.ts
5d: Invoking the callback logic is part of provider's. I think it has been tested in provider unit test cases. The unit test for public API `onComplete` is just testing whether it invokes the underlying provider's method.

packages/interactions/__tests__/lex-v1/apis/sendMessage.test.ts
5d: The type signagure is restrict `message` to be `string | InteractionsMessage`. For behaviors on different message type, that's the logic in provider class. It should have been tested in provider's unit test cases.

packages/interactions/__tests__/lex-v1/apis/onComplete.test.ts
5d: The API `onComplete` is just a wrapping on provider's `onComplete` method. I think we only need to check the wrapping is invoking the correct provider method in unit test. We should have done the unit test on provider class to check the provider implementation.

packages/interactions/__tests__/lex-v1/apis/onComplete.test.ts
5d: The unit test is on the module of `onComplete`. It doesn't need to know `Interactions`. We could add unit test on `Interactions` module to check whether `send` is invoke the correct `sendMessage` method.

packages/interactions/__tests__/lex-v1/utils/resolveBotConfig.test.ts
5d: I was trying to test the order when we using Array to store the botConfig. As we changed to object, I think we don't need this unit test case anymore. I will remove it.

packages/interactions/src/index.ts
5d: yeah, I've changed the `types/index.ts` to only expose needed types. I will change the `index.ts` as well.

packages/interactions/src/lex-v2/apis/sendMessage.ts
5d: will update the name.

packages/interactions/src/utils/index.ts
5d: I think this folder is not for internal use. Should we also change to named exports?

packages/interactions/src/lex-v1/index.ts
5d: Yes, we just wrap the APIs into an object to mimic V5 API.

packages/interactions/src/utils/index.ts
ashwinkumar6: Thanks for updating it. I think will keeps things uniform

packages/interactions/src/lex-v2/apis/sendMessage.ts
ashwinkumar6: Thanks for updating

packages/interactions/src/lex-v2/apis/onComplete.ts
ashwinkumar6: I think adding a `provider` folder will keep it uniform across all packages

packages/interactions/__tests__/lex-v1/utils/resolveBotConfig.test.ts
ashwinkumar6: thanks

packages/interactions/__tests__/lex-v1/apis/onComplete.test.ts
ashwinkumar6: That's a good point.  Thanks for clarifying that,
lgtm

packages/interactions/__tests__/lex-v1/apis/onComplete.test.ts
ashwinkumar6: ahhh we're retaining existing provider tests as well. Got it thanks

packages/interactions/__tests__/lex-v1/apis/sendMessage.test.ts
ashwinkumar6: ahhh we're retaining existing provider tests as well. Got it thanks

packages/interactions/__tests__/lex-v1/apis/onComplete.test.ts
ashwinkumar6: ahhh we're retaining existing provider tests as well. Got it thanks

packages/core/src/singleton/Interactions/types.ts
ashwinkumar6: yep makes sense, but i'm wondering if we should avoid `{}`

5d: As we are doing merging at category configuration level, running `Amplify.configure({Interactions:{}})` will have no effect on the configuration.

packages/core/src/parseAWSExports.ts
5d: this is only for the legacy CLI config, it has no side effect as the data is confirming the interface shape (`LexV1BotConfig`). And user will not need to provide the `name` field.

ashwinkumar6 (dismissed):
Sorry missed the part where we're retaining existing lexV1 and lexV2 provider unit tests.
So for wrapper APIs, we can just test if the provider's method is called with correct params

lgtm 💯

ashwinkumar6 (approved):
lgtm

stocaaro (approved):

packages/core/src/parseAWSExports.ts
elorzafe: this is not a type error because is `aws_bots_config` is `any`?

packages/interactions/package.json
```diff
@@ -44,49 +70,59 @@
...
+		"uuid": "^9.0.0",
 		"tslib": "^2.5.0"
 	},
-	"peerDependencies": {
-		"@aws-amplify/core": "^6.0.0"
-	},
 	"devDependencies": {
-		"@aws-amplify/core": "6.0.0"
+		"@aws-amplify/core": "6.0.0",
+		"typescript": "^5.2.2"
```
elorzafe, line 88: Should we use the same version as the other packages?

elorzafe (approved):
Just have a minor comment, but looks good 

Thanks @5d 🎖️

packages/interactions/package.json
5d: yes, will update the typescript version to `5.0.2` in the follow up PR.

packages/core/src/parseAWSExports.ts
5d: sorry, I didn't notice this comment. The `aws_bots_config` is any here. I only checked whether it's an array. As this is part of old aws-exports config, I assume it has good shape. But we can add the data shape validation check if needed.

PR #12299, opened 2023-10-13
fix(auth): remove username en/decode logic 

israx:
#### Description of changes

Currently the library create auth keys to store tokens and uses the `username` as identifier. E.g 
`CognitoIdentityServiceProvider.${userPoolClientId}.${username}.idToken`

This username is encoded when setting tokens and decoded when getting tokens. This works fine in `LocalStorage`. The issue happens in Cookies as `js-cookie`(the package used by the library) will encode the keys and if username is an `email` it will double encode the `@` symbol.

This removes the encoding logic and attaches the plain username in the keys.

**NOTE:**

This change when published to `console-preview` will log out any users that were authenticated with the encoded key format.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
+++ b/packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
@@ -71,7 +71,7 @@ export class DefaultTokenStore implements AuthTokenStore {
 				refreshToken,
 				deviceMetadata: (await this.getDeviceMetadata()) ?? undefined,
 				clockDrift,
-				username: decodeURIComponent(await this.getLastAuthUser()),
+				username: await this.getLastAuthUser(),
 			};
 		} catch (err) {
 			return null;
@@ -81,8 +81,7 @@ export class DefaultTokenStore implements AuthTokenStore {
 		assert(tokens !== undefined, TokenProviderErrorCode.InvalidAuthTokens);
 		await this.clearTokens();
 
-		const lastAuthUser =
-			(tokens.username && encodeURIComponent(tokens.username)) ?? 'username';
+		const lastAuthUser = tokens.username;
 		await this.getKeyValueStorage().setItem(
 			this.getLastAuthUserKey(),
 			lastAuthUser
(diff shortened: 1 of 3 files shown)


elorzafe (approved):
Looks good,

Thanks @israx 🎖️

packages/auth/src/providers/cognito/utils/refreshAuthTokens.ts
```diff
@@ -69,6 +69,6 @@ export const refreshAuthTokens: TokenRefresher = async ({
 		idToken,
 		clockDrift,
 		refreshToken: refreshTokenString,
-		username: `${accessToken.payload.username}`,
```
HuiSF, line 72: Tested against the new backend setup, the auth keys are consistent with this fix 👍

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
```diff
@@ -81,8 +81,7 @@ export class DefaultTokenStore implements AuthTokenStore {
 		assert(tokens !== undefined, TokenProviderErrorCode.InvalidAuthTokens);
 		await this.clearTokens();
 
-		const lastAuthUser =
-			(tokens.username && encodeURIComponent(tokens.username)) ?? 'username';
```
ashwinkumar6, line 85: QQ,
Do we also need to update anythere here
1. [signInWithRedirect1](https://github.com/aws-amplify/amplify-js/blob/550cc9dc27c0909bf73618d94b1b5264aa02404c/packages/auth/src/providers/cognito/apis/signInWithRedirect.ts#L216-L218)
2. [signInWithRedirect2](https://github.com/aws-amplify/amplify-js/blob/550cc9dc27c0909bf73618d94b1b5264aa02404c/packages/auth/src/providers/cognito/apis/signInWithRedirect.ts#L270-L272)

HuiSF: +1, do we need to fallback to `username`, if `username` is missing from token payload, isn't considered as an error?

israx: I don't think so. The username stored in `LastAuthUser` will still be consistent. I'll give it a test to confirm

israx: tested signInWithRedirect and seems working fine on my end

israx: Just check, it does default to 'username'

HuiSF (approved):

PR #12335, opened 2023-10-17
fix: Make Reachability node safe

jimblanc:
#### Description of changes

Makes the `Reachability` utility node safe to allow it to be built on Next.js applications, even when not using SSR.

Previously this behavior was enforced by this check which was removed to avoid a dependency on `process`: https://github.com/aws-amplify/amplify-js/blob/main/packages/core/src/Util/Reachability.ts#L16

Tested `Reachability` in a Web application to make sure the observable was still firing by disconnecting/reconnecting my network. Associated DS integ test now passing.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/core/src/Reachability/Reachability.ts
+++ b/packages/core/src/Reachability/Reachability.ts
@@ -9,7 +9,7 @@ export class Reachability {
 	private static _observers: Array<CompletionObserver<NetworkStatus>> = [];
 
 	networkMonitor(_?: unknown): Observable<NetworkStatus> {
-		const globalObj = isWebWorker() ? self : window;
+		const globalObj = isWebWorker() ? self : typeof window !== 'undefined' && window;
 
 		if (!globalObj) {
 			return from([{ online: true }]);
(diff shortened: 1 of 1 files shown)


packages/core/src/Reachability/Reachability.ts
```diff
@@ -1,15 +1,19 @@
...
 
 export class Reachability {
 	private static _observers: Array<CompletionObserver<NetworkStatus>> = [];
 
 	networkMonitor(_?: unknown): Observable<NetworkStatus> {
-		const globalObj = isWebWorker() ? self : window;
+		const globalObj = isWebWorker() ? self : typeof window !== 'undefined' && window;
+
+		if (!globalObj) {
+			return from([{ online: true }]);
```
jimblanc, line 15: This behavior was previously covered by this check: https://github.com/aws-amplify/amplify-js/blob/main/packages/core/src/Util/Reachability.ts#L16

RN uses `Reachability.native.ts` so not a concern. Web Worker is explicitly covered. On Node.js (where window is not available) it will return `online: true` which is aligned with v5 behavior.

david-mcafee (approved):
LGTM! 🎉

PR #12340, opened 2023-10-18
chore: disable/delete SSR tests

israx:
#### Description of changes

**Disable test**:
- `auth-cdn`: Based on a previous discussion it is unclear whether we still want to support CDN or not. Thus, this PR momentarily disables it. If it will be supported, then we would need to generate the CDN v6 artifacts.

**Deleted test**
- `custom-auth-ssr`: it tests a basic auth flow in the client and the `getServerSideProps` functionality, which is already being tested.
- `integ_react_amazon_cognito_identity_js_cookie_storage` and `integ_react_amazon_cognito_identity_js`: it tests functionality in `amazon_cognito_identity_js` package, which is not supported in v6. However these tests also test some device and user management functionalities. So instead of refactoring them we should create a better test coverage for auth.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


stocaaro (approved):

elorzafe (approved):
Looks good!

PR #12364, opened 2023-10-19
fix(auth): assign `USERNAME` param returned from service to RespondToAuthChallenge calls

israx:
#### Description of changes

This PR fixes MFA and device tracking flows during the following conditions:
- userpool is configured to sign-in with `phone_number` or `email`
- userpool is configured to sign-in with an alias

Cognito will return a `userSub` or the corresponding `username` during sign-in when any of the stated configs above are met.

**IMPORTANT:**

Device tracking will be supported, however end-users won't be able to bypass MFA under the following conditions:
- userpool is configured with any of the options listed above
- `signIn` API is called with `USER_PASSWORD_AUTH` authFlow.

Customers will be able to call `listDevices` as `ConfirmDevice` will be kicked off. However bypassing MFA is not possible because `InitiateAuth` won't have access to a `DEVICE_KEY` when authFlow is `USER_PASSWORD_AUTH`.

Currently v5 throws an error during sign-in making it not possible to be `Authenticated`.

- Tested all AuthFlows in a sample App with Device tracking enabled

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


jimblanc (approved):
Makes sense to me, thanks for the fix

PR #12381, opened 2023-10-20
Chore/cleanup type versions

jimblanc:
#### Description of changes

Cleans up our `typesVersions` fields across the library and removes the catch-all `*` rule. Types from the package root are specified via the top-level `typings` field so the catch-all is not required. We still need to keep `typesVersions` however for some versions of React Native & TypeScript to ensure consistent behavior.

Tested imports & builds in a NextJS, React, and RN sample apps.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api/index.v37.d.ts
jimblanc, line 1: Confirmed with the Data team that we no longer require this.

stocaaro (approved):

ashwinkumar6 (approved):
Thank you !

cshfang (approved):

PR #12388, opened 2023-10-20
fix(core): give ampilfyUuid clear type info

5d:
#### Description of changes

Currently, the `amplifyUuid` is alias of dependency `uuid`'s `v4` method, which has [optional `options?: V4Options` parameter defined](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/uuid/index.d.ts#L31). 
image

As we are not exposing `V4Options` type, it will cause compile issue in user App. This PR defines `amplifyUuid` function signature without `V4Options` type.

- tested at local

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/core/src/utils/amplifyUuid/index.ts
```diff
@@ -4,6 +4,6 @@
 import './polyfill';
 import { v4 } from 'uuid';
 
-const amplifyUuid = v4;
+const amplifyUuid = (): string => v4();
```
stocaaro, line 7: Why not type this directly without the wrapper?
```
const amplifyUuid: () => string = v4;
```

HuiSF (dismissed):

packages/core/src/utils/amplifyUuid/index.ts
5d: I think it's the same thing. Defining it with annonymous function looks more explicit for me.

jimblanc (dismissed):

packages/core/src/utils/amplifyUuid/index.ts
stocaaro: I think I'll grant you that adding a function layer is more explicit since its actually adding executing code to the js runtime, where typing the assignment is only adding build time type checking that doesn't ship to the end user. They're the same behavior, but the typing version achieves the same outcome with a smaller bundle size and fewer cpu cycles, with the needed typesafety.

Maybe this is premature optimization, but I think we should prefer to add type guarantees without adding js size as a pattern for the library.

5d: @stocaaro The bundle size is a good point. I've tested the function assignment style reduced 5 bytes of the bundle size. I was not able to confirm the CPU cycles. However, I think the mordern interpreters should be able to inline the anonymous function, the CPU cost should be the same. I will update the PR, thanks for the comments! 👍

stocaaro (approved):

jimblanc (approved):

HuiSF (approved):

AllanZhengYP (approved):

PR #12415, opened 2023-10-24
fix(auth): handle multiple  redirect links

israx:
#### Description of changes

**Web:**
- it gets the URI that comes from the same origin, otherwise throws 

use cases supported by Amplify:
| Origin    | Destination |  Same origin|
| ----------- | ----------- | -----------|
| www.a.com     | www.a.com        | yes
| www.a.com     | www.a.com/app       | yes
| www.a.com/app  | www.a.com/another-app       | yes
| www.a.com/app  | dev.a.com/app       | no

Websites with different origins don't share elements in storage, thus Amplify is unable to access oauth keys.

**RN:**
- it gets the first URI that is not http or https

Test in a sample app with code and token responseType

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/auth/src/providers/cognito/apis/signOut.ts
```diff
@@ -123,10 +124,10 @@ async function oAuthSignOutRedirect(
 	preferPrivateSession: boolean
 ) {
 	assertOAuthConfig(authConfig);
-
+	const getRedirectSignOut = getSignInRedirect;
```
cshfang, line 127: Instead of doing this - maybe we can just change the utility function name to something more generic like `getRedirectUrl` and just use it directly in both APIs

packages/auth/src/providers/cognito/utils/oauth/getSignInRedirect.native.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { invalidRedirectException } from '../../../../errors/constants';
+
+/** @internal */
+export function getSignInRedirect(redirects: string[]): string {
+	const redirect = redirects.find(
+		redirect => !redirect.startsWith('http') && !redirect.startsWith('https')
```
cshfang, line 9: This would be redundant, I think. As you can't have a string that starts with `'https'` without it also starting with `'http'`.

But I think maybe we should fully qualify the search for protocols i.e.
```ts
const redirect = redirects.find(
		redirect => !redirect.startsWith('https://') && !redirect.startsWith('http://')
```
Since we don't actually want to filter out a scheme like `'httpjustkidding://'`

packages/auth/src/providers/cognito/utils/oauth/getSignInRedirect.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { invalidRedirectException } from '../../../../errors/constants';
+
+/** @internal */
+export function getSignInRedirect(redirects: string[]): string {
+	const redirect = redirects.find(redirect => {
+		return redirect.startsWith(String(window.location.origin + '/'));
```
cshfang, line 9: QQ: If there's no match with the current location.. does it necessarily invalidate the redirect url? Is it possible they *want* to redirect to a different url? I'm wondering if we should start with this and additionally fallback to a best guess rather than throw an error

elorzafe: I think that makes, sense what is flutter doing?

packages/auth/src/providers/cognito/utils/oauth/getSignInRedirect.native.ts
elorzafe: Good catch @cshfang, could there be an app name be `http` or `https`?

packages/auth/src/providers/cognito/utils/oauth/getSignInRedirect.ts
israx: Flutter is matching the current window location and also throws. I added a fallback case but I kept the `throw` logic in case nothing is found in the `signInRedirect` array config option. Otherwise it will return `undefined` and customers will still face an error while in the `oauth` flow.

israx: Thinking twice I think it is not a good idea to add a fallback because the `signInRedirect` should be coming from the same origin, v6 doesn't support oauth flows coming outside of Amplify, thus customers can have their `redirectSignIn` url with tokens hanging in the browser.

israx: added an additional fallback to match the same domain

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.ts
```diff
@@ -0,0 +1,32 @@
...
+	} else if (redirectUrlFromDifferentOrigin) {
+		throw invalidOriginException;
+	}
+	throw invalidRedirectException;
+}
+
+// origin + pathname => https://example.com/app
+const isSameOriginAndPathName = (redirect: string) =>
+	redirect.startsWith(
+		String(window.location.origin + window.location.pathname ?? '/')
```
cshfang, line 26: Nit: Could you use string interpolation to avoid the `String()` cast? More importantly - it might make the expression itself a little clearer (I *think* this means if there's no pathname then use `'/'`) and doing something like
```ts
`${window.location.origin}${window.location.pathname ?? '/'}`
```
might make that marginally clearer?

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
```diff
@@ -0,0 +1,16 @@
...
+export function getRedirectUrl(redirects: string[]): string {
+	const redirect = redirects?.find(
+		redirect =>
+			!redirect.startsWith('http://') && !redirect.startsWith('https://')
+	);
+	if (!redirect) {
+		throw invalidRedirectException;
+	}
+	return redirect;
+}
```
elorzafe, line 16: This will always use the first non `http` `https` url, this means if you have more than the rest will be ignored

elorzafe: Is it possible to detect the app name?

israx: yeah this is specific for RN, I don't think there is a way to distinguish more than 1 non http or non https. The conditions here are less restrictive

cshfang (approved):

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.ts
```diff
@@ -0,0 +1,32 @@
...
+	);
+// domain => outlook.live.com, github.com
+const isTheSameDomain = (redirect: string) =>
+	redirect.includes(String(window.location.hostname));
```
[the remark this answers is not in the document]
israx: Yeah I think if we want to be less restrictive, the next case basically wouldn't match an url coming from the same origin, thus not something that Amplify supports at the moment

elorzafe (approved):
Looks good

Thanks @israx 🏅

PR #12433, opened 2023-10-25
fix(analytics):  throw `Error` if `domElementId` type is not `string`

kvramyasri7:
#### Description of changes

This change removes type check `domElementId` in `autoTackMedia` util function from `record` api

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/analytics/src/providers/personalize/utils/autoTrackMedia.ts
```diff
@@ -187,7 +187,7 @@ export const autoTrackMedia = async (
 		return;
 	}
 
-	if (typeof domElementId === 'string' && !domElementId) {
+	if (!domElementId) {
```
5d, line 190: later at line#197 it cast domElementId type to string. will that work without checking the type infront?

kvramyasri7: I believe that should be working  as the type of `properties: Record<string, unknown>`. which has  I do not see any red underscore or TS complaints😅.  That is a record which is `{}` can be casted to a string.

5d: ok, just wondering if a user give the value with number `123`. When force type casting to string type, what could happen? I guess browser would just log an error in console. Not sure how react native would handle this.

kvramyasri7: Very Good Point. umm, RN I did not think about it. 😭 Any tip ?

kvramyasri7: I believe it does the same what the other logger messages do and just return from there isn't it?

packages/analytics/src/providers/personalize/utils/autoTrackMedia.ts
```diff
@@ -187,13 +188,21 @@ export const autoTrackMedia = async (
...
 		logger.debug(
 			"Missing domElementId field in 'properties' for MediaAutoTrack event type."
 		);
 		return;
 	}
 
+	if (typeof domElementId !== 'string') {
+		throw new AnalyticsError({
```
5d, line 199: For Analytics category, I think we only throw errors for config validation.

5d (changes requested):

packages/analytics/src/providers/personalize/utils/autoTrackMedia.ts
kvramyasri7: umm, but why is it? any specific reason. Let me go and re-change it from `AnalyticsError` to `AmplifyError`

5d: For Analytics, the APIs are designed in fire and forget style, which is returning void. The customer won't have chance to catch those errors. It's better to just log and return here.

kvramyasri7: Good to know. Thanks Di. Updated.

packages/analytics/src/providers/personalize/utils/autoTrackMedia.ts
```diff
@@ -187,13 +187,21 @@ export const autoTrackMedia = async (
...
-	if (typeof domElementId === 'string' && !domElementId) {
+	if (!domElementId) {
 		logger.debug(
 			"Missing domElementId field in 'properties' for MediaAutoTrack event type."
 		);
 		return;
 	}
 
+	if (typeof domElementId !== 'string') {
+		throw new AmplifyError({
```
5d, line 198: Could you give me an example code how customer could catch the error you thrown here?

kvramyasri7: Web Or RN ?

5d: Either of them, I just want to understand how throwing `AmplifyError` would help here for customer to do error handling.

PR #12439, opened 2023-10-26
feat(auth): add signInDetails to AuthUser interface

israx:
#### Description of changes

Adds signInDetails to the `AuthUser` interface. This change will provide information used during the sign-in flow, such as `loginId` and `authFlowType`

**DX:**
1. userpool configure to sign-in with either phone_number or email.
Screenshot 2023-10-26 at 9 36 48 AM

2. userpool configure to sign-in with alias
Screenshot 2023-10-26 at 9 38 03 AM

- unit test
- smoke test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/src/providers/cognito/apis/internal/getCurrentUser.ts
```diff
@@ -1,22 +1,38 @@
...
+	if (isTokensWithSignInDetails(tokens)) {
+		authUser.signInDetails = tokens.signInDetails;
+	}
+	return authUser;
 };
+
+function isTokensWithSignInDetails(
+	tokens: Record<string, unknown>
+): tokens is AuthTokens & { signInDetails: CognitoAuthSignInDetails } {
+	return !!tokens && !!(tokens as any).signInDetails;
```
elorzafe, line 37: Do you need to check what `signInDetails`?

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
```diff
@@ -65,14 +65,22 @@ export class DefaultTokenStore implements AuthTokenStore {
...
 				accessToken,
 				idToken,
 				refreshToken,
 				deviceMetadata: (await this.getDeviceMetadata()) ?? undefined,
 				clockDrift,
 				username: await this.getLastAuthUser(),
 			};
+
+			if (signInDetails) {
+				tokens.signInDetails = JSON.parse(signInDetails);
```
elorzafe, line 81: Do we need to check if this throws?

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
```diff
@@ -125,6 +133,12 @@ export class DefaultTokenStore implements AuthTokenStore {
 				tokens.deviceMetadata.randomPassword
 			);
 		}
+		if (!!tokens.signInDetails) {
+			await this.getKeyValueStorage().setItem(
+				authKeys.signInDetails,
+				JSON.stringify(tokens.signInDetails)
```
elorzafe, line 139: Do we need to check if this throws?

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
jimblanc: Looks like it's wrapped in a try-catch if that's adequate?

jimblanc (dismissed):
LGTM

packages/auth/src/providers/cognito/apis/internal/getCurrentUser.ts
cshfang, line 37: Nit: Could be more concise
```ts
	return !!(tokens as any)?.signInDetails;
```

packages/auth/src/providers/cognito/apis/internal/getCurrentUser.ts
```diff
@@ -1,22 +1,38 @@
...
 		username: username as string,
 		userId: sub as string,
 	};
+	if (isTokensWithSignInDetails(tokens)) {
+		authUser.signInDetails = tokens.signInDetails;
+	}
+	return authUser;
 };
+
+function isTokensWithSignInDetails(
```
cshfang, line 34: Nit: Does this need to be a type predicate? Is there any reason above to narrow type? I'm just wondering if this function can be something like `getSignInDetailsFromTokens` and just return either `CognitoAuthSignInDetails | undefined` - which would help consolidate some logic above

packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts
```diff
@@ -62,7 +63,9 @@ export class TokenOrchestrator implements AuthTokenOrchestrator {
 
 	async getTokens(
 		options?: FetchAuthSessionOptions
-	): Promise<AuthTokens | null> {
+	): Promise<
+		(AuthTokens & { signInDetails?: CognitoAuthSignInDetails }) | null
```
cshfang, line 67: Nit: I'm sure there was a reason for this but why can't `signInDetails` just be an optional property on `AuthTokens`?

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
elorzafe: if there is something that was modified I would handle that there. can you change that @israx ?

packages/auth/src/providers/cognito/types/outputs.ts
```diff
@@ -29,7 +29,7 @@ export type FetchUserAttributesOutput = AuthUserAttributes<UserAttributeKey>;
 /**
  * Output type for Cognito getCurrentUser API.
  */
-export type GetCurrentUserOutput = AuthUser;
```
cshfang, line 32: Nit: Is `AuthUser` still used?

cshfang (dismissed):
Had some questions/suggestions but I don't _think_ they're blocking

elorzafe (approved):
Looks good it seems weird that `signIndetails` gets modified externally

packages/auth/src/providers/cognito/types/outputs.ts
israx: It is used to extend the `CognitoAuthUser`

packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts
israx: So if customers are extending the TokenProvider interface, they don't have to add signInDetails

packages/auth/src/providers/cognito/apis/internal/getCurrentUser.ts
israx: I think for now, this should be good, thinking if in the feature signInDetails gets extended, then we would need to account for it too

israx: changed the logic to include @cshfang feedback, this code is not longer valid.

jimblanc (approved):

elorzafe (approved):

cshfang (approved):

PR #12481, opened 2023-10-31
fix(core): remove used symbol and remove usage of Symbol.for

HuiSF:
#### Description of changes

* Removed usage of `Symbol.for` as the symbol that is supposed to use internally can be overridden
* Removed unused the symbol constant

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/core/src/Hub/index.ts
+++ b/packages/core/src/Hub/index.ts
@@ -16,8 +16,8 @@ import {
 } from './types';
 
 export const AMPLIFY_SYMBOL = (
-	typeof Symbol !== 'undefined' && typeof Symbol.for === 'function'
-		? Symbol.for('amplify_default')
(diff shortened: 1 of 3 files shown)


packages/core/src/Hub/index.ts
```diff
@@ -17,7 +17,7 @@ import {
 
 export const AMPLIFY_SYMBOL = (
 	typeof Symbol !== 'undefined' && typeof Symbol.for === 'function'
```
jimblanc, line 19: Can we remove the `typeof Symbol.for` check as well

HuiSF: Good catch

jimblanc (approved):

erinleigh90 (approved):

cshfang (approved):

PR #12484, opened 2023-10-31
chore: add missing auto sign-in unit tests

israx:
#### Description of changes

add the following tests cases for auto sign-in:
- Auto sign-in should not resolve if a different user is confirmed
- Auto sign-in should not be interrupted if `confirmSignUp` fails
- `signUp` should return "COMPLETE_AUTO_SIGN_IN" step when `autoSignIn` is enabled with verification link.

yarn test 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/__tests__/providers/cognito/autoSignIn.test.ts
```diff
@@ -64,13 +66,157 @@ describe('Auto sign-in API Happy Path Cases:', () => {
...
+		signUpSpy.mockClear();
 	});
 
 	test('Auto sign-in should resolve to a signIn output', async () => {
 		const signInOutput = await autoSignIn();
 		expect(signInOutput).toEqual(authAPITestParams.signInResult());
 		expect(handleUserSRPAuthflowSpy).toBeCalledTimes(1);
 	});
+
+	test('Auto sign-in should not resolve if a different user is confirmed', async () => {
```
erinleigh90, line 78: I don't know if this is what is being tested here. It looks like we never test auto sign-in after the first user is confirmed.

erinleigh90 (changes requested):

PR #12497, opened 2023-11-02
fix(auth): vend the helpful types for custom providers

Samaritan1011001:
#### Description of changes

Vends the helpful types for constructing custom auth providers

yarn link

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


elorzafe (dismissed):

israx (dismissed):

elorzafe (approved):

stocaaro (approved):

jimblanc (approved):

packages/auth/src/index.ts
```diff
@@ -75,4 +75,13 @@ export {
 
 export { AuthError } from './errors/AuthError';
 
-export { fetchAuthSession } from '@aws-amplify/core';
+export {
+	fetchAuthSession,
+	decodeJWT,
+	AWSCredentialsAndIdentityIdProvider,
+	GetCredentialsOptions,
+	AWSCredentialsAndIdentityId,
```
jimblanc, line 83: Discussed offline, but should should rename these to not be AWS specific since they're intended for custom providers which may or may not use AWS

jimblanc (changes requested):

packages/auth/src/index.ts
elorzafe: I think this is valid because we are expecting AWSCredentials, we are not mentioning what underlying service is used to generate credentials, it could STS, Cognito Identity Pools or other

Samaritan1011001: Agreed. Pointed this out myself and making the change already.

elorzafe (approved):

jimblanc (approved):

PR #12519, opened 2023-11-06
fix: add module loader for web browser native module

cshfang:
#### Description of changes

This PR refactors the Amplify Web Browser native module loading to use the module loading pattern introduced in `@aws-amplify/react-native`. This was necessary given that Expo applications by default do not lazy load modules and was thusly throwing an error even when `signInWithRedirect` was not used by the application.

* `yarn test`
* Tested against EAS built Android and iOS apps - importing the native module should now only be required if executing `signInWithRedirect`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/react-native/src/moduleLoaders/loadAmplifyWebBrowser.ts
+++ b/packages/react-native/src/moduleLoaders/loadAmplifyWebBrowser.ts
@@ -1,0 +1,29 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import type { WebBrowserModule } from '@aws-amplify/rtn-web-browser';
+
+export const loadAmplifyWebBrowser = () => {
+	try {
+		// metro bundler requires static string for loading module.
+		// See: https://facebook.github.io/metro/docs/configuration/#dynamicdepsinpackages
+		const module = require('@aws-amplify/rtn-web-browser')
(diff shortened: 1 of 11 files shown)


packages/react-native/src/moduleLoaders/loadAmplifyWebBrowser.ts
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import type { WebBrowserModule } from '@aws-amplify/rtn-web-browser';
+
+export const loadAmplifyWebBrowser = () => {
+	try {
+		// metro bundler requires static string for loading module.
+		// See: https://facebook.github.io/metro/docs/configuration/#dynamicdepsinpackages
+		const module = require('@aws-amplify/rtn-web-browser')
+			?.module as WebBrowserModule;
```
HuiSF, line 11: nit: the type casting could be done in the truthy check below? At this line, the type should be `WebBrowserModule | undefined`

HuiSF (approved):

packages/react-native/src/moduleLoaders/loadAmplifyWebBrowser.ts
cshfang: Let me do this in a different PR so I can also apply to PN

jimblanc (approved):
LGTM

PR #12577, opened 2023-11-16
fix: export generateClient return types; don't throw error in genClient

iartemiev:
#### Description of changes

* export generateClient return types
* don't throw error if genClient is called before Amplify.configure. Silent return instead
* Bump data-schema dep versions

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


svidgen (approved):
Happy to have this ship to unblock, but would be good to do a little cleanup and update the date for the "timing" on that error.

PR #12584, opened 2023-11-17
add assertUserNotAuthenticated option to signInWithRedirect

dehli:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
+++ b/packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
@@ -44,7 +44,9 @@ export async function signInWithRedirect(
 	assertTokenProviderConfig(authConfig);
 	assertOAuthConfig(authConfig);
 	store.setAuthConfig(authConfig);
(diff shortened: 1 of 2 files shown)


cwomack:
Hey, @dehli 👋. We appreciate you opening this PR and will review it internally and reach out if there's further questions or suggestions.

cwomack:
@dehli, can you provide more context as to why just enforcing Google as your only federated provider for sign-in would not work here? We're not fully understanding what the benefit here would be to allow for an additional call of `signInWithRedirect` when you already have a user signed in with valid credentials.

dehli:
Hey @cwomack, for sure! The specific usecase is that you're logged in with your email/password but trying to access something that can only be accessed when logged in via Google. 

We have a saas tool that allows you to login with email/password or Google. Some of our customers would like to require that their data is only accessible by users who have logged in with Google. We don't want to force all users to login with Google which is why we needed this functionality. 

Let me know if that doesn't make sense or if you can think of a better workaround!

PR #12586, opened 2023-11-17
graphql: utilize default auth

iartemiev:
#### Description of changes

We don't utilize the `defaultAuthMode` inside of `_graphql`, which prevents us from signing the request correctly when IAM is the default auth mode

This PR defaults `authMode` to `defaultAuthMode` unless an explicit `authMode` was passed in at the client or call site.

* ~Also: updates `pretty-quick` dep. Necessary after Prettier was upgraded to unblock pre-commit husky hook.~ Nvm. pretty-quick doesn't seem to support prettier@3 at all. We'll need to find modify the pre-commit hook with a shell command to run prettier or use a newer pretty-quick equivalent tool

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/api-graphql/src/internals/InternalGraphQLAPI.ts
+++ b/packages/api-graphql/src/internals/InternalGraphQLAPI.ts
@@ -227,7 +227,7 @@ export class InternalGraphQLAPIClass {
 
 	private async _graphql<T = any>(
 		amplify: AmplifyClassV6,
-		{ query, variables, authMode }: GraphQLOptions,
+		{ query, variables, authMode: explicitAuthMode }: GraphQLOptions,
 		additionalHeaders: CustomHeaders = {},
 		abortController: AbortController,
 		customUserAgentDetails?: CustomUserAgentDetails,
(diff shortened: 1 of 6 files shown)


stocaaro (dismissed):

alharris-at (approved):

packages/api-graphql/src/internals/InternalGraphQLAPI.ts
```diff
@@ -238,8 +238,11 @@ export class InternalGraphQLAPIClass {
 			endpoint: appSyncGraphqlEndpoint,
 			customEndpoint,
 			customEndpointRegion,
+			defaultAuthMode
 		} = resolveConfig(amplify);
 
+		const authMode = explicitAuthMode ?? defaultAuthMode;
+
```
iartemiev, line 245: Around line 327 of this file we see that we skip setting `signingServiceInfo` when `authMode` is undefined. That's the source of the bug.

david-mcafee (approved):

svidgen (approved):

PR #12688, opened 2023-12-07
fix(datastore): Change outbox comparison to allow partial matching

stocaaro:
#### Description of changes
**Brief:** When multiple object updates for a single DataStore object are queued in the outbox to be synced to AppSync, comparison logic is overly strict and as a result only the first update is successfully applied. As a result Datastore appears to accept each update before reverting to the first update made while the outbox was queuing.

We have a number of tests the exercise variations on the problem and may be helpful to review. [Example](https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/__tests__/connectivityHandling.test.ts#L1129-L1180)

**Context:** When model changes are queued up in the outbox, `_version` field updates are passed on to other updates to the same object that are queued up in the outbox **IF** the object being processed is exactly and deeply equal to the return value from the AppSync update.

**Problem:** Updates send partial models to AppSync, for which AppSync returns complete models. These complete models may contain the same fields that are sent for update, but by also including all other fields, these fail the equality match and as a result we don't update the other objects queued up in the outbox. Since these objects don't receive the updated `_version`, they are rejected by AppSync conflict resolution and the object in Datastore reverts to the last saved content.

**Proposed fix:** Instead of comparing all fields, compare only the fields being updated when checking equality before updating object versions in the outbox.

**Example:**
Object sent to AppSync as an update (when updating only the title
`{title: 'post title 0', id: '1a94e616-506c-47f0-8dbc-db81f264e7e8'}`

Object returned from AppSync after successfully updating
`{id: '1a94e616-506c-47f0-8dbc-db81f264e7e8', title: 'post title 0', order: 348, altText: 'Alt Text - 617', blogPostsId: '52ba510d-5233-4866-a531-1563291d72b1', …}`

While these are obviously not an exact full equality match, all fields from the input object are exactly the same as the output object.

- Updated unit tests
- Added tests around the new `objectMatches` comparison function
- Tested in a sample application

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/connectivityHandling.test.ts
+++ b/packages/datastore/__tests__/connectivityHandling.test.ts
@@ -978,8 +978,7 @@ describe('DataStore sync engine', () => {
 				const retrieved = await DataStore.query(Post, postId);
 
 				await DataStore.save(
-					// @ts-ignore
-					Post.copyOf(retrieved, updated => {
+					Post.copyOf(retrieved!, updated => {
 						updated.title = updatedTitle;
 					})
 				);
@@ -1022,8 +1021,7 @@ describe('DataStore sync engine', () => {
 				// Validate that `query` returns the latest `title` and `_version`:
 				const queryResult = await DataStore.query(Post, postId);
 				expect(queryResult?.title).toEqual(title);
-				// @ts-ignore
-				expect(queryResult?._version).toEqual(version);
+				expect(queryResult!['_version'].toEqual(version));
 
 				if (blogId) expect(queryResult?.blogId).toEqual(blogId);
 			};
@@ -1040,9 +1038,7 @@ describe('DataStore sync engine', () => {
 						if (opType === 'UPDATE') {
 							const response: SubscriptionLogTuple = [
 								element.title,
-								// No, TypeScript, there is a version:
-								// @ts-ignore
-								element._version,
+								element['_version'],
 							];
 							// Track sequence of versions and titles
 							subscriptionLog.push(response);
@@ -1115,13 +1111,13 @@ describe('DataStore sync engine', () => {
 						['post title 0', 1],
 						['post title 1', 1],
 						['post title 2', 1],
-						['post title 0', 3],
+						['post title 2', 3],
 					]);
 
 					expectFinalRecordsToMatch({
 						postId: original.id,
 						version: 3,
-						title: 'post title 0',
+						title: 'post title 2',
 					});
 				});
 				test('rapid mutations on fast connection when initial create is not pending', async () => {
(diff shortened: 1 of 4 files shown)


packages/datastore/__tests__/connectivityHandling.test.ts
```diff
@@ -2187,9 +2186,8 @@ describe('DataStore sync engine', () => {
 			};
 
 			await resyncWith([
-				syncExpression(
-					LegacyJSONPost,
-					p => p?.title.eq("whatever, it doesn't matter.")
+				syncExpression(LegacyJSONPost, p =>
+					p?.title.eq("whatever, it doesn't matter.")
```
stocaaro, line 2190: Auto-formatting change

packages/datastore/src/sync/outbox.ts
```diff
@@ -193,10 +193,12 @@ class MutationEventOutbox {
 		// Don't sync the version when the data in the response does not match the data
 		// in the request, i.e., when there's a handled conflict
 		//
-		// NOTE: `incomingData` contains all the fields in the record, and `outgoingData`
-		// only contains updated fields, resulting in an error when doing a comparison
-		// of two equal mutations. Fix this, or mitigate otherwise.
```
stocaaro, line 198: This comment essentially describes what was needed here. Changing it to describe how/why this update works

packages/datastore/src/util.ts
```diff
@@ -473,6 +473,32 @@ export function sortCompareFunction<T extends PersistentModel>(
...
+	valB: object,
+	nullish: boolean = false
+) {
+	const aKeys = Object.keys(valA);
+
+	for (const key of aKeys) {
+		const aVal = valA[key];
+		const bVal = valB[key];
+
+		if (!valuesEqual(aVal, bVal, nullish)) {
```
stocaaro, line 494: Doing a deep bi-directional equality check on all fields keeps parity with what existed before and works in all cases I've experimented with. I have a couple cases I intend to get under integ testing today. Happy to have suggestions if others have thoughts on cases to cover.

svidgen (changes requested):
There's at least one case that still doesn't make sense to me.

https://github.com/aws-amplify/amplify-js/blob/ebf57ef182747f5ceff89308a3d05da1aff30122/packages/datastore/__tests__/connectivityHandling.test.ts#L1570-L1582

If everything is working correctly, I wouldn't expect `post title 0` to resurface at the end there.

stocaaro:
The case testcase pointed out by Jon is interesting. I had interpreted it incorrectly as being caused because the fake service wasn't feeding updates from the second client back to the DS client. The Fake service is much more accurate than I was thinking/expecting.

Whats actually happening here is that the version number in the fake service is incremented the version number on failures. The second client is posting an update with an old version, which isn't accepted, but does increment the version number. This incremented version number doesn't make it out to the DS client before the subsequent updates are attempted, so they also fail, (again incrementing the version number), which is how we end up with a final version number of 5.

From the service perspective:
- `['original title', 1]` - Succeed in creating post with title 'original title'
- `['post title 0', 2]` - Succeed in updating post with title 'post title 0'
- `['post title 0', 3]` - Fail to update post with title 'update from second client'
- `['post title 0', 4]` - Fail to update post with title 'post title 1'
- `['post title 0', 5]` - Fail to update post with title 'post title 2'

Issue #12665, opened 2023-12-02
Multi-level item in `selectionSet` is not received.


erickriva:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

GraphQL API

### Environment information

<details>

```
  System:
    OS: Windows 11 10.0.23590
    CPU: (16) x64 AMD Ryzen 9 5900HX with Radeon Graphics
    Memory: 13.21 GB / 23.91 GB
  Binaries:
    Node: 18.18.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.8.1 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.11.0 - ~\AppData\Local\pnpm\pnpm.EXE
    bun: Not Found
    Watchman: Not Found
  npmPackages:
    @aws-amplify/backend: 0.5.5
    @aws-amplify/backend-cli: 0.9.1
    aws-amplify: 6.0.5
    aws-cdk: Not Found
    aws-cdk-lib: 2.111.0
    typescript: 5.3.2
```

</details>

### Describe the bug

A multi-level item in `selectionSet` is not received.

### Expected behavior

Expected to receive data for any level of nested query.

### Reproduction steps

- Declare schema as following:
```javascript
import { a } from "@aws-amplify/backend";

const schema = a.schema({
  CommunityPost: a.model({
    id: a.id().required(),
    poll: a.hasOne("CommunityPoll"),
  }),

  CommunityPoll: a.model({
    id: a.id().required(),
    question: a.string().required(),
    answers: a.hasMany("CommunityPollAnswer").arrayRequired().valueRequired(),
  }),

  CommunityPollAnswer: a.model({
    id: a.id().required(),
    answer: a.string().required(),
    votes: a.hasMany("CommunityPollVote").arrayRequired().valueRequired(),
  }),

  CommunityPollVote: a.model({ id: a.id().required() }),
});
```

[...]

PR #12701, opened 2023-12-13
fix: Multi-level item in selectionSet is not received

erickriva:
Fix for issue #12665


packages/api-graphql/src/internals/APIClient.ts
```diff
@@ -360,57 +360,64 @@ export function customSelectionSetToIR(
 	modelName: string,
 	selectionSet: string[]
 ): Record<string, string | object> {
-	const modelDefinition = modelDefinitions[modelName];
-	const { fields: modelFields } = modelDefinition;
+	const dotNotationToObject = (
+		path: string,
+		modelName?: string
```
iartemiev, line 365: Let's make this param required. That'll resolve the build error
```suggestion
		modelName: string
```

iartemiev (changes requested):
Thank you for the contribution! The PR looks good overall. I suggested a couple of changes to get the build passing.

iartemiev:
@erickriva - bundle size check failed. I just pushed up a commit increasing the limit for this package.

iartemiev (approved):

packages/api-graphql/src/internals/APIClient.ts
```diff
@@ -472,6 +479,29 @@ export function selectionSetIRToString(
...
+	const isObject = (obj: any) => obj && typeof obj === 'object';
+
+	Object.keys(source).forEach((key: keyof T) => {
+		const targetValue = target[key];
+		const sourceValue = source[key];
+
+		if (isObject(targetValue) && isObject(sourceValue)) {
+			target[key] = deepMergeSelectionSetObjects(
+				Object.assign({}, targetValue),
+				sourceValue
```
svidgen, line 495: I might be having a slow brain day. But, the method is `source, target`, and we're flipping that here when recurse. This ends up mutating `source`, doesn't it? I'm not sure I can explain how this is correct if it is. If you can confirm this is the intended call, can you help me understand?

Assuming I'm just missing the point, it would be helpful to add a comment explaining why this is correct. I would anticipate being confused about this again 6 months from now otherwise.

packages/api-graphql/src/internals/APIClient.ts
```diff
@@ -472,6 +479,29 @@ export function selectionSetIRToString(
 	return res.join(' ');
 }
 
+function deepMergeSelectionSetObjects<T extends Record<string, any>>(
```
svidgen, line 482:
```suggestion
/**
 * Recursively merges selection set objects from `source` onto `target`.
 *
 * `target` will be updated. `source` will be left alone
 *
 * @param source The object to merge into target.
 * @param target The object to be mutated.
 */
function deepMergeSelectionSetObjects<T extends Record<string, any>>(
```

svidgen (commented):
I think I'm 98% there. I'm just a little confused on one point around the recursive merge at the moment.

I've also included a suggested docstring, but this was done in passing. It's a nit, largely written so I could summarize what I was looking at.

packages/api-graphql/src/internals/APIClient.ts
erickriva: You're right. The params had inverted order, I forgot that line from my tests. Besides that, `Object.assign` is not really necessary, so I get rid of it.
Your comment is accurate. `source` is not mutated, and `target` receives props from `source`.

packages/api-graphql/src/internals/APIClient.ts
erickriva: JSDoc added. Thank you.

packages/api-graphql/src/internals/APIClient.ts
```diff
@@ -472,6 +479,34 @@
...
+	const isObject = (obj: any) => obj && typeof obj === 'object';
+
+	Object.keys(source).forEach((key: keyof T) => {
+		const targetValue = target[key];
+		const sourceValue = source[key];
+
+		if (isObject(targetValue) && isObject(sourceValue)) {
+			target[key] = deepMergeSelectionSetObjects(sourceValue, targetValue);
+		} else {
+			target[key] = sourceValue;
```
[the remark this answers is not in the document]
iartemiev: @erickriva - would you be able to refactor as suggested here?

iartemiev (changes requested):
Blocking on the prototype pollution issue discovered by CodeQL

packages/api-graphql/src/internals/APIClient.ts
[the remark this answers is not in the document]
erickriva: I'm looking into it.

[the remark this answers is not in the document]
erickriva: @iartemiev I have read a lot about it and the main recommendation is to use `Map` instead of `Object`.
There is also an recommendation for `Object.freeze` if an object is absolutely necessary.
What do you think? Can we can go to `Map`, or it needs to keep using `Object`? Is there another option you can think?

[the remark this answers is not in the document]
erickriva: Found another much easier solution and `Map` was not necessary.
Took more time than I thought because CodeQL was not running in my account repos to test it and verify it was fixed before commit it here, but finally it worked.

packages/api-graphql/src/internals/APIClient.ts
```diff
@@ -472,6 +479,37 @@
...
+	Object.keys(source).forEach((key: keyof T) => {
+		const targetValue = target[key];
+		const sourceValue = source[key];
+
+		if (isObject(targetValue) && isObject(sourceValue)) {
+			target[key] = deepMergeSelectionSetObjects(sourceValue, targetValue);
+		} else {
+			// This verification avoids 'Prototype Pollution' issue
+			if (source.hasOwnProperty(key)) {
+				target[key] = sourceValue;
```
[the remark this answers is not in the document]
erickriva: CodeQL reporting this error again here, but it does not report (with this exact same code) when I create a repository to test specifically this function 😒

iartemiev:
@erickriva the prototype pollution mitigation looks good! The only apparent remaining blocker is a minor lint error:
> ERROR: src/internals/APIClient.ts:496:11 - Identifier 'key' is never reassigned; use 'const' instead of 'let'.

You can verify locally by running `yarn lint`

packages/api-graphql/src/internals/APIClient.ts
```diff
@@ -472,6 +479,34 @@ export function selectionSetIRToString(
...
+ * @param source The object to merge into target.
+ * @param target The object to be mutated.
+ */
+function deepMergeSelectionSetObjects<T extends Record<string, any>>(
+	source: T,
+	target: T
+) {
+	const isObject = (obj: any) => obj && typeof obj === 'object';
+
+	for (let key in source) {
```
iartemiev, line 496: To fix lint error:
```suggestion
	for (const key in source) {
```

iartemiev (approved):

svidgen (approved):
Awesome! Thank-you!

PR #12722, opened 2023-12-19
Use getRedirectUrl in all cases where redirectUri is passed

lukedigby:
#### Description of changes
This PR uses getRedirectUrl to resolve the correct url when multiple are provided in oauth config. The function was implemented in #12415 but two usages were missed. These missed usages cause errors when handling responses from oauth redirects. The PR adds these usages, completing the functionality.

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
+++ b/packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
@@ -127,7 +127,7 @@ export async function oauthSignIn({
 			currentUrl: url,
 			clientId,
 			domain,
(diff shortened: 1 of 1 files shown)


cwomack:
Hey, @lukedigby 👋.  Just wanted to follow up on this PR as the suggested changes are [currently reflected](https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/providers/cognito/apis/signInWithRedirect.ts) within the most recent version of Amplify.  I'll close this PR out since this should already be accomplished, but let me know if you find otherwise or have any further questions when using the most recent version of Amplify!

PR #12728, opened 2023-12-19
refactor(datastore): Move conflict resolution tests to their own file

stocaaro:
#### Description of changes
Outcome:
- The sequence / conflict resolution testing is separated out from other connection testing
- There are no changes in this aside from migrating the tests and the `before` / `after` blocks needed to keep them working.

#### Description of how you validated changes
This change is a pure refactor of test code. It is validated as correct since all tests continue to pass.

#### Related changes
1. :seedling:  **refactor(datastore): Move conflict resolution tests to their own file**
2. https://github.com/stocaaro/amplify-js/pull/84
3. https://github.com/stocaaro/amplify-js/pull/85
4. https://github.com/stocaaro/amplify-js/pull/86
5. https://github.com/stocaaro/amplify-js/pull/87

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #12732, opened 2023-12-20
refactor(datastore): Refactor conflict resolution tests optimize for concise/readable

stocaaro:
#### Description of changes

Outcome: 
- ~1300 lines -> ~700 lines aimed at maintaining readability. All assertions are the same as they where before this refactor.
- All test assertions are maintained with the same expected result for each test

We want to add OCC tests in a subsequent change. Rather than duplicating these tests as they where, I have refactored them, extracting out shared logic that isn't core to the testing story into a test harness aimed to enable/clarify these tests specifically.

The test suite continues to run to success

#### Related changes
2. :seedling:  **refactor(datastore): Refactor conflict resolution tests towards readability**
3. https://github.com/stocaaro/amplify-js/pull/85
4. https://github.com/stocaaro/amplify-js/pull/86

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/__tests__/conflictResolutionBehavior.test.ts
+++ b/packages/datastore/__tests__/conflictResolutionBehavior.test.ts
@@ -66,7 +66,7 @@ describe('DataStore sync engine', () => {
 		describe('observed rapid single-field mutations with variable connection latencies', () => {
 			describe('single client updates', () => {
 				test('rapid mutations on poor connection when initial create is not pending', async () => {
-					harness.connectionSpeed = 'slow';
+					harness.userInputLatency = 'fasterThanOutbox';
 					harness.latency = 'high';
 					const postHarness = await harness.createPostHarness({
 						title: 'original title',
@@ -89,13 +89,13 @@ describe('DataStore sync engine', () => {
 						['post title 0', 3],
 					]);
 
-					postHarness.expectCurrentToMatch({
-						version: 3,
+					expect(await postHarness.currentContents).toMatchObject({
+						_version: 3,
 						title: 'post title 0',
 					});
 				});
 				test('rapid mutations on fast connection when initial create is not pending', async () => {
-					harness.connectionSpeed = 'fast';
+					harness.userInputLatency = 'slowerThanOutbox';
 					harness.latency = 'low';
 					const postHarness = await harness.createPostHarness({
 						title: 'original title',
@@ -120,13 +120,13 @@ describe('DataStore sync engine', () => {
 						['post title 0', 4],
 					]);
 
-					postHarness.expectCurrentToMatch({
-						version: 4,
+					expect(await postHarness.currentContents).toMatchObject({
+						_version: 4,
 						title: 'post title 0',
 					});
 				});
 				test('rapid mutations on poor connection when initial create is pending', async () => {
-					harness.connectionSpeed = 'slow';
+					harness.userInputLatency = 'fasterThanOutbox';
 					harness.latency = 'high';
 					const postHarness = await harness.createPostHarness({
 						title: 'original title',
@@ -149,13 +149,13 @@ describe('DataStore sync engine', () => {
 						['post title 2', 1],
 					]);
 
-					postHarness.expectCurrentToMatch({
-						version: 1,
+					expect(await postHarness.currentContents).toMatchObject({
+						_version: 1,
 						title: 'post title 2',
 					});
 				});
 				test('rapid mutations on fast connection when initial create is pending', async () => {
-					harness.connectionSpeed = 'fast';
+					harness.userInputLatency = 'slowerThanOutbox';
 					harness.latency = 'low';
 					const postHarness = await harness.createPostHarness({
 						title: 'original title',
@@ -178,13 +178,13 @@ describe('DataStore sync engine', () => {
 						['post title 2', 1],
 					]);
 
-					postHarness.expectCurrentToMatch({
-						version: 1,
+					expect(await postHarness.currentContents).toMatchObject({
+						_version: 1,
 						title: 'post title 2',
 					});
 				});
 				test('observe on poor connection with awaited outbox', async () => {
-					harness.connectionSpeed = 'slow';
+					harness.userInputLatency = 'fasterThanOutbox';
 					harness.latency = 'high';
 					const postHarness = await harness.createPostHarness({
 						title: 'original title',
@@ -217,8 +217,8 @@ describe('DataStore sync engine', () => {
 						['post title 2', 4],
 					]);
 
-					postHarness.expectCurrentToMatch({
-						version: 4,
+					expect(await postHarness.currentContents).toMatchObject({
+						_version: 4,
 						title: 'post title 2',
 					});
 				});
@@ -249,8 +249,8 @@ describe('DataStore sync engine', () => {
 						['post title 2', 4],
 					]);
 
-					postHarness.expectCurrentToMatch({
-						version: 4,
+					expect(await postHarness.currentContents).toMatchObject({
+						_version: 4,
 						title: 'post title 2',
 					});
 				});
@@ -266,7 +266,7 @@ describe('DataStore sync engine', () => {
 			describe('Multi-client updates', () => {
 				describe('Updates to the same field', () => {
 					test('rapid mutations on poor connection when initial create is not pending', async () => {
-						harness.connectionSpeed = 'slow';
+						harness.userInputLatency = 'fasterThanOutbox';
 						harness.latency = 'high';
 						const postHarness = await harness.createPostHarness({
 							title: 'original title',
@@ -293,13 +293,13 @@ describe('DataStore sync engine', () => {
 							['update from second client', 4],
 						]);
 
-						postHarness.expectCurrentToMatch({
-							version: 4,
+						expect(await postHarness.currentContents).toMatchObject({
+							_version: 4,
 							title: 'update from second client',
 						});
 					});
 					test('rapid mutations on fast connection when initial create is not pending', async () => {
-						harness.connectionSpeed = 'fast';
+						harness.userInputLatency = 'slowerThanOutbox';
 						harness.latency = 'low';
 						const postHarness = await harness.createPostHarness({
 							title: 'original title',
@@ -327,13 +327,13 @@ describe('DataStore sync engine', () => {
 							['post title 0', 5],
 						]);
 
-						postHarness.expectCurrentToMatch({
-							version: 5,
+						expect(await postHarness.currentContents).toMatchObject({
+							_version: 5,
 							title: 'post title 0',
 						});
 					});
 					test('observe on poor connection with awaited outbox', async () => {
-						harness.connectionSpeed = 'fast';
+						harness.userInputLatency = 'slowerThanOutbox';
 						harness.latency = 'high';
 						const postHarness = await harness.createPostHarness({
 							title: 'original title',
@@ -369,13 +369,13 @@ describe('DataStore sync engine', () => {
 							['post title 2', 5],
 						]);
 
-						postHarness.expectCurrentToMatch({
-							version: 5,
+						expect(await postHarness.currentContents).toMatchObject({
+							_version: 5,
 							title: 'post title 2',
 						});
 					});
 					test('observe on fast connection with awaited outbox', async () => {
-						harness.connectionSpeed = 'fast';
+						harness.userInputLatency = 'slowerThanOutbox';
 						harness.latency = 'low';
 
 						const postHarness = await harness.createPostHarness({
@@ -412,8 +412,8 @@ describe('DataStore sync engine', () => {
 							['post title 2', 5],
 						]);
 
-						postHarness.expectCurrentToMatch({
-							version: 5,
+						expect(await postHarness.currentContents).toMatchObject({
+							_version: 5,
 							title: 'post title 2',
 						});
 					});
@@ -432,7 +432,7 @@ describe('DataStore sync engine', () => {
 					 * ultimately resulting in different final states.
 					 */
 					test('poor connection, initial create is not pending, external request is first received update', async () => {
-						harness.connectionSpeed = 'fast';
+						harness.userInputLatency = 'slowerThanOutbox';
 						harness.latency = 'high';
 						const postHarness = await harness.createPostHarness({
 							title: 'original title',
@@ -464,14 +464,14 @@ describe('DataStore sync engine', () => {
 							['original title', 'update from second client', 4],
 						]);
 
-						postHarness.expectCurrentToMatch({
-							version: 4,
+						expect(await postHarness.currentContents).toMatchObject({
+							_version: 4,
 							title: 'original title',
 							blogId: 'update from second client',
 						});
 					});
 					test('poor connection, initial create is not pending, external request is second received update', async () => {
-						harness.connectionSpeed = 'slow';
+						harness.userInputLatency = 'fasterThanOutbox';
 						harness.latency = 'high';
 						const postHarness = await harness.createPostHarness({
 							title: 'original title',
@@ -509,14 +509,14 @@ describe('DataStore sync engine', () => {
 							['post title 0', 'update from second client', 5],
 						]);
 
-						postHarness.expectCurrentToMatch({
-							version: 5,
+						expect(await postHarness.currentContents).toMatchObject({
+							_version: 5,
 							title: 'post title 0',
 							blogId: 'update from second client',
 						});
 					});
 					test('rapid mutations on fast connection when initial create is not pending (second field is `null`)', async () => {
-						harness.connectionSpeed = 'fast';
+						harness.userInputLatency = 'slowerThanOutbox';
 						harness.latency = 'low';
 
 						const postHarness = await harness.createPostHarness({
@@ -549,8 +549,8 @@ describe('DataStore sync engine', () => {
 							['post title 0', 'update from second client', 5],
 						]);
 
-						postHarness.expectCurrentToMatch({
-							version: 5,
+						expect(await postHarness.currentContents).toMatchObject({
+							_version: 5,
 							title: 'post title 0',
 							blogId: 'update from second client',
 						});
@@ -562,7 +562,7 @@ describe('DataStore sync engine', () => {
 					 * in different behavior.
 					 */
 					test('rapid mutations on fast connection when initial create is not pending (second field has initial value)', async () => {
-						harness.connectionSpeed = 'fast';
+						harness.userInputLatency = 'slowerThanOutbox';
 						harness.latency = 'low';
 
 						const postHarness = await harness.createPostHarness({
@@ -596,14 +596,14 @@ describe('DataStore sync engine', () => {
 							['post title 0', 'original blogId', 5],
 						]);
 
-						postHarness.expectCurrentToMatch({
-							version: 5,
+						expect(await postHarness.currentContents).toMatchObject({
+							_version: 5,
 							title: 'post title 0',
 							blogId: 'original blogId',
 						});
 					});
 					test('observe on poor connection with awaited outbox', async () => {
-						harness.connectionSpeed = 'slow';
+						harness.userInputLatency = 'fasterThanOutbox';
 						harness.latency = 'high';
 						const postHarness = await harness.createPostHarness({
 							title: 'original title',
@@ -641,14 +641,14 @@ describe('DataStore sync engine', () => {
 							['post title 2', 'update from second client', 5],
 						]);
 
-						postHarness.expectCurrentToMatch({
-							version: 5,
+						expect(await postHarness.currentContents).toMatchObject({
+							_version: 5,
 							title: 'post title 2',
 							blogId: 'update from second client',
 						});
 					});
 					test('observe on fast connection with awaited outbox', async () => {
-						harness.connectionSpeed = 'slow';
+						harness.userInputLatency = 'fasterThanOutbox';
 						harness.latency = 'low';
 						const postHarness = await harness.createPostHarness({
 							title: 'original title',
@@ -686,8 +686,8 @@ describe('DataStore sync engine', () => {
 							['post title 2', 'update from second client', 5],
 						]);
 
-						postHarness.expectCurrentToMatch({
-							version: 5,
+						expect(await postHarness.currentContents).toMatchObject({
+							_version: 5,
 							title: 'post title 2',
 							blogId: 'update from second client',
 						});
--- a/packages/datastore/__tests__/conflictResolutionBehavior.test.ts
+++ b/packages/datastore/__tests__/conflictResolutionBehavior.test.ts
@@ -1,76 +1,20 @@
-import {
-	warpTime,
-	unwarpTime,
-	pause,
-	getDataStore,
-	waitForEmptyOutbox,
-	graphqlServiceSettled,
-} from './helpers';
+import { warpTime, unwarpTime, pause } from './helpers';
+import { UpdateSequenceHarness } from './helpers/UpdateSequenceHarness';
 
 describe('DataStore sync engine', () => {
-	// establish types :)
-	let {
-		DataStore,
-		errorHandler,
-		schema,
-		connectivityMonitor,
-		Model,
-		ModelWithExplicitOwner,
-		ModelWithExplicitCustomOwner,
-		ModelWithMultipleCustomOwner,
-		BasicModel,
-		BasicModelWritableTS,
-		LegacyJSONPost,
-		LegacyJSONComment,
-		Post,
-		Comment,
-		HasOneParent,
-		HasOneChild,
-		CompositePKParent,
-		CompositePKChild,
-		graphqlService,
-		simulateConnect,
-		simulateDisconnect,
-		simulateDisruption,
-		simulateDisruptionEnd,
-	} = getDataStore({ online: true, isNode: false });
+	let harness: UpdateSequenceHarness;
 
 	beforeEach(async () => {
 		// we don't need to see all the console warnings for these tests ...
 		(console as any)._warn = console.warn;
 		console.warn = () => {};
 
-		({
-			DataStore,
-			errorHandler,
-			schema,
-			connectivityMonitor,
-			Model,
-			ModelWithExplicitOwner,
-			ModelWithExplicitCustomOwner,
-			ModelWithMultipleCustomOwner,
-			BasicModel,
-			BasicModelWritableTS,
-			LegacyJSONPost,
-			LegacyJSONComment,
-			Post,
-			Comment,
-			Model,
-			HasOneParent,
-			HasOneChild,
-			CompositePKParent,
-			CompositePKChild,
-			graphqlService,
-			simulateConnect,
-			simulateDisconnect,
-			simulateDisruption,
-			simulateDisruptionEnd,
-		} = getDataStore({ online: true, isNode: false }));
-		await DataStore.start();
+		harness = new UpdateSequenceHarness();
+		await harness.startDatastore();
 	});
 
 	afterEach(async () => {
-		await DataStore.clear();
+		await harness.destroy();
 		console.warn = (console as any)._warn;
 	});
 	describe('connection state change handling', () => {
@@ -81,7 +25,6 @@ describe('DataStore sync engine', () => {
 		afterEach(async () => {
 			unwarpTime();
 		});
-
 		/**
 		 * NOTE:
 		 * The following test assertions are based on *existing* behavior, not *correct*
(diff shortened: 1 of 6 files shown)


packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -0,0 +1,266 @@
...
+
+	/**
+	 * Query post, update, then increment counter.
+	 * @param postId - id of the post to update
+	 * @param updatedTitle - title to update the post with
+	 */
+	async revisePost(postId: string, updatedTitle: string) {
+		if (this.connectionSpeed === 'fast') {
+			await pause(200);
+		}
```
svidgen, line 217: I'm still not super in love with this. If this is the only place `connectionSpeed` is used, I would vote for `connectionSpeed` be called `userInputLatency` or something like that.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -121,470 +64,150 @@ describe('DataStore sync engine', () => {
...
-					 * will be different (here they are merged in the outbox). See the
-					 * tests following this one.
-					 */
-					graphqlService.setLatencies({
-						request: latency,
-						response: latency,
-						subscriber: latency,
-						jitter,
+					harness.connectionSpeed = 'slow';
+					harness.latency = 'high';
```
svidgen, line 70: Can we make this configuration more transparent without losing the semantics? I'd really like to be able to more easily see what connection speed and latency do specifically to timings, even if only on hover.

svidgen (approved):
I have a couple opinions, but non-blocking. Overall I think this is trending in the right direction, and don't want to impede the positive changes.

packages/datastore/__tests__/helpers/util.ts
```diff
@@ -459,15 +459,13 @@ export async function waitForSyncQueriesReady(verbose = false) {
...
 type GraphQLServiceSettledParams = {
 	graphqlService: any;
-	expectedNumberOfUpdates: number;
-	externalNumberOfUpdates: number;
+	expectedCallCount: number;
 	modelName: string;
 };
 
-export async function graphqlServiceSettled({
+export async function waitForExpectModelUpdateGraphqlCallCount({
```
stocaaro, line 466: This function deserves better comments and some internal annotations to describe the expected behavior. Ex. Why do we now pass `5_000` instead of `undefined` to the retry function?

packages/datastore/__tests__/helpers/fakes/graphqlService.ts
```diff
@@ -260,6 +261,22 @@ export class FakeGraphQLService {
 		};
 	}
 
+	private makeOCCConflictUnhandeled(existingObject, call) {
```
stocaaro, line 264: Add a comment

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -0,0 +1,266 @@
...
+		this.datastoreFake = getDataStore({
+			online: true,
+			isNode: false,
+		});
+
+		this.subscriptionLogSubscription = this.datastoreFake.DataStore.observe(
+			this.datastoreFake.Post
+		).subscribe(({ opType, element }) => {
+			if (opType === 'UPDATE') {
+				// const response: SubscriptionLogTuple = [
```
stocaaro, line 120: I think this was removed in a subsequent change. Should have been removed here.

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -0,0 +1,266 @@
...
+
+	async externalPostUpdate({
+		originalPostId,
+		updatedFields,
+		version,
+	}: ExternalPostUpdateParams) {
+		await this.datastoreFake.graphqlService.externalGraphql(
+			{
+				query: `
+                    mutation operation($input: UpdatePostInput!, $condition: ModelPostConditionInput) {
```
stocaaro, line 179: This query is essentially static. Pull it to the top of the file to keep this code concise.

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -0,0 +1,266 @@
...
+
+	expectCurrentToMatch(values: { version; title; blogId? }) {
+		this.harness.expectFinalRecordsToMatch({
+			postId: this.original.id,
+			...values,
+		});
+	}
+}
+
+export class UpdateSequenceHarness {
```
stocaaro, line 65: Needs a robust comment with example use to ensure that its not hard to understand and work with later.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: Yep. It was on my notes to do so and I missed circling back to make that happen. Will see whats possible today.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -121,470 +64,150 @@ describe('DataStore sync engine', () => {
...
-					 * Because we have increased the latency, and don't wait for the outbox
-					 * to clear on each mutation, the outbox will merge some of the mutations.
-					 * In this example, we expect the number of requests received to be one less than
-					 * the actual number of updates. If we were running this test without
-					 * increased latency, we'd expect more requests to be received.
-					 */
-					await graphqlServiceSettled({
-						graphqlService,
-						expectedNumberOfUpdates: expectedNumberOfUpdates - 1,
-						externalNumberOfUpdates: 0,
```
david-mcafee, line 329: Unless I'm missing something (and it's quite possible that I am!), I'm not seeing that we're making this assertion in any of these tests. Is there a reason for removing this? This becomes important in later tests, because we want to ensure that we're indeed receiving the "external" update on the fake service. When I was fine tuning these tests and there was any change in a latency or pause, the fake service would miss the update.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -121,470 +64,150 @@ describe('DataStore sync engine', () => {
...
 					});
-
+					await harness.outboxSettled();
 					// Note: We do NOT wait for the outbox to be empty here, because
 					// we want to test concurrent updates being processed by the outbox.
+					await postHarness.revise('post title 0');
+					await postHarness.revise('post title 1');
+					await postHarness.revise('post title 2');
+					await harness.outboxSettled();
+					await harness.expectUpdateCallCount(2);
```
david-mcafee, line 82: I would rename this, since this is doing more than verifying the expected update call count - it's also ensuring that the fake service has settled, which wasn't immediately clear to me. Also, I would personally prefer to keep the comment here, or at least be able to view it on hover.

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	/**
+	 * Query post, update, then increment counter.
+	 * @param postId - id of the post to update
+	 * @param updatedTitle - title to update the post with
+	 */
+	async revisePost(postId: string, updatedTitle: string) {
+		if (this.connectionSpeed === 'fast') {
+			await pause(200);
```
david-mcafee, line 216: This is confusing to me. I'd rather we explicitly perform the pause in the tests so that it's clear what we're doing and why (i.e. we're simulating an artificial "button click"). This also seems really easy to break, and not know why it's breaking, IMO.

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -0,0 +1,266 @@
...
+				this.datastoreFake.Post.copyOf(retrieved, updated => {
+					updated.title = updatedTitle;
+				})
+			);
+		}
+		if (this.settleOutboxAfterRevisions) {
+			await this.outboxSettled();
+		}
+
+		this.expectedNumberOfInternalUpdates++;
```
david-mcafee, line 233: I'd explain why here

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -0,0 +1,266 @@
...
+		);
+		if (retrieved) {
+			const x = await this.datastoreFake.DataStore.save(
+				this.datastoreFake.Post.copyOf(retrieved, updated => {
+					updated.title = updatedTitle;
+				})
+			);
+		}
+		if (this.settleOutboxAfterRevisions) {
+			await this.outboxSettled();
```
david-mcafee, line 230: I'm confused what's happening here, since the tests explicitly wait on consecutive updates without a param. IMO following this pattern makes it more difficult to track down what's happening.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -121,470 +64,150 @@ describe('DataStore sync engine', () => {
...
-					await waitForEmptyOutbox();
-
-					await revPost(original.id, 'post title 1');
-					await waitForEmptyOutbox();
-
-					await revPost(original.id, 'post title 2');
-					await waitForEmptyOutbox();
-					//endregion
+					await postHarness.revise('post title 0');
+					await harness.outboxSettled();
```
david-mcafee, line 202: Doesn't `revise` optionally wait on the outbox? I think we should either update `revise` to not do that, or remove this here. I'd favor the former.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
david-mcafee: Also, when I originally wrote these tests, @svidgen suggested that we use a variable (that is incremented) on each revision so that if we decided to change the number of revisions in the future, we wouldn't have to also update these values. I'd suggest we stick with that approach, but that's more of a nit.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -121,470 +64,150 @@ describe('DataStore sync engine', () => {
...
-					 * updates on a record that is still in the outbox.
-					 */
-
-					//region perform consecutive updates
-					await pause(200);
-					await revPost(original.id, 'post title 0');
-
-					await pause(200);
-					await revPost(original.id, 'post title 1');
+					harness.connectionSpeed = 'fast';
```
david-mcafee, line 158: I'd probably add a comment here explaining that this changes the behavior of `revise`

packages/datastore/__tests__/helpers/util.ts
```diff
@@ -459,15 +459,13 @@ export async function waitForSyncQueriesReady(verbose = false) {
...
 	graphqlService,
-	expectedNumberOfUpdates,
-	externalNumberOfUpdates,
```
david-mcafee, line 470: We need to know the number of external updates received in order to verify that all subscription messages have been sent - otherwise we risk running into a scenario where the test appears to be "complete", but if we'd waited a bit longer, we'd get additional subscription messages that would end up reverting the original change (i.e. the behavior we're attempting to test here).

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -121,470 +64,150 @@ describe('DataStore sync engine', () => {
...
-						request: latency,
-						response: latency,
-						subscriber: latency,
-						jitter,
+					harness.connectionSpeed = 'slow';
+					harness.latency = 'high';
+					const postHarness = await harness.createPostHarness({
+						title: 'original title',
+						blogId: 'blog id',
 					});
```
david-mcafee, line 192: We intentionally set the latencies _after_ the initial update. See comment above. We want to make sure that the initial save happens successfully and doesn't get merged with the updates (on purpose).

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -594,60 +217,29 @@ describe('DataStore sync engine', () => {
...
-					/**
-					 * We wait for the empty outbox on each mutation, because
-					 * we want to test non-concurrent updates (i.e. we want to make
-					 * sure all the updates are going out and are being observed)
-					 */
-
-					//region perform consecutive updates
-					await revPost(original.id, 'post title 0');
-					await waitForEmptyOutbox();
+					harness.latency = 'low';
```
david-mcafee, line 226: Same here.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -121,470 +64,150 @@ describe('DataStore sync engine', () => {
...
-					 * will be different (here they are merged in the outbox). See the
-					 * tests following this one.
-					 */
-					graphqlService.setLatencies({
-						request: latency,
-						response: latency,
-						subscriber: latency,
-						jitter,
+					harness.connectionSpeed = 'slow';
+					harness.latency = 'high';
```
david-mcafee, line 188: See next comment

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -674,178 +265,100 @@ describe('DataStore sync engine', () => {
...
-
-						await waitForEmptyOutbox();
-
-						graphqlService.setLatencies({
-							request: latency,
-							response: latency,
-							subscriber: latency,
-							jitter,
+						harness.connectionSpeed = 'slow';
+						harness.latency = 'high';
```
david-mcafee, line 270: I won't put comments on all of these, but we shouldn't set this yet (see above comment).

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
david-mcafee, line 70: We previously set the latencies _after_ the initial create. We want to make sure that the initial save happens successfully and doesn't get merged with the updates (i.e. we're not concerned with testing the original save in any way in most of these tests, except for the ones where we do not await the outbox). If there is a reason for moving this to the beginning of the test, I would 1) add a very long pause at the end of this test to make sure no other subscription messages eventually come through, and 2) would add back the check on the fake service to validate that external messages are coming through (that is, on the tests that have external updates). The hardest thing about these tests is ensuring that your assertions truly test the final state of things.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -121,470 +64,150 @@ describe('DataStore sync engine', () => {
...
-					 * however, the `expectedNumberOfUpdates` received by the fake service
-					 * will be different (here they are merged in the outbox). See the
-					 * tests following this one.
-					 */
-					graphqlService.setLatencies({
-						request: latency,
-						response: latency,
-						subscriber: latency,
-						jitter,
+					harness.connectionSpeed = 'slow';
```
david-mcafee, line 69: To me this is less readable than before. On one of my previous revisions it was suggested that I have a flag for whether or not to pause, but that become more confusing and less readable - it becomes unclear what's happening in the test (i.e. are we pausing, are we not pausing, and why). Since it's _already_ unclear what's happening with DataStore in this tested behavior, I think it's better to be more explicit, even if that means more repeated lines of test code. Additionally, if someone is updating these and uses the wrong value (since `connectionSpeed` could be confused with setting the actual connection speed), it's possible we'd have incorrect test results.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -856,47 +369,39 @@ describe('DataStore sync engine', () => {
...
 						});
 					});
 					test('observe on fast connection with awaited outbox', async () => {
-						const original = await DataStore.save(
-							new Post({
-								title: 'original title',
-								blogId: 'blog id',
-							})
-						);
+						harness.connectionSpeed = 'fast';
```
david-mcafee, line 378: This is incorrect. Since you are setting connectionSpeed to `fast` you are now also adding an unnecessary pause between revisions, even though we are already waiting on the empty outbox.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -952,170 +432,124 @@ describe('DataStore sync engine', () => {
...
-						);
-
-						await waitForEmptyOutbox();
-
-						graphqlService.setLatencies({
-							request: latency,
-							response: latency,
-							subscriber: latency,
-							jitter,
+						harness.connectionSpeed = 'fast';
```
david-mcafee: This is also incorrect - we should not be pausing between updates. Instead, we should rely on the outbox being settled.

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
stocaaro: I have made the name change we discussed.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -952,170 +432,124 @@ describe('DataStore sync engine', () => {
...
-						await waitForEmptyOutbox();
-
-						await graphqlServiceSettled({
-							graphqlService,
-							expectedNumberOfUpdates: expectedNumberOfUpdates - 1,
-							externalNumberOfUpdates: 1,
-							modelName: 'Post',
-						});
+						await harness.outboxSettled();
+						await harness.expectUpdateCallCount(3);
```
david-mcafee, line 455: My worry here is that we're combing the number of expected updates with the number of external updates, but we have no way to validate what is what. In this case, we make 3 revisions by the primary client but expect 2 updates to be sent to the service for the primary client, and a single updates to be made by the external client.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -952,170 +432,124 @@ describe('DataStore sync engine', () => {
...
-						await revPost(original.id, 'post title 0');
-
-						await revPost(original.id, 'post title 1');
+						await postHarness.revise('post title 0');
+						await postHarness.revise('post title 1');
 
 						/**
 						 * Ensure that the external update is received after the
-						 * primary client's first update.
+						 * primary client's first update. TODO - Can we just settle the outbox?
```
david-mcafee, line 486: No, because that never enters the outbox.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -952,170 +432,124 @@ describe('DataStore sync engine', () => {
...
 					});
 					test('rapid mutations on fast connection when initial create is not pending (second field is `null`)', async () => {
-						const original = await DataStore.save(
-							new Post({
-								title: 'original title',
-							})
-						);
-
-						await waitForEmptyOutbox();
+						harness.connectionSpeed = 'fast';
```
david-mcafee: This is problematic because we will end up pausing on the third revision. If you look at the original test, we do _not_ pause on the third revision (on purpose), and expect the external update to go out at roughly the same time. We also do not pause after the last revision and instead wait on the outbox.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: If I changed the name to `expectGraphqlSettledWithUpdateCallCount`, would this be clear enough to communicate whats happening here?

I went back and forth with the call count tracking. It gets trickier later on with OCC because OCC fails on collision which triggers a retry shifting the count.

I understand how we landed on tracking updates as variables, but even in the prior tests we where adjusting in some tests. I found it harder to navigate than just having a single static value per test, which is what prompted my refactor to simplify in this direction.

Really appreciate the context. I'll setup some time to chat about this.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: You found this in the later comment thread above. Archiving for now.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: Good call. This is now added as a comment that shows in the on hover tooltip for this.

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
david-mcafee: I don't understand, sorry! Can you explain?

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	async expectUpdateCallCount(expectedCallCount: number) {
+		/**
+		 * Because we have increased the latency, and don't wait for the outbox
+		 * to clear on each mutation, the outbox will merge some of the mutations.
+		 * In this example, we expect the number of requests received to be one less than
+		 * the actual number of updates. If we were running this test without
+		 * increased latency, we'd expect more requests to be received.
+		 */
```
david-mcafee, line 211: This comment doesn't belong here, since we use this util in each test, yet some tests do not increase the latency / do wait for the outbox to clear after each mutation.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: You are on point. Somehow the feature to use the flag was added in this change and adopted in the next one. I'm going to need to do a big rebase anyway, so no reason not to pull it forward since it doesn't cause these test expectations to change.

Will post an update with this change in a few.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
david-mcafee: I should probably elaborate on this more to be clear :). Though today we do not make any additional validations on the fake service, we _should_ (was previously discussed on my PR, and if there's nothing in the backlog, there should be - will check after this review). Nonetheless, I think it's more readable if we are explicit in the tests about how many updates we expect, and how many external updates we expect.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -594,60 +217,29 @@ describe('DataStore sync engine', () => {
...
-					/**
-					 * We wait for the empty outbox on each mutation, because
-					 * we want to test non-concurrent updates (i.e. we want to make
-					 * sure all the updates are going out and are being observed)
-					 */
-
-					//region perform consecutive updates
-					await revPost(original.id, 'post title 0');
-					await waitForEmptyOutbox();
+					harness.latency = 'low';
```
david-mcafee, line 226: This is reset by the fake service, so unnecessary.

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
stocaaro: I have removed this because it was disused when I migrated to explicitly declaring the count here, which was prompted by the incongruity between revision calls and graphql calls.

I suppose we could input more information to calculate the number of calls, but I think we would want to make explicit. Something like:
`revisions + external calls - local merges + retried exceptions`

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -121,470 +64,150 @@ describe('DataStore sync engine', () => {
...
-					 * will be different (here they are merged in the outbox). See the
-					 * tests following this one.
-					 */
-					graphqlService.setLatencies({
-						request: latency,
-						response: latency,
-						subscriber: latency,
-						jitter,
+					harness.userInputLatency = 'fasterThanOutbox';
+					harness.latency = 'high';
```
david-mcafee, line 70: I'm guessing my previous comment was removed since this was updated, but we should not be setting the latency prior to the first save, since we want to ensure that nothing is accidentally preventing the initial save from happening. We should set the latencies _after_ the initial save. If I remember correctly, if we set the latency before the first save, we potentially enter a situation where the outbox is empty, but we are not guaranteed that the fake service has completed the save.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: Good call. Fixing. Also, related to some of the above threads, I agree that create doesn't need to be subject to the latency specified in the test. Similar, `userInputLatency` and `settleOutboxAfterRevisions` settings only apply to revisions. I'd like to keep them together, but am apply a revision to all tests to move these settings below the create for all the reasons you've identified. I'll post this update after I review the rest of your comments.

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -0,0 +1,266 @@
...
+			attributes.map(attribute => post[attribute])
+		);
+	}
+
+	get latency(): 'low' | 'high' {
+		return this.latencyValue;
+	}
+
+	set latency(value: 'low' | 'high') {
+		if (value === 'low') {
```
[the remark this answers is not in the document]
stocaaro: In my thinking, I'd like to keep setting this to `low` in all tests where its not `high` for clarity for the person working on these tests next, so that its clear that each test is a explicitly `high` or `low`. This is inconsistent with how I've used `settleOutboxAfterRevisions`, which defaults to false and is turned true in tests where we want this behavior. I might consider making `latency` a bool renamed something like `elevatedLatency`, which defaults to false and isn't included in tests where latency isn't elevated.  Either, both of these could just be command functions, eg: `harness.settleOutboxAfterRevisions()` and `harness.elevatedLatency()`.

Thoughts?

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: Yeah. I discovered this in testing, but apparently forgot to remove the TODO. Doing so now.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: Discussed elsewhere. We should catch up on this one.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: Ack changing

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: Ack. Updated

david-mcafee (changes requested):
At a high-level, I would argue that some of these changes make the tests _less_ readable than before, but perhaps that’s just me!!! Understanding what’s happening in DataStore at any given moment is quite difficult with all of the async concurrent behavior, and though abstracting the repeated logic out of the tests means less lines of code, I would argue that I (personally) have a harder time understanding what’s happening at each step, and I worry that we'll make updates to the tests in the future that will result in different behavior being tested (and thus would no longer identify a regression). If I was paged at 3am and had to figure out what’s broken with DataStore, I’d want the tests to be as easy to reason about as possible, with no additional cognitive overload, even if that means tests are repeating themselves too much. 

At a minimum, I think we should 1) keep the manual pauses between DataStore updates (as an example, some of these tests now pause when they should not), 2) re-add some of the comments back into the tests (see specific feedback below, but adding comments to the helpers does not work because we do not always wait on the outbox, etc), and 3) go back to setting the latencies _after_ the initial save has completed.

There are a few places where we are _potentially_ no longer testing the same behavior as before. Even though the assertions are the same, that doesn’t necessarily mean that DataStore is processing the updates in the same way - for instance, the assertions across most of the tests are the same, even though the tests are testing different DataStore patterns. I also think we should continue to supply the "external" update number to the assertion testing that the fake service has settled so that, in the future, we can update the fake service to differentiate between primary client / external client updates, since that's a weak spot in these tests (and I would argue one of the most important updates we should make before committing the fix).

Lastly, since this is a pretty big refactor, I think it would be smart to test these "manually" by adding a very long pause at the end of each of these tests to make sure no other subscription messages eventually come through.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: I would be interested in your thoughts with the change to `userInputLatency` and the addition of comments that appear on hover.
image

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: Update. More thorough comment on a different thread.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: Ack. New update is in that does this.

packages/datastore/__tests__/helpers/util.ts
stocaaro: We discussed this and I have discussed this with Jon. I'm recommending we settle on having a single number now and augmenting this as needed. It's simpler to work with, but potentially harder to reason around. As discussed elsewhere both merges and retries impact this.

We have discussed the benefit of digging in and adding tests for merges and retries (as a maybe later opportunity).

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
stocaaro: We discussed this out of band and settled on adding a comment above the `postHarness.revise` function that articulated which settings impact revision operations in what ways.

image

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
stocaaro: I've posed this to Jon. The pause to bring things down to human time had raised questions/concerns earlier. He has recommended changing this to `userInputLatency`, which has been done and with your input, the `revise` function comment now clarifies the impact of this setting on the revise function.

I think this is a good midpoint to settle on for this change, but happy to iterate further if you have more input.

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -0,0 +1,327 @@
...
+ */
+export class UpdateSequenceHarness {
+	private datastoreFake: ReturnType<typeof getDataStore>;
+
+	/**
+	 * Should we inject latency before each standard client `Post` revision?
+	 *
+	 * `slowerThanOutbox` will add 200ms of latency before each revision against datastore
+	 */
+	userInputLatency: 'fasterThanOutbox' | 'slowerThanOutbox' =
```
david-mcafee, line 136: This isn't necessarily faster or slower than the outbox (if that were the case, we could simply wait on the outbox). Thoughts on making this a boolean?

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -0,0 +1,327 @@
...
+	 * @returns A post harness wrapped around a newly created post
+	 */
+	async createPostHarness(
+		args: ConstructorParameters<typeof Post>[0],
+		settleOutbox: boolean = true
+	) {
+		const original = await this.datastoreFake.DataStore.save(
+			new this.datastoreFake.Post(args)
+		);
+		if (settleOutbox) {
```
david-mcafee, line 241:
```suggestion
// We set this to `false` when we want to test updating a record that is still in the outbox.
		if (settleOutbox) {
```

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -674,178 +265,97 @@ describe('DataStore sync engine', () => {
...
-							jitter,
+						const postHarness = await harness.createPostHarness({
+							title: 'original title',
+							blogId: 'blog id',
 						});
 
-						//region perform consecutive updates
-						await revPost(original.id, 'post title 0');
-						await waitForEmptyOutbox();
+						harness.userInputLatency = 'slowerThanOutbox';
```
david-mcafee, line 341: We don't pause between revisions on the original test - is there a reason to do that here?

```suggestion
						harness.userInputLatency = 'fasterThanOutbox';
```

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -952,171 +425,124 @@ describe('DataStore sync engine', () => {
...
 							version: 1,
 						});
 
-						await revPost(original.id, 'post title 2');
-						//endregion
+						await postHarness.revise('post title 2');
 
-						await waitForEmptyOutbox();
+						await harness.outboxSettled();
+						await harness.expectGraphqlSettledWithUpdateCallCount(3);
```
david-mcafee, line 448: I may be misunderstanding something, but we previously made three revisions from the original client, and one from the external client. Due to the merge, we subtract one from the original revisions, and end up with 4.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -952,171 +425,124 @@ describe('DataStore sync engine', () => {
...
 
-						await graphqlServiceSettled({
-							graphqlService,
-							expectedNumberOfUpdates: expectedNumberOfUpdates - 1,
-							externalNumberOfUpdates: 1,
-							modelName: 'Post',
-						});
-
-						expect(subscriptionLog).toEqual([
-							['original title', null, undefined],
```
david-mcafee, line 995: Was this incorrect?

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -952,171 +425,124 @@ describe('DataStore sync engine', () => {
...
-							graphqlService,
-							expectedNumberOfUpdates,
-							externalNumberOfUpdates: 1,
-							modelName: 'Post',
-						});
+						await harness.outboxSettled();
+						await harness.expectGraphqlSettledWithUpdateCallCount(4);
 
-						expect(subscriptionLog).toEqual([
-							['original title', null, undefined],
```
david-mcafee, line 1057: Same here?

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -952,171 +425,124 @@ describe('DataStore sync engine', () => {
...
-							graphqlService,
-							expectedNumberOfUpdates: numberOfUpdates,
-							externalNumberOfUpdates: 1,
-							modelName: 'Post',
-						});
+						await harness.outboxSettled();
+						await harness.expectGraphqlSettledWithUpdateCallCount(4);
 
-						expect(subscriptionLog).toEqual([
-							['original title', null, undefined],
```
david-mcafee, line 1109: Same here

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -1128,101 +554,71 @@ describe('DataStore sync engine', () => {
...
 
-						await graphqlServiceSettled({
-							graphqlService,
-							expectedNumberOfUpdates: numberOfUpdates,
-							externalNumberOfUpdates: 1,
-							modelName: 'Post',
-						});
-
-						expect(subscriptionLog).toEqual([
-							['original title', 'original blogId', undefined],
```
david-mcafee, line 1168: I won't leave a comment on all of these, perhaps let's discuss offline?

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
svidgen: Changing this was my suggestion. But, if it helps speed things along, I'm OK either way. We'll have plenty of time for more refactors later if we think this PR is moving things in a better (more readable/grokkable) direction.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: This case was including creates, unlike the above tests. I normalized to only include the create event, which dropped the initial create request.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
david-mcafee: Nevermind, I can't count!

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
stocaaro: We've updated this to change the language and idioms slightly. The result will be the same.

david-mcafee (approved):
LGTM!! 🎉

svidgen (approved):

PR #12740, opened 2023-12-21
refactor(datastore): Refactor test names and make subscription events more sample app like

stocaaro:
#### Description of changes
Outcome:
- Test names changed to align with the harness features used per test
- Wait for [all subscriptions to be published](https://github.com/stocaaro/amplify-js/blob/03-consecutive-updates-take2/packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts#L221) before proceeding (remove a race condition between test assertions and subscription updates)

Tests updated in reference to how users receive settled updates twice in sample apps (once via mutation response and then later via subscription update). 

#### Related changes

3. :seedling:  **refactor(datastore): Refactor test names and make subscriptions more latent / realistic**
4. https://github.com/stocaaro/amplify-js/pull/88

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -213,6 +218,7 @@ export class UpdateSequenceHarness {
 	 */
 	async outboxSettled() {
 		await waitForEmptyOutbox();
+		await Promise.all(subscriptionDeliveryPromiseList);
```
david-mcafee, line 221: Can you add a comment explaining what this is doing, and why?

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -55,6 +59,7 @@ mutation operation($input: UpdatePostInput!, $condition: ModelPostConditionInput
  */
 const jitter: number = 0;
 const highLatency = 1000;
+const highSubscriptionLatency = 1100;
```
david-mcafee, line 62: Since this change requires us to update our test assertions (i.e. more sub messages are coming through), can you add a comment here explaining _why_ we have a different value for this from the other latencies?

packages/datastore/__tests__/helpers/fakes/graphqlService.ts
```diff
@@ -39,6 +39,15 @@ const defaultLatencies: FakeLatencies = {
 	jitter: 5,
 };
 
+export let subscriptionDeliveryPromiseList: Promise<void>[] = [];
```
david-mcafee, line 42: Can you add a comment above this explaining what this is?

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -86,14 +86,15 @@ describe('DataStore sync engine', () => {
 						['post title 1', 1],
 						['post title 2', 1],
 						['post title 0', 3],
+						['post title 0', 3],
```
david-mcafee, line 89: Is this coming through because of the increased subscription latency, the fact that we are now waiting on all subs to be published when awaiting the outbox, or both?

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -210,18 +212,21 @@ describe('DataStore sync engine', () => {
 						['original title', 1],
 						['post title 0', 1],
 						['post title 0', 2],
+						['post title 0', 2],
```
david-mcafee, line 215: In this situation, I'm assuming we are now getting not only the local sub messages, but also the sub message from the service since we've changed the behavior of `outboxSettled`? Don't need a comment in the code, but can you explain why that's better, here?

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -354,19 +360,22 @@ describe('DataStore sync engine', () => {
 							['original title', 1],
 							['post title 0', 1],
 							['post title 0', 2],
+							['post title 0', 2],
 							['post title 1', 2],
 							['post title 1', 3],
 							['update from second client', 4],
+							['update from second client', 4],
```
david-mcafee, line 367: I'm trying to understand why we now get two sub messages for the external update. Going to think about this more, but maybe you know why.

david-mcafee (changes requested):
Since these changes result in different test assertions, I think I'd like to see a few comments explaining exactly what's happening (and why) - specific comments requesting those changes are included inline.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: Correct. When the subscription arrives after the mutation response has been processed out of the outbox, then datastore registers the update twice, like what I have observed in the sample apps I've been checking these against.

packages/datastore/__tests__/helpers/fakes/graphqlService.ts
stocaaro: On it.

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
stocaaro: Good call.

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
stocaaro: I'm going to split this out and change the function names a bit.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: - The local change doesn't increment the version, so `['post title 0', 1]` is the local update.
- When the mutation response is process and dequeued successfully from the inbox we get the first updated version message `['post title 0', 2]`
- When the update subscription received the update (and there is no corresponding change pending in the outbox) then we get the duplicate message `['post title 0', 2]`

This is the sequence of events that occurs when we settle both the outbox and the subscription messages between revisions while the latency of subscription messages is higher than the mutation responses. (otherwise, the subscription update is dropped in favor of the outbox mutation being processed).

I'm preferring a change to this behavior because the OCC tests coming in the next change run into issues when the subscriptions are processed first and this behavior is more consistent with how sample apps work (it is possible, but unlikely that the subscription update would process first).

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: I'm not sure, just looking at it. I'll go do some debugging and get back to you.

stocaaro: Alright. I don't love this, but the first `['update from second client', 4]` is actually the subscription message from the prior 'post title 1' update.

[Just before sending subscription updates](https://github.com/stocaaro/amplify-js/blob/03-consecutive-updates-take2/packages/datastore/src/datastore/datastore.ts#L2141-L2158), datastore looks up the actual value in storage for the item right now. It sends the current value along instead of the received item content.

In this case, the external update happens in the 100ms gap between when 'post title 1''s mutation returns and is process and when the subscription event for 'post title 1' arrives, and so that subscription event carries the update value.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -65,7 +65,7 @@ describe('DataStore sync engine', () => {
 		 */
 		describe('observed rapid single-field mutations with variable connection latencies', () => {
 			describe('single client updates', () => {
-				test('rapid mutations on poor connection when initial create is not pending', async () => {
+				test('no input delay, high latency where we wait for the create to clear the outbox', async () => {
```
david-mcafee, line 68: nit (and not a blocker): a comment summarizing what we mean by "no input delay" at this first test may be useful. Took me a second to remember what this was, and why it existed.

packages/datastore/__tests__/helpers/UpdateSequenceHarness.ts
```diff
@@ -139,13 +146,13 @@ export class UpdateSequenceHarness {
 		this.isUserInputDelayed = true;
 	}
 
-	private isOutboxSettledAfterRevisions: boolean = false;
+	private isSettledAfterRevisions: boolean = false;
 
 	/**
 	 * Do we want to settle the outbox after each `Post` revision call?
```
david-mcafee, line 152: Since this now also awaits subscription responses, I'd probably update this comment:

```suggestion
	 * Determines whether we settle the outbox and await for the subscription response after each mutation.
```

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
david-mcafee: Thanks for digging into this / adding the comment!

david-mcafee (dismissed):
Thanks for responding to my previous questions on this one! Left two minor (non-blocking) suggestions.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: Yep. In our last revision before the break, we moved input delay to being a flag that can be turned on instead of a field that is set to true or false in each test. Since the `no input delay` tests don't have any harness call to hover over for comment, I think its valuable to get a comment in here for the first test, so I've added one.

iartemiev (approved):

david-mcafee (approved):

PR #12789, opened 2024-01-03
fix: Recreate Analytics clients when credentials change

jimblanc:
#### Description of changes

This change fixes an issue where Analytics clients for Kinesis, Personalize, and Kinesis Firehose weren't being updated when credentials change (e.g. after expiration). Pinpoint is not impacted by this issue as it uses the legacy buffer that specifically handles this case.

Tested providers in sample applications by waiting for credentials to expire and verifying that new events can be recorded. E2E test suite run which covers Pinpoint, Personalize, and Kinesis.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/analytics/src/providers/kinesis-firehose/utils/getEventBuffer.ts
```diff
@@ -23,7 +29,7 @@ const eventBufferMap: Record<
 	string,
 	EventBuffer<KinesisFirehoseBufferEvent>
 > = {};
-const cachedClients: Record<string, FirehoseClient> = {};
+const cachedClients: Record<string, [FirehoseClient, AWSCredentials]> = {};
```
jimblanc, line 32: Need to cache credentials in-memory alongside the associated client instance in order to determine if they changed. SDK clients do not provide a synchronous method for fetching their configured credentials.

erinleigh90: Makes sense to me

erinleigh90 (dismissed):
LGTM!

cshfang (commented):
General comment - is it at all possible (and what are the tradeoffs if so) to simply not cache clients at all?

Also seems kind of evident we could benefit from making these providers a bit more DRY to avoid having to modify the same thing in multiple places but that's probably outside of the scope of this PR

jimblanc:
> General comment - is it at all possible (and what are the tradeoffs if so) to simply not cache clients at all?
> 
> Also seems kind of evident we could benefit from making these providers a bit more DRY to avoid having to modify the same thing in multiple places but that's probably outside of the scope of this PR

We could avoid caching any clients by just creating a new one each time we have events to send, but I think the performance & memory benefits from not creating/destroying clients for each batch make it worthwhile. Agree with more DRY, but out of scope for sev-2 (and to reduce the risk surface of the fix).

packages/analytics/src/providers/kinesis-firehose/utils/getEventBuffer.ts
```diff
@@ -76,18 +82,28 @@ export const getEventBuffer = ({
 	userAgentValue,
 }: KinesisFirehoseEventBufferConfig): EventBuffer<KinesisFirehoseBufferEvent> => {
 	const sessionIdentityKey = [region, identityId].filter(id => !!id).join('-');
+	const cachedClient = cachedClients[sessionIdentityKey];
```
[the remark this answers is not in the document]
jimblanc: > Does this need to be tied to per-client? I.e. is it possible for one cached client to become stale due to credential changes without the others to also?

Potentially, the reason that these maps exist is to ensure that all events are transmitted if the identity ID changes, which may have different sets of credentials.

packages/analytics/src/providers/kinesis/utils/getEventBuffer.ts
```diff
@@ -67,19 +73,29 @@ export const getEventBuffer = ({
...
-	if (!eventBufferMap[sessionIdentityKey]) {
+	if (!eventBufferMap[sessionIdentityKey] || credentialsHaveChanged) {
 		const getKinesisClient = (): IAnalyticsClient<KinesisBufferEvent> => {
-			if (!cachedClients[sessionIdentityKey]) {
-				cachedClients[sessionIdentityKey] = new KinesisClient({
-					credentials,
-					region,
-					customUserAgent: userAgentValue,
-				});
+			if (!cachedClient || credentialsHaveChanged) {
```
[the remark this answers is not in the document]
jimblanc: Should be able to use the destructured client ref, updated the other providers to be consistent

packages/analytics/src/providers/kinesis-firehose/utils/getEventBuffer.ts
jimblanc: Commented in thread below

packages/analytics/src/providers/kinesis-firehose/utils/getEventBuffer.ts
```diff
@@ -82,17 +84,17 @@ export const getEventBuffer = ({
 	userAgentValue,
 }: KinesisFirehoseEventBufferConfig): EventBuffer<KinesisFirehoseBufferEvent> => {
 	const sessionIdentityKey = [region, identityId].filter(id => !!id).join('-');
-	const cachedClient = cachedClients[sessionIdentityKey];
+	const [ cachedClient, cachedCredentials ] = cachedClients[sessionIdentityKey] ?? [];
```
erinleigh90, line 87: Much easier to read!

erinleigh90 (dismissed):
Looks even better to me!

packages/analytics/src/providers/kinesis-firehose/utils/getEventBuffer.ts
```diff
@@ -76,18 +84,28 @@ export const getEventBuffer = ({
...
+						region,
+						credentials,
+						customUserAgent: userAgentValue,
+					}),
+					credentials
+				];
 			}
 
-			const firehoseClient = cachedClients[sessionIdentityKey];
+			const firehoseClient = cachedClients[sessionIdentityKey][0];
```
cshfang, line 108: Clarity nit again:
```suggestion
			const [ firehoseClient ] = cachedClients[sessionIdentityKey];
```

cshfang (dismissed):
Just a nit but I think this lgtm

cshfang (approved):

erinleigh90 (approved):

PR #12795, opened 2024-01-04
test(datastore): Add OptimisticConcurrency tests

stocaaro:
#### Description of changes
Outcomes:
- Update graphql service fake to mimic OptimisticConcurrency(OCC) conflict resolution
- Add OptimisticConcurrency (OCC) tests for all cases covered by the existing Automerge tests
- Update service call cont expectation to reflect multiple event types, provide more rich feedback for failures

By comparing sample app results to test expectations and iterating until they align.

There is existing odd behavior where errors occur at in low latency conditions where the retry message includes null values for fields, which overwrites external updates. I will do sample app testing to confirm that this is consistent with the real world behavior.

#### Related changes

3. https://github.com/aws-amplify/amplify-js/pull/12740
4. :seedling:  **test(datastore): Add OptimisticConcurrency tests**

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -690,4 +691,681 @@ describe('DataStore sync engine', () => {
...
+					 * the second client's update to `blogId` "reverts" the primary
+					 * client's changes. This is because the external update takes
+					 * effect while the primary client's updates are still in flight.
+					 * By the time the primary client's update reaches the service,
+					 * the `_version` has changed. As a result, auto-merge chooses
+					 * the existing server-side `title` over the proposed updated
+					 * `title`. The following two tests demonstrate this behavior,
+					 * with a difference in the timing of the external request,
+					 * ultimately resulting in different final states.
+					 */
```
david-mcafee, line 1089: This comment only applies to the auto-merge tests, and should be removed for the OCC tests.

```suggestion
```

david-mcafee (changes requested):
Overall this looks great! I do think removing the incorrect comment is a blocker, since it would lead to significant confusion (at least for me!) down the line.

stocaaro:
@david-mcafee - Appreciate the suggestions.  I've accepted both suggested changes and have re-read all of the comments to make sure nothing else thats coming across is auto-merge centric.

david-mcafee (dismissed):
LGTM! 🎉

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
```diff
@@ -690,4 +691,669 @@ describe('DataStore sync engine', () => {
...
+						title: 'post title 2',
+					});
+				});
+				test("no input delay, high latency where we don't wait for the create to clear the outbox", async () => {
+					const postHarness = await harness.createPostHarness(
+						{
+							title: 'original title',
+							blogId: 'blog id',
+						},
+						false
```
svidgen, line 779: I'm guessing this is the "we don't wait for the create to clear the outbox" part of the test. Uncertain whether this warrants some clarification in the tests, but on first read of the tests above, I found myself puzzling over where/how where converse of this happening. It *might* be worth making this field required and "naming" it.

Or, probably better, just using the same `await harness.fullSettle()` call in the converse cases if that does what it looks like it does.

If you agree with some extra explicitness here, I'd probably defer the change to a followup after we're done resolving the core issue.

packages/datastore/__tests__/helpers/fakes/graphqlService.ts
```diff
@@ -474,6 +490,29 @@ export class FakeGraphQLService {
...
+				...this.populatedFields(updated),
+				_version: updated._version + 1,
+				_lastChangedAt: new Date().getTime(),
+				updatedAt: new Date().toISOString(),
+			};
+		} else {
+			merged = {
+				...this.populatedFields(existing),
+			};
+		}
```
svidgen, line 511: This `else` branch looks like a no-op. Since the fake is branching on `canOccMerge()` first anyway, I think I'd prefer that this `throw` to alert us that "we shouldn't be here", unless I'm failing to see or grok a use-case for this branch.

I think this is the only area of the PR that concerns me at the moment. Unless I'm just having a slow brain day, can we either turn this into a `throw` or add a comment explaining the branch?

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
stocaaro: Yeah. Prior to my work there was a line or two to settle each of the tests (except for the 2 tests where we don't settle after create).

I had considered making this a named arg, but settled on this with a description in the function comment.
image

I would agree that in PR, this looks very abstract, where in editor, it's pretty well documented for posterity.

There are now 32 tests, of which 4 skip settle after create.  I like the factoring of explicitly disabling the settling behavior instead of injecting it.

Given this conect, thoughts/preferences?

packages/datastore/__tests__/helpers/fakes/graphqlService.ts
stocaaro: Rereading this, I think this branch will never be reached since I have extracted the `canOccMerge` function, we're essentially doing the version check twice. I'll get a refactor out around this.

packages/datastore/__tests__/conflictResolutionBehavior.test.ts
svidgen: Given the docstrings/tooltips, let's defer refactoring unless it notably impedes future PR-ing.

svidgen (approved):

iartemiev (approved):

PR #12802, opened 2024-01-05
fix(auth): improve error handling while refreshing auth tokens

israx:
#### Description of changes

Main changes: 
 - improves error handling when a user wants to refresh a session while authenticated with `implicit grant` oauth flow.
 - dispatches a `tokenRefresh_failure` event when the refresh_token is expired. 

- add unit tests
- smoked test- changed the expiration date of the `refresh_token` in the Cognito console and was able to see that the `refreshToken_failure` event was dispatched 

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/auth/src/errors/constants.ts
```diff
@@ -25,3 +25,4 @@ export const invalidOriginException = new AuthError({
 	recoverySuggestion: 'Please call signInWithRedirect from the same origin.',
 });
 export const OAUTH_SIGNOUT_EXCEPTION = 'OAuthSignOutException';
+export const TOKEN_REFRESH_EXCEPTION = 'TOKEN_REFRESH_EXCEPTION';
```
Samaritan1011001, line 28: Should the value follow the same pattern as above? `TokenRefreshException`

packages/auth/src/providers/cognito/utils/types.ts
```diff
@@ -53,15 +54,28 @@ export function assertIdTokenInAuthTokens(
 	}
 }
 
+export const oAuthTokenRefreshException = new AuthError({
+	name: TOKEN_REFRESH_EXCEPTION,
+	message: `Token refresh is not supported once authenticated with 'implicit grant'(token) oauth flow.
```
Samaritan1011001, line 59: Do we know for sure this is the only time we don't have have refresh_token with us? Just making sure since we are suggesting a very specific flow type is the cause here.

packages/auth/__tests__/providers/cognito/refreshToken.test.ts
```diff
@@ -13,103 +16,175 @@ describe('refreshToken', () => {
...
+					AuthParameters: { REFRESH_TOKEN: 'refreshtoken' },
+					ClientId: 'aaaaaaaaaaaa',
+					UserContextData: { EncodedData: 'abcd' },
+				})
+			);
+			window['AmazonCognitoAdvancedSecurityData'] = undefined;
+		});
+	});
+
+	describe('negative cases', () => {
```
Samaritan1011001, line 121: In all the test cases, should we checking for the failure hub event to be dispatched as well along with the throwing of the error?

israx: the refreshToken doesn't dispatches hub events. Probably we can test that use case in the token orchestrator

packages/auth/src/providers/cognito/utils/types.ts
israx: Yeah good point. As this code is only triggered while refreshing cognito tokens, then it is expected that the sign-in flow would've returned a `refresh_token` , and if the user authenticated with `code grant` flow, then a `refresh_token` is expected too

Samaritan1011001 (dismissed):

ashika112 (approved):
LGTM

packages/aws-amplify/package.json
```diff
@@ -384,7 +384,7 @@
 			"name": "[Auth] confirmSignIn (Cognito)",
 			"path": "./dist/esm/auth/index.mjs",
 			"import": "{ confirmSignIn }",
-			"limit": "25.70 kB"
+			"limit": "25.89 kB"
```
ashika112, line 387: Question: why are we increasing the limits in this PR?

israx: some APIs depend on `fetchAuthSession` and adding additional code might result in a bundle size bump

packages/auth/src/providers/cognito/utils/types.ts
jimblanc:
```suggestion
	message: `Token refresh is not supported when authenticated with the 'implicit grant' (token) oauth flow. 
```

nit

jimblanc (approved):
lgtm, one nit

jimblanc (approved):

Samaritan1011001 (approved):

packages/auth/src/providers/cognito/utils/types.ts
```diff
@@ -113,3 +127,16 @@ export interface OAuthStore {
 	clearOAuthInflightData(): Promise<void>;
 	clearOAuthData(): Promise<void>;
 }
+function isAuthenticated(tokens?: CognitoAuthTokens | null) {
+	return tokens?.accessToken && tokens?.idToken;
```
elorzafe, line 131: This should be `||`

PR #12845, opened 2024-01-12
chore(auth): debounce refreshAuthTokens

israx:
#### Description of changes

The purpose of this PR is to debounce sync concurrent calls to the `fetchAuthSession` API. This approach is similar to the current debouncing logic [implemented](https://github.com/aws-amplify/amplify-js/blob/v5-stable/packages/auth/src/Auth.ts#L1878) in v5

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


AllanZhengYP (approved):
Good to see timer is removed. Thank you!

packages/core/src/utils/deDupCallback.ts
```diff
@@ -0,0 +1,34 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * returns inflight promise if there hasn't been resolved yet
+ *
+ * @param callback - callback to be deDup.
+ * @returns - the return type of the callback
+ */
+export const deDupCallback = <F extends (...args: any[]) => any>(
```
HuiSF, line 10: I think the naming could be `dedupeAsyncRequests` as now the callback is expected to return a promise (so can update the callback interface, too) that can be repeatedly awaited for. (dedupe is a more legit word I think, too 😅), wdyt?

packages/datastore/src/sync/processors/subscription.ts
```diff
@@ -279,7 +279,7 @@ class SubscriptionProcessor {
 
 				try {
 					// retrieving current token info from Cognito UserPools
-					const session = await await fetchAuthSession();
```
HuiSF, line 282: 😬

packages/core/__tests__/utils/debounceCallback.test.ts
```diff
@@ -0,0 +1,46 @@
+import { debounceCallback } from '../../src/utils/debounceCallback';
```
HuiSF, line 1: probably need an update 😅

packages/core/__tests__/utils/debounceCallback.test.ts
```diff
@@ -0,0 +1,46 @@
...
+		const debouncedCallback = debounceCallback(mockServiceCallback);
+		for (let i = 0; i < numberOfConcurrentCalls; i++) {
+			debouncedCallback();
+		}
+
+		await debouncedCallback();
+
+		debouncedCallback();
+		expect(mockServiceCallback).toHaveBeenCalledTimes(2);
+	});
```
HuiSF, line 36: This test may not be accurate - the `mockServiceCallback` called times should increase after the promised returned by it resolves, and when `debouncedCallback` gets called again, but in this test, the `mockServiceCallback` doesn't even return a promise.

HuiSF (commented):
Ty for improving the impl.

packages/core/src/utils/deDupCallback.ts
```diff
@@ -0,0 +1,34 @@
...
+			});
+		}
+
+		try {
+			return await inflightPromise;
+		} finally {
+			inflightPromise = undefined;
+		}
+	};
+};
```
HuiSF, line 34: nit: thinking this way of using the generic can make ts run slightly faster around this function. wdyt?
```suggestion
export const deDupCallback = <F extends (...args: any[]) => any>(
	callback: F
) => {
	let inflightPromise: Promise<Awaited<ReturnType<F>>> | undefined;
	return async (...args: Parameters<F>): Promise<Awaited<ReturnType<F>>> => {
		if (inflightPromise) return inflightPromise;

		if (!inflightPromise) {
			inflightPromise = new Promise(async (resolve, reject) => {
				try {
					const result = await callback(...args);
					resolve(result);
				} catch (error) {
					reject(error);
				}
			});
		}

		try {
			return await inflightPromise;
		} finally {
			inflightPromise = undefined;
		}
	};
};
```

packages/core/__tests__/utils/debounceCallback.test.ts
israx: the `mockServiceCallback` doesn't return a promise but the `deDupe` function does because it creates a promise on top of the `callback` passed. What I'm trying to do is to call the `deDuped` function concurrently and then calling it again when it was able to **resolve**.

packages/core/src/utils/deDupeAsyncRequests.ts
```diff
@@ -0,0 +1,42 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// this will make the tsc-complience-test to pass
```
cshfang, line 4: Nit:
```suggestion
// this will make the tsc-compliance-test to pass
```

packages/core/__tests__/utils/deDupeAsyncRequests.test.ts
```diff
@@ -0,0 +1,50 @@
...
+		mockServiceCallback.mockImplementation(async () => {});
+	});
+	afterEach(() => {
+		mockServiceCallback.mockClear();
+	});
+
+	it('should allow to invoke the callback when there is no concurrent calls', async () => {
+		const debouncedCallback = deDupeAsyncRequests(mockServiceCallback);
+
+		debouncedCallback();
```
cshfang, line 17: Nit: A bunch of these still refer to "debounce" and should be updated to "dedupe" for clarity

cshfang (dismissed):

package.json
```diff
@@ -114,7 +114,7 @@
 		"tslint-config-airbnb": "^5.8.0",
 		"typedoc": "^0.17.0",
 		"typescript": "^4.3.5",
-		"typescript-coverage-report": "^0.6.4",
+		"typescript-coverage-report": "^0.8.0",
```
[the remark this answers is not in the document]
israx: nope, I can have another PR for it

packages/core/src/utils/deDupeAsyncRequests.ts
```diff
@@ -0,0 +1,42 @@
...
+				} catch (error) {
+					reject(error);
+				}
+			});
+		}
+
+		try {
+			return await inflightPromise;
+		} finally {
+			inflightPromise = undefined;
```
HuiSF, line 39: Should this be add to the try-catch block inside the Promise impl.?

israx: yeah good point, I will test this

packages/core/src/utils/deDupeAsyncRequests.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @returns - the return type of the callback
+ */
+export const deDupeAsyncRequests = <A extends any[], R>(
+	callback: (...args: A) => Promise<R>
+) => {
+	let inflightPromise: Promise<Awaited<R>> | undefined;
+	return async (...args: A): Promise<Awaited<R>> => {
+		if (inflightPromise) return inflightPromise;
+
+		if (!inflightPromise) {
```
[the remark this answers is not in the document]
israx: good catch

AllanZhengYP (dismissed):

packages/core/src/utils/deDupeAsyncFunction.ts
```diff
@@ -10,33 +10,29 @@ type Awaited<T> = T extends null | undefined
 			: never // the argument to `then` was not callable
 	  : T; //
 /**
- * returns inflight promise if there hasn't been resolved yet
+ * returns in-flight promise if there is one
  *
- * @param callback - callback to be deDup.
+ * @param callback - callback to be deduped.
```
cshfang, line 15: Doesn't match the actual param

packages/core/src/utils/deDupeAsyncFunction.ts
```diff
@@ -10,33 +10,29 @@ type Awaited<T> = T extends null | undefined
...
  * @returns - the return type of the callback
  */
-export const deDupeAsyncRequests = <A extends any[], R>(
-	callback: (...args: A) => Promise<R>
+export const deDupeAsyncFunction = <A extends any[], R>(
+	fun: (...args: A) => Promise<R>
```
cshfang, line 19: Nit: `asyncFunction`?

packages/core/src/utils/deDupeAsyncFunction.ts
```diff
@@ -10,33 +10,29 @@ type Awaited<T> = T extends null | undefined
...
+				inflightPromise = undefined;
+			}
+		});
 
-		try {
-			return await inflightPromise;
-		} finally {
-			inflightPromise = undefined;
-		}
+		return await inflightPromise;
```
cshfang, line 36: Nit:
```suggestion
		return inflightPromise;
```

cshfang (approved):

elorzafe (approved):
Looks good,

nit: why not have a unit for refreshAuthTokens calls are deduped?

iartemiev (approved):

Issue #12776, opened 2024-01-01
Amplify Datastore not working when I start app without internet


digvijaysinh-york:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

DataStore

### Amplify Categories

auth

### Environment information

<details>

```
# Put output below this line
System:
    OS: macOS 14.1.1
    CPU: (8) arm64 Apple M2
    Memory: 103.50 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.2.0 - /opt/homebrew/bin/node
    npm: 10.2.3 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 120.0.6099.129
    Safari: 17.1
  npmPackages:
    @aws-amplify/datastore-storage-adapter: ^2.1.5 => 2.1.9 
    @aws-amplify/react-native: ^1.0.5 => 1.0.9 
    @babel/core: ^7.20.0 => 7.23.6 
    @babel/preset-env: ^7.20.0 => 7.23.6 
    @babel/runtime: ^7.20.0 => 7.23.6 
    @react-native-async-storage/async-storage: ^1.21.0 => 1.21.0 
    @react-native-community/netinfo: ^11.1.1 => 11.2.1 
    @react-native/eslint-config: ^0.72.2 => 0.72.2 
    @react-native/metro-config: ^0.72.11 => 0.72.11 
    @tsconfig/react-native: ^3.0.0 => 3.0.2 
    @types/react: ^18.0.24 => 18.2.46 
    @types/react-test-renderer: ^18.0.0 => 18.0.7 
    HelloWorld:  0.0.1 
    aws-amplify: ^6.0.5 => 6.0.9 
    aws-amplify/adapter-core:  undefined ()
    aws-amplify/analytics:  undefined ()
    aws-amplify/analytics/kinesis:  undefined ()
    aws-amplify/analytics/kinesis-firehose:  undefined ()
[...]

PR #12325, opened 2023-10-17
Reenable dataastore unit tests

elorzafe:
#### Description of changes

Reenabling datastore unit tests

#### Issue #, if available

There are 4 tests that are presenting issues compared to v5

splitted `DataStore.ts` tests into different modules per `describe` section
```
Test Suites: 29 passed, 29 total
Tests:       18 skipped, 1112 passed, 1130 total
Snapshots:   8 passed, 8 total
Time:        146.002s, estimated 246s
```

v5 has 1116 tests
#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #12861, opened 2024-01-18
fix(data): update DataStore to send correct Control Messages when starting

david-mcafee:
#### Description of changes

Fixes a DataStore bug where starting DataStore offline in a React Native app causes DataStore to hang indefinitely (and not work).

The control messages were originally updated [here](https://github.com/aws-amplify/amplify-js/pull/12025) when removing the Node check, and reworked [here](https://github.com/aws-amplify/amplify-js/pull/12325) to get DataStore tests passing (however, this introduced the above-mentioned bug).

## This PR:
| Environment  | Control Message |
| ------------- | ------------- |
| browser / RN / et al  | `SYNC_ENGINE_STORAGE_SUBSCRIBED`  |
| Node  | `SYNC_ENGINE_SYNC_QUERIES_READY` |

## Before this PR:
| Environment  | Control Message |
| ------------- | ------------- |
| browser  | `SYNC_ENGINE_STORAGE_SUBSCRIBED`  |
| Node / RN / et al  | `SYNC_ENGINE_SYNC_QUERIES_READY` |

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/12776

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/datastore/src/datastore/datastore.ts
+++ b/packages/datastore/src/datastore/datastore.ts
@@ -1561,9 +1561,10 @@ class DataStore {
 								// In Node, we need to wait for queries to be synced to prevent returning empty arrays.
 								// In non-Node environments (the browser or React Native), we can begin returning data
 								// once subscriptions are in place.
-								const readyType = isNode
-									? ControlMessage.SYNC_ENGINE_SYNC_QUERIES_READY
-									: ControlMessage.SYNC_ENGINE_STORAGE_SUBSCRIBED;
+								const readyType =
+									isNode && !isBrowser()
+										? ControlMessage.SYNC_ENGINE_SYNC_QUERIES_READY
+										: ControlMessage.SYNC_ENGINE_STORAGE_SUBSCRIBED;
 
 								if (type === readyType) {
 									this.initResolve();
--- a/packages/datastore/src/datastore/datastore.ts
+++ b/packages/datastore/src/datastore/datastore.ts
@@ -72,7 +72,7 @@ import {
 } from '../types';
 // tslint:disable:no-duplicate-imports
 import type { __modelMeta__ } from '../types';
-import { isNode } from '../util';
+import { isNode } from './utils';
 
 import {
 	DATASTORE,
@@ -1563,7 +1563,6 @@ class DataStore {
 									? ControlMessage.SYNC_ENGINE_SYNC_QUERIES_READY
 									: ControlMessage.SYNC_ENGINE_STORAGE_SUBSCRIBED;
 
-								// debugger;
 								if (type === readyType) {
 									this.initResolve();
 								}
--- a/packages/datastore/src/datastore/datastore.ts
+++ b/packages/datastore/src/datastore/datastore.ts
@@ -72,6 +72,7 @@ import {
 } from '../types';
 // tslint:disable:no-duplicate-imports
 import type { __modelMeta__ } from '../types';
+import { isNode } from '../util';
 
 import {
 	DATASTORE,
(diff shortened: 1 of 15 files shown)


packages/datastore/src/datastore/datastore.ts
```diff
@@ -1553,10 +1553,17 @@ class DataStore {
...
 						.subscribe({
 							next: ({ type, data }) => {
-								// In the Browser, we can begin returning data once subscriptions are in place.
-								const readyType = isBrowser()
-									? ControlMessage.SYNC_ENGINE_STORAGE_SUBSCRIBED
-									: ControlMessage.SYNC_ENGINE_SYNC_QUERIES_READY;
+								const isNode =
+									typeof process !== 'undefined' &&
+									process.versions != null &&
+									process.versions.node != null;
```
svidgen, line 1559: 1. Are there other places where `isBrowser()` is used?
    1. If not, let's change or remove the util function.
    2. If so, let's put work on the appropriate backlogs please to confirm those branches behave correctly for RN/Expo.
2. Should we drop this into an `isNode()` util function?
3. Do we need/want to consider a single util that returns a more complex environment attributes descriptor? (e.g., `{isNode, isNative}`)
4. Bonus points if you can find a node docs or GH issue link we can put here to show that this is a "canonical" way to confirm we're in a server-side/`node` environment.

packages/datastore/src/datastore/datastore.ts
```diff
@@ -1553,10 +1553,17 @@ class DataStore {
...
+
+								// In Node, we need to wait for queries to be synced to prevent returning empty arrays.
+								// In non-Node environments (the browser or React Native), we can begin returning data
+								// once subscriptions are in place.
```
svidgen, line 1563: Thanks for adding a comment!

svidgen (approved):
Thanks, David. A few small suggestion if they make sense to you and can be done quickly. Otherwise, let's just get this shipped and into customers' hands ASAP!

svidgen:
> Thanks, David. A few small suggestion if they make sense to you and can be done quickly. Otherwise, let's just get this shipped and into customers' hands ASAP!

Oh, I didn't see we had failing tests now. That's strange, this fix makes a ton of sense. Maybe the tests are just spoofing environmental stuff assuming the old util is in place.

david-mcafee:
> > Thanks, David. A few small suggestion if they make sense to you and can be done quickly. Otherwise, let's just get this shipped and into customers' hands ASAP!
> 
> Oh, I didn't see we had failing tests now. That's strange, this fix makes a ton of sense. Maybe the tests are just spoofing environmental stuff assuming the old util is in place.

Yep, that was the issue - `isNode` and `isBrowser` were both returning `true` in a test environment. I've updated the check so the tests will pass, just revalidating that change against my React Native sample.

packages/datastore/src/datastore/datastore.ts
```diff
@@ -1553,10 +1553,18 @@ class DataStore {
...
+								const readyType =
+									isNode && !isBrowser()
+										? ControlMessage.SYNC_ENGINE_SYNC_QUERIES_READY
+										: ControlMessage.SYNC_ENGINE_STORAGE_SUBSCRIBED;
```
svidgen, line 1567: I think this latest revision makes this conditional confusing. It doesn't look like it should matter. "Why doesn't `isNode === !isBrowser()`?"

We should either:

1. Remove the conditional entirely, because we don't officially support DataStore SSR in v6
2. Make `isNode()` a util that the tests can mock.

I would stronger favor option 2 for this PR with a followup to discuss the consequences of removing the check entirely. E.g., if a customers attempts to use DataStore in node, despite our us not supporting it, what's the best way we can fail for customers?

packages/datastore/src/datastore/datastore.ts
david-mcafee: 1. We are using this elsewhere, so we'll want to take a look at what changed in the two PRs linked in the description to validate that nothing else is broken. I'll put an item in our backlog.
2. Done
3. I think it makes sense to stick with the `isBrowser` util exported from `core`, unless further refactoring warrants that. If it were possible to check that we were in a RN environment in some way, I think that would make sense. Given that there is not, I would vote we leave as is.
4. I did some digging on this previously, and couldn't find a "canonical" answer of any kind. This check existed in DataStore prior to the V6 PRs linked above, so at least this is a "tried-and-tested" method in our library.

packages/datastore/src/datastore/datastore.ts
david-mcafee: Went with option 2, and I've added a spike to the backlog for further investigation. As part of that investigation, I'd like to also take a look at _all_ instances of `isNode` that were removed / the current usage of `isBrowser` to ensure that no other defects were introduced.

svidgen (approved):

PR #12922, opened 2024-01-29
fix: Lambda auth config value

iartemiev:
#### Description of changes

Both CLI & Gen2 emit Lambda auth config for GraphQL as `AWS_LAMBDA`, but we were looking for `LAMBDA` when parsing the config object. As a result the library was not respecting the default auth mode when it was configured as Lambda.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


stocaaro (approved):
Is there a testing story around this parsing logic that we might extend to cover this change so we don't have future regressions around this?

stocaaro (approved):

svidgen (approved):

PR #12930, opened 2024-01-31
fix(core): Amplify.configure dispatches Hub event with unparsed config object

HuiSF:

* unit testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/src/singleton/Amplify.ts
+++ b/packages/core/src/singleton/Amplify.ts
@@ -72,7 +72,7 @@ export class AmplifyClass {
 			'core',
 			{
 				event: 'configure',
-				data: resourcesConfig,
+				data: this.resourcesConfig,
 			},
 			'Configure',
 			AMPLIFY_SYMBOL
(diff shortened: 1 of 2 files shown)


packages/core/src/singleton/Amplify.ts
```diff
@@ -72,7 +72,7 @@ export class AmplifyClass {
 			'core',
 			{
 				event: 'configure',
-				data: resourcesConfig,
+				data: this.resourcesConfig,
```
jimblanc, line 75: Is this technically a breaking change?

HuiSF: Not really, this change is for fixing when call `.configure()` on the Amplify singleton imported directly from core (internal usage).

When calling `.configure` on the Amplify singleton proxy imported from `aws-amplify`, the resourceConfig param that reaches the core singleton, is always parsed.

jimblanc: What if customers are using the data from the `configure` event though?

HuiSF: > What if customers are using the data from the `configure` event though?

If they are using `Amplify` imported from `aws-amplify` (which should be the case as we don't document to use `Amplify` imported from core), the `configure` event gets the data as `ResourceConfig`.

jimblanc: Got it, thanks for confirming. Just wanted to check!

jimblanc (approved):

kvramyasri7 (approved):

PR #12935, opened 2024-02-01
fix(api-graphql): generate client.models has no properties

HuiSF:
#### Description of changes

`Amplify.configure()` and `generateClient()` calls order can be randomly bundled by script bundler when they are located in separate files, as from the code path perspective they don't have a dependent relationship, however, `generateClient()` is actually depending on `Amplify.configure()` to get a valid GraphQL provider configuration. To resolve this, this PR added functionality to `generateClient` as the following:

* When `generateClient` gets called, if it can get a valid GraphQL provider configuration, it generates the properties for `client.models`.
* Otherwise, it sets up a listener to the `core` Hub event: `configure`, when the listener receives a valid GraphQL provider configuration, it generates the properties for `client.models`. The listener modifies `client.models` by reference.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/api-graphql/src/internals/generateClient.ts
```diff
@@ -34,7 +36,50 @@ export function generateClient<T extends Record<any, any> = never>(
...
+>(clientRef: any) => {
+	Hub.listen('core', coreEvent => {
+		const { event, data } = coreEvent.payload;
+
+		if (event !== 'configure') {
+			return;
+		}
+
+		// data is guaranteed to be `ResourcesConfig` when the event is `configure`
+		const resourceConfig = data as ResourcesConfig;
```
svidgen, line 65: I think it would be slick if we start building up a library of small type-guard utils that canonically categorize objects like this.

E.g., something like:

```typescript
if (!isConfigPayload(event)) return;

function isConfigEvent(payload: any): payload is { event: 'configure', data: ResourcesConfig } {
  return payload.event === 'configure';
}
```

packages/api-graphql/src/internals/generateClient.ts
```diff
@@ -34,7 +36,50 @@ export function generateClient<T extends Record<any, any> = never>(
...
+
+	if (!config.API?.GraphQL) {
+		client.models = emptyModels as ModelTypes<never>;
+		generateModelsPropertyOnAmplifyConfigure<T>(client);
+	} else {
+		client.models = generateModelsProperty<T>(
+			client,
+			config.API?.GraphQL,
+		);
+	}
```
svidgen, line 49: In the case where we register the listener, it looks like it subscribes to config changes perpetually. I'm curious if this breaks expectations if a customer intends to create a client and then re-configure another one with another API.

On the other hand, if we don't intend to support that pattern, it makes sense to subscribe to config updates *either way*, IMO:

1. Generate models, defaulting to empty object if not configured.
1. Start listening for config changes.

An unlikely edge, I imagine. But I'm eager for clear-cut definition of the intended behavior that reconciles the differences.

packages/api-graphql/src/internals/generateClient.ts
```diff
@@ -34,7 +36,68 @@ export function generateClient<T extends Record<any, any> = never>(
...
+}
+
+const emptyModels = new Proxy(
+	{},
+	{
+		get() {
+			throw new Error(
+				'Could not generate client. This is likely due to Amplify.configure() not being called prior to generateClient().',
+			);
+		},
```
iartemiev, line 101: > script bundlers may randomly arrange their orders in the production bundle

Is it possible to have a race condition here, causing the error to get thrown because a customer accesses `client.models.*` before the `configure` hub event fired? 

For example, I call `Amplify.configure()` with a valid config in one file and then in another:
```ts
const client = generateClient<Schema>();
const data = await client.models.Post.list();
``` 
If the bundler arranges the files in such a way that the above snippet gets called first, will the customer see an error thrown?

iartemiev (approved):
LGTM. Left 1 nit and 1 question; neither are blocking.

packages/api-graphql/src/internals/generateClient.ts
HuiSF: The race condition is going to happen when someone makes back-to-back calls exactly like this at the module level. For calls made at the module level, we would need to do the following in the same file to maintain the order of calls:

```ts
Amplify.configure();
const client = generateClient<Schema>();
const data = await client.models.Post.list();
```

Let's discuss.

iartemiev: Sounds good. Let's summarize in the PR description:
1. Which usage patterns does this fix resolve?
2. What are the remaining sharp edges?

iartemiev (approved):

packages/api-graphql/src/internals/generateClient.ts
iartemiev: Discussed offline. Approving

stocaaro (approved):

PR #12967, opened 2024-02-06
feat(api-graphql): add model index query support

iartemiev:
#### Description of changes

Runtime implementation for [this feature](https://github.com/aws-amplify/amplify-api-next/pull/90).

```ts
// Given this gen2 schema
const schema = a.schema({
  Post: a
    .model({
      title: a.string().required(),
      description: a.string(),
      viewCount: a.integer(),
      updatedAt: a.string(),
    })
    .secondaryIndexes([ a.index('title').sortKeys(['viewCount']) ])
});

// We can now perform the following index query using Amplify JS:
const { data: posts } = await client.models.Post.listByTitleAndViewCount(
  { title: 'My Post' viewCount: {gt: 4} }
);
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/src/internals/generateModelsProperty.ts
```diff
@@ -67,11 +71,46 @@ export function generateModelsProperty<T extends Record<any, any> = never>(
...
+					};
+				});
+
+			return idxs || [];
+		};
+
+		const secondaryIdxs = getSecondaryIndexesFromSchemaModel(model);
+
+		for (const idx of secondaryIdxs) {
+			models[name][idx.queryField] = indexQueryFactory(
```
stocaaro, line 107: I was surprised TS is cool with this, but it looks like models is `any` above and specified by the return type here, which doesn't give our customers any type interface protection, does it?  How do we get guarantees that all of the typed index methods from the type will be populated at runtime?

packages/api-graphql/src/internals/operations/indexQuery.ts
```diff
@@ -0,0 +1,164 @@
...
+		return _indexQuery(client, modelIntrospection, model, indexMeta, {
+			...args,
+			...options,
+		});
+	};
+
+	return context ? indexQueryWithContext : indexQuery;
+}
+
+async function _indexQuery(
```
stocaaro, line 67: This function is ~100 lines and carries a lot of different responsibilities. Could we decompose it into a few well names pieces?

packages/api-graphql/src/internals/operations/indexQuery.ts
```diff
@@ -0,0 +1,164 @@
...
+					};
+				}
+			}
+
+			return {
+				data: data[key],
+				nextToken: data[key].nextToken,
+				extensions,
+			};
+		}
```
stocaaro, line 154: Is this the only place where we restructure graphql results like this? It reads like something we would have in a couple places and should reuse by importing/calling here.

packages/api-graphql/src/internals/generateModelsProperty.ts
```diff
@@ -67,11 +71,46 @@ export function generateModelsProperty<T extends Record<any, any> = never>(
...
+
+		const getSecondaryIndexesFromSchemaModel = (model: SchemaModel) => {
+			const idxs = model.attributes
+				?.filter(
+					attr =>
+						attr.type === 'key' &&
+						// presence of `name` property distinguishes GSI from primary index
+						attr.properties?.name &&
+						attr.properties?.queryField &&
+						attr.properties?.fields.length > 0,
```
stocaaro, line 88: This should live in a private function to guard `attr`.

stocaaro (commented):
I'm torn on this one. There are two lines I'm tempted to hold here.
1. We can and should decompose functions down into well named, single responsibility private functions to increase the readability.
2. Casting `any` into a type gives us no protection and this code is touching the piece that is making promises it isn't guaranteeing.

I'm inclined to hold the line on the 1 and accept that 2 is a bigger issue that deserves individual treatment, but both seem worth discussion / alignment.

packages/api-graphql/src/internals/operations/indexQuery.ts
iartemiev: Updated

packages/api-graphql/src/internals/generateModelsProperty.ts
iartemiev: Updated

packages/api-graphql/src/internals/operations/indexQuery.ts
```diff
@@ -0,0 +1,182 @@
...
+			};
+		} else {
+			const initialized = modelInitializer(flattenedResult);
+
+			return {
+				data: initialized,
+				nextToken: data[key].nextToken,
+				extensions,
+			};
+		}
```
stocaaro, line 95: Would a ternary be lighter without losing clarity?
```suggestion
		return {
			data: selectionSet ? flattenedResult : modelInitializer(flattenedResult),
			nextToken: data[key].nextToken,
			extensions,
		};
```

stocaaro (approved):
Appreciate the updates. One nit suggestion, but otherwise happy to carry on.

stocaaro (approved):

jimblanc (approved):
Approving changes to `aws-amplify` and `core`.

Issue #12890, opened 2024-01-24
Ability to use https redirectSignOutUrls when using amplify v6


cranberyxl:
### Is this related to a new or existing framework?

React Native

### Is this related to a new or existing API?

Authentication

### Is this related to another service?

Cognito

### Describe the feature you'd like to request

When using a federated SSO client we need to be able to signout of the upstream provider inside of the webview. TO do this we want to be able to call something like https://upstreamSSOProvider.com/logout?post_logout_redirect_uri=mobile%3A%2F%2F, however https urls are rejected on [react native](https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts#L10). The upstream logout prodivder will ultimately redirect back to the `mobile://` url as the cognito logout would do. 

### Describe the solution you'd like

The ability to override the `invalidRedirectException` thrown [here](https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts#L10), using the provided signout uri

### Describe alternatives you've considered

1. Stay on v5 where we can use our own webview implementation
2. Find a way to use the rtn-browser directly to do this

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

cwomack:
Heyllo, @cranberyxl 👋 and thank you for opening this issue.  Can you confirm that you're only seeing this issue when you upgraded to v6, and if you downgrade back to v.5x that the problem goes away?

cranberyxl:
The problem goes away in v5 only because we can use a custom webview solution like `react-native-inappbrowser-reborn` and as far as I can tell, the no https restriction was added in v6

cwomack:
@cranberyxl, thank you for the confirmation.  I'll mark this as a bug then and review this with the team internally.  We'll comment back with updates or further questions as we make progress.

elorzafe:
[cranberyxl](https://github.com/cranberyxl) when you configured SSO provider on Amazon Cognito you dont need to go to the sign out endpoint manually, you can configure that directly on Cognito. Have you tried that?

More information [on Cognito docs](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-saml-idp-sign-out.html)

cranberyxl:
[...]

Issue #12956, opened 2024-02-05
Throws InvalidRedirectException when attempting social sign in on Capacitor app


borgoat:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Angular

### Amplify APIs

Authentication, Analytics, GraphQL API, Storage

### Amplify Version

v6

### Amplify Categories

auth

### Backend

CDK

### Environment information

<details>

```
  System:
    OS: macOS 14.2.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 3.89 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.10.0 - ~/.asdf/installs/nodejs/20.10.0/bin/node
    npm: 10.2.3 - ~/.asdf/plugins/nodejs/shims/npm
    pnpm: 8.11.0 - ~/.asdf/installs/nodejs/20.10.0/bin/pnpm
    bun: 0.5.7 - ~/.bun/bin/bun
  Browsers:
    Chrome: 121.0.6167.139
    Safari: 17.2.1
  npmPackages:
    @angular-devkit/build-angular: 17.1.0 => 17.1.0 
    @angular-devkit/core: 17.1.0 => 17.1.0 
    @angular-devkit/schematics: 17.1.0 => 17.1.0 
    @angular-eslint/eslint-plugin: 17.2.0 => 17.2.0 
    @angular-eslint/eslint-plugin-template: 17.2.0 => 17.2.0 
    @angular-eslint/template-parser: 17.2.0 => 17.2.0 
    @angular/animations: 17.1.0 => 17.1.0 
    @angular/cdk: 17.1.0 => 17.1.0 
    @angular/cdk-experimental: 17.1.0 => 17.1.0 
    @angular/cli: 17.1.0 => 17.1.0 
    @angular/common: 17.1.0 => 17.1.0 
    @angular/compiler: 17.1.0 => 17.1.0 
    @angular/compiler-cli: 17.1.0 => 17.1.0 
[...]

PR #12969, opened 2024-02-07
fix(auth): Allow Web Urls on signInWithRedirect in Mobile Apps

elorzafe:
#### Description of changes

Adds fallback URL in case customer wants to redirect native app to a website

#### Issue #, if available

fixes: https://github.com/aws-amplify/amplify-js/issues/12890

React-Native CLI App
#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
```diff
@@ -10,7 +10,12 @@ export function getRedirectUrl(redirects: string[]): string {
 			!redirect.startsWith('http://') && !redirect.startsWith('https://')
 	);
 	if (!redirect) {
-		throw invalidRedirectException;
-	}
+		if (redirects.length === 0) {
+			throw invalidRedirectException;
+		} else {
+			// fallsback to available URL https://github.com/aws-amplify/amplify-js/issues/12890
```
Samaritan1011001, line 16:
```suggestion
			// fallsback to available redirect URL. Addresses: https://github.com/aws-amplify/amplify-js/issues/12890
```
or something on those lines 😅

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
```diff
@@ -10,7 +10,12 @@ export function getRedirectUrl(redirects: string[]): string {
...
+			return redirects[0];
```
Samaritan1011001, line 17: Always return the first url? What if they want to call a particular url instead?

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
```diff
@@ -10,7 +10,12 @@ export function getRedirectUrl(redirects: string[]): string {
...
+		}
+	} 
 	return redirect;
```
Samaritan1011001, line 20: These changes will still return the `mobile` url right? 
Ex: Giving ["mobile://", "http://"]  as input the output is still "mobile://" url
In the issue linked, it seems like they want to hit "https://" first then "mobile://"

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
elorzafe: how a customer would select that?

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
elorzafe: This will still choose first `mobile` url because is filtered. Now if you want to force the web url, that seems something that could be address by an API change

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
kvramyasri7: +2

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
jimblanc: Correct, with the current API there's no way to actively select which URL gets used. We (along with other platforms) use a heuristic to determine which URL to use (filtering out dev URLs, etc.).

jimblanc (commented):
Logically makes sense to me. Working on testing the change now and will hold approval until then. One small nit.

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
kvramyasri7: By re-directing to the right url is understanding please correct if I am wrong @elorzafe 😕

daochaojian:
Facing same error. I think this PR is not enough to fix this issue. The  function 'openAuthSessionIOS'  from [here](https://github.com/aws-amplify/amplify-js/blob/528a9842c99978e7317ef893f55298f38c97bfc3/packages/rtn-web-browser/src/apis/openAuthSessionAsync.ts#L31C7-L31C25) will filter all https and http redirectUrl.
Maybe this code will make it work:
`
	const redirectUrl = redirectUrls.find(
		// take the first non-web url as the deeplink
		item => !item.startsWith('https://') && !item.startsWith('http://'),
	) ?? redirectUrls?.[0]; // return first url when there is no custom url scheme like myApp://
`
I didn't know much background knowledge about this repository. Just hope it helps.

Samaritan1011001:
Feature has been released in `aws-amplify@6.6.0` , you can now provide a redirectUrl when signing out.

Documentation is now available - 
[Gen2](https://docs.amplify.aws/react/build-a-backend/auth/concepts/external-identity-providers/#redirect-urls)
[Gen1](https://docs.amplify.aws/gen1/react/build-a-backend/auth/add-social-provider/#redirect-urls-1)

5t33:
@Samaritan1011001 This has not been fixed, as mentioned by @borgoat here:
https://github.com/aws-amplify/amplify-js/pull/12969#issuecomment-1948600131 

The fix is for https://github.com/aws-amplify/amplify-js/issues/12890 _not_ https://github.com/aws-amplify/amplify-js/issues/12956

Issue #8384, opened 2021-06-03
Potential bug: Calling signOut() before currentAuthenticatedUser() does not log the user out on react-native app restart


bertrand-caron:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

Authentication

### Amplify Categories

auth

### Environment information

<details>

```
# Put output below this line

  System:
    OS: macOS 11.4
    CPU: (8) arm64 Apple M1
    Memory: 95.45 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.0 - ~/.nvm/versions/node/v14.16.0/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.16.0/bin/yarn
    npm: 6.14.11 - ~/.nvm/versions/node/v14.16.0/bin/npm
  Browsers:
    Chrome: 91.0.4472.77
    Firefox: 88.0.1
    Safari: 14.1.1
  npmPackages:
    @aws-amplify/api: 4.0.1 => 4.0.1 
    @aws-amplify/auth: 4.0.1 => 4.0.1 
    @aws-amplify/core: 4.0.1 => 4.0.1 (4.0.3)
    @babel/core: ^7.8.4 => 7.12.16 (7.14.3)
    @babel/runtime: ^7.8.4 => 7.12.13 (7.14.0)
    @fortawesome/fontawesome-svg-core: ^1.2.35 => 1.2.35 
    @fortawesome/free-regular-svg-icons: ^5.15.3 => 5.15.3 
    @fortawesome/free-solid-svg-icons: ^5.15.3 => 5.15.3 
    @fortawesome/react-native-fontawesome: ^0.2.6 => 0.2.6 
    @onfido/react-native-sdk: ^1.3.4 => 1.3.4 
    @react-native-async-storage/async-storage: ^1.14.1 => 1.14.1 
    @react-native-community/datetimepicker: ^3.4.2 => 3.4.2 
[...]

PR #13005, opened 2024-02-13
fix(auth): storage not synced before accessing when the app boots from killed state

Samaritan1011001:
#### Description of changes

**[V5 only bug fix]**
We sync the storage before it's read so when the LastAuthUser is accessed it's reading the latest values. 

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/8384

1. Linked the changed library to RN sample app
2. Sign in a user
3. Killed the app with swipe up
4. Open the app
5. Sign out (dispatches a Hub event an inspection of local storage shows the user is deleted)
6. Call currentAuthenticatedUser (observe there is no user)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


elorzafe (dismissed):

packages/auth/src/Auth.ts
```diff
@@ -2155,16 +2155,23 @@ export class AuthClass {
...
 			logger.debug('failed to clear cached items');
 		}
 
 		if (this.userPool) {
-			const user = this.userPool.getCurrentUser();
-			if (user) {
-				await this.cognitoIdentitySignOut(opts, user);
+			try {
+				await this._storageSync;
+			} catch (e) {
```
HuiSF, line 2168: Isn't the `e` already an `Error` object so `new Error()` is not needed?

packages/auth/src/Auth.ts
```diff
@@ -2155,16 +2155,23 @@ export class AuthClass {
...
+				logger.debug('Failed to sync cache info into memory', e);
+				throw new Error(e);
+			}
+			storedCurrUser = this.userPool.getCurrentUser();
+			if (storedCurrUser) {
+				await this.cognitoIdentitySignOut(opts??{}, storedCurrUser);
```
HuiSF, line 2174: nit
```suggestion
				await this.cognitoIdentitySignOut(opts ?? {}, storedCurrUser);
```

packages/auth/src/Auth.ts
Samaritan1011001: good call, I used it from another place. Throwing e now like in other places again.

packages/auth/src/Auth.ts
```diff
@@ -2155,16 +2155,23 @@ export class AuthClass {
...
-			const user = this.userPool.getCurrentUser();
-			if (user) {
-				await this.cognitoIdentitySignOut(opts, user);
+			try {
+				await this._storageSync;
+			} catch (e) {
+				logger.debug('Failed to sync cache info into memory', e);
+				throw e;
+			}
+			storedCurrUser = this.userPool.getCurrentUser();
```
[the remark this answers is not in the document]
Samaritan1011001: If is used outside of this if block to dispatch hub event. Previously it emitted with `this.user` which was unavailable when the app booted from a killed state. But i'll revert that change to keep the fix minimal.

packages/auth/src/Auth.ts
```diff
@@ -2155,16 +2155,23 @@ export class AuthClass {
 	 * @return - A promise resolved if success
 	 */
 	public async signOut(opts?: SignOutOpts): Promise<any> {
+		let storedCurrUser;
```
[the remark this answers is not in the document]
Samaritan1011001: because it comes from the storage but based on the other comment, I'll simply keep what was before.

elorzafe (approved):

jimblanc (dismissed):
MSTM

packages/auth/src/Auth.ts
```diff
@@ -2154,17 +2154,23 @@ export class AuthClass {
...
+			try {
+				await this._storageSync;
+			} catch (e) {
+				logger.debug('Failed to sync cache info into memory', e);
+				throw e;
+			}
 			const user = this.userPool.getCurrentUser();
 			if (user) {
-				await this.cognitoIdentitySignOut(opts, user);
+				await this.cognitoIdentitySignOut(opts ?? {}, user);
```
ashika112, line 2173: do you need this? since you have default val at L2157.

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -3680,7 +3713,7 @@ describe('auth unit test', () => {
 			);
 		});
 
-		test.only('User Pools and Identity Pools', async () => {
```
jimblanc, line 3683: Oop 😅

jimblanc (approved):

ashika112 (approved):

PR #13013, opened 2024-02-14
fix(core): eslint migration caused wrong CustomScope type

HuiSF:
#### Description of changes

The `CustomScope` should be a string `type CustomScope = string & {}` was used to ensure `OAuthScope` won't resolve as a `string`.

The eslint migration required change of using `{}`, and the change was wrong.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/src/singleton/Auth/types.ts
+++ b/packages/core/src/singleton/Auth/types.ts
@@ -185,7 +185,9 @@ interface CustomProvider {
 	custom: string;
 }
 
-type CustomScope = string & Record<string, string>;
+// CustomScope is a non-nullish string, using `& NonNullable<unknown>` to ensure
+// `OAuthScope` is not resolved as a `string`
+type CustomScope = string & NonNullable<unknown>;
 export type OAuthScope =
 	| 'email'
 	| 'openid'


jimblanc (approved):

ashika112 (approved):
Good Catch.

packages/core/src/singleton/Auth/types.ts
```diff
@@ -185,7 +185,9 @@ interface CustomProvider {
 	custom: string;
 }
 
-type CustomScope = string & Record<string, string>;
+// CustomScope is a non-nullish string, using `& NonNullable<unknown>` to ensure
+// `OAuthScope` is not resolved as a `string`
+type CustomScope = string & NonNullable<unknown>;
```
cshfang, line 190: Thoughts on a util type? It seems like we'll run into this at least a couple more times with the Auth category.

cshfang (approved):

packages/core/src/singleton/Auth/types.ts
HuiSF: Agreed, will add and refactor when migrate the auth package accordingly.

PR #13029, opened 2024-02-19
Custom queries mutations

svidgen:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/api-graphql/__tests__/fixtures/modeled/schema.ts
+++ b/packages/api-graphql/__tests__/fixtures/modeled/schema.ts
@@ -70,6 +70,64 @@ const schema = a.schema({
 	CommunityPollVote: a
 		.model({ id: a.id().required() })
 		.authorization([a.allow.public('apiKey'), a.allow.owner()]),
+
+	// custom queries and mutations
+	EchoResult: a.customType({
+		resultContent: a.string().required(),
+	}),
+
+	// custom query returning a non-model type
+	echo: a
+		.query()
+		.arguments({
+			argumentContent: a.string().required(),
+		})
+		.returns(a.ref('EchoResult'))
+		.function('echoFunction')
+		.authorization([a.allow.public()]),
+
+	// custom query returning a primitive type
+	echoString: a
+		.query()
+		.arguments({
+			inputString: a.string().required(),
+		})
+		.returns(a.string())
+		.function('echoFunction')
+		.authorization([a.allow.public()]),
+
+	// custom mutation returning a non-model type
+	PostLikeResult: a.customType({
+		likes: a.integer().required(),
+	}),
+	likePost: a
+		.mutation()
+		.arguments({
+			postId: a.id().required(),
+		})
+		.returns(a.ref('PostLikeResult'))
+		.function('echoFunction')
+		.authorization([a.allow.private()]),
+
+	// custom mutation returning a model type
+	// NOTE: I'm not sure we can/should actually support this.
+	// What are the implications of returning a model? What if it has
+	// relationships? When/Where are those resolved? Does this work today?
+	// TODO: Need to test this case directly.
+	Post: a
+		.model({
+			id: a.id().required(),
+			content: a.string(),
+		})
+		.authorization([a.allow.public('apiKey'), a.allow.owner()]),
+	listPostReturnPost: a
+		.mutation()
+		.arguments({
+			postId: a.id().required(),
+		})
+		.returns(a.ref('Post'))
+		.function('echoFunction')
+		.authorization([a.allow.private()]),
 });
 
 export type Schema = ClientSchema<typeof schema>;
--- a/packages/api-graphql/src/internals/generateClient.ts
+++ b/packages/api-graphql/src/internals/generateClient.ts
@@ -3,6 +3,10 @@
 import { graphql, cancel, isCancelError } from './v6';
 import { generateModelsProperty } from './generateModelsProperty';
 import {
+	generateCustomQueriesProperty,
+	generateCustomMutationsProperty,
+} from './generateCustomOperationsProperty';
+import {
 	V6Client,
 	__amplify,
 	__authMode,
@@ -38,8 +42,8 @@ export function generateClient<T extends Record<any, any> = never>(
 	} as any;
 
 	client.models = generateModelsProperty<T>(client, params);
-	client.queries = generateQueriesProperty<T>(client, params);
-	// client.queries = generateMutationsProperty<T>(client, params);
+	client.queries = generateCustomQueriesProperty<T>(client, params);
+	client.mutations = generateCustomMutationsProperty<T>(client, params);
 
 	return client as V6Client<T>;
 }
--- a/packages/api-graphql/src/internals/generateClient.ts
+++ b/packages/api-graphql/src/internals/generateClient.ts
@@ -10,6 +10,7 @@ import {
 	__headers,
 } from '../types';
 import { ClientGenerationParams } from './types';
+import { generateQueriesProperty } from './generateQueriesProperty';
 
 /**
  * @private
--- a/packages/api-graphql/src/internals/generateClient.ts
+++ b/packages/api-graphql/src/internals/generateClient.ts
@@ -32,9 +32,13 @@ export function generateClient<T extends Record<any, any> = never>(
 		cancel,
 		isCancelError,
 		models: {},
+		queries: {},
+		mutations: {},
 	} as any;
 
 	client.models = generateModelsProperty<T>(client, params);
+	client.queries = generateQueriesProperty<T>(client, params);
+	// client.queries = generateMutationsProperty<T>(client, params);
 
 	return client as V6Client<T>;
 }
--- a/packages/api-graphql/src/internals/generateCustomOperationsProperty.ts
+++ b/packages/api-graphql/src/internals/generateCustomOperationsProperty.ts
@@ -1,0 +1,92 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import {
+	ModelTypes,
+	CustomQueries,
+	CustomMutations,
+} from '@aws-amplify/data-schema-types';
+import { graphQLOperationsInfo, ModelOperation } from './APIClient';
+import { ClientGenerationParams } from './types';
+import { V6Client, __authMode, __authToken } from '../types';
+
+import { customOpFactory } from './operations/custom';
+import { ModelIntrospectionSchema } from '@aws-amplify/core/internals/utils';
+
+type OpTypes = 'queries' | 'mutations';
+
+type CustomOpsProperty<
+	T extends Record<any, any>,
+	OpType extends OpTypes,
+> = OpType extends 'queries' ? CustomQueries<T> : CustomMutations<T>;
+
+const operationTypeMap = {
+	queries: 'query',
+	mutations: 'mutation',
+} as const;
+
+export function generateCustomOperationsProperty<
+	T extends Record<any, any>,
+	OpType extends OpTypes,
+>(
+	client: V6Client<Record<string, any>>,
+	params: ClientGenerationParams,
+	operationsType: OpType
+): OpType extends 'queries' ? CustomQueries<T> : CustomMutations<T> {
+	const config = params.amplify.getConfig();
+
+	if (!config.API?.GraphQL) {
+		// breaks compatibility with certain bundler, e.g. Vite where component files are evaluated before
+		// the entry point causing false positive errors. Revisit how to better handle this post-launch
+
+		// throw new Error(
+		// 	'The API configuration is missing. This is likely due to Amplify.configure() not being called
+		// prior to generateClient().'
+		// );
+		return {} as CustomOpsProperty<T, OpType>;
+	}
+
+	const modelIntrospection: ModelIntrospectionSchema | undefined =
+		config.API.GraphQL.modelIntrospection;
+
+	if (!modelIntrospection) {
+		return {} as CustomOpsProperty<T, OpType>;
+	}
+
+	const ops = {} as CustomOpsProperty<T, OpType>;
+	for (const operation of Object.values(modelIntrospection[operationsType])) {
+		// There's no common ancestry between CustomOpsProperty (CustomQueries, CustomMutations)
+		// and the model introspection schema, and no way to ensure the types *actually* match.
+		// They *should* match as long as the customer is using types from the gen2 schema builder
+		// and the config generated from gen2.
+		(ops as any)[operation.name] = customOpFactory(
+			client,
+			modelIntrospection,
+			operationTypeMap[operationsType],
+			operation
+		);
+	}
+
+	return ops;
+}
+
+export function generateCustomMutationsProperty<T extends Record<any, any>>(
+	client: V6Client<Record<string, any>>,
+	params: ClientGenerationParams
+) {
+	return generateCustomOperationsProperty<T, 'mutations'>(
+		client,
+		params,
+		'mutations'
+	);
+}
+
+export function generateCustomQueriesProperty<T extends Record<any, any>>(
+	client: V6Client<Record<string, any>>,
+	params: ClientGenerationParams
+) {
+	return generateCustomOperationsProperty<T, 'queries'>(
+		client,
+		params,
+		'queries'
+	);
+}
--- a/packages/api-graphql/src/internals/operations/custom.ts
+++ b/packages/api-graphql/src/internals/operations/custom.ts
@@ -1,0 +1,414 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { AmplifyServer } from '@aws-amplify/core/internals/adapter-core';
+import {
+	initializeModel,
+	generateGraphQLDocument,
+	buildGraphQLVariables,
+	flattenItems,
+	authModeParams,
+	ModelOperation,
+	getCustomHeaders,
+} from '../APIClient';
+import {
+	AuthModeParams,
+	ClientWithModels,
+	GraphQLOptionsV6,
+	GraphQLResult,
+	ListArgs,
+	QueryArgs,
+	V6Client,
+	V6ClientSSRRequest,
+} from '../../types';
+import {
+	ModelIntrospectionSchema,
+	SchemaModel,
+	SchemaNonModel,
+	CustomOperation,
+	CustomOperationArgument,
+} from '@aws-amplify/core/internals/utils';
+
+const OBJECT_TYPE_STRINGS = ['AWSJSON'] as const;
+const NUMERIC_TYPE_STRINGS = ['Int', 'Float', 'AWSTimeStamp'] as const;
+const BOOLEAN_TYPE_STRINGS = ['Boolean'] as const;
+const STRING_TYPE_STRINGS = [
+	'ID',
+	'String',
+	'AWSPhone',
+	'AWSDate',
+	'AWSTime',
+	'AWSDateTime',
+	'AWSEmail',
+	'AWSURL',
+	'AWSIPAddress',
+] as const;
+
+/**
+ * Experimental.
+ *
+ * Playing with the idea of using this as a look up for `typeof arg[f]` to
+ * validate whether the field is of the correct type.
+ *
+ * Problem is `null` and `undefined`, which are valid for any nullable field.
+ */
+const JS_TYPES_TO_GRAPHAL_TYPES = {
+	object: OBJECT_TYPE_STRINGS,
+	number: NUMERIC_TYPE_STRINGS,
+	boolean: BOOLEAN_TYPE_STRINGS,
+	string: STRING_TYPE_STRINGS,
+	symbol: [],
+	undefined: [],
+	function: [],
+	bigint: [],
+};
+
+type PrimitiveTypeString =
+	| StringTypeString
+	| NumericTypeString
+	| BooleanTypeString
+	| ObjectTypeString;
+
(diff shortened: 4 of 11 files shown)


packages/api-graphql/src/internals/operations/custom.ts
```diff
@@ -0,0 +1,324 @@
...
+					headers
+			  )) as GraphQLResult<any>)
+			: ((await (client as V6Client<Record<string, any>>).graphql(
+					{
+						...auth,
+						query,
+						variables,
+					},
+					headers
+			  )) as GraphQLResult<any>);
```
stocaaro, line 291: nit: This function is getting a bit long. A helper to manage the context vs no context graphql call might be an okay place to extract.

packages/api-graphql/src/internals/operations/custom.ts
```diff
@@ -0,0 +1,324 @@
...
+
+		// flatten response
+		if (data) {
+			const [key] = Object.keys(data);
+			const flattenedResult = flattenItems(data)[key];
+
+			// TODO: custom selection set
```
stocaaro, line 298: The selection set logic above doesn't cover this?

packages/api-graphql/src/internals/operations/custom.ts
```diff
@@ -0,0 +1,324 @@
...
+
+			return { data: initialized, extensions };
+		} else {
+			return { data: null, extensions };
+		}
+	} catch (error: any) {
+		if (error.errors) {
+			// graphql errors pass through
+			return error as any;
+		} else {
```
stocaaro, line 319: nit: Since the condition returns, this else isn't doing anything.

packages/api-graphql/src/types/index.ts
```diff
@@ -97,14 +99,13 @@ type PagedList<T, TYPENAME> = {
...
-		: T extends {}
-			? {
-					[K in keyof T]: WithListsFixed<T[K]>;
-				}
-			: T;
+type WithListsFixed<T> = T extends PagedList<infer IT, infer NAME>
+	? PagedList<Exclude<IT, null | undefined>, NAME>
+	: T extends {}
+	  ? {
+				[K in keyof T]: WithListsFixed<T[K]>;
```
stocaaro, line 106: Tabs and spaces living together. Also, classic double tab for emphasis.  Why isn't prettier doing things on your behalf here?

packages/api-graphql/src/types/index.ts
```diff
@@ -120,12 +121,14 @@ type NeverEmpty<T> = {
...
-				[K in keyof T]-?: WithListsFixed<Exclude<T[K], null | undefined>>;
-			}
-		: T;
+type FixedQueryResult<T> = Exclude<
+	T[keyof T],
+	null | undefined
+> extends PagedList<any, any>
+	? {
+			[K in keyof T]-?: WithListsFixed<Exclude<T[K], null | undefined>>;
+	  }
```
stocaaro, line 130: ?

packages/api-graphql/src/types/index.ts
```diff
@@ -238,16 +241,15 @@ export type UnknownGraphQLResponse =
...
-			? IN
-			: TYPED_GQL_STRING extends GeneratedSubscription<infer IN, any>
-				? IN
-				: FALLBACK_TYPES extends GraphQLOperationType<infer IN, any>
-					? IN
-					: any;
+> = TYPED_GQL_STRING extends GeneratedQuery<infer IN, any>
+	? IN
+	: TYPED_GQL_STRING extends GeneratedMutation<infer IN, any>
+	  ? IN
```
stocaaro, line 247: ?

stocaaro: Different from the above, it seems like you're using spaces after the initial tab. You've revised towards this, but application of this is inconsistent in this change. Consistency?

stocaaro (commented):
I made a pass. Have a few nits and a couple more substantive comments.

packages/api-graphql/src/types/index.ts
stocaaro: Meant more re: Tabs and spaces, although `]-?` isn't something I recognize.

packages/api-graphql/__tests__/fixtures/modeled/schema.ts
```diff
@@ -87,6 +87,68 @@ const schema = a.schema({
 			a.index('title'),
 			a.index('description').sortKeys(['viewCount']),
 		]),
+
+	// custom queries and mutations
```
david-mcafee, line 91: Nit:
```suggestion
	//#region Custom queries and mutations
```

packages/api-graphql/__tests__/internals/generateClient.test.ts
```diff
@@ -5173,4 +5175,245 @@ describe('generateClient', () => {
...
+
+			const client = generateClient<Schema>({
+				amplify: Amplify,
+			});
+			const result = await client.queries.echoString({
+				inputString: 'echo argumentContent value',
+			});
+
+			expect(normalizePostGraphqlCalls(spy)).toMatchSnapshot();
+			expect(result?.data).toMatchSnapshot();
```
david-mcafee, line 5238: Also a nit, but any reason not to check for `echo result content` instead of matching a snapshot, like you've done in the other tests?

packages/api-graphql/__tests__/internals/generateClient.test.ts
```diff
@@ -5173,4 +5175,245 @@ describe('generateClient', () => {
...
+				}
+			});
+			const result = await client.queries.echo({
+				argumentContent: 'echo argumentContent value',
+			});
+
+			expect(normalizePostGraphqlCalls(spy)).toMatchSnapshot();
+		});
+
+		test('includes call level headers', async () => {
```
david-mcafee, line 5353: Thoughts on additionally testing that headers passed to `Amplify.configure` are present?

packages/api-graphql/__tests__/internals/generateClient.test.ts
```diff
@@ -5173,4 +5175,245 @@ describe('generateClient', () => {
...
+				authToken: 'my-auth-token'
+			});
+			const result = await client.queries.echo({
+				argumentContent: 'echo argumentContent value',
+			});
+
+			expect(normalizePostGraphqlCalls(spy)).toMatchSnapshot();
+		});
+
+		test('uses call site authMode', async () => {
```
david-mcafee, line 5397: Similarly, thoughts on testing that the Amplify configuration auth mode is passed through?

packages/api-graphql/src/internals/generateClient.ts
```diff
@@ -26,7 +35,7 @@ import { isConfigureEventWithResourceConfig } from './utils/isConfigureEventWith
  * @returns
  */
 export function generateClient<T extends Record<any, any> = never>(
-	params: ClientGenerationParams,
+	params: ClientGenerationParams
```
david-mcafee, line 38: Nit - looks like you have trailing commas disabled in VS Code?
```suggestion
	params: ClientGenerationParams,
```

packages/api-graphql/src/types/index.ts
```diff
@@ -184,7 +187,7 @@ export interface AWSAppSyncRealTimeProviderOptions {
 
 export type AWSAppSyncRealTimeProvider = {
 	subscribe(
-		options?: AWSAppSyncRealTimeProviderOptions,
+		options?: AWSAppSyncRealTimeProviderOptions
```
david-mcafee, line 190:
```suggestion
		options?: AWSAppSyncRealTimeProviderOptions,
```

david-mcafee (changes requested):
This PR looks good to me, and most of my comments are non-blocking nits. Non-blocking, but it looks like something is up with your Prettier configuration. The only truly blocking comment is for removal of the `console.log`.

packages/api-graphql/src/internals/generateCustomOperationsProperty.ts
```diff
@@ -0,0 +1,89 @@
...
+
+	if (!modelIntrospection) {
+		return {} as CustomOpsProperty<T, OpType>;
+	}
+
+	// digging operations out here to type guard
+	const operations = modelIntrospection[operationsType];
+	if (!operations) {
+		return {} as CustomOpsProperty<T, OpType>;
+	}
```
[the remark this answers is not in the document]
svidgen: I'll add some comments here. There are a few reasons this might fall through.

packages/api-graphql/src/internals/generateCustomOperationsProperty.ts
```diff
@@ -0,0 +1,89 @@
...
+
+	const ops = {} as CustomOpsProperty<T, OpType>;
+	for (const operation of Object.values(operations)) {
+		(ops as any)[operation.name] = customOpFactory(
```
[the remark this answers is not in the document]
svidgen: I think it's visually better to start with `{} as any`, which ends up being the same lie either way. The only difference in my mind that pushes me towards `{} as SomeIntendedType` with *more* casts later on is that it communicates intent of the variable and *forces* me to check the "yes, this is on purpose" box every time by casting.

packages/api-graphql/src/internals/operations/custom.ts
```diff
@@ -0,0 +1,324 @@
...
+}
+
+/**
+ * Checks whether a specific argument is defined on an arguments object.
+ *
+ * @param args Arguments object to check.
+ * @param argDef Argument definition to check for.
+ * @returns true if argument is present; false if not.
+ * @throws If argument is required and not present.
+ */
```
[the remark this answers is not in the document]
svidgen: Oh, I mostly just left things unannotated that were copy-pasta'd mostly-as-is from other parts of the codebase. With an exception or two, I documented the new, more distinct utils I added because I legitimately thought I'd need the reorientation the next time I looked at them.

*But*, it seems worthwhile to circle around with some more docstrings.

packages/api-graphql/src/internals/operations/custom.ts
```diff
@@ -0,0 +1,324 @@
...
+function hasValidArgument<ArgumentType extends CustomOperationArgument>(
+	args: any,
+	argDef: ArgumentType
+): args is Record<ArgumentType['name'], any> {
+	const argValue = args[argDef.name];
+	const argJSType = typeof argValue;
+	if (argJSType !== 'undefined') {
+		return true;
+	} else if (argDef.isRequired) {
+		throw new Error(`Argument ${argDef.name}: ${argDef.type} is required.`);
```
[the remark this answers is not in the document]
svidgen: On second though, some of this becomes trivial to inline after that separation. FWIW, part of the reason this was initially "un-golfed" is that I was working on more extensive argument type validation &mdash; I might pick that back up later. But, for now it seemed like a rabbit hole.

packages/api-graphql/src/internals/operations/custom.ts
```diff
@@ -0,0 +1,324 @@
...
+ * ```json
+ * "InputString: String!"
+ * ```
+ *
+ * @param operation Operation object from model introspection schema.
+ * @returns "outer" arguments string
+ */
+function outerArguments(operation: CustomOperation): string {
+	const args = Object.entries(operation.arguments)
+		.map(([k, v]) => {
```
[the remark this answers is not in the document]
svidgen: Maybe I'm having a slow brain day, but I'm at all not sure what you mean! But, please follow up with me if you have ideas here.

packages/api-graphql/src/internals/operations/custom.ts
svidgen: Copy-pasta from other builders. As a nit, I'll leave this alone for now. Though, I do agree this and the other builders are due for some decomposition.

packages/api-graphql/src/internals/operations/custom.ts
svidgen: I disagree! But, maybe the point you're making is that I don't have a test proving that these branches are distinct... so, I'll work on that!

packages/api-graphql/src/types/index.ts
svidgen: Weird ... my prettier has apparently been displeased with me and has given up.

image

svidgen: > That's it. You've written too much garbage. I quit.

-- Prettier in February of 2024 before breathing it's last.

packages/api-graphql/src/types/index.ts
svidgen: Yeah. I gotta try to fix prettier. I actually really don't know why it has bailed on me.

packages/api-graphql/src/types/index.ts
svidgen: Recurring theme. I'll try to fix prettier. I actually didn't notice it had quit on me.

packages/api-graphql/src/internals/operations/custom.ts
svidgen: Logic above is for default selection set only. Custom selection set requires both types work and selection set builder updates.

svidgen: Updating the comment to clarify.

packages/api-graphql/__tests__/fixtures/modeled/schema.ts
svidgen: Also requires an endregion, so applying separately. I'm not sure I wanted to regionalize this, but I don't have a strong opinion on it at the moment. I'll add it.

packages/api-graphql/__tests__/internals/generateClient.test.ts
svidgen: Just missed it. I started with snapshots across the board and missed this one. I'll fix it!

packages/api-graphql/src/internals/generateClient.ts
svidgen: Yeah. Prettier is aggressively unhappy with me. Downright combative at this point. Not sure what I'm missing...

packages/api-graphql/src/internals/generateClient.ts
```diff
@@ -38,13 +47,21 @@ export function generateClient<T extends Record<any, any> = never>(
...
 		models: emptyProperty as ModelTypes<never>,
 		enums: emptyProperty as EnumTypes<never>,
+		queries: emptyProperty as CustomQueries<never>,
+		mutations: emptyProperty as CustomMutations<never>,
 	} as any;
 
 	const apiGraphqlConfig = params.amplify.getConfig().API?.GraphQL;
 
 	if (isApiGraphQLConfig(apiGraphqlConfig)) {
+		console.log('here');
```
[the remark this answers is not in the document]
svidgen: 😅

packages/api-graphql/__tests__/internals/generateClient.test.ts
svidgen: Looking into this.

svidgen: I think this is actually covered by all the base cases, where configured auth mode and apikey are clearly being used.

packages/api-graphql/src/internals/operations/custom.ts
svidgen: Or did you mean to suggest this change?

```ts
if (error.errors) {
  return error as any;
}
throw error;
```

svidgen: Ok. I've added tests. Regardless of whether that's what you meant, it seemed pertinent.

I'm leaving the syntax for now. Just keeping patterns from which this was copy-pasta'd from for now.

packages/api-graphql/src/internals/generateCustomOperationsProperty.ts
```diff
@@ -0,0 +1,89 @@
...
+): OpType extends 'queries' ? CustomQueries<T> : CustomMutations<T> {
+	if (!config) {
+		// breaks compatibility with certain bundler, e.g. Vite where component files are evaluated before
+		// the entry point causing false positive errors. Revisit how to better handle this post-launch
+
+		// throw new Error(
+		// 	'The API configuration is missing. This is likely due to Amplify.configure() not being called
+		// prior to generateClient().'
+		// );
+		return {} as CustomOpsProperty<T, OpType>;
```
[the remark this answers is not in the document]
svidgen: Copy pasta from one of the other generate files. But, I've now updated comment to be more meaningful here.

packages/api-graphql/src/internals/operations/custom.ts
```diff
@@ -0,0 +1,324 @@
...
+ *   //    ^? const s: string
+ * }
+ * ```
+ *
+ * @param o Object to inspect
+ * @param field Field to look for
+ * @returns Boolean: `true` if the `o[field]` is a `string`
+ */
+function hasStringField<Field extends string>(
+	o: any,
```
HuiSF, line 104: Any possibility not using `any`? JS codebase is in the progress of migration to stricter linter rules where `any` is forbidden.

For example, at this particular spot, the type of `o` could be `Record<unknown, unknown>`?

svidgen: The amount of casting we're doing in general, including all the `any`'s is starting to feel somewhat egregious and a little risky. So, I agree with your general point. But, I think we should make an effort of it... I'm actually planning on spending a little time composing a sampling of the scary type/runtime mismatches that concern me, and we can plan something from there for the *general* case.

For this particular spot, I've taken a couple passes at changing the `any` to something less broad, and it's more of a riddle than it might be worth right now. It doesn't like `Record<unknown, unknown>`, and *similar* changes looks like they end up punting the `any` elsewhere and causing a rippling effect. It looks like it'll be a similar story with some of the other `any`'s laying around.

And for what it's worth, I don't actually think this particular instance is high risk, since the runtime logic in the guard aligns very narrowly with the type assurance it offer. It's largely *intended* to take "anything" and just tell you whether "the thing" you give it has a particular `[T]: string` field on it. Not a high risk `any`, IMO.

packages/api-graphql/__tests__/fixtures/modeled/schema.ts
david-mcafee: Also not a strong opinion - meant to add `//#eddregion` as a suggestion.

packages/api-graphql/src/internals/operations/custom.ts
svidgen: (I *am* taking a pass at clearing up the low hanging fruit on code I touched though. Stay tuned!)

packages/api-graphql/__tests__/internals/generateClient.test.ts
david-mcafee: I've added these these previously in other tests, if an example would help.

david-mcafee (dismissed):
LGTM!

packages/api-graphql/src/internals/operations/custom.ts
svidgen: Not as much wiggle room as I'd have liked on some of those any's. But, I squashed a *few* of them.

svidgen: Wait. Push failed ..

svidgen: Ok. Pushed. Sorry for the noise.

david-mcafee (dismissed):

packages/api-graphql/src/internals/generateClient.ts
```diff
@@ -38,13 +47,20 @@ export function generateClient<T extends Record<any, any> = never>(
...
 	const apiGraphqlConfig = params.amplify.getConfig().API?.GraphQL;
 
 	if (isApiGraphQLConfig(apiGraphqlConfig)) {
 		client.models = generateModelsProperty<T>(client, apiGraphqlConfig);
 		client.enums = generateEnumsProperty<T>(apiGraphqlConfig);
+		client.queries = generateCustomQueriesProperty<T>(client, apiGraphqlConfig);
+		client.mutations = generateCustomMutationsProperty<T>(
+			client,
+			apiGraphqlConfig,
+		);
```
iartemiev, line 63: Need to do something similar in the [server client](https://github.com/aws-amplify/amplify-js/blob/main/packages/api-graphql/src/internals/server/generateClientWithAmplifyInstance.ts), but we can do that in a subsequent PR

iartemiev (dismissed):

packages/api-graphql/src/internals/generateClient.ts
svidgen: I won't raise a fuss if someone merges this. But I'll otherwise plan on addressing in this pr early Monday

packages/api-graphql/src/internals/operations/custom.ts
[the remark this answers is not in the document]
stocaaro: Nah. Pretty sure this was my slow brain day. I was thinking it was starting with pairs, not tearing down an object.

stocaaro (approved):
LGTM

packages/api-graphql/src/internals/operations/custom.ts
stocaaro: I am in alignment with this. I would be a fan of adding as much strong typing as we can and where we can't having a dev mode that turns on runtime type validation so that we can run some amount of minimal interface testing that raises when types don't match expectations.

The two cases that are hard:
- Dynamically typed interfaces, like offering custom methods for retrieving certain graphql queries.
- JSON input parsing.

In both cases, we could document the type interface in a tool like Zod (or using the same strategies) and use the same interface for enforcing runtime type guarantees when it doesn't meet the type expectations.

iartemiev (approved):

stocaaro (approved):

PR #13073, opened 2024-03-01
Feat: DownloadData refactor

ashika112:
#### Description of changes
This PR adds support for the new `path` parameter as input for `downloadData` API

- Unit tests are added 
- Manually validated the usage in a react sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/storage/src/providers/s3/apis/downloadData.ts
```diff
@@ -42,32 +52,46 @@ import { logger } from '../../../utils';
...
-		const { bucket, keyPrefix, s3Config } = await resolveS3ConfigAndInput(
-			Amplify,
-			downloadDataOptions,
+	async (): Promise<
+		StorageDownloadDataOutput<StorageItemKey | StorageItemPath>
+	> => {
+		const { options: downloadDataOptions } = downloadDataInput;
+		const { bucket, keyPrefix, s3Config, identityId } =
+			await resolveS3ConfigAndInput(Amplify, downloadDataOptions);
+		const { inputType, objectKey } = validateStorageOperationInput(
```
jimblanc, line 87: Nit: Type guard might be a better/more consistent mechanism for doing this (e.g. isInputWithKey, etc)

packages/storage/src/providers/s3/types/inputs.ts
```diff
@@ -60,7 +64,13 @@ export type RemoveInput = StorageRemoveInput<RemoveOptions>;
...
  * Input type for S3 downloadData API.
  */
-export type DownloadDataInput = StorageDownloadDataInput<DownloadDataOptions>;
+export type DownloadDataInput = StrictUnion<
+	DownloadDataInputKey | DownloadDataInputPath
+>;
+export type DownloadDataInputKey =
+	StorageDownloadDataInputKey<DownloadDataOptionsKey>;
+export type DownloadDataInputPath =
+	StorageDownloadDataInputPath<DownloadDataOptions>;
```
jimblanc, line 73: Nit: Should this have `Path` at the end for consistency?

packages/storage/src/providers/s3/types/outputs.ts
```diff
@@ -8,12 +8,36 @@ import {
...
+	 * VersionId used to reference a specific version of the object.
+	 */
+	versionId?: string;
+	/**
+	 * A standard MIME type describing the format of the object data.
+	 */
+	contentType?: string;
+}
+
+export interface ItemPath extends StorageItemPath {
```
jimblanc, line 40: Duplicate symbol, also do we need to duplicate `Item` at all? Seems like the types are identical

jimblanc (commented):
Some nits & questions. Also, it looks like we have a handful of linting errors

packages/storage/src/providers/s3/apis/downloadData.ts
```diff
@@ -42,32 +52,46 @@ import { logger } from '../../../utils';
  * }
  *```
  */
-export const downloadData = (input: DownloadDataInput): DownloadDataOutput => {
+
+interface DownloadData {
+	(input: DownloadDataInputPath): DownloadDataOutputPath;
+	(input: DownloadDataInputKey): DownloadDataOutputKey;
```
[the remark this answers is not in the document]
ashika112: yet to add those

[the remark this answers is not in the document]
ashika112: Added them

packages/storage/src/providers/s3/types/inputs.ts
```diff
@@ -13,7 +16,8 @@ import {
 import {
 	CopyDestinationOptions,
 	CopySourceOptions,
-	DownloadDataOptions,
+	DownloadDataOptionsKey,
+	DownloadDataOptions as DownloadDataOptionsPath,
```
jimblanc, line 20: Nit: Why not just rename this symbol?

packages/storage/src/providers/s3/types/options.ts
```diff
@@ -18,11 +18,13 @@ interface CommonOptions {
 	useAccelerateEndpoint?: boolean;
 }
 
+/** @deprecated This will be removed in next major version. */
 type ReadOptions =
 	| { accessLevel?: 'guest' | 'private' }
 	| { accessLevel: 'protected'; targetIdentityId?: string };
 
 interface WriteOptions {
+	/** @deprecated This will be removed in next major version. */
```
jimblanc, line 27: Nit: Can we just deprecate the entire interface like we did with ReadOptions?

packages/storage/src/providers/s3/types/outputs.ts
```diff
@@ -5,15 +5,29 @@ import {
...
+/** @deprecated Use{@link ItemPath} instead. */
+export interface Item extends StorageItemKey {
+	/**
+	 * VersionId used to reference a specific version of the object.
+	 */
+	versionId?: string;
+	/**
+	 * A standard MIME type describing the format of the object data.
+	 */
+	contentType?: string;
```
jimblanc, line 27: Worth extracting out these common fields for Item & ItemPath like you did with `StorageItemBase`?

packages/storage/src/providers/s3/utils/storageInputValidation.ts
```diff
@@ -0,0 +1,39 @@
...
+		const { path } = input;
+
+		return {
+			inputType: STORAGE_INPUT_PATH,
+			objectKey: typeof path === 'string' ? path : path({ identityId }),
+		};
+	} else if (isInputWithKey(input)) {
+		return { inputType: STORAGE_INPUT_KEY, objectKey: input.key };
+	} else {
+		throw new Error('invalid input');
```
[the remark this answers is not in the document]
ashika112: good call

packages/storage/src/providers/s3/types/options.ts
ashika112: So only if it is put inside the interface you see the deprecation warning on the parameter level. The reason it is on top level in Read Options is because I cant do the same with union type.
REF: https://github.com/microsoft/tsdoc/issues/131

packages/storage/src/providers/s3/apis/downloadData.ts
```diff
@@ -52,22 +102,25 @@ export const downloadData = (input: DownloadDataInput): DownloadDataOutput => {
...
+		const { options: downloadDataOptions } = downloadDataInput;
+		const { bucket, keyPrefix, s3Config, identityId } =
+			await resolveS3ConfigAndInput(Amplify, downloadDataOptions);
+		const { inputType, objectKey } = validateStorageOperationInput(
+			downloadDataInput,
+			identityId,
 		);
-		const finalKey = keyPrefix + key;
+		const finalKey =
+			inputType === STORAGE_INPUT_KEY ? keyPrefix + objectKey : objectKey;
```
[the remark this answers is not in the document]
ashika112: Yeah I initially had it like that.But the way copy and list constructs the key is different so i pulled it out. Also, another reason i kept it here was for post deprecation. Did not want to extract the core logic out now to bring it back agin later.

packages/storage/src/providers/s3/apis/downloadData.ts
```diff
@@ -52,22 +102,25 @@ export const downloadData = (input: DownloadDataInput): DownloadDataOutput => {
 		},
 	});
 
-	return downloadTask;
+	return downloadTask as Output;
```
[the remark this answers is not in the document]
ashika112: Yes it is because of limitation from overloading through interface. TS is confused as to where it should place the return type between `DownloadDataOutputKey` & `DownloadDataOutputPath`.

packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
```diff
@@ -177,3 +178,128 @@ describe('downloadData', () => {
...
+	[
+		{
+			path: 'path',
+			expectedKey: 'path'
+		},
+		{
+			path: () => 'path',
+			expectedKey: 'path'
+		}
+	].forEach(({ path, expectedKey }) => {
```
HuiSF, line 217: nit: use `test.each` ([reference](https://jestjs.io/docs/api#testeachtablename-fn-timeout))

So we can avoid having tests with the same description on line 218

packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
```diff
@@ -177,3 +178,128 @@ describe('downloadData', () => {
...
+		await job();
+
+		expect(getObject).toHaveBeenCalledWith(
+			expect.anything(),
+			expect.objectContaining({
+				Range: `bytes=${start}-${end}`,
+			}),
+		);
+	});
+});
```
HuiSF, line 305: nit: new line at the end of a file

packages/storage/__tests__/providers/s3/apis/utils/storageInputValidation.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    const input = { key: 'testKey' };
+    const result = validateStorageOperationInput(input);
+    expect(result).toEqual({ inputType: STORAGE_INPUT_KEY, objectKey: 'testKey' });
+  });
+
+  it('should throw an error when input is invalid', () => {
+    const input = {invalid: 'test' } as any;
+    expect(() => validateStorageOperationInput(input)).toThrow('invalid input');
+  });
+});
```
HuiSF, line 30: nit: new line

packages/storage/src/providers/s3/types/options.ts
HuiSF: Can we also include "what you should use instead" in the deprecation message?

packages/storage/src/providers/s3/utils/storageInputValidation.ts
HuiSF, line 1: nit: file name should reflect the exported member (in this case `validateStorageOperationInput` is the only exported member)

packages/storage/src/types/options.ts
```diff
@@ -4,6 +4,7 @@
 import { StorageAccessLevel } from '@aws-amplify/core';
 
 export interface StorageOptions {
+	/** @deprecated This will be removed in next major version. */
```
HuiSF, line 7: Instead of "next major version" can we just say `v7` to make it clearer? wdyt?

packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
ashika112: retained the same format from test above. i like this better

packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
ashika112: curious: how is this not caught in linting 🤔

packages/storage/src/types/options.ts
jimblanc: IMO we should leave this open ended in-case we end up not removing it right away. I'd even say "may be removed"

packages/storage/src/providers/s3/types/options.ts
```diff
@@ -87,11 +89,14 @@ export type GetUrlOptions = ReadOptions &
 /**
  * Input options type for S3 downloadData API.
  */
-export type DownloadDataOptions = ReadOptions &
-	CommonOptions &
+export type DownloadDataOptions = CommonOptions &
 	TransferOptions &
 	BytesRangeOptions;
 
+/** @deprecated Use {@link DownloadDataOptions} instead. */
```
jimblanc, line 96: DownloadDataOptionsPath?

packages/storage/src/types/inputs.ts
```diff
@@ -7,6 +7,28 @@ import {
...
 
+/** @deprecated Use {@link StorageOperationInputPath} instead. */
+export interface StorageOperationInputKey {
+	/** @deprecated Use `path` instead. */
+	key: string;
+}
+export interface StorageOperationInputPath {
+	path: string | (({ identityId }: { identityId?: string }) => string);
+}
+export interface StorageOperationOptions<Options> {
```
jimblanc, line 18: Should this go in `options.ts`?

packages/storage/src/providers/s3/apis/downloadData.ts
```diff
@@ -4,45 +4,95 @@
...
 import { getObject } from '../utils/client';
 import { getStorageUserAgentValue } from '../utils/userAgent';
 import { logger } from '../../../utils';
+import {
+	StorageDownloadDataOutput,
+	StorageItemKey,
+	StorageItemPath,
+} from '../../../types';
+import { STORAGE_INPUT_KEY } from '../utils/constants';
+import { DownloadDataInputKey, DownloadDataInputPath } from '../types/inputs';
```
jimblanc, line 19: Should we export these from index.ts and import from there for consistency?

packages/storage/src/providers/s3/apis/downloadData.ts
```diff
@@ -4,45 +4,95 @@
...
+import {
+	DownloadDataOutputKey,
+	DownloadDataOutputPath,
+} from '../types/outputs';
```
jimblanc, line 23: Should we export these from index.ts and import from there for consistency?

packages/storage/src/providers/s3/types/options.ts
```diff
@@ -18,11 +18,13 @@ interface CommonOptions {
 	useAccelerateEndpoint?: boolean;
 }
 
+/** @deprecated This will be removed in next major version. */
 type ReadOptions =
 	| { accessLevel?: 'guest' | 'private' }
 	| { accessLevel: 'protected'; targetIdentityId?: string };
 
 interface WriteOptions {
```
ashwinkumar6, line 26: Should we add the depreciation flag in two places

1. **On top of the type**: strike-threw when the entire type is referenced elsewhere
2. **At the particular property**: strike-threw when the particular property is used. Especially with unions types

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const isInputWithPath = (input: Input): input is StorageOperationInputPath => {
+	return input.path !== undefined;
+};
+
+export const validateStorageOperationInput = (
+	input: Input,
+	identityId?: string,
+) => {
+	assertValidationError(
```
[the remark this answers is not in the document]
ashika112: While the assertion error is for vanilla JS, there is another check with typeguard at the top.

packages/storage/src/types/inputs.ts
ashika112: This was originally here. Also, it not really actual option type, it is part of input. I think it is best to keep it here. Maybe I can rename to `StorageOperationOptionInput` ?

packages/storage/src/providers/s3/types/options.ts
ashika112: I have added on comment above. There is [ongoing issue](https://github.com/microsoft/tsdoc/issues/131) on how union type can't have this and so we have only at top of types in handful of places for such case but generic rule would be use as see fit. Property level & top level together is preferable but if not achievable top level is ok.

packages/storage/src/providers/s3/types/inputs.ts
```diff
@@ -60,7 +64,14 @@ export type RemoveInput = StorageRemoveInput<RemoveOptions>;
...
  */
-export type DownloadDataInput = StorageDownloadDataInput<DownloadDataOptions>;
+export type DownloadDataInput = StrictUnion<
+	DownloadDataInputKey | DownloadDataInputPath
+>;
+/** @deprecated Use {@link DownloadDataInputPath} instead. */
+export type DownloadDataInputKey =
+	StorageDownloadDataInputKey<DownloadDataOptionsKey>;
+export type DownloadDataInputPath =
+	StorageDownloadDataInputPath<DownloadDataOptionsPath>;
```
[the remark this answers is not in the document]
ashika112: These seem too exhaustive for our case. I would just keep it on top level.

packages/storage/src/providers/s3/apis/downloadData.ts
ashika112: Yes. I see this is missed but all need to come from `index.ts`

packages/storage/src/types/options.ts
ashika112: I will update with wording from product and it will be final.

packages/storage/src/providers/s3/apis/downloadData.ts
```diff
@@ -52,22 +100,25 @@ export const downloadData = (input: DownloadDataInput): DownloadDataOutput => {
...
-	async () => {
-		const { bucket, keyPrefix, s3Config } = await resolveS3ConfigAndInput(
-			Amplify,
-			downloadDataOptions,
+	(downloadDataInput: DownloadDataInput, abortSignal: AbortSignal) =>
+	async (): Promise<
+		StorageDownloadDataOutput<StorageItemKey | StorageItemPath>
+	> => {
+		const { options: downloadDataOptions } = downloadDataInput;
+		const { bucket, keyPrefix, s3Config, identityId } =
```
HuiSF, line 112: So we are not really splitting the implementation for key and path use cases here, and still resolves the prefix. I'm wondering if it worth investment now to split the implementation now for

1. performance - when use only path
2. easy deprecation process in the future just remove the key branch and then forget about it

ashika112: My thoughts, its only key prefix concat mostly not too worried about performance since we need other params any ways here. That said, I have been encouraging everyone if they find a gap/worthy case to extract it out to do it. wdyt?

packages/storage/src/providers/s3/apis/downloadData.ts
```diff
@@ -16,20 +23,13 @@ import {
...
-} from '../types/outputs';
 
 interface DownloadData {
 	/**
 	 * Download S3 object data to memory
 	 *
 	 * @param input - The DownloadDataInputPath object.
 	 * @returns A cancelable task exposing result promise from `result` property.
-	 * @throws service: {@link S3Exception} - thrown when checking for existence of the object
-	 * @throws validation: {@link StorageValidationErrorCode } - Validation errors
```
jimblanc: Can we just remove the links? Still some value in indicating that something may be thrown in specific circumstances. E.g. what I'm doing in upload:

```
	 * @throws Service: `S3Exception` thrown when checking for existence of the object.
	 * @throws Validation: `StorageValidationErrorCode` thrown when a validation error occurs.
```

packages/storage/src/providers/s3/apis/downloadData.ts
```diff
@@ -62,8 +62,6 @@ interface DownloadData {
 	 *
 	 * @param input - The DownloadDataInputKey object.
 	 * @returns A cancelable task exposing result promise from `result` property.
-	 * @throws service: {@link S3Exception} - thrown when checking for existence of the object
-	 * @throws validation: {@link StorageValidationErrorCode } - Validation errors
```
jimblanc: Ditto

packages/storage/src/providers/s3/apis/downloadData.ts
ashika112: Doing in sep PR if its ok

jimblanc (approved):

packages/storage/src/errors/types/validation.ts
```diff
@@ -49,4 +50,7 @@ export const validationErrorMap: AmplifyErrorMap<StorageValidationErrorCode> = {
 		message:
 			'Upload source type can only be a `Blob`, `File`, `ArrayBuffer`, or `string`.',
 	},
+	[StorageValidationErrorCode.InvalidStorageOperationInput]: {
+		message: 'Missing path or key parameter in Input',
```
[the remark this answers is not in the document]
ashika112: I will add to docs PR im opening up next.

HuiSF (approved):

PR #13078, opened 2024-03-04
fix(api-graphql): custom operation selection set on nested custom types

HuiSF:
#### Description of changes

* Updated custom operation impl. to use pre-existing selection set generation utility function so it can handle nested custom types
* Added unit tests against the nested custom types scenarios
* Bump data schema packages versions to gain new implementations

1. unit tests
2. manual E2E tests in a sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/src/internals/operations/custom.ts
```diff
@@ -251,11 +233,12 @@ function operationSelectionSet(
 	) {
 		return '';
 	} else if (hasStringField(operation.type, 'nonModel')) {
-		const model = modelIntrospection.nonModels[operation.type.nonModel];
-		return `{${modelishTypeSelectionSet(model)}}`;
+		const nonModel = modelIntrospection.nonModels[operation.type.nonModel];
+		return `{${selectionSetIRToString(
```
stocaaro, line 237: We're wrapping strings in surrounding curly brackets, `{}`, in several places. I find the character sequence `` `{${`` a bit hard to read, which makes me wonder if a few extra chars for `wrapInCurlyBrackets(x)` might be more readable.

Not specific to this PR, but a thing I thought of.

stocaaro (approved):
Makes sense. I think all of the below are more nit than concern. Good to see this ship as is or with any attention to these comments.

iartemiev (approved):
LGTM. 
As Aaron suggested, would be nice to import from a top-level `utils/index.ts` instead of sub-dirs, but that's non-blocking

HuiSF:
Thanks if this is non-blocking I will open a separate PR to refactor further (also wanted to extra the selection set utils out from that long source file)

svidgen (approved):

PR #13044, opened 2024-02-22
chore: Improve hot-fix & release process to generate a PR for merging back to main

jimblanc:
#### Description of changes

This PR improves our hot-fix and release processes by reducing friction when unreleased changes are present on `main`. 

It makes the following changes to the process:
- Generates a temporary merge branch to eliminate risk of potentially merging unreleased changes on `main` into `release`
- Creates a PR for merging the temporary merge branch back into `main` upon conclusion of a release

Benefits of this process:
- Don't need to worry about wiping out unreleased changes on `main` when executing a hot-fix
- No longer would need to freeze `main` when executing releases (beyond perhaps allowing e2e tests to stabilize)

This change will be accompanied with corresponding runbook changes & team training sessions.

#### Description of how you validated changes
Tested workflow on my fork.

Example of generated PR: https://github.com/jimblanc/amplify-js/pull/5
Example workflow execution: https://github.com/jimblanc/amplify-js/actions/runs/8008326884/job/21874530318

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #13095, opened 2024-03-07
chore: Refactor release workflow to allow workflow execution on generated PRs

jimblanc:
#### Description of changes

This PR proposes a change to our usage of `[ci skip]` in releases by introducing a new `[skip release]` syntax which will skip the `push-latest-release.yml` workflow specifically (which is what we're achieving with our current usage of `[ci skip]`).

This is intended to address the following issue:
- A recent change (https://github.com/aws-amplify/amplify-js/pull/13044) refactored our release process to create a PR for merging `release` back into `main` instead of force-pushing.
- Because we're using `[ci skip]` in the "update API docs" & "Publish" commits which are generated during releases, the resulting PR does not trigger the `codeql` & `pr` workflows required to actually merge the PR as per the branch protection rules on `main`. As a result, admin intervention or additional commit is required to merge the PR.

Alternatives considered:
- Use `pull_request_target` instead of `pull_request` for the CodeQL & PR workflows, which bypasses the `[ci skip]` check
  - This was discarded because `pull_request_target` changes the security posture of the repo for PRs originating from forks
- Automatically merge the generated PR using `gh pr merge`
  - This was discarded because there may be merge conflicts due to changes on `main` (particularly in the case of hot-fixes) that would require manual intervention anyway

Tested that this logic correctly skips the release workflow when `[skip release]` is present in the head commit of the triggering branch: https://github.com/aws-amplify/amplify-js/actions/runs/8180781321

Tested that PRs into `main` containing `[skip release]` still trigger the required PR workflows. Also verified that branches containing `[ci skip]` exhibit the same behavior we observed in the release (i.e. stuck PR workflows).

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/.github/workflows/push-latest-release.yml
+++ b/.github/workflows/push-latest-release.yml
@@ -12,8 +12,8 @@ on:
 
 jobs:
   e2e:
-    # Skip the release workflow if the most recent commit in the triggering branch contains [skip build]
-    if: ${{ ! contains(github.event.commits[0].message, '[skip release]') }}
+    # Skip the release workflow if the head commit in the triggering branch contains [skip release]
+    if: ${{ !contains(github.event.head_commit.message, '[skip release]') }}
     secrets: inherit
     uses: ./.github/workflows/callable-release-verification.yml
   release:
--- a/.github/workflows/push-latest-release.yml
+++ b/.github/workflows/push-latest-release.yml
@@ -12,6 +12,8 @@ on:
 
 jobs:
   e2e:
+    # Skip the release workflow if the most recent commit in the triggering branch contains [skip build]
+    if: ${{ ! contains(github.event.commits[0].message, '[skip release]') }}
     secrets: inherit
     uses: ./.github/workflows/callable-release-verification.yml
   release:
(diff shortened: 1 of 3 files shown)


.github/workflows/push-latest-release.yml
```diff
@@ -12,6 +12,8 @@ on:
 
 jobs:
   e2e:
+    # Skip the release workflow if the head commit in the triggering branch contains [skip release]
+    if: ${{ !contains(github.event.head_commit.message, '[skip release]') }}
```
ashwinkumar6, line 16: nit question,
If we have multiple variables would we prefer `my commit message [skip release] [skip e2e]` or `my commit message [skip release e2e]`

package.json
```diff
@@ -28,7 +28,7 @@
 		"unlink-all": "lerna exec --no-bail --parallel -- yarn unlink; exit 0",
 		"publish:preid": "./scripts/preid-env-vars-exist.sh && lerna publish --canary --force-publish --dist-tag=${PREID_PREFIX} --preid=${PREID_PREFIX}${PREID_HASH_SUFFIX} --yes",
 		"publish:main": "lerna publish --canary --force-publish --dist-tag=unstable --preid=unstable${PREID_HASH_SUFFIX} --yes",
-		"publish:release": "lerna publish --conventional-commits --message 'chore(release): Publish [ci skip]' --yes",
+		"publish:release": "lerna publish --conventional-commits --message 'chore(release): Publish [skip release]' --yes",
```
ashwinkumar6, line 31: nit question,
If we have multiple variables would we prefer `my commit message [skip release] [skip e2e]` or `my commit message [skip release e2e]`

jimblanc: It's a good question, I think I'd prefer to keep it narrowly tailored to the overall "intents" in our CI 🤔

stocaaro (approved):
I would be a fan of putting a PR/human into this process to avoid using the write token for security reasons.  This isn't changing the existing process, so as an amendment I don't see any risk in moving forward as is.

elorzafe (approved):
Looks good

Thanks @jimblanc 🎖️

PR #13099, opened 2024-03-09
feat(storage): uploadData path parameter support

jimblanc:
#### Description of changes

This PR adds support for the Gen2 `path` parameter to the `uploadData` API for single part & multi-part uploads.

- Existing key-based unit & E2E tests pass without modification
- Basic happy case testing in the `storage-gen2` sample app (additional testing underway)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/storage/src/providers/s3/apis/uploadData/multipart/uploadCache.ts
```diff
@@ -112,10 +112,18 @@ export const getUploadsCacheKey = ({
 	accessLevel,
 	key,
 }: UploadsCacheKeyOptions) => {
+	let levelStr;
 	const resolvedContentType =
 		contentType ?? file?.type ?? 'application/octet-stream';
-	const levelStr = accessLevel === 'guest' ? 'public' : accessLevel;
+
+	if (accessLevel === undefined) {
+		levelStr = 'custom';
```
jimblanc, line 120: This value is used to construct the cache key for multi-part uploads in local or async storage. In the case where no access level is provided, I use `custom` to indicate that access rules are being managed via the gen2 resources.

packages/storage/src/providers/s3/apis/uploadData/multipart/uploadCache.ts
```diff
@@ -112,10 +112,18 @@ export const getUploadsCacheKey = ({
...
+	} else {
+		levelStr = accessLevel === 'guest' ? 'public' : accessLevel;
+	}
+
 	const baseId = `${size}_${resolvedContentType}_${bucket}_${levelStr}_${key}`;
```
jimblanc, line 125: Thought exercise for the reader: Is it okay if we use the full object path (for the case of `path` uploads) here in place of `${key}`, or should we try to clip off the end of the path? It should still work either way 🤔 Might be worth mentioning in the threat model.

packages/storage/__tests__/providers/s3/apis/uploadData/index.test.ts
```diff
@@ -121,3 +123,103 @@ describe('uploadData', () => {
 		});
 	});
 });
+
+describe('uploadData with path', () => {
```
jimblanc, line 127: Duplicated `key` test suite above & added functionality for `path` behavior in order to ease cleanup when the time comes to remove `key` support and to avoid accidentally dropping test coverage.

packages/storage/__tests__/providers/s3/apis/uploadData/multipartHandlers.test.ts
```diff
@@ -629,3 +632,507 @@ describe('getMultipartUploadHandlers', () => {
 		});
 	});
 });
+
+describe('getMultipartUploadHandlers with path', () => {
```
jimblanc, line 636: Duplicated `key` test suite above & added functionality for `path` behavior in order to ease cleanup when the time comes to remove `key` support and to avoid accidentally dropping test coverage.

packages/storage/__tests__/providers/s3/apis/uploadData/putObjectJob.test.ts
```diff
@@ -130,3 +131,92 @@ describe('putObjectJob', () => {
 		expect(calculateContentMd5).toHaveBeenCalledWith('data');
 	});
 });
+
+describe('putObjectJob with path', () => {
```
jimblanc, line 135: Duplicated `key` test suite above & added functionality for `path` behavior in order to ease cleanup when the time comes to remove `key` support and to avoid accidentally dropping test coverage.

packages/storage/src/providers/s3/apis/uploadData/putObjectJob.ts
jimblanc, line 1: This file is used for single-part uploads

packages/storage/src/providers/s3/apis/uploadData/multipart/uploadHandlers.ts
jimblanc, line 1: This file is used for multi-part uploads

packages/storage/__tests__/providers/s3/apis/uploadData/index.test.ts
```diff
@@ -15,6 +15,7 @@ jest.mock('../../../../../src/providers/s3/utils/');
 jest.mock('../../../../../src/providers/s3/apis/uploadData/putObjectJob');
 jest.mock('../../../../../src/providers/s3/apis/uploadData/multipart');
 
+const testPath = 'testPath/object';
```
ashwinkumar6, line 18: nit
would be nice to also add a test with path as a function to check if the function got passed to the respective handler. 

Since we also have some top level assertions to make sure key | path are present, if these assertions have issues this test might be able to catch it :thinking:

ashwinkumar6 (dismissed):
Great work @jimblanc !! :rocket:
LGTM

Just a few minor nit question

packages/storage/__tests__/providers/s3/apis/uploadData/index.test.ts
ashika112: i think to jim's point it might be ok 🤔

packages/storage/__tests__/providers/s3/apis/uploadData/putObjectJob.test.ts
```diff
@@ -130,3 +131,92 @@ describe('putObjectJob', () => {
...
+	test.each([
+		{
+			path: testPath,
+			expectedKey: testPath,
+		},
+		{
+			path: () => testPath,
```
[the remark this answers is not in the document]
ashika112: i think it doesn't matter. technically this is allowed too

packages/storage/src/providers/s3/apis/uploadData/multipart/initialUpload.ts
```diff
@@ -54,7 +54,7 @@ export const loadOrCreateMultipartUpload = async ({
 	metadata,
 	abortSignal,
 }: LoadOrCreateMultipartUploadOptions): Promise<LoadOrCreateMultipartUploadResult> => {
-	const finalKey = keyPrefix + key;
+	const finalKey = keyPrefix !== undefined ? keyPrefix + key : key;
```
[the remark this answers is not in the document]
ashika112: I mean it looks like we are making this oblivious of key/path implementation. So just this should suffice? 🤔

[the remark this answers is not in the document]
jimblanc: Correct, `keyPrefix` will only exist for `key` use-cases. IMO I think this logic is a little more maintainable & intuitive (i.e. if a prefix exists, prepend it, otherwise carry on)

packages/storage/__tests__/providers/s3/apis/uploadData/index.test.ts
jimblanc: Added another test

packages/storage/src/providers/s3/apis/uploadData/multipart/uploadCache.ts
ashika112: should we just add this. I think its better to have a hash maybe attached to `custom`?

ashwinkumar6 (approved):

ashika112 (approved):
🚢 🚀  YAy! we are there!

PR #13100, opened 2024-03-09
feat(storage): list API to accept both prefix and path

Samaritan1011001:
#### Description of changes

List API will now accept either prefix or path as input. Depending on which is given, their respective types are suggested in the IDE and output has key or path respectively. 
Some notes, 
- If prefix is given, options show the accessLevel, targetIdentityId (this is as is currently in V6) 
- if no input object is given or its empty, default guest prefix is assumed (this is as is currently in V6).
- If path is given, options does not show accessLevel, targetIdentityId (gen2 feature)

Below table shows the different type changes in `index.ts`. 

Note: None of the exposed type names have changed. Only the first two rows are the ones exposed via `aws-amplify/storage`. 

| **file**                                  | **feature branch(gen2-storage)**                                      | **PR change**                                                                                                                                                            |
|-------------------------------------------|-----------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| src/index.ts                              | ListAllInput, ListPaginateInput,<br>ListAllOutput, ListPaginateOutput | No change                                                                                                                                                                |
| src/providers/s3/index.ts                 | ListAllInput, ListPaginateInput,<br>ListAllOutput, ListPaginateOutput | No change                                                                                                                                                                |
| src/types/index.ts                        | StorageListInput                                                      | StorageListInputPrefix, StorageListInputPath                                                                                                                             |
| (options) src/providers/s3/types/index.ts | ListAllOptions, ListPaginateOptions                                   | ListAllOptionsPrefix, ListPaginateOptionsPrefix, <br>ListAllOptionsPath, ListPaginateOptionsPath                                                                         |
| (inputs) src/providers/s3/types/index.ts  | ListAllInput, ListPaginateInput                                       | ListAllInput, ListPaginateInput,<br>ListAllInputPath, ListPaginateInputPath,<br>ListAllInputPrefix,ListPaginateInputPrefix                                               |
| (outputs) src/providers/s3/types/index.ts | ListOutputItem, ListAllOutput, ListPaginateOutput                                                         | ListOutputItemKey, ListOutputItemPath<br>ListAllOutput, ListPaginateOutput<br>ListAllOutputPrefix, ListAllOutputPath<br>ListPaginateOutputPath, ListPaginateOutputPrefix |

Linked to a sample app and could complete the list operation both ways

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/storage/src/providers/s3/types/options.ts
```diff
@@ -57,19 +57,34 @@ export type GetPropertiesOptions = ReadOptions & CommonOptions;
 export type RemoveOptions = WriteOptions & CommonOptions;
 
 /**
- * Input options type for S3 list API.
+ * Input options type for S3 list API to list all items for a prefix.
```
ashika112, line 60:
```suggestion
 * Input options type with prefix for S3 list all API.
```

ashika112: you can make similar change for all doc string

packages/storage/src/providers/s3/types/inputs.ts
```diff
@@ -47,14 +50,12 @@ export type GetPropertiesInput =
...
+ * Input type for S3 list API.
+ * It may contain the full path or just the prefix.
  */
-export type ListAllInput = StorageListInput<ListAllOptions>;
-
-/**
- * Input type for S3 list API. Lists bucket objects with pagination.
- */
-export type ListPaginateInput = StorageListInput<ListPaginateOptions>;
+export type ListInput =
```
ashika112, line 56: i think we should keep this separated as before and only union path/prefix on each level. This will also keep consistency with how we make changes in other APIs

packages/storage/src/providers/s3/types/outputs.ts
```diff
@@ -67,15 +83,24 @@ export type GetPropertiesOutput = ItemKey;
...
  */
-export type ListPaginateOutput = StorageListOutput<ListOutputItem> & {
+export type ListPaginateOutput = StorageListOutput<
+	ListOutputItemKey | ListOutputItemPath
+> & {
 	nextToken?: string;
 };
 
+/**
+ * Output type for S3 list API.
```
ashika112, line 100: same with output

packages/storage/src/providers/s3/apis/list.ts
```diff
@@ -4,38 +4,50 @@
...
+	 * To list all the pages, you can set `listAll` to true in `options` to get all the files from S3.
+	 * For pagination, pageSize defaults to 1000. Additionally, the result will include a nextToken if there are more items to retrieve.
+	 * @param input - The StorageListInputPath object.
+	 * @returns when listAll is true, A list of keys and metadata for all objects in path
+	 * @returns when listAll is false, A list of keys and metadata with nextToken
 	 * @throws service: {@link S3Exception} - S3 service errors thrown when checking for existence of bucket
 	 * @throws validation: {@link StorageValidationErrorCode } - thrown when there are issues with credentials
 	 */
-	(input?: ListPaginateInput): Promise<ListPaginateOutput>;
+	(
```
ashika112, line 31: u dont want to change the interface. this will break customers

ashika112 (changes requested):

packages/storage/__tests__/providers/s3/apis/list.test.ts
```diff
@@ -98,6 +99,15 @@ describe('list API', () => {
 			{
 				expectedPath: `public/`,
 			},
+			{
+				path: undefined,
```
ashika112, line 103: Also this path and other path seems confusing. Could we rename the tests to reflect `path` only for the input parameter and name other differently

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -20,25 +24,48 @@ import {
...
-	prefix: string;
+	generatedPrefix: string;
+	storageInputType: string;
 }
 
 export const list = async (
 	amplify: AmplifyClassV6,
-	input?: ListAllInput | ListPaginateInput,
-): Promise<ListAllOutput | ListPaginateOutput> => {
-	const { options = {}, prefix: path = '' } = input ?? {};
```
ashika112, line 36: Same here. It is weird we had path but can we rename that to something so we can differentiate btw `path` as input vs this.

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
```diff
@@ -22,19 +38,35 @@ export const validateStorageOperationInput = (
...
 	identityId?: string,
 ) => {
-	assertValidationError(
-		!!(input as Input).key || !!(input as Input).path,
-		StorageValidationErrorCode.InvalidStorageOperationInput,
-	);
-
 	if (isInputWithPath(input)) {
-		const { path } = input;
+		const { path } = input as StorageOperationInputPath;
```
ashika112, line 42: So here, can we just built a similar util that change the original? Since we dont want input to be strict union of the three for others outside of list? Wdyt?

packages/storage/src/providers/s3/apis/internal/list.ts
Samaritan1011001: Yeah, it was confusing, that's why I removed the renaming to path. 
Now, we get the prefix or path from input based on the input via `validateStorageOperationInput`

packages/storage/src/providers/s3/apis/list.ts
Samaritan1011001: Got it. I reverted it to the previous state but now the `ListAllInput` is union of path and prefix types. Same for `ListPaginationInput`

packages/storage/src/providers/s3/types/inputs.ts
Samaritan1011001: Updated it, lmk if that's better.

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
Samaritan1011001: That's because in the case of empty input to list API it will assert No key or path which we don't want. That's why I put the assertion in it's respective branches.

packages/storage/__tests__/providers/s3/apis/list.test.ts
```diff
@@ -269,6 +274,146 @@ describe('list API', () => {
...
+			},
+			{
+				path: `${privatePrefix}${defaultIdentityId}/${key}`,
+			},
+			{
+				path: `${protectedPrefix}${defaultIdentityId}/${key}`,
+			},
+			{
+				path: `${protectedPrefix}${targetIdentityId}/${key}`,
+			},
```
ashika112, line 296: I dont think we need to call multiple strings here like test on accessLevels. Just - path as string , path as function should be good.

packages/storage/src/providers/s3/apis/list.ts
```diff
@@ -16,18 +15,18 @@ import { list as listInternal } from './internal/list';
 
 interface ListApi {
 	/**
-	 * List files with given prefix in pages
+	 * List files with given prefix or path in pages
```
ashika112, line 18: On this you are gonna need 4 overloads -> separately for `path` vs `perfix` for `listAll` and `listPaginated`

packages/storage/src/providers/s3/apis/server/list.ts
```diff
@@ -1,6 +1,5 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-
 import {
 	AmplifyServer,
```
ashika112, line 4: same on overloads in this file and u need to separate out deprecation warnings as well

packages/storage/src/providers/s3/types/outputs.ts
```diff
@@ -67,12 +83,16 @@ export type GetPropertiesOutput = ItemKey;
 /**
  * Output type for S3 list API. Lists all bucket objects.
  */
-export type ListAllOutput = StorageListOutput<ListOutputItem>;
```
ashika112, line 70: shouldn't `ListOutputItem` be removed?

packages/storage/src/providers/s3/index.ts
```diff
@@ -15,8 +15,6 @@ export {
 	UploadDataInput,
 	DownloadDataInput,
 	RemoveInput,
-	ListAllInput,
```
ashika112, line 18: Again i guess these shouldn't be removed

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
ashika112: I think the check for trailing slash might cover this ?

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
```diff
@@ -1,30 +1,62 @@
...
+
+const isInputWithPrefix = (
+	input: Input,
+): input is StorageOperationInputPrefix => 'prefix' in input;
 
 export const validateStorageOperationInput = (
 	input: Input,
 	identityId?: string,
 ) => {
-	assertValidationError(
```
ashika112, line 15: this shouldn't be removed right

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
```diff
@@ -1,30 +1,62 @@
...
-		!!(input as Input).key || !!(input as Input).path,
-		StorageValidationErrorCode.InvalidStorageOperationInput,
-	);
-
 	if (isInputWithPath(input)) {
-		const { path } = input;
+		const { path } = input as StorageOperationInputPath;
+		assertValidationError(
+			!!path,
```
ashika112, line 34: minor difference between this and other validation. Previous validation where checking if path even exists while these one check for value of the parameter

packages/storage/__tests__/providers/s3/apis/list.test.ts
```diff
@@ -141,12 +150,10 @@ describe('list API', () => {
 				});
 				expect.assertions(4);
```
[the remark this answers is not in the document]
kvramyasri7: yeah!!.. They are helpful to count the assertions. This is the standard we had when you are on leave Ashikaaa..

packages/storage/__tests__/providers/s3/apis/list.test.ts
```diff
@@ -293,5 +438,15 @@ describe('list API', () => {
 				expect(error.$metadata.httpStatusCode).toBe(404);
 			}
 		});
+		it('should throw InvalidStorageOperationInput error when the path is empty', async () => {
+			expect.assertions(1);
+			try {
+				await list({ path: '' });
+			} catch (error: any) {
+				expect(error.name).toBe(
+					StorageValidationErrorCode.InvalidStorageOperationInput,
```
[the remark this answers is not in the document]
kvramyasri7: +1

packages/storage/__tests__/providers/s3/apis/list.test.ts
kvramyasri7: I recommend to test everything as recommended by @stocaaro  for more reffernce.

packages/storage/src/index.ts
```diff
@@ -15,8 +15,6 @@ export {
 	UploadDataInput,
 	DownloadDataInput,
 	RemoveInput,
-	ListAllInput,
```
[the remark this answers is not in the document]
kvramyasri7: I think we can remove this....

packages/storage/src/providers/s3/types/outputs.ts
kvramyasri7: Noo....

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
kvramyasri7: Yesss.....  Ashikaaa....

kvramyasri7 (approved):

packages/storage/src/providers/s3/apis/server/list.ts
kvramyasri7: ohhh!!! Thanks for the suggestion Ashikaaaaaa

packages/storage/src/index.ts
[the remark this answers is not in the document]
ashwinkumar6: I think removing anything from 
`packages/storage/src/index.ts` or `packages/storage/src/providers/s3/index.ts` would be a breaking change. We have an export test for all API exposed but no

packages/storage/__tests__/providers/s3/apis/list.test.ts
ashwinkumar6: agree with Ashika,
Question: 
1. Is it possible to have path as undefined (`await list()`).
2. If i have a gen2 backend and i call `await list()` would this make use of prefix or path. Would we face error if it uses prefix
3. post prefix depreciation would `await list()` be supported

packages/storage/__tests__/providers/s3/apis/list.test.ts
```diff
@@ -269,6 +274,146 @@ describe('list API', () => {
...
+						Contents: [{ ...listObjectClientBaseResultItem, Key: path }],
+						NextContinuationToken: nextToken,
+					};
+				});
+				expect.assertions(4);
+				let response = await list({
+					path,
+				});
+				expect(response.items).toEqual([
+					{ ...listResultItem, path: path ?? '' },
```
ashwinkumar6, line 312: would need to dig in to recollect but when would path be `''` 
(for prefix and path test)

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -82,15 +111,22 @@ const _listAll = async ({
 		continuationToken = pageNextToken;
 	} while (continuationToken);
 
-	return {
-		items: listResult,
-	};
+	if (storageInputType === STORAGE_INPUT_PREFIX) {
+		return {
+			items: listResult as ListOutputItemKey[],
```
ashwinkumar6, line 116: would we be able to get by without using `as` maybe with type narrowing

packages/storage/src/providers/s3/types/outputs.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { StrictUnion } from '@aws-amplify/core/internals/utils';
```
ashwinkumar6, line 4: i think this can be cleaned up

packages/storage/src/providers/s3/types/outputs.ts
```diff
@@ -25,10 +27,24 @@ export interface ItemBase {
 	contentType?: string;
 }
 
+/**
+ * type for S3 list item with key.
+ */
+export type ListOutputItemKey = Omit<ItemKey, 'metadata'>;
```
ashwinkumar6, line 33: should we had `deprecated` doc string for this

packages/storage/__tests__/providers/s3/apis/list.test.ts
Samaritan1011001: 1. yes, it takes the default prefix : `public/`
2. It would use a default prefix: `public/`
3. I will say no but that's a future decision to take in the next major version as it is breaking.

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
Samaritan1011001: Created a separate function as discussed.

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
Samaritan1011001: No longer valid as this is separate function now that checks for only undefined.

packages/storage/__tests__/providers/s3/apis/list.test.ts
Samaritan1011001: Thanks for catching. Updated this.

packages/storage/src/providers/s3/apis/internal/list.ts
Samaritan1011001: Separated out the functions and duplicated it to be more clear.

packages/storage/src/providers/s3/types/outputs.ts
Samaritan1011001: What do you mean?

ashwinkumar6: Sorry I mean I think this isn't used to we can remove it

kvramyasri7: What??? @ashwinkumar6

packages/storage/src/providers/s3/types/options.ts
```diff
@@ -65,19 +65,34 @@ export type GetPropertiesOptions = ReadOptions & CommonOptions;
 export type RemoveOptions = WriteOptions & CommonOptions;
 
 /**
- * Input options type for S3 list API.
+ * Input options type with prefix for S3 list all API.
  */
-export type ListAllOptions = StorageListAllOptions &
+export type ListAllOptionsPrefix = StorageListAllOptions &
```
ashika112, line 70: missed deprecation warning

packages/storage/src/providers/s3/types/options.ts
```diff
@@ -65,19 +65,34 @@ export type GetPropertiesOptions = ReadOptions & CommonOptions;
...
 	ReadOptions &
 	CommonOptions;
 
 /**
- * Input options type for S3 list API.
+ * Input options type with prefix for S3 list API to paginate items.
  */
-export type ListPaginateOptions = StorageListPaginateOptions &
+export type ListPaginateOptionsPrefix = StorageListPaginateOptions &
```
ashika112, line 77: missed deprecation warning

packages/storage/src/providers/s3/utils/validateStorageInputWithPrefix.ts
```diff
@@ -0,0 +1,45 @@
...
+import { STORAGE_INPUT_PATH, STORAGE_INPUT_PREFIX } from './constants';
+
+// Local assertion function with StorageOperationPrefixInputType as Input
+export const isInputWithPath = (
+	input: StorageOperationPrefixInputType,
+): input is StorageOperationInputPath => {
+	return input.path !== undefined;
+};
+
+export const validateStorageInputPrefix = (
```
ashika112, line 20: can we name this `validateStorageListOperationInput` and also update type name accordingly?

packages/storage/src/providers/s3/utils/validateStorageInputWithPrefix.ts
```diff
@@ -0,0 +1,45 @@
...
+) => {
+	if (isInputWithPath(input)) {
+		assertValidationError(
+			(input as StorageOperationPrefixInputType).path !== '',
+			StorageValidationErrorCode.InvalidStorageOperationInput,
+		);
+		const { path } = input;
+		const objectKey = typeof path === 'string' ? path : path({ identityId });
+
+		// Assert if there is a leading slash in the input path
```
ashika112, line 32:
```suggestion
		// Assert on no leading slash in the path parameter
```

ashika112: Nice. I missed it on mine.

packages/storage/__tests__/providers/s3/apis/list.test.ts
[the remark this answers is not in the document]
ashika112: @Samaritan1011001 can u include these as well in revision. Atleast for `path` cases

packages/storage/src/types/inputs.ts
```diff
@@ -14,7 +14,9 @@ import {
 export type StorageOperationInputType = StrictUnion<
 	StorageOperationInputKey | StorageOperationInputPath
 >;
-
+export type StorageOperationPrefixInputType = StrictUnion<
```
ashwinkumar6, line 17: I'm wondering if we can improve this type name 
- This type is a union of path and prefix
- There's another type with a very similar name 'StorageOperationInputPrefix'

If this type is only used by the `validateStorageInputPrefix()` util, maybe we could name it based on that

packages/storage/src/providers/s3/utils/validateStorageInputWithPrefix.ts
ashika112: We will do `validateStorageOperationInputWithPrefix`

packages/storage/src/providers/s3/utils/client/utils/parsePayload.ts
```diff
@@ -39,7 +39,7 @@ export const parseXmlBody = async (response: HttpResponse): Promise<any> => {
 		try {
 			return parser.parse(data);
 		} catch (error) {
-			throw new Error('Failed to parse XML response.');
+			throw new Error(`Failed to parse XML response: ${error}`);
```
ashwinkumar6, line 42: good catch

packages/storage/src/providers/s3/apis/list.ts
```diff
@@ -1,41 +1,69 @@
...
-export const list: ListApi = (
+export const list: ListApi = <
+	Output extends ListAllOutput | ListPaginateOutput,
+>(
 	input?: ListAllInput | ListPaginateInput,
-): Promise<ListAllOutput | ListPaginateOutput> => {
-	return listInternal(Amplify, input ?? {});
+): Promise<Output> => {
+	return listInternal(Amplify, input ?? {}) as Promise<Output>;
 };
```
ashwinkumar6, line 69:
```suggestion
): Promise<Output> => listInternal(Amplify, input ?? {}) as Promise<Output>
```

ashwinkumar6: super nit

packages/storage/src/providers/s3/apis/server/list.ts
```diff
@@ -1,53 +1,89 @@
...
+>(
 	contextSpec: AmplifyServer.ContextSpec,
 	input?: ListAllInput | ListPaginateInput,
-): Promise<ListAllOutput | ListPaginateOutput> => {
+): Promise<Output> => {
 	return listInternal(
 		getAmplifyServerContext(contextSpec).amplify,
 		input ?? {},
-	);
+	) as Promise<Output>;
```
ashwinkumar6, line 88: super nit:
same as previous comment

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -20,25 +28,37 @@ import {
...
+		keyPrefix: generatedPrefix,
+		identityId,
 	} = await resolveS3ConfigAndInput(amplify, options);
+
+	const { inputType, objectKey } = validateStorageOperationInputWithPrefix(
+		input,
+		identityId,
+	);
+	const path =
+		inputType === STORAGE_INPUT_PREFIX
```
ashwinkumar6, line 58: super nit,
should we store `inputType === STORAGE_INPUT_PREFIX` in a separate variable like isStorageWithPrefix since it's used in multiple places

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -50,27 +70,39 @@ export const list = async (
 	}
 	const listParams = {
 		Bucket: bucket,
-		Prefix: `${prefix}${path}`,
+		Prefix: path,
```
ashwinkumar6, line 73: nit:
should we directly define path here like
```
Prefix: isStorageWithPrefix ? `${generatedPrefix}${objectKey}` : objectKey;
```

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -50,27 +70,39 @@ export const list = async (
...
+					listParams,
+					generatedPrefix,
+				})
+		: inputType === STORAGE_INPUT_PREFIX
+			? _listPrefix({ s3Config, listParams, generatedPrefix })
+			: _listPath({ s3Config, listParams, generatedPrefix });
 };
 
-const _listAll = async ({
+const _listAllPrefix = async ({
```
ashwinkumar6, line 96: nit
should we also add deprication doc strings to these internal util functions, might be easier to cleanup

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -87,11 +119,11 @@ const _listAll = async ({
 	};
 };
 
-const _list = async ({
+const _listPrefix = async ({
```
ashwinkumar6, line 122: nit:
same as above (deprication doc string)

packages/storage/__tests__/providers/s3/apis/list.test.ts
```diff
@@ -99,33 +99,42 @@ describe('list API', () => {
 				expectedPath: `public/`,
 			},
 			{
-				path,
-				expectedPath: `public/${path}`,
+				key: undefined,
```
ashwinkumar6, line 102: QQ, is this and the case above the same

packages/storage/src/providers/s3/utils/validateStorageInputWithPrefix.ts
```diff
@@ -0,0 +1,45 @@
...
+	StorageOperationInputPath,
+	StorageOperationPrefixInputType,
+} from '../../../types/inputs';
+import { assertValidationError } from '../../../errors/utils/assertValidationError';
+import { StorageValidationErrorCode } from '../../../errors/types/validation';
+
+import { STORAGE_INPUT_PATH, STORAGE_INPUT_PREFIX } from './constants';
+
+// Local assertion function with StorageOperationPrefixInputType as Input
+export const isInputWithPath = (
```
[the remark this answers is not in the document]
Samaritan1011001: It's only ever used in this file so I prefer we just keep it here. But good catch, I'll make it local and not export it.

packages/storage/__tests__/providers/s3/apis/list.test.ts
```diff
@@ -293,5 +440,14 @@ describe('list API', () => {
 				expect(error.$metadata.httpStatusCode).toBe(404);
 			}
 		});
+		it('should throw InvalidStorageOperationInput error when the path is empty', async () => {
+			try {
```
ashwinkumar6, line 444: nit
I think in error cases to make sure the assertion in catch block is we can add `expect.assertions(<count>)`

packages/storage/src/types/inputs.ts
Samaritan1011001: Ty, renamed it to `StorageOperationInputWithPrefixPath` . These are internal types so it should be good enough and seeing the type definition is self explanatory on what's happening.

ashwinkumar6 (dismissed):
couple of nits
LGTM, great job !!! 💯

packages/storage/src/providers/s3/apis/internal/list.ts
Samaritan1011001: Just two places, don't see much advantage in storing it in a variable other than one less comparison.

packages/storage/src/providers/s3/apis/internal/list.ts
Samaritan1011001: Think the function name along with the code flow is self explanatory. I'll add in a comment above it for clarity if you would like.

Samaritan1011001: Oh the deprecation message, I see, well one would remove everything prefix so again think it's self explanatory with the funciton name

packages/storage/src/providers/s3/utils/validateStorageInputWithPrefix.ts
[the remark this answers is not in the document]
ashika112: +1. Also this will be only need till Prefix is all deprecated.

packages/storage/__tests__/providers/s3/apis/list.test.ts
Samaritan1011001: Good catch, ty

packages/storage/src/providers/s3/apis/internal/list.ts
ashika112: +1 we can add this. Can we verify if this is done in copy key as well.? 

I think this will help make deprecation easier later.

ashwinkumar6: Added on `copyWithKey` as well :white_check_mark:

ashwinkumar6 (approved):

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -50,27 +67,39 @@ export const list = async (
...
+		Prefix: isInputWithPrefix ? `${generatedPrefix}${objectKey}` : objectKey,
 		MaxKeys: options?.listAll ? undefined : options?.pageSize,
 		ContinuationToken: options?.listAll ? undefined : options?.nextToken,
 	};
 	logger.debug(`listing items from "${listParams.Prefix}"`);
 
 	return options.listAll
-		? _listAll({ s3Config, listParams, prefix })
-		: _list({ s3Config, listParams, prefix });
+		? isInputWithPrefix
```
jimblanc, line 77: Nit: This ternary is pretty complex, I think I'd prefer just if/else blocks. Additionally, this would let you extract the input objects which are identical.

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -112,15 +141,74 @@ const _list = async ({
...
+			size: item.Size,
+		})),
+		nextToken: response.NextContinuationToken,
+	};
+};
+
+const _listAllPath = async ({
+	s3Config,
+	listParams,
+	generatedPrefix,
```
jimblanc, line 158: Question: What does generatedPrefix mean in the context of `path`? Is this just the path the customer passed in? Also do we even need this param? It looks like `_listPath` doesn't use it internally.

packages/storage/src/providers/s3/types/options.ts
```diff
@@ -65,19 +65,36 @@ export type GetPropertiesOptions = ReadOptions & CommonOptions;
...
 	ReadOptions &
 	CommonOptions;
 
+/**
+ * Input options type with path for S3 list all API.
+ */
+export type ListAllOptionsPath = Omit<StorageListAllOptions, 'accessLevel'> &
```
jimblanc, line 86: Sanity check: Where does `targetIdentityId` get pulled in? Do we need to remove that anywhere?

packages/storage/src/providers/s3/utils/validateStorageOperationInputWithPrefix.ts
```diff
@@ -0,0 +1,45 @@
...
+	StorageOperationInputPath,
+	StorageOperationInputWithPrefixPath,
+} from '../../../types/inputs';
+import { assertValidationError } from '../../../errors/utils/assertValidationError';
+import { StorageValidationErrorCode } from '../../../errors/types/validation';
+
+import { STORAGE_INPUT_PATH, STORAGE_INPUT_PREFIX } from './constants';
+
+// Local assertion function with StorageOperationInputWithPrefixPath as Input
+const _isInputWithPath = (
```
jimblanc, line 14: Isn't this duplicated with the existing `isInputWithPath`?

packages/storage/src/providers/s3/utils/validateStorageOperationInputWithPrefix.ts
```diff
@@ -0,0 +1,45 @@
...
+	input: StorageOperationInputWithPrefixPath,
+): input is StorageOperationInputPath => {
+	return input.path !== undefined;
+};
+
+export const validateStorageOperationInputWithPrefix = (
```
jimblanc, line 20: Nit: Any way we can refactor this to use `validateStorageOperationInput`, or at least reduce code duplication? E.g. this function is missing a recent validation check that was added to `validateStorageOperationInput`. Would be ideal to not have to maintain validation rules in two places

jimblanc (commented):
Looking pretty good, some questions & nits below

packages/storage/src/providers/s3/types/options.ts
Samaritan1011001: It comes from the [ReadOptions](https://github.com/aws-amplify/amplify-js/pull/13100/files/34ad9f6a6fad8bf3c25f9091487c46512eeab932#diff-380e33ff82765e24f6b3c2872d359ee7946bf0109f3f1f96af5def4cf374170bL22) which is not included in the path types like the one above.

packages/storage/src/providers/s3/utils/validateStorageOperationInputWithPrefix.ts
Samaritan1011001: I kept the same name and local. The difference is the type of the Input. Did this so this can be removed together when deprecating the `prefix` code path later.

packages/storage/src/providers/s3/utils/validateStorageOperationInputWithPrefix.ts
Samaritan1011001: Agreed. But it was decided with @ashika112 that duplication is better in this case because it will make it easier to remove prefix things. Plus adding this to `validateStorageOperationInput` made it more complex because it is valid to have empty `prefix` in list API which is not the same as `key` on other APIs

jimblanc (approved):

packages/storage/src/providers/s3/utils/validateStorageOperationInputWithPrefix.ts
Samaritan1011001: @jimblanc Upon more thought, I included the validation for prefix & path as well. 
https://github.com/aws-amplify/amplify-js/pull/13100/files#diff-5262fafa91521452d88353f3bd256f307a638e57d38d8e67dfda317c30be6bcaR25

packages/storage/src/errors/types/validation.ts
```diff
@@ -62,6 +65,15 @@ export const validationErrorMap: AmplifyErrorMap<StorageValidationErrorCode> = {
 	[StorageValidationErrorCode.InvalidStorageOperationInput]: {
 		message: 'Missing path or key parameter in Input.',
 	},
+	[StorageValidationErrorCode.NoPrefix]: {
+		message: 'Missing `prefix` parameter in Input.',
+	},
+	[StorageValidationErrorCode.NoPath]: {
+		message: 'Missing `path` parameter in Input.',
+	},
+	[StorageValidationErrorCode.PathAndPrefixNotAllowed]: {
```
ashika112, line 74: I know this is still in discussion if we need separate validation or not in Jim's PR. Lets finalize this in that PR and add change in Jim's PR if needed. @jimblanc

packages/storage/src/errors/types/validation.ts
```diff
@@ -62,6 +65,15 @@ export const validationErrorMap: AmplifyErrorMap<StorageValidationErrorCode> = {
 	[StorageValidationErrorCode.InvalidStorageOperationInput]: {
 		message: 'Missing path or key parameter in Input.',
 	},
+	[StorageValidationErrorCode.NoPrefix]: {
+		message: 'Missing `prefix` parameter in Input.',
```
jimblanc, line 69: Nit: Can we remove the ticks to be consistent with `InvalidStorageOperationInput` (or vice-versa)

packages/storage/src/providers/s3/utils/validateStorageOperationInputWithPrefix.ts
```diff
@@ -0,0 +1,50 @@
...
+	input: StorageOperationInputWithPrefixPath,
+	identityId?: string,
+) => {
+	// Assert that both prefix and path are not present together
+	assertValidationError(
+		!('prefix' in input && 'path' in input),
```
jimblanc, line 26: Technically this won't catch `undefined` or `null` values. Maybe we just want to do something like: `!(input as Input).key || !(input as Input).path,`

ashwinkumar6 (dismissed):

ashika112 (approved):

jimblanc (approved):
LGTM one small nit

packages/storage/src/errors/types/validation.ts
```diff
@@ -66,14 +64,9 @@ export const validationErrorMap: AmplifyErrorMap<StorageValidationErrorCode> = {
...
-		message: 'Missing `prefix` parameter in Input.',
-	},
-	[StorageValidationErrorCode.NoPath]: {
-		message: 'Missing `path` parameter in Input.',
-	},
-	[StorageValidationErrorCode.PathAndPrefixNotAllowed]: {
-		message: 'Input cannot contain both `path` and `prefix` parameters.',
+	[StorageValidationErrorCode.InvalidStorageOperationPrefixInput]: {
+		message:
+			'Path or prefix parameter must be specified in the input. Both can not be specified at the same time.',
```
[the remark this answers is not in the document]
Samaritan1011001: Updated it to `Both path and prefix can not be specified at the same time.`

ashwinkumar6 (approved):

ashika112 (approved):

jimblanc (approved):

PR #13101, opened 2024-03-11
chore(api-graphql): enable eslint and remove tslint

HuiSF:
#### Description of changes

* Enabled eslint for the `api-graphql` package codebase and removed the tslint cnofig
* Ran `eslint --fix` to auto-fix linter issues
* Manually fixed remaining linter reported errors

* Unit tests are still passing
* Ran E2E test against the feature branch all tests are passing https://github.com/aws-amplify/amplify-js/actions/runs/8210556406

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -63,17 +63,17 @@ export interface ObserverQuery {
...
 }
 
 const standardDomainPattern =
-	/^https:\/\/\w{26}\.appsync\-api\.\w{2}(?:(?:\-\w{2,})+)\-\d\.amazonaws.com(?:\.cn)?\/graphql$/i;
+	/^https:\/\/\w{26}\.appsync-api\.\w{2}(?:(?:-\w{2,})+)-\d\.amazonaws.com(?:\.cn)?\/graphql$/i;
 
 const customDomainPath = '/realtime';
 
-interface DataObject {
+interface DataObject extends Record<string, unknown> {
```
HuiSF, line 76: The object that conforms to the `DataObject` are assigned to some other objects that has `Record` as the base type.

packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -98,7 +98,7 @@ export interface AWSAppSyncRealTimeProviderOptions {
 	variables?: Record<string, DocumentType>;
 	apiKey?: string;
 	region?: string;
-	libraryConfigHeaders?(): {} | (() => Promise<{}>);
+	libraryConfigHeaders?(): Promise<Record<string, unknown> | Headers>;
```
HuiSF, line 101: This is aligning with [here](https://github.com/aws-amplify/amplify-js/blob/1570b12a12e782e53dcf37c22c620835587e1a07/packages/core/src/singleton/API/types.ts#L9-L12) so the object refered by the `libraryConfigHeaders` can be assigned without type incompatibility error.

packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -924,7 +932,9 @@ export class AWSAppSyncRealTimeProvider {
 		Record<string, unknown> | undefined
 	> {
 		const headerHandler: {
-			[key in GraphQLAuthMode]: (arg0: AWSAppSyncRealTimeAuthInput) => {};
```
HuiSF, line 919: The original return type ~= any non-nullable object which was not able to catch type incompatibility.

packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -978,7 +988,7 @@ export class AWSAppSyncRealTimeProvider {
 		host,
 	}: AWSAppSyncRealTimeAuthInput) {
 		const dt = new Date();
-		const dtStr = dt.toISOString().replace(/[:\-]|\.\d{3}/g, '');
+		const dtStr = dt.toISOString().replace(/[:-]|\.\d{3}/g, '');
```
HuiSF, line 991: Unnecessary escape on `-`

packages/api-graphql/src/internals/APIClient.ts
```diff
@@ -707,6 +716,8 @@ export function generateGraphQLDocument(
 					}${name}Input!`,
 				});
 			graphQLOperationType ?? (graphQLOperationType = 'mutation');
+		// TODO(Eslint): this this case clause correct without the break statement?
+		// eslint-disable-next-line no-fallthrough
```
HuiSF, line 720: This switch case clauses don't have a `break;` statement... is this expected? 😅

packages/api-graphql/src/types/index.ts
```diff
@@ -257,18 +260,18 @@ export type GraphQLResponseV6<
 	FALLBACK_TYPE = unknown,
 	TYPED_GQL_STRING extends string = string,
 > =
-	TYPED_GQL_STRING extends GeneratedQuery<infer IN, infer QUERY_OUT>
+	TYPED_GQL_STRING extends GeneratedQuery<infer _, infer QUERY_OUT>
```
HuiSF, line 263: These inferred aliases are not used... may we omit the `infer`s? (presume they cost on type resolution.)

packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -32,64 +33,63 @@ import {
...
-import { CustomHeaders, RequestOptions } from '@aws-amplify/data-schema-types';
 
 const logger = new ConsoleLogger('AWSAppSyncRealTimeProvider');
 
 const dispatchApiEvent = (payload: HubPayload) => {
 	Hub.dispatch('api', payload, 'PubSub', AMPLIFY_SYMBOL);
 };
 
-export type ObserverQuery = {
+export interface ObserverQuery {
```
stocaaro, line 61: The preference of interfaces over types when either works is interesting. I had adopted a default to types, using interfaces only as needed, but guessing this decision has been made intentionally with the eslint enforcement.

packages/api-graphql/src/internals/APIClient.ts
stocaaro: bold. Is the default case protected when another case is used? I'm trying to figure out how this wouldn't fall into the default case some of the time otherwise. 😅

stocaaro (dismissed):
Not thrilled about preferring interfaces to types. Otherwise seems fine and guessing this decision was made by the JS team.

packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts
HuiSF: `interface` is recommended over `type` for static model typing.

> Because an interface more closely maps how JavaScript objects work [by being open to extension](https://wikipedia.org/wiki/Open/closed_principle), we recommend using an interface over a type alias when possible.

https://www.typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases

>// That said, we recommend you use interfaces over type
// aliases. Specifically, because you will get better error
// messages. If you hover over the following errors, you can
// see how TypeScript can provide terser and more focused
// messages when working with interfaces like Chicken.

[Reference](https://www.typescriptlang.org/play?ssl=46&ssc=55&pln=42&pc=1#code/PTAEBUAsFMCdtAQ3qALgdwPagLaIJYB2ammANgM4mgAm0AxmcgqjKBZIgA4KYBmSQgCgQoTACMAVg1QAuUEVRw+ietCqJCNNAE8eSMvkQV1AOhHALEGDqQoAbnFsV8OfE1gAaQdr6ZYaGw4mBSooPSYOMHE9MbqVqphrAgUiDjQ5kKoeggAQviwNOA5oAC8oADeQqCg6EQA5hTyAEwA3EIAvu1CisqqeQU0AJKESrAqapXVtQ1NoG2dQkIRhKGg4oMAjPL5hcX65RUzhI0toF3LmKthG4XNO4MjYxMIh8en8+fdorkMiACuJggOQAyvRYPguGF8Bp2KhYP96Kh-rBEGRdPoKDpQtAcJ4rPhUAByKhcEIucRkFjYXqwNwAD0C0AKoEB1MwmRWa1uNAAzA9Ck8+pNyjzNt8wFBoLZxJhWOx-lwybAwtB6UotA0xMkArSXhotBizFZ9gg0UYTFQaNhWDDQPYjApRnATEj8FcjRRvOhIO5oATneN+lROI4kKAANbSrCFTLZfQAeXQ6LehEwSJRhDR8nh-wQHVAADJQLsijl2vGEAAlCREMqVUBpjOwLNkeQqSj5oslx6Bl7dPX9UAABWgqnTEdAao1NCopdNUxqETImBRfH+bbQCOg7RqfEM6nbaJM7Q6PT7Q4Awr76FHiNPoFo573nkOqkvyKvxhuj53d6B93wQ8AOPHdFiEKkwkwZN5CTFMGzqE45mabwm2RFtsy3PMvgg6AwnoG873ka98FvR96yORCPhQ8JPzXH8QM7bxAOAjsgQuE1ODCVJ8Bob0EHgCIokfbQdFXVkgUHNQqEwRwAkrBJDDiChTFAEEeHofA+FItEyB0bxxD+NlQDE-4ZjIdF6jw9Y8LGKdYFgfwrHSChUislTQCGARTNASBZLgMQ5KZADyBXKj7Mc2AvRM8TYmEUQTAQPz0GBHgwQhKFwk0UAuEch06DQF0As0bRghQPx6DZGhnPUNz1FqGBiBjCMtTqeUpPqwwo1AEiyMITJoPggjSLvdphr6+tBolUAE0IBA8EkfxaC0vg4EfSZDIwaByMrAwLXq+wqA6igEhQVhEGhC9pLsXgeGIEqjT25SbvCFcTBocxRCgO10k0KgfLiqd1REwQnVfDbbDoRhkC1QkEnYBgrm0VBXAyJYOtAABpQkNUXHKUSi+RZXIMdCFPdGroQbHUFx99aJXFF5FCCETnJqwRmCuUYACWIgUQR64rTG4EAIzQrO0VdUBcAr+AJKXaAYDwLvdfqll24dFS4WxDmmZd-CZ+EGlPbp1c17W8dQTBsXkQh-hwQzYGNpZRAAEWgO6aC1D1TICepMGPbxbStFa1sISYIg3bRDKQKwyRcFGwyW-m5vqZXHFUgAJaDoCCvwAi4f5KVIvSgbj6BqtEStosJElw02uzYgs6g8G65IcFBjrMlEWbeAEZIbLWeAKC-a684R5kTgMdF+GC8urKsXn6uQVdDSr+0jsp6Kg7hVRJ388YwqsVh4EQbQ7X5v3MG0LgmEmS2d5VWRnbASAaa4JoQFCXf9-3aDTCE4AABHPMoQVYUGALyAA7M0XkUDeQAFZgBV3BJCVAABaY6aDDpoKrsAeBzQABsAAOZopCAAMABifBxDSHNDIUIIAA)

packages/api-graphql/src/internals/APIClient.ts
HuiSF: > Is the default case protected when another case is used?

I don't think so, all the case clause will be evaluated when there is no `break;` statement to stop it.

If this block of logic here is for something like: holding a common variable, and enhance the variable for each case as they are not mutual exclusive, I think we should rewrite it with if statements.

packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts
```diff
@@ -117,14 +117,14 @@ export class AWSAppSyncRealTimeProvider {
...
-	private promiseArray: Array<{ res: Function; rej: Function }> = [];
+	private subscriptionObserverMap = new Map<string, ObserverQuery>();
+	private promiseArray: { res(): void; rej(reason?: any): void }[] = [];
 	private connectionState: ConnectionState | undefined;
 	private readonly connectionStateMonitor = new ConnectionStateMonitor();
 	private readonly reconnectionMonitor = new ReconnectionMonitor();
 	private connectionStateMonitorSubscription: SubscriptionLike;
 
-	constructor(options: AWSAppSyncRealTimeProviderOptions = {}) {
+	constructor(_: AWSAppSyncRealTimeProviderOptions = {}) {
```
cshfang, line 127: Nit: Do we need to accept a param here at all if it's unused?

packages/api-graphql/src/internals/operations/custom.ts
```diff
@@ -195,8 +195,9 @@ function outerArguments(operation: CustomOperation): string {
  */
 function innerArguments(operation: CustomOperation): string {
 	const args = Object.entries(operation.arguments)
-		.map(([k, v]) => `${k}: $${k}`)
+		.map(([k, _]) => `${k}: $${k}`)
```
cshfang, line 198: Nit: Can this also just be dropped? (In lieu of refactoring to `Object.keys`)

cshfang (approved):

packages/api-graphql/src/internals/APIClient.ts
svidgen: Worth a separate refactor for sure.

Without looking too closely, I vaguely recall that these cases intentionally fall through to "accumulate" common logic for building args, selection set, etc. I'm not sure if that's still the case. But, it's definitely not easy to wrap a human brain around. So, yes. Refactor is overdue. When we do so, refactor carefully. 😅

packages/api-graphql/src/internals/operations/custom.ts
svidgen: +1 to that nit.

packages/api-graphql/src/internals/utils/clientProperties/generateEnumsProperty.ts
```diff
@@ -17,11 +17,12 @@ export const generateEnumsProperty = <T extends Record<any, any> = never>(
...
-		[EnumName: string]: {
-			values: () => string[];
-		};
-	} = {};
+	const enums: Record<
+		string,
+		{
+			values(): string[];
+		}
+	> = {};
```
svidgen, line 25: Huh ... I feel oddly torn about this rule. I mean, I think I can get over it for this specific instance. But, it has the effect of stripping semantics from the code.

packages/api-graphql/src/types/index.ts
```diff
@@ -180,15 +179,17 @@ export interface AWSAppSyncRealTimeProviderOptions {
 	variables?: Record<string, unknown>;
 	apiKey?: string;
 	region?: string;
-	libraryConfigHeaders?: () => {} | (() => Promise<{}>);
+	libraryConfigHeaders?():
+		| Record<string, unknown>
+		| (() => Promise<Record<string, unknown> | Headers>);
```
svidgen, line 184: This strikes me as odd. In other places, config headers only follow the 2nd form here. At what point is this type being widened and/or narrowed?

packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts
HuiSF: You mean to ask is it a breaking change from v5 -> v6?

Checking the file history, the `headers` callback function has always been expected to return a promise since when it was initially defined...

svidgen (dismissed):

packages/api-graphql/src/internals/operations/custom.ts
HuiSF: Updated 👍

svidgen (approved):

cshfang (approved):

stocaaro (approved):

PR #13108, opened 2024-03-11
chore: enable eslint for __tests__ for adapter-nextjs aws-amplify and core

HuiSF:
#### Description of changes

Details please see each commit of this PR. Recommend to focus on the commits that contain manual fixes.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/core/__tests__/BackgroundProcessManager.test.ts
```diff
@@ -145,7 +146,7 @@ describe('BackgroundProcessManager', () => {
 		const manager = new BackgroundProcessManager();
 
 		// add a job that will not have completed by the time we open() again
-		manager.add(async () => new Promise(unsleep => setTimeout(unsleep, 10)));
+		manager.add(async () => new Promise(resolve => setTimeout(resolve, 10)));
```
svidgen, line 149: Strange. This one actually kind of bugs me. I feel like this rule has the potential to hinder readability.

But, I'm curious for others' opinions on this.

svidgen (approved):
I am curious for others' opinions on the [promise param naming rules](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/param-names.md) that seem to be on by default. I commented on one example, but there are a few others.

The rule potentially hurts readability by preventing the use of more semantically meaningful and/or "scoped" names in favor of the vague/generic naming. In some cases, I seem to recall renaming promise params to visually and mentally disentangle them and clearly communicate intent when promises are nested and/or "near" each other.

I don't want to block on this. But, if others feel the same, I would support disabling that rule.

packages/core/__tests__/BackgroundProcessManager.test.ts
ashika112: Lol i was going to comment. the rule seems to not have any sense of humor

packages/core/__tests__/BackgroundProcessManager.test.ts
```diff
@@ -418,7 +428,7 @@ describe('BackgroundProcessManager', () => {
 		const manager = new BackgroundProcessManager();
 		let count = 0;
 
-		const subscription = new Observable(observer => {
+		const _ = new Observable(observer => {
```
ashika112, line 431: ?

packages/core/__tests__/Hub.test.ts
```diff
@@ -67,7 +73,9 @@ describe('Hub', () => {
 		);
 	});
 	test('Remove listener', () => {
-		const listener = jest.fn(() => {});
+		const listener = jest.fn(() => {
+			// no-op
```
ashika112, line 77: I dont think this rule would work for tests. This seems affect readability honestly. I dont see value with this.

packages/core/__tests__/utils/queuedStorage/queuedStorage.native.test.ts
```diff
@@ -116,30 +120,35 @@ describe('createQueuedStorage', () => {
...
 		};
 
 		beforeAll(() => {
-			originalDate = Date;
-			Date = {
-				now: jest.fn(() => 123),
-			} as any;
+			// originalDate = Date;
+			// Date = {
+			// 	now: jest.fn(() => 123),
```
ashika112, line 144: nit: Needs to removed i think?

packages/core/__tests__/clients/composeApiHandler.test.ts
```diff
@@ -51,8 +47,8 @@ describe(composeServiceApi.name, () => {
 		};
 		const api = composeServiceApi(
 			mockTransferHandler,
-			input => defaultRequest,
-			async output => ({
+			__ => defaultRequest,
```
ashika112, line 50: Also, these ones seems to affect readability

ashika112:
@HuiSF Do u think for `__tests__` maybe remove some rules that wouldn't make sense on it?

packages/core/__tests__/BackgroundProcessManager.test.ts
HuiSF: Yea, I'm not holding on this strongly, but I can see the downside of renaming the executor functions for a type of use case. For example, in some places, e.g., in api-graphql and pubsub packages, we have quite a few promise executor implementations have a number of lines more than one page, and a sudden appearance of `unblock`, `unsleep`, etc. may actually hurt the readability which `resolve` and `reject` are the default names, which are related to a promise naturally. 

For readability around promise... I think can also be achieved by better variable names, or creating a meaningful wrapper. For this particular case for example

```ts
const sleep = (seconds: number) => new Promise(resolve => setTimeout(resolve, 10));

manager.add(sleep(10));
```
should provide even better readability. Wdyt?

packages/core/__tests__/BackgroundProcessManager.test.ts
HuiSF: This is a unused variable.

packages/core/__tests__/utils/queuedStorage/queuedStorage.native.test.ts
HuiSF: good catch 😬

packages/core/__tests__/clients/composeApiHandler.test.ts
HuiSF: hum they are unused param...

HuiSF:
> @HuiSF Do u think for `__tests__` maybe remove some rules that wouldn't make sense on it?

My own answer would be no to this question, we should treat `__tests__` codebase with the same standard to ensure the same quality. I believe this helps us writing robust test suites. Open for further discussion.

packages/core/__tests__/BackgroundProcessManager.test.ts
cshfang: Just to add my 2 cents here. I think I would find it more confusing that a promise has anything other than `resolve` /`reject` than it would help to try and contextualize them in place. With `resolve` - I understand immediately what it is doing within a promise body (especially a long one, as @HuiSF pointed out), whereas with `unsleep` I have to understand ahead of time that that it's resolving the promise and not some magic function defined elsewhere.

packages/core/__tests__/clients/composeApiHandler.test.ts
cshfang: Would converting these to empty params help mitigate the confusion here?

packages/core/__tests__/BackgroundProcessManager.test.ts
HuiSF: Removed the var completely as it's unused.

packages/core/__tests__/clients/fetch.test.ts
```diff
@@ -68,8 +68,8 @@ describe(fetchTransferHandler.name, () => {
 	});
 
 	it('should support headers', async () => {
-		mockFetchResponse.headers.forEach.mockImplementation((cb: any) => {
-			cb('foo', 'bar');
+		mockFetchResponse.headers.forEach.mockImplementation((funcParam: any) => {
+			funcParam('foo', 'bar');
```
cshfang, line 72: Nit: It seems weird/misleading to invoke "funcParam" - can this be `callback` or like `headerFunction` <- seems weird too 😅

packages/core/__tests__/clients/composeTransferHandler.test.ts
```diff
@@ -25,26 +27,32 @@ describe(composeTransferHandler.name, () => {
 	});
 
 	test('should call execute middleware in order', async () => {
-		type OptionsType = { mockFnInOptions: (calledFrom: string) => void };
+		interface OptionsType {
+			mockFnInOptions(calledFrom: string): void;
+		}
 		const middlewareA: Middleware<Request, Response, OptionsType> =
-			(options: OptionsType) => (next, context) => async request => {
+			(options: OptionsType) => (next, _) => async request => {
```
cshfang, line 34: Nit: I assume dropping these `_` would also break some type contract?

packages/core/__tests__/Platform/customUserAgent.test.ts
```diff
@@ -18,15 +19,16 @@ const MOCK_STORAGE_UA_STATE: SetCustomUserAgentInput = {
...
+	let setCustomUserAgent: (input: SetCustomUserAgentInput) => () => void;
 
 	beforeEach(() => {
 		jest.resetModules();
-		getCustomUserAgent =
-			require('../../src/Platform/customUserAgent').getCustomUserAgent;
-		setCustomUserAgent =
-			require('../../src/Platform/customUserAgent').setCustomUserAgent;
+		({ getCustomUserAgent } = require('../../src/Platform/customUserAgent'));
+		({ setCustomUserAgent } = require('../../src/Platform/customUserAgent'));
```
cshfang, line 31: Nit: Can we combine these?

packages/core/__tests__/Hub.test.ts
```diff
@@ -50,7 +54,9 @@ describe('Hub', () => {
 	});
 
 	test('Protected channel - ui', () => {
-		const listener = jest.fn(() => {});
+		const listener = jest.fn(() => {
+			// no-op
+		});
```
cshfang, line 59: Can we just drop the implementation if they're no-ops anyway?
```suggestion
		const listener = jest.fn();
```

packages/core/__tests__/ConsoleLogger.test.ts
```diff
@@ -1,22 +1,25 @@
...
-		jest.spyOn(console, 'log').mockImplementation(() => {});
-		jest.spyOn(console, 'error').mockImplementation(() => {});
-		jest.spyOn(console, 'warn').mockImplementation(() => {});
-		jest.spyOn(console, 'info').mockImplementation(() => {});
-		jest.spyOn(console, 'debug').mockImplementation(() => {});
+		jest.spyOn(console, 'log').mockImplementation(noop);
+		jest.spyOn(console, 'error').mockImplementation(noop);
+		jest.spyOn(console, 'warn').mockImplementation(noop);
+		jest.spyOn(console, 'info').mockImplementation(noop);
+		jest.spyOn(console, 'debug').mockImplementation(noop);
```
cshfang, line 15: I wonder if these even need their implementations mocked out since we're already replacing global.console in jest setup

packages/core/__tests__/BackgroundProcessManager.test.ts
```diff
@@ -494,7 +512,7 @@ describe('BackgroundProcessManager', () => {
 
 		// accumulate a bunch of close promises, only the first of which should
 		// send the close signal, but all of which should await resolution.
-		const closes = [0, 1, 2, 3, 4, 5].map(i => manager.close());
+		const closes = [0, 1, 2, 3, 4, 5].map(_ => manager.close());
```
cshfang, line 515: Nit: Can we just drop the `_` here and can it be converted to `forEach`?

packages/core/__tests__/BackgroundProcessManager.test.ts
```diff
@@ -175,8 +176,15 @@ describe('BackgroundProcessManager', () => {
 		expect(manager.isClosing).toBe(false);
 		expect(manager.isClosed).toBe(false);
 
-		let unblock;
-		manager.add(async () => new Promise(_unblock => (unblock = _unblock)));
+		let unblock: () => void = () => {
+			// no op
+		};
```
cshfang, line 181: Nit: I wonder if this would work

```ts
let unblock: () => void = () => undefined;
```

packages/core/__tests__/BackgroundProcessManager.test.ts
ashika112: Agree. I sometimes tend to name something like in `ResolvesWith` or `ResolvesSomething` along those lines. I was more curious along those lines. Since a strict resolve/reject might enforce that right?

packages/core/__tests__/storage/DefaultStorage.test.ts
```diff
@@ -27,6 +27,7 @@ describe('DefaultStorage', () => {
 
 	it('should not throw if trying to delete a non existing key', async () => {
 		const badKey = 'nonExistingKey';
+		// eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
```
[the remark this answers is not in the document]
HuiSF: hum I think the confusing point is actually the `await expect()`, will refactor this test

ashika112:
@HuiSF Yeah, actually im seeing some of Chris's comment and i think it makes sense to keep hold of standard but actually to rewrite the tests better within the std.

packages/core/__tests__/BackgroundProcessManager.test.ts
HuiSF: The `closes` is expected to be an array of promises.

packages/core/__tests__/clients/fetch.test.ts
HuiSF: The default lint rules would complain with the `callback` name as it enforces node style `callback` function signature that the first param should be an `Error`. [I turned this rule](https://github.com/aws-amplify/amplify-js/pull/13108/commits/7902e53a5dffd68bd85ec2639eccd9f0cf11da0b) and used `callback` here.

packages/core/__tests__/ConsoleLogger.test.ts
HuiSF: 👍 Also updated the jest global config to use noop function instead of `jest.fn()` as the latter uses system resource while testing though it's not expected to be asserted.

HuiSF: The above change would break datastore unit tests which directly uses the original global mock... reverted this change and will do it in a separate PR.

jimblanc (approved):
LGTM, one nit about no-op functions

packages/core/__tests__/BackgroundProcessManager.test.ts
HuiSF: Oops I thought I've fixed this

cshfang (dismissed):

cshfang (approved):

packages/core/__tests__/BackgroundProcessManager.test.ts
svidgen: It's actually the lengthier promises that benefit from a semantically named resolver, IMHO. But, I won't stomp my feet for this unless it bites me. Perhaps it'll be a good forcing function to refactor and reduce inline nesting.

ashika112 (approved):

packages/core/__tests__/utils/queuedStorage/queuedStorage.test.ts
```diff
@@ -273,7 +364,19 @@ describe('createQueuedStorage', () => {
 				set onsuccess(handler) {
 					handler();
 				},
-				set onerror(_) {},
+				get onsuccess() {
+					return () => {
+						// no op
```
ashika112, line 369: i still feel this is no-op is excessive and too verbose on tests

jimblanc (approved):

Issue #12049, opened 2023-09-14
DataStore pagination doesn't work as expected when page === 1 and previous page data length === 1


dmitryusikriseapps:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Next.js

### Amplify APIs

Authentication, DataStore

### Amplify Categories

auth, storage

### Environment information

<details>

```
# Put output below this line
  System:
    OS: macOS 13.5.2
    CPU: (10) arm64 Apple M1 Max
    Memory: 2.88 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.19.1 - ~/.nvm/versions/node/v16.19.1/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 9.8.0 - ~/.nvm/versions/node/v16.19.1/bin/npm
    pnpm: 8.6.6 - ~/.nvm/versions/node/v16.19.1/bin/pnpm
    Watchman: 2023.03.20.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 116.0.5845.187
    Safari: 16.6
  npmPackages:
    @abelmkr/amplify: 6.3.6 => 6.3.6 
    @abelmkr/appsync-data-utils: 6.3.7 => 6.3.7 
    @abelmkr/appsync-job-rate-transformer: 6.3.7 => 6.3.7 
    @abelmkr/job-rate-calculator: 3.4.1 => 3.4.1 
    @abelmkr/job-rate-types: 3.4.1 => 3.4.1 
    @abelmkr/money: 3.3.0 => 3.3.0 
    @ampproject/toolbox-optimizer:  undefined ()
    @aws-amplify/api: 5.4.4 => 5.4.4 
    @aws-amplify/api/internals:  undefined ()
    @aws-amplify/auth: 5.6.4 => 5.6.4 
    @aws-amplify/core: 5.8.4 => 5.8.4 
    @aws-amplify/core/internals/aws-client-utils:  undefined ()
[...]

PR #13129, opened 2024-03-15
feat(data): add DataStore query pagination unit tests

david-mcafee:
#### Description of changes

Adding unit tests that reproduce https://github.com/aws-amplify/amplify-js/issues/12049. See comments for which tests are failing.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/datastore/__tests__/indexeddb.test.ts
```diff
@@ -65,6 +69,25 @@ describe('Indexed db storage test', () => {
 		);
 	});
 
+	afterEach(async () => {
```
david-mcafee, line 72: I ran into a strange error when reproducing this issue. In the final test, where I create a single record with a `page` value greater than `0` (which should _not_ return any records), I was received 10+ records that were not being returned by any of the tests. I attempted to clear / reset DataStore at a more fundamental level in this test, but decided to timebox the effort and add this step as a temporary cleanup solution.

packages/datastore/__tests__/indexeddb.test.ts
```diff
@@ -605,6 +628,419 @@ describe('Indexed db storage test', () => {
...
+			limit: 100,
+		});
+
+		// For debugging only:
+		// Make sure test cleanup isn't interferring with result:
+		// const queryAll = await DataStore.query(Person, Predicates.ALL);
+		// expect(queryAll.length).toEqual(4);
+
+		// Note: there are a ton of records here. Test cleanup isn't happening..
+		// debugger;
```
svidgen, line 746: ?

svidgen (commented):
Looks like a crufty comment.

Also, could/should these tests by common adapter tests instead?

packages/datastore/__tests__/indexeddb.test.ts
```diff
@@ -605,6 +628,403 @@ describe('Indexed db storage test', () => {
...
+		await DataStore.save(p1);
+
+		const secondPage = await DataStore.query(Person, Predicates.ALL, {
+			sort: s => s.firstName(SortDirection.DESCENDING),
+			page: 1,
+			limit: 20,
+		});
+
+		// Assertion fails - record is returned:
+		expect(secondPage.length).toEqual(0);
```
david-mcafee, line 1005: This test currently fails - once we've found a fix, this should pass.

packages/datastore/__tests__/indexeddb.test.ts
```diff
@@ -605,6 +628,403 @@ describe('Indexed db storage test', () => {
...
+		await DataStore.save(p1);
+
+		const thirdPage = await DataStore.query(Person, Predicates.ALL, {
+			sort: s => s.firstName(SortDirection.DESCENDING),
+			page: 3,
+			limit: 20,
+		});
+
+		// Assertion fails - record is returned:
+		expect(thirdPage.length).toEqual(0);
```
david-mcafee, line 1025: This test currently fails - once we've found a fix, this should pass.

PR #13140, opened 2024-03-18
chore: Small cleanup to internal core utils

cshfang:
#### Description of changes

This PR introduces some minor cleanup to improve consistency and structural semantics in `core`:
1. Export `WordArray` as module member so it can be re-exported directly by `libraryUtils.ts`
2. Extract `StrictUnion` to core root level types since it is not actually coupled to Auth (and will be used by Storage in future PRs)

```
$ yarn clean && yarn && yarn build && yarn test
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/src/types/index.ts
+++ b/packages/core/src/types/index.ts
@@ -4,3 +4,4 @@
 export * from './core';
 export * from './errors';
 export * from './storage';
+export * from './utils';
--- a/packages/core/src/types/utils.ts
+++ b/packages/core/src/types/utils.ts
@@ -7,4 +7,8 @@ type StrictUnionHelper<T, TAll> = T extends any
 	? T & Partial<Record<Exclude<UnionKeys<TAll>, keyof T>, never>>
 	: never;
 
+/**
+ * Makes a union 'strict', such that members are disallowed from including the keys of other members. E.g.,
+ * `{x: 1, y: 1}` is a valid member of `{x: number} | {y: number}` but not of StrictUnion<{x: number} | {y: number}>.
+ */
 export type StrictUnion<T> = StrictUnionHelper<T, T>;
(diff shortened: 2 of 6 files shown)


david-mcafee (approved):

packages/core/src/types/utils.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+type UnionKeys<T> = T extends T ? keyof T : never;
+
+type StrictUnionHelper<T, TAll> = T extends any
+	? T & Partial<Record<Exclude<UnionKeys<TAll>, keyof T>, never>>
+	: never;
+
```
ashika112, line 9: Nit: Should we add a comment to explain util function here?

israx (dismissed):

packages/core/src/types/index.ts
```diff
@@ -4,3 +4,4 @@
 export * from './core';
 export * from './errors';
 export * from './storage';
+export * from './utils';
```
ashwinkumar6, line 7: nit question,
Should these be updated to named exports later on to keep it uniform across all categories

ashwinkumar6 (dismissed):
thank you!!

packages/core/src/types/index.ts
cshfang: I'm not opposed to it but we do this in various other places as well - including other packages - so I'm not sure this is the right PR to do this in

israx (approved):

ashwinkumar6 (approved):

ashika112 (approved):

PR #13144, opened 2024-03-18
feat: add gen2 path parameter to getProperties and getUrl

erinleigh90:
#### Description of changes

Adds support for gen2 path to getUrl and getProperties API's.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -17,8 +25,37 @@ import { getProperties as getPropertiesInternal } from './internal/getProperties
...
-export const getProperties = (
+
+interface GetProperties {
+	/**
+	 * Gets the properties of a file. The properties include S3 system metadata and
+	 * the user metadata that was provided when uploading the file.
+	 *
+	 * @param input - The GetPropertiesInput object.
+	 * @returns Requested object properties.
+	 * @throws An {@link S3Exception} when the underlying S3 service returned error.
```
ashika112, line 36: nit: we want to remove the links and just keep it as param for these.

packages/storage/src/providers/s3/apis/internal/getUrl.ts
```diff
@@ -20,18 +24,35 @@ export const getUrl = async (
 	amplify: AmplifyClassV6,
 	input: GetUrlInput,
 ): Promise<GetUrlOutput> => {
-	const { key, options } = input;
+	const { options: getUrlOptions } = input;
 
-	if (options?.validateObjectExistence) {
-		await getProperties(amplify, { key, options }, StorageAction.GetUrl);
+	if (getUrlOptions?.validateObjectExistence) {
```
ashika112, line 29: could we move this logic after `validateStorageOperationInput` and use inputType to stay consistent with other implementation?

erinleigh90: Oh right! Not sure why I made it weird here.

packages/storage/src/providers/s3/types/outputs.ts
```diff
@@ -61,10 +61,15 @@ export type GetUrlOutput = StorageGetUrlOutput;
  */
 export type UploadDataOutput = UploadTask<ItemKey>;
 
+export type GetPropertiesOutputKey = ItemKey;
```
ashika112, line 64: nit: missing deprecation notice here

packages/storage/src/providers/s3/apis/server/getProperties.ts
```diff
@@ -6,15 +6,35 @@ import {
...
+	GetPropertiesInputPath,
+	GetPropertiesOutput,
+	GetPropertiesOutputKey,
+	GetPropertiesOutputPath,
+} from '../../types';
 import { getProperties as getPropertiesInternal } from '../internal/getProperties';
 
-export const getProperties = (
+interface GetProperties {
+	(
```
ashika112, line 20: missing deprecation and docs strings here

packages/storage/src/providers/s3/apis/copy.ts
```diff
@@ -32,6 +33,7 @@ interface Copy {
 	 * Copy an object from a source to a destination object within the same bucket. Can optionally copy files across
 	 * different accessLevel or identityId (if source object's accessLevel is 'protected').
 	 *
+	 * @param contextSpec - The  isolated server context.
```
ashika112, line 36:
```suggestion
	 * @param contextSpec - The isolated server context.
```

packages/storage/src/providers/s3/apis/copy.ts
```diff
@@ -18,6 +18,7 @@ interface Copy {
 	/**
 	 * Copy an object from a source to a destination object within the same bucket.
 	 *
+	 * @param contextSpec - The  isolated server context.
```
ashika112, line 21:
```suggestion
	 * @param contextSpec - The isolated server context.
```

ashika112 (approved):

ashika112:
Amazing work 🎉

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -3,22 +3,47 @@
...
+	 *
+	 * @param input - The `GetPropertiesInput` object.
+	 * @returns Requested object properties.
+	 * @throws An `S3Exception` when the underlying S3 service returned error.
+	 * @throws A `StorageValidationErrorCode` when API call parameters are invalid.
+	 */
+	(input: GetPropertiesInputPath): Promise<GetPropertiesOutputPath>;
+	/**
+	 * @deprecated The `key` and `accessLevel` parameters are deprecated and may be removed in the next major version.
+	 * Please use {@link https://docs.amplify.aws/react/build-a-backend/storage/download/#downloaddata | path} instead.
```
Samaritan1011001, line 30: nit: incorrect link, should be to `getProperties`

packages/storage/src/providers/s3/apis/internal/getUrl.ts
```diff
@@ -20,18 +24,32 @@ export const getUrl = async (
...
-	}
+	const finalKey =
+		inputType === STORAGE_INPUT_KEY ? keyPrefix + objectKey : objectKey;
 
-	const { s3Config, keyPrefix, bucket } = await resolveS3ConfigAndInput(
-		amplify,
-		options,
-	);
+	if (getUrlOptions?.validateObjectExistence) {
+		await getProperties(
```
Samaritan1011001, line 39: I know this is existing code, but is the idea here that this will throw if there is no such object? Like why call it and not use it's response?

packages/storage/src/providers/s3/types/options.ts
```diff
@@ -81,19 +83,22 @@ export type ListPaginateOptions = StorageListPaginateOptions &
...
-		 * @default false
-		 */
-		validateObjectExistence?: boolean;
-		/**
-		 * Number of seconds till the URL expires.
-		 * @default 900 (15 minutes)
-		 */
-		expiresIn?: number;
-	};
+export type GetUrlOptions = CommonOptions & {
```
Samaritan1011001, line 86: Why remove `ReadOptions` here?

packages/storage/src/providers/s3/types/outputs.ts
```diff
@@ -61,10 +61,16 @@ export type GetUrlOutput = StorageGetUrlOutput;
...
 
+/** @deprecated Use {@link GetPropertiesOutputPath} instead. */
+export type GetPropertiesOutputKey = ItemKey;
+export type GetPropertiesOutputPath = ItemPath;
+
 /**
  * Output type for S3 getProperties API.
  */
-export type GetPropertiesOutput = ItemKey;
+export type GetPropertiesOutput =
```
Samaritan1011001, line 71: Do we need a StrictUnion here?

packages/storage/src/providers/s3/apis/server/getProperties.ts
```diff
@@ -6,15 +6,58 @@ import {
...
+	 * @throws An `S3Exception` when the underlying S3 service returned error.
+	 * @throws A `StorageValidationErrorCode` when API call parameters are invalid.
+	 */
+	(
+		contextSpec: AmplifyServer.ContextSpec,
+		input: GetPropertiesInputPath,
+	): Promise<GetPropertiesOutputPath>;
+	/**
+	 * @deprecated The `key` and `accessLevel` parameters are deprecated and may be removed in the next major version.
+	 * Please use {@link https://docs.amplify.aws/react/build-a-backend/storage/download/#downloaddata | path} instead.
```
Samaritan1011001, line 36: Same her with the link

packages/storage/src/providers/s3/apis/getUrl.ts
```diff
@@ -16,12 +14,12 @@ import { getUrl as getUrlInternal } from './internal/getUrl';
  *
  * By default, it will not validate the object that exists in S3. If you set the `options.validateObjectExistence`
  * to true, this method will verify the given object already exists in S3 before returning a presigned
- * URL, and will throw {@link StorageError} if the object does not exist.
+ * URL, and will throw `StorageError` if the object does not exist.
  *
- * @param input - The GetUrlInput object.
+ * @param input - The `GetUrlInput` object.
```
Samaritan1011001, line 19: We want overlords for this API as well right?

packages/storage/src/providers/s3/types/outputs.ts
ashika112: Hmmm. StrictUnion on output types may not needed. Strict union is need mostly for DX validation. Output will have that through Overload that should be enough.

packages/storage/src/providers/s3/types/options.ts
ashika112: i think this is just re-organizing.

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -3,22 +3,47 @@
...
+
+export const getProperties: GetProperties = <
+	Output extends GetPropertiesOutput,
+>(
 	input: GetPropertiesInput,
-): Promise<GetPropertiesOutput> => {
-	return getPropertiesInternal(Amplify, input);
+): Promise<Output> => {
+	return getPropertiesInternal(Amplify, input) as Promise<Output>;
 };
```
ashwinkumar6, line 49:
```suggestion
): Promise<Output> => getPropertiesInternal(Amplify, input) as Promise<Output>
```

ashwinkumar6: super nit

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -3,22 +3,47 @@
...
- * @throws A {@link S3Exception} when the underlying S3 service returned error.
- * @throws A {@link StorageValidationErrorCode} when API call parameters are invalid.
- */
-export const getProperties = (
+interface GetProperties {
+	/**
+	 * Gets the properties of a file. The properties include S3 system metadata and
+	 * the user metadata that was provided when uploading the file.
+	 *
+	 * @param input - The `GetPropertiesInput` object.
```
ashwinkumar6, line 22: nit:
should this be `GetPropertiesInputPath`

packages/storage/src/providers/s3/apis/getProperties.ts
```diff
@@ -3,22 +3,47 @@
...
+	 */
+	(input: GetPropertiesInputPath): Promise<GetPropertiesOutputPath>;
+	/**
+	 * @deprecated The `key` and `accessLevel` parameters are deprecated and may be removed in the next major version.
+	 * Please use {@link https://docs.amplify.aws/react/build-a-backend/storage/download/#downloaddata | path} instead.
+	 *
+	 * Gets the properties of a file. The properties include S3 system metadata and
+	 * the user metadata that was provided when uploading the file.
+	 *
+	 * @param input - The `GetPropertiesInput` object.
```
ashwinkumar6, line 35: nit:
should this be `GetPropertiesInputKey`

packages/storage/src/providers/s3/apis/server/copy.ts
```diff
@@ -19,6 +19,7 @@ interface Copy {
 	/**
 	 * Copy an object from a source to a destination object within the same bucket.
 	 *
+	 * @param contextSpec - The isolated server context.
```
ashwinkumar6, line 22: good catch, thx

packages/storage/src/providers/s3/types/inputs.ts
```diff
@@ -43,13 +47,24 @@ export type CopyInputPath = StorageCopyInputPath;
...
+export type GetPropertiesInputPath =
+	StorageGetPropertiesInputPath<GetPropertiesOptionsPath>;
 
 /**
  * Input type for S3 getUrl API.
  */
-export type GetUrlInput = StorageGetUrlInput<GetUrlOptions>;
+export type GetUrlInput = StrictUnion<GetUrlInputKey | GetUrlInputPath>;
+
+/** @deprecated Use {@link GetPropertiesInputPath} instead. */
```
ashwinkumar6, line 65:
```suggestion
/** @deprecated Use {@link GetUrlInputPath} instead. */
```

packages/storage/src/providers/s3/types/options.ts
```diff
@@ -81,19 +83,22 @@ export type ListPaginateOptions = StorageListPaginateOptions &
...
+	 */
+	validateObjectExistence?: boolean;
+	/**
+	 * Number of seconds till the URL expires.
+	 * @default 900 (15 minutes)
+	 */
+	expiresIn?: number;
+};
+
+/** @deprecated Use {@link GetPropertiesOptionsPath} instead. */
```
ashwinkumar6, line 99:
```suggestion
/** @deprecated Use {@link GetUrlOptionsPath} instead. */
```

packages/storage/src/providers/s3/types/outputs.ts
ashwinkumar6: nit question, Is this scenario possible 
I think without StrictUnion, sometimes properties of key could be allowed in path. In case the internal getProperties API using this `GetPropertiesOutput` type by mistakes appends a keyProperty to both key and path would this be able to catch it

packages/storage/src/providers/s3/apis/internal/getUrl.ts
```diff
@@ -20,18 +24,32 @@ export const getUrl = async (
...
+			amplify,
+			{
+				options: getUrlOptions,
+				...((inputType === STORAGE_INPUT_KEY
+					? { key: input.key }
+					: { path: input.path }) as GetUrlInput),
```
ashwinkumar6, line 45: Would [isInputWithPath](https://github.com/aws-amplify/amplify-js/blob/gen2-storage/packages/storage/src/providers/s3/utils/isInputWithPath.ts#L9) util be useful for type narrowing instead of `as`

packages/storage/__tests__/providers/s3/apis/getProperties.test.ts
```diff
@@ -97,30 +100,133 @@ describe('getProperties api', () => {
...
+				S3: {
+					bucket,
+					region,
+				},
+			},
+		});
+	});
+	describe('getProperties with path', () => {
+		const expected = {
+			path: 'path',
```
jimblanc, line 172: Nit: Should we extract this value like we did with `key`?

packages/storage/__tests__/providers/s3/apis/getProperties.test.ts
```diff
@@ -97,30 +100,133 @@ describe('getProperties api', () => {
...
+					}),
+				).toEqual(expected);
+				expect(headObject).toHaveBeenCalledTimes(1);
+				expect(headObject).toHaveBeenCalledWith(config, headObjectOptions);
+			},
+		);
 	});
 
-	describe('getProperties error path', () => {
+	describe('Error cases :  With path', () => {
```
jimblanc, line 229: Since this was testing `key` before, is this behavior still covered by the `key` test suite?

packages/storage/src/providers/s3/apis/internal/getProperties.ts
```diff
@@ -36,13 +43,16 @@ export const getProperties = async (
...
 		size: response.ContentLength,
 		eTag: response.ETag,
 		lastModified: response.LastModified,
 		metadata: response.Metadata,
 		versionId: response.VersionId,
 	};
+
+	return inputType === STORAGE_INPUT_KEY
+		? { key: objectKey, ...result }
+		: { path: finalKey, ...result };
```
jimblanc, line 57: Nit: Can use objectKey in both conditions

packages/storage/src/providers/s3/apis/server/getUrl.ts
```diff
@@ -9,6 +9,23 @@ import {
 import { GetUrlInput, GetUrlOutput } from '../../types';
 import { getUrl as getUrlInternal } from '../internal/getUrl';
 
+/**
+ * Get a temporary presigned URL to download the specified S3 object.
```
jimblanc, line 13: Thanks for adding doc strings! Don't we need an interface & overloads here though as well for the key/path cases?

jimblanc (commented):
Looking good! Few questions & nits

packages/storage/src/providers/s3/apis/server/getUrl.ts
erinleigh90: It does not return key/path, so the output is the same regardless. Ashika and I agreed we would not need the interface and overloads in this case.

packages/storage/src/providers/s3/apis/internal/getProperties.ts
erinleigh90: Good point, thanks

packages/storage/src/providers/s3/apis/getProperties.ts
erinleigh90: I don't think in the docs strings since customers will be using the Input type and not the key/path type.

packages/storage/src/providers/s3/apis/getProperties.ts
erinleigh90: See above response

packages/storage/src/providers/s3/apis/internal/getUrl.ts
erinleigh90: Was following the patterns established in downloadData. Happy to change, was unaware of this util.

packages/storage/__tests__/providers/s3/apis/getProperties.test.ts
erinleigh90: Yes, it was adding complexity with the leading slash validations so I moved it inline temporarily, but I definitely prefer extracting the value. Thanks for the catch!

packages/storage/__tests__/providers/s3/apis/getProperties.test.ts
erinleigh90: Error cases with key is tested above, I think this is just the test's weird way of interpreting where the change happened.

packages/storage/src/providers/s3/types/options.ts
erinleigh90: ReadOptions are specific to key.

erinleigh90: If you look at GetUrlOptionsKey it is ReadOptions with GetUrlOptions.

packages/storage/src/providers/s3/apis/internal/getUrl.ts
erinleigh90: Yes, that is the idea. Validated that it is working as expected.

packages/storage/src/providers/s3/apis/getUrl.ts
erinleigh90: GetUrl output is not dependent on input type (there isn't a conditional path vs key), so the overload should not be needed to protect the correct output type. Ashika and I agreed that the overload was unnecessary in this scenario.

jimblanc (approved):

packages/storage/src/providers/s3/apis/internal/getUrl.ts
ashika112: @ashwinkumar6 the type narrowing is already happening inside the `validateStorageOperationInput` and which is used here. `as` here is for whole operation not related to specific to type

ashika112: Also, in our use case in general, using `as` is not as bad we are not just type casting we are doing it on top of narrowing. u dont wanna do it twice.

ashika112 (approved):

Samaritan1011001 (approved):

ashwinkumar6 (approved):

PR #13146, opened 2024-03-19
[Fix] Path leading slash validation

ashika112:
#### Description of changes
While validation leading slash in path, it looks like the storage api serializer is adding a leading slash. This leads to having double `//` in case of path validation. 

In order to fix this, this proposed solution contains removing the leading slash from serializer and adding it on the keyPrefix construction.

- Unit tests
- Manual test in sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.github/integ-config/integ-all.yml
+++ b/.github/integ-config/integ-all.yml
@@ -730,6 +730,15 @@ tests:
     spec: multi-part-copy
     browser: *minimal_browser_list
 
+  # GEN2 STORAGE
+  - test_name: integ_react_storage
+    desc: 'React Storage Gen2'
+    framework: react
+    category: storage
+    sample_name: [storage-gen2]
+    spec: storage-gen2
+    browser: *minimal_browser_list
+
   # INAPPMESSAGING
   - test_name: integ_in_app_messaging
     desc: 'React InApp Messaging'
--- a/packages/storage/__tests__/providers/s3/utils/client/S3/cases/getObject.ts
+++ b/packages/storage/__tests__/providers/s3/utils/client/S3/cases/getObject.ts
@@ -182,7 +182,7 @@ const getObjectCustomEndpoint: ApiFunctionalTestCase<typeof getObject> = [
 	} as Parameters<typeof getObject>[0],
 	{
 		Bucket: 'bucket',
-		Key: '/key',
+		Key: 'key',
 	},
 	expect.objectContaining({
 		url: expect.objectContaining({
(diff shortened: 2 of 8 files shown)


packages/storage/__tests__/providers/s3/utils/client/S3/cases/getObject.ts
```diff
@@ -182,7 +182,7 @@ const getObjectCustomEndpoint: ApiFunctionalTestCase<typeof getObject> = [
 	} as Parameters<typeof getObject>[0],
 	{
 		Bucket: 'bucket',
-		Key: 'key',
+		Key: '/key',
```
ashika112, line 185: Need to validate this change

kvramyasri7: Nice !!

.github/integ-config/integ-all.yml
```diff
@@ -730,6 +730,15 @@ tests:
     spec: multi-part-copy
     browser: *minimal_browser_list
 
+  # GEN2 STORAGE
```
ashika112, line 733: Also, adding this section which was missed with integ test PR.

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
```diff
@@ -27,7 +27,7 @@ export const validateStorageOperationInput = (
 
 		return {
 			inputType: STORAGE_INPUT_PATH,
-			objectKey,
+			objectKey: objectKey.slice(1),
```
ashika112, line 30: I know this looks weird. Bare with me for a bit. Once we finalize on a approach with backend will decide on our. For time being, all libraries are gonna have validation and remove the leading slash while passing for request

erinleigh90: Should we add a TODO or something easily searchable so it doesn't fall through the cracks?

Samaritan1011001: Maybe a todo will help keep track to change it later?

ashika112: I am actively on it right now. So i am ok with it. Best/Worst case this will be the solution 😂

Samaritan1011001: For now we assert them to put in a leading slash and we remove it, but then add it back in the serializer? 
Maybe todos in all three places will keep us reminded.

ashika112: Just slice here. We dont have a good path forward and hence i dont want to put TODO and leave it in there.

erinleigh90: Makes sense.

erinleigh90 (approved):

Samaritan1011001 (approved):

ashwinkumar6 (approved):

PR #13150, opened 2024-03-19
Feat: Gen2config parse

elorzafe:
#### Description of changes

Adds support for Gen2 config.
Categories supported on this draft PR:

- [x] Auth
- [x] Storage
- [x] Analytics
- [x] Geo
- [x] Data
- [x] Notifications
- [ ] custom

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/aws-amplify/__tests__/initSingleton.test.ts
+++ b/packages/aws-amplify/__tests__/initSingleton.test.ts
@@ -72,36 +72,36 @@ describe('initSingleton (DefaultAmplify)', () => {
 	describe('Gen2 Amplify configure', () => {
 		it('should use Gen2 config type', () => {
 			const gen2Config: Gen2Config = {
-				"$id": "https://amplify.aws/2024-02/outputs-schema.json",
-				"storage": {
-					"aws_region": "us-east-1",
-					"name": "my-bucket-name"
+				'version': '1',
+				'storage': {
+					'aws_region': 'us-east-1',
+					'bucket_name': 'my-bucket-name'
 				},
-				"auth": {
-					"user_pool_id": 'us-east-1:',
-					"user_pool_client_id": "xxxx",
-					"aws_region": "us-east-1",
-					"identity_pool_id": "test"
+				'auth': {
+					'user_pool_id': 'us-east-1:',
+					'user_pool_client_id': 'xxxx',
+					'aws_region': 'us-east-1',
+					'identity_pool_id': 'test'
 				},
-				"analytics": {
+				'analytics': {
 					amazon_pinpoint: {
 						app_id: 'xxxxx',
 						aws_region: 'us-east-1'
 					}
 				},
-				"geo": {
+				'geo': {
 					aws_region: 'us-east-1',
 					maps: {
-						items: [{ name: 'map1', style: 'color' }],
+						items: { 'map1': { name: 'map1', style: 'color' } },
 						default: 'map1'
 					},
 					geofence_collections: {
-						items: ["a", "b", "c"],
-						default: "a"
+						items: ['a', 'b', 'c'],
+						default: 'a'
 					},
 					search_indices: {
-						items: ["a", "b", "c"],
-						default: "a"
+						items: ['a', 'b', 'c'],
+						default: 'a'
 					}
 				}
 			};
@@ -111,8 +111,8 @@ describe('initSingleton (DefaultAmplify)', () => {
 			expect(AmplifySingleton.configure).toHaveBeenCalledWith({
 				Storage: {
 					S3: {
-						bucket: "my-bucket-name",
-						region: "us-east-1"
+						bucket: 'my-bucket-name',
+						region: 'us-east-1'
 					},
 				},
 				Auth: {
@@ -124,33 +124,36 @@ describe('initSingleton (DefaultAmplify)', () => {
 				},
 				Analytics: {
 					Pinpoint: {
-						appId: "xxxxx",
-						region: "us-east-1",
+						appId: 'xxxxx',
+						region: 'us-east-1',
 					},
 				},
 				Geo: {
 					LocationService: {
-						"geofenceCollections": {
-							"default": "a",
-							"items": [
-								"a",
-								"b",
-								"c",
+						'geofenceCollections': {
+							'default': 'a',
+							'items': [
+								'a',
+								'b',
+								'c',
 							],
 						},
-						"maps": {
-							"default": "map1",
-							"items": {
-								"map1": "color",
+						'maps': {
+							'default': 'map1',
+							'items': {
+								'map1': {
+									'name': 'map1',
+									'style': 'color'
+								},
 							},
 						},
-						"region": "us-east-1",
-						"searchIndices": {
-							"default": "a",
-							"items": [
-								"a",
-								"b",
-								"c",
+						'region': 'us-east-1',
+						'searchIndices': {
+							'default': 'a',
+							'items': [
+								'a',
+								'b',
+								'c',
 							],
 						},
 					},
--- a/packages/aws-amplify/__tests__/initSingleton.test.ts
+++ b/packages/aws-amplify/__tests__/initSingleton.test.ts
@@ -13,7 +13,7 @@ import {
 } from '../src/auth/cognito';
 
 import { Amplify } from '../src';
-import { Gen2Config } from '@aws-amplify/core/internals/utils';
+import { AmplifyOutputs } from '@aws-amplify/core/internals/utils';
 
 jest.mock('@aws-amplify/core');
 jest.mock('../src/auth/cognito', () => ({
@@ -71,7 +71,7 @@ describe('initSingleton (DefaultAmplify)', () => {
 
 	describe('Gen2 Amplify configure', () => {
 		it('should use Gen2 config type', () => {
-			const gen2Config: Gen2Config = {
+			const gen2Config: AmplifyOutputs = {
 				'version': '1',
 				'storage': {
 					'aws_region': 'us-east-1',
--- a/packages/aws-amplify/__tests__/initSingleton.test.ts
+++ b/packages/aws-amplify/__tests__/initSingleton.test.ts
@@ -69,9 +69,9 @@ describe('initSingleton (DefaultAmplify)', () => {
 		mockAmplifySingletonGetConfig.mockReset();
 	});
 
-	describe('Gen2 Amplify configure', () => {
-		it('should use Gen2 config type', () => {
-			const gen2Config: AmplifyOutputs = {
+	describe('Amplify configure with AmplifyOutputs format', () => {
+		it('should use AmplifyOutputs config type', () => {
+			const amplifyOutputs: AmplifyOutputs = {
 				'version': '1',
 				'storage': {
 					'aws_region': 'us-east-1',
@@ -106,7 +106,7 @@ describe('initSingleton (DefaultAmplify)', () => {
 				}
 			};
 
-			Amplify.configure(gen2Config);
+			Amplify.configure(amplifyOutputs);
 
 			expect(AmplifySingleton.configure).toHaveBeenCalledWith({
 				Storage: {
--- a/packages/aws-amplify/__tests__/initSingleton.test.ts
+++ b/packages/aws-amplify/__tests__/initSingleton.test.ts
@@ -13,6 +13,7 @@ import {
 } from '../src/auth/cognito';
 
 import { Amplify } from '../src';
+import { Gen2Config } from '@aws-amplify/core/internals/utils';
 
 jest.mock('@aws-amplify/core');
 jest.mock('../src/auth/cognito', () => ({
@@ -66,6 +67,96 @@ describe('initSingleton (DefaultAmplify)', () => {
 		mockCognitoUserPoolsTokenProviderSetKeyValueStorage.mockReset();
 		mockAmplifySingletonConfigure.mockReset();
 		mockAmplifySingletonGetConfig.mockReset();
+	});
+
+	describe('Gen2 Amplify configure', () => {
+		it('should use Gen2 config type', () => {
+			const gen2Config: Gen2Config = {
+				"$id": "https://amplify.aws/2024-02/outputs-schema.json",
+				"storage": {
+					"aws_region": "us-east-1",
+					"name": "my-bucket-name"
+				},
+				"auth": {
+					"user_pool_id": 'us-east-1:',
+					"user_pool_client_id": "xxxx",
+					"aws_region": "us-east-1",
+					"identity_pool_id": "test"
+				},
+				"analytics": {
+					amazon_pinpoint: {
+						app_id: 'xxxxx',
+						aws_region: 'us-east-1'
+					}
+				},
+				"geo": {
+					aws_region: 'us-east-1',
+					maps: {
+						items: [{ name: 'map1', style: 'color' }],
+						default: 'map1'
+					},
+					geofence_collections: {
+						items: ["a", "b", "c"],
+						default: "a"
+					},
+					search_indices: {
+						items: ["a", "b", "c"],
+						default: "a"
+					}
+				}
+			};
+
+			Amplify.configure(gen2Config);
+
+			expect(AmplifySingleton.configure).toHaveBeenCalledWith({
+				Storage: {
+					S3: {
+						bucket: "my-bucket-name",
+						region: "us-east-1"
+					},
+				},
+				Auth: {
+					Cognito: {
+						identityPoolId: 'test',
+						userPoolId: 'us-east-1:',
+						userPoolClientId: 'xxxx'
+					}
+				},
+				Analytics: {
+					Pinpoint: {
+						appId: "xxxxx",
+						region: "us-east-1",
+					},
+				},
+				Geo: {
+					LocationService: {
+						"geofenceCollections": {
+							"default": "a",
+							"items": [
+								"a",
+								"b",
+								"c",
+							],
+						},
+						"maps": {
+							"default": "map1",
+							"items": {
+								"map1": "color",
+							},
+						},
+						"region": "us-east-1",
+						"searchIndices": {
+							"default": "a",
+							"items": [
+								"a",
+								"b",
+								"c",
+							],
+						},
+					},
+				}
+			}, expect.anything());
+		})
 	});
 
 	describe('DefaultAmplify.configure()', () => {
--- a/packages/aws-amplify/src/initSingleton.ts
+++ b/packages/aws-amplify/src/initSingleton.ts
@@ -8,8 +8,11 @@ import {
 	defaultStorage,
 } from '@aws-amplify/core';
 import {
+	Gen2Config,
 	LegacyConfig,
+	isGen2Config,
 	parseAWSExports,
+	parseGen2Config,
 } from '@aws-amplify/core/internals/utils';
 
 import {
@@ -31,13 +34,15 @@ export const DefaultAmplify = {
 	 * Amplify.configure(config);
 	 */
 	configure(
-		resourceConfig: ResourcesConfig | LegacyConfig,
+		resourceConfig: ResourcesConfig | LegacyConfig | Gen2Config,
 		libraryOptions?: LibraryOptions,
 	): void {
 		let resolvedResourceConfig: ResourcesConfig;
 
 		if (Object.keys(resourceConfig).some(key => key.startsWith('aws_'))) {
 			resolvedResourceConfig = parseAWSExports(resourceConfig);
+		} else if (isGen2Config(resourceConfig)) {
+			resolvedResourceConfig = parseGen2Config(resourceConfig);
 		} else {
 			resolvedResourceConfig = resourceConfig as ResourcesConfig;
 		}
--- a/packages/aws-amplify/src/initSingleton.ts
+++ b/packages/aws-amplify/src/initSingleton.ts
@@ -8,11 +8,11 @@ import {
 	defaultStorage,
 } from '@aws-amplify/core';
 import {
-	Gen2Config,
+	AmplifyOutputs,
 	LegacyConfig,
-	isGen2Config,
+	isAmplifyOutputs,
 	parseAWSExports,
-	parseGen2Config,
+	parseAmplifyOutputs,
 } from '@aws-amplify/core/internals/utils';
 
 import {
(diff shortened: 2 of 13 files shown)


packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,245 @@
...
+					KinesisFirehose,
+				};
+			} else {
+				config.Analytics!.KinesisFirehose = KinesisFirehose;
+			}
+		}
+	}
+
+	if (resourceConfig.geo) {
+		const geoProperties = resourceConfig.geo;
```
cshfang, line 176: Nit: Looks like we can save some duplication and improve readability a bit if we destructure. I think this is broadly applicable to all of the other category configs above, especially the analytics ones where we have to drill multiple levels into provider configs.
```suggestion
		const { aws_region, geofence_collections, maps, search_indices } = resourceConfig.geo;
```

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,245 @@
...
+import { Gen2AuthProperties } from './singleton/gen2/types';
+import { Gen2Config, LegacyConfig, ResourcesConfig } from './singleton/types';
+
+export function isGen2Config(
+	resourceConfig: ResourcesConfig | LegacyConfig | Gen2Config,
+): resourceConfig is Gen2Config {
+	return !!(resourceConfig as Gen2Config).$id;
+}
+
+export function parseGen2Config(resourceConfig: Gen2Config): ResourcesConfig {
```
cshfang, line 11: I found it really confusing to name this param `resourceConfig` since it's so close to the `ResourcesConfig` type we're trying to map this **to**. I think we should be able to just keep it concise like we do with the parseAWSExports and refer to the incoming config as just `config`.
```suggestion
export function parseGen2Config(config: Gen2Config): ResourcesConfig {
```

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,245 @@
+import { OAuthProvider } from './libraryUtils';
+import { Gen2AuthProperties } from './singleton/gen2/types';
+import { Gen2Config, LegacyConfig, ResourcesConfig } from './singleton/types';
+
+export function isGen2Config(
+	resourceConfig: ResourcesConfig | LegacyConfig | Gen2Config,
```
cshfang, line 6: Similar comment to below but it didn't strike me as a big deal until the function body grew and the parameter was oft used. It does make even more sense here since the incoming parameter can be any type of config.
```suggestion
	config: ResourcesConfig | LegacyConfig | Gen2Config,
```

packages/core/src/singleton/types.ts
```diff
@@ -36,6 +42,14 @@ export interface LegacyConfig {
 	aws_project_region?: string;
 }
 
+export interface Gen2Config {
+	$id?: 'https://amplify.aws/2024-02/outputs-schema.json';
```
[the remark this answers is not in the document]
elorzafe: I updated that to `version`, `$id` is not used

haverchuck:
This PR doesn't include Predictions or Interactions?

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+				requireSpecialCharacters: password_policy.require_symbols,
+				minLength: password_policy.min_length ?? 6,
+			};
+		}
+
+		if (mfa_configuration) {
+			config.Auth.Cognito.mfa = {
+				status: getMfaStatus(mfa_configuration),
+				smsEnabled: mfa_methods?.some(method => method === 'SMS'),
+				totpEnabled: mfa_methods?.some(method => method === 'TOTP'),
```
jimblanc, line 78: I think we can get by without the functions here, can't we just do `includes('SMS')`, etc?

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+				redirectSignIn: oauth.redirect_sign_in_uri,
+				redirectSignOut: oauth.redirect_sign_out_uri,
+				responseType: oauth.response_type,
+				scopes: oauth.scopes,
+				providers: getOAuthProviders(oauth.identity_providers),
+			};
+		}
+
+		if (username_attributes) {
+			if (username_attributes.some(user => user === 'EMAIL')) {
```
jimblanc, line 100: Ditto for these next usages of `some()`

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+		if (standard_required_attributes) {
+			config.Auth.Cognito.userAttributes = standard_required_attributes.reduce(
+				(acc, curr) => ({ ...acc, [curr]: { required: true } }),
+				{},
+			);
+		}
+
+		if (!config.Auth.Cognito.loginWith) {
+			delete config.Auth.Cognito.loginWith;
+		}
```
jimblanc, line 122: Can we get rid of this if we gate the the `loginWith` population behind the `username_attributes` & `oauth` checks? Seems a little weird

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+				appId: amazon_pinpoint.app_id,
+				region: amazon_pinpoint.aws_region,
+			};
+			if (!config.Analytics) {
+				config.Analytics = {
+					Pinpoint,
+				};
+			} else {
+				config.Analytics!.Pinpoint = Pinpoint;
+			}
```
jimblanc, line 139: Why is Analytics different than the others? Can't we just do `Analytics = { Pinpoint: { ... }}`

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+		case 'API_KEY':
+			return 'apiKey';
+		case 'AWS_IAM':
+			return 'iam';
+		case 'AWS_LAMBDA':
+			return 'lambda';
+		case 'OPENID_CONNECT':
+			return 'oidc';
+		default:
+			throw new Error('Invalid AuthMode configured');
```
jimblanc, line 234: Sanity check: Should this use `AmplifyError` or gets wrapped somewhere else up the stack?

packages/core/src/parseAmplifyGen2.ts
cshfang, line 78: I think we can just use `includes` to simplify this
```suggestion
				smsEnabled: mfa_methods?.includes('SMS'),
				totpEnabled: mfa_methods?.includes('TOTP'),
```

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+				config.Auth.Cognito.loginWith.email = true;
+			}
+
+			if (username_attributes.some(user => user === 'PHONE_NUMBER')) {
+				config.Auth.Cognito.loginWith.phone = true;
+			}
+
+			if (username_attributes.some(user => user === 'USERNAME')) {
+				config.Auth.Cognito.loginWith.username = true;
+			}
```
cshfang, line 110: Nit: Ditto
```suggestion
			if (username_attributes.includes('EMAIL')) {
				config.Auth.Cognito.loginWith.email = true;
			}

			if (username_attributes.includes('PHONE_NUMBER')) {
				config.Auth.Cognito.loginWith.phone = true;
			}

			if (username_attributes.includes('USERNAME')) {
				config.Auth.Cognito.loginWith.username = true;
			}
```

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+				},
+			};
+		} else {
+			config.Auth = {
+				Cognito: {
+					userPoolId: user_pool_id,
+					userPoolClientId: user_pool_client_id,
+				},
+			};
+		}
```
cshfang, line 62: Seems like you can reduce some duplication since both conditions result in a `Cognito` object with `userPoolId` and `userPoolClientId`
```suggestion
		config.Auth = {
			Cognito: {
				userPoolId: user_pool_id,
				userPoolClientId: user_pool_client_id,
			},
		};
		if (identity_pool_id) {
			config.Auth.Cognito.identityPoolId = identity_pool_id,
		}
```

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+	config: ResourcesConfig | LegacyConfig | Gen2Config,
+): config is Gen2Config {
+	return (config as Gen2Config).version === '1';
+}
+
+export function parseGen2Config(gen2Config: Gen2Config): ResourcesConfig {
+	const config: ResourcesConfig = {};
+
+	if (gen2Config?.storage) {
+		/* eslint-disable camelcase */
```
cshfang, line 23: This should either be moved to the top of the file or converted to `eslint-disable-next-line` as necessary.

packages/core/src/libraryUtils.ts
```diff
@@ -21,7 +21,8 @@ export {
 	deDupeAsyncFunction,
 } from './utils';
 export { parseAWSExports } from './parseAWSExports';
-export { LegacyConfig } from './singleton/types';
+export { isGen2Config, parseGen2Config } from './parseAmplifyGen2';
+export { LegacyConfig, Gen2Config } from './singleton/types';
```
cshfang, line 25: I left a comment on the file as well but I think we should move this Gen2Config into the gen2 subfolder and just keep things isolated.

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+	AuthType,
+	Gen2AuthMFAConfiguration,
+	Gen2OAuthIdentityProviders,
+} from './singleton/gen2/types';
+import { Gen2Config, LegacyConfig, ResourcesConfig } from './singleton/types';
+
+export function isGen2Config(
+	config: ResourcesConfig | LegacyConfig | Gen2Config,
+): config is Gen2Config {
+	return (config as Gen2Config).version === '1';
```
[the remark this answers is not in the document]
elorzafe: I will update this logic a bit in the following commit

packages/core/src/parseAmplifyGen2.ts
elorzafe: Originally support more providers, but those are not included anymore, good catch!

packages/core/src/parseAmplifyGen2.ts
elorzafe: I will add an exhaustive check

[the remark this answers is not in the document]
elorzafe: I updated this with an exhaustive check

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+					region: aws_region,
+				},
+			};
+			if (!config.Notifications) {
+				config.Notifications = {
+					InAppMessaging,
+				};
+			} else {
+				config.Notifications.InAppMessaging = InAppMessaging;
+			}
```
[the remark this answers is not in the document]
elorzafe: good catch thanks!

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+		}
+
+		if (standard_required_attributes) {
+			config.Auth.Cognito.userAttributes = standard_required_attributes.reduce(
+				(acc, curr) => ({ ...acc, [curr]: { required: true } }),
+				{},
+			);
+		}
+
+		if (!config.Auth.Cognito.loginWith) {
```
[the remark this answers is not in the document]
elorzafe: updated thanks!

packages/core/src/parseAmplifyGen2.ts
elorzafe: There was an issue with the Union type but I reduced this as much as possible

Thanks

packages/core/src/singleton/gen2/types.ts
```diff
@@ -0,0 +1,103 @@
+import { ModelIntrospectionSchema } from '../API/types';
+import { AuthStandardAttributeKey } from '../Auth/types';
+
+export type Gen2OAuthIdentityProviders =
```
calebpollman, line 4: Suggest removing the `Gen2` prefix and dropping the pluralization here (and for the rest of the string unions in this file). There is no `AuthIdentityProviders` type in the JS lib currently, and plural form is confusing in usage since it is refers to a singular value:

```ts
// but it's just one!
const authIdentityProviders: AuthIdentityProviders = 'GOOGLE`;

// prevents applying a consistent naming convention when assigning an array
const authIdentityProviders: AuthIdentityProviders[] = ['GOOGLE`, 'AMAZON'];

// how to name this?
type AuthIdentityProviders = AuthIdentityProviders[];
```

packages/core/src/singleton/gen2/types.ts
```diff
@@ -0,0 +1,103 @@
...
+	| 'GOOGLE'
+	| 'FACEBOOK'
+	| 'LOGIN_WITH_AMAZON'
+	| 'SIGN_IN_WITH_APPLE';
+
+type Gen2AuthUsernameAttributes = 'EMAIL' | 'PHONE_NUMBER' | 'USERNAME';
+
+type Gen2AuthUserVerificationTypes = 'EMAIL' | 'PHONE_NUMBER';
```
calebpollman, line 12: Prefer naming the type to its use case, e.g. `UserVerificationMethod`

packages/aws-amplify/src/initSingleton.ts
```diff
@@ -31,13 +34,15 @@ export const DefaultAmplify = {
 	 * Amplify.configure(config);
 	 */
 	configure(
-		resourceConfig: ResourcesConfig | LegacyConfig,
+		resourceConfig: ResourcesConfig | LegacyConfig | Gen2Config,
 		libraryOptions?: LibraryOptions,
 	): void {
 		let resolvedResourceConfig: ResourcesConfig;
 
 		if (Object.keys(resourceConfig).some(key => key.startsWith('aws_'))) {
```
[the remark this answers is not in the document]
calebpollman: +1

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+			const Pinpoint = {
+				appId: amazon_pinpoint.app_id,
+				region: amazon_pinpoint.aws_region,
+			};
+			if (!config.Analytics) {
+				config.Analytics = {
+					Pinpoint,
+				};
+			} else {
+				config.Analytics!.Pinpoint = Pinpoint;
```
calebpollman: Why is there a need for a type assertion here?

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+		if (provider === 'FACEBOOK') {
+			oauthProviders.push('Facebook');
+		}
+		if (provider === 'SIGN_IN_WITH_APPLE') {
+			oauthProviders.push('Apple');
+		}
+	});
+
+	return oauthProviders;
+}
```
calebpollman, line 258: Why not use `Array.map` and simplify the lookup logic by assign an object with the IdentityProvider values mapped to their ResourcesConfig name?
```suggestion
const providerNames: Record<Gen2OAuthIdentityProviders, OAuthProvider> = {
	GOOGLE: 'Google',
	LOGIN_WITH_AMAZON: 'Amazon',
	FACEBOOK: 'Facebook',
	SIGN_IN_WITH_APPLE: 'Apple',
};
function getOAuthProviders(
	providers?: Gen2OAuthIdentityProviders[],
): OAuthProvider[] {
	return (providers ?? []).map(provider => providerNames[provider]);
}
```

calebpollman (commented):
Thanks for doing this, left mostly call outs around type names and testability. 

Also don't think we need to use "Gen2" in the types, utilities and file names here to avoid confusion and coupling in the future, in particular if we end up needing to publicly vend the new types or `parseGen2Config` in the future. Not the greatest suggestion, but something like `parseOutputsConfig` would make more sense imo since it would align to the new configuration file name

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -6,21 +12,22 @@ import {
...
 } from './singleton/gen2/types';
-import { Gen2Config, LegacyConfig, ResourcesConfig } from './singleton/types';
+import { LegacyConfig, ResourcesConfig } from './singleton/types';
 
 export function isGen2Config(
 	config: ResourcesConfig | LegacyConfig | Gen2Config,
 ): config is Gen2Config {
-	return (config as Gen2Config).version === '1';
+	// version format initially will be '1' but is expected to be something like x.y where x is major and y minor version
+	return ('' + (config as Gen2Config).version).startsWith('1');
```
jimblanc, line 24: Why the empty string?

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -231,7 +229,9 @@ function getGraphQLAuthMode(authType: AuthType): GraphQLAuthMode {
 		case 'OPENID_CONNECT':
 			return 'oidc';
 		default:
-			throw new Error('Invalid AuthMode configured');
+			// This makes sure all AuthTypes are handled.
+			const exhaustiveCheck: never = authType;
+			throw new Error(`Unhandled GraphQL Auth Mode: ${exhaustiveCheck}`);
```
jimblanc, line 234: Nit

```suggestion
			throw new Error(`Unsupported GraphQL Auth Mode: ${exhaustiveCheck}`);
```

packages/core/src/parseAmplifyGen2.ts
jimblanc, line 1: Missing license header in this file (failing test)

packages/core/src/singleton/gen2/types.ts
jimblanc, line 1: Missing license header in this file (failing test)

jimblanc (commented):
Looking good! Few nits & callouts

packages/core/src/parseAmplifyGen2.ts
elorzafe: in case `version` is `undefined`

packages/core/src/singleton/gen2/types.ts
elorzafe: Good catch, I will update that!

packages/aws-amplify/__tests__/initSingleton.test.ts
```diff
@@ -68,6 +69,96 @@ describe('initSingleton (DefaultAmplify)', () => {
...
+							"default": "a",
+							"items": [
+								"a",
+								"b",
+								"c",
+							],
+						},
+					},
+				}
+			}, expect.anything());
```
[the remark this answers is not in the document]
elorzafe: I am just testing the resource parameters, there are other tests that validate that

packages/core/src/parseAmplifyGen2.ts
```diff
@@ -0,0 +1,267 @@
...
+export function isGen2Config(
+	config: ResourcesConfig | LegacyConfig | Gen2Config,
+): config is Gen2Config {
+	return (config as Gen2Config).version === '1';
+}
+
+export function parseGen2Config(gen2Config: Gen2Config): ResourcesConfig {
+	const config: ResourcesConfig = {};
+
+	if (gen2Config?.storage) {
```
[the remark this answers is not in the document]
elorzafe: Good idea, thanks

packages/core/src/parseAmplifyGen2.ts
elorzafe: Nice suggestion, I would do this!

packages/core/src/parseAmplifyOutputs.ts
```diff
@@ -0,0 +1,363 @@
...
+	GOOGLE: 'Google',
+	LOGIN_WITH_AMAZON: 'Amazon',
+	FACEBOOK: 'Facebook',
+	SIGN_IN_WITH_APPLE: 'Apple',
+};
+
+function getOAuthProviders(
+	providers?: AmplifyOutputsOAuthIdentityProvider[],
+): OAuthProvider[] {
+	return (providers ?? []).map(provider => providerNames[provider]);
```
cshfang, line 353: Nit: Just use a default value instead
```suggestion
	providers: AmplifyOutputsOAuthIdentityProvider[] = [],
): OAuthProvider[] {
	return providers.map(provider => providerNames[provider]);
}
```

cshfang (commented):
Left a few more comments but it's overall looking pretty close. Can you please also just do a search for "gen2" across these changes to make sure they're updated? (Stray variable names, test description strings, etc)

jimblanc (commented):
Looking good! Agree with Chris' suggestions for cleanup in `parseAmplifyOutputs.ts`

packages/core/src/parseAmplifyOutputs.ts
```diff
@@ -0,0 +1,363 @@
...
+		);
+	}
+
+	return authConfig;
+}
+
+export function parseAnalytics(
+	amplifyOutputsAnalyticsProperties?: AmplifyOutputsAnalyticsProperties,
+): AnalyticsConfig | undefined {
+	if (!amplifyOutputsAnalyticsProperties?.amazon_pinpoint) {
```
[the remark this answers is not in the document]
elorzafe: yes, that was the intention, because is the only one that has different providers

jimblanc (dismissed):

packages/core/src/singleton/API/types.ts
```diff
@@ -47,7 +47,7 @@ interface APIGraphQLConfig {
 	modelIntrospection?: ModelIntrospectionSchema;
```
ashika112, line 47: i thought this was decided not to exposed through schema?

ashika112 (approved):

packages/core/src/parseAmplifyOutputs.ts
```diff
@@ -0,0 +1,347 @@
...
+	if (!amplifyOutputsDataProperties) {
+		return undefined;
+	}
+
+	const {
+		aws_region,
+		default_authorization_type,
+		url,
+		api_key,
+		model_introspection,
```
ashika112, line 220: this bit got changed in schema now right

jimblanc (approved):

packages/core/src/parseAmplifyOutputs.ts
```diff
@@ -0,0 +1,347 @@
...
+	} = amplifyOutputsDataProperties;
```
david-mcafee, line 221: Missing both `customEndpoint` and `customEndpointRegion`

packages/core/src/parseAmplifyOutputs.ts
```diff
@@ -0,0 +1,347 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
david-mcafee, line 1: This is missing [REST API config](https://github.com/aws-amplify/amplify-js/blob/main/packages/core/src/singleton/API/types.ts#L50-L68)

PR #13154, opened 2024-03-21
feat(api-graphql): custom subscriptions

iartemiev:
~Draft until this is merged: https://github.com/aws-amplify/amplify-api-next/pull/131~

TODO: 
- [x] Update `data-schema` & `data-schema-types` dep versions after merging ^

#### Description of changes

Adds custom subscriptions support to data-client.

Given this schema
```ts
const schema = a.schema({
...,
onPostLiked: a
  .subscription()
  .for(a.ref('likePostReturnPost'))
  .returns(a.ref('Post'))
  .handler(a.handler.custom({ entry: './jsResolver_base.js' })),

onPostUpdated: a
  .subscription()
  .for(a.ref('Post').mutations(['update']))
  .arguments({ postId: a.string() })
  .returns(a.ref('Post'))
  .handler(a.handler.custom({ entry: './jsResolver_base.js' })),

}).authorization([a.allow.private()]),
```

data-client makes the following subscription functions available:

```ts
const sub = client.subscriptions.onPostLiked().subscribe({ next: ..., error: ... })
const sub2 = client.subscriptions.onPostUpdated({ postId: 'abc123' }).subscribe({ next: ..., error: ... })
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,6 @@
 {
 	"editor.detectIndentation": false,
 	"editor.insertSpaces": false,
-	"editor.tabSize": 4,
 	"prettier.requireConfig": true,
 	"typescript.tsdk": "node_modules/typescript/lib",
 	"formattingToggle.affects": [
--- a/packages/api-graphql/__tests__/fixtures/modeled/schema.ts
+++ b/packages/api-graphql/__tests__/fixtures/modeled/schema.ts
@@ -120,7 +120,7 @@ const schema = a.schema({
 			argumentContent: a.string().required(),
 		})
 		.returns(a.ref('EchoResult'))
-		.function('echoFunction')
+		.handler(a.handler.function('echoFunction'))
 		.authorization([a.allow.public()]),
 
 	// custom query returning a primitive type
@@ -130,7 +130,7 @@ const schema = a.schema({
 			inputString: a.string().required(),
 		})
 		.returns(a.string())
-		.function('echoFunction')
+		.handler(a.handler.function('echoFunction'))
 		.authorization([a.allow.public()]),
 	echoNestedCustomTypes: a
 		.query()
@@ -138,7 +138,7 @@ const schema = a.schema({
 			input: a.string().required(),
 		})
 		.returns(a.ref('ProductTrackingMeta'))
-		.function('echoFunction')
+		.handler(a.handler.function('echoFunction'))
 		.authorization([a.allow.public()]),
 	echoModelHasNestedTypes: a
 		.query()
@@ -146,7 +146,7 @@ const schema = a.schema({
 			input: a.string().required(),
 		})
 		.returns(a.ref('Product'))
-		.function('echoFunction')
+		.handler(a.handler.function('echoFunction'))
 		.authorization([a.allow.public()]),
 	// custom mutation returning a non-model type
 	PostLikeResult: a.customType({
@@ -158,7 +158,7 @@ const schema = a.schema({
 			postId: a.id().required(),
 		})
 		.returns(a.ref('PostLikeResult'))
-		.function('echoFunction')
+		.handler(a.handler.function('echoFunction'))
 		.authorization([a.allow.private()]),
 
 	// custom mutation returning a model type
@@ -182,9 +182,21 @@ const schema = a.schema({
 			postId: a.id().required(),
 		})
 		.returns(a.ref('Post'))
-		.function('echoFunction')
+		.handler(a.handler.function('echoFunction'))
 		.authorization([a.allow.private()]),
 
+	onPostLiked: a
+		.subscription()
+		.for(a.ref('likePostReturnPost'))
+		.returns(a.ref('Post'))
+		.handler(a.handler.custom({ entry: './jsResolver_base.js' })),
+
+	onPostUpdated: a
+		.subscription()
+		.for(a.ref('Post').mutations(['update']))
+		.arguments({ postId: a.string() })
+		.returns(a.ref('Post'))
+		.handler(a.handler.custom({ entry: './jsResolver_base.js' })),
 	//#endregion
 });
 
--- a/packages/api-graphql/__tests__/fixtures/modeled/schema.ts
+++ b/packages/api-graphql/__tests__/fixtures/modeled/schema.ts
@@ -83,9 +83,9 @@ const schema = a.schema({
 			viewCount: a.integer(),
 			status: a.enum(['draft', 'pending', 'published']),
 		})
-		.secondaryIndexes([
-			a.index('title'),
-			a.index('description').sortKeys(['viewCount']),
+		.secondaryIndexes(index => [
+			index('title'),
+			index('description').sortKeys(['viewCount']),
 		]),
 	Product: a
 		.model({
--- a/packages/api-graphql/__tests__/internals/generateClient.test.ts
+++ b/packages/api-graphql/__tests__/internals/generateClient.test.ts
@@ -10,7 +10,6 @@ import {
 	expectSubWithlibraryConfigHeaders,
 } from '../utils/expects';
 import { Observable, from } from 'rxjs';
-import * as internals from '../../src/internals';
 
 const serverManagedFields = {
 	id: 'some-id',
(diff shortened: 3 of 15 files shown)


.vscode/settings.json
```diff
@@ -1,7 +1,6 @@
 {
 	"editor.detectIndentation": false,
 	"editor.insertSpaces": false,
-	"editor.tabSize": 4,
```
iartemiev, line 4: The workspace settings defined in this file override local VS Code settings, so this effectively forces everyone to use `tabSize: 4`. Removing it so contributors can use their tab size of choice.

packages/api-graphql/src/internals/operations/custom.ts
```diff
@@ -85,38 +99,61 @@ import {
...
+			}
+		}
+
+		if (argsDefined) {
+			if (useContext) {
+				arg = args[1] as QueryArgs;
+			} else {
+				arg = args[0] as QueryArgs;
+			}
+		}
```
svidgen, line 132: I can't brain it right now, but it would be slick if there were a way to ingest these variations with some type guarding.

svidgen (dismissed):
No blockers jumping out at me.

packages/api-graphql/__tests__/fixtures/modeled/schema.ts
```diff
@@ -83,9 +83,9 @@ const schema = a.schema({
 			viewCount: a.integer(),
 			status: a.enum(['draft', 'pending', 'published']),
 		})
-		.secondaryIndexes([
-			a.index('title'),
-			a.index('description').sortKeys(['viewCount']),
+		.secondaryIndexes(index => [
+			index('title'),
+			index('description').sortKeys(['viewCount']),
```
iartemiev, line 88: Lines us the DX with https://github.com/aws-amplify/amplify-api-next/pull/132

packages/api-graphql/__tests__/fixtures/modeled/schema.ts
```diff
@@ -120,7 +120,7 @@ const schema = a.schema({
 			argumentContent: a.string().required(),
 		})
 		.returns(a.ref('EchoResult'))
-		.function('echoFunction')
+		.handler(a.handler.function('echoFunction'))
```
iartemiev, line 123: Migrating these to the new DX as well. We still have tests in `data-schema` that ensure the now-deprecated top-level `.function` functionality is intact.

packages/api-graphql/__tests__/internals/generateClient.test.ts
```diff
@@ -5392,6 +5395,121 @@ describe('generateClient', () => {
...
+					done();
+				},
+				error(error) {
+					expect(error).toBeUndefined();
+					done('bad news!');
+				},
+			});
+
+			sub.unsubscribe();
+		});
```
HuiSF, line 5451: nit: since the expect assertions are happening inside the subscribe async, can we add `expect.assertions();` to ensure these assertions have actually happened?

packages/api-graphql/__tests__/internals/generateClient.test.ts
```diff
@@ -5392,6 +5395,121 @@ describe('generateClient', () => {
...
+						done();
+					},
+					error(error) {
+						expect(error).toBeUndefined();
+						done('bad news!');
+					},
+				});
+
+			sub.unsubscribe();
+		});
```
HuiSF, line 5511: nit: ditto

packages/api-graphql/__tests__/internals/generateClient.test.ts
iartemiev: The `done()` call ensures it. Test does not complete until that gets executed

packages/api-graphql/__tests__/internals/generateClient.test.ts
```diff
@@ -5181,6 +5182,8 @@ describe('generateClient', () => {
 		beforeEach(() => {
 			jest.clearAllMocks();
 			jest.resetAllMocks();
+			jest.restoreAllMocks();
```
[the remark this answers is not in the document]
iartemiev: The new `spy` for the `graphql` method I added ([here](https://github.com/aws-amplify/amplify-js/pull/13154/files#diff-eb4d95fe48d39dd28d6e4254dc39b982f3caf4f45298629638581d98eb86d50bR5416)) to ensure the generated GQL doc looks correct.  Other subscriptions tests were broken without adding this

packages/api-graphql/src/internals/operations/custom.ts
iartemiev: Agreed. `contextSpec` and `options` have a predictable shape, so we can guard those.

iartemiev: (will update shortly)

packages/api-graphql/__tests__/internals/generateClient.test.ts
svidgen: I think I also favor either just returning a promise or using `done` over expecting a number of assertions. I want my test to clearly indicate in "logic" that it's done one way or the other.

But, maybe I'm a little ignorant here. If I have `expect.assertions(n)` and I add a new assertion, does `expect.assertions(n)` fail when `n + 1` assertions are made?

svidgen: (Or does it only ensure *at least* `n` assertions are made?)

packages/api-graphql/src/internals/generateCustomOperationsProperty.ts
```diff
@@ -10,16 +14,21 @@ import { V6Client, __amplify } from '../types';
...
 
 type CustomOpsProperty<
 	T extends Record<any, any>,
 	OpType extends OpTypes,
-> = OpType extends 'queries' ? CustomQueries<T> : CustomMutations<T>;
+> = OpType extends 'queries'
+	? CustomQueries<T>
+	: OpType extends 'mutations'
+		? CustomMutations<T>
+		: CustomSubscriptions<T>;
```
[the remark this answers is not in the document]
iartemiev: Doesn't hurt to make it exhaustive. I've got an update queued up once the other PR is merged.

packages/api-graphql/__tests__/internals/generateClient.test.ts
stocaaro: I think its looking for an exact match. I'm not sure if it leaves time to fail on n+1 if the test is complete.

packages/api-graphql/src/internals/generateCustomOperationsProperty.ts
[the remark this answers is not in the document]
iartemiev: Updated

packages/api-graphql/src/internals/operations/custom.ts
iartemiev: I added a type guard for `contextSpec`, but `options` and `args` are actually not deterministically distinguishable via type guard, because `options` consists of only optional params and we can't be sure that `args` won't contain an argument with the same name. Updated with a slight improvement, but not fully type guarded.

packages/api-graphql/__tests__/internals/generateClient.test.ts
HuiSF: Oops, missed the usage of `done`. If so, do we need `expect(error).toBeUndefined();` in the catch clause? It's a bit confusing reading this test to interpret what are expected.

iartemiev: @HuiSF - the `expect(error).toBeUndefined()` will give us a slightly more helpful error message as opposed to just taking out the `done()` from the catch and letting the test timeout. With the expect we'll know where to look right away.

packages/api-graphql/src/internals/operations/custom.ts
svidgen: Gotcha. Could a type guard be applied to `...args` as a whole with guarding based on length and the shape of the first arg?

(Not a blocker.)

HuiSF (approved):

packages/api-graphql/src/internals/operations/custom.ts
iartemiev: That's an interesting idea. I think it can work. Mind if I play around with that as a follow up?

stocaaro (approved):

PR #13158, opened 2024-03-21
fix(auth): add inflightPromise in token orchestrator

israx:
#### Description of changes

The main issue is due to a race condition between `attemptCompleteOAuthFlow` and `TokenOrchestrator.getTokens`

Ideally when there is an inflight OAuth flow, `attemptCompleteOAuthFlow` should always be called first to register any upcoming async calls coming from `TokenOrchestrator.getTokens`. This behavior is consistent in React but the affected frameworks are bundled different.

The fix at the moment is to check if `OAuth` is inflight directly from the TokenOrchestrator and add an inflight promise if so.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts
+++ b/packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts
@@ -23,6 +23,8 @@ import {
 	DeviceMetadata,
 	TokenRefresher,
 } from './types';
+import { oAuthStore } from '../utils/oauth';
+import { addInflightPromise } from '../utils/oauth/inflightPromise';
 
 export class TokenOrchestrator implements AuthTokenOrchestrator {
 	private authConfig?: AuthConfig;
@@ -30,6 +32,14 @@ export class TokenOrchestrator implements AuthTokenOrchestrator {
 	tokenRefresher?: TokenRefresher;
 	waitForInflightOAuth: () => Promise<void> = async () => {
 		// no-op
+		if (!(await oAuthStore.loadOAuthInFlight())) {
+			return;
+		}
+		
+		return new Promise<void>((resolve, _reject) => {
+			addInflightPromise(resolve);
+		});
+
 	};
 
 	setAuthConfig(authConfig: AuthConfig) {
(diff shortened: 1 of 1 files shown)


packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts
```diff
@@ -23,13 +23,23 @@ import {
...
 	tokenStore?: AuthTokenStore;
 	tokenRefresher?: TokenRefresher;
 	waitForInflightOAuth: () => Promise<void> = async () => {
 		// no-op
+		if (!(await oAuthStore.loadOAuthInFlight())) {
+			return;
+		}
+		
+		return new Promise<void>((resolve, _reject) => {
+			addInflightPromise(resolve);
```
HuiSF, line 40: Should we remove the `addInflightPromise()` call from `attempCompleteOAuthFlow` then?

packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts
```diff
@@ -23,13 +23,23 @@ import {
...
 } from './types';
+import { oAuthStore } from '../utils/oauth';
+import { addInflightPromise } from '../utils/oauth/inflightPromise';
 
 export class TokenOrchestrator implements AuthTokenOrchestrator {
 	private authConfig?: AuthConfig;
 	tokenStore?: AuthTokenStore;
 	tokenRefresher?: TokenRefresher;
 	waitForInflightOAuth: () => Promise<void> = async () => {
 		// no-op
```
jimblanc, line 34: Nit: Not a no-op anymore

jimblanc (commented):
If we're planning on hot-fixing this, should target `release` branch

packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts
israx: yeah good point

HuiSF: `cognitoUserPoolsTokenProvider.setWaitForInflightOAuth` should be removed as well as its single use case is being replace by this. 👍

PR #13162, opened 2024-03-22
fix(auth): add inflight promise in token orchestrator

israx:
#### Description of changes

The main issue is due to a race condition between attemptCompleteOAuthFlow and TokenOrchestrator.getTokens

Ideally when there is an inflight OAuth flow, attemptCompleteOAuthFlow should always be called first to register any upcoming async calls coming from TokenOrchestrator.getTokens. This behavior is consistent in React but the affected frameworks are bundled different.

The fix at the moment is to check if OAuth is inflight directly from the TokenOrchestrator and add an inflight promise if so.

#### verification
- smoke test on sign-in with oauth e2e test app
- smoke test on React, Vue and Angular

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/auth/__tests__/providers/cognito/tokenOrchestrator.test.ts
```diff
@@ -124,5 +141,18 @@ describe('TokenOrchestrator', () => {
...
+		it('Should call addInflightPromise when OAuth is inflight', async () => {
+			 mockAuthTokenStore.loadTokens.mockResolvedValue(validAuthTokens);
+			(oAuthStore.loadOAuthInFlight as jest.Mock).mockResolvedValue(true);
+         
+			await tokenOrchestrator.getTokens();
+
+			expect(mockAddInflightPromise).toHaveBeenCalledTimes(1);
+	
+			const inflightPromiseResolver = mockAddInflightPromise.mock.calls[0][0];
+			inflightPromiseResolver();
```
HuiSF, line 154: I think the test should assert `tokenOrchestrator.getTokens()` to resolve and get tokens after `inflightPromiseResolver()` resolves the blocking promise

packages/auth/__tests__/providers/cognito/utils/oauth/attemptCompleteOAuthFlow.test.ts
```diff
@@ -108,8 +99,8 @@ describe('attemptCompleteOAuthFlow', () => {
 		expect(mockCompleteOAuthFlow).not.toHaveBeenCalled();
 	});
 
-	it('sets inflight oauth promise and invokes `completeOAuthFlow` to complete an inflight oauth process', async () => {
-		(oAuthStore.loadOAuthInFlight as jest.Mock).mockResolvedValueOnce(true);
+	it('invokes `completeOAuthFlow` to complete an inflight oauth process', async () => {
+		(oAuthStore.loadOAuthInFlight as jest.Mock).mockResolvedValueOnce(Promise.resolve(true));
```
HuiSF, line 103: nit: As we use `mockResolvedValueOnce` here, probably don't need the `Promise.resolve()` right?

```suggestion
		(oAuthStore.loadOAuthInFlight as jest.Mock).mockResolvedValueOnce(true);
```

packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts
```diff
@@ -29,7 +31,16 @@ export class TokenOrchestrator implements AuthTokenOrchestrator {
...
+		if (!(await oAuthStore.loadOAuthInFlight())) {
+			return;
+		}
+		// when there is valid oauth config and there is an inflight oauth flow, try
+		// to block async calls that require fetching tokens before the oauth flow completes
+		// e.g. getCurrentUser, fetchAuthSession etc.
+
+		return new Promise<void>((resolve, _reject) => {
+			addInflightPromise(resolve);
+		});
```
HuiSF, line 43: We probably need to cache this promise and always return the same on when `waitForInflightOAuth` gets called multiple times instead of keeping pushing new promises.

HuiSF (commented):
Thanks for the fix!

packages/auth/src/providers/cognito/utils/oauth/completeOAuthFlow.ts
```diff
@@ -236,12 +235,6 @@ const completeFlow = async ({
 	//  `fetchAuthSession` can be resolved
 	resolveAndClearInflightPromises();
 
-	// when the oauth flow is completed, there should be nothing to block the async calls
-	// that involves fetchAuthSession in the `TokenOrchestrator`
-	cognitoUserPoolsTokenProvider.setWaitForInflightOAuth(async () => {
```
Samaritan1011001, line 241: Don't we need to set this elsewhere if not here?

elorzafe (approved):
Looks good
thanks @israx 🎖️

jimblanc (approved):
MSTM

packages/auth/src/providers/cognito/utils/oauth/completeOAuthFlow.ts
Samaritan1011001: I'll take that as a no then

PR #13163, opened 2024-03-22
chore: Add validation to ensure key & path not specified at the same time

jimblanc:
#### Description of changes

Adds runtime validation to ensure that key & path aren't specified at the same time to protect JavaScript customers from undefined behavior.

Local testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/storage/src/errors/types/validation.ts
+++ b/packages/storage/src/errors/types/validation.ts
@@ -60,7 +60,8 @@ export const validationErrorMap: AmplifyErrorMap<StorageValidationErrorCode> = {
 			'Upload source type can only be a `Blob`, `File`, `ArrayBuffer`, or `string`.',
 	},
 	[StorageValidationErrorCode.InvalidStorageOperationInput]: {
-		message: 'Missing path or key parameter in Input.',
+		message:
+			'Missing path or key parameter in Input. Key and path can not be specified at the same time.',
 	},
 	[StorageValidationErrorCode.InvalidStoragePathInput]: {
 		message: 'Input `path` does not allow a leading slash (/).',
--- a/packages/storage/src/errors/types/validation.ts
+++ b/packages/storage/src/errors/types/validation.ts
@@ -61,7 +61,7 @@ export const validationErrorMap: AmplifyErrorMap<StorageValidationErrorCode> = {
 	},
 	[StorageValidationErrorCode.InvalidStorageOperationInput]: {
 		message:
-			'Missing path or key parameter in Input. Key and path can not be specified at the same time.',
+			'Path or key parameter must be specified in the input. Both can not be specified at the same time.',
 	},
 	[StorageValidationErrorCode.InvalidStoragePathInput]: {
 		message: 'Input `path` does not allow a leading slash (/).',
--- a/packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
+++ b/packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
@@ -12,9 +12,11 @@ export const validateStorageOperationInput = (
 	input: Input,
 	identityId?: string,
 ) => {
-	// Validate key or path present
 	assertValidationError(
-		!!(input as Input).key || !!(input as Input).path,
+		// Key present without a path
+		(!!(input as Input).key && !(input as Input).path) ||
+			// Path present without a key
+			(!(input as Input).key && !!(input as Input).path),
 		StorageValidationErrorCode.InvalidStorageOperationInput,
 	);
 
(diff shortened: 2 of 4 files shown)


packages/storage/src/errors/types/validation.ts
```diff
@@ -60,7 +60,8 @@ export const validationErrorMap: AmplifyErrorMap<StorageValidationErrorCode> = {
 			'Upload source type can only be a `Blob`, `File`, `ArrayBuffer`, or `string`.',
 	},
 	[StorageValidationErrorCode.InvalidStorageOperationInput]: {
-		message: 'Missing path or key parameter in Input.',
+		message:
+			'Path or key parameter must be specified in the input. Both can not be specified at the same time.',
```
erinleigh90, line 64: Nit, non-blocking: This is more efficient, but wondering if the error message is too broad here..

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
```diff
@@ -12,11 +12,18 @@ export const validateStorageOperationInput = (
...
 		!!(input as Input).key || !!(input as Input).path,
 		StorageValidationErrorCode.InvalidStorageOperationInput,
 	);
 
+	// Validate key & path not present at the same time
+	assertValidationError(
+		(input as Input).key === undefined || (input as Input).path === undefined,
+		StorageValidationErrorCode.InvalidStorageOperationInput,
+	);
+
```
erinleigh90, line 26: If we are sending the same message, should we consolidate the validation as well?

erinleigh90 (approved):
Had some nit questions, but nothing blocking. Good with it as is.

packages/storage/src/errors/types/validation.ts
jimblanc: Yeah was thinking about this one. It's the same error code really, the path or key was specified incorrectly 🤔 I could break `InvalidStorageOperationInput` into two separate error codes if that's preferable, but I think it's fairly intuitive in this case.

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
jimblanc: I prefer to keep them separate I think (assuming we don't break up the error code) to improve readability

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
```diff
@@ -12,11 +12,18 @@ export const validateStorageOperationInput = (
...
 ) => {
+	// Validate key or path present
 	assertValidationError(
 		!!(input as Input).key || !!(input as Input).path,
 		StorageValidationErrorCode.InvalidStorageOperationInput,
 	);
 
+	// Validate key & path not present at the same time
+	assertValidationError(
+		(input as Input).key === undefined || (input as Input).path === undefined,
```
cshfang, line 23: Nit: Do we want to allow `null` to count as "not present"?

cshfang (dismissed):

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
Samaritan1011001: We could just check for the existence of the keys in them instead of the values. 
Something on these lines ? `!('prefix' in input && 'path' in input)`

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
```diff
@@ -12,11 +12,18 @@ export const validateStorageOperationInput = (
...
+		StorageValidationErrorCode.InvalidStorageOperationInput,
```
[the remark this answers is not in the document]
Samaritan1011001: Oh, didn't realise the message was changed. All good.

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
jimblanc: Only issue with this is it would trip up if the one of those is present but set to `undefined` 🤔 I think we need to be more explicit with the check, updated PR.

packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts
```diff
@@ -13,7 +13,16 @@ export const validateStorageOperationInput = (
...
-		!!(input as Input).key || !!(input as Input).path,
+		// Key present without a path
+		(!!(input as Input).key && !(input as Input).path) ||
+			// Path present without a key
+			(!(input as Input).key && !!(input as Input).path),
+		StorageValidationErrorCode.InvalidStorageOperationInput,
+	);
+
+	// Validate key & path not present at the same time
+	assertValidationError(
```
cshfang: Do you still need this given the above?

cshfang (approved):

ashika112 (approved):

PR #13164, opened 2024-03-22
feat(data): add 'sortDirection' support for list and index queries

david-mcafee:
#### Description of changes

feat(data): add 'sortDirection' support for list and index queries

[Corresponding types PR](https://github.com/aws-amplify/amplify-api-next/pull/135)

# ~~Note: PR checks will fail until the [corresponding types PR](https://github.com/aws-amplify/amplify-api-next/pull/135) has been merged~~

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/__tests__/internals/generateClient.test.ts
```diff
@@ -394,6 +394,7 @@ describe('generateClient', () => {
 			const client = generateClient<Schema>({ amplify: Amplify });
 			const { data } = await client.models.Todo.list({
 				filter: { name: { contains: 'name' } },
+				sortDirection: 'ASC',
```
dzhan-aws, line 397: Correct me if I'm wrong but here it seems like we are adding `sortDirection: 'ASC'` to make the tests pass, which makes me feel like this parameter is required instead of optional, does it mean customers need to always pass this `sortDirection` value?

david-mcafee: The tests pass without this parameter. However, similar to how we include `filter` I've included `sortDirection` in each test. If anyone feels strongly that this should not be included, I can instead create a sort direction-specific test.

packages/api-graphql/src/internals/APIClient.ts
```diff
@@ -732,6 +732,15 @@ export function generateGraphQLDocument(
 			graphQLArguments ??
 				(graphQLArguments = {
 					filter: `Model${name}FilterInput`,
+					sortDirection: 'ModelSortDirection',
+					...[primaryKeyFieldName, ...sortKeyFieldNames].reduce(
+						(acc: Record<string, any>, fieldName) => {
+							acc[fieldName] = `${fields[fieldName].type}!`;
```
iartemiev, line 738: The `!` after the type makes these required input fields for list operations. AppSync will reject this query if the PK/SK aren't present in variables. 

They need to be optional:
```suggestion
							acc[fieldName] = `${fields[fieldName].type}`;
```

iartemiev (changes requested):

packages/api-graphql/__tests__/internals/generateClient.test.ts
david-mcafee: Comments are outdated as new tests have been added to this PR.

iartemiev (approved):

packages/api-graphql/__tests__/internals/generateClient.test.ts
```diff
@@ -382,6 +382,82 @@ describe('generateClient', () => {
...
+								cpk_cluster_key: '1',
+								cpk_sort_key: 'b',
+							},
+							{
+								__typename: 'ThingWithCustomPk',
+								...serverManagedFields,
+								cpk_cluster_key: '1',
+								cpk_sort_key: 'c',
+							},
+						],
```
svidgen, line 408: You're not asserting on the correct parsing of these return items in any way. No need to rev at this point. But, going forward, I would generally skip the extra LOC.

packages/api-graphql/__tests__/internals/generateClient.test.ts
```diff
@@ -5098,6 +5174,166 @@ describe('generateClient', () => {
 				}),
 			);
 		});
+
+		/**
+		 * Though sorting happens at the AppSync level, we still test the return
+		 * result in order to protect against regressions to the sort order.
+		 */
+		test('PK and SK index query, with sort direction (ascending)', async () => {
```
svidgen, line 5182: I'm not sure I'm convinced of the value of this test and the one that follows. Are we trying to ensure the client doesn't change the ordering of the results? If so, these tests don't really convince me. E.g., the client could still be performing its own sorting on the result and pass these tests, but fail on more complex scenarios.

If we're worried about accidentally introducing sorting, I think I'd rather see a more concise test that "fuzzes" results with larger, auto-generated `items` collections. Rinse/Repeat 3x or something.

```ts
for (const i in [1,2,3]) {
  for (const d of ['ASC', 'DESC']) {
    test(`Client respects sort order returned by server (${d} run ${i})`, async () => {
      // generate 10x result, randomly sorted
      // establish mocking
      // run query with `sortDirection: d`
      // ensure client returns results back in same order as mocked
     })
  }
}
```

I don't want to block the whole PR on this. But, if we're actually worried that the client could change the order of results, let's follow up with something more convincing please. Else, I'd actually suggest scrapping these two tests.

(Or, have I misunderstood the point of these?)

packages/api-graphql/__tests__/internals/server/generateClientWithAmplifyInstance.test.ts
```diff
@@ -90,7 +90,7 @@ describe('server generateClient', () => {
 						}),
 						body: {
 							query: expect.stringContaining(
-								'listTodos(filter: $filter, limit: $limit, nextToken: $nextToken)',
+								'query ($filter: ModelTodoFilterInput, $sortDirection: ModelSortDirection, $id: ID, $limit: Int, $nextToken: String)',
```
svidgen, line 93: Here and below, I don't understand how/why this changed. I see all the fixtures are naming their tests `query`, so this certainly seems right. But, this doesn't look like a new convention, and the test body doesn't appear to have changed.

How was this testing passing prior to this change? Is this actually asserting what we expect? Please help me understand!

svidgen (commented):
Just one concern that prevents me from hitting the approve bubble:

The Q below about `listTodos` -> `query`. I don't understand how/why that changed. If it was incorrect previously, why weren't the tests failing? (Or were they? But, if they were, how did they make it into `main`???)

packages/api-graphql/__tests__/internals/server/generateClientWithAmplifyInstance.test.ts
iartemiev: Both/either would be valid here since we're matching on a substring via `expect.stringContaining`. 

The full document should look something like:

```gql
query ($filter: ModelTodoFilterInput, $sortDirection: ModelSortDirection, $id: ID, $limit: Int, $nextToken: String) {
  listTodos(
    filter: $filter,
    sortDirection: $sortDirection, 
    id: $id,
    limit: $limit,
    nextToken: $nextToken) { 
        items { ... }
    }
}
```

So I'd expect the assert to succeed for both `query...` and `listTodos...`.

That being said, the diff is confusing. Since we care about both the operation-level variables and the query field variables, let's include both in the assertion, like we have in `generateClient.test.ts.snap`

packages/api-graphql/__tests__/internals/generateClient.test.ts
david-mcafee: You're correct, and I had actually meant to remove these.

packages/api-graphql/__tests__/internals/generateClient.test.ts
david-mcafee: I think that the most important part of these tests is that they assert that a result is returned when including `sortDirection`. Correct me if I'm wrong, but I don't see a world were we perform any kind of sorting client-side, so removing that piece of these tests is fine with me (was including as a safeguard because it was an easy addition, but agree that the testing against regressions isn't exhaustive). 

However, I would still be in favor of having these tests as they are the only one that tests this functionality "E2E", but I can remove the testing of the sort order.

packages/api-graphql/__tests__/internals/server/generateClientWithAmplifyInstance.test.ts
david-mcafee: I was actually just adding both in right now, thanks for following up with this comment, Ivan!

svidgen: Ohhh ... right. That makes sense. I didn't catch that we were matching a completely separate line.

I think that satisfies me for this PR.

To seed a TODO for subsequent PR's, let's start moving in the direction of having *a few* snapshot tests with tests that assert on specific details about the graphql we're trying to clamp down on. E.g., let's move towards actually *parsing* the graphql and assert on the AST. We're already doing some of this in DataStore tests and now in the `data-schema` integ tests. Shouldn't be too much of a lift to start moving on that direction.

svidgen (approved):

packages/api-graphql/__tests__/internals/generateClient.test.ts
svidgen: OK. I'm not sure these are critical, but not enough doubt to ask to remove them then. Maybe we just need to work on our test naming!

svidgen: Well, and actually in looking again, what caught my eye was the comment that does suggest the purpose of these tests is to assert on something related to sort order...

It *might* actually be worth either a rev or quick followup PR to clarify what these tests are clamping down on. (And whether that's distinct from already-clamped behavior in other tests.)

david-mcafee: Updated.

svidgen (approved):

iartemiev (approved):

PR #13179, opened 2024-03-26
fix(data): update all operations to return flattened response, `null`, or empty array response on error

david-mcafee:
## NOTE: 
These changes have been ported over to the `amplify-api-next` repo, and a new PR has been created for the tests added to this repo:
https://github.com/aws-amplify/amplify-api-next/pull/187
https://github.com/aws-amplify/amplify-js/pull/13279

fix(data): update all operations to return flattened response, null, or empty array response on error

### Note: unit tests will not pass until the [related types PR](https://github.com/aws-amplify/amplify-api-next/pull/145) has been merged.

**As of today, there are several response shapes for `data` returned to the customer when an error is present. `data` can be:**
1. `null`
2. an empty object
3. "populated" but with a `null` value: `{ getPost: null }`
4. an actual record: `{ getPost: { id: '1', title: 'Hello, World!' } }`

This PR handles each of the above scenarios for all model operations by flattening the error response when necessary, and returning either `null` or an empty array unless actual data is present.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/src/internals/operations/get.ts
```diff
@@ -24,18 +24,19 @@ import {
...
 } from '@aws-amplify/core/internals/utils';
+import { handleGraphQlError } from './utils';
 
 export function getFactory(
 	client: ClientWithModels,
 	modelIntrospection: ModelIntrospectionSchema,
 	model: SchemaModel,
 	operation: ModelOperation,
-	useContext = false
+	useContext = false,
```
david-mcafee, line 34: We have always used trailing commas, but [this PR](https://github.com/aws-amplify/amplify-js/pull/13017) removed them during a Prettier upgrade. 

Will discuss during standup, but it's worth us making a decision on this now so we can either 1) reformat the library, or 2) change our local config to avoid these changes being included in future PRs

packages/api-graphql/src/internals/operations/utils.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export function handleGraphQlError(error: any) {
+	if (error.errors) {
+		// graphql errors pass through
+		return {
+			...error,
+			data: [],
```
svidgen, line 9: Just to get ahead of this: This only makes sense for list-type queries. Needs to probably just be `null` for non-list queries.

david-mcafee: Yep, still working on this! Thanks for the early review, nonetheless!

packages/api-graphql/src/internals/operations/utils.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+// import { GraphQLFormattedError } from '@aws-amplify/data-schema-types';
+
+/**
+ * Handle errors for list and index query operations
+ */
+export function handleListGraphQlError(error: any) {
```
david-mcafee, line 8: I considered using a single utility that accepted a flag (something like `isList`), but decided against it as this approach seems slightly less error prone / more straightforward to parse. This is only a slight preference, so if someone feels more strongly that `handleListGraphQlError` and `handleSingularGraphQlError` should be a single utility function, I'm happy to combine them.

packages/api-graphql/src/internals/operations/list.ts
```diff
@@ -124,12 +127,67 @@ async function _list(
...
+			const [key] = Object.keys(data);
+
+			if (data[key]?.items) {
+				const flattenedResult = flattenItems(data)[key];
+
+				/**
+				 * `flattenedResult` could be `null` here (e.g. `data: { getPost: null }`)
+				 * if `flattenedResult`, result is an actual record:
+				 */
+				if (flattenedResult) {
```
david-mcafee, line 152: This is required for `get`, currently testing if this is also required for `list`

david-mcafee: Just tested with a sample app - this is indeed required for `list`.

packages/api-graphql/src/internals/operations/custom.ts
```diff
@@ -413,12 +416,49 @@ async function _op(
...
+		if (data && !isEmpty(data) && errors) {
+			const [key] = Object.keys(data);
+			const flattenedResult = flattenItems(data)[key];
+
+			/**
+			 * `flattenedResult` could be `null` here (e.g. `data: { getPost: null }`)
+			 * if `flattenedResult`, result is an actual record:
+			 */
+			if (flattenedResult) {
+				// TODO: custom selection set. current selection set is default selection set only
```
david-mcafee, line 440: This is the same TODO that can be found with the existing implementation in the `try` block

packages/api-graphql/src/internals/operations/get.ts
```diff
@@ -129,12 +132,50 @@ async function _get(
...
+
+			/**
+			 * `flattenedResult` could be `null` here (e.g. `data: { getPost: null }`)
+			 * if `flattenedResult`, result is an actual record:
+			 */
+			if (flattenedResult) {
+				if (options?.selectionSet) {
+					return { data: flattenedResult, errors };
+				} else {
+					// TODO: refactor to avoid destructuring here
```
david-mcafee, line 159: This is the same TODO that can be found with the existing implementation in the `try` block

PR #13188, opened 2024-03-28
release(required): api-graphql - default selection set regression

iartemiev:
#### Issue #, if available

When generating the default selection set for a model, we were attempting to inject the `groupsField` of a static group auth rule e.g. `a.allow.specificGroup('Admin')`. Only dynamic group model introspection auth attributes contain a `groupsField` value, so we were inadvertently adding an explicit `undefined` to the default selection set, which was causing unexpected behavior.

This PR adds an additional check to make sure we're only adding the `groupsField` value when it's defined.

* Red/green TDD
* Manual e2e testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


stocaaro (approved):

jimblanc (approved):
Approving as a hot-fix, deferring to data team on the contents of the change

Issue #13191, opened 2024-03-29
iOS: UI Freezes on Canceling Sign-In via In-App Browser with Amplify


canrollas:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

Authentication

### Amplify Version

v6

### Amplify Categories

auth

### Backend

Amplify CLI

### Environment information

<details>

```


  System:
    OS: macOS 14.1.2
    CPU: (8) arm64 Apple M1
    Memory: 75.36 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.6.1 - /opt/homebrew/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.2.4 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 123.0.6312.87
    Safari: 17.1.2
  npmPackages:
    @aws-amplify/auth: ^6.0.23 => 6.0.23 
    @aws-amplify/auth/cognito:  undefined ()
    @aws-amplify/auth/cognito/server:  undefined ()
    @aws-amplify/auth/enable-oauth-listener:  undefined ()
    @aws-amplify/auth/server:  undefined ()
    @aws-amplify/core: ^6.0.23 => 6.0.23 
    @aws-amplify/core/internals/adapter-core:  undefined ()
    @aws-amplify/core/internals/aws-client-utils:  undefined ()
    @aws-amplify/core/internals/aws-client-utils/composers:  undefined ()
    @aws-amplify/core/internals/aws-clients/cognitoIdentity:  undefined ()
    @aws-amplify/core/internals/aws-clients/pinpoint:  undefined ()
    @aws-amplify/core/internals/providers/pinpoint:  undefined ()
    @aws-amplify/core/internals/utils:  undefined ()
    @aws-amplify/core/server:  undefined ()
    @aws-amplify/react-native: ^1.0.24 => 1.0.24 
[...]

PR #13196, opened 2024-03-30
feat(auth): fix inflight promise on RN

Samaritan1011001:
#### Description of changes

Currently we are only setting the [OAuthInflight flag](https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/providers/cognito/apis/signInWithRedirect.ts#L90) and waiting for it in the [TokenOrchestrator](https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts#L35) to get resolved(success/error/cancellation). This flow works well for web. However, for RN, we do not need this OAuthInflight flag to be set nor do we need to wait for it to get resolved. This fix involves removing both these actions for RN using the `isBrowser()` method.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13191

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
+++ b/packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
@@ -6,6 +6,7 @@ import {
 	AuthAction,
 	assertOAuthConfig,
 	assertTokenProviderConfig,
+	isBrowser,
 	urlSafeEncode,
 } from '@aws-amplify/core/internals/utils';
 
@@ -87,7 +88,7 @@ const oauthSignIn = async ({
 	const { value, method, toCodeChallenge } = generateCodeVerifier(128);
 	const redirectUri = getRedirectUrl(oauthConfig.redirectSignIn);
 
-	oAuthStore.storeOAuthInFlight(true);
+	if (isBrowser()) oAuthStore.storeOAuthInFlight(true);
 	oAuthStore.storeOAuthState(state);
 	oAuthStore.storePKCE(value);
 
--- a/packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts
+++ b/packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts
@@ -10,6 +10,7 @@ import {
 import {
 	AMPLIFY_SYMBOL,
 	assertTokenProviderConfig,
+	isBrowser,
 	isTokenExpired,
 } from '@aws-amplify/core/internals/utils';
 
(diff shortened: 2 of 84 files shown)


canrollas:
Hi, Thanks for your remarkable solution. I override it in my code by detecting the platform and changing the test-flight like this, but your solution seems more solid. I hope that will be fixed as soon as possible. 
```javascript
if (Platform.OS === 'ios') {
		oAuthStore.storeOAuthInFlight(false); // This is the first step of the OAuth flow, so we are not in flight
	} else {
		oAuthStore.storeOAuthInFlight(true); // This is the first step of the OAuth flow, so we are not in flight
	}
```

packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts
```diff
@@ -32,25 +33,29 @@ export class TokenOrchestrator implements AuthTokenOrchestrator {
 	tokenStore?: AuthTokenStore;
 	tokenRefresher?: TokenRefresher;
 	inflightPromise: Promise<void> | undefined;
-	waitForInflightOAuth: () => Promise<void> = async () => {
-		if (!(await oAuthStore.loadOAuthInFlight())) {
-			return;
-		}
+	waitForInflightOAuth: () => Promise<void> = isBrowser()
```
israx, line 36: I think it is best to separate this code into 2 files. It could be
- Web: `waitForInflightOAuth.ts`
- RN: `waitForInflightOAuth.native.ts`

It is more readable and helps with the bundle size as well.

packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
```diff
@@ -87,7 +88,7 @@ const oauthSignIn = async ({
 	const { value, method, toCodeChallenge } = generateCodeVerifier(128);
 	const redirectUri = getRedirectUrl(oauthConfig.redirectSignIn);
 
-	oAuthStore.storeOAuthInFlight(true);
+	if (isBrowser()) oAuthStore.storeOAuthInFlight(true);
```
israx, line 91: Probably this is not even needed

packages/auth/src/providers/cognito/tokenProvider/TokenOrchestrator.ts
canrollas: I believe this would be a concrete solution, but it is a major bug and has to be implemented asap in newer versions. Kind regards.

jimblanc: I like this idea 🤔 It also removes ambiguity around the usage of `isBrowser` on other environments (e.g. Node) and makes the change very specific to RN.

cshfang: I usually agree that we should isolate the logic via `.native.` extension but I think the risk here outweighs the benefit.

If we were to separate this out, the web function would have to accept a reference to this instance's `inflightPromise` to read/set it. I think it may actually add complexity to have a pure function be extracted out. Furthermore, I think it would fragment the logic between here and where we enable the OAuth listener.

It may be the right long term fix but we would have to evaluate more holistically than just updating the logic in this one place.

packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
israx: o nop only the `isBrowser` checker

jimblanc (commented):
Looks good, but bundle test is failing

Samaritan1011001:
Closing this in favor of this new PR, same set of changes just targeting release as the base branch. 
https://github.com/aws-amplify/amplify-js/pull/13204

PR #13239, opened 2024-04-11
Fix Overload return types for storage APIs

ashika112:
#### Description of changes
Storage APIs return type for unionized input can potentially break existing customers. This PR fixes the return types and adds path and key to all return types.

- Manual testing
- Unit Testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/storage/src/providers/s3/apis/downloadData.ts
```diff
@@ -160,6 +153,6 @@ const downloadDataJob =
 		};
 
 		return inputType === STORAGE_INPUT_KEY
-			? { key: objectKey, ...result }
-			: { path: finalKey, ...result };
+			? { key: objectKey, path: finalKey, ...result }
+			: { path: finalKey, key: finalKey, ...result };
```
erinleigh90, line 157: Can we just return key for key and path for path even in the path scenario? Seems weird to return the path for key, even though I know that we don't want them to use key going forward. Then this could be simplified as below:

`return { key: objectKey, path: finalKey, ...result }`

packages/storage/src/providers/s3/types/outputs.ts
```diff
@@ -28,137 +25,52 @@ export interface ItemBase {
...
+ * type for S3 list item.
  */
-export type ItemWithKey = ItemBase & StorageItemWithKey;
+export type ListOutputItem = Omit<ItemWithKeyAndPath, 'metadata'>;
 
-/**
- * type for S3 list item with path.
- */
-export type ItemWithPath = ItemBase & StorageItemWithPath;
+export type ItemWithKeyAndPath = ItemBase & StorageItem;
```
erinleigh90, line 32: Can this just be `Item` now? For every other scenario where we return both key and path we are not clarifying that it has both, so feels redundant.

erinleigh90 (commented):
In general the initial reason we wanted the overloads was due to the output mapping. Now that we don't need those mappings do we need function overloads in the same way? I do know it improves the doc strings and DX, just want to ask the question since overloads can be harder to maintain going forward.

packages/storage/src/providers/s3/apis/downloadData.ts
ashika112: this is to align with other libraries. Because technically, for path users there is no key if that makes sense

packages/storage/src/providers/s3/types/outputs.ts
ashika112: Since this is nit. I ll follow up on sep PR.

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -214,6 +214,7 @@ const _listWithPath = async ({
 	return {
 		items: response.Contents.map(item => ({
 			path: item.Key!,
+			key: item.Key!,
```
Samaritan1011001, line 217: We might want to figure out a way to have this value to not include the `generatedPrefix` in it. 

With these changes, there will be a discrepancy in the output. When they use prefix as `/photos`, we return key as `/photos` but if they use path `public/photos` we return key as `public/photos`

Samaritan1011001 (changes requested):
Want to make sure we are aware of this diff in output values for key

packages/storage/src/providers/s3/apis/internal/copy.ts
```diff
@@ -70,14 +68,14 @@ const copyWithPath = async (
 		s3Config,
 	});
 
-	return { path: finalCopyDestination };
+	return { path: finalCopyDestination, key: finalCopyDestination };
```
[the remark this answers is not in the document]
ashwinkumar6: lgtm since this is an expected behavior

Samaritan1011001 (approved):

ashwinkumar6 (approved):

packages/storage/src/providers/s3/apis/downloadData.ts
erinleigh90: In light of above this then becomes a harmless 
NIT:
Can we eventually simplify the output to be 
`return { key: objectKey, path: finalKey, ...result }`

packages/storage/__tests__/providers/s3/apis/list.test.ts
```diff
@@ -147,7 +147,9 @@ describe('list API', () => {
 					prefix: key,
 					options: options as ListPaginateOptionsWithPrefix,
 				});
-				expect(response.items).toEqual([{ ...listResultItem, key: key ?? '' }]);
+				expect(response.items).toEqual([
```
reesscot, line 150: Can you actually destructure out key and path to prove in your test that the types are working correctly? By doing your test this way, you don't allow TS to typecheck that `key` exists on the output.

packages/storage/src/providers/s3/types/outputs.ts
```diff
@@ -28,137 +25,52 @@ export interface ItemBase {
...
+export type GetPropertiesOutput = ItemWithKeyAndPath;
 
 /**
- * Output type for S3 list API. Lists bucket objects with pagination.
+ * Output type for S3 Copy API.
  */
-export type ListPaginateOutput = StrictUnion<
-	ListPaginateOutputWithPath | ListPaginateOutputWithPrefix
->;
+export type CopyOutput = Pick<ItemWithKeyAndPath, 'key' | 'path'>;
```
jimblanc, line 59: Nit: I realize this utility type usage was in the original PRs, but looking at it again I wonder if we can just have a explicitly defined type for this pattern. For example:

```
export type StorageItemLocation = {
	/** @deprecated This may be removed in next major version */
	key: string;
	path: string;
};
export type StorageItem = StorageItemBase & StorageItemLocation;
```

Okay maybe doing it in a separate PR if you agree

packages/storage/src/types/outputs.ts
```diff
@@ -27,25 +27,14 @@ export interface StorageItemBase {
...
-	 * Path of the object.
-	 */
-	path: string;
-};
-
 /**
  * A storage item can be identified either by a key or a path.
  */
-export type StorageItem = StorageItemWithKey | StorageItemWithPath;
+export type StorageItem = StorageItemBase & {
```
jimblanc, line 33: I wonder, can we combine `StorageItemBase` with `ItemBase` (seems a little redundant naming wise)? They both seem to contain meta-data about an object. Again okay refactoring in a follow up PR.

jimblanc (commented):
Few nits but looking good

packages/storage/__tests__/providers/s3/apis/remove.test.ts
```diff
@@ -84,12 +83,13 @@ describe('remove API', () => {
 				},
 			].forEach(({ options, expectedKey }) => {
 				const accessLevel = options?.accessLevel ?? 'default';
+				const removeResultKey = { key };
```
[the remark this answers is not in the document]
ashika112: this was all already there. Not much cleaning up honestly.

packages/storage/__tests__/providers/s3/apis/remove.test.ts
```diff
@@ -121,8 +117,13 @@ describe('remove API', () => {
 					path: ({ identityId }: any) => `protected/${identityId}/${key}`,
 				},
 			].forEach(({ path }) => {
+				const resolvePath =
```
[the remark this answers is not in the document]
ashika112: Same was already there

packages/storage/__tests__/providers/s3/apis/list.test.ts
```diff
@@ -147,7 +147,9 @@ describe('list API', () => {
...
+					{ ...listResultItem, key: key ?? '', path: expectedPath ?? '' },
```
calebpollman, line 151: Prefer assertion to expect static value, preferably typed

packages/storage/src/providers/s3/apis/downloadData.ts
ashika112: I prefer keeping verbose since i feel like `return { key: objectKey, path: finalKey, ...result }`  this suggests both are different but in cases of path it is not.

I feel like the verbosity will make us look again on implementation to see what is happening

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -150,23 +145,28 @@ const _listWithPrefix = async ({
...
-				: item.Key!,
-			eTag: item.ETag,
-			lastModified: item.LastModified,
-			size: item.Size,
-		})),
+				: item.Key!;
+
+			return {
+				key: finalKey,
+				path: item.Key!,
```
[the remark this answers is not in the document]
ashika112: This is what it has always been. Looking at the type from s3 itself, ListV2API actually returns a optional key. 

i think we decided to assert non-null to make sure a value is returned to customer maybe bcoz it is required on us.

REF: https://github.com/aws-amplify/amplify-js/blob/main/packages/storage/src/providers/s3/apis/internal/list.ts#L116

packages/storage/src/types/outputs.ts
ashika112: Yeah did want to change bunch of stuff in terms of type cleanup here. But these two where separated since v6, i dont know if there was more reasoning to that.

packages/storage/src/providers/s3/types/outputs.ts
ashika112: i think it is reasonable, I can add it to the test upgrade PR.

packages/storage/__tests__/providers/s3/apis/list.test.ts
```diff
@@ -174,15 +176,17 @@ describe('list API', () => {
 					};
 				});
 				const customPageSize = 5;
-				const response = await list({
+				const response: ListPaginateOutput = await list({
```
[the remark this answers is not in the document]
ashika112: Yes Typescript has problem inferring it.

jimblanc (approved):

calebpollman (commented):
Thanks for doing this! Potentially out of scope for this PR but something that i'd like to see addressed is a clean up of the typing on the Storage function declarations, and removal of the casting of the output types in the function bodies. Until the outputs are no longer casted it's hard to say that what we are shipping is type safe. Thoughts?

packages/storage/__tests__/providers/s3/apis/list.test.ts
ashika112: Hey @reesscot , I adding more test around finding TS issues that has come bitten us multiple times around wrapper.

Working on sep PR -> https://github.com/ashika112/amplify-js/pull/3

I will include UI folks in that one as well. Unless you feel it should go with the same PR

packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
```diff
@@ -153,6 +154,7 @@ describe('downloadData with key', () => {
 		expect(getObject).toHaveBeenCalledTimes(1);
 		expect(result).toEqual({
 			key,
+			path: `public/${key}`,
```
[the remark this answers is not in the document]
ashika112: nice catch.

packages/storage/__tests__/providers/s3/apis/remove.test.ts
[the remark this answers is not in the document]
jimblanc: Technically before it was a function, in which the verb made more sense

packages/storage/__tests__/providers/s3/apis/list.test.ts
reesscot: +1, it feels a little odd to allow this to be an empty string in a test we control?

packages/storage/__tests__/providers/s3/apis/list.test.ts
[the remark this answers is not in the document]
reesscot: Do we have an existing issue to track down why it's not being inferred?

packages/storage/__tests__/providers/s3/apis/list.test.ts
ashika112: Im working on sep PR to update test. But this has just copy atm - Can u provide more feedback there?

https://github.com/ashika112/amplify-js/pull/3/files

packages/storage/__tests__/providers/s3/apis/list.test.ts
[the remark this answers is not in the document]
ashika112: I m removing this. this was an IDE bug. tried couple of fresh build not an issue. Happens only inbtw builds

calebpollman (approved):

reesscot (approved):

PR #13246, opened 2024-04-12
Chore: Update Storage test suite

ashika112:
#### Description of changes
This PR aims to cover storage API tests to assert types and remove type casting on outputs.

This is by no means an exhaustive type test. But aims to cover some aspects of it. 

- Unit Tests
- Integ Tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/storage/__tests__/providers/s3/apis/copy.test.ts
+++ b/packages/storage/__tests__/providers/s3/apis/copy.test.ts
@@ -10,6 +10,8 @@ import { copy } from '../../../../src/providers/s3/apis';
 import {
 	CopySourceOptionsWithKey,
 	CopyDestinationOptionsWithKey,
+	CopyInput,
+	CopyOutput,
 } from '../../../../src/providers/s3/types';
 
 jest.mock('../../../../src/providers/s3/utils/client');
@@ -48,6 +50,10 @@ const copyObjectClientBaseParams = {
 	Bucket: bucket,
 	MetadataDirective: 'COPY',
 };
+
+// Adding the wrapper to catch some of type misses we had on wrappers before
+const copyWrapper = async (input: CopyInput): Promise<CopyOutput> =>
+	copy(input);
 
 describe('copy API', () => {
 	beforeAll(() => {
@@ -167,7 +173,7 @@ describe('copy API', () => {
 
 					it(`should copy ${source.accessLevel} ${targetIdentityIdMsg} -> ${destination.accessLevel}`, async () => {
 						expect(
-							await copy({
+							await copyWrapper({
 								source: {
 									...(source as CopySourceOptionsWithKey),
 									key: sourceKey,
@@ -223,7 +229,7 @@ describe('copy API', () => {
 					expectedDestinationPath,
 				}) => {
 					expect(
-						await copy({
+						await copyWrapper({
 							source: { path: sourcePath },
 							destination: { path: destinationPath },
 						}),
@@ -257,7 +263,7 @@ describe('copy API', () => {
 			const sourceKey = 'SourceKeyNotFound';
 			const destinationKey = 'destinationKey';
 			try {
-				await copy({
+				await copyWrapper({
 					source: { key: sourceKey },
 					destination: { key: destinationKey },
 				});
@@ -276,7 +282,7 @@ describe('copy API', () => {
 			expect.assertions(2);
 			try {
 				// @ts-expect-error
-				await copy({
+				await copyWrapper({
 					source: { path: 'sourcePath' },
 					destination: { key: 'destinationKey' },
 				});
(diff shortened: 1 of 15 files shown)


packages/storage/__tests__/providers/s3/apis/uploadData/putObjectJob.test.ts
```diff
@@ -146,33 +164,51 @@ describe('putObjectJob with path', () => {
...
 						onProgress,
 						useAccelerateEndpoint,
 					},
 				},
 				abortController.signal,
 			);
-			const result = await job();
-			expect(result).toEqual({
+			const {
+				key,
```
reesscot, line 193: Quesiton: Are these test files type checked in CI/CD or just in VS Code? If a type error shows up in a test, will the build fail?

packages/storage/__tests__/providers/s3/apis/copy.test.ts
```diff
@@ -166,18 +172,17 @@ describe('copy API', () => {
...
-								},
-								destination: {
-									...(destination as CopyDestinationOptionsWithKey),
-									key: destinationKey,
-								},
-							}),
-						).toEqual(copyResult);
+						const { key, path } = await copyWrapper({
+							source: {
+								...(source as CopySourceOptionsWithKey),
```
calebpollman, line 177: If we cast our test input how are we guaranteeing that there is no typing issue with the function input typing?

ashika112: I thought about this. this is how we do all across our libraries. There are some tests that capture the options value down the chain without casting. Not sure if its exhaustive. 

Also, i noticed in storage it is only on options and that is bcoz of the [Array] we input in the mock.

packages/storage/__tests__/providers/s3/apis/uploadData/putObjectJob.test.ts
ashika112: hmm.. unfortunately i dont think this will fail our tests we dont have checks enabled. I think it will only be in VSC. I ll check with the team.

ashika112: There is bunch of changes going in enabling the linting rules and ts validation on tests. that should catch these issues.

packages/storage/__tests__/providers/s3/apis/copy.test.ts
calebpollman: Think this PR provides a good opportunity to improve gaps in at the Storage unit tests irregardless of the rest of the library. Test inputs of the tested function should not be casted, with intentionally invalid inputs using `@ts-expect-error`, e.g.:
```ts
it('should throw if provided both a `key` and `path` in the input`, () => {
  const invalidInput: UseGetUrlInput = {
    path: 'test-path',
    // @ts-expect-error force invalid input
    key: 'test-key',
  }
  // ...do test stuff
})
```

ashika112: We do have some tests like this to do validate assertions /error we throw. I didnt see any tests casting, including `path` just use options directly. The ones casting options here are for key, it is particularly doing it for the ones where we are pulling array of options. There is a type error on `accessLevel` field, if we explicitly write the option into the functions ts knows to infer. But doing that seemed a lot. 

If you see places outside options with key casting, i would be happy to change to look into those ones.

packages/storage/__tests__/providers/s3/apis/uploadData/putObjectJob.test.ts
haverchuck: Is there a reason we don't have checks enabled?

packages/storage/__tests__/providers/s3/apis/copy.test.ts
reesscot: But isn't the need for casting a symptom of the way this is being tested? Do we really need to be using a forEach loop here? Wouldn't the types work better if these were tested individually by use case?

ashika112: So with `key` in input parameter, there is various accessLevel combination that needs to be tested and hence the `forEach` i m sure there must be ways to work around this. I haven't yet had the time to prioritize this and doing that refactor may take longer since that is how every single test is in lib. 

TAYQ: Yes, removing cast will make the test the type properly but i would rather do it in steps over no test improvements. Right now the tests adding path as inputs i have updated and to some extend on ones on key as well.

packages/storage/__tests__/providers/s3/apis/uploadData/putObjectJob.test.ts
ashika112: we have never had it. Hui's work on enabling eslint on tests, is started is done in phases. But that work is not prioritized atm and will be resumed soon to enable on all test suites.

ashika112:
Closing this. It is now folded in gen2-storage branch in another PR and all casting where removed. Ref: https://github.com/aws-amplify/amplify-js/pull/13270

PR #13234, opened 2024-04-09
Fix: relax config for AmplifyOutputs

elorzafe:
#### Description of changes

Relax types of `AmplifyOutputs` to prevent TypeScript problems while importing JSON file, also updating with latests schema.

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://amplify.aws/2024-02/outputs-schema.json",
  "title": "AWS Amplify Backend Outputs",
  "description": "Config format for Amplify Gen 2 client libraries to communicate with backend services.",
  "type": "object",
  "additionalProperties": false,
  "properties": {
    "$schema": {
      "description": "JSON schema",
      "type": "string"
    },
    "version": {
      "description": "Version of this schema",
      "const": "1"
    },
    "analytics": {
      "description": "Outputs manually specified by developers for use with frontend library",
      "type": "object",
      "additionalProperties": false,
      "properties": {
        "amazon_pinpoint": {
          "type": "object",
          "additionalProperties": false,
          "properties": {
            "aws_region": {
              "description": "AWS Region of Amazon Pinpoint resources",
              "$ref": "#/$defs/aws_region"
            },
            "app_id": {
              "type": "string"
            }
          },
          "required": ["aws_region", "app_id"]
        }
      }
    },
    "auth": {
      "description": "Outputs generated from defineAuth",
      "type": "object",
      "additionalProperties": false,
      "properties": {
        "aws_region": {
          "description": "AWS Region of Amazon Cognito resources",
          "$ref": "#/$defs/aws_region"
        },
        "authentication_flow_type": {
          "description": "Authentication flow types",
          "type": "string",
          "enum": ["USER_SRP_AUTH", "CUSTOM_AUTH"],
          "default": "USER_SRP_AUTH"
        },
        "user_pool_id": {
          "description": "Cognito User Pool ID",
          "type": "string"
        },
        "user_pool_client_id": {
          "description": "Cognito User Pool Client ID",
          "type": "string"
        },
        "identity_pool_id": {
          "description": "Cognito Identity Pool ID",
          "type": "string"
        },
        "password_policy": {
          "description": "Cognito User Pool password policy",
          "type": "object",
          "additionalProperties": false,
          "properties": {
            "min_length": {
              "type": "integer",
              "minimum": 6,
              "maximum": 99
            },
            "require_numbers": {
              "type": "boolean"
            },
            "require_lowercase": {
              "type": "boolean"
            },
            "require_uppercase": {
              "type": "boolean"
            },
[...]

PR #13247, opened 2024-04-12
fix: Fix SSR & AmplifyOutput types when using Gen2 configuration files

jimblanc:
#### Description of changes

This change updated the `AmplifyOutputs` type & SSR utilities to fully support the Gen2 configuration file format.

Changes:
- Loosen up types on the `model_introspection` field in `AmplifyOutputs`
  - This is required as TypeScript is not able to properly infer `ModelIntrospectionSchema` from the JSON file. This is only required for configuration file parsing, ultimately the type present in `ResourcesConfig` (which the library uses internally after config file parsing) will be `ModelIntrospectionSchema`. For more context see: https://github.com/aws-amplify/amplify-js/pull/13234
- Add support for `AmplifyOutputs` to SSR utilities to enable server-side API usage against a Gen2 back-end
- Add support for `AmplifyOutputs` to the Amplify singleton
  - **Important context:** This change is the root cause of the apparent bundle size increases observed in this PR. Previously, support for `AmplifyOutputs` was only added to `initSingleton` (https://github.com/aws-amplify/amplify-js/pull/13150), which is the primary method by which customers access the singleton. As our APIs interact with the `Amplify` singleton directly (to retrieve configuration, for example), the bundle size impact of adding Gen2 parsing to `initSingleton` (which customers would almost always use in v6) was obfuscated from the bundle size tests. The changes to the bundle limits in this PR simply bring the tests in-line with what real customers would see in practice as a result of the Gen2 changes.
- Standardized config file parsing across the library
  - Previously we were doing this individually in three different places (i.e. `initSingleton`, `Amplify`, and `adapter-nextjs`), standardized all three places to use the same config parsing utility.

- Tested Gen2 config file parsing in a sample app with both auth & data resources configured
- Smoke tested a handful of SSR APIs using a Gen2 configuration file

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/adapter-nextjs/src/utils/getAmplifyConfig.ts
jimblanc, line 1: This SSR-specific utility has been replaced with the universal `parseAmplifyConfig` utility.

packages/aws-amplify/package.json
jimblanc, line 1: See PR description for context on apparent bundle size increases. The delta in these increases line up pretty well with the size of the new Gen2 parser.

packages/core/src/singleton/Amplify.ts
```diff
@@ -48,16 +49,10 @@ export class AmplifyClass {
 	 * @param libraryOptions - Additional options for customizing the behavior of the library.
 	 */
 	configure(
-		resourcesConfig: ResourcesConfig | LegacyConfig,
+		resourcesConfig: ResourcesConfig | LegacyConfig | AmplifyOutputs,
 		libraryOptions?: LibraryOptions,
 	): void {
-		let resolvedResourceConfig: ResourcesConfig;
```
jimblanc, line 54: Amplify singleton did not previously support Gen2 configuration files.

packages/core/__tests__/utils/parseAmplifyConfig.test.ts
```diff
@@ -0,0 +1,73 @@
+import { ResourcesConfig } from '../../src';
+import { parseAmplifyConfig } from '../../src/libraryUtils';
+import { parseAWSExports } from '../../src/parseAWSExports';
+import { isAmplifyOutputs, parseAmplifyOutputs } from '../../src/parseAmplifyOutputs';
+
+jest.mock('../../src/parseAWSExports');
+jest.mock('../../src/parseAmplifyOutputs');
+
+const testAmplifyOutputs = {
```
Samaritan1011001, line 9: Do we want to test the parsing of the complete config given that we check for each block in the `parseAmplifyOutputs` function?

Samaritan1011001 (approved):
thank you!

packages/adapter-nextjs/src/utils/getAmplifyConfig.ts
ashika112: Wondering if there is anything more to why this was not moved to an utility and use across in first place?

packages/core/__tests__/utils/parseAmplifyConfig.test.ts
jimblanc: IMO we don't need to here considering each parse function has its own tests. This test is simply ensuring that the correct parser is selected for a given config shape

packages/core/src/parseAmplifyOutputs.ts
```diff
@@ -215,7 +216,7 @@ function parseData(
 		defaultAuthMode: getGraphQLAuthMode(default_authorization_type),
 		region: aws_region,
 		apiKey: api_key,
-		modelIntrospection: model_introspection,
+		modelIntrospection: model_introspection as ModelIntrospectionSchema,
```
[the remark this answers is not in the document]
jimblanc: If we're not able to infer the type correctly from the JSON, we don't really have a choice IMO. This is similar to what we're doing already in `parseAWSExports` (expect there it's `any`). We can't lean on TypeScript alone to validate config structure regardless for our non-TS customers, e.g. how we make runtime assertions about configuration state elsewhere in the library.

packages/adapter-nextjs/src/utils/getAmplifyConfig.ts
jimblanc: Checked with Hui, originally we were reading the configuration in a different way for SSR, but that changed before launch. No reason to not unify at this time as far as I can tell.

HuiSF (commented):
Thanks for consolidating the parser function, it looks much cleaner now. 

One thing we've might missed is that we are currently exporting the original `parseAmplifyConfig` function from the path `aws-amplify/utils`, to support the generic SSR adapter [use cases](https://docs.amplify.aws/react/build-a-backend/server-side-rendering/nuxt/#implement-01amplify-apisserverts). 

We may need to update the function being exported here:

https://github.com/aws-amplify/amplify-js/blob/1625424a5f04ed6eb2ad9386ffc31d8e25f88e90/packages/aws-amplify/src/utils/index.ts#L21C10-L21C25

packages/core/__tests__/utils/parseAmplifyConfig.test.ts
```diff
@@ -0,0 +1,73 @@
...
+	}
+};
+
+describe('parseAmplifyConfig', () => {
+	const mockParseAWSExports = parseAWSExports as jest.Mock;
+	const mockParseAmplifyOutputs = parseAmplifyOutputs as jest.Mock;
+	const mockIsAmplifyOutputs = isAmplifyOutputs as unknown as jest.Mock;
+
+	beforeEach(() => {
+		jest.clearAllMocks();
```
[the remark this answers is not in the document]
jimblanc: I'd prefer to keep this in place to be able to make assertions about how many times something is called.

jimblanc:
> Thanks for consolidating the parser function, it looks much cleaner now.
> 
> One thing we've might missed is that we are currently exporting the original `parseAmplifyConfig` function from the path `aws-amplify/utils`, to support the generic SSR adapter [use cases](https://docs.amplify.aws/react/build-a-backend/server-side-rendering/nuxt/#implement-01amplify-apisserverts).
> 
> We may need to update the function being exported [here](https://github.com/aws-amplify/amplify-js/blob/1625424a5f04ed6eb2ad9386ffc31d8e25f88e90/packages/aws-amplify/src/utils/index.ts#L21C10-L21C25).

Good catch with the generic SSR adapter use-case, updated to export `parseAmplifyConfig` directly. While the type surface is technically different, it should still work for anyone passing in a config object as described in the docs.

HuiSF (approved):

ashika112 (approved):

Samaritan1011001 (approved):

PR #13249, opened 2024-04-14
fix(api-graphql): incorrect list sk arg type

iartemiev:
#### Description of changes

We were generating list query arguments for sort keys using the underlying field's primitive type as the argument type.

List queries on models with composite primary keys expect a _condition_ input argument type for the sort key(s).

See this diff for current vs expected behavior: https://github.com/aws-amplify/amplify-js/compare/data-fix-sk-args?expand=1#diff-fa24101f776e07b54f8416746bbed348b96dd7e855c0eed2e5609bcb501a9b46L4095-R4095

* updated e2e test to include this use case - https://github.com/aws-amplify/amplify-js-samples-staging/pull/845

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/api-graphql/__tests__/internals/APIClient.test.ts
+++ b/packages/api-graphql/__tests__/internals/APIClient.test.ts
@@ -563,7 +563,7 @@ describe('generateGraphQLDocument()', () => {
 					modelOperation
 				);
 
-				expect(document.includes(expectedArgs)).toBe(true);
+				expect(document).toEqual(expect.stringContaining(expectedArgs))
 			}
 		);
 	});
--- a/packages/api-graphql/__tests__/internals/__snapshots__/generateClient.test.ts.snap
+++ b/packages/api-graphql/__tests__/internals/__snapshots__/generateClient.test.ts.snap
@@ -4092,11 +4092,11 @@ exports[`generateClient basic model operations can list() with sortDirection (AS
       "abortController": AbortController {},
       "options": {
         "body": {
-          "query": "query ($cpk_cluster_key: String, $cpk_sort_key: ModelStringKeyConditionInput, $sortDirection: ModelSortDirection, $filter: ModelThingWithCustomPkFilterInput, $limit: Int, $nextToken: String) {
+          "query": "query ($cpk_cluster_key: String, $sortDirection: ModelSortDirection, $cpk_sort_key: ModelStringKeyConditionInput, $filter: ModelThingWithCustomPkFilterInput, $limit: Int, $nextToken: String) {
   listThingWithCustomPks(
     cpk_cluster_key: $cpk_cluster_key
-    cpk_sort_key: $cpk_sort_key
     sortDirection: $sortDirection
+    cpk_sort_key: $cpk_sort_key
     filter: $filter
     limit: $limit
     nextToken: $nextToken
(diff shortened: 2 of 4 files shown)


packages/api-graphql/__tests__/internals/APIClient.test.ts
```diff
@@ -563,7 +563,7 @@ describe('generateGraphQLDocument()', () => {
 					modelOperation
 				);
 
-				expect(document.includes(expectedArgs)).toBe(true);
+				expect(document).toEqual(expect.stringContaining(expectedArgs))
```
iartemiev, line 566: Same effect, but this outputs a helpful diff if the object doesn't match, whereas previously it would only tell us got `false` expected `true`.

packages/api-graphql/__tests__/internals/__snapshots__/generateClient.test.ts.snap
```diff
@@ -4092,12 +4092,12 @@ exports[`generateClient basic model operations can list() with sortDirection (AS
       "abortController": AbortController {},
       "options": {
         "body": {
-          "query": "query ($filter: ModelThingWithCustomPkFilterInput, $sortDirection: ModelSortDirection, $cpk_cluster_key: String, $cpk_sort_key: String, $limit: Int, $nextToken: String) {
+          "query": "query ($cpk_cluster_key: String, $cpk_sort_key: ModelStringKeyConditionInput, $sortDirection: ModelSortDirection, $filter: ModelThingWithCustomPkFilterInput, $limit: Int, $nextToken: String) {
```
iartemiev, line 4095: `Model${fieldType}KeyConditionInput` is the expected type for SKs in `list` operations

packages/api-graphql/src/internals/APIClient.ts
```diff
@@ -721,6 +721,44 @@ export function generateGraphQLDocument(
...
+					acc[fieldName] = `${fieldType}!`;
+				} else if (op === 'list') {
+					acc[fieldName] = `Model${fieldType}KeyConditionInput`;
+				}
+
+				return acc;
+			},
+			{},
+		);
+	};
```
HuiSF, line 746: nit: can use `Object.fromEntries` to save a few bytes :D (still over the limit a bit though...)
```suggestion
	const generateSkArgs = (op: 'get' | 'list') =>
		Object.fromEntries(
			sortKeyFieldNames.map(fieldName => {
				const fieldType = fields[fieldName].type;

				return [
					fieldName,
					op === 'get' ? `${fieldType}!` : `Model${fieldType}KeyConditionInput`,
				];
			}),
		);
```

packages/api-graphql/src/internals/APIClient.ts
```diff
@@ -721,6 +721,44 @@ export function generateGraphQLDocument(
...
+
+	if (isCustomPrimaryKey) {
+		Object.assign(getPkArgs, {
+			...generateSkArgs('get'),
```
HuiSF, line 750: nit: probably we don't need the spread operator here

packages/api-graphql/src/internals/APIClient.ts
```diff
@@ -721,6 +721,44 @@ export function generateGraphQLDocument(
...
+		});
+
+		Object.assign(listPkArgs, {
+			// PK is only included in list query field args in the generated GQL
+			// when explicitly specifying PK with .identifier(['fieldName']) or @primaryKey in the schema definition
+			[primaryKeyFieldName]: `${fields[primaryKeyFieldName].type}`, // PK is always a nullable arg for list (no `!` after the type)
+			...generateSkArgs('list'),
```
HuiSF, line 757: nit: ditto, `generateSkArgs('list')` can be the 3rd param of this `Object.assign()`

packages/api-graphql/src/internals/APIClient.ts
iartemiev: I think this is hard to read/grok. I'd rather keep it very simple since we've had lots of bugs in this code already.

packages/api-graphql/src/internals/APIClient.ts
iartemiev: Updated

packages/api-graphql/src/internals/APIClient.ts
iartemiev: Updated

HuiSF (approved):

jimblanc (approved):
Approving umbrella package.json changes

elorzafe (approved):

david-mcafee (approved):

PR #13263, opened 2024-04-18
refactor: move data-driven api-graphql property builders to data-schema package

svidgen:
#### Description of changes

Moves data-driven `client` properties from `generateClient()` to the `data-schema` package and bumps the dep.

Leaves a good bulk of data-driven `generateClient()` tests in place *for now* until we can determine a minimal set of tests necessary to prevent core client functionality from breaking data-driven use-cases. The majority of these tests will ultimately be moved to `data-schema` and/or the associated integ test package.

The changes to `data-schema` also require changes to schema definition. Tests are updated accordingly.

A few type casts were changed to `as any` casts in an effort to reduce TS type check counts. As types improve on the `data-schema` side, we can revisit these casts.

1. Manual e2e testing.
2. e2e integ test will also be updated. (WIP)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/__tests__/utils/index.ts
```diff
@@ -33,7 +33,7 @@ export function normalizePostGraphqlCalls(spy: jest.SpyInstance<any, any>) {
 		const [_, postOptions] = call;
 		const userAgent = postOptions?.options?.headers?.['x-amz-user-agent'];
 		if (userAgent) {
-			const staticUserAgent = userAgent.replace(/\/[\d.]+/g, '/latest');
+			const staticUserAgent = userAgent.replace(/\/[\w\d.+-]+/g, '/latest');
```
svidgen, line 36: I don't completely remember why I needed this. Something about running with tagged releases or versions locally against verdaccio caused the normalization to leave version artifacts. Failure to consume dashes, etc..

david-mcafee (approved):

packages/api-graphql/__tests__/GraphQLAPI.test.ts
```diff
@@ -1142,7 +1142,7 @@ describe('API test', () => {
 				[__amplify]: AmplifyCore,
 				graphql,
 				cancel,
-			} as V6Client;
+			} as unknown as V6Client;
```
jimblanc, line 1145: Sanity check: Why do we need to do this? Just making sure it's not indicative of a type error somewhere

packages/api/package.json
```diff
@@ -68,6 +68,8 @@
 	},
 	"homepage": "https://aws-amplify.github.io/",
 	"devDependencies": {
+		"@rollup/plugin-typescript": "11.1.5",
+		"rollup": "^4.9.6",
```
jimblanc, line 72: Nit: Do we need these? Have them as dev deps in the mono-repo as well

scripts/tsc-compliance-test/package.json
```diff
@@ -4,7 +4,7 @@
 	"license": "MIT",
 	"private": true,
 	"devDependencies": {
-		"@types/node": "^16.11.7",
+		"@types/node": "16.18.82",
```
jimblanc, line 7: I'm surprised this change was needed 🤔 Could we just update it in package lock and keep the range?

yarn.lock
jimblanc, line 1: You're already on it, but can we stay with the yarn registry

jimblanc (commented):
Focused review on dependencies, package.json files, and mono-repo changes.

yarn.lock
svidgen: Yep. Trying to fix this now.

svidgen: Re-pulled from `main` and restart VS Code. Maybe `yarn` defers to an environment variable that I didn't remember I'd set? 🤷

scripts/tsc-compliance-test/package.json
svidgen: Are you are we don't want to pin it? If moving past this version breaks the checks, wouldn't we want this pinned so the next chump doesn't have to re-diagnose the problem?

packages/api-graphql/__tests__/GraphQLAPI.test.ts
svidgen: Yeah. This client mock isn't *quite* a `V6Client`. But, it's close enough for the test.

yarn.lock
svidgen: Huh ... actually. I didn't notice, but `yarn` still some of the entries with `npmjs.org`. No idea where it's pulling that config from quiet yet.

svidgen: Still baffled. I've replaced the registry manually.

scripts/tsc-compliance-test/package.json
svidgen: As discussed offline, going beyond this version breaks things. I tried 18 as well. I'd argue pretty strongly in favor of leaving this as-is to prevent any further yarn lock regenerations from causing problems again.

david-mcafee (approved):

jimblanc (approved):

packages/api-graphql/__tests__/server/generateClient.test.ts.bak
HuiSF, line 1: Is this filed needed? 😅

svidgen: doh!

Issue #13274, opened 2024-04-20
Amplify Authenticator: Buttons "frozen" after back clicking social login 


alex-breen:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Vite

### What browsers are you seeing the problem on?

Chrome, Firefox

### Please describe your bug.

Steps to reproduce: 
1. Load login page running `Authenticator` .
2. Click `Sign in with Google`.
3. Click `back` on your browser to return to the login page.
4. Click `Sign in with Google`.

Observe:
1. All buttons on page are unresponsive.
2. No errors logged.
3. Only workaround is to reload page.

### What's the expected behaviour?

Buttons work after if you click back to the login page. 

### Help us reproduce the bug!

Strangely, this occurs only on a production build, but not on a dev build. 
Ie: Using vite, if you `npm run dev` the behavior is not observed. But if you `npm run build` and `npm run preview` the frozen behavior is observed. 

Reproduction: I've pasted code snippets. But basically this is the came code as documented in v6 of https://docs.amplify.aws/react/start/getting-started/ and https://ui.docs.amplify.aws/react/connected-components/authenticator.

### Code Snippet

```typescript
// Put your code below this line.

-------- index.html  -----------
<!doctype html>
<html lang="en">

<head>
  <meta charset="UTF-8" />
  <link rel="icon" type="image/svg+xml" href="/vite.svg" />
  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
  <link rel="manifest" href="manifest.json" />
  <title>Vite + React</title>
</head>

<body>
  <div id="root"></div>
  <script type="module" src="/src/main2.jsx"></script>
</body>

</html>

--------  main2.jsx  -----------
import React from 'react'
import ReactDOM from 'react-dom/client'
import App3 from './App3.jsx'

import { Amplify } from 'aws-amplify';
[...]

PR #13278, opened 2024-04-22
fix(auth): listen for `pageshow` event listener

israx:
#### Description of changes

Currently when bfcache is enabled and the user clicks the back button, amplify is not able to detect the cancellation because `Amplify.configure` is never called. This change listens for the `pageshow` window event and clears any inflight oauth keys

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13274

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.navite.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
cshfang, line 1: *native.ts

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { createOAuthError } from './createOAuthError';
+import { handleFailure,  } from './index';
+
+export const listenForOAuthFlowCancellation = (store: OAuthStore) => {
+	async function handleCancelOauthFlow(event: PageTransitionEvent) {
+		const isOAuthInFlight = await store.loadOAuthInFlight();
+		if (event.persisted && isOAuthInFlight) {  
+			await handleFailure(createOAuthError('User cancelled OAuth flow.'));
+			window.removeEventListener('pageshow', handleCancelOauthFlow);
```
Samaritan1011001, line 14: Should this be removed regardless this `if` clause is visited? Currently it will get removed only if this particular scenario happens (i.e it was cancelled or back pressed)

cshfang: I was thinking about this too, but I guess if the app isn't reopened via bfcache, then it would presumably be a new window? I was trying to think through the various scenarios and this seems safe but can we just double check to make sure there's no possibility of a memory leak?

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { OAuthStore } from '../types';
+
+import { createOAuthError } from './createOAuthError';
+import { handleFailure,  } from './index';
```
calebpollman, line 7: Not related to the PR but think the noise in this PR (and others) applied by linting errors could be fixed by running the lint command as a precommit hook cc @jimblanc @HuiSF

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const listenForOAuthFlowCancellation = (store: OAuthStore) => {
+	async function handleCancelOauthFlow(event: PageTransitionEvent) {
+		const isOAuthInFlight = await store.loadOAuthInFlight();
+		if (event.persisted && isOAuthInFlight) {  
+			await handleFailure(createOAuthError('User cancelled OAuth flow.'));
+			window.removeEventListener('pageshow', handleCancelOauthFlow);
+		}
+	}
+	window.addEventListener('pageshow', handleCancelOauthFlow);
```
HuiSF, line 17: > The pageshow event is sent to a [Window](https://developer.mozilla.org/en-US/docs/Web/API/Window) when the browser displays the window's document due to navigation.
> 
> This includes:
> * Initially loading the page
> * Navigating to the page from another page in the same window or tab
> * Restoring a frozen page on mobile OSes
> * Returning to the page using the browser's forward or back buttons

This seemed too broad to me than what we are looking for (the last case).
Any possibility to use `popstate` event which is more closely tight to the navigation?

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.navite.ts
israx: upsi

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
```diff
@@ -0,0 +1,18 @@
...
+import { OAuthStore } from '../types';
+
+import { createOAuthError } from './createOAuthError';
+import { handleFailure,  } from './index';
+
+export const listenForOAuthFlowCancellation = (store: OAuthStore) => {
+	async function handleCancelOauthFlow(event: PageTransitionEvent) {
+		const isOAuthInFlight = await store.loadOAuthInFlight();
+		if (event.persisted && isOAuthInFlight) {  
+			await handleFailure(createOAuthError('User cancelled OAuth flow.'));
```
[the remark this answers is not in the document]
israx: Yeah this is a good point. However we are currently surfacing the same `signInWithRedirect_failure` hub event when bfcache is not enabled. I just wanted to keep consistency

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
israx: yeah that is a fair claim. If the app is not reopened as Chris stated above, then it will be a new window, which will reset the window context, hence removing any listeners. I can double check and also trying to remove the event listener regardless of any condition.

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
israx: I see the `popstate` will be fired only if the history state has changed, also it doesn't provide a way to identify whether the page was initially being loaded or not, whereas the `pageshow` does. Although `pageshow` has more triggers, this event will most likely be triggered when the user has clicked the back button.

PR #13284, opened 2024-04-23
fix(auth): listen for pageshow event listener

israx:
#### Description of changes

Currently when bfcache is enabled and the user clicks the back button, amplify is not able to detect the cancellation because Amplify.configure is never called. This change listens for the pageshow window event and clears any inflight oauth keys

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13274

smoke testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.native.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export const listenForOAuthFlowCancelation = () => {
```
cshfang, line 4:
```suggestion
export const listenForOAuthFlowCancellation = () => {
```

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { OAuthStore } from '../types';
+
+import { createOAuthError } from './createOAuthError';
+
+import { handleFailure } from './index';
```
cshfang, line 8: Nit: For consistency with above and generally preferring directly accessing sibling functions
```suggestion
import { handleFailure } from './handleFailure';
```

packages/auth/__tests__/providers/cognito/signInWithRedirect.test.ts
```diff
@@ -188,6 +189,23 @@ describe('signInWithRedirect', () => {
...
 		});
+
+		it('invokes handleFailure when user cancelled the oauth flow', async () => {
+			const error = new Error('OAuth flow was cancelled.')
+			const mockOpenAuthSessionResult = {
+				type: undefined,
+			};
+			mockCreateOAuthError.mockReturnValueOnce(error);
+			mockOpenAuthSession.mockResolvedValueOnce(mockOpenAuthSessionResult);
+			oAuthStore.loadOAuthInFlight = jest.fn().mockResolvedValueOnce(true);
```
HuiSF, line 200: nit: `oAuthStore.loadOAuthInFlight` is mocked already at line 65.

packages/auth/__tests__/providers/cognito/signInWithRedirect.test.ts
```diff
@@ -188,6 +189,23 @@ describe('signInWithRedirect', () => {
...
+			window.addEventListener = jest.fn((event: string, cb: any) => {
```
HuiSF, line 201: nit: can we restore the mock after test?

packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
```diff
@@ -110,6 +111,11 @@ const oauthSignIn = async ({
 	// TODO(v6): use URL object instead
 	const oAuthUrl = `https://${domain}/oauth2/authorize?${queryString}`;
 
+	// this will be effective only on the following scenarios:
```
cshfang, line 114: Nit: Slight wording change to avoid confusion over "effectiveness"
```suggestion
	// this will only take effect in the following scenarios:
```

packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
```diff
@@ -110,6 +111,11 @@ const oauthSignIn = async ({
...
+	// 1. the user cancels the OAuth flow on web, and
```
cshfang, line 115: Nit: Should this be narrowed down to the back button? E.g.
```suggestion
	// 1. the user cancels the OAuth flow on web via back button, and
```

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
```diff
@@ -0,0 +1,20 @@
...
+export const listenForOAuthFlowCancellation = (store: OAuthStore) => {
+	async function handleCancelOAuthFlow(event: PageTransitionEvent) {
+		const isBackEventWithBfcache = event.persisted;
+		if (isBackEventWithBfcache && (await store.loadOAuthInFlight())) {
+			const error = createOAuthError('User cancelled OAuth flow.');
+			await handleFailure(error);
+		}
+		window.removeEventListener('pageshow', handleCancelOAuthFlow);
+	}
+	window.addEventListener('pageshow', handleCancelOAuthFlow);
```
HuiSF, line 19: We are assuming this listener to be added in browser env... so we may want to guard `window.addEventListener` to be accessible to avoid any unexpectedly unhandled errors.

israx: Not sure if a guard is needed here. there are currently two files `cancelOAuthFlow` and `cancelOAuthFlow.native` that separate web and rn

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { OAuthStore } from '../types';
+
+import { createOAuthError } from './createOAuthError';
+
+import { handleFailure } from './index';
+
+export const listenForOAuthFlowCancellation = (store: OAuthStore) => {
+	async function handleCancelOAuthFlow(event: PageTransitionEvent) {
+		const isBackEventWithBfcache = event.persisted;
```
cshfang, line 12: Nit: Is this technically misleading naming? I the `persisted` property could in theory indicate forward event, too, right? I'm only nitpicking at this because, in the case we ever need to debug this, if the naming implies something too specific here it could obfuscate some behavior.
```suggestion
		const isBackEventWithBfcache = event.persisted;
```

packages/auth/__tests__/providers/cognito/signInWithRedirect.test.ts
```diff
@@ -188,6 +189,23 @@ describe('signInWithRedirect', () => {
 				);
 			});
 		});
+
+		it('invokes handleFailure when user cancelled the oauth flow', async () => {
```
cshfang, line 193: Super nit
```suggestion
		it('invokes handleFailure when user cancels the oauth flow', async () => {
```

packages/auth/__tests__/providers/cognito/signInWithRedirect.test.ts
israx: It is mocked but not returning `true` by default.

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.native.ts
israx: good catch

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
israx: I already tried this import but it makes the unit test to fail.  It is not mocking it correctly

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
israx: seems the suggestion is the same. Changed to `isBfcache`

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
calebpollman: Can you update the mock import?

israx: done

cshfang (approved):

calebpollman (commented):
After thinking about the issue more, not sure that i feel comfortable adding an additional async side effect to `signInWithRedirect`. The proposed solution makes sense on its own, but concerned that when used in conjunction with `attemptCompleteOAuthFlow` that we are increasing the potential for exposing race conditions

packages/auth/src/providers/cognito/utils/oauth/cancelOAuthFlow.ts
```diff
@@ -0,0 +1,20 @@
...
+		if (isBackEventWithBfcache && (await store.loadOAuthInFlight())) {
+			const error = createOAuthError('User cancelled OAuth flow.');
+			await handleFailure(error);
```
calebpollman, line 15: [Adding for context](https://github.com/aws-amplify/amplify-js/pull/13278#discussion_r1575566948)

Where are we currently handling cancelation events as failures? See that `handleFailure` is called in the catch of `attemptCompleteOAuthFlow`, which makes sense. Also see the call to `handleFailure` in the `catch` block of `signinWithRedirect`, is that what you are referring to?

israx: yeah so basically [this](https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/providers/cognito/utils/oauth/completeOAuthFlow.ts#L89) is where we are handling the cancellation flow. This error will be caught by the catch of `attemptCompleteOAuthFlow`

AllanZhengYP (approved):

calebpollman (approved):

packages/auth/__tests__/providers/cognito/signInWithRedirect.test.ts
HuiSF: `(oAuthStore.loadOAuthInFlight as jest.mock).mockResolvedValueOnce(true);` should work?

HuiSF (approved):

haverchuck (approved):

elorzafe (approved):

PR #13288, opened 2024-04-24
fix(auth): Impossibility to sign-in when RandomKeyPassword missing

israx:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/auth/__tests__/providers/cognito/signInWithSRP.test.ts
+++ b/packages/auth/__tests__/providers/cognito/signInWithSRP.test.ts
@@ -14,8 +14,18 @@ import {
 import { AuthError } from '../../../src';
 import { createKeysForAuthStorage } from '../../../src/providers/cognito/tokenProvider/TokenStore';
 import * as clients from '../../../src/providers/cognito/utils/clients/CognitoIdentityProvider';
-
 jest.mock('../../../src/providers/cognito/utils/dispatchSignedInHubEvent');
+
+jest.mock('../../../src/providers/cognito/utils/srp', () => {
+	return {
+		...jest.requireActual('../../../src/providers/cognito/utils/srp'),
+		getAuthenticationHelper: jest.fn(() => ({
+			A: { toString: jest.fn() },
+			getPasswordAuthenticationKey: jest.fn(),
+		})),
+		getSignatureString: jest.fn(),
+	}
+})
 jest.mock('@aws-amplify/core/internals/utils', () => ({
 	...jest.requireActual('@aws-amplify/core/internals/utils'),
 	isBrowser: jest.fn(() => false),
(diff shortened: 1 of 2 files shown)


packages/auth/__tests__/providers/cognito/signInWithSRP.test.ts
```diff
@@ -56,13 +66,27 @@ function setDeviceKeys() {
 		mockedDeviceMetadata.randomPasswordKey,
 	);
 }
+type DeviceKey = 'deviceKey' | 'deviceGroupKey' | 'randomPasswordKey';
```
ashika112, line 69: nit: could we do something like ? this way we just define the keys in one place. Bascially opposite of what you are doing

```
const deviceKeys = ['deviceKey', 'deviceGroupKey', 'randomPasswordKey'] as const;

type DeviceKey = typeof deviceKeys[number]
```

packages/auth/__tests__/providers/cognito/signInWithSRP.test.ts
```diff
@@ -163,6 +187,58 @@ describe('signIn API happy path cases', () => {
...
+		test.each(deviceKeys)('respondToAuthChallenge should not include device key in the request if any device key in storage is deleted', async deviceKey => {
+           deleteDeviceKey(deviceKey);
+			await signIn({
+				username: lastAuthUser,
+				password: 'XXXXXXXX',
+			});
+
+			expect(respondToAuthChallengeAuthSpy).toHaveBeenCalledTimes(1);
+			const deviceKeyFromRequest = respondToAuthChallengeAuthSpy.mock.calls[0][1].ChallengeResponses?.DEVICE_KEY
+			expect(deviceKeyFromRequest).toBe(undefined)
```
ashika112, line 237: curious: wouldn't it be undefined pre this change as well? since u r coalescing to undefined? im asssuming the test is adding just the case now

israx: I added a test above that doesn't delete the device keys, and checks for an actual value.

israx: updated the description to specify the tests that were added

packages/auth/__tests__/providers/cognito/signInWithSRP.test.ts
israx: I'd prefer to keep the explicit type. There are only a few instances in the library where we use the `typeof` operator.

AllanZhengYP (approved):
Looks good to me. Thank you @israx

packages/auth/__tests__/providers/cognito/signInWithSRP.test.ts
ashika112: I guess im wondering more on the line test case for return is null like when all three are removed.

ashika112 (approved):

packages/auth/__tests__/providers/cognito/signInWithSRP.test.ts
israx: ooh I think what you mean. So if any key or all of them are missing then the underlying method to get keys will always return null, as shown bellow. This will prevent the library to add any device keys to the request. Hence the request will return undefined if a parameter is not found.

```
const deviceKeys = await tokenOrchestrator.getDeviceKeys(); // keys || null

if(deviceKeys){
 // add keys to request
}

```

PR #13299, opened 2024-04-26
feat(api-graphql): alias authMode identityPool -> iam

iartemiev:
#### Description of changes

* support `authMode: 'identityPool'`, which is an alias for `'iam'`

* added unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/core/src/singleton/API/types.ts
```diff
@@ -82,6 +82,7 @@ export type GraphQLAuthMode =
 	| 'oidc'
 	| 'userPool'
 	| 'iam'
```
HuiSF, line 84: Should we start marking `iam` as `@deprecated` so we can remove it from the next major version?

svidgen (approved):

packages/core/src/singleton/API/types.ts
iartemiev: I tried that, but it doesn't actually surface over union types AFAICT

iartemiev: Maybe still worth it as a reminder to ourselves. I'll update

iartemiev: Screenshot 2024-04-26 at 2 14 30 PM

svidgen (approved):

HuiSF (approved):

PR #13315, opened 2024-04-30
fix(core): Always set login methods when parsing Gen2 config files

jimblanc:
#### Description of changes

This change addresses a discrepancy when parsing Gen2 configuration files relative to Gen1. With this change, supported login attributes will always be set when parsing Gen2 configuration files when `username_attributes` is present in the config file. Before this change, login attributes may be `undefined` if that particular attribute was not present in the configuration file.

Local testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


ashika112 (approved):

israx (dismissed):

calebpollman (commented):
Nice! We caught this issue by diffing the result of `getConfig` when provided an `aws-exports` config for feature X (e.g. set login method to email and require phone number as a sign up attribute) against an `outputs` config configured for feature X. The result of `getConfig` is the single source of truth for configuration whether it's provided an `aws-exports` or `outputs` config, to ensure backwards compatibility we need to add unit testing for `parseAmplifyConfig` that assert the same return result for a feature when provided both shapes of configuration

packages/core/__mocks__/configMocks/amplify_outputs.json
jimblanc, line 1: Moved from `__tests__` and replaced all of the lorem ipsum stuff with more meaningful mock values.

packages/core/__mocks__/configMocks/README.md
jimblanc, line 1: Unit test would catch any updates not made to both files.

packages/core/src/parseAmplifyOutputs.ts
```diff
@@ -135,17 +135,12 @@ function parseAuth(
 		};
 	}
 
-	if (username_attributes?.includes('email')) {
+	if (username_attributes) {
```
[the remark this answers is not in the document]
jimblanc: That could happen as far as the schema, and in that case all three login attributes would be false (I don't know if the back-end would allow this though in practice). There was that discussion if this case could be considered "username" sign-in but we decided to wait for back-end support for that.

Good call on the test, went through and added a more complete comparison test using comparable fully specified configuration files.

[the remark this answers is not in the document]
ashika112: I think backend will always send [username_attributes] since login by email is default for auth.

packages/core/__tests__/utils/parseAmplifyConfigCompatibility.test.ts
```diff
@@ -0,0 +1,12 @@
+import { parseAmplifyConfig } from '../../src/libraryUtils';
+import mockAmplifyOutputs from '../../__mocks__/configMocks/amplify_outputs.json';
+import mockAmplifyConfiguration from '../../__mocks__/configMocks/amplifyconfiguration.json';
+
+describe('parseAmplifyConfig backwards compatibility', () => {
+	it('returns identical ResourcesConfig for equivalent CLI & Gen2 configuration files', () => {
+		const cliConfig = parseAmplifyConfig(mockAmplifyConfiguration);
+		const gen2Config = parseAmplifyConfig(mockAmplifyOutputs);
+
+		expect(gen2Config).toEqual(cliConfig);
```
ashika112, line 10: Nice 💯

packages/core/__mocks__/configMocks/amplifyconfiguration.json
```diff
@@ -0,0 +1,69 @@
+{
```
ashika112, line 1: We came across [this in cli ](https://github.com/aws-amplify/amplify-cli/blob/769211a99dd703f49e607b6e5da619204b920a4f/packages/amplify-frontend-javascript/lib/frontend-config-creator.js#L20)while working on geo issue to find possible config from backend: Maybe worth manual cross_check . Seems like all of it covered

ashika112 (approved):

israx (approved):

calebpollman (commented):
Thanks for adding the compatibility unit test, do think more granular testing of specific configurations should be added in the future but out of scope for this PR

israx (approved):

calebpollman (approved):

ashika112 (approved):

Issue #13272, opened 2024-04-21
Can't update optOut for AWS Pinpoint endpoint without address (APNs device token)


victoravr:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

Push Notifications

### Amplify Version

v6

### Amplify Categories

notifications

### Backend

CDK

### Environment information

<details>

```
  System:
    OS: macOS 14.4.1
    CPU: (8) arm64 Apple M2
    Memory: 58.22 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.7.3 - /opt/homebrew/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.5.0 - /opt/homebrew/bin/npm
    Watchman: 2024.04.15.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 124.0.6367.62
    Edge: 124.0.2478.51
    Safari: 17.4.1
  npmPackages:
    @aws-amplify/react-native: ^1.0.28 => 1.0.28 
    @aws-amplify/rtn-push-notification: ^1.2.28 => 1.2.28 
    @aws-amplify/ui-react-native: ^2.1.5 => 2.1.5 
    @aws-crypto/sha256-js: ^5.2.0 => 5.2.0 (3.0.0)
    @babel/core: ^7.23.6 => 7.24.0 
    @babel/preset-env: ^7.23.6 => 7.24.0 
    @babel/preset-react: ^7.23.3 => 7.23.3 
    @babel/runtime: ^7.23.6 => 7.24.0 
    @react-native-async-storage/async-storage: ^1.23.1 => 1.23.1 
    @react-native-community/netinfo: ^11.2.1 => 11.3.1 
    @react-native-community/slider: ^4.5.0 => 4.5.0 
    @react-native-picker/picker: ^2.6.1 => 2.6.1 
    @react-native/babel-preset: 0.73.21 => 0.73.21 
[...]

PR #13319, opened 2024-04-30
fix: Prevent creating multiple endpoint ids

cshfang:
#### Description of changes

Currently, when an endpoint is created by `updateEndpoint`, it is not cached until a full network round trip to the service. This results in a large window during which additional `updateEndpoint` calls will create new endpoint ids due to none being found in the cache.

This PR adds a utility which will hold created endpoint ids in memory until it is successfully cached or if the endpoint updating has failed.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13272

* `yarn test`
* Added thrown error expectation to original code and asserted the resulting behavior of the `updateEndpoint` function is the same (async void which throws uncaught)
* Tested in sample app and validated that network race condition was mitigated - the `updateEndpoint` calls triggered by `initializePushNotifications` and an adjacent `identifyUser` call were no longer creating separate endpoints.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/__tests__/providers/pinpoint/utils/createEndpointId.test.ts
+++ b/packages/core/__tests__/providers/pinpoint/utils/createEndpointId.test.ts
@@ -42,11 +42,13 @@ describe('Pinpoint Provider Util: createEndpointId', () => {
 		expect(mockAmplifyUuid).toHaveBeenCalled();
 	});
 
-	it('clears a created endpoint id for a category', () => {
-		const newUuid = `new-${uuid}`;
-		mockAmplifyUuid.mockReturnValue(newUuid);
-		clearCreatedEndpointId(appId, category);
+	describe('clearCreatedEndpointId()', () => {
+		it('can create a new endpoint id for a category after clearing', () => {
+			const newUuid = `new-${uuid}`;
+			mockAmplifyUuid.mockReturnValue(newUuid);
+			clearCreatedEndpointId(appId, category);
 
-		expect(createEndpointId(appId, category)).toBe(newUuid);
+			expect(createEndpointId(appId, category)).toBe(newUuid);
+		});
 	});
 });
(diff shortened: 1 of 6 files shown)


packages/core/__tests__/providers/pinpoint/utils/createEndpointId.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+	it('returns a new endpoint id for a category', () => {
+		mockAmplifyUuid.mockReturnValue(uuid);
+
+		expect(createEndpointId(appId, category)).toBe(uuid);
+	});
+
+	it('returns the same endpoint id for a category', () => {
+		const newUuid = `new-${uuid}`;
+		mockAmplifyUuid.mockReturnValue(newUuid);
```
HuiSF, line 25: nit: I think `expect(mockAmplifyUuid).not.toHaveBeenCalled()` is more valuable for this particular scenario

HuiSF (dismissed):

packages/core/__tests__/providers/pinpoint/utils/createEndpointId.test.ts
cshfang: Good call

packages/core/__tests__/providers/pinpoint/utils/createEndpointId.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+	it('returns a new endpoint id for a different category', () => {
+		const newUuid = `new-${uuid}`;
+		const newCategory = 'PushNotification';
+		mockAmplifyUuid.mockReturnValue(newUuid);
+
+		expect(createEndpointId(appId, newCategory)).toBe(newUuid);
+	});
+
+	it('clears a created endpoint id for a category', () => {
```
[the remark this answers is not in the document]
cshfang: The goal of the test is to check that it cleared. There's just not other obvious way of verifying it.

ashika112 (dismissed):

ashika112 (approved):

HuiSF (approved):

Issue #13174, opened 2024-03-25
Pinpoint endpoint user ID keeps getting set to identityId


nigelAt8Seats:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

Analytics, Push Notifications

### Amplify Version

v6

### Amplify Categories

analytics, notifications

### Backend

Amplify CLI

### Environment information

<details>

```
# Put output below this line
npx envinfo --system --binaries --browsers --npmPackages --duplicates --npmGlobalPackages

  System:
    OS: Linux 6.2 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz
    Memory: 12.59 GB / 31.18 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 18.18.2 - ~/.asdf/installs/nodejs/18.18.2/bin/node
    Yarn: 1.22.21 - ~/code/8/horizons/node_modules/.bin/yarn
    npm: 9.8.1 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 122.0.6261.128
  npmPackages:
    @aws-amplify/react-native: ^1.0.5 => 1.0.13
    @aws-amplify/rtn-push-notification: ^1.2.5 => 1.2.13
    @babel/core: ^7.20.0 => 7.23.9
    @babel/preset-env: ^7.20.0 => 7.23.9
    @babel/runtime: ^7.20.0 => 7.23.9
    @gorhom/bottom-sheet: ^4.4.7 => 4.6.0
    @hookform/resolvers: ^3.1.1 => 3.3.4
    @hookform/resolvers/ajv:  1.0.0
    @hookform/resolvers/arktype:  1.0.0
    @hookform/resolvers/class-validator:  1.0.0
    @hookform/resolvers/computed-types:  1.0.0
[...]

PR #13330, opened 2024-05-02
fix: Prevent overwriting with fallbacks when updating endpoint

cshfang:
#### Description of changes

When calling `initializePushNotifications`, the library calls an internal `updateEndpoint` API. Anytime `updateEndpoint` is called, it automatically populates the `demographic` property with client info and the `userId` with the identity id. When calling `identifyUser`, these values can be provided explicitly but will always be pre-populated. This can have unexpected consequences since `initializePushNotifications` is typically called on every app launch - and it will opaquely overwrite existing values previously set on the endpoint.

This PR updates the `updateEndpoint` API to only pre-populate these fields when **creating** an endpoint to avoid repeatedly overwriting the endpoint with the default/fallback values.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13174

* `yarn test`
* Added unit tests that were expected to fail prior to code changes and verified they are now passing
* Tested on sample app to see that `demographic` and `userId` are only on the network request when an endpoint has yet to be cached (endpoint creation) and subsequent calls do not.

```
// Endpoint creation
{
  "RequestId": "8808844f-5e12-440b-a914-d3ae7191b0f8",
  "EffectiveDate": "2024-05-02T19:55:22.795Z",
  "ChannelType": "APNS_SANDBOX",
  "Address": "<apns-token>",
  "Demographic": {
    "AppVersion": "ios/17.0",
    "Make": null,
    "Model": null,
    "ModelVersion": "17.0",
    "Platform": "ios"
  },
  "User": { "UserId": "<cognito-identity-id>" }
}
```

```
// Endpoint update
{
  "RequestId": "8178652f-5842-44b6-a02c-27616f4d6cb9",
  "EffectiveDate": "2024-05-02T19:57:28.751Z",
  "ChannelType": "APNS_SANDBOX",
  "Address": "<apns-token>",
}

```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/src/providers/pinpoint/apis/updateEndpoint.ts
+++ b/packages/core/src/providers/pinpoint/apis/updateEndpoint.ts
@@ -1,12 +1,13 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { amplifyUuid } from '../../../utils/amplifyUuid';
-import { getClientInfo } from '../../../utils/getClientInfo';
 import {
 	UpdateEndpointInput,
 	updateEndpoint as clientUpdateEndpoint,
 } from '../../../awsClients/pinpoint';
+import { UserProfile } from '../../../types';
+import { amplifyUuid } from '../../../utils/amplifyUuid';
+import { getClientInfo } from '../../../utils/getClientInfo';
 import { PinpointUpdateEndpointInput } from '../types';
 import { cacheEndpointId } from '../utils/cacheEndpointId';
 import {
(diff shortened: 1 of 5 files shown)


packages/core/src/providers/pinpoint/apis/updateEndpoint.ts
```diff
@@ -46,22 +47,34 @@ export const updateEndpoint = async ({
 		name,
 		plan,
 	} = userProfile ?? {};
-	const clientInfo = getClientInfo();
+
+	// only automatically populate the endpoint with client info and identity id upon endpoint creation to
+	// avoid overwriting the endpoint with these values every time the endpoint is updated
+	const demographicsFromClientInfo: UserProfile['demographic'] = {};
+	const resolvedUserId = createdEndpointId ? userId ?? identityId : userId;
```
erinleigh90, line 54: Nit: Did we decide that `||` is better to use than `??`
```suggestion
	const resolvedUserId = createdEndpointId ? userId || identityId : userId;
```

erinleigh90: To make sure I understand: the `identityId` property is always a default value and `userId` is always customer provided or empty, correct?

erinleigh90 (approved):
Thanks, Chris! Good catch.

Had one nit and one question but LGTM!

packages/core/src/providers/pinpoint/apis/updateEndpoint.ts
cshfang: If this call creates an endpoint
* use `userId` if provided, otherwise fall back to `identityId`

If this call updates an endpoint
* use `userId` if provided but don't fall back to anything

packages/core/src/providers/pinpoint/apis/updateEndpoint.ts
```diff
@@ -70,31 +83,37 @@ export const updateEndpoint = async ({
...
+					Model: mergedDemographic.model,
+					ModelVersion: mergedDemographic.modelVersion,
+					Platform: mergedDemographic.platform,
+					PlatformVersion: mergedDemographic.platformVersion,
+					Timezone: mergedDemographic.timezone,
+				},
+			}),
+			...(location && {
+				Location: {
+					City: location.city,
```
Samaritan1011001, line 101: nit: no need for null check on location like before?

Samaritan1011001 (approved):

packages/core/src/providers/pinpoint/apis/updateEndpoint.ts
cshfang: Line 99 should short circuit this

jimblanc (approved):

PR #13338, opened 2024-05-03
fix(v3): updates dependencies and unit tests

israx:
Please make sure to read the Pull Request Guidelines:
https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#pull-requests
-->
#### System Requirements
- Node 16.14.0

#### Description of changes

 - ignores deprecated UI packages from lerna to allow building the library
 - fixes `auth`, `predictions` `graph-ql` and `datastore` unit tests

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/api-graphql/__tests__/GraphQLAPI-test.ts
```diff
@@ -757,7 +757,7 @@ describe('API test', () => {
 					variables,
 					authMode: GRAPHQL_AUTH_MODE.OPENID_CONNECT,
 				})
-			).rejects.toThrowError('No federated jwt');
+			).rejects.toThrowError('No current user');
```
jimblanc, line 760: Sanity check: What did this change? Shouldn't the behavior be the same?

packages/auth/src/Auth.ts
```diff
@@ -1386,7 +1386,9 @@ export class AuthClass {
 		logger.debug('Getting current session');
 		// Purposely not calling the reject method here because we don't need a console error
 		if (!this.userPool) {
-			return Promise.reject();
```
jimblanc, line 1389: Why did we need to change this?

jimblanc (commented):
Looks pretty good, thank you! Holding off on approving until we confirm the strategy with the UI packages.

One other validation idea we might want to consider: Would it be possible to diff the artifacts produced by this build with the artifacts available on the latest version of v3 on NPM? This would help us ensure behavior remains the same in lieu of complete E2E tests.

packages/auth/src/Auth.ts
israx: most APIs, if not all, will construct the same error when a user pool is not configured, and the test is currently asserting for that specific error.

packages/api-graphql/__tests__/GraphQLAPI-test.ts
israx: Seems the underlying error message changed from `No federated jwt` to `No current user`.

.yarnrc
```diff
@@ -1,2 +1,2 @@
---add.no-lockfile true
---install.no-lockfile true
\ No newline at end of file
+--add.no-lockfile false
```
ashika112, line 1: I think it is safe to remove this file if we dont need the override. The default behavior should be to add lock-file

packages/auth/__tests__/auth-unit-test.ts
```diff
@@ -3257,7 +3257,7 @@ describe('auth unit test', () => {
 				Username: 'username',
 				Pool: userPool,
 			});
-			const errorMessage = new NoUserPoolError(AuthErrorTypes.EmptyCode);
+			const errorMessage = new NoUserPoolError(AuthErrorTypes.MissingAuthConfig);
```
ashika112, line 3260: why this change?

package.json
```diff
@@ -36,7 +36,22 @@
 	},
 	"workspaces": {
 		"packages": [
-			"packages/*"
+			"packages/amazon-cognito-identity-js",
```
ashika112, line 39: Why are we making it explicit now?

ashika112: NVM got it . UI ones

package.json
```diff
@@ -91,14 +106,20 @@
 		"webpack": "^4.32.0",
 		"webpack-bundle-analyzer": "^3.3.2",
 		"webpack-cli": "^3.1.0",
-		"winston": "^3.2.1"
+		"winston": "^3.2.1",
+		"@types/lodash": "4.14.182"
```
ashika112, line 110: can u also explain need for the extra package here?

packages/auth/__tests__/auth-unit-test.ts
israx: The test case is checking for an incorrect auth configuration. This test was using a wrong `AuthErrorTypes`

package.json
israx: It is to help the library to build. V4 currently added this dev dep

packages/auth/src/Auth.ts
israx: Also the test is checking for no userPool error as well

jimblanc: Makes sense, I'm just wondering why the error behavior changed?

jimblanc (commented):
Looking good, just had a few more questions about changes to `Auth.ts`, ideally we'd be able to get by here without changing any of the implementation. Also, it seems like this branch kicked off some old workflows which are failing. Do we need to disable those? It looks like they run any time someone adds a comment to this PR 😅

packages/auth/src/Auth.ts
israx: yeah I don't have an answer respect to that, there is inconsistency between the tests and the code. 

We can either:
1. change/delete the test to match the code.
2. change the code to match the test case.

I would go for number 2 as it is not breaking and is consistent how the other APIs handle the same error when there is no userpool.

jimblanc (approved):
Discussed concerns async, LGTM. If you haven't yet could you please also make sure the branch can be published using Verdaccio and try installing it in a sample app?

PR #13353, opened 2024-05-09
fix: Wait for endpoint creation to identify user

cshfang:
#### Description of changes

When the Push Notification category is first creating a Pinpoint endpoint upon receiving a device token, it is possible for an `identifyUser` call to be attempting to update the endpoint to be resolved first over the network. This is troublesome for at least two known reasons:
1. The `identifyUser` call could be attempting to set the `OptOut` property to `'NONE'` and if it is resolved first by the service, the service will reject the request due to the `Address` not having yet been set.
2. The `identifyUser` call could contain `userId` or `demographic` information and if it resolved first by the service, the request made by the library to register the device token would contain default values which would unintentionally overwrite the ones set by `identifyUser`

This PR mitigates this by having `identifyUser` calls (for Push Notification channels) await the resolution of device registration during endpoint creation (i.e. this blocking does not occur during endpoint updates).

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13174
https://github.com/aws-amplify/amplify-js/issues/13272

* Added unit tests for added use cases
* `yarn test`
* Tested in a sample React Native application to assert `identifyUser` calls were blocked on the endpoint creation triggered by device registration but not on subsequent updates

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/notifications/__tests__/pushNotifications/providers/pinpoint/utils/inflightDeviceRegistration.test.ts
+++ b/packages/notifications/__tests__/pushNotifications/providers/pinpoint/utils/inflightDeviceRegistration.test.ts
@@ -1,0 +1,69 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+	getInflightDeviceRegistration,
+	rejectInflightDeviceRegistration,
+	resolveInflightDeviceRegistration,
+} from '../../../../../src/pushNotifications/providers/pinpoint/utils/inflightDeviceRegistration';
+import { InflightDeviceRegistration } from '../../../../../src/pushNotifications/providers/pinpoint/types';
+
+describe('inflightDeviceRegistration', () => {
+	describe('resolveInflightDeviceRegistration', () => {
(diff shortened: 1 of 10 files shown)


packages/notifications/__tests__/pushNotifications/providers/pinpoint/utils/inflightDeviceRegistration.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+		it('should reject the promise', async () => {
+			const blockedFunction = jest.fn();
+			const promise = getInflightDeviceRegistration()?.then(() => {
+				blockedFunction();
+			});
+
+			expect(blockedFunction).not.toHaveBeenCalled();
+			rejectInflightDeviceRegistration(new Error());
+			await expect(promise).rejects.toThrow('Failed to register device');
```
HuiSF, line 65: nit: worth to also assert the thrown error contains the `underlyError` as the `new Error()`?

HuiSF (dismissed):

HuiSF (approved):

israx (approved):

PR #13383, opened 2024-05-15
fix(v2): update dependencies and code 

israx:
Enable publish to npm tag stable-2.0 from branch 2.0-stable

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/api/src/API.ts
```diff
@@ -315,6 +315,15 @@ export default class APIClass {
...
 			case 'AMAZON_COGNITO_USER_POOLS':
+				
+				try {
+					const session = await Auth.currentSession();
+					headers = {
+						Authorization: session.getAccessToken().getJwtToken(),
+					};
+				} catch (e) {
+					throw new Error(GraphQLAuthError.NO_CURRENT_USER);
+				}
```
israx, line 326: The current issue is that `API.graphql` will throw an unhandled promise when there is no user logged in and `authenticationType` is `AMAZON_COGNITO_USER_POOLS` . This is making a unit test to fail. 

The solution was to implement the same [v3 ](https://github.com/aws-amplify/amplify-js/blob/version-3/packages/api-graphql/src/GraphQLAPI.ts#L164) approach.

packages/api/src/API.ts
```diff
@@ -315,6 +315,15 @@ export default class APIClass {
 				};
 				break;
 			case 'AMAZON_COGNITO_USER_POOLS':
+				
+				try {
+					const session = await Auth.currentSession();
```
jimblanc, line 320: Can we consolidate this with the logic on lines 327-330?

jimblanc (commented):
Looks good, one nit

packages/api/src/API.ts
```diff
@@ -336,10 +345,9 @@ export default class APIClass {
 	 */
 	getGraphqlOperationType(operation) {
 		const doc = parse(operation);
-		const {
-			definitions: [{ operation: operationType }],
-		} = doc;
-
+		const definitions =
+			doc.definitions as ReadonlyArray<OperationDefinitionNode>;
+		const [{ operation: operationType }] = definitions;
```
cshfang, line 350: Nit: Suggest making this more readable i.e.
```ts
		const [ definition ] =
			doc.definitions as ReadonlyArray<OperationDefinitionNode>;
		return definition.operation
```

jimblanc (approved):

packages/api/src/API.ts
israx: The change only adds a type but doesn't make any change to the code. I'll leave it as is for now

jimblanc (approved):

PR #13387, opened 2024-05-15
fix(ssr-adapter): refreshToken may return the same tokens across requests

HuiSF:
#### Description of changes

1. Exported a non-deduped token refresher function 
2. Updated the SSR adapter to use the non-deduped token refresher function to create context
3. Fixed that in a `getServerSideProps` context, refresh tokens were not being sent back to client cookie store

1. unit tests
2. manual testing with Next.js sample apps

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/auth/src/providers/cognito/utils/refreshAuthTokens.ts
+++ b/packages/auth/src/providers/cognito/utils/refreshAuthTokens.ts
@@ -77,3 +77,4 @@ const refreshAuthTokensFunction: TokenRefresher = async ({
 };
 
 export const refreshAuthTokens = deDupeAsyncFunction(refreshAuthTokensFunction);
+export const refreshAuthTokensWithoutDedupe = refreshAuthTokensFunction;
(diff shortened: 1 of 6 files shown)


packages/auth/src/providers/cognito/utils/refreshAuthTokens.ts
```diff
@@ -77,3 +77,4 @@ const refreshAuthTokensFunction: TokenRefresher = async ({
 };
 
 export const refreshAuthTokens = deDupeAsyncFunction(refreshAuthTokensFunction);
+export const refreshAuthTokensWithoutDedupe = refreshAuthTokensFunction;
```
cshfang, line 80: Nit: Just rename `refreshAuthTokensFunction` -> `refreshAuthTokensWithoutDedupe` and export it

jimblanc (commented):
Change looks good, need to target `release` with your PR though.

packages/auth/src/providers/cognito/utils/refreshAuthTokens.ts
HuiSF: Will do this later, as I'm thinking of a different solution to create dedupe edition for SSR use cases.

jimblanc (approved):

cshfang (approved):

PR #13430, opened 2024-05-24
chore(analytics): enable linting on __tests__

HuiSF:
#### Description of changes

See each commit message for details.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/analytics/__tests__/providers/personalize/apis/flushEvents.test.ts
+++ b/packages/analytics/__tests__/providers/personalize/apis/flushEvents.test.ts
@@ -46,7 +46,7 @@ describe('Analytics Personalize API: flushEvents', () => {
 		await new Promise(process.nextTick);
 		expect(mockResolveConfig).toHaveBeenCalledTimes(1);
 		expect(mockResolveCredentials).toHaveBeenCalledTimes(1);
-		const { trackingId, ...configWithoutTrackingId } = mockPersonalizeConfig;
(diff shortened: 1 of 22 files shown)


packages/analytics/__tests__/providers/personalize/apis/flushEvents.test.ts
```diff
@@ -46,7 +46,7 @@ describe('Analytics Personalize API: flushEvents', () => {
 		await new Promise(process.nextTick);
 		expect(mockResolveConfig).toHaveBeenCalledTimes(1);
 		expect(mockResolveCredentials).toHaveBeenCalledTimes(1);
-		const { trackingId, ...configWithoutTrackingId } = mockPersonalizeConfig;
+		const { trackingId: _, ...configWithoutTrackingId } = mockPersonalizeConfig;
```
ashika112, line 49: Do we even need to keep trackingId if we are not using it?

HuiSF: This expression was for removing the `trackingId` property while creating a new copy from the `mockPersonalizeConfig`, I didn't want to refactor it completely from the original code.

ashika112: i think this might be missed artifact right? We should either check if trackingId's value or remove it , IMO. but non blocking so i will approve this PR :)

ashika112 (approved):

cshfang (approved):

PR #13433, opened 2024-05-27
Md5 calculation blob data/storage

sushpatg:
#### Description of changes

Currently when we enable "isObjectLockEnabled" to true. Amplify Library auto-calculates MD5 header: [Code Reference](https://github.com/aws-amplify/amplify-js/blob/0b72b32348dc63bbc11596a062338ee9ddfa9177/packages/storage/src/providers/s3/utils/md5.ts#L8)

However, this calculation is only working when passing String data. When we select file using the following [documented code](https://docs.amplify.aws/gen1/javascript/build-a-backend/storage/upload/#browser-uploads) it throws BadDigest error.
```
import { uploadData } from 'aws-amplify/storage';

const uploadDataInBrowser = async (event) => {
  if (event?.target?.files) {
    const file = event.target.files[0];
    uploadData({
      path: file.name,
      data: file
    });
  }
};
```

On debugging the MD5 implementation, could observe that [reader.readAsDataURL(blob)](https://github.com/aws-amplify/amplify-js/blob/0b72b32348dc63bbc11596a062338ee9ddfa9177/packages/storage/src/providers/s3/utils/md5.ts#L48C3-L48C29) returns base64 string and we are encoding it with [utf-8](https://github.com/aws-amplify/amplify-js/blob/0b72b32348dc63bbc11596a062338ee9ddfa9177/packages/storage/src/providers/s3/utils/client/runtime/base64/index.browser.ts#L11) which is causing the error,

With the below PR, I'm converting the base64 string to binary string which is returned to Uint8Array
 containing the UTF-8 encoded bytes of the input string.

#### Issue #, if available

[13434](https://github.com/aws-amplify/amplify-js/issues/13434)

Have tested code for png, json, txt file extensions and also directly passing data uploadData API call

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


cshfang:
Hi @sushpatg, thank you for opening this PR. However, while digging into this issue to validate the fix, we found that the issue extends into the React Native implementation of this calculation as well.

Rather than asking you to also account for this and add unit tests as well, please take a look at this PR (https://github.com/aws-amplify/amplify-js/pull/13458) opened instead.

sushpatg:
> Hi @sushpatg, thank you for opening this PR. However, while digging into this issue to validate the fix, we found that the issue extends into the React Native implementation of this calculation as well.
> 
> Rather than asking you to also account for this and add unit tests as well, please take a look at this PR (#13458) opened instead. If the proposed changes would address the original issue, I would like to close this PR in favor of the more holistic one if that is alright with you. Thanks again for looking into this!

Thank you @cshfang, will follow up [13458](https://github.com/aws-amplify/amplify-js/pull/13458)

ashika112:
i think this issue can be closed in favor of merged (https://github.com/aws-amplify/amplify-js/pull/13458) Feel free to open up ticket if you find anything else.

Issue #13434, opened 2024-05-27
MD5 not calculating correctly when selecting file from broswer, failing with BadDigest Error


sushpatg:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Storage

### Amplify Version

v6

### Amplify Categories

storage

### Backend

Amplify CLI

### Environment information

<details>

```
System:
    OS: macOS 14.4.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 86.14 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.1 - /opt/homebrew/opt/node@20/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.2.4 - /opt/homebrew/opt/node@20/bin/npm
  Browsers:
    Chrome: 125.0.6422.112
    Safari: 17.4.1
  npmPackages:
    @aws-amplify/ui-react: ^6.1.12 => 6.1.12 
    @aws-amplify/ui-react-internal:  undefined ()
    @smithy/md5-js: ^3.0.0 => 3.0.0 
    @testing-library/jest-dom: ^5.17.0 => 5.17.0 
    @testing-library/react: ^13.4.0 => 13.4.0 
    @testing-library/user-event: ^13.5.0 => 13.5.0 
    aws-amplify: ^6.3.4 => 6.3.4 
    aws-amplify/adapter-core:  undefined ()
    aws-amplify/analytics:  undefined ()
    aws-amplify/analytics/kinesis:  undefined ()
    aws-amplify/analytics/kinesis-firehose:  undefined ()
    aws-amplify/analytics/personalize:  undefined ()
    aws-amplify/analytics/pinpoint:  undefined ()
    aws-amplify/api:  undefined ()
    aws-amplify/api/server:  undefined ()
    aws-amplify/auth:  undefined ()
    aws-amplify/auth/cognito:  undefined ()
[...]

PR #13458, opened 2024-06-02
fix(storage): Fix MD5 calculation

cshfang:
#### Description of changes

When `isObjectLockEnabled` is set to `true` in configuration, the Amplify library calculates and appends `ContentMD5` automatically. However, as a customer discovered, this MD5 calculation was not always being performed correctly and the upload task results in a BadDigest error from the service.

Upon investigation, it appears that the category went under a refactor which saw the MD5 calculations utilities be renamed and relocated. However, during this refactor, the implementations intended for web and native were accidentally flipped:
* MD5Utils -> utils/md5.native
* MD5Utils.native -> utils/md5

After undoing this switch, the library behaved as expected on Web but the native implementations continued to fail. Upon further deep dive, it was discovered that the `readAsArrayBuffer` gap which necessitated the diverged native path in the first place is only experienced in React Native version < 0.71.

This PR:
* Unflips the web implementation for MD5 calculation
* Wraps the attempt to use `readAsArrayBuffer` in a try/catch and only invokes the `readAsDataURL` strategy as a fallback if the prior function fails
* Adds unit tests to prevent future regressions

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13434

Validated that MD5 calculation was not working in sample apps prior to changes but now work with the proposed fix for:
* React web app
* React native 0.74 app (current)
* React native 0.71 app (latest version to not support `readAsArrayBuffer` in its FileReader)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/storage/__tests__/providers/s3/utils/md5.native.test.ts
+++ b/packages/storage/__tests__/providers/s3/utils/md5.native.test.ts
@@ -1,0 +1,131 @@
+import { Buffer } from 'buffer';
+
+import { Md5 } from '@smithy/md5-js';
+
+import { calculateContentMd5 } from '../../../../src/providers/s3/utils/md5.native';
+import { toBase64 } from '../../../../src/providers/s3/utils/client/utils';
+
+jest.mock('@smithy/md5-js');
+jest.mock('../../../../src/providers/s3/utils/client/utils');
+jest.mock('buffer');
+
+interface MockFileReader {
+	error?: any;
+	result?: any;
+	onload?(): void;
+	onabort?(): void;
+	onerror?(): void;
+	readAsArrayBuffer?(): void;
+	readAsDataURL?(): void;
+}
+
+// The FileReader in React Native 0.71 did not support `readAsArrayBuffer`. This native implementation accomodates this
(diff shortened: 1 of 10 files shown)


israx (approved):

packages/storage/__tests__/providers/s3/utils/md5.native.test.ts
```diff
@@ -0,0 +1,131 @@
...
+	onabort?(): void;
+	onerror?(): void;
+	readAsArrayBuffer?(): void;
+	readAsDataURL?(): void;
+}
+
+// The FileReader in React Native 0.71 did not support `readAsArrayBuffer`. This native implementation accomodates this
+// by attempting to use `readAsArrayBuffer` and changing the file reading strategy if it throws an error.
+// TODO: This file should be removable when we drop support for React Native 0.71
+describe('calculateContentMd5 (native)', () => {
```
jimblanc, line 25: Thank you for the thorough testing this PR ❤️

erinleigh90 (approved):
LGTM!

jimblanc (approved):
Thanks! Looks good

ashika112 (approved):

Issue #13466, opened 2024-06-04
configuring oidc breaks authenricator component


dataminion:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React, Angular

### Which UI component?

Authenticator

### How is your app built?

vit/webpack

### What browsers are you seeing the problem on?

Chrome, Safari

### Which region are you seeing the problem in?

us east 2

### Please describe your bug.

I started with first the angular then switched to the react quickstart template repositories in an attempt to trouble shoot this issue. I would set up the basics then when I added the oidc provider to the backend everything deploys and succeeds but the page fails to render. The console then throws the following error:

`Missing onError handler for invocation 'authenticator.setup.getConfig:invocation[0]', error was 'TypeError: undefined is not an object (evaluating 'provider.toString')'. Stacktrace was '@aws-amplify_ui-angular.js:12987:120'`

I have narrowed the issue down to the authenticator component as when it alone is removed from the html the page loads as expected. I want to be very clear the oidc provider is properly configured on both ends.

I have been able to use the functions from the documentation:

```typescript
await signInWithRedirect({
        provider: {
          custom: 'MicrosoftEntraID'
        }
});

const attr: any = await fetchUserAttributes();

await signOut();
```

to redirect to my provider then come back to the page, proving that the provider is properly configured. 
This puts me in the frustrating position of having to manually handle the auth state.

### What's the expected behaviour?

Honestly it was pretty clear from the documentation that oidc isn't fully implemented in the UI so I don't expect my ideal behavior for example cognito auto generates the following login landing page

<img width="749" alt="Screenshot 2024-06-04 at 4 38 57 PM" src="https://github.com/aws-amplify/amplify-ui/assets/13214966/67e5a741-d611-41f2-852a-72fc47bcd591">

but my basic expectation was that the authenticator component would function as normal and that I would be able to add a login with sso button using the footer slot

### Help us reproduce the bug!

[...]

PR #13469, opened 2024-06-05
release(required): Parsing custom oAuth in amplify_outputs

ashika112:
#### Description of changes
The `parseAmplifyOutputs` function does not take into account any custom oAuth Provider and this lead to providers in oAuth config having undefined values breaking the UI Authenticator component.

In additon to the fix for Amplify outputs, the PR loosens up the OAuth config type.  Since looking into the issue, realized our parse utility provided for adding custom oAuth as string to list of providers outside of our `CustomProvider` in `OAuthConfig`. 

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13466

- unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/src/parseAmplifyOutputs.ts
+++ b/packages/core/src/parseAmplifyOutputs.ts
@@ -315,8 +315,8 @@ const providerNames: Record<string, OAuthProvider> = {
 	SIGN_IN_WITH_APPLE: 'Apple',
 };
 
-function getOAuthProviders(providers: string[] = []): OAuthProvider[] {
-	return providers.map(provider => providerNames[provider]);
+function getOAuthProviders(providers: string[] = []): string[] {
+	return providers.map(provider => providerNames[provider] ?? provider);
 }
 
 function getMfaStatus(
--- a/packages/core/src/singleton/Auth/types.ts
+++ b/packages/core/src/singleton/Auth/types.ts
@@ -179,7 +179,7 @@ export interface OAuthConfig {
 	redirectSignIn: string[];
 	redirectSignOut: string[];
 	responseType: 'code' | 'token';
-	providers?: (OAuthProvider | CustomProvider)[];
+	providers?: (OAuthProvider | CustomProvider)[] | string[];
 }
 
 export type OAuthProvider = 'Google' | 'Facebook' | 'Amazon' | 'Apple';
--- a/packages/aws-amplify/package.json
+++ b/packages/aws-amplify/package.json
@@ -317,7 +317,7 @@
 			"name": "[Analytics] identifyUser (Pinpoint)",
 			"path": "./dist/esm/analytics/index.mjs",
 			"import": "{ identifyUser }",
-			"limit": "15.57 kB"
+			"limit": "15.58 kB"
 		},
 		{
 			"name": "[Analytics] enable",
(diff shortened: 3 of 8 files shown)


jimblanc (approved):
Change looks good to me pending additional testing.

packages/core/src/singleton/Auth/types.ts
```diff
@@ -179,7 +179,7 @@ export interface OAuthConfig {
 	redirectSignIn: string[];
 	redirectSignOut: string[];
 	responseType: 'code' | 'token';
-	providers?: (OAuthProvider | CustomProvider)[];
+	providers?: (OAuthProvider | CustomProvider)[] | string[];
```
reesscot, line 182: I thought the custom providers had to be in a CustomProvider object?

packages/core/src/parseAmplifyOutputs.ts
```diff
@@ -315,8 +315,8 @@ const providerNames: Record<string, OAuthProvider> = {
 	SIGN_IN_WITH_APPLE: 'Apple',
 };
 
-function getOAuthProviders(providers: string[] = []): OAuthProvider[] {
-	return providers.map(provider => providerNames[provider]);
+function getOAuthProviders(providers: string[] = []): string[] {
+	return providers.map(provider => providerNames[provider] ?? provider);
```
calebpollman, line 319: Concerned that this is going to break the `Authenticator`. Need to see a passing test run through our CI before  approval

calebpollman (changes requested):
Left a blocking comment

packages/core/src/parseAmplifyOutputs.ts
ashika112: how can we do that? @calebpollman

Also , the output of this and gen1 parser is the same. It wasn't before. I updated our mock comparison to check the same [here] (https://github.com/aws-amplify/amplify-js/pull/13469/files#diff-b729da5fc469a5a8411d893651231dee6138148eb593396542ae3a11346ef96eR25)

and [here](https://github.com/aws-amplify/amplify-js/pull/13469/files#diff-b729da5fc469a5a8411d893651231dee6138148eb593396542ae3a11346ef96eR25)

ashika112: and that gets compared in this test https://github.com/aws-amplify/amplify-js/blob/main/packages/core/__tests__/utils/parseAmplifyConfigCompatibility.test.ts#L10

packages/core/src/singleton/Auth/types.ts
ashika112: long story short, what we were doing in gen1 has an issue. This loosens the type to account for issue coming from [this typecasting ](https://github.com/aws-amplify/amplify-js/pull/13469/files#diff-fa4b039ec83dc5858b6f57efdb0d580f57133da49c31ab9b5836aa218358118eL328)

reesscot: Since we don't actually ever return a CustomProvider, shouldn't it be removed from the type?

israx (approved):

packages/core/src/singleton/Auth/types.ts
ashika112: This is type is publicly available through `Amplify.configure()` and `Amplify.getConfig()` any change to that would be breaking change

ashika112:
Closing in favor of https://github.com/aws-amplify/amplify-js/pull/13473

PR #13474, opened 2024-06-06
release(required): Parsing custom oAuth in amplify_outputs

ashika112:
#### Description of changes
The `parseAmplifyOutputs` function does not take into account any custom oAuth Provider and this lead to providers in oAuth config having undefined values breaking the UI Authenticator component.

In additon to the fix for Amplify outputs, the PR loosens up the OAuth config type.  Since looking into the issue, realized our parse utility provided for adding custom oAuth as string to list of providers outside of our `CustomProvider` in `OAuthConfig`. 

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13466

- unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/__mocks__/configMocks/amplifyconfiguration.json
+++ b/packages/core/__mocks__/configMocks/amplifyconfiguration.json
@@ -22,38 +22,32 @@
 		"responseType": "token"
 	},
 	"aws_cognito_verification_mechanisms": ["EMAIL"],
-	"aws_cognito_social_providers": ["FACEBOOK", "APPLE", "GOOGLE"],
+	"aws_cognito_social_providers": ["FACEBOOK", "APPLE", "GOOGLE", "Auth0"],
 	"aws_mobile_analytics_app_id": "mock-pinpoint-app-id",
 	"aws_mobile_analytics_app_region": "us-west-2",
 	"aws_user_files_s3_bucket": "mock-storage-bucket",
 	"aws_user_files_s3_bucket_region": "us-west-2",
 	"aws_user_pools_id": "mock-cup-id",
 	"aws_user_pools_web_client_id": "mock-cup-client-id",
-	"geo": { 
+	"geo": {
 		"amazon_location_service": {
 			"search_indices": {
-				"items": [
-					"mock-geo-search-item",
-					"mock-geo-search-item-alt"
-				],
+				"items": ["mock-geo-search-item", "mock-geo-search-item-alt"],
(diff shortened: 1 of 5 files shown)


packages/core/__mocks__/configMocks/amplifyconfiguration.json
ashika112, line 1: just lint changes on this file

jimblanc (dismissed):
LGTM

jimblanc (approved):

jimblanc (approved):

packages/core/src/parseAmplifyOutputs.ts
```diff
@@ -316,7 +316,13 @@ const providerNames: Record<string, OAuthProvider> = {
 };
 
 function getOAuthProviders(providers: string[] = []): OAuthProvider[] {
-	return providers.map(provider => providerNames[provider]);
+	return providers.reduce<OAuthProvider[]>((oAuthProviders, provider) => {
```
israx, line 319: is this approach correct ? wouldn't reduce ignore any other providers that are not included in the `providerNames` object ?

packages/core/src/parseAmplifyOutputs.ts
```diff
@@ -316,7 +316,13 @@ const providerNames: Record<string, OAuthProvider> = {
...
+		if (providerNames[provider] !== undefined) {
+			oAuthProviders.push(providerNames[provider]);
+		}
+
+		return oAuthProviders;
+	}, []);
```
svidgen, line 325: Essentially this?

```suggestion
	return providers.map(provider => providerNames[provider]).filter(p => p);
```

svidgen (approved):

packages/core/src/parseAmplifyOutputs.ts
ashika112: yes ! reduce reduces the two iteration :P

calebpollman, line 325: Can simplify this with `filter`:
```suggestion
	return providers.filter(provider => !!providerNames[provider]);
```

svidgen: 🤷

ashika112: slight note we would need map and filter in this case :P

calebpollman (approved):

Issue #10469, opened 2022-10-13
Ability to resume authentication flow


LouisYoungGMSL:
### Is this related to a new or existing framework?

React

### Is this related to a new or existing API?

Authentication

### Describe the feature you'd like to request

We'd like to be able to resume an authentication flow when completing a custom challenge (MFA in this case) that requires us to navigate away from the current login page. The MFA page would redirect us back to the login page where we would then expect to send the custom challenge response via Amplify to Cognito.

Unfortunately, all session information prior to logging in is currently lost from Amplify when navigating away in the middle of a authentication flow which means we cannot simply pick up where we left off.

### Describe the solution you'd like

A possible solution might be to introduce a mechanism to somehow temporarily persist the relevant session details that are used to authenticate (possibly using a browser storage API) so that it can be retrieved at a later time when resuming an authentication flow (such as after a redirect from/to the application).

This would potentially require additional APIs to allow us to store a session and retrieve an existing session such that we can support navigating away from an authentication flow without losing that vital context.

### Describe alternatives you've considered

I've tested a proof of concept of the suggested possible solution above by persisting the user manually. This isn't ideal as the user returned by `Auth.signIn` is an instance of `CognitoUser` that can't be serialised and stored and then retrieved and deserialised properly.

To work around this, I've attempted to hydrate/reconstruct a `CognitoUser` by using the properties from the user returned by `Auth.signIn` that I've persisted in a browser storage API. From what I can see, there isn't a documented way to do this.

```ts
// Logic to determine whether initially signing in or continuing MFA journey.
if (resumingLogin) {
  resumeSignIn()
} else {
  doSignIn()
}

const doSignIn = () => {
  // Navigate away to MFA provider.
  ...
}

const resumeSignIn = () => {
  const deserialisedUser = ...;
  const challengeResponse = ...; // In our case, we'd retrieve this from the URL query parameters.


  const cognitoUserPool = new CognitoUserPool({
    UserPoolId: "...",
    ClientId: "...",
  });

  const cognitoUser = new CognitoUser({
    Username: deserialisedUser.username,
    Pool: cognitoUserPool,
  });

  cognitoUser.authenticationFlowType = deserialisedUser.authenticationFlowType;
  cognitoUser.challengeName = deserialisedUser.challengeName;
  cognitoUser.challengeParam = deserialisedUser.challengeParam;
  cognitoUser.Session = deserialisedUser.Session;
  
[...]

PR #13483, opened 2024-06-07
feat(auth): Enable resumable SignIn

joon-won:
#### Description of changes

- Enable v6 users to have an option to continue SignIn process after a full browser reload by storing, then utilizing session in session storage

#### Issue #, if available

- #10469 
- #12810 

- Unit tests (IP)
- Smoke Test on sample application

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +67,61 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+	const expiry = sessionStorage.getItem(SIGN_IN_STATE_KEYS.expiry) as string;
+	if (isExpired(expiry)) {
+		logger.warn('Sign-in session expired');
+		clearPersistedSignInState();
+
+		return getDefaultState();
+	}
+
+	const username =
+		sessionStorage.getItem(SIGN_IN_STATE_KEYS.username) ?? ('' as string);
```
israx, line 106: are we sure we want to default to an empty string ? What would happen with the underling API calls that use this value ?

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +67,61 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+	const challengeName = (sessionStorage.getItem(
+		SIGN_IN_STATE_KEYS.challengeName,
+	) ?? '') as ChallengeName;
```
israx, line 109: ditto about the defaulting to an empty string

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +67,61 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+	const signInSession =
+		sessionStorage.getItem(SIGN_IN_STATE_KEYS.signInSession) ?? ('' as string);
```
israx, line 111: ditto about the defaulting to an empty string

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +67,61 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+
+	// Clear SignInStage from sessionStorage after re-hydration
+	clearPersistedSignInState();
```
israx, line 114: we should not do this. We might be creating undesired side effects and losing state as result. Let's consider the following scenarios:

1. `confirmSignIn` ->  confirmSignIn fails -> `confirmSign`  
2. `confirmSignIn` ->  confirmSignIn fails -> user reloads page -> `confirmSign`  

Please let's test both cases and make sure state is not lost

israx (commented):
Thanks @joon-won . We are getting close. A couple of things, 

1. we should consider creating an internal sync `sessionStorage` util so we don't have change sync behavior of the `signInStore`

2. we must avoid losing or modifying state at all cost. Left comments bellow and pointed out the tests needed

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +67,61 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
 				username: action.value,
 			};
 		case 'SET_INITIAL_STATE':
-			return defaultState();
+			return getInitialState();
```
israx, line 70: probably rename it to `setInitialState`

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -27,6 +29,22 @@ type Store<State, Action> = (reducer: Reducer<State, Action>) => {
 
 type Reducer<State, Action> = (state: State, action: Action) => State;
 
+const logger = new ConsoleLogger('Auth signInStore');
+
+// Minutes until stored session invalidates
+const EXPIRATION_MINUTES = 5;
```
israx, line 35: are we sure 5 minutes is the right default time ? The authentication session expiry time can be set from 3 to 15 minutes. We might to consider increasing it a bit more

joon-won: Great point, can there any potential issue if I match it with the maximum possible expiry time, 15 mins?

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +67,61 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+
+const getDefaultState = (): SignInState => ({
+	username: undefined,
+	challengeName: undefined,
+	signInSession: undefined,
+});
+
+// Hydrate signInStore from sessionStorage
+const getInitialState = (): SignInState => {
+	const expiry = sessionStorage.getItem(SIGN_IN_STATE_KEYS.expiry) as string;
```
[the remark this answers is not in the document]
joon-won: I see, thanks for pointing this out. 
Seems like sync sessionStorage util will be required to resolve the issue, as removing the cast from current code will cause an error

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -27,6 +29,22 @@ type Store<State, Action> = (reducer: Reducer<State, Action>) => {
...
+// Minutes until stored session invalidates
+const EXPIRATION_MINUTES = 5;
+const MS_TO_EXPIRY = 1000 * 60 * EXPIRATION_MINUTES;
+const SIGN_IN_STATE_KEYS = [
+	'username',
+	'challengeName',
+	'signInSession',
+	'expiry',
+].reduce((keys: Record<string, string>, key) => {
+	keys[key] = `CognitoSignInState.${key}`;
```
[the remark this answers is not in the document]
joon-won: Working on it!

packages/auth/src/providers/cognito/utils/signInStore.ts
joon-won: This is not ideal, yet not defaulting it to an empty string will cause error due to `SignInState` implementation `username?: string` part. This is one of the original implementation so decided not to change it

packages/auth/src/providers/cognito/utils/signInStore.ts
joon-won: Yeah, this one also comes out from the same issue. Original implementation of `SignInState` will require default value for it to best of my understanding (IDE is complaining about it). Modifying the current implementation of it was an option I've considered, yet I went very conservative about it because I wasn't sure if it's in the design scope

packages/auth/src/providers/cognito/utils/signInStore.ts
joon-won: Same issue as above 2, removed casting.

packages/auth/src/providers/cognito/utils/signInStore.ts
joon-won: Oh, thanks for pointing it out. I've totally missed out on the possibility. Removed the line clearing the sessionStorage

packages/auth/src/providers/cognito/utils/signInStore.ts
israx: can it be default to an `undefined` value instead ?

joon-won: Yup it can be. I've defaulted them to `undefined`

packages/auth/src/providers/cognito/utils/signInStore.ts
joon-won: For future reference, we've decided to default it to 3 minute

packages/auth/src/providers/cognito/utils/signInStore.ts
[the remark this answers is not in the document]
joon-won: For future reference, `SIGN_IN_STATE_KEYS` is now an object

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +63,70 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+
+const getDefaultState = (): SignInState => ({
+	username: undefined,
+	challengeName: undefined,
+	signInSession: undefined,
+});
+
+// Hydrate signInStore from Synced Session Storage
+const setInitialState = (): SignInState => {
+	const expiry = syncSessionStorage.getItem(signInStateKeys.expiry) as string;
```
cshfang, line 104: I think this is incorrectly cast. The result of this *might* be `null` so casting it to string is actually unsafe. What should happen instead is to update your `isExpired` function to expect `null` as a possible input and handling it appropriately

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +63,70 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+
+const clearPersistedSignInState = (keys: Record<string, string>) => {
+	for (const key in keys) {
+		syncSessionStorage.removeItem(key);
+	}
+};
+
+// Clear saved sign in states from both memory and Synced Session Storage
+export function cleanActiveSignInState(): void {
+	signInStore.dispatch({ type: 'SET_INITIAL_STATE' });
```
cshfang, line 87: We already went over this but this used to dispatch an action to clear the state. This was *the same thing* as setting to initial state before because the initial state was ALWAYS the default (empty) state. However, this is no longer true - so we actually want to dispatch (and add) an action like.. `SET_EMPTY_STATE` or `RESET_STATE` to actually "clear out" the state. This was a miss by me in my original POC!

cshfang (changes requested):

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +63,70 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
 
-function defaultState(): SignInState {
+const isExpired = (expiryDate: string): boolean => {
+	const expiryTimestamp = Number(expiryDate);
+	const currentTimestamp = Date.now();
+
+	return expiryTimestamp <= currentTimestamp;
+};
+
+const clearPersistedSignInState = (keys: Record<string, string>) => {
```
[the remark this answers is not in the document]
joon-won: Gotcha, the idea was to give better visibility to users over what's passed. Nonetheless, I agree we won't have any other usability for the method.

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +63,70 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+	for (const key in keys) {
```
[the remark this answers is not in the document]
joon-won: Per our conversation above removed parameter passing and made it just refer to `SIGN_IN_KEYS`

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +63,70 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+		syncSessionStorage.removeItem(key);
+	}
+};
+
+// Clear saved sign in states from both memory and Synced Session Storage
+export function cleanActiveSignInState(): void {
+	signInStore.dispatch({ type: 'SET_INITIAL_STATE' });
+	clearSignInStateKeysFromSessionStorage();
```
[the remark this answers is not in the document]
joon-won: Removed method `clearSignInStateKeysFromSessionStorage()` and make it call `clearPersistedSignInState()` directly as we have discarded idea of exporting `clearSignInStateKeysFromSessionStorage()`

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +63,70 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+}
+
+// Clear stored values for sign in state keys in Synced Session Storage
+export function clearSignInStateKeysFromSessionStorage(): void {
+	clearPersistedSignInState(signInStateKeys);
+}
```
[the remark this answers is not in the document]
joon-won: Per our previous conversation, `clearPersistedSignInState()` no longer takes params.

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +63,70 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+
+const getDefaultState = (): SignInState => ({
+	username: undefined,
+	challengeName: undefined,
+	signInSession: undefined,
+});
+
+// Hydrate signInStore from Synced Session Storage
+const setInitialState = (): SignInState => {
```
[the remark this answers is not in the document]
joon-won: Makes sense, modified it to `initializeState()`, this will give better image

packages/auth/src/providers/cognito/utils/signInStore.ts
joon-won: That's a great point, I'll add one more safety layer onto it, trying to prevent potential risks coming out of `syncSessionStorage.getItem(signInStateKeys.expiry)` being null

packages/auth/src/providers/cognito/utils/signInStore.ts
joon-won: Thanks for helping me out on debugging! I probably needed longer time to find it out myself. Learned a lesson here about dispatcher and reducer 👍

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +64,74 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
 				username: action.value,
 			};
 		case 'SET_INITIAL_STATE':
-			return defaultState();
+			return initializeState();
+		case 'RESET_STATE':
+			return getDefaultState();
 		default:
```
joon-won, line 70: NIT: this default state is never reachable due to `signInReducer` requiring a type of `SignInState`. This is the only line case in this file that's not being covered

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +64,74 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+		username,
+		challengeName,
+		signInSession,
 	};
-}
+};
 
 const createStore: Store<SignInState, SignInAction> = reducer => {
-	let currentState = reducer(defaultState(), { type: 'SET_INITIAL_STATE' });
+	let currentState = reducer(getDefaultState(), { type: 'SET_INITIAL_STATE' });
```
israx, line 134: Thinking twice, we can probably dispatch a different action if the session was already persisted. That way we can addressed the concerns above

```suggestion
const action = isStatePersisted() ? {type:'SET_PERSISTED_STATE'} : { type: 'SET_INITIAL_STATE' };
let currentState = reducer(getDefaultState(), action );
```

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +67,61 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
 
-function defaultState(): SignInState {
+const isExpired = (expiryDate: string): boolean => {
+	const expiryTimestamp = Number(expiryDate);
+	const currentTimestamp = Date.now();
+
+	return expiryTimestamp <= currentTimestamp;
+};
+
+const clearPersistedSignInState = () => {
```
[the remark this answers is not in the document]
israx: I think we still need to address this comment. I was suggesting to pass the `syncSessionStorage` as an argument. 

```suggestion
const clearPersistedSignInState = (storage:SyncStorageInterface) => {
```

israx (changes requested):
Thanks @joon-won for putting these together. 

Sorry for not catching this earlier but I realize that the `SET_INITIAL_STATE` action is already being dispatched and modifying the state based on the `expiry` time might lead to mutating the state. I think is worth considering adding a new action and dispatching it if the session was already persisted.

packages/auth/__tests__/providers/cognito/signInResumable.test.ts
```diff
@@ -0,0 +1,299 @@
...
+
+	test('State is set after calling setActiveSignInState', async () => {
+		setActiveSignInState(user1);
+		const localSignInState = signInStore.getState();
+
+		expect(localSignInState).toEqual(user1);
+		cleanActiveSignInState();
+	});
+
+	test('State is updated after ', async () => {
```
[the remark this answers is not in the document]
joon-won: Sorry, seems it was cut off. Good finding!

packages/core/__tests__/storage/SyncSessionStorage.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+	test('should not throw if trying to delete a non existing key', () => {
+		const badKey = 'nonExistingKey';
+
+		expect(() => {
+			sessionStorage.removeItem(badKey);
+		}).not.toThrow();
+	});
+
+	test('syncSessionStorage instance should show the same behavior', () => {
```
HuiSF, line 91: Is this test merely for `syncSessionStorage instanof SyncSessionStorage === true`?

packages/core/__tests__/storage/storage-mechanisms-node-runtime.test.ts
```diff
@@ -26,4 +30,13 @@ describe('test mechanisms', () => {
 			expect(error.name).toBe(AmplifyErrorCode.PlatformNotSupported);
 		}
 	});
+
+	test('test syncSessionStorage operations in node environment', () => {
```
HuiSF, line 34: Does this test even run? It gets the fallback InMemoryStorage in Node env.

packages/auth/src/providers/cognito/utils/signInStore.ts
HuiSF: You can add this info as a code comment :)

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +64,74 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+	for (const key in signInStateKeys) {
+		syncSessionStorage.removeItem(key);
+	}
+};
+
+// Clear saved sign in states from both memory and Synced Session Storage
+export function cleanActiveSignInState(): void {
```
HuiSF, line 92: nit: use consistent function expiression pattern throughout

packages/core/__tests__/storage/storage-mechanisms-node-runtime.test.ts
```diff
@@ -3,7 +3,11 @@
  */
 
 import { AmplifyError, AmplifyErrorCode } from '../../src/libraryUtils';
-import { defaultStorage, sessionStorage } from '../../src/storage';
+import {
```
cshfang, line 6: I think we can just delete this whole test file. Even in Node environment, these would just fallback to in-memory versions. The fact that these are passing just means the `catch` blocks aren't running. So, effectively, this entire test suite is useless.

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -81,8 +148,22 @@ export function setActiveSignInState(state: SignInState): void {
...
+	persistSignInState(syncSessionStorage, state);
 }
+
+// Save local state into Session Storage
+const persistSignInState = (
+	storage = syncSessionStorage,
+	{
+		challengeName = '' as ChallengeName,
+		signInSession = '',
+		username = '',
```
HuiSF, line 162: Are these inital state correct? would this cause storing empty strings always into session storage with the `signInStateKeys`?

packages/core/__tests__/storage/SyncSessionStorage.test.ts
```diff
@@ -0,0 +1,123 @@
...
+		const newUserName = 'joonchoi+tobedeleted';
+		sessionStorage.setItem(signInStateKeys.username, newUserName);
+		expect(sessionStorage.getItem(signInStateKeys.username)).toEqual(
+			newUserName,
+		);
+		sessionStorage.removeItem(signInStateKeys.username);
+		expect(sessionStorage.getItem(signInStateKeys.username)).toBeNull();
+	});
+
+	test('should clear out storage', () => {
```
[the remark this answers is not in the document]
cshfang: Nit to comment: it('clears...

packages/auth/src/providers/cognito/utils/signInStore.ts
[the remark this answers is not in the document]
cshfang: This is probably my fault because my argument is "Do we really need to pass this in? This function isn't exported from the module and the module scope contains the reference to the `syncSessionStorage` instance already right?"

packages/core/__tests__/storage/SyncSessionStorage.test.ts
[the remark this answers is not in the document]
HuiSF: Updated 😊

packages/auth/src/providers/cognito/utils/signInStore.ts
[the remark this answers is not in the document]
cshfang: The defaults were because `setItem` is typed as non-nullable. We can use a guard clause in the body instead to early return - but is even a real situation where this would be called with a partial state?

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -81,8 +148,22 @@ export function setActiveSignInState(state: SignInState): void {
...
 
-export function cleanActiveSignInState(): void {
-	signInStore.dispatch({ type: 'SET_INITIAL_STATE' });
+	// Save the local state into Synced Session Storage
+	persistSignInState(syncSessionStorage, state);
 }
+
+// Save local state into Session Storage
+const persistSignInState = (
+	storage = syncSessionStorage,
```
cshfang, line 158: Imo - this is redundant for the same reason I gave above around the `syncSessionStorage` reference already being accessible via the module scope. I do understand the rationale of making things more pure but here, we should be able to treat it as a member method of this module.

packages/auth/src/providers/cognito/utils/signInStore.ts
cshfang: It's not a super necessary refactor imo, but I am onboard with the change. Essentially, taking the logic out of the mutation function and making a determination of which type of action to dispatch makes sense.

We can repurpose `RESET_STATE` as `SET_DEFAULT_STATE` so it makes more sense for either use case (clear and setting an initial empty state)

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -49,22 +64,74 @@ const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
...
+	username: undefined,
+	challengeName: undefined,
+	signInSession: undefined,
+});
+
+// Hydrate signInStore from Synced Session Storage
+const initializeState = (): SignInState => {
+	if (!syncSessionStorage) {
+		return getDefaultState();
+	}
```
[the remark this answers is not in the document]
israx: let's delete this code please

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -83,6 +159,17 @@ export function setActiveSignInState(state: SignInState): void {
...
 }
 
-export function cleanActiveSignInState(): void {
-	signInStore.dispatch({ type: 'SET_INITIAL_STATE' });
-}
+// Save local state into Session Storage
+const persistSignInState = ({
+	challengeName = '' as ChallengeName,
+	signInSession = '',
+	username = '',
```
israx, line 166: As previously discussed, let's avoid setting default values

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -27,44 +30,117 @@ type Store<State, Action> = (reducer: Reducer<State, Action>) => {
...
 const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
 	switch (action.type) {
 		case 'SET_SIGN_IN_SESSION':
 			return {
 				...state,
 				signInSession: action.value,
 			};
+
 		case 'SET_SIGN_IN_STATE':
+			persistSignInState(state);
```
israx, line 54: the `state` being passed in is not the latest `state` from the store. The `persistSignInState` should take the `action` as an argument instead.

```suggestion
persistSignInState(action.value);
```

packages/core/__tests__/storage/storage-mechanisms-node-runtime.test.ts
```diff
@@ -1,29 +0,0 @@
-/**
```
israx, line 1: why are we deleting this file? — We should keep the same coverage

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -111,32 +108,32 @@ const getDefaultState = (): SignInState => ({
...
-	if (!syncSessionStorage) {
-		return getDefaultState();
-	}
-
 	const expiry = syncSessionStorage.getItem(signInStateKeys.expiry);
-	if (isExpired(expiry)) {
-		logger.warn('Sign-in session expired');
-		clearPersistedSignInState();
+	if (expiry) {
+		if (isExpired(expiry)) {
```
israx, line 113: you can collapse the 2 if statements into one

```suggestion
if (expiry&&isExpired(expiry)) {
```

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -161,13 +158,19 @@ export function setActiveSignInState(state: SignInState): void {
...
+	challengeName,
+	signInSession,
+	username,
 }: SignInState) => {
-	syncSessionStorage.setItem(signInStateKeys.username, username);
-	syncSessionStorage.setItem(signInStateKeys.challengeName, challengeName);
-	syncSessionStorage.setItem(signInStateKeys.signInSession, signInSession);
+	if (username) {
+		syncSessionStorage.setItem(signInStateKeys.username, username);
+	}
```
israx, line 167:
```suggestion
username && syncSessionStorage.setItem(signInStateKeys.username, username);
```

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -161,13 +158,19 @@ export function setActiveSignInState(state: SignInState): void {
...
+	if (signInSession) {
+		syncSessionStorage.setItem(signInStateKeys.signInSession, signInSession);
```
israx, line 169: same suggestion of using the && operator

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -161,13 +158,19 @@ export function setActiveSignInState(state: SignInState): void {
...
+	}
+	if (challengeName) {
```
israx, line 171: same suggestion of using the && operator

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -161,13 +158,19 @@ export function setActiveSignInState(state: SignInState): void {
...
+		syncSessionStorage.setItem(signInStateKeys.challengeName, challengeName);
+	}
 	syncSessionStorage.setItem(
 		signInStateKeys.expiry,
 		String(Date.now() + MS_TO_EXPIRY),
```
israx, line 176: you want to update the session expiry only when a new session is passed

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -27,44 +30,112 @@ type Store<State, Action> = (reducer: Reducer<State, Action>) => {
...
+	signInSession: 'CognitoSignInState.signInSession',
+	expiry: 'CognitoSignInState.expiry',
+};
+
 const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
 	switch (action.type) {
 		case 'SET_SIGN_IN_SESSION':
 			return {
 				...state,
 				signInSession: action.value,
```
israx, line 50: I think we need to persist the session here too.

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -27,44 +30,112 @@ type Store<State, Action> = (reducer: Reducer<State, Action>) => {
...
+	username: 'CognitoSignInState.username',
+	challengeName: 'CognitoSignInState.challengeName',
+	signInSession: 'CognitoSignInState.signInSession',
+	expiry: 'CognitoSignInState.expiry',
+};
+
 const signInReducer: Reducer<SignInState, SignInAction> = (state, action) => {
 	switch (action.type) {
 		case 'SET_SIGN_IN_SESSION':
 			return {
```
cshfang, line 48: Why don't we need to persist when session is set?

packages/auth/__tests__/providers/cognito/signInResumable.test.ts
```diff
@@ -0,0 +1,278 @@
...
+		expect(localSignInState).toEqual({
+			challengeName: undefined,
+			signInSession: undefined,
+			username: undefined,
+		});
+		cleanActiveSignInState();
+	});
+
+	test('State is set after calling setActiveSignInState', async () => {
+		setActiveSignInState(user1);
```
israx, line 101: can you spy `persistSignInState` method and make sure is called with the right parameters?

packages/auth/__tests__/providers/cognito/signInResumable.test.ts
```diff
@@ -0,0 +1,278 @@
...
+			user1.signInSession,
+		);
+		syncSessionStorage.setItem(
+			signInStateKeys.challengeName,
+			user1.challengeName,
+		);
+		syncSessionStorage.setItem(
+			signInStateKeys.expiry,
+			(new Date().getTime() + 5000).toString(),
+		);
```
israx, line 213: Based on the test description, these values shouldn't be set manually but instead set when `signInStore.dispatch` is called

packages/auth/__tests__/providers/cognito/signInResumable.test.ts
```diff
@@ -0,0 +1,278 @@
...
+		expect(localSignInState).toEqual({
+			username: user1.username,
+			challengeName: user1.challengeName,
+			signInSession: user1.signInSession,
+		});
+
+		signInStore.dispatch({
+			type: 'SET_SIGN_IN_SESSION',
+			value: newSignInSessionID,
+		});
```
israx, line 229: can you spy `persistSignInState` method and make sure is called with the right parameters when `signInStore.dispatch` is called here?

packages/core/__tests__/storage/SyncSessionStorage.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+	it('clears all items', () => {
+		sessionStorage.clear();
+		expect(sessionStorage.getItem(signInStateKeys.username)).toBeNull();
+		expect(sessionStorage.getItem(signInStateKeys.challengeName)).toBeNull();
+		expect(sessionStorage.getItem(signInStateKeys.signInSession)).toBeNull();
+		expect(sessionStorage.getItem(signInStateKeys.expiry)).toBeNull();
+	});
+
+	it('will throw if trying to delete a non existing key', () => {
```
cshfang, line 82: Nit: Are we expecting this *not* to throw - like the expect below?

packages/auth/__tests__/providers/cognito/signInResumable.test.ts
```diff
@@ -0,0 +1,278 @@
...
+
+		expect(localSignInState).toEqual({
+			username: user1.username,
+			challengeName: user1.challengeName,
+			signInSession: user1.signInSession,
+		});
+		cleanActiveSignInState();
+	});
+
+	test('Stored session is expired thus State should default to undefined', async () => {
```
israx, line 171: nit
```suggestion
	test('should return undefined sign-in state values when stored session is expired', async () => {
```

packages/auth/__tests__/providers/cognito/signInResumable.test.ts
```diff
@@ -0,0 +1,278 @@
...
+			username,
+			signInDetails: {
+				loginId: username,
+				authFlowType: 'USER_SRP_AUTH',
+			},
+		});
+		handleUserSRPAuthflowSpy.mockClear();
+	});
+
+	test('Stored session is not expired thus State should be rehydrated', () => {
```
israx, line 142: Nit. The wording could be improved for better clarity.

```suggestion
The stored sign-in state should be rehydrated if the sign-in session is still valid.
```

packages/auth/__tests__/providers/cognito/signInResumable.test.ts
```diff
@@ -0,0 +1,278 @@
...
+			signInStateKeys.challengeName,
+			user1.challengeName,
+		);
+		syncSessionStorage.setItem(
+			signInStateKeys.expiry,
+			(new Date().getTime() - 500000).toString(),
+		);
+		signInStore.dispatch({
+			type: 'SET_INITIAL_STATE',
+		});
```
israx, line 187: ditto about extracting this logic into a `helper` or putting it in a `beforeEach`

packages/auth/__tests__/providers/cognito/signInResumable.test.ts
```diff
@@ -0,0 +1,278 @@
...
+			user1.challengeName,
+		);
+		syncSessionStorage.setItem(
+			signInStateKeys.expiry,
+			(new Date().getTime() + 5000).toString(),
+		);
+
+		signInStore.dispatch({
+			type: 'SET_INITIAL_STATE',
+		});
```
israx, line 217: ditto about extracting this logic into a `helper` or putting it in a `beforeEach`, and removing `signInStore.dispatch`

cshfang (approved):

israx (approved):

packages/aws-amplify/jest.config.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
 	coverageThreshold: {
 		global: {
 			branches: 85,
-			functions: 65.5,
+			functions: 63.3,
```
jimblanc, line 6: Where'd the drop come from? Anyway to bring this back up?

jimblanc (dismissed):

packages/core/src/storage/SyncKeyValueStorage.ts
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	/**
+	 * This is used to set a specific item in storage
+	 * @param {string} key - the key for the item
+	 * @param {object} value - the value
+	 * @returns {string} value that was set
+	 */
+	setItem(key: string, value: string) {
+		if (!this.storage) throw new PlatformNotSupportedError();
```
HuiSF, line 24: Can implement a getter function to handle the exception throwing at one single place (saving a few bits probably :D ).

packages/core/__tests__/storage/SyncSessionStorage.test.ts
```diff
@@ -0,0 +1,89 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { SyncSessionStorage } from '../../src/storage/SyncSessionStorage';
+
+jest.mock('@aws-amplify/core/internals/utils', () => ({
+	...jest.requireActual('@aws-amplify/core/internals/utils'),
+	isBrowser: jest.fn(() => false),
```
HuiSF, line 7: Why is this mock needed? The `SyncSessionStorage` impl. doesn't seem rely on this function.

packages/core/__tests__/storage/SyncSessionStorage.test.ts
```diff
@@ -0,0 +1,89 @@
...
+		signInSession: '888577-ltfgo-42d8-891d-666l858766g7',
+		expiry: '1234567',
+	};
+
+	beforeEach(() => {
+		sessionStorage = new SyncSessionStorage();
+	});
+
+	afterEach(() => {
+		sessionStorage.clear();
```
HuiSF, line 31: nit: A new instance of `SyncSessionStorage` is being created before each test, this client probably is not needed?

packages/core/__tests__/storage/SyncSessionStorage.test.ts
```diff
@@ -0,0 +1,89 @@
...
+	});
+
+	it('can set and retrieve item by key', () => {
+		sessionStorage.setItem(signInStateKeys.username, user1.username);
+		sessionStorage.setItem(signInStateKeys.challengeName, user1.challengeName);
+		sessionStorage.setItem(signInStateKeys.signInSession, user1.signInSession);
+		sessionStorage.setItem(signInStateKeys.expiry, user1.expiry);
+
+		expect(sessionStorage.getItem(signInStateKeys.username)).toEqual(
```
HuiSF, line 40: nit: could use `strictEqual` for this type of tests

packages/core/__tests__/storage/SyncSessionStorage.test.ts
```diff
@@ -0,0 +1,89 @@
...
+		sessionStorage.setItem(signInStateKeys.username, newUserName);
+		expect(sessionStorage.getItem(signInStateKeys.username)).toEqual(
+			newUserName,
+		);
+		sessionStorage.removeItem(signInStateKeys.username);
+		expect(sessionStorage.getItem(signInStateKeys.username)).toBeNull();
+	});
+
+	it('clears all items', () => {
+		sessionStorage.clear();
```
HuiSF, line 75: In this test, there were no key-values written into the newly created instance of the `SyncSessionStorage` before calling the `clear()`, this test may not be accurate.

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -27,44 +30,118 @@ type Store<State, Action> = (reducer: Reducer<State, Action>) => {
...
+const logger = new ConsoleLogger('Auth signInStore');
+
+// Minutes until stored session invalidates
+const EXPIRATION_MINUTES = 3;
+const MS_TO_EXPIRY = 1000 * 60 * EXPIRATION_MINUTES;
+const signInStateKeys = {
+	username: 'CognitoSignInState.username',
+	challengeName: 'CognitoSignInState.challengeName',
+	signInSession: 'CognitoSignInState.signInSession',
+	expiry: 'CognitoSignInState.expiry',
```
HuiSF, line 42: nit: can assign `CognitoSignInState` to a constant and construct the strings here (can save a few bytes I presume)

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -27,44 +30,118 @@ type Store<State, Action> = (reducer: Reducer<State, Action>) => {
...
+const clearPersistedSignInState = () => {
+	for (const storedKey of Object.values(signInStateKeys)) {
+		syncSessionStorage.removeItem(storedKey);
+	}
+};
+
+// Clear saved sign in states from both memory and Synced Session Storage
+export function cleanActiveSignInState(): void {
+	signInStore.dispatch({ type: 'RESET_STATE' });
+	clearPersistedSignInState();
```
HuiSF, line 106: I think `clearPersistedSignInState()` should be invoked at line 82, under the `case 'RESET_STATE'` event in the reducer.

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -27,44 +30,118 @@ type Store<State, Action> = (reducer: Reducer<State, Action>) => {
...
+	challengeName: undefined,
+	signInSession: undefined,
+});
+
+// Hydrate signInStore from Synced Session Storage
+const initializeState = (): SignInState => {
+	const expiry = syncSessionStorage.getItem(signInStateKeys.expiry);
+
+	if (!expiry || (expiry && isExpired(expiry))) {
+		logger.warn('Session Expired');
```
HuiSF, line 120: sanity check: do we really need this log? (it increases bundle size)

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -27,44 +30,118 @@ type Store<State, Action> = (reducer: Reducer<State, Action>) => {
 
 type Reducer<State, Action> = (state: State, action: Action) => State;
 
+const logger = new ConsoleLogger('Auth signInStore');
+
+// Minutes until stored session invalidates
+const EXPIRATION_MINUTES = 3;
+const MS_TO_EXPIRY = 1000 * 60 * EXPIRATION_MINUTES;
```
HuiSF, line 37:
```suggestion
const MS_TO_EXPIRY = 3 * 60 * 1000; // 3 mins
```

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -27,44 +30,116 @@ type Store<State, Action> = (reducer: Reducer<State, Action>) => {
...
+const clearPersistedSignInState = () => {
+	for (const storedKey of Object.values(signInStateKeys)) {
+		syncSessionStorage.removeItem(storedKey);
+	}
+};
+
+// Clear saved sign in states from both memory and Synced Session Storage
+export function cleanActiveSignInState(): void {
+	signInStore.dispatch({ type: 'RESET_STATE' });
+	clearPersistedSignInState();
```
HuiSF, line 106: Since you've moved this call to the reducer, we can probably remove it here.

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -30,16 +30,14 @@ type Store<State, Action> = (reducer: Reducer<State, Action>) => {
 
 type Reducer<State, Action> = (state: State, action: Action) => State;
 
-const logger = new ConsoleLogger('Auth signInStore');
-
 // Minutes until stored session invalidates
-const EXPIRATION_MINUTES = 3;
-const MS_TO_EXPIRY = 1000 * 60 * EXPIRATION_MINUTES;
+const MS_TO_EXPIRY = 3 * 60 * 1000; // 3 mins
+const tgtState = 'CognitoSignInState';
```
israx, line 35: Constants should follow an UPPER_CASE_SNAKE syntax

packages/auth/src/providers/cognito/apis/signInWithCustomSRPAuth.ts
```diff
@@ -89,6 +86,7 @@ export async function signInWithCustomSRPAuth(
 			signInDetails,
 		});
 		if (AuthenticationResult) {
+			signInStore.dispatch({ type: 'RESET_STATE' });
```
cshfang, line 89: Why did this move up in the block?

packages/auth/src/providers/cognito/apis/signInWithUserPassword.ts
```diff
@@ -84,6 +81,7 @@ export async function signInWithUserPassword(
 			signInDetails,
 		});
 		if (AuthenticationResult) {
+			signInStore.dispatch({ type: 'RESET_STATE' });
```
cshfang, line 84: Ditto here

cshfang (approved):

HuiSF (approved):

aca-stev:
Hi, I noticed this PR was previously merged but later reverted. Could you share why it was reverted and if there are plans to  re-integrate? Thanks!

joon-won:
Hey @aca-stev, this PR was previously reverted due to our internal plan change. This is now re-released with [Amplify JS 6.11.0](https://github.com/aws-amplify/amplify-js/releases/tag/aws-amplify%406.11.0), please feel free to reach out if you have any further questions or concerns!

PR #13508, opened 2024-06-18
fix(aws-amplify): createKeyValueStorageFromCookieStorageAdapter misses default path and secure values

HuiSF:
#### Description of changes

Added default `path: '/'` and `secure: true` to the cookies setting options from the server side.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13509

1. E2E tests
2. Manual testing with a Next.js sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


cshfang (dismissed):

ashika112 (dismissed):

jimblanc (changes requested):
Change looks good to me, but as this is auth related please hold off until we have sign-off.

cshfang (approved):

israx (dismissed):

ashika112 (approved):

ashika112 (approved):

packages/aws-amplify/src/adapter-core/storageFactories/createKeyValueStorageFromCookieStorageAdapter.ts
```diff
@@ -25,6 +27,11 @@ export const createKeyValueStorageFromCookieStorageAdapter = (
 ): KeyValueStorageInterface => {
 	return {
 		setItem(key, value) {
+			// Delete the cookie item first then set it. This results:
+			// SetCookie: key=;expires=1970-01-01;(path='current-path') <- remove path'ed cookies
+			// SetCookie: key=value;expires=Date.now() + 365 days;path=/;secure=true
+			cookieStorageAdapter.delete(key);
```
ashika112, line 33: in case this fails, then self recovery will be the fallback right?

HuiSF: You meant the `.delete` operation itself would fail? If so, this operation just sets client cookie with a expiry time in the past so browser will remove it. I can't think a reason it may fail. (unless the HTTP response never gets delivered to the client)

cshfang (approved):

AllanZhengYP (approved):

AllanZhengYP (approved):

haverchuck:
Have we tested this for both a) apps that were spun up before the fix, with tokens already persisted, and b) new apps?

HuiSF:
> Have we tested this for both a) apps that were spun up before the fix, with tokens already persisted, and b) new apps?

@haverchuck here's the manual testing setup:

### Set 1

1. A next.js sample app with the page router installed the `latest` version of `aws-amplify` and `@aws-amplify/adapter-nextjs`
2. access route `/fetch-auth-session/test` with expired access token, server-side refreshed tokens are set back in the browser cookie store with `/fetch-auth-session` path (wrong cookie persisted)
3. Swap the `aws-amplify` and `@aws-amplify/adapter-nextjs` with a version that contains this PR fix
4. Wait till the tokens set in step 2 expired, and refresh the page, the originally cookie item set with path `/fetch-auth-session` got removed, and the newly refreshed tokens are set in the cookie store with `/` path.

### Set 3

1. A next.js sample with the page router installed a version of `aws-amplify` and `@aws-amplify/adapter-nextjs` that contains this PR fix
2. access route `/fetch-auth-session/test`
3. Wait till the access token expires, then refresh the page, the newly refreshed tokens are set in the cookie store with `/` path.

haverchuck (approved):

PR #13512, opened 2024-06-18
feat(auth): HostedUI oidc signout

Samaritan1011001:
#### Description of changes

Updates the signOut API to allow customers to provide which redirect url to use enabling sign out of users from both Cognito Hosted UI and the upstream OIDC provider.

Sample app with HostedUI setup with Auth0 as the OIDC provider.
- sign in using signWithRedirect
- sign out with preferredRedirectUrl as the Auth0's logout endpoint
- call signInWithRedirect to see that the Auth0 login again requires to enter user credentials.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.ts
```diff
@@ -6,18 +6,33 @@ import {
...
-		redirects?.find(isTheSameDomain);
-	const redirectUrlFromDifferentOrigin =
-		redirects?.find(isHttps) ?? redirects?.find(isHttp);
-	if (redirectUrlFromTheSameOrigin) {
-		return redirectUrlFromTheSameOrigin;
-	} else if (redirectUrlFromDifferentOrigin) {
-		throw invalidOriginException;
+export function getRedirectUrl(
+	redirects: string[],
+	_preferredSignOutUrl?: string,
```
cshfang, line 11: What's the reason for adding the `_`?

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.ts
```diff
@@ -6,18 +6,33 @@ import {
...
+	if (_preferredSignOutUrl) {
+		const redirectUrl = redirects?.find(
+			redirect => redirect === _preferredSignOutUrl,
+		);
+		if (!redirectUrl) {
+			throw invalidRedirectException;
+		}
+
+		return redirectUrl;
+	} else {
```
cshfang, line 22: Nit: If clause above has early return, so nesting the rest in the `else` block should be unnecessary

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
```diff
@@ -4,14 +4,27 @@
...
 		throw invalidRedirectException;
 	}
+	if (preferredSignOutUrl) {
+		preferredRedirectUrl = redirects?.find(
+			redirect => redirect === preferredSignOutUrl,
+		);
+		if (!preferredRedirectUrl) {
+			throw invalidRedirectException;
+		}
+	}
```
cshfang, line 27: Couldn't this have used the exact same strategy as the web version?
```ts
if (preferredSignOutUrl) {
	const redirectUrl = redirects?.find(
		redirect => redirect === _preferredSignOutUrl,
	);

	if (!redirectUrl) {
		throw invalidRedirectException;
	}

	return redirectUrl;
}

// existing previous logic
```

israx (commented):
Thanks Manoj. Left a couple of comments

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.ts
Samaritan1011001: I want to say the linter was complaining but not able to repro it so I removed the _

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.ts
Samaritan1011001: Yes but I feel its better for readability with the else block there. It's clear which of the two scenarios to check when reading it.

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
Samaritan1011001: Yes it can be the same, but in the mobile case we should fail early if there is no app scheme found in the array. It does matter what the preferred url is if there is no app scheme in the array for us to register. In the web version, the preferred url takes precedence.

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
```diff
@@ -4,14 +4,27 @@
...
-	if (!redirectUrl) {
+	if (!appSchemeRedirectUrl) {
 		throw invalidRedirectException;
 	}
+	if (preferredSignOutUrl) {
+		preferredRedirectUrl = redirects?.find(
+			redirect => redirect === preferredSignOutUrl,
+		);
+		if (!preferredRedirectUrl) {
+			throw invalidRedirectException;
```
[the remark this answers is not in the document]
Samaritan1011001: That's fair. Updated.

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.ts
```diff
@@ -2,22 +2,38 @@
...
+	if (preferredSignOutUrl) {
+		const redirectUrl = redirects?.find(
+			redirect => redirect === preferredSignOutUrl,
+		);
+		if (!redirectUrl) {
+			throw invalidPreferredRedirectUrlException;
+		}
+
+		return redirectUrl;
+	} else {
```
HuiSF, line 23: nit: you meant the following?

```suggestion
	if (preferredSignOutUrl) {
		if (redirects?.includes(preferredSignOutUrl)) {
			return redirectUrl;
		}

		throw invalidPreferredRedirectUrlException;
	} else {
```

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
```diff
@@ -1,17 +1,38 @@
...
+	const appSchemeRedirectUrl = redirects?.find(
 		redirect =>
 			!redirect.startsWith('http://') && !redirect.startsWith('https://'),
 	);
-	if (!redirectUrl) {
-		throw invalidRedirectException;
+	if (!appSchemeRedirectUrl) {
+		throw invalidAppSchemeException;
 	}
+	if (preferredSignOutUrl) {
```
HuiSF, line 26: nit: could move this to the top to avoid executing the look up of `appSchemeRedirectUrl`?

packages/auth/__tests__/providers/cognito/utils/oauth/getRedirectUrl.test.ts
```diff
@@ -0,0 +1,66 @@
...
+	beforeEach(() => {
+		windowSpy = jest.spyOn(window, 'window', 'get');
+	});
+
+	afterEach(() => {
+		windowSpy.mockRestore();
+	});
+
+	it('should return the redirect url that has the same origin and same pathName', () => {
+		windowSpy.mockImplementation(() => ({
```
HuiSF, line 21: nit: could use `mockReturnValueOnce`

packages/auth/__tests__/providers/cognito/utils/oauth/getRedirectUrl.test.ts
```diff
@@ -0,0 +1,66 @@
...
+			location: {
+				origin: 'https://example.com/',
+				pathname: 'app',
+			},
+		}));
+		expect(getRedirectUrl(mockRedirectUrls)).toStrictEqual(mockRedirectUrls[0]);
+	});
+
+	it('should throw an invalid origin exception if there is no url that is the same origin and pathname', () => {
+		windowSpy.mockImplementation(() => ({
```
HuiSF, line 31: nit: ditto

packages/auth/__tests__/providers/cognito/utils/oauth/getRedirectUrl.test.ts
```diff
@@ -0,0 +1,66 @@
...
+			},
+		}));
+		expect(() => getRedirectUrl(mockRedirectUrls)).toThrow(
+			invalidOriginException,
+		);
+	});
+
+	it('should throw an invalid redirect exception if there is no url that is the same origin/pathname and is also not http or https', () => {
+		const mockNonHttpRedirectUrls = ['test-non-http-string'];
+		windowSpy.mockImplementation(() => ({
```
HuiSF, line 44: nit: ditto

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.ts
Samaritan1011001: That reads well, thanks, I'll update it.

packages/auth/src/providers/cognito/utils/oauth/getRedirectUrl.native.ts
Samaritan1011001: We need `appSchemeRedirectUrl` irrespective of the `preferredSignOutUrl`. We don't return the appScheme above if you notice. Flow is like, throw `appSchemeNotFound` error since that check is always required then check for the `preferredSignOutUrl` and return it if present, if not proceed with the appScheme.

package.json
```diff
@@ -137,5 +137,8 @@
 		"nx": "16.7.0",
 		"xml2js": "0.5.0",
 		"tar": "6.2.1"
+	},
+	"overrides": {
+		"tar": "6.2.1"
```
israx, line 142: is there any reason to add a resolution at the root here?

Samaritan1011001: So, this is required to override the version when using `npm` because `resolutions` block doesn't work with `npm`. 
And this version is the secure version of `tar` that needs to be used. Found via `npm audit`

.github/workflows/callable-e2e-test-detox.yml
```diff
@@ -13,11 +13,15 @@ on:
...
+      host_signout_page:
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   e2e-test:
     name: E2E-Detox ${{ inputs.test_name }}
-    runs-on: macos-latest
+    runs-on: macos-latest-large
```
Samaritan1011001, line 24: Using this large image allows the detox to load the RN app on startup. Without, tests take atleast 2 trys to succeed.

.github/workflows/callable-e2e-test-detox.yml
```diff
@@ -70,6 +74,14 @@ jobs:
           JEST_JUNIT_OUTPUT_NAME: detox-test-results.xml
         working-directory: ${{ inputs.working_directory }}
         shell: bash
+      - name: Start the http-server and host the oidc signout page locally (background)
```
Samaritan1011001, line 77: Step to boot the localhost server.

packages/auth/src/types/inputs.ts
```diff
@@ -46,7 +46,10 @@ export interface AuthSignInInput<
 	options?: ServiceOptions;
 }
 export interface AuthSignOutInput {
-	global: boolean;
+	global?: boolean;
+	oauth?: {
```
Samaritan1011001, line 50: API signature changes

packages/rtn-web-browser/package.json
```diff
@@ -25,7 +25,7 @@
 	},
 	"devDependencies": {
 		"@types/react-native": "0.70.0",
-		"react-native": "0.72.3",
+		"react-native": "0.72.15",
```
Samaritan1011001, line 28: Needed to close a vulnerability on the bad version.

.github/workflows/callable-e2e-test-detox.yml
```diff
@@ -70,6 +74,14 @@ jobs:
...
+        if: ${{ inputs.host_signout_page }}
+        run: |
+          cd oidc-signout-hosted-page/ &&
+          npx http-server -p 8000 &
```
israx, line 81: is this command going to run at the root of the RN app or the monorepo ?

israx (dismissed):

packages/auth/src/types/inputs.ts
```diff
@@ -46,7 +46,10 @@ export interface AuthSignInInput<
 	options?: ServiceOptions;
 }
 export interface AuthSignOutInput {
-	global: boolean;
+	global?: boolean;
```
cshfang, line 49: Is this possibly a breaking change
```ts
const doSomethingIfGlobal = (global: boolean) => {
  if (global) {
    // do something
  }
}

const mySignOut = (input: AuthSignOutInput) => {
  doSomethingIfGlobal(input.global);
}
```

israx: Yeah good catch. I guess we have a trade off between a possible breaking change for an uncommon use case, vs forcing the usage of a parameter that is not needed.

israx: I would rather avoid forcing the usage of the `global` flag to prevent confusion and potential impact on the UX for authenticated users if passed in without knowing what it does.

Samaritan1011001: Yeah, the reasoning behind making it optional is this. 
`If the global parameter is NOT optional, customers will have to provide a value for global if they want to provide a preferredRedirectUrl.`

.github/workflows/callable-e2e-test-detox.yml
HuiSF: Is it possible to encapsulate the http server bootstrap command into a package.json script under the RN test app?

package.json
HuiSF: I think we kept only resolutions based on the assumption that `yarn` is the driver here. If we want to also use npm, should we create a 100% mapping from `resolutions`?

packages/auth/__tests__/providers/cognito/utils/oauth/getRedirectUrl.native.test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+	it('should return preferredRedirectUrl if it matches at least one of the redirect urls from config', () => {
+		const configRedirectUrl = mockRedirectUrls[2];
+
+		expect(getRedirectUrl(mockRedirectUrls, configRedirectUrl)).toStrictEqual(
+			configRedirectUrl,
+		);
+	});
+
+	it('should throw an exception when there is no url with no http or https as prefix irrespective if a preferredSignOutUrl is given or not', () => {
```
HuiSF, line 23: sanity check: this description sounds like that `getRedirectUrl.native.test.ts` always expects at least one custom app scheme, is this correct? Is a universal link (starts with `https`) supposed to work, too?

packages/auth/src/providers/cognito/utils/oauth/oAuthSignOutRedirect.ts
```diff
@@ -12,11 +12,12 @@ import { getRedirectUrl } from './getRedirectUrl';
 export const oAuthSignOutRedirect = async (
 	authConfig: CognitoUserPoolConfig,
 	preferPrivateSession = false,
+	preferredSignOutUrl?: string,
```
HuiSF, line 15: nit: can we use named parameter here?

```
{
  authConfig: CognitoUserPoolConfig,
  preferPrivateSession?: boolean,
  preferredSignOutUrl: string | undefined // required param but allow undefined
}
```

Concern: `preferredSignOutUrl` is currently a optional and positional parameter, calling `oAuthSignOutRedirect` without passing anything as the 3rd param is totally fine. With this, for example, while refactoring someone may forget to pass this param, and can't catch it early.

packages/auth/__tests__/providers/cognito/utils/oauth/getRedirectUrl.native.test.ts
Samaritan1011001: Yeah, it always expect at-least one custom app scheme, it is the current behavior and we are not changing that with this change. We don't support universal links yet.

.github/workflows/push-integ-test.yml
```diff
@@ -8,7 +8,7 @@ concurrency:
 on:
   push:
     branches:
-      - replace-with-your-branch
+      - fix/redirect-rn
```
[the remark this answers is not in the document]
Samaritan1011001: It is intentional but I will revert it soon.

.github/workflows/callable-e2e-test-detox.yml
Samaritan1011001: Root of the RN app. 

Hui, yeah, I thought of it initially, should be possible.

package.json
Samaritan1011001: I would lean towards adding the other `overrides` as the need arises. `tar` version is the only one showing up so far for `npm audit`.

.github/workflows/callable-e2e-test-detox.yml
ashika112: And the CI workflow works as expected?

Samaritan1011001: Yes, it does.

.github/workflows/push-integ-test.yml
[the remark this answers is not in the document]
Samaritan1011001: Reverted

.github/workflows/callable-e2e-test-detox.yml
Samaritan1011001: @HuiSF Done.

.github/workflows/push-integ-test.yml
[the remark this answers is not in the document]
israx: does it need to be reverted again?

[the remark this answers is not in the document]
Samaritan1011001: Yes, I will remove this once this [PR](https://github.com/aws-amplify/amplify-js-samples-staging/pull/949/files) gets merged.

.github/workflows/push-preid-release.yml
```diff
@@ -34,5 +34,4 @@ jobs:
     uses: ./.github/workflows/callable-npm-publish-preid.yml
     # The preid should be detected from the branch name recommending feat/{PREID}/whatever as branch naming pattern
     #   if your branch doesn't follow this pattern, you can override it here for your branch.
-    with:
-      preid: ${{ needs.parse-preid.outputs.preid }}
+    with: preid:${{ needs.parse-preid.outputs.preid }}
```
HuiSF, line 37: Is this change intentional?

packages/auth/src/providers/cognito/utils/oauth/handleOAuthSignOut.native.ts
```diff
@@ -5,20 +5,24 @@ import { CognitoUserPoolConfig } from '@aws-amplify/core';
...
 import { DefaultOAuthStore } from '../../utils/signInWithRedirectStore';
+import { TokenOrchestrator } from '../../tokenProvider';
 
 import { completeOAuthSignOut } from './completeOAuthSignOut';
 import { oAuthSignOutRedirect } from './oAuthSignOutRedirect';
 
 export const handleOAuthSignOut = async (
 	cognitoConfig: CognitoUserPoolConfig,
 	store: DefaultOAuthStore,
+	// No-op here as it's only used in the non-native implementation
```
HuiSF, line 16: nit: if we covert the positional params into named params then we don't need to pass in this?

.github/workflows/push-preid-release.yml
Samaritan1011001: No, reverted.

packages/auth/src/providers/cognito/utils/oauth/handleOAuthSignOut.native.ts
Samaritan1011001: Discussed offline, will keep it the same to reduce complexity on the changes.

HuiSF (dismissed):

israx (dismissed):

cshfang (approved):

packages/auth/src/providers/cognito/apis/signOut.ts
```diff
@@ -60,21 +60,20 @@ export async function signOut(input?: SignOutInput): Promise<void> {
...
 				oAuthStore,
 				tokenOrchestrator,
+				input?.oauth?.preferredSignOutUrl,
 			)) ?? {};
 		if (type === 'error') {
 			throw new AuthError({
 				name: OAUTH_SIGNOUT_EXCEPTION,
-				message:
-					'An error occurred when attempting to log out from OAuth provider.',
+				message: `An error occurred when attempting to log out from OAuth provider. ${error}`,
```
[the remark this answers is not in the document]
Samaritan1011001: This diff seems to be outdated, exposing the error is not in main so I left it as is.

.github/workflows/push-integ-test.yml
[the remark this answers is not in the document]
Samaritan1011001: Merged that PR, so reverted this.

jimblanc (approved):

HuiSF (approved):

PR #13517, opened 2024-06-19
feat(storage): add delimiter option to list API

israx:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -206,27 +216,61 @@ const _listWithPath = async ({
...
+	commonPrefixes?: CommonPrefix[],
+): string[] | undefined {
+	const mappedSubpaths = commonPrefixes?.map(({ Prefix }) => Prefix);
+
+	return mappedSubpaths?.filter(
+		(excludedSubpath): excludedSubpath is string => !!excludedSubpath,
+	);
+}
+
+function parseExcludedSubpaths(excludedSubpaths?: string[]) {
```
ashika112, line 262: why keep this separate? why not just have `mapCommonPrefixesToExcludedSubpaths` return list of string[]

packages/storage/src/types/options.ts
```diff
@@ -10,12 +10,64 @@ export interface StorageOptions {
...
+			 *	console.log(excludedSubpaths)
+			 * ```
+			 */
+			strategy: 'exclude';
+			/**
+			 * Allows to customize the default delimiter (`/`).
+			 *
+			 * @example
+			 * ```ts
+			 *const { excludedSubpaths } = await list({
```
HuiSF, line 59: nit: 
```suggestion
			 * const { excludedSubpaths } = await list({
```

packages/storage/src/types/options.ts
```diff
@@ -10,12 +10,64 @@ export interface StorageOptions {
...
+	  }
+	| {
+			/**
+			 * When passed, the output of the list API will include a list of `excludedSubpaths`
+			 * that are delimited by the `/` character.
+			 *
+			 *
+			 * @example
+			 * ```ts
+			 *const { excludedSubpaths } = await list({
```
HuiSF, line 41: nit:
```suggestion
			 * const { excludedSubpaths } = await list({
```

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -206,27 +216,61 @@ const _listWithPath = async ({
...
+	return excludedSubpaths && excludedSubpaths.length > 0
+		? { excludedSubpaths }
+		: {};
+}
+
+function getDelimiter(
```
HuiSF, line 268: nit: we could keep function expression pattern consistently

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -79,13 +81,15 @@ export const list = async (
 		Prefix: isInputWithPrefix ? `${generatedPrefix}${objectKey}` : objectKey,
 		MaxKeys: options?.listAll ? undefined : options?.pageSize,
 		ContinuationToken: options?.listAll ? undefined : options?.nextToken,
+		Delimiter: getDelimiter(options.subpathStrategy),
```
HuiSF, line 84: What will be the value for the delimiter query parameter when the `getDelimiter()` returns `undefined?`

ashika112: From the looks of it, undefined seems to passed to the request too.

packages/storage/src/providers/s3/apis/internal/list.ts
jimblanc: +1, these always seem to be used together

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -206,27 +216,61 @@ const _listWithPath = async ({
...
+	return excludedSubpaths && excludedSubpaths.length > 0
+		? { excludedSubpaths }
+		: {};
+}
+
+function getDelimiter(
+	subpathStrategy?: StorageSubpathStrategy,
+): string | undefined {
+	if (subpathStrategy?.strategy === 'exclude') {
+		const delimiter = subpathStrategy?.delimiter ?? '/';
```
jimblanc, line 272: Nit, would prefer to move `/` into a constant like `DEFAULT_DELIMITER` to improve maintainability/readability. Also I think you can just return this expression directly.

packages/storage/src/types/options.ts
```diff
@@ -10,12 +10,64 @@ export interface StorageOptions {
...
 	nextToken?: string;
+	subpathStrategy?: StorageSubpathStrategy;
 };
 
 export type StorageRemoveOptions = StorageOptions;
+
+export type StorageSubpathStrategy =
+	| {
+			/**
+			 * When passed, the output of the list API will include all the subpaths in
```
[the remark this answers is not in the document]
jimblanc: +1, maybe something like "Default behavior. Includes all subpaths for a given page in the result."

packages/storage/src/types/options.ts
```diff
@@ -10,12 +10,64 @@ export interface StorageOptions {
...
+			 * the items list.
+			 */
+			strategy: 'include';
+	  }
+	| {
+			/**
+			 * When passed, the output of the list API will include a list of `excludedSubpaths`
+			 * that are delimited by the `/` character.
```
jimblanc, line 36:
```suggestion
			 * When passed, the output of the list API will include a list of `excludedSubpaths`
			 * that are delimited by the `/` (by default) character.
```

packages/storage/src/types/options.ts
```diff
@@ -10,12 +10,64 @@ export interface StorageOptions {
...
+			 *
+			 *
+			 * @example
```
jimblanc, line 39: I'm torn if these examples should go on the API definition for discoverability (particularly on API docs), or if the type is sufficient. Wdyt (I haven't tested how it manifests in the IDE)?

packages/storage/src/types/options.ts
```diff
@@ -10,12 +10,64 @@ export interface StorageOptions {
...
+			 *			}
+			 *		}
+			 *	});
+			 *
+			 *	console.log(excludedSubpaths)
+			 * ```
+			 */
+			strategy: 'exclude';
+			/**
+			 * Allows to customize the default delimiter (`/`).
```
jimblanc, line 55: Nit

```suggestion
			 * Deliminate with with a custom delimiter character.
```

packages/storage/src/providers/s3/apis/internal/list.ts
israx: yeah `undefined` is passed when `subpathStrategy` is undefined or `include`

packages/storage/src/types/options.ts
israx: Leaving these examples on the type should be sufficient. They will manifest when the API is used. 
Screenshot 2024-06-25 at 10 15 28 AM

jimblanc (dismissed):
Looks good, one nit

packages/storage/src/providers/s3/apis/internal/list.ts
HuiSF: Do you have an signed url example for this API call?

packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -206,27 +216,53 @@ const _listWithPath = async ({
...
 };
+
+const mapCommonPrefixesToExcludedSubpaths = (
+	commonPrefixes?: CommonPrefix[],
+): string[] | undefined => {
+	const mappedSubpaths = commonPrefixes?.map(({ Prefix }) => Prefix);
+
+	return mappedSubpaths?.filter(
+		(excludedSubpath): excludedSubpath is string => !!excludedSubpath,
+	);
```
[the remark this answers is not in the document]
HuiSF: nit: `return mappedSubpaths;`

packages/storage/src/providers/s3/apis/internal/list.ts
ashika112: nit: i know its connivence but i think the intend is clear if we add delimiter only when we need it into the `listInputArgs`

israx: It is not included. Bellow attached the examples. Btw the `ListObjectV2` API is already taking `undefined` values.

- `Delimiter` receives an undefined value.
Screenshot 2024-06-25 at 2 37 15 PM

- `Delimiter` receives a value.     
Screenshot 2024-06-25 at 2 41 14 PM 1

haverchuck:
Do we have tests for server side usage?

404ghosting:
WOW Dustin, good catch!

404ghosting:
@ashwinkumar6 please provide your valuable feedback

packages/storage/src/types/options.ts
HuiSF: What if someone doesn't use a ts enabled IDE :D

HuiSF (dismissed):

ashika112 (approved):
Thanks @israx Great work on this :tada:

HuiSF (approved):

packages/storage/src/types/options.ts
israx: our official docs will be the best option then : )

jimblanc (approved):

haverchuck (approved):

PR #13522, opened 2024-06-19
rm "username" from being parsed in amplify_outputs.json `auth.username_attributes`

josefaidt:
#### Description of changes

modifies `parseAmplifyOutputs` to look for an empty array in `auth.username_attributes` instead of `["username"]` to determine whether logging in with username is enabled

this is related to a similar change in the backend repo to remove `username` from the list of `username_attributes` https://github.com/aws-amplify/amplify-backend/pull/1636

valid `username_attributes` properties include `email` and `phone_number` https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cognito.CfnUserPool.html#usernameattributes

these values come from the L1 CfnUserPool property, therefore will not include `username` but the absence of attributes that can be used in place of the username
https://github.com/aws-amplify/amplify-backend/blob/main/packages/auth-construct/src/construct.ts?rgh-link-date=2024-06-19T22%3A39%3A40Z#L952

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/core/src/parseAmplifyOutputs.ts
+++ b/packages/core/src/parseAmplifyOutputs.ts
@@ -141,7 +141,7 @@ function parseAuth(
 			email: username_attributes.includes('email'),
 			phone: username_attributes.includes('phone_number'),
 			// Signing in with a username is not currently supported in Gen2, this should always evaluate to false
-			username: username_attributes.includes('username'),
+			username: username_attributes.length === 0,
 		};
 	}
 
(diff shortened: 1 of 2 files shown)


packages/core/src/parseAmplifyOutputs.ts
```diff
@@ -141,7 +141,7 @@ function parseAuth(
 			email: username_attributes.includes('email'),
 			phone: username_attributes.includes('phone_number'),
 			// Signing in with a username is not currently supported in Gen2, this should always evaluate to false
-			username: username_attributes.includes('username'),
+			username: username_attributes.length === 0,
```
israx, line 144: if the backend is changing, then this fix makes sense. But I wonder if we should also support previous versions to avoid any breaking change.

``` suggestion
			username: username_attributes.length === 0 || username_attributes.includes('username') ,
```

josefaidt: this was written with the outputs schema that had a reference to "username" for this value, however "username" would not be emitted from the construct. for this field, the absence of username attributes indicates "username"

PR #13526, opened 2024-06-20
chore: Add meta-data to core to ensure it is always released

jimblanc:
#### Description of changes

This change introduces a meta-data file to `@aws-amplify/core` which will be automatically updated and committed before releases are run to ensure that a new version of `core` is published with a correct `version.ts` file (even if there were no other changes in `@aws-amplify/core`). Resolves an issue where particular releases of Amplify would use the wrong version in the user-agent headers if there were no changes to `core` in that release. This issue started happening when we fixed conventional commits in https://github.com/aws-amplify/amplify-js/pull/13228. Before this change, all packages were getting published every time which obfuscated the problem.

Downsides:
- More noisy release process as 1 more commit will be generated automatically
- As core will receive a version bump every release, pretty much every other package in the repo will release as most depend on `core` (although this is technically more correct given the current UA versioning logic)

Alternatives considered:
- Force publish `@aws-amplify/core` using `--force-publish`
  - Problem: Disables conventional commits as per Lerna docs & testing.
- Force commit updated `version.ts` before release runs 
  - Problem: Off-by-one error as Lerna hasn't updated the version of `aws-amplify` yet.
- Use a Lerna life-cycle hook to update `version.ts` before release runs
  - Problem: Chicken and egg problem. There is no hook that runs before Lerna determines which packages have changed (and should be released), so even if we update `version.ts` it's too late in the process and `core` would be skipped.
- Import `aws-amplify` version directly from package.json
  - Problem: Bundle bloat and security concerns as package.json would be exposed in customer bundles. Also implicit circular dependency.
- Refactor version logic into `aws-amplify`
  - Problem: This logic is used in packages that are outside of the umbrella package, so this approach would not work universally

PR that introduced original `version.ts` logic: https://github.com/aws-amplify/amplify-js/pull/11301

Tested release workflow on Verdaccio. Confirmed that `version.ts` was published with the correct version.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


ashika112 (approved):

ashwinkumar6 (approved):
Thanks for getting this in !!

PR #13574, opened 2024-07-09
[Multi-bucket] Update options to include bucket

ashika112:
#### Description of changes
- Update the common options to include `bucket`
- Update test cases to test bucket as string and object
- This PR includes changes for all API except Copy which will be followed up separately.

- New unit tests added
- Manual testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/src/singleton/AmplifyOutputs/types.ts
+++ b/packages/core/src/singleton/AmplifyOutputs/types.ts
@@ -46,6 +46,11 @@ export interface AmplifyOutputsAuthProperties {
 export interface AmplifyOutputsStorageProperties {
 	aws_region: string;
 	bucket_name: string;
+	buckets?: {
+		name: string;
+		bucket_name: string;
+		aws_region: string;
+	}[];
 }
 
 export interface AmplifyOutputsGeoProperties {
--- a/packages/core/src/singleton/AmplifyOutputs/types.ts
+++ b/packages/core/src/singleton/AmplifyOutputs/types.ts
@@ -46,7 +46,7 @@ export interface AmplifyOutputsAuthProperties {
 export interface AmplifyOutputsStorageBucketProperties {
 	/** Friendly bucket name provided in Amplify Outputs */
 	name: string;
-	/** Actual s3 bucket name given */
+	/** Actual S3 bucket name given */
 	bucket_name: string;
 	/** Region for the bucket */
 	aws_region: string;
--- a/packages/core/src/singleton/AmplifyOutputs/types.ts
+++ b/packages/core/src/singleton/AmplifyOutputs/types.ts
@@ -44,13 +44,19 @@ export interface AmplifyOutputsAuthProperties {
 }
 
 export interface AmplifyOutputsStorageBucketProperties {
+	/** Friendly bucket name provided in Amplify Outputs */
 	name: string;
+	/** Actual s3 bucket name given */
 	bucket_name: string;
+	/** Region for the bucket */
 	aws_region: string;
 }
 export interface AmplifyOutputsStorageProperties {
+	/** Default region for Storage */
 	aws_region: string;
+	/** Default bucket for Storage */
 	bucket_name: string;
+	/** List of buckets for Storage */
 	buckets?: AmplifyOutputsStorageBucketProperties[];
 }
 
--- a/packages/core/src/singleton/AmplifyOutputs/types.ts
+++ b/packages/core/src/singleton/AmplifyOutputs/types.ts
@@ -43,14 +43,15 @@ export interface AmplifyOutputsAuthProperties {
 	mfa_methods?: string[];
 }
 
+export interface AmplifyOutputsStorageBucketProperties {
+	name: string;
+	bucket_name: string;
+	aws_region: string;
+}
 export interface AmplifyOutputsStorageProperties {
 	aws_region: string;
 	bucket_name: string;
-	buckets?: {
-		name: string;
-		bucket_name: string;
-		aws_region: string;
-	}[];
+	buckets?: AmplifyOutputsStorageBucketProperties[];
 }
 
 export interface AmplifyOutputsGeoProperties {
--- a/packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
+++ b/packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
@@ -368,7 +368,7 @@ describe('downloadData with path', () => {
 		);
 	});
 
-	it('should override bucket in putObject call when bucket is passed in option', async () => {
+	it('should override bucket in getObject call when bucket is passed in option', async () => {
 		(getObject as jest.Mock).mockResolvedValueOnce({ Body: 'body' });
 		const abortController = new AbortController();
 		const bucketInfo: BucketInfo = {
--- a/packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
+++ b/packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
@@ -63,7 +63,7 @@ const mockDownloadResultBase = {
 const mockFetchAuthSession = Amplify.Auth.fetchAuthSession as jest.Mock;
 const mockCreateDownloadTask = createDownloadTask as jest.Mock;
 const mockValidateStorageInput = validateStorageOperationInput as jest.Mock;
-const mockGetConfig = Amplify.getConfig as jest.Mock;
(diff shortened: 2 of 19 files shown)


packages/core/src/singleton/AmplifyOutputs/types.ts
```diff
@@ -44,13 +44,19 @@ export interface AmplifyOutputsAuthProperties {
 }
 
 export interface AmplifyOutputsStorageBucketProperties {
+	/** Friendly bucket name provided in Amplify Outputs */
 	name: string;
+	/** Actual s3 bucket name given */
```
HuiSF, line 49: nit: 
```suggestion
	/** Actual S3 bucket name given */
```

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
```diff
@@ -101,3 +109,25 @@ export const resolveS3ConfigAndInput = async (
...
+			StorageValidationErrorCode.InvalidStorageBucket,
+		);
+
+		return { bucket: bucketConfig.bucketName, region: bucketConfig.region };
+	}
+
+	if (typeof apiOptions.bucket === 'object') {
+		return {
+			bucket: apiOptions.bucket.bucketName,
+			region: apiOptions.bucket.region,
```
HuiSF, line 130: Sanity check: doesn't need to validate the object shape?

packages/storage/src/errors/types/validation.ts
```diff
@@ -70,4 +71,8 @@ export const validationErrorMap: AmplifyErrorMap<StorageValidationErrorCode> = {
 	[StorageValidationErrorCode.InvalidStoragePathInput]: {
 		message: 'Input `path` does not allow a leading slash (/).',
 	},
+	[StorageValidationErrorCode.InvalidStorageBucket]: {
+		message:
+			'Unable to lookup bucket from provided name in Amplify configuration',
```
HuiSF, line 76: nit:
```suggestion
			'Unable to lookup bucket from provided name in Amplify configuration.',
```

packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
```diff
@@ -220,6 +222,70 @@ describe('downloadData with key', () => {
 			}),
 		);
 	});
+
+	describe('bucket passed in options', () => {
```
HuiSF, line 226: Thanks for adding additional tests. 

One thought here is that the bucket info overriding is carried by the function [`resolveS3ConfigAndInput.ts`](https://github.com/aws-amplify/amplify-js/pull/13574/files#diff-2c611974cdca69f851063ca24bbb25068839e263d0a3b4099a7955417f118c70) should we primarily test this function, and verify whether other consumer functions are invoking this function correctly, so we don't need to repeat the same tests in every test suite (as they are basically testing the `resolveS3ConfigAndInput` function)?

HuiSF (commented):
Thanks!

packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
ashika112: So couple of pointers here, i think what called out makes sense.
1. I still have not made changes to COPY api which needs resolution in both source and destination. Per design we might end up extracting `resolveBucketConfig` logic out and make it per API level if something pops out from Copy implement so for now i have it top level since that makes most sense. So i wrote the test per APi level so if we make the change it can catch any changes.

2. We will still need test APi level here but rather in that case will check if API options is passed right. I think both of these achieve the same result in my mind.

That said, with next PR on copy API if this still stays here i can add in test specific to that function to check high level. But i think it is still beneficial per API level as well. wdyt?

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
ashika112: I didnt think we checked object structure like that? Do we do this somewhere. I thought usually type is enough.

We do have assertion already if there is no bucket or region in our code that is checked after this piece of code.

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
```diff
@@ -62,8 +63,16 @@ export const resolveS3ConfigAndInput = async (
...
+	let { bucket, region, dangerouslyConnectToHttpEndpointForTesting, buckets } =
 		amplify.getConfig()?.Storage?.S3 ?? {};
+
+	const resolvedBucketConfig =
+		apiOptions?.bucket && resolveBucketConfig(apiOptions, buckets);
+
+	if (resolvedBucketConfig) {
+		({ bucket, region } = resolvedBucketConfig);
+	}
+
```
israx, line 75: instead of mutating the config object, I think it is better to have a common config resolver and move all of this logic inside. 

```suggestion
const { bucket, region, dangerouslyConnectToHttpEndpointForTesting } = resolveStorageConfig(amplify, apiOptions)
```

ashika112: hmm. This is our extracted config resolver right? why abstract it again. I can update the logic to avoid mutating though but in any case i would need to get the config and use that as a fallback.

packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
jimblanc: Not a big deal, but is there a way to compose this such that the same logic can be re-used between suites?

packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
```diff
@@ -75,6 +76,7 @@ describe('downloadData with key', () => {
 				S3: {
 					bucket,
 					region,
+					buckets: { 'default-bucket': { bucketName: bucket, region } },
```
jimblanc, line 79: Might be covered elsewhere, but is it worth having a test or two to ensure backwards compatibility with old configuration files?

packages/storage/src/errors/types/validation.ts
```diff
@@ -70,4 +71,8 @@ export const validationErrorMap: AmplifyErrorMap<StorageValidationErrorCode> = {
 	[StorageValidationErrorCode.InvalidStoragePathInput]: {
 		message: 'Input `path` does not allow a leading slash (/).',
 	},
+	[StorageValidationErrorCode.InvalidStorageBucket]: {
+		message:
+			'Unable to lookup bucket from provided name in Amplify configuration.',
```
jimblanc, line 76: Nit: Save a few bytes since this gets bundled universally

```suggestion
			'Bucket not found in configuration.',
```

jimblanc (commented):
Looking pretty good to me, few questions & nits

packages/storage/src/errors/types/validation.ts
ashika112: The ask from API BR was to keep it verbose since the whole string friendly name thing is not easily evident.

packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
ashika112: i thought about it so most of our test actually test against just `bucket and region` only multi-bucket tests utilizes this value. as for overall test, in parse logic we have test for with and without buckets in it. Are u thinking about something in specific here?

Or another thing we could do is add this values in only during multi-bucket test suite.

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
israx: We can always keep breaking down internal implementations to make them reusable and more readable.

ashika112: Agree just dont see a huge difference in doing that here, since its already just adding a single line here.

packages/storage/__tests__/providers/s3/apis/downloadData.test.ts
ashika112: Yeah i thought about it but we kinda do this in all test suites. Also, i see value in testing this API level like we are doing other places. let me take another look at this. All of these are testing their input on actual s3 call , so i think even if we compose we still need to pass that part in

jimblanc (approved):

israx (approved):

PR #13590, opened 2024-07-12
feat(storage): integrate locationCredentialsProvider

israx:
#### Description of changes

Integrates the `locationCredentialsProvider` into the Storage APIs

The following changes will throw an exception,
 - if and only if `locationCredentialsProvider` is defined, and
 	- `path` is not a string;
 	- `bucket` is an empty `string` or is not defined;
 - if `bucket` and `region` are not defined from the Amplify singleton or API input

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/storage/src/providers/s3/apis/internal/copy.ts
+++ b/packages/storage/src/providers/s3/apis/internal/copy.ts
@@ -36,15 +36,6 @@ export const copy = async (
 		: copyWithKey(config, input);
 };
 
-export const copyV2 = async (
-	config: S3InternalConfig,
-	input: CopyInput | CopyWithPathInput,
-): Promise<CopyOutput | CopyWithPathOutput> => {
-	return isCopyInputWithPath(input)
-		? copyWithPath(config, input)
-		: copyWithKey(config, input);
-};
-
 const copyWithPath = async (
 	config: S3InternalConfig,
 	input: CopyWithPathInput,
--- a/packages/storage/src/providers/s3/apis/internal/copy.ts
+++ b/packages/storage/src/providers/s3/apis/internal/copy.ts
@@ -1,7 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { AmplifyClassV6 } from '@aws-amplify/core';
 import { StorageAction } from '@aws-amplify/core/internals/utils';
 
 import {
@@ -12,7 +11,6 @@ import {
 } from '../../types';
 import { ResolvedS3Config } from '../../types/options';
 import {
-	createStorageConfiguration,
 	isInputWithPath,
 	resolveS3ConfigAndInput,
 	validateStorageOperationInput,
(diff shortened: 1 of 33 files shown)


packages/storage/src/providers/s3/apis/internal/types/index.ts
```diff
@@ -23,9 +29,25 @@ type S3LibraryOptions = NonNullable<LibraryOptions['Storage']>['S3'];
  *
  * @internal
  */
-export interface S3InternalConfig {
-	serviceOptions: S3ServiceOptions;
-	libraryOptions: S3LibraryOptions;
+export interface S3InternalConfig extends S3ServiceOptions, S3LibraryOptions {
```
jimblanc, line 29: Why flatten this out further? It seems a little more organized to keep `serviceOptions` & `libraryOptions` in their own fields.

packages/storage/src/providers/s3/apis/internal/copy.ts
```diff
@@ -23,25 +21,36 @@ import { copyObject } from '../../utils/client/s3data';
...
 	input: CopyInput | CopyWithPathInput,
 ): Promise<CopyOutput | CopyWithPathOutput> => {
 	return isCopyInputWithPath(input)
-		? copyWithPath(amplify, input)
-		: copyWithKey(amplify, input);
+		? copyWithPath(config, input)
+		: copyWithKey(config, input);
+};
+
+export const copyV2 = async (
```
jimblanc, line 39: What is this? It appears identical to `copy`

packages/storage/src/providers/s3/utils/config.ts
```diff
@@ -44,16 +68,194 @@ const createDefaultIdentityIdProvider = (amplify: AmplifyClassV6) => {
  */
 export const createStorageConfiguration = (
 	amplify: AmplifyClassV6,
+	apiInput: unknown,
+	permission: Permission,
```
jimblanc, line 72: I wonder if `permission` should just be optional so we don't need to pass in "dummy" values for our normal APIs.

packages/storage/src/providers/s3/utils/config.ts
```diff
@@ -44,16 +68,198 @@ const createDefaultIdentityIdProvider = (amplify: AmplifyClassV6) => {
...
+): S3InternalConfig => {
+	const options = getOptionsFromAPIInput(apiInput) ?? {};
+	const { locationCredentialsProvider } = options;
+	const libraryOptions = amplify.libraryOptions?.Storage?.S3 ?? {};
+	const serviceOptions = getServiceOptions(amplify, options);
+	const identityIdProvider = createDefaultIdentityIdProvider(amplify);
+	const { bucket } = serviceOptions;
+
+	const credentialsProvider = locationCredentialsProvider
+		? createCustomCredentialsProvider({
```
jimblanc, line 82: As per the tech design, I was under the impression that consumers would pass in an initialized credential provider with the permission & scope already associated.

packages/storage/src/providers/s3/apis/internal/copy.ts
israx: sorry was trying to merge config and input instead of having to params, but forgot to delete this code.

packages/storage/src/providers/s3/utils/config.ts
israx: The permission is defined per API and scope is resolved from API input

packages/storage/src/providers/s3/utils/config.ts
israx: every Storage API needs to pass a permission to resolve the `locationCredentialsProvider`

packages/storage/src/providers/s3/apis/internal/types/index.ts
israx: This feedback came from @AllanZhengYP, the argument was that `serviceOptions` and `libraryOptions` are related with the configuration from the singleton and would be best just to pass the specific storage configuration. Open to discussion

packages/storage/src/providers/s3/utils/config.ts
jimblanc: Discussed offline, concern resolved.

packages/storage/src/providers/s3/utils/config.ts
ashika112: +1. I dont think we should pass in a value on all API that is not provided by the customer and we can not assume permission for them. we can keep it optional and throw an error from resolve for locationCreds if there is no permission in there. I think that will be better

ashika112:
Closing this since it is no longer valid.

PR #13605, opened 2024-07-17
feat(storage): enables location credentials provider

israx:
#### Description of changes

The following PR enables the location credentials provider on the `resolveS3ConfigAndInput` helper. This helper is used by all of the storage APIs.

##### Main changes:

- Removed `identityId` assertion in the `resolveS3ConfigAndInput` helper  — doubled check with @AllanZhengYP already. The assertion of the `identityId` is not needed by the underlying service APIs. 
- The `credentialsProvider` in the `resolveS3ConfigAndInput` function is asserting for a `deprecated` input or a callback `path` if and only if a `locationCredentialsProvider` is passed in the Storage option inputs.

- unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
+++ b/packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
@@ -9,6 +9,11 @@ import {
 	StorageValidationErrorCode,
 	validationErrorMap,
 } from '../../../../../src/errors/types/validation';
+import {
+	CallbackPathStorageInput,
+	DeprecatedStorageInput,
+} from '../../../../../src/providers/s3/utils/resolveS3ConfigAndInput';
+import { INVALID_STORAGE_INPUT } from '../../../../../src/errors/constants';
 
 jest.mock('@aws-amplify/core', () => ({
 	ConsoleLogger: jest.fn(),
@@ -76,13 +81,11 @@ describe('resolveS3ConfigAndInput', () => {
 		}
 	});
 
-	it('should throw if identityId is not available', async () => {
+	it('should not throw if identityId is not available', async () => {
 		mockFetchAuthSession.mockResolvedValueOnce({
 			credentials,
 		});
-		await expect(resolveS3ConfigAndInput(Amplify, {})).rejects.toMatchObject(
-			validationErrorMap[StorageValidationErrorCode.NoIdentityId],
-		);
+		expect(async () => resolveS3ConfigAndInput(Amplify, {})).not.toThrow();
 	});
 
 	it('should resolve bucket from S3 config', async () => {
@@ -213,5 +216,73 @@ describe('resolveS3ConfigAndInput', () => {
 			targetIdentityId,
 		});
 		expect(keyPrefix).toEqual('prefix');
+	});
+
+	describe('with locationCredentialsProvider', () => {
+		const mockLocationCredentialsProvider = jest
+			.fn()
+			.mockReturnValue({ credentials });
+		it('should resolve credentials without Amplify singleton', async () => {
+			mockGetConfig.mockReturnValue({
+				Storage: {
+					S3: {
+						bucket,
+						region,
+					},
+				},
+			});
+			const { s3Config } = await resolveS3ConfigAndInput(Amplify, {
+				locationCredentialsProvider: mockLocationCredentialsProvider,
+			});
+
+			if (typeof s3Config.credentials === 'function') {
+				const result = await s3Config.credentials();
+				expect(mockLocationCredentialsProvider).toHaveBeenCalled();
+				expect(result).toEqual(credentials);
+			} else {
+				throw new Error('Expect credentials to be a function');
+			}
+		});
+
+		describe('with deprecated or callback paths as inputs', () => {
+			const key = 'mock-value';
+			const prefix = 'mock-value';
+			const path = () => 'path';
+			const deprecatedInputs: DeprecatedStorageInput[] = [
+				{ prefix },
+				{ key },
+				{
+					source: { key },
+					destination: { key },
+				},
+			];
+			const callbackPathInputs: CallbackPathStorageInput[] = [
+				{ path },
+				{
+					destination: { path },
+					source: { path },
+				},
+			];
+
+			const testCases = [...deprecatedInputs, ...callbackPathInputs];
+
+			it.each(testCases)('should throw when input is %s', async input => {
+				const { s3Config } = await resolveS3ConfigAndInput(
+					Amplify,
+					{ locationCredentialsProvider: mockLocationCredentialsProvider },
+					input,
+				);
+				if (typeof s3Config.credentials === 'function') {
+					await expect(s3Config.credentials()).rejects.toThrow(
+						expect.objectContaining({
+							name: INVALID_STORAGE_INPUT,
+							message: 'The storage input needs to pass a string path.',
+						}),
+					);
+				} else {
+					throw new Error('Expect credentials to be a function');
+				}
+			});
+		});
 	});
 });
--- a/packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
+++ b/packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
@@ -182,7 +182,7 @@ describe('resolveS3ConfigAndInput', () => {
 	it('should resolve prefix with given access level', async () => {
 		mockDefaultResolvePrefix.mockResolvedValueOnce('prefix');
 		const { keyPrefix } = await resolveS3ConfigAndInput(Amplify, {
-			accessLevel: 'someLevel' as any,
+			options: { accessLevel: 'someLevel' as any },
 		});
 		expect(mockDefaultResolvePrefix).toHaveBeenCalledWith({
 			accessLevel: 'someLevel',
@@ -232,7 +232,9 @@ describe('resolveS3ConfigAndInput', () => {
 				},
 			});
 			const { s3Config } = await resolveS3ConfigAndInput(Amplify, {
-				locationCredentialsProvider: mockLocationCredentialsProvider,
+				options: {
+					locationCredentialsProvider: mockLocationCredentialsProvider,
+				},
 			});
 
 			if (typeof s3Config.credentials === 'function') {
@@ -267,18 +269,16 @@ describe('resolveS3ConfigAndInput', () => {
 			const testCases = [...deprecatedInputs, ...callbackPathInputs];
 
 			it.each(testCases)('should throw when input is %s', async input => {
-				const { s3Config } = await resolveS3ConfigAndInput(
-					Amplify,
-					{ locationCredentialsProvider: mockLocationCredentialsProvider },
-					input,
-				);
+				const { s3Config } = await resolveS3ConfigAndInput(Amplify, {
+					...input,
+					options: {
+						locationCredentialsProvider: mockLocationCredentialsProvider,
+					},
+				});
 				if (typeof s3Config.credentials === 'function') {
 					await expect(s3Config.credentials()).rejects.toThrow(
 						expect.objectContaining({
 							name: INVALID_STORAGE_INPUT,
-							message: 'The input needs to have a path as a string value.',
-							recoverySuggestion:
-								'Please provide a valid path as a string value for the input.',
 						}),
 					);
 				} else {
--- a/packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
+++ b/packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
@@ -246,6 +246,23 @@ describe('resolveS3ConfigAndInput', () => {
 			}
 		});
 
+		it('should not throw when path is pass as a string', async () => {
+			const { s3Config } = await resolveS3ConfigAndInput(Amplify, {
+				path: 'my-path',
+				options: {
+					locationCredentialsProvider: mockLocationCredentialsProvider,
+				},
+			});
+
+			if (typeof s3Config.credentials === 'function') {
+				const result = await s3Config.credentials();
+				expect(mockLocationCredentialsProvider).toHaveBeenCalled();
+				expect(result).toEqual(credentials);
+			} else {
+				throw new Error('Expect credentials to be a function');
+			}
+		});
+
 		describe('with deprecated or callback paths as inputs', () => {
 			const key = 'mock-value';
 			const prefix = 'mock-value';
--- a/packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
+++ b/packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
@@ -276,7 +276,9 @@ describe('resolveS3ConfigAndInput', () => {
 					await expect(s3Config.credentials()).rejects.toThrow(
 						expect.objectContaining({
 							name: INVALID_STORAGE_INPUT,
-							message: 'The storage input needs to pass a string path.',
+							message: 'The input needs to have a path as a string value.',
+							recoverySuggestion:
+								'Please provide a valid path as a string value for the input.',
 						}),
 					);
 				} else {
--- a/packages/storage/src/providers/s3/apis/downloadData.ts
+++ b/packages/storage/src/providers/s3/apis/downloadData.ts
@@ -115,7 +115,11 @@ const downloadDataJob =
 	> => {
 		const { options: downloadDataOptions } = downloadDataInput;
 		const { bucket, keyPrefix, s3Config, identityId } =
-			await resolveS3ConfigAndInput(Amplify, downloadDataOptions);
+			await resolveS3ConfigAndInput(
+				Amplify,
+				downloadDataOptions,
+				downloadDataInput as DownloadDataInput,
+			);
 		const { inputType, objectKey } = validateStorageOperationInput(
 			downloadDataInput,
 			identityId,
--- a/packages/storage/src/providers/s3/apis/downloadData.ts
+++ b/packages/storage/src/providers/s3/apis/downloadData.ts
@@ -118,7 +118,7 @@ const downloadDataJob =
 			await resolveS3ConfigAndInput(
 				Amplify,
 				downloadDataOptions,
-				downloadDataInput as DownloadDataInput,
+				downloadDataInput,
 			);
 		const { inputType, objectKey } = validateStorageOperationInput(
 			downloadDataInput,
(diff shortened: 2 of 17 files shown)


packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
```diff
@@ -76,13 +81,11 @@ describe('resolveS3ConfigAndInput', () => {
 		}
 	});
 
-	it('should throw if identityId is not available', async () => {
+	it('should not throw if identityId is not available', async () => {
```
israx, line 84: The `resolveS3ConfingAndInput` helper is not asserting for the identityId anymore

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
```diff
@@ -38,13 +59,13 @@ interface ResolvedS3ConfigAndInput {
...
 	amplify: AmplifyClassV6,
 	apiOptions?: S3ApiOptions,
+	input?: DeprecatedStorageInput | CallbackPathStorageInput,
 ): Promise<ResolvedS3ConfigAndInput> => {
 	/**
 	 * IdentityId is always cached in memory so we can safely make calls here. It
 	 * should be stable even for unauthenticated users, regardless of credentials.
 	 */
 	const { identityId } = await amplify.Auth.fetchAuthSession();
-	assertValidationError(!!identityId, StorageValidationErrorCode.NoIdentityId);
```
AllanZhengYP, line 47: We confirmed that this validation is extra by validating every reference of identity id. The identity id is modeled as optional through the codebase. When the identity id is required, like prefix resolver, we have the same assertion throwing same error.

AllanZhengYP (commented):
It looks good to me, just a couple nits.

This approach makes sense to me. If we cannot prevent complexity, I'd rather contain the complexity inside this function and refactor it later, than scatter it in multiple modules.

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
```diff
@@ -38,13 +59,13 @@ interface ResolvedS3ConfigAndInput {
 export const resolveS3ConfigAndInput = async (
 	amplify: AmplifyClassV6,
 	apiOptions?: S3ApiOptions,
+	input?: DeprecatedStorageInput | CallbackPathStorageInput,
```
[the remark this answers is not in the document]
israx: would like to leave it as `input`. I have other functions that make reference to the `input` param

packages/storage/src/providers/s3/types/inputs.ts
```diff
@@ -35,19 +35,24 @@ import {
 	UploadDataOptionsWithPath,
 } from '../types';
 
+import { CommonOptions } from './options';
+
 // TODO: support use accelerate endpoint option
```
AllanZhengYP, line 40: Since we don't have people asking for this option, it makes sense to keep leaving it out of the `options`.

AllanZhengYP (dismissed):

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
ashika112: @AllanZhengYP will that be the case in Path callback as well? if they are using identityId and it is not preset they will have an error thrown out?

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
```diff
@@ -101,3 +128,70 @@ export const resolveS3ConfigAndInput = async (
...
+			typeof (input as StorageOperationInputWithPath).path === 'function') ||
+		((input as StorageCopyInputWithPath)?.destination?.path &&
+			typeof (input as StorageCopyInputWithPath).destination?.path ===
+				'function') ||
+		((input as StorageCopyInputWithPath)?.source?.path &&
+			typeof (input as StorageCopyInputWithPath).source?.path === 'function')
+	);
+};
+
+const isDeprecatedInput = (
```
ashika112, line 152: name is misleading since callbackPath is not deprecated

packages/storage/src/providers/s3/apis/internal/remove.ts
```diff
@@ -25,7 +25,7 @@ export const remove = async (
 ): Promise<RemoveOutput | RemoveWithPathOutput> => {
 	const { options = {} } = input ?? {};
 	const { s3Config, keyPrefix, bucket, identityId } =
-		await resolveS3ConfigAndInput(amplify, options);
+		await resolveS3ConfigAndInput(amplify, options, input as RemoveInput);
```
ashika112, line 28: why are we typecasting here? shouldn't the input take both?

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
```diff
@@ -101,3 +128,70 @@ export const resolveS3ConfigAndInput = async (
...
+	return (
+		isInputWithKey(input) ||
+		isInputWithPrefix(input) ||
+		isInputWithCopySourceOrDestination(input)
+	);
+};
+const assertStorageInput = (
+	input?: DeprecatedStorageInput | CallbackPathStorageInput,
+) => {
+	if (isDeprecatedInput(input) || isInputWithCallbackPath(input)) {
```
ashika112, line 164: why not use assertValidationError here?

packages/storage/src/types/inputs.ts
```diff
@@ -85,12 +85,14 @@ export type StorageUploadDataInputWithPath<Options> =
 export interface StorageCopyInputWithKey<
 	SourceOptions extends StorageOptions,
 	DestinationOptions extends StorageOptions,
-> {
+	Options,
+> extends StorageOperationOptionsInput<Options> {
```
ashika112, line 89: can we just add `options: Options` here

packages/storage/src/types/inputs.ts
```diff
@@ -85,12 +85,14 @@ export type StorageUploadDataInputWithPath<Options> =
...
 	source: SourceOptions;
 	destination: DestinationOptions;
 }
 
-export interface StorageCopyInputWithPath {
+export interface StorageCopyInputWithPath<Options = unknown>
```
ashika112, line 94: why set it to unknown?

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
```diff
@@ -101,3 +128,70 @@ export const resolveS3ConfigAndInput = async (
...
+		throw new StorageError({
+			name: INVALID_STORAGE_INPUT,
+			message: 'The storage input needs to pass a string path.',
+		});
+	}
+};
+
+const isInputWithKey = (
```
ashika112, line 172: nit: should we move these to separate utils and test them out separately?

packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
ashika112: while this might not be needed, if customers are already relyign on this message for something, would removing it cause a breaking change for them?

packages/storage/src/providers/s3/apis/internal/remove.ts
jimblanc: +1, can we get by without the cast?

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
```diff
@@ -101,3 +128,45 @@ export const resolveS3ConfigAndInput = async (
...
+	return !!options?.locationCredentialsProvider;
+};
+
+const isInputWithCallbackPath = (input?: CallbackPathStorageInput) => {
+	return (
+		((input as StorageOperationInputWithPath)?.path &&
+			typeof (input as StorageOperationInputWithPath).path === 'function') ||
+		((input as StorageCopyInputWithPath)?.destination?.path &&
+			typeof (input as StorageCopyInputWithPath).destination?.path ===
+				'function') ||
```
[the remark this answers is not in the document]
jimblanc: Agreed, this did end up getting implemented? See the feedback was resolved

jimblanc (commented):
Overall looks good! Agree with some of the feedback from other reviewers (left comments)

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
AllanZhengYP: Oh this is a good point. The path callback's input identity id is optional. However, since we used to validate identity id here, the path callbacks's input identity id is never actually optional although it's typed as optional. By removing the validation here, we probably should validate identity id in `validateStorageOptionalInput()`:https://github.com/aws-amplify/amplify-js/blob/58b86bb8930061d406fe0f03a7d335c4fb40060a/packages/storage/src/providers/s3/utils/validateStorageOperationInput.ts#L25

packages/storage/src/types/inputs.ts
AllanZhengYP: +1 I think this is the better approach.

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
ashika112: yes we made it named param since we might need to make way for additional values here but not sure why we made it optional. Now that we are here more , i think it shouldn't have been optional 😓

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
```diff
@@ -23,6 +36,14 @@ interface ResolvedS3ConfigAndInput {
...
 	identityId?: string;
 }
+export type DeprecatedStorageInput =
+	| StorageOperationInputWithKey
+	| StorageOperationInputWithPrefix
+	| CopyInput;
+
+export type CallbackPathStorageInput =
+	| StorageOperationInputWithPath
+	| StorageCopyInputWithPath;
```
Samaritan1011001, line 46: Should this be `CopyWithPathInput`?

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
```diff
@@ -101,3 +128,70 @@ export const resolveS3ConfigAndInput = async (
...
+		isInputWithCopySourceOrDestination(input)
+	);
+};
+const assertStorageInput = (
+	input?: DeprecatedStorageInput | CallbackPathStorageInput,
+) => {
+	if (isDeprecatedInput(input) || isInputWithCallbackPath(input)) {
+		throw new StorageError({
+			name: INVALID_STORAGE_INPUT,
+			message: 'The storage input needs to pass a string path.',
```
Samaritan1011001, line 167: Nit: Since it can be either string or function, this message can be updated a bit or have it generic with a specific recoverySuggestion.

Samaritan1011001 (commented):
Looks good, few questions. Thanks Israel!

packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
AllanZhengYP: @ashika112 I don't think removing the validation here itself is a breaking change, unless they are using Storage APIs intentionally to check for identity ID. However, it's not the intended usage of the Storage APIs.

ashika112: Oh i mean , if customer lets say is specifically finding this error to avoid making a call right now, wouldn't it break? this is propogated all the way to customer. Again coming from path callback perspective, but if this check is moved to path validation i think it is fine there. but wondering if there is any other use case like that

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
israx: The `assertValidationError` uses a map and enum, expanding any of those will increase the bundle size. That is why I would like to avoid using it

packages/storage/src/types/inputs.ts
israx: The `StorageOperationOptionsInput` is already used by other inputs. Want to keep consistency

Screenshot 2024-07-19 at 3 04 03 PM

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
israx: The 	`StorageCopyInputWithPath` type is a type that we are alredy exposing

packages/storage/src/types/inputs.ts
israx: The `StorageCopyInputWithPath` doesn't extend the `CommonOptions` as the other Inputs, hence it doesn't have any constrains. The `unknown` type is a placeholder that allows to use the `StorageCopyInputWithPath` type without providing a generic type.

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
israx: These helpers are mainly used to test `deprecated` inputs, and I added a test that will test all the possible permutations.

packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
```diff
@@ -214,4 +217,74 @@ describe('resolveS3ConfigAndInput', () => {
...
+
+			it.each(testCases)('should throw when input is %s', async input => {
+				const { s3Config } = await resolveS3ConfigAndInput(
+					Amplify,
+					{ locationCredentialsProvider: mockLocationCredentialsProvider },
+					input,
+				);
+				if (typeof s3Config.credentials === 'function') {
+					await expect(s3Config.credentials()).rejects.toThrow(
+						expect.objectContaining({
```
ashika112, line 277: nit:we could just check for name here since if we make changes to message and recovery in future we need to come update it here. that is not better pratice

packages/storage/src/providers/s3/apis/downloadData.ts
```diff
@@ -115,7 +115,11 @@ const downloadDataJob =
 	> => {
 		const { options: downloadDataOptions } = downloadDataInput;
 		const { bucket, keyPrefix, s3Config, identityId } =
-			await resolveS3ConfigAndInput(Amplify, downloadDataOptions);
+			await resolveS3ConfigAndInput(
```
ashika112, line 118: General comment: can be sep PR - should we just destructure the options inside `resolveS3ConfigAndInput` since we are passing whole input now

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
ashika112: Yes but any validation we already perform it through the assertValidation why change the pattern here and now?

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
ashika112: Hmmm i notice the test so the test right now only covers the happy path cases what if the value passed is invalid is not covered. I dont mind this staying here but we should cover all branches in tests so we dont introduce bugs in future

packages/storage/src/types/inputs.ts
ashika112: hmm also copy with key shouldn't need to add an options right?

packages/storage/src/types/inputs.ts
ashika112: Neither does `copywithkey` sorry im confused. I still dont see the need here, pass a generic and use it. if need arises we can do this.

ashika112 (dismissed):
My other comments are non-blocking and ok to approve here. Lets try to follow up the other ones or refute them. thanks

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
AllanZhengYP: I don't think we need to export util functions just to unit test it. However it makes sense to have the unit test of `resolveS3ConfigAndInput` to cover these code paths dealing with deprecated inputs.

packages/storage/src/types/inputs.ts
israx: removed the extra type

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
israx: @ashika112 — The `resolveS3ConfigAndInput` test is also testing negative cases, specially when `locationCredentialsProvider` is passed and the input is invalid

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
```diff
@@ -23,6 +50,30 @@ interface ResolvedS3ConfigAndInput {
...
+	| ListPaginateInput
+	| GetPropertiesInput
+	| CopyInput;
+
+export type CallbackPathStorageInput =
+	| StorageOperationInputWithPath
+	| StorageCopyInputWithPath;
+
+type StorageInput =
+	| DeprecatedStorageInput
```
ashika112, line 69: I think types here can be simplified. DMed u on this

packages/storage/src/providers/s3/utils/resolveS3ConfigAndInput.ts
ashika112: Replied on the tests on my question

packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
```diff
@@ -214,4 +217,74 @@ describe('resolveS3ConfigAndInput', () => {
...
+			if (typeof s3Config.credentials === 'function') {
+				const result = await s3Config.credentials();
+				expect(mockLocationCredentialsProvider).toHaveBeenCalled();
+				expect(result).toEqual(credentials);
+			} else {
+				throw new Error('Expect credentials to be a function');
+			}
+		});
+
+		describe('with deprecated or callback paths as inputs', () => {
```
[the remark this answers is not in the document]
israx: `path` is independent when `locationCredentialsProvider` is called.

ashika112 (approved):

packages/storage/__tests__/providers/s3/apis/utils/resolveS3ConfigAndInput.test.ts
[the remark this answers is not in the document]
ashika112: yes but it should not throw here tho right? i think we should have a test exercising that route

ashika112 (approved):

AllanZhengYP (approved):

PR #13606, opened 2024-07-17
fix(storage): omit subPathStrategy when prefix is defined

israx:
#### Description of changes

The `subpathStrategy` option is currently added to the `prefix` input of the `list`, which is deprecated already. This change is excluding the `subpathStrategy` key when the `list` API has `prefix` as part of the input.

- smoke test
***before***:
Screenshot 2024-07-17 at 3 31 31 PM

***after***:
Screenshot 2024-07-17 at 3 38 35 PM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/storage/src/providers/s3/apis/internal/list.ts
+++ b/packages/storage/src/providers/s3/apis/internal/list.ts
@@ -20,7 +20,11 @@ import {
 	resolveS3ConfigAndInput,
 	validateStorageOperationInputWithPrefix,
 } from '../../utils';
-import { ResolvedS3Config } from '../../types/options';
+import {
+	ListAllOptionsWithPath,
+	ListPaginateOptionsWithPath,
+	ResolvedS3Config,
+} from '../../types/options';
 import {
 	ListObjectsV2Input,
 	ListObjectsV2Output,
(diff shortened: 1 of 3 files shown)


packages/storage/src/providers/s3/apis/internal/list.ts
```diff
@@ -81,7 +81,9 @@ export const list = async (
 		Prefix: isInputWithPrefix ? `${generatedPrefix}${objectKey}` : objectKey,
 		MaxKeys: options?.listAll ? undefined : options?.pageSize,
 		ContinuationToken: options?.listAll ? undefined : options?.nextToken,
-		Delimiter: getDelimiter(options.subpathStrategy),
+		Delimiter: getDelimiter(
+			(options as ListAllWithPathInput['options'])?.subpathStrategy,
```
israx, line 85: The service `Delimiter` option is shared for both `key` and `path` options. This is fine as `getDelimiter` will resolve to `undefined` if it is not resolved, and don't being passed to the server – This was currently [discussed](https://github.com/aws-amplify/amplify-js/pull/13517#discussion_r1649483125).

ashika112: rather than casting i would make use of `isInputWithPrefix` or `isInputWithPath` here and use that to narrow down the type and make the call

ashika112:
The PR is also missing removing the `excludedSubpaths` from output types

packages/storage/src/providers/s3/apis/internal/list.ts
israx: A type guard here will just create extra code that is not needed. It'd make sense to add it if we are conditionally rendering something or if `getDelimiter` is accepting a `Truthy` value. 

I expanded the `getDelimiter` param type to accept `options` instead. This removes the need to cast the `subpathStrategy`

ashwinkumar6 (approved):

ashika112 (approved):

HuiSF (approved):

PR #13607, opened 2024-07-17
[Multi-Bucket] Update Copy API

ashika112:
#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/storage/src/providers/s3/apis/internal/copy.ts
+++ b/packages/storage/src/providers/s3/apis/internal/copy.ts
@@ -10,7 +10,7 @@ import {
 	CopyWithPathInput,
 	CopyWithPathOutput,
 } from '../../types';
-import { ResolvedS3Config } from '../../types/options';
+import { ResolvedS3Config, StorageBucket } from '../../types/options';
 import {
 	isInputWithPath,
 	resolveS3ConfigAndInput,
@@ -25,6 +25,18 @@ import { logger } from '../../../../utils';
 const isCopyInputWithPath = (
 	input: CopyInput | CopyWithPathInput,
 ): input is CopyWithPathInput => isInputWithPath(input.source);
+
+const storageBucketAssertion = (
+	sourceBucket?: StorageBucket,
+	destBucket?: StorageBucket,
+) =>
+	// Throw assertion error when either one of bucket options is empty
+	{
+		assertValidationError(
+			!!(sourceBucket && destBucket) || !!(!destBucket && !sourceBucket),
+			StorageValidationErrorCode.InvalidCopyOperationStorageBucket,
+		);
+	};
 
 export const copy = async (
 	amplify: AmplifyClassV6,
@@ -41,12 +53,7 @@ const copyWithPath = async (
 ): Promise<CopyWithPathOutput> => {
 	const { source, destination } = input;
 
-	// Throw assertion error when either one of bucket options is empty
-	assertValidationError(
-		!!(source.bucket && destination.bucket) ||
-			!!(!destination.bucket && !source.bucket),
-		StorageValidationErrorCode.InvalidCopyOperationStorageBucket,
-	);
+	storageBucketAssertion(source.bucket, destination.bucket);
 
 	const { bucket: sourceBucket, identityId } = await resolveS3ConfigAndInput(
 		amplify,
@@ -94,12 +101,7 @@ export const copyWithKey = async (
 ): Promise<CopyOutput> => {
 	const { source, destination } = input;
 
-	// Throw assertion error when either one of bucket options is empty
-	assertValidationError(
-		!!(source.bucket && destination.bucket) ||
-			!!(!destination.bucket && !source.bucket),
-		StorageValidationErrorCode.InvalidCopyOperationStorageBucket,
-	);
+	storageBucketAssertion(source.bucket, destination.bucket);
 
 	assertValidationError(!!source.key, StorageValidationErrorCode.NoSourceKey);
 	assertValidationError(
--- a/packages/storage/src/providers/s3/apis/internal/copy.ts
+++ b/packages/storage/src/providers/s3/apis/internal/copy.ts
@@ -43,19 +43,19 @@ const copyWithPath = async (
 
 	// Throw assertion error when either one of bucket options is empty
 	assertValidationError(
-		!!(source.options?.bucket && destination.options?.bucket) ||
-			!!(!destination.options?.bucket && !source.options?.bucket),
+		!!(source.bucket && destination.bucket) ||
+			!!(!destination.bucket && !source.bucket),
 		StorageValidationErrorCode.InvalidCopyOperationStorageBucket,
 	);
 
 	const { bucket: sourceBucket, identityId } = await resolveS3ConfigAndInput(
 		amplify,
-		input.source.options,
+		input.source,
 	);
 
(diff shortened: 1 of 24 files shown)


packages/storage/src/providers/s3/types/inputs.ts
```diff
@@ -20,6 +20,8 @@ import {
 import {
 	CopyDestinationOptionsWithKey,
```
ashika112, line 21: all these internal names for options need to be changed to match input and output naming type i will open a PR directly against main for it. I just named them right for ones im adding

packages/storage/src/providers/s3/apis/internal/copy.ts
```diff
@@ -26,6 +26,18 @@ const isCopyInputWithPath = (
...
 ): input is CopyWithPathInput => isInputWithPath(input.source);
 
+const storageBucketAssertion = (
+	sourceBucket?: StorageBucket,
+	destBucket?: StorageBucket,
+) =>
+	// Throw assertion error when either one of bucket options is empty
+	{
+		assertValidationError(
+			!!(sourceBucket && destBucket) || !!(!destBucket && !sourceBucket),
```
Samaritan1011001, line 36: Will it throw if both are empty?

packages/storage/src/providers/s3/apis/internal/copy.ts
```diff
@@ -26,6 +26,18 @@ const isCopyInputWithPath = (
 	input: CopyInput | CopyWithPathInput,
 ): input is CopyWithPathInput => isInputWithPath(input.source);
 
+const storageBucketAssertion = (
+	sourceBucket?: StorageBucket,
+	destBucket?: StorageBucket,
+) =>
+	// Throw assertion error when either one of bucket options is empty
```
Samaritan1011001, line 33: By bucket options do you just mean the bucket itself or is it a miss that the options are not accessed in this validation logic?

packages/storage/src/providers/s3/apis/internal/copy.ts
```diff
@@ -58,14 +80,14 @@ const copyWithPath = async (
...
-	const finalCopySource = `${bucket}/${sourcePath}`;
+	const finalCopySource = `${sourceBucket}/${sourcePath}`;
 	const finalCopyDestination = destinationPath;
 	logger.debug(`copying "${finalCopySource}" to "${finalCopyDestination}".`);
 
 	await serviceCopy({
 		source: finalCopySource,
 		destination: finalCopyDestination,
-		bucket,
+		bucket: destBucket,
```
Samaritan1011001, line 90: should this be the `destBucket`? 

Previously, the bucket value from `resolveS3ConfigAndInput` was used to form the `finalCopySource` and also provide it here but now, `finalCopySource` has `sourceBucket` and this has `destBucket`.

Samaritan1011001 (commented):
Thank you, had a few questions.

packages/storage/src/providers/s3/apis/internal/copy.ts
ashika112: bucket as input option for source and destination.

packages/storage/src/providers/s3/apis/internal/copy.ts
ashika112: it shoud not throw when both are empty or both are present basically. It should throw only when one of source or destination has an option

ashika112: this is by design since we dont want to assume for customer leading to data leaking

packages/storage/src/providers/s3/apis/internal/copy.ts
ashika112: Yes so S3 expectation is we provide the value to which bucket u like to copy. Previously we only had one bucket so it did not matter but with multi bucket . Customer can provide source and destination to be different hence in this case we want to copy it over to destination

ashika112: Ref: https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html#API_CopyObject_RequestSyntax

AllanZhengYP (approved):

packages/storage/src/providers/s3/apis/internal/copy.ts
HuiSF, line 36: Why do we need `!!`, doesn't `&&` and `!` evaluates booleans already?

packages/storage/src/providers/s3/apis/internal/copy.ts
```diff
@@ -40,8 +52,18 @@ const copyWithPath = async (
...
+
+	const { bucket: sourceBucket, identityId } = await resolveS3ConfigAndInput(
+		amplify,
+		input.source,
+	);
+
+	const { s3Config, bucket: destBucket } = await resolveS3ConfigAndInput(
+		amplify,
+		input.destination,
+	);
```
HuiSF, line 66: nit: these two async calls can be parallelized correct?

packages/storage/src/providers/s3/apis/internal/copy.ts
```diff
@@ -40,8 +52,18 @@ const copyWithPath = async (
...
+	storageBucketAssertion(source.bucket, destination.bucket);
+
+	const { bucket: sourceBucket, identityId } = await resolveS3ConfigAndInput(
+		amplify,
+		input.source,
+	);
+
+	const { s3Config, bucket: destBucket } = await resolveS3ConfigAndInput(
+		amplify,
+		input.destination,
```
jjarvisp, line 65: since `source` and `destination` are already destructured from `input` above, any reason not to use those variables for consistency?

packages/storage/src/providers/s3/apis/internal/copy.ts
ashika112: Yes. We do the same already in copywithKey and when called twice no extra network calls are made

packages/storage/src/providers/s3/apis/internal/copy.ts
ashika112: Yeah I think so. I added it since everywhere in storage we seem to be doing that. I thought it was for more readable pattern

ashika112: Oh never mind was checking, `!!(sourceBucket && destBucket)` this part still needs !! or else it resolves to storageBucket

ashika112: Please let me know if im missing something.

packages/storage/src/providers/s3/apis/internal/copy.ts
ashika112: Just missed it with while copying over

packages/storage/src/providers/s3/apis/internal/copy.ts
HuiSF: I see, so the assertion is for check whether these vars hold strings? If that, for better readability around checking multiples string vars at the same time, I think making it a full expression is better with "!== undefined". 😅

ashika112: actuaklly yeah that will read better. ll update thanks

AllanZhengYP (approved):

HuiSF (approved):

PR #13637, opened 2024-07-24
fix redirect issue on Android

geovannifuentesQred:
#### Description of changes
Allow the ability to skip the redirect url validation and use the first redirect url instead.

I enabled this for Android and I could verify that uses the first redirect url and skips http/https validation.
This is needed since Chrome Custom Tabs does not allow to redirect back into the app.
Our solution includes a custom web page that redirects back to the app automatically and we need to setup both redirect urls.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
+++ b/packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { Platform } from 'react-native';
 import { Amplify, OAuthConfig } from '@aws-amplify/core';
 import {
 	AuthAction,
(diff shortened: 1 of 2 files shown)


Samaritan1011001:
Hi, thank you for opening this PR. I would like to understand the use-case a little more. 

> This is needed since Chrome Custom Tabs does not allow to redirect back into the app.
From my understanding you should be able to redirect back to the app. 
1. Are you saying it's a Custom Tabs limitation? 
2. Are you using web links or Android app links? 

> Our solution includes a custom web page that redirects back to the app automatically and we need to setup both redirect urls.

Can you expand on this? Why would a custom web page be needed in the middle?

geovannifuentesQred:
Hi @Samaritan1011001, 
We were trying to use both Web Links and App Links but there's a restriction in the Amplify SDK for https as redirect urls.
Using a custom web page allows us to redirect back to app since we are using Signicat and the redirect they are doing is based on form action, and the redirect fails because of Content Security Policy. This only happens on Chrome (Custom Tabs), not in other browsers like Firefox.
Ideally we would like to use App Links but we are restricted by the SDK right now.

soberm:
Hi @geovannifuentesQred,

Thank you for taking the time to submit this PR. I think this is already possible with the changes introduced in https://github.com/aws-amplify/amplify-js/pull/14448. Hence, I will be closing this PR.

Issue #12849, opened 2024-01-13
Unable to override oauth domain in v6 using documented method


alex-breen:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication

### Amplify Version

v6

### Amplify Categories

auth

### Backend

Amplify CLI

### Environment information

<details>

```
# Put output below this line
System:
    OS: Windows 11 10.0.22621
    CPU: (6) x64 Intel(R) Core(TM) i5-9600K CPU @ 3.70GHz
    Memory: 4.09 GB / 15.94 GB
  Binaries:
    Node: 18.18.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.1.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 120.0.6099.216
    Edge: Chromium (120.0.2210.133)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @aws-amplify/ui-react: ^6.1.1 => 6.1.1 
    @aws-amplify/ui-react-internal:  undefined ()
    @aws-amplify/ui-react-storage: ^3.0.10 => 3.0.10 
    @types/react: ^18.2.43 => 18.2.47 
    @types/react-dom: ^18.2.17 => 18.2.18 
    @vitejs/plugin-react: ^4.2.1 => 4.2.1 
    aws-amplify: ^6.0.12 => 6.0.12 
    aws-amplify/adapter-core:  undefined ()
    aws-amplify/analytics:  undefined ()
    aws-amplify/analytics/kinesis:  undefined ()
    aws-amplify/analytics/kinesis-firehose:  undefined ()
    aws-amplify/analytics/personalize:  undefined ()
    aws-amplify/analytics/pinpoint:  undefined ()
    aws-amplify/api:  undefined ()
    aws-amplify/api/server:  undefined ()
    aws-amplify/auth:  undefined ()
    aws-amplify/auth/cognito:  undefined ()
[...]

PR #13642, opened 2024-07-24
fix(singleton): allow oauth listener to be called with the latest config

israx:
#### Description of changes

`Amplify.configure` will call `oAuthListener` and it will pass the Auth config. This `oAuthListener` is responsable for completing an OAuth flow. However when `Amplify.configure` is called more than 1 time, the `oAuthListener` will be called only with the first Auth config.

This fix makes sure the `oAuthListener` is called only and only if a valid OAuth `config` and `oAuthListener` are already defined. And ensures the `oAuthListener` is called with the latest config of Amplify.

- Before
`Amplify.configure` with **config1** -> `Amplify.configure` with config2 -> call `oAuthListener` with **config1**

- After
`Amplify.configure` with config1 -> `Amplify.configure` with **config2** -> call `oAuthListener` with **config2**

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/12849

- smoke testing
- unit tests

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/core/src/singleton/Amplify.ts
+++ b/packages/core/src/singleton/Amplify.ts
@@ -13,14 +13,14 @@ import {
 } from './types';
 import { AuthClass } from './Auth';
 import { ADD_OAUTH_LISTENER } from './constants';
+import { OAuthListener } from './Auth/types';
 
 export class AmplifyClass {
-	private oAuthListener:
-		| ((authConfig: AuthConfig['Cognito']) => void)
-		| undefined = undefined;
+	private oAuthListener: OAuthListener | undefined = undefined;
 
 	resourcesConfig: ResourcesConfig;
 	libraryOptions: LibraryOptions;
+	private listenerTimeout?: ReturnType<typeof setTimeout>;
 
 	/**
 	 * Cross-category Auth utilities.
@@ -53,7 +53,6 @@ export class AmplifyClass {
 		libraryOptions?: LibraryOptions,
 	): void {
 		const resolvedResourceConfig = parseAmplifyConfig(resourcesConfig);
-
 		this.resourcesConfig = resolvedResourceConfig;
 
 		if (libraryOptions) {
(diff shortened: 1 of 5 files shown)


AllanZhengYP (dismissed):

packages/core/src/singleton/Amplify.ts
```diff
@@ -98,17 +100,27 @@ export class AmplifyClass {
...
+	private notifyOAuthListener(
+		amplify: AmplifyClass,
+		config: AuthConfig['Cognito'],
+		oAuthListener: OAuthListener,
+	) {
+		if (amplify.listenerTimeout) {
+			amplify.clearListenerTimeout();
 		}
+		// The setTimeout will only be called as long as an oAuthLister and oauth config
+		// are defined. The code executed by the time out will be only fired once.
```
HuiSF, line 112: This sounds like it only works with certain preconditions, e.g. assuming `Amplify.configure()` gets called back to back. While this trick seemed to be working, I'm hesitant to make this a solution...

With Amplify Gen2, `amplify_outputs.json` can be overridden from the `backend.ts` ([reference](https://docs.amplify.aws/react/build-a-backend/auth/use-existing-cognito-resources/#use-auth-resources-with-an-amplify-backend)). With Amplify Gen1, should be able to change the url in `amplifyconfiguration.json` as well, would be these be legit workarounds to solve what the PR is trying to solve?

cshfang: I have to echo this sentiment. When I looked at this PR this was my initial concern as well. Code after the configure call that may have previously executed after the `this.oAuthListener` call was made may now execute after the call due to it being pulled out of the current execution cycle. While I'm not certain that is practically concerning, it feels quite risky.

packages/core/src/singleton/Amplify.ts
```diff
@@ -98,17 +100,27 @@ export class AmplifyClass {
...
-	private notifyOAuthListener() {
-		if (
-			!this.resourcesConfig.Auth?.Cognito.loginWith?.oauth ||
-			!this.oAuthListener
-		) {
-			return;
+	private notifyOAuthListener(
+		amplify: AmplifyClass,
+		config: AuthConfig['Cognito'],
+		oAuthListener: OAuthListener,
```
cshfang, line 106: Making this pure would make sense if it were removed completely from the class, but since it is a private member of the class, this seems confusing. For example, it's not really clear to me why `oAuthListener` is passed separately from `this`. The `clearListenerTimeout` is then also tacked onto the instance but it's obfuscated here.

cshfang (changes requested):

PR #13727, opened 2024-08-19
feat(api-graphql): pass authToken via subprotocol

iartemiev:
#### Description of changes

AppSync Realtime is changing the way it accepts auth tokens when establishing subscription connections.

Today auth tokens are passed in the connection url as a query string param, e.g. `wss://api.example.com/graphql?header=<base64_encoded_token> `

Going forward, the service expects the client to pass the auth token through as a WS subprotocol:
```ts
const url = 'wss://api.example.com/graphql'
const protocols = ['graphql-ws', 'base64url_encoded_token']

new WebSocket(url, protocols)
```

This change is an implementation detail that does not affect the behavior of the AppSyncRealtime client. AppSync will continue to accept the auth token in query params for existing APIs, so older versions of the Amplify library will work as expected.

> [!NOTE]  
> The `WebSocket` Web API prohibits separator characters in subprotocol values, e.g. `/`, `=`, which are part of the standard base64 char set. With this change, the auth token is encoded as [base64url](https://datatracker.ietf.org/doc/html/rfc4648#section-5), which lacks any prohibited characters in its alphabet. AppSync now supports this encoding for auth headers.

The change is also being back-ported to v5 here: https://github.com/aws-amplify/amplify-js/pull/13726

* manual testing
* updated unit tests
* E2Es: https://github.com/aws-amplify/amplify-js/actions/runs/10579328718?pr=13727

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/__tests__/AWSAppSyncRealTimeProvider.test.ts
```diff
@@ -298,8 +298,50 @@ describe('AWSAppSyncRealTimeProvider', () => {
...
+					base64url encoding is expected to drop padding chars `=`
+					*/
+
+					expect(newSocketSpy).toHaveBeenNthCalledWith(
+						1,
+						'wss://testaccounturl123456789123.appsync-realtime-api.us-east-1.amazonaws.com/graphql',
+						[
+							'graphql-ws',
+							'header-eyJBdXRob3JpemF0aW9uIjoibXktY3VzdG9tLWF1dGgtdG9rZW4iLCJob3N0IjoidGVzdGFjY291bnR1cmwxMjM0NTY3ODkxMjMuYXBwc3luYy1hcGkudXMtZWFzdC0xLmFtYXpvbmF3cy5jb20ifQ',
+						],
```
david-mcafee, line 344: I realize this is already not covered by existing unit tests, but thoughts on adding a custom domain test here?

david-mcafee (approved):
LGTM, just left one non-blocking comment!

stocaaro (dismissed):

packages/api-graphql/__tests__/AWSAppSyncRealTimeProvider.test.ts
iartemiev: Added

stocaaro (approved):

david-mcafee (approved):

ashika112 (approved):
Approving on bundle size increase

packages/core/src/utils/convert/base64/base64Encoder.ts
```diff
@@ -2,18 +2,37 @@
...
-	convert(input, { urlSafe } = { urlSafe: false }) {
+	/**
+	 * Convert input to base64-encoded string
+	 * @param input - string to convert to base64
+	 * @param options - encoding options that can optionally produce a base64url string
+	 * @returns base64-encoded string
+	 */
+	convert(
+		input,
+		options: Base64EncoderConvertOptions = {
```
HuiSF, line 18: nit: `options` should've been typed by the `Base64Encoder`

packages/core/src/utils/convert/types.ts
```diff
@@ -3,6 +3,7 @@
 
 export interface Base64EncoderConvertOptions {
 	urlSafe: boolean;
+	skipPadding?: boolean;
```
HuiSF, line 6: oops, should the `urlSafe` be an optional property too, otherwise when you specify `skipPadding` you'd need to pass in a value for `urlSafe` as well. This is a internal tooling, I think it's safe to make a necessary change.

HuiSF (approved):

packages/core/src/utils/convert/types.ts
iartemiev: This was intentional. `skipPadding` would only be valid if `urlSafe` is set to `true`.

packages/core/src/utils/convert/base64/base64Encoder.ts
iartemiev: It is, I just wanted it to be more explicit, but I can follow up and remove in another PR if you prefer

PR #13735, opened 2024-08-20
feat(api-graphql): AppSync realtime - include custom headers as query string params

iartemiev:
#### Description of changes

Today custom request headers set via:
```ts
// client level
generateClient<Schema>({
  headers: {
    'My-Custom-Header': 'my value',
  },
});

// OR

// request level
await client.models.Blog.get(
  { id: 'myBlogId' },
  {
    headers: {
      'My-Custom-Header': 'my value',
    },
  }
);
```
Get applied to GraphQL requests for queries and mutations. ([Docs link](https://docs.amplify.aws/react/build-a-backend/data/connect-to-API/#set-custom-request-headers))

However, the headers are ignored for subscription connection requests to AppSync, becuase the WebSocket API does not support setting headers directly.

The AppSync team is asking us to give customers a way to pass these through as query string params instead.

E.g.
```ts
await client.models.Blog.onCreate(
  {
    headers: {
      'My-Custom-Header': 'my value',
    },
  }
);
```
Should produce this query string that's added to the subscription connect url: `"&My-Custom-Header=my%20value"`

> [!NOTE]  
> The service will ignore any query params it does not recognize. So including additional params will not cause any unwanted side effects

* Unit tests
* Manual testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/api-graphql/src/internals/v6.ts
+++ b/packages/api-graphql/src/internals/v6.ts
@@ -106,6 +106,7 @@ export function graphql<
 	const internals = getInternals(this as any);
 	options.authMode = options.authMode || internals.authMode;
 	options.authToken = options.authToken || internals.authToken;
+	const headers = additionalHeaders || internals.headers;
 
 	/**
 	 * The correctness of these typings depends on correct string branding or overrides.
@@ -116,7 +117,7 @@ export function graphql<
 		// TODO: move V6Client back into this package?
 		internals.amplify as any,
 		options,
-		additionalHeaders,
+		headers,
 	);
 
 	return result as any;
--- a/packages/api-graphql/__tests__/AWSAppSyncRealTimeProvider.test.ts
+++ b/packages/api-graphql/__tests__/AWSAppSyncRealTimeProvider.test.ts
@@ -276,38 +276,6 @@ describe('AWSAppSyncRealTimeProvider', () => {
 					);
 				});
 
-				test('custom headers are applied in query string', async () => {
-					expect.assertions(1);
-
-					const newSocketSpy = jest
-						.spyOn(provider, 'getNewWebSocket')
-						.mockImplementation(() => {
-							fakeWebSocketInterface.newWebSocket();
-							return fakeWebSocketInterface.webSocket as WebSocket;
-						});
-
-					provider
-						.subscribe({
-							appSyncGraphqlEndpoint: 'http://localhost:8080',
-							additionalHeaders: {
-								'x-amz-user-agent': 'aws-amplify/6.4.0 api/1 framework/2',
-								'ex-machina': 'is a good movie',
-								// This should NOT get included in the querystring
-								Authorization: 'abc12345',
-							},
-						})
-						.subscribe({ error: () => {} });
-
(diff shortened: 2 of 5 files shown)


packages/api-graphql/src/internals/v6.ts
```diff
@@ -106,6 +106,7 @@ export function graphql<
 	const internals = getInternals(this as any);
 	options.authMode = options.authMode || internals.authMode;
 	options.authToken = options.authToken || internals.authToken;
+	const headers = additionalHeaders || internals.headers;
```
iartemiev, line 109: This fixes an existing bug - we don't actually pass headers defined in `generateClient` through to `graphql` today.

svidgen: We have tests in the `amplify-api-next` defined behavior tests for client-level headers on modeled operations in `2-expected-use/auth-modes.ts`.

Is the gap *just* for the raw `graphql()` operation? Or, are those modeled operation tests silently failing?

svidgen: Maybe I'm confused, actually. What I thought you were saying seems to be covered in the core `graphql()` case also. We have a test [here](https://github.com/aws-amplify/amplify-js/blob/main/packages/api-graphql/__tests__/GraphQLAPI.test.ts#L1206) that appears to cover headers supplies at various levels as well.

iartemiev: 1. Added here - https://github.com/aws-amplify/amplify-js/pull/13735/commits/3668cdfbd1336a4dfe114c808bc4f2450cd62b80
2. For posterity - modeled operations perform an additional step where they extract the headers from the client, then pass them through explicitly via the 2nd param to `client.graphql` (example for [list](https://github.com/aws-amplify/amplify-api-next/blob/main/packages/data-schema/src/runtime/internals/operations/list.ts#L94)). So they don't rely on the client-bound headers being passed through automatically, i.e. the patch introduced in this PR.

svidgen (approved):

stocaaro (approved):
LGTM

luunminh:
Can anyone provide some sample code. I tried to pass auth token but its not work.
```
export const WSClient = generateClient({
  headers: async (requestOptions) => {
    const token = await TokenService.getToken();
    console.log('headers: ~ token:', token);
    return {
      Authorization: token as string,
    };
  },
});

const useGraphQLSubscription = <TSubscriptionVariables = any, TSubscriptionResponse = any>({
  mutation,
  variables,
  callback,
  enabled = true,
}: Props<TSubscriptionVariables, TSubscriptionResponse>) => {
  useEffect(() => {
    if (!enabled) return;

    const subscription = WSClient.graphql({
      query: mutation,
      variables: variables || ({} as any),
    }).subscribe({
      next: ({ data }) => {
        callback(data as TSubscriptionResponse);
      },
      error: (error) => console.warn(error),
    });

    return () => {
      subscription.unsubscribe();
    };
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [callback, variables, enabled]);
};
```

Issue #13650, opened 2024-07-26
Custom userPoolEndpoint does not take effect in Next.js adapter and custom identityPoolEndpoint is unsupported


tonestrike:
### Is this related to a new or existing framework?

Next.js

### Is this related to a new or existing API?

Authentication

### Describe the feature you'd like to request

This is my configuration:

```
{
	Auth: {
		Cognito: {
			userPoolEndpoint: process.env.NEXT_PUBLIC_COGNITO_POOL_ENDPOINT,
			userPoolId: process.env.NEXT_PUBLIC_COGNITO_USER_POOL_ID!,
			identityPoolId: process.env.NEXT_PUBLIC_COGNITO_IDENTITY_POOL_ID!,
			userPoolClientId: process.env.NEXT_PUBLIC_COGNITO_CLIENT_ID!,
			signUpVerificationMethod: 'code',
			allowGuestAccess: true,
			loginWith: {
				email: true,
			},
		},
	},
}
```

My `userPoolEndpoint` points to LocalStack. Everything works correctly on the client side, but on the server side, Amplify is not able to parse the cookies and return the current user.

If I use an AWS hosted Cognito User Pool, everything works as expected. My assumption is that `userPoolEndpoint` is not supported on the server side.

This is the error:
```
[ResourceNotFoundException: IdentityPool 'us-east-1:7b676de5-ed4a-4750-95f5-223e8d95102d' not found.] {
  name: 'ResourceNotFoundException',
  $metadata: {
  attempts: 1,
  httpStatusCode: 400,
  requestId: '58fa10e9-e0be-4a81-b785-7611deed288c',
  extendedRequestId: undefined,
  cfId: undefined
}
```

This is my middleware:
```

export async function middleware(request: NextRequest) {
	const response = NextResponse.next()

	const authenticated = await runWithAmplifyServerContext({
		nextServerContext: { request, response },
		operation: async (contextSpec) => {
			try {
				const session = await fetchAuthSession(contextSpec)

				return session.tokens?.accessToken !== undefined && session.tokens?.idToken !== undefined
			} catch (error) {
				console.log(error)
				return false
			}
		},
	})

	if (authRoutes.includes(request.nextUrl.pathname)) {
		if (authenticated) {
			return NextResponse.redirect(new URL('/', request.url))
		}

		return response
	}

	if (authenticated) {
		return response
	}

	return NextResponse.redirect(new URL('/login', request.url))
}
```

### Describe the solution you'd like

[...]

PR #13739, opened 2024-08-20
fix(auth): custom `userPoolEndpoint` cannot be applied on the server-side

HuiSF:
#### Description of changes

This PR re-organized the Cognito Identity service clients, and turn them into factory function. These factories expecting an `endpointResolver` to be injected as dependency. So they can be created within the Auth category API calls, and ensure it uses the always up-to-date endpoint information.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13650

* Updated and added unit tests
* Manual testing calling the Auth category APIs and verify a custom user pool endpoint is applied

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/auth/src/providers/cognito/utils/signInHelpers.ts
```diff
@@ -1037,11 +1126,17 @@ export async function assertUserNotAuthenticated() {
...
-	accessToken?: string,
-): Promise<DeviceMetadata | undefined> {
+export async function getNewDeviceMetadata({
+	userPoolId,
+	userPoolEndpoint,
+	newDeviceMetadata,
+	accessToken,
+}: {
+	userPoolId: string;
+	userPoolEndpoint: string | undefined;
```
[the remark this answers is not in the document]
HuiSF: This is for enforcing passing the `userPoolEndpoint` parameter at the callsties of the `getNewDeviceMetadata()` function, the value can be `undefined`, in case we forget to do it somewhere, sometime.

packages/auth/src/foundation/core/parsers/index.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
HuiSF: It shortens import path on some occasions which is a nice way to organize code.
Though need to be care full import unexpected side effect?

e.g.
```
index.ts 
  module1.ts // <- has a side effect executes on module loads
  module2.ts
```
I want to use a function provided from module2, if import from `index.ts`, the side effect lives in module1 may get attached.

packages/auth/src/foundation/factories/serviceClients/cognitoIdentityProvider/createAssociateSoftwareTokenClient.ts
```diff
@@ -0,0 +1,30 @@
...
+	buildUserPoolSerializer,
+} from './shared/serialization';
+import {
+	AssociateSoftwareTokenCommandInput,
+	AssociateSoftwareTokenCommandOutput,
+} from './types/Sdk';
+import { DEFAULT_SERVICE_CLIENT_API_CONFIG } from './constants';
+
+export const createAssociateSoftwareTokenClient = (
+	config: ServiceClientAPIConfig,
```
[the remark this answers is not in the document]
HuiSF: Updated the type name to `ServiceClientFactoryInput`

packages/auth/__tests__/foundation/core/parsers/getRegion.test.ts
```diff
@@ -0,0 +1,47 @@
...
+		expect(region).toEqual('us-west-2');
+	});
+
+	test.each([undefined, 'invalid-id-123'])(
+		`throws an error when the user pool id is invalid as %p`,
+		userPoolId => {
+			expect(() => getRegionFromUserPoolId(userPoolId)).toThrow(
+				new AuthError({
+					name: 'InvalidUserPoolId',
+					message: 'Invalid user pool id provided.',
```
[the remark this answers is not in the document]
HuiSF: This is the original error set up, will not change it in this PR.

packages/auth/src/foundation/core/cognitoUserPoolEndpointResolver.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
HuiSF: Flattened it a bit!

packages/auth/__tests__/foundation/core/parsers/getRegion.test.ts
```diff
@@ -0,0 +1,47 @@
+import { AuthError } from '../../../../src/errors/AuthError';
```
cshfang, line 1: Filename is outdated

packages/auth/__tests__/foundation/factories/serviceClients/cognitoIdentityProvider/shared/serialization/buildEmptyResponseDeserializer.test.ts
```diff
@@ -0,0 +1,55 @@
+import {
```
cshfang, line 1: Filenames are outdated

packages/auth/src/foundation/factories/serviceClients/cognitoIdentityProvider/shared/serde/createEmptyResponseDeserializer.ts
```diff
@@ -8,16 +8,14 @@ import {
...
+	<Output>(): ((response: HttpResponse) => Promise<Output | undefined>) =>
+	async (response: HttpResponse): Promise<Output | undefined> => {
 		if (response.statusCode >= 300) {
 			const error = await parseJsonError(response);
 			assertServiceError(error);
 			throw new AuthError({ name: error.name, message: error.message });
 		} else {
-			return undefined as any;
+			return undefined;
 		}
```
cshfang, line 20: Nit: I meant - is this else clause even needed? Doesn't this already implicitly return undefined if it falls through?

packages/auth/src/foundation/core/parsers/index.ts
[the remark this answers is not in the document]
cshfang: Seems like a good reason to consider dropping the practice overall except for maybe exposing types and controlling top level exports 🤔

packages/auth/src/providers/cognito/utils/signInHelpers.ts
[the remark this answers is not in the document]
cshfang: If it's typed and enforced, how could we forget to do it?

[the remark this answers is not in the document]
HuiSF: Resolved offline

packages/auth/src/foundation/factories/serviceClients/cognitoIdentityProvider/shared/serde/createEmptyResponseDeserializer.ts
HuiSF: Oh muhaha, yes!

packages/auth/src/providers/cognito/factories/createCognitoUserPoolEndpointResolver.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { EndpointResolverOptions } from '@aws-amplify/core/internals/aws-client-utils';
+import { AmplifyUrl } from '@aws-amplify/core/internals/utils';
+
+import { cognitoUserPoolEndpointResolver } from '../../../foundation/cognitoUserPoolEndpointResolver';
+
+export const createCognitoUserPoolEndpointResolver =
+	({ endpointOverride }: { endpointOverride: string | undefined }) =>
```
israx, line 9: why not just making the `endpointOverride` key optional instead of forcing to pass a value?

HuiSF: See this [comment](https://github.com/aws-amplify/amplify-js/pull/13739#discussion_r1725948322), it answers this.

packages/auth/src/foundation/factories/serviceClients/cognitoIdentityProvider/shared/serde/createEmptyResponseDeserializer.ts
HuiSF: Actually remove the `return undefined` changes the function signature to `return void`. I will not change this code here in case changing return type cause issue else where.

cshfang (approved):

packages/auth/src/providers/cognito/apis/confirmResetPassword.ts
```diff
@@ -53,10 +54,14 @@ export async function confirmResetPassword(
 		userPoolId,
 		userPoolClientId,
 	});
-
+	const confirmForgotPassword = createConfirmForgotPasswordClient({
+		endpointResolver: createCognitoUserPoolEndpointResolver({
+			endpointOverride: userPoolEndpoint,
+		}),
+	});
```
AllanZhengYP, line 61: My hot take on this approach: the endpoint makes more sense to be a part of the `confirmForgotPassword` client. Re-creating the endpoint resolver in every api implementation seems redundant. By this I mean each cognito idp clients can bake-in this `endpointOverride` option and accept it along with `region` and `userAgentValue`.

packages/auth/src/foundation/factories/serviceClients/cognitoIdentityProvider/createAssociateSoftwareTokenClient.ts
AllanZhengYP, line 1: The `Client` in the naming is a bit repetitive. Since we already have `serviceClients` in path, should we just drop the `Client`?

israx (dismissed):

packages/auth/src/providers/cognito/apis/confirmResetPassword.ts
HuiSF: I understood your suggestion. 

> Re-creating the endpoint resolver in every API implementation seems redundant.

This instance recreation ensures the resolver and the service client gain a configuration that's always up-to-date. After each API call completes, the created instances will be GC's. This behavior on the resolver is exactly the same as the original implementation; the newly added behavior is to create a new client for every API call instead of using a singleton. The latter seemed redundant but here we ensure the clients can respond to the updates from the injected resolver, I think it worth it, plus it doesn't introduce any performance drop.

> Each cognito idp client can bake in this endpointOverride option and accept it along with region and userAgentValue.

I oppose this, as we inject the resolver as a dependency on the service client, and the latter only needs to consume the result returned by the resolver, so we separate the concerns and make the service client handle only the request parameters and sending the request instead of handling endpoint override as an addition.

AllanZhengYP: Synced offline. The main difference is whether to supply the endpoint resolver logic to service client at every call, or keep the endpoint resolver the same but accept `customEndpoint` configuration when calling the service client. The S3 client uses the later approach([link](https://github.com/aws-amplify/amplify-js/blob/0d2aa5d71a6b995e302c7404becd8345763eae43/packages/storage/src/providers/s3/utils/client/base.ts#L62)).

This is not a blocker. We will align the implementation in separately.

AllanZhengYP (approved):

ashika112 (approved):

packages/auth/src/foundation/factories/serviceClients/cognitoIdentityProvider/createAssociateSoftwareTokenClient.ts
HuiSF: We can adjust the naming when we start doing the logical separation refactoring again. 👍

packages/auth/src/providers/cognito/apis/confirmSignIn.ts
```diff
@@ -27,12 +27,12 @@ import { assertValidationError } from '../../../errors/utils/assertValidationErr
 import { AuthValidationErrorCode } from '../../../errors/types/validation';
 import { AuthErrorCodes } from '../../../common/AuthErrorStrings';
 import { cacheCognitoTokens } from '../tokenProvider/cacheTokens';
+import { tokenOrchestrator } from '../tokenProvider';
```
Samaritan1011001, line 30: nit: unused?

Samaritan1011001, line 30: nit: unused?

HuiSF: I'm pretty sure it's used at line 98

ashika112 (approved):

cshfang (approved):

israx (approved):

PR #13760, opened 2024-08-27
feat(auth): [EMAIL MFA] Sign In / Confirm Sign In With MFA_SETUP

jjarvisp:
#### Description of changes

The purpose of this pull request is to expand the MFA_SETUP challenge to support setting up and signing in with multiple MFA setup options - EMAIL and TOTP. 

- Manual Tests
- Unit Tests

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


israx (changes requested):
Thanks for putting this together. 

Keeping track of state at 2 different places can be problematic. It introduces complexity when debugging, refactoring, and scaling. It can also lead to bugs. My concern is that when adding resumable-sign-in, the new store would also need to store state in sessionStorage.

To avoid all of that, we can keep handling the same state that is managed by the `signInStore` and keep the sign-in flow between the `getSignInResult` and `handleChallengeName` helpers. If we want to add some extra state, we can add some `metadata` in the signInStore — I don't see the need but I'll leave it up to you.

Bellow I added a flow diagram that shows the sign-in flow between the `handleChangeName` and `getSignInResult` helpers when challengeName is `MFA_SETUP`

sign-in-flow drawio

packages/auth/src/providers/cognito/utils/signInStore.ts
```diff
@@ -20,12 +20,12 @@ type SignInAction =
...
 	| { type: 'SET_SIGN_IN_SESSION'; value?: string };
 
-type Store<State, Action> = (reducer: Reducer<State, Action>) => {
+export type Store<State, Action> = (reducer: Reducer<State, Action>) => {
 	getState(): ReturnType<Reducer<State, Action>>;
 	dispatch(action: Action): void;
 };
 
-type Reducer<State, Action> = (state: State, action: Action) => State;
+export type Reducer<State, Action> = (state: State, action: Action) => State;
```
israx, line 28: if these types are not used outside this module, we can remove the `export`

israx (commented):
Thanks for the thorough testing. Can we also add test cases for `confirmSignIn MFA_SETUP challenge negative path cases` ? — Making sure `confirmSignIn` is still callable when an invalid response is passed in.

packages/auth/src/providers/cognito/utils/signInStore.ts
jjarvisp: Nice catch, this has been updated.

israx (approved):

PR #13762, opened 2024-08-28
add process script and update typedoc config

jacoblogan:
#### Description of changes
As part of the api docs migration typedoc will be exported as a json file and then flattened to be used by the docs repo to generate the api docs pages.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/scripts/process-typedoc.mjs
+++ b/scripts/process-typedoc.mjs
@@ -1,0 +1,75 @@
+import references from '../docs/reference.json' assert { type: 'json' };
+import { writeFileSync } from 'fs';
+
+// build flat object for easier faster lookups
+const flatReferences = {};
+
+const recursivelyPopulateFlatObject = referenceObject => {
+	if (!referenceObject) return;
+	if (referenceObject['id']) {
+		const copy = recursivelyStripObject(structuredClone(referenceObject));
+		flatReferences[referenceObject['id']] = copy;
+	}
+
+	for (let key in referenceObject) {
+		if (referenceObject.hasOwnProperty(key)) {
+			if (Array.isArray(referenceObject[key])) {
+				referenceObject[key].forEach(child => {
+					recursivelyPopulateFlatObject(child);
+				});
+			} else if (
+				typeof referenceObject[key] === 'object' &&
+				referenceObject[key] !== null
+			) {
+				recursivelyPopulateFlatObject(referenceObject[key]);
+			}
+		}
+	}
+};
+
+const recursivelyStripObject = referenceObject => {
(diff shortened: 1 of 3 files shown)


scripts/process-typedoc.mjs
```diff
@@ -0,0 +1,77 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import references from '../docs/reference.json' assert { type: 'json' };
+import { writeFileSync } from 'fs';
+
+// build flat object for easier faster lookups
+const flatReferences = {};
+
+const recursivelyPopulateFlatObject = referenceObject => {
```
jimblanc, line 9: Some inline documentation on the various steps in this script would be really helpful for future maintainers and those not familiar with the API docs, for example: https://github.com/aws-amplify/amplify-js/blob/main/scripts/set-preid-versions.sh

Also could you please document in the script which versions of TypeDoc this is compatible with?

scripts/process-typedoc.mjs
```diff
@@ -0,0 +1,77 @@
...
+	return cat;
+});
+
+try {
+	writeFileSync(
+		'docs/parsedJson.json',
+		JSON.stringify(flatReferences, null, 2),
+		'utf8',
+	);
+	console.log('Data successfully saved to disk');
```
jimblanc, line 74: Could we make these messages a little more descriptive to help debugging if something goes wrong, e.g.:

```suggestion
	console.log('Successfully saved parsed API information');
```

scripts/process-typedoc.mjs
```diff
@@ -0,0 +1,77 @@
...
+} catch (error) {
+	console.log('An error has occurred ', error);
```
jimblanc, line 76: Sanity check: If something failed in this script, will the release process be disrupted? Or does it depend on the eventual workflow implementation?

scripts/process-typedoc.mjs
jacoblogan: Thanks inline documentation has been added to the script

scripts/process-typedoc.mjs
jacoblogan: You are correct the error here is swallowed up and will not disrupt the release process.  The workflow implementation will look for changes to the typedoc output, but as discussed earlier we don't want this process to block the release.

scripts/process-typedoc.mjs
jimblanc, line 1: File missing copyright header at top

PR #13780, opened 2024-09-04
update release workflow

jacoblogan:
#### Description of changes
update release workflow to dispatch docs action

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


.github/workflows/callable-npm-publish-release.yml
```diff
@@ -54,12 +57,68 @@ jobs:
...
           yarn run docs
           git add ./docs/api/
-          git commit -m "chore(release): Update API docs [skip release]"
+
+      # Check if the API Docs have been updated
+      - name: Check for updated api docs
+        working-directory: ./amplify-js
+        id: api_docs_updated
+        run: |
+          if git diff --cached --quiet -- ${{ env.API_FILE }}; then
```
jimblanc, line 72: Will this ever not be false? Doesn't TypeDoc incorporate the commit ID when generating (thus would get updated for every release)? Might only be true for the full HTML generated docs

.github/workflows/callable-npm-publish-release.yml
```diff
@@ -54,12 +57,68 @@ jobs:
...
+          else
+            echo "updated=true" >> $GITHUB_OUTPUT
+            echo "API docs have been updated"
+          fi
+
+      # Commit docs changes
+      - name: Commit docs changes
+        working-directory: ./amplify-js
+        run: |
+          git commit -m "chore(release): update API docs [skip release]"
```
jimblanc, line 84: Nit

```suggestion
          git commit -m "chore(release): Update API docs [skip release]"
```

jimblanc (commented):
Makes sense to me, few nits & questions

.github/workflows/callable-npm-publish-release.yml
jacoblogan: Yes good catch the packageVersion is being included in the JSON output so this would always be false.  The packageVersion is not being used in the docs template so I've updated the script to strip out the packageVersion key from the resulting parsedJson file and changed this comparison to be with the parsedJson file that is being generated without the packageVersion keys

jimblanc (approved):

PR #13826, opened 2024-09-18
chore: update contributing guide vite troubleshooting

scanlonp:
My library changes were not showing up in my locally deployed app. Vite caching turned out to be the issue. Add a note to help future contributors avoid this problem.


Samaritan1011001 (dismissed):

CONTRIBUTING.md
```diff
@@ -126,6 +126,16 @@ In your sample project, you can now link specific packages
 yarn link @aws-amplify/auth
 ```
 
+If you are testing with a Vite sample project, such as from the quickstart guide, and your changes to the library are not showing up in your local app when running `npm run dev`, Vite caching may be the culprit. Try editing this line in your sample project's `package.json`.
```
jjarvisp, line 129: One recommendation, pretty much everywhere else in the contributing guide refers to using `yarn` as the package manager, so it's probably a good idea to keep it consistent.

scanlonp: Good call out. The quickstart guide uses npm, but there is no difference in behavior. I'll add that.

scanlonp, line 129:
```suggestion
If you are testing with a Vite sample project, such as from the quickstart guide, and your changes to the library are not showing up in your local app when running `yarn run dev`, Vite caching may be the culprit. Try editing this line in your sample project's `package.json`.
```

AllanZhengYP (approved):

jjarvisp (approved):

ashika112 (approved):

PR #13847, opened 2024-09-24
fix(rtn-web-browser): prevent CustomTab to show in recent used app

AnantTiwari001:
#### Description of changes
Add extra flag to the custom browser Tab to exclude the app from recently used apps.

Went through the android documentation and tested it locally on various devices.


cshfang (approved):
Makes sense to me! Will add approval but could we verify that this doesn't have unintended consequences (like hiding the host app from recent?) as well?

AnantTiwari001:
@cshfang Yeah I do understand the risk of unintended effects. 
Altho I do have tested it manually in both emulator and physical devices. And the Flag has been added specifically to the custom tab intent which won't leak, to other intents. Perhaps [here is the link for documentation of flag in question.](https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_EXCLUDE_FROM_RECENTS)

HuiSF (approved):
Thanks for your contribution!

PR #13851, opened 2024-09-25
feat(storage): export Default Part Size constant

Samaritan1011001:
#### Description of changes

Exposes an existing constant publicly to allow customers to use it when determining their path forward. This constant is internally used to determine if an upload task will be single part or multi part.

tested the import in a sample vite app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/storage/src/types/common.ts
```diff
@@ -47,11 +47,16 @@ export interface TransferTask<Result> {
 	 * Promise that resolves when the transfer task is completed. The promise will be rejected if the task is canceled.
 	 */
 	result: Promise<Result>;
+
+	/**
+	 * Boolean indicator showing if the transfer task is resumable.
```
jimblanc, line 52: Nit

```suggestion
	 * Indicates if the transfer task is resumable.
```

packages/storage/src/providers/s3/utils/transferTask.ts
```diff
@@ -77,7 +77,7 @@ export const createUploadTask = <Result>({
 	onCancel,
 	onResume,
 	onPause,
-	isMultipartUpload,
+	isMultipartUpload = false,
```
jimblanc, line 80: Since this is currently used in this function, do we want to set the default here? I think I'd prefer to just so this logic when setting `isResumable`, e.g `!!isMultipartUpload`

packages/storage/src/types/common.ts
Samaritan1011001: Thank you Jim. Resolving as the PR direction changed.

packages/storage/src/providers/s3/utils/transferTask.ts
Samaritan1011001: Thank you Jim. Resolving as the PR direction changed.

packages/storage/src/providers/s3/utils/constants.ts
```diff
@@ -13,6 +13,9 @@ const MiB = 1024 * 1024;
 const GiB = 1024 * MiB;
 const TiB = 1024 * GiB;
 
+/**
+ * Default part size in MB that is used to determine if an upload task is single part or multi part.
```
HuiSF, line 17: nit: can we also make it more explicit: "talk to UI before changing the value."

Samaritan1011001: In the doc strings? This is for the public and not just internal.

HuiSF: I guess we can add a line with `//` on top of the doc string

jimblanc (approved):

HuiSF (approved):

Issue #13853, opened 2024-09-26
Storage.get no longer working after V5.3.23 update


BuildingFiles:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

Storage

### Amplify Version

v5

### Amplify Categories

storage

### Backend

Amplify CLI

### Environment information

<details>

```
# Put output below this line
  System:
    OS: Windows 11 10.0.22631
    CPU: (16) x64 AMD Ryzen 7 5700G with Radeon Graphics
    Memory: 43.65 GB / 63.89 GB
  Binaries:
    Node: 20.17.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.8.3 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 129.0.6668.70
    Edge: Chromium (127.0.2651.74)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    @aws-amplify/cli: ^12.8.2 => 12.8.2
    @aws-amplify/react-native: ^1.1.5 => 1.1.5
    @babel/core: ^7.12.9 => 7.20.12
    @babel/plugin-proposal-export-namespace-from: ^7.18.9 => 7.18.9
    @expo/config-plugins: ~6.0.0 => 6.0.2
    @react-native-async-storage/async-storage: 1.17.11 => 1.17.11
    @react-native-community/netinfo: 9.3.7 => 9.3.7
    @react-native-masked-view/masked-view: 0.2.8 => 0.2.8
    @react-native-picker/picker: 2.4.8 => 2.4.8
    @react-navigation/bottom-tabs: ^5.11.15 => 5.11.15
    @react-navigation/drawer: ^5.12.9 => 5.12.9
    @react-navigation/native: ^5.9.8 => 5.9.8
    @react-navigation/stack: ^5.14.9 => 5.14.9
    @types/react: ~18.0.24 => 18.0.27
    HelloWorld:  0.0.1
[...]

PR #13881, opened 2024-10-03
fix(storage): browser module resoltuion is not picked up by metro web

HuiSF:
#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13853

When using the library with Expo, and run `npx expo start`, observed that the Storage uses `fetch` instead of `xhr` for making network requests. This caused that the `Storage.get()` returns a non `Blob`, `response.Body`. 

The metro bundler respect the `browser` module in `package.json`, but observed it didn't pick up the resolution during testing. The cause is that the module entry doesn't contain an extension name.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/storage/package.json
+++ b/packages/storage/package.json
@@ -6,7 +6,7 @@
   "module": "./lib-esm/index.js",
   "typings": "./lib-esm/index.d.ts",
   "browser": {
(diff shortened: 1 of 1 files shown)


jimblanc (approved):

AllanZhengYP (approved):
This change makes sense to me. This issue was not found earlier because most bundlers default to esm entry point. We should add the metro web app to our test.

PR #13903, opened 2024-10-09
fix: Update useragent string with npm version

ashika112:
#### Description of changes
- Fixes the issue with userAgent string getting encoded for the signer
   - Geo category recently had a [failing test  ](https://github.com/aws-amplify/amplify-js/actions/runs/11228288984/job/31212052785). The root cause of the issue was that the userAgent string was not properly encoded especially special character `+`. Lerna adds this special character `+` to the version and npm strips the value of post the `+` [[Ref]](https://github.com/lerna/lerna/issues/1893#issuecomment-901039466) we will do the same thing since we care about version in NPM for our metrics.
- Update `yarn.lock` since some version were backtracked weirdly last time we updated it 

- Unit test
- Manual test in sample geo app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/core/src/Platform/index.ts
```diff
@@ -11,7 +11,7 @@ import { getCustomUserAgent } from './customUserAgent';
 const BASE_USER_AGENT = `aws-amplify`;
 
 class PlatformBuilder {
-	userAgent = `${BASE_USER_AGENT}/${version}`;
+	userAgent = `${BASE_USER_AGENT}/${version.replace(/\+.*/, '')}`;
```
reesscot, line 14: Will changing this cause any issues with our metrics? Will we need to query for multiple versions of the version string now?

packages/core/__tests__/Platform/userAgent.test.ts
```diff
@@ -3,16 +3,17 @@ import {
...
 import {
 	clearCache,
 	detectFramework,
 } from '../../src/Platform/detectFramework';
 import * as detection from '../../src/Platform/detection';
 import { getCustomUserAgent } from '../../src/Platform/customUserAgent';
+import { version } from '../../src/Platform/version';
 
 jest.mock('../../src/Platform/customUserAgent');
+const expectedVersion = version.replace(/\+.*/, '');
```
HuiSF, line 16: nit: For testing purposes this could be a some hardcoded strings

ashika112: i was going to do that but then we will have to mock the version in first place so for now kept the same logic

stocaaro: You could push the version replacement into a function and then test that function with a variety of inputs/outputs.

packages/core/src/Platform/index.ts
stocaaro: As we discussed, this is going to leave the preid versions with the hash duplication. Probably non-impacting, but a bit ugly. If PM is bought in on this variation, its not a blocker for any technical reasons.

stocaaro (approved):
LGTM. Made a suggestion around the testing. I'm on the fence about whether thats more than a nit, so approving and good either way on if that happens.

packages/core/src/Platform/index.ts
ashika112: @stocaaro confirmed this wont cause an issue there since the current change will still preserve the hash. Its a duplicated bit of hash that is removed. Etc,

if version is `6.5.1-multi-bucket.7bc5681.0+7bc5681` -> `6.5.1-multi-bucket.7bc5681.0`, hash `7bc5681` is preserved :)

stocaaro (approved):

HuiSF (approved):

Issue #13791, opened 2024-09-07
Cache.clear() doesn't work


lwang-79:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Cache

### Amplify Version

v6

### Amplify Categories

Not applicable

### Backend

None

### Environment information

<details>

```
# Put output below this line
  System:
    OS: macOS 14.6.1
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 77.84 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.17.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.7.0 - /usr/local/bin/npm
    pnpm: 7.26.3 - /usr/local/bin/pnpm
  Browsers:
    Chrome: 128.0.6613.120
    Safari: 17.6
  npmPackages:
    %name%:  0.1.0 
    @aws-amplify/backend: ^1.2.1 => 1.2.1 
    @aws-amplify/backend-cli: ^1.2.5 => 1.2.5 
    @testing-library/jest-dom: ^5.17.0 => 5.17.0 
    @testing-library/react: ^13.4.0 => 13.4.0 
    @testing-library/user-event: ^13.5.0 => 13.5.0 
    @types/jest: ^27.5.2 => 27.5.2 
    @types/node: ^16.18.108 => 16.18.108 (20.16.5)
    @types/react: ^18.3.5 => 18.3.5 
    @types/react-dom: ^18.3.0 => 18.3.0 
    aws-amplify: ^6.6.0 => 6.6.0 
    aws-amplify/adapter-core:  undefined ()
    aws-amplify/analytics:  undefined ()
    aws-amplify/analytics/kinesis:  undefined ()
    aws-amplify/analytics/kinesis-firehose:  undefined ()
    aws-amplify/analytics/personalize:  undefined ()
[...]

PR #13926, opened 2024-10-15
fix(core): fixed cache.clear() not working as expected

yuhengshs:
#### Description of changes

The purpose of this PR is to fix the issue that `Cache.clear()` is not working as expected by not clearing the cache.

- Added `prefix` to the key to make sure the item is successfully identified and removed.
- Updated the unit test to prevent missing `prefix` and not clearing cache

#### Issue #, if available

[13791](https://github.com/aws-amplify/amplify-js/issues/13791)

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Unit Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)


packages/core/__tests__/Cache/StorageCacheCommon.test.ts
```diff
@@ -587,13 +587,24 @@ describe('StorageCacheCommon', () => {
...
+			mockGetAllCacheKeys.mockResolvedValue(initialKeys);
+
 			await cache.clear();
+
 			expect(loggerSpy.debug).toHaveBeenCalledWith('Clear Cache');
-			expect(mockKeyValueStorageRemoveItem).toHaveBeenCalledTimes(2);
+			expect(mockKeyValueStorageRemoveItem).toHaveBeenCalledTimes(
+				initialKeys.length,
+			);
+			mockKeyValueStorageGetItem.mockResolvedValue(null);
```
ashika112, line 603: I think mockResolve a function and expecting the value here does not provide us with any value. We would want to want to call the original implementation without mock to test the expected value so we catch the error next time around

ashika112 (dismissed):

packages/core/__tests__/Cache/StorageCacheCommon.test.ts
```diff
@@ -587,13 +587,22 @@ describe('StorageCacheCommon', () => {
...
+
 			await cache.clear();
+
 			expect(loggerSpy.debug).toHaveBeenCalledWith('Clear Cache');
-			expect(mockKeyValueStorageRemoveItem).toHaveBeenCalledTimes(2);
+			expect(mockKeyValueStorageRemoveItem).toHaveBeenCalledTimes(
+				initialKeys.length,
+			);
+			expect(await cache.getCurrentCacheSize()).toBe(0);
+			expect(await cache.getItem('some-key')).toBeNull();
```
jjarvisp, line 604: Because we're mocking the `getAllCacheKeys` fn these expects will pass regardless of whether `cache.clear()` was successful because there's nothing in the cache instance. Is there a way we can test the cache without mocking?

ashika112: i think the `getItem()` should catch that right? unless we are mocking that somewhere as well.

+1 tho. We want to make sure the test actually catches the failure case.

yuhengshs: Updated to use real implementation to test cache!

ashika112 (approved):

jjarvisp (approved):

PR #13997, opened 2024-11-09
Fix oidc authToken not getting passed through to websocket connection open for Events Api

NWylynko:
#### Description of changes

While attempting to use the new Event Api with Appsync, I wanted to authenticate using clerk.com, using there jwt templates feature I created an aws template and then came back to aws and setup the OpenID Connect authorization mode. But despite getting the token generated and passed through to the events.connect() method as the authToken, it would always fail to connect with error: "Required Headers are missing". Doing some digging in the code found that the options.authToken was effectively being ignored, and then on top of that it was still just trying to use the Amplify auth token. So I have updated the code to use the supplied jwt token.

I updated the raw js files in my project i am working on and was able to get the client to authorise with the websocket and receive messages, I can also see in the network tab of the dev tools that the jwt generated by Clerk is being passed through the headers of the connection.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


kumarmedasani (approved):

packages/api-graphql/src/Providers/AWSWebSocketProvider/authHeaders.ts
```diff
@@ -110,7 +117,7 @@ export const awsRealTimeHeaderBasedAuth = async ({
 	const headerHandler = {
 		apiKey: awsRealTimeApiKeyHeader,
 		iam: awsRealTimeIAMHeader,
-		oidc: awsAuthTokenHeader,
+		oidc: oidcAuthTokenHeader,
```
iartemiev, line 120: This change breaks the expected behavior for OIDC auth for both Events and GraphQL subscriptions. By default, this auth mode is meant to work with the OIDC configuration [provided through Amplify Backend](https://docs.amplify.aws/react/build-a-backend/data/customize-authz/using-oidc-authorization-provider/) and sign in/sign out handled via [Amplify Auth](https://docs.amplify.aws/react/build-a-backend/auth/connect-your-frontend/using-the-authenticator/). When configured this way, we automatically extract the OIDC access token from the currently signed in user in the `awsAuthTokenHeader` function and pass it through to the subscription auth token. This functionality must remain intact. 

That being said, I agree that we should allow a fallback to a manually-managed `authToken` passed in through the client's public API. I think we can do that by conditionally calling either the existing `customAuthHeader` or `awsAuthTokenHeader` depending on whether an explicit `authToken` was specified at the client call site.

iartemiev (changes requested):

PR #14053, opened 2024-12-04
refactor(auth): clean signIn state after all relevant async works finish

zhouzh1:
#### Description of changes

- Moved all the `cleanActiveSignInState` method invocations to after the `cacheCognitoTokens` method invocations, so that we can ensure the signIn state is only cleaned when all the tasks which're possible to throw error have been successfully finished, otherwise it would cause the relevant APIs isn't able to be retried on errors (e.g. network error).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


zhouzh1:
@cshfang @HuiSF @pranavosu Could you help take a look if this PR makes sense?

packages/auth/src/providers/cognito/apis/confirmSignIn.ts
```diff
@@ -121,7 +120,8 @@ export async function confirmSignIn(
 				}),
 				signInDetails,
 			});
-
+			cleanActiveSignInState();
```
Magminder, line 123: Wouldn't it clean something that `cacheCognitoTokens` does? Is `getNewDeviceMetadata` the only thing that makes requests? If so, maybe we just call it to save result before `cleanActiveSignInState` and leave the order as it is?

jjarvisp:
Hi @zhouzh1, thanks for the pull request! These changes look reasonable; will review internally with the team and follow up if we need anything else.

jjarvisp:
Hey @zhouzh1, are you able to clean up the linting errors so we can get the unit tests running? You should just need to run `yarn lint:fix --scope @aws-amplify/auth`

jjarvisp (approved):

jjarvisp:
Sample integration test run: https://github.com/aws-amplify/amplify-js/actions/runs/12206362394

HuiSF (approved):

zhouzh1:
@jjarvisp I tried the `yarn lint:fix --scope @aws-amplify/auth` command, but seems that all the displayed linting errors aren't related to my change (my change is simple, I guess it wouldn't introduce linting error), they're all relevant with the failed module path resolving.

Magminder (approved):

zhouzh1:
@jjarvisp @HuiSF One more question, I am thinking that is this `cleanActiveSignInState` invocation in the `catch` clause necessary? Because it will cause the associated API not able to be retried as well, and also, I saw that there is no such a invocation in the `confirmSignIn` API's corresponding `catch` clause.
image

jjarvisp (approved):

PR #14132, opened 2025-01-10
fix(core): generateRandomString uses Math.random()

HuiSF:
#### Description of changes

Update the existing `generateRandomString()` implementation to use `crypto.getRandomValues()`.

- unit tests
- manual testing with a sample app integrating Hosted UI.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/core/__tests__/utils/generateRandomString.test.ts
+++ b/packages/core/__tests__/utils/generateRandomString.test.ts
@@ -3,28 +3,48 @@
 
 import { generateRandomString } from '../../src/libraryUtils';
 
-describe('generateRandomString()', () => {
-	it('generates a string with the specified length', () => {
-		let counter = 0;
-		while (counter++ < 50) {
-			expect(generateRandomString(20).length).toEqual(20);
-		}
+describe('generateRandomString', () => {
+	const getRandomValuesSpy = jest.spyOn(crypto, 'getRandomValues');
+	beforeAll(() => {
+		// Mock crypto.getRandomValues
+		getRandomValuesSpy.mockImplementation(bufferView => {
+			if (!bufferView) {
+				return null;
+			}
+			const array = new Uint8Array(bufferView.buffer);
+			for (let i = 0; i < array.byteLength; i++) {
+				array[i] = Math.floor(Math.random() * 256);
+			}
+
+			return array;
+		});
 	});
 
-	it('generates correct string', () => {
-		const mathRandomSpy = jest.spyOn(Math, 'random');
-		let counter = 1;
-		mathRandomSpy.mockImplementation(() => {
-			const returnValue = counter;
-			counter += 5;
+	afterEach(() => {
+		getRandomValuesSpy.mockClear();
+	});
 
-			return parseFloat(`0.${returnValue}`);
-		});
+	it('generates a string of the specified length', () => {
+		const length = 10;
+		const result = generateRandomString(length);
+		expect(result).toHaveLength(length);
+	});
 
+	it('uses only characters from the charset', () => {
+		const result = generateRandomString(20);
+		const charset =
+			'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+		expect(result).toMatch(new RegExp(`^[${charset}]+$`));
+	});
+
+	it('calls crypto.getRandomValues with Uint8Array', () => {
+		generateRandomString(5);
+		expect(crypto.getRandomValues).toHaveBeenCalledWith(expect.any(Uint8Array));
+	});
+
+	it('generates different strings on subsequent calls', () => {
 		const result1 = generateRandomString(10);
-		counter = 1;
-		const result2 = generateRandomString(20);
-
-		expect(result2.substring(0, 10)).toEqual(result1);
+		const result2 = generateRandomString(10);
+		expect(result1).not.toEqual(result2);
 	});
 });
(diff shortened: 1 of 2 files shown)


ashwinkumar6 (approved):
changes lgtm !!
(will need to double check polyfill in RN)

packages/core/__tests__/utils/generateRandomString.test.ts
```diff
@@ -3,28 +3,48 @@
...
+	afterEach(() => {
+		getRandomValuesSpy.mockClear();
+	});
 
-			return parseFloat(`0.${returnValue}`);
-		});
+	it('generates a string of the specified length', () => {
+		const length = 10;
+		const result = generateRandomString(length);
+		expect(result).toHaveLength(length);
```
cshfang, line 30: Nit: I think it's probably ok to just be concise here
```suggestion
		expect(generateRandomString(length)).toHaveLength(10);
```

packages/core/__tests__/utils/generateRandomString.test.ts
```diff
@@ -3,28 +3,48 @@
...
+	});
 
-		const result1 = generateRandomString(10);
-		counter = 1;
-		const result2 = generateRandomString(20);
+	it('uses only characters from the charset', () => {
+		const result = generateRandomString(20);
+		const charset =
+			'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
```
cshfang, line 36: Nit: Should the charset be exported and used as a source of truth for this?

packages/core/__tests__/utils/generateRandomString.test.ts
```diff
@@ -3,28 +3,48 @@
...
+		expect(result).toMatch(new RegExp(`^[${charset}]+$`));
+	});
 
-		expect(result2.substring(0, 10)).toEqual(result1);
+	it('calls crypto.getRandomValues with Uint8Array', () => {
+		generateRandomString(5);
+		expect(crypto.getRandomValues).toHaveBeenCalledWith(expect.any(Uint8Array));
+	});
+
+	it('generates different strings on subsequent calls', () => {
```
cshfang, line 45: Is this a meaningful assertion given the mocked `getRandomValues`?

packages/core/__tests__/utils/generateRandomString.test.ts
HuiSF: I did this sorta elaborately - as the charset should NOT change often, when it changes, there should be something to break right away as a reminder for "why did this change".

packages/core/__tests__/utils/generateRandomString.test.ts
```diff
@@ -3,28 +3,48 @@
...
+	const getRandomValuesSpy = jest.spyOn(crypto, 'getRandomValues');
+	beforeAll(() => {
+		// Mock crypto.getRandomValues
+		getRandomValuesSpy.mockImplementation(bufferView => {
+			if (!bufferView) {
+				return null;
+			}
+			const array = new Uint8Array(bufferView.buffer);
+			for (let i = 0; i < array.byteLength; i++) {
+				array[i] = Math.floor(Math.random() * 256);
```
cshfang, line 16: Maybe instead of trying to replicate randomness, we should trust that `getRandomValues` would indeed give us random values and we mock here with something static that we can assert our actual logic (what we end up doing with the random values) is correct

packages/core/__tests__/utils/generateRandomString.test.ts
cshfang: That's a fair rationale 👍

packages/core/__tests__/utils/generateRandomString.test.ts
```diff
@@ -13,7 +18,7 @@ describe('generateRandomString', () => {
 			}
 			const array = new Uint8Array(bufferView.buffer);
 			for (let i = 0; i < array.byteLength; i++) {
-				array[i] = Math.floor(Math.random() * 256);
+				array[i] = mockRandomValues[i];
```
cshfang, line 21: Nit: Can't this entire mock implementation just return `mockRandomValues`?

packages/core/__tests__/utils/generateRandomString.test.ts
```diff
@@ -37,14 +46,9 @@ describe('generateRandomString', () => {
...
-	it('calls crypto.getRandomValues with Uint8Array', () => {
-		generateRandomString(5);
-		expect(crypto.getRandomValues).toHaveBeenCalledWith(expect.any(Uint8Array));
-	});
-
-	it('generates different strings on subsequent calls', () => {
-		const result1 = generateRandomString(10);
-		const result2 = generateRandomString(10);
-		expect(result1).not.toEqual(result2);
+	it('generates random strings with calculated indexes (num % STATE_CHARSET.length)', () => {
```
cshfang, line 49: Nit-ish: Can our "random values" actually be an exhaustive array that results in every character in the charset being used? Can also combine the previous test that way?

cshfang (approved):

packages/core/__tests__/utils/generateRandomString.test.ts
HuiSF: I think the mock implementation needs to mock the the exact behavior of getRandomValues:

1. modify the passed in typed array in place
2. return the same passed in typed array

cshfang: Good point

PR #14136, opened 2025-01-13
fix: move package exports for react native

timostroehlein:
#### Description of changes

Currently the `react-native` package exports are incorrectly order. The `react-native` package export should come before the `require` and `import` exports, since the condition names are resolved in order and [default](https://metrobundler.dev/docs/configuration/#unstable_conditionnames-experimental) to `['require', 'react-native']`.

I've only validated the changes by adjusting the exports using yarn patch in our repo.

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/analytics/package.json
+++ b/packages/analytics/package.json
@@ -43,33 +43,33 @@
 	"exports": {
 		".": {
 			"types": "./dist/esm/index.d.ts",
+			"react-native": "./src/index.ts",
 			"import": "./dist/esm/index.mjs",
-			"require": "./dist/cjs/index.js",
-			"react-native": "./src/index.ts"
+			"require": "./dist/cjs/index.js"
 		},
 		"./pinpoint": {
 			"types": "./dist/esm/providers/pinpoint/index.d.ts",
+			"react-native": "./src/providers/pinpoint/index.ts",
 			"import": "./dist/esm/providers/pinpoint/index.mjs",
-			"require": "./dist/cjs/providers/pinpoint/index.js",
-			"react-native": "./src/providers/pinpoint/index.ts"
+			"require": "./dist/cjs/providers/pinpoint/index.js"
 		},
 		"./kinesis": {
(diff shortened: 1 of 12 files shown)


HuiSF:
Hi @timostroehlein thank you for opening this PR, much appreciated. Have you actually encountered react-native bundle build issues with the existing `exports` settings? 

Following the Metro documentation, `unstable_conditionNames` takes only effect when `unstable_enablePackageExports` is set to `true`. Do you currently enable this option?

timostroehlein:
Hi @HuiSF, the issue actually isn't build related. The main issue was the inconsistency between the exports. Let me try to explain the main issue. The issue only happens if `unstable_enablePackageExports ` is enabled and `unstable_conditionNames ` is set to `['require', 'react-native']`.

For example the [@aws-amplify/core](https://github.com/aws-amplify/amplify-js/blob/931bfeccc6f8d2a03e390df1a2db9650f7bc4504/packages/core/package.json) package doesn't define any package exports, but has a `react-native` export field at root exporting ts files. The order of the exports on root doesn't matter, so even with package exports enabled, metro will import from the `react-native` entry point, which is correct in this case.

On the other hand, most of the packages, e.g. the [@aws-amplify/auth](https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/package.json) package does define package exports. Those exports are resolved in order, since the `unstable_conditionNames` in metro are set to `['require', 'react-native']` (order doesn't matter here), it will import from the require/cjs import. That isn't the expected behaviour, since we want to import from `react-native`.

In our case the Amplify config was undefined when calling `signIn` from `@aws-amplify/auth`, even though we called `Amplify.configure`, caused by both cjs and ts inconsistently referenced.

timostroehlein:
Hi @HuiSF,
Is there still something from my side missing? If so, please let me know.
I would highly appreciate it if we can get this merged, because then we can get rid of our patch we temporarily use to fix this issue on our side.

timostroehlein:
Has been resolved in the meantime by another PR

PR #14137, opened 2025-01-13
fix(deps): fix more implicit deps and add linting

pranavosu:
#### Description of changes

Fix implicit dependencies for packages (except adapter-nextjs<sup>1</sup>, and react-native adjacent<sup>2</sup>) to better support Yarn Pnp and pnpm.

1. adapter-nextjs ( waiting on https://github.com/aws-amplify/amplify-js/pull/14141 merge into main)
2. RN code needs to be separated from core to make native dependencies explicit. We can't have all consumers of the core package install native deps. Additionaly, RN doesn't support yarn PnP or pnpm so there's less upside 

 `yarn && yarn lint`

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


AllanZhengYP (approved):

ashika112 (dismissed):

joon-won (approved):

stocaaro (dismissed):

AllanZhengYP (approved):

packages/rtn-push-notification/package.json
```diff
@@ -23,10 +23,17 @@
...
+		"lodash": "^4.17.21"
+	},
 	"devDependencies": {
 		"@types/react-native": "0.70.0",
+		"react-native": "0.71.0",
 		"typescript": "5.0.2"
 	},
+	"peerDependencies": {
+		"react-native": ">=0.70"
+	},
```
cshfang, line 36: Is this strictly necessary?

packages/api/package.json
```diff
@@ -82,6 +83,11 @@
 	"dependencies": {
 		"@aws-amplify/api-graphql": "4.7.2",
 		"@aws-amplify/api-rest": "4.0.67",
+		"@aws-amplify/data-schema": "^1.7.0",
+		"rxjs": "^7.8.1",
 		"tslib": "^2.5.0"
+	},
+	"peerDependencies": {
```
cshfang, line 90: Could this cause issues with duplicate cores - e.g. during a MV bump this is forgotten?

eslint.config.mjs
```diff
@@ -294,4 +294,15 @@ export default [
 			'jsdoc/no-undefined-types': 1,
 		},
 	},
+	{
+		ignores: [
+			'**/**.{native,android,ios}.**',
```
cshfang, line 299: Does this mean these files aren't linted?

ashika112: Yes from the looks of it? anything specifically concerning?

packages/api/package.json
ashika112: we do this in all our packages already. I dont see how this addition would be any different? 👀

ashika112: But answer the actual question, yes. If we miss updating anywhere it will cause duplicate dependencies but i think that is caught in our tests

packages/rtn-push-notification/package.json
pranavosu: it's not strictly necessary because the `@aws-amplify/react-native` package has a peer dependency on `"react-native": ">=0.70"` but it is correct in isolation because this package uses `react-native`

cshfang: I don't think we have this in the web browser package either. Hesitant to use peer deps given the inconsistent behavior we sometimes see with different package managers and it gives us more places to update this when none of the `rtn-` packages really work in isolation anyway. Just a little scared

eslint.config.mjs
cshfang: Why wouldn't we want to lint say `getRedirectUrl.native.ts`? (I could be misinterpreting what's happening here)

packages/api/package.json
cshfang: I'm a little concerned about adding peer deps in general given their behavior at times but it's a little extra daunting when we have two deps on core here and they're already "mismatched". Should be fine, but I'm jut wondering if we're adding some baggage needlessly.

cshfang (approved):

packages/api/package.json
pranavosu: That's fair but it's useful for modern package managers like `pnpm`, so there is some benefit. I think it's better to not rely on implicit dependencies which could stop working with future versions of npm

packages/rtn-push-notification/package.json
pranavosu: For some reason I though rtn-web-browser had a peer dep but it's just a dev dep 🤦🏽  I can create a separate PR for that. If these are truly meant to be used with the `@amplify-aws/react-native` package then there should be no issues adding these peer deps. I haven't seen any in testing 🤞🏽

stocaaro (approved):

PR #14164, opened 2025-01-24
fix(api-graphql): Data messages should maintain the keep alive status

stocaaro:
#### Description of changes
**Problem:** In low bandwidth or high js thread contention cases, the AppSync websocket can get backed up. If it gets backed up to the point where 1 minute of data can't be processed in 5 minutes, then no keep alive messages arrive which leads the client to think that the connection is dead, prompting disconnect.

**Solution:** Any receipt of data should count as a keep alive message.

**Next Problem:** We receive substantial more data than keep alive messages and using the existing keep alive maintenance logic with each fresh piece of data is expensive.  *In testing for 500,000 messages received the processing time goes from ~2100ms up to ~3000ms by adding this logic*.

**Next Solution:** Revise the keep alive maintenance logic to capture a timestamp, then have a heartbeat interval that reviews the keep alive alert and disconnect conditions every 5 seconds. *As a result, the processing time of maintaining the keep alive heartbeat for each receipt of data for 500,000 messages is ~2150ms*

A unit test was added to validate the change. Tested failing without the logic update, passing with the data maintenance logic.

I will be performing manual app testing on Monday 1/27.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
```diff
@@ -83,9 +83,8 @@ export abstract class AWSWebSocketProvider {
 
 	protected awsRealTimeSocket?: WebSocket;
 	private socketStatus: SOCKET_STATUS = SOCKET_STATUS.CLOSED;
-	private keepAliveTimeoutId?: ReturnType<typeof setTimeout>;
-	private keepAliveTimeout = DEFAULT_KEEP_ALIVE_TIMEOUT;
-	private keepAliveAlertTimeoutId?: ReturnType<typeof setTimeout>;
+	private kaTimestamp?: number;
+	private keepAliveHeartbeatIntervalId?: ReturnType<typeof setInterval>;
```
cshfang, line 87: Nit: Stick with one convention for clarity (prefer latter - verbose but expressive)

Also nit: I wonder if just having `number` as the type would be sufficient, when reading this outside of an IDE, I just find myself having to go figure out what the return type of setInterval calls is

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
```diff
@@ -577,13 +572,43 @@ export abstract class AWSWebSocketProvider {
 		errorType: string;
 	};
 
+	private maintainKeepAlive() {
```
cshfang, line 575: Nit: `keepKeepAliveAlive`

Jk 😛

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
```diff
@@ -577,13 +572,43 @@ export abstract class AWSWebSocketProvider {
...
+			this.kaTimestamp &&
+			currentTime - this.kaTimestamp > DEFAULT_KEEP_ALIVE_ALERT_TIMEOUT
+		) {
+			this.connectionStateMonitor.record(CONNECTION_CHANGE.KEEP_ALIVE_MISSED);
+		} else {
+			this.connectionStateMonitor.record(CONNECTION_CHANGE.KEEP_ALIVE);
+		}
+
+		// Recognize we are disconnected if we haven't seen messages in the keep alive timeout period
+		if (
```
cshfang, line 593: Sanity check because of unfamiliarity but are we expecting that we might record KA/KA missed above and then disconnect here?

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
```diff
@@ -686,13 +702,21 @@ export abstract class AWSWebSocketProvider {
...
 		}
 
 		this.socketStatus = SOCKET_STATUS.CLOSED;
 	}
 
+	private _closeSocket() {
+		if (this.keepAliveHeartbeatIntervalId) {
+			clearInterval(this.keepAliveHeartbeatIntervalId);
+			this.keepAliveHeartbeatIntervalId = undefined;
+		}
```
cshfang, line 716: Nit: I think you can ditch the `if` clause for brevity as `clearInterval` won't complain on nullish input

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
```diff
@@ -912,7 +938,11 @@ export abstract class AWSWebSocketProvider {
 			return;
 		}
 
-		this.keepAliveTimeout = connectionTimeoutMs;
+		// Setup a keep alive heartbeat for this connection
```
cshfang, line 941: Super duper nit:
```suggestion
		// Set up a keep alive heartbeat for this connection
```

zxl629 (approved):
LGTM!

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
stocaaro: Agreed on `ka` vs `keepAlive`.

The return type of `setInterval` is `NodeJS.Timer` which is an object wrapping ref/unref functions. We could drop to having the type be `NodeJS.Timer` directly. In an earlier version of this code.

I had accidentally left this `ReturnType<typeof setTimeout>`, which has the same return type and worked, but having it clearly be the return type of the function call I'll be assigning to the variable carries extra context that I think is valuable.

I'm a fan of the current pattern, I would suppose outside an IDE, whatever the return type is, it tells me this variable hold the return value of a `setInterval` call, which I think is contextually valuable.

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
stocaaro: The reconnect change introduced multiple chatty connection state observables, like keep alive. Rather than adding logic to track current state and check the state to avoid sending a duplicate state, the states observable is [pipe'ed through a filter](https://github.com/aws-amplify/amplify-js/blob/86d787bfb57beb472c5617cf8c1bc3f2cd2f6c57/packages/api-graphql/src/utils/ConnectionStateMonitor.ts#L109-L131) that removes duplicates before they are surfaced to the user.

In this instance, we could check is the connection is times out and return. I can't think of a case where the disconnect would occur on the same check as the first keep alive alert. Its possible that there are suspense/recovery patterns that don't break the web socket that I'm not aware of.

Since KA missed and disconnect are both checks against time, I wouldn't expect KA -> Disconnect logically, but if the timeout constants where changed, I accept that this outcome is possible given how the code is structured right now.

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
cshfang: Well I hadn't considered TypeScript's feelings. Don't worry about it then

cshfang (approved):

zxl629 (approved):

PR #14222, opened 2025-02-18
(chore): Added early error message for configuration error

yuhengshs:
#### Description of changes

The purpose of this PR is to add early error message when `Amplify.configure()` is not called, an error will be thrown early to the customers and notify them Amplify.configure() is not called.
Unit test is added accordingly.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/analytics/src/providers/kinesis-firehose/utils/resolveConfig.ts
```diff
@@ -11,6 +11,7 @@ import {
 import { DEFAULT_KINESIS_FIREHOSE_CONFIG } from './constants';
 
 export const resolveConfig = () => {
+	Amplify.assertConfigured();
```
stocaaro, line 14: Quite a few of these occur right before we call `Amplify.getConfig()`. Do we always want to assert configured with getConfig? If so, could we include this in the getConfig behavior to remove some of the amount of change happening here?


Also, there are a number of places where this isn't couple with getConfig. This seems counterintuitive, right? Are these other callsites still calling getConfig, but a few layers of abstraction removed? Is there a pattern for accessing the config different from using getConfig?

stocaaro (commented):
The amount of change here makes me wonder if we're not missing something. How do we know that we haven't missed protecting a call site that uses config? Since this is spread over all of the categories independently, are we retesting all categories to ensure this behaves as expected in each case?

soberm:
Closed in favor of https://github.com/aws-amplify/amplify-js/pull/14424

PR #14231, opened 2025-02-20
fix(adapter-nextjs): secure: false is not set for localhost on sign-in

HuiSF:
#### Description of changes

* Move the `secure` cookie attribute value calculation into the functions that actually generate the cookie attributes to prevent misses

- local testing with a Next.js sample app running on localhost
- e2e test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/adapter-nextjs/src/auth/utils/authFlowProofCookies.ts
+++ b/packages/adapter-nextjs/src/auth/utils/authFlowProofCookies.ts
@@ -11,6 +11,8 @@ import {
 	STATE_COOKIE_NAME,
 } from '../constant';
 
+import { isSSLOrigin } from './origin';
+
 export const createSignInFlowProofCookies = ({
 	state,
 	pkce,
@@ -37,12 +39,12 @@ export const createSignOutFlowProofCookies = () => [
 
 export const createAuthFlowProofCookiesSetOptions = (
 	setCookieOptions: CookieStorage.SetCookieOptions,
-	overrides?: Pick<CookieStorage.SetCookieOptions, 'secure'>,
+	origin: string,
 ) => ({
 	domain: setCookieOptions?.domain,
 	path: '/',
 	httpOnly: true,
-	secure: overrides?.secure ?? true,
+	secure: isSSLOrigin(origin),
 	sameSite: 'lax' as const,
 	maxAge: AUTH_FLOW_PROOF_MAX_AGE,
 });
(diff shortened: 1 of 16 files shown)


packages/adapter-nextjs/src/auth/utils/authFlowProofCookies.ts
```diff
@@ -37,12 +39,12 @@ export const createSignOutFlowProofCookies = () => [
...
 export const createAuthFlowProofCookiesSetOptions = (
 	setCookieOptions: CookieStorage.SetCookieOptions,
-	overrides?: Pick<CookieStorage.SetCookieOptions, 'secure'>,
+	origin: string,
 ) => ({
 	domain: setCookieOptions?.domain,
 	path: '/',
 	httpOnly: true,
-	secure: overrides?.secure ?? true,
+	secure: isSSLOrigin(origin),
```
ashika112, line 47: curious: why were these overrides to begin with?

packages/adapter-nextjs/src/auth/utils/tokenCookies.ts
```diff
@@ -55,13 +56,13 @@ export const createTokenRemoveCookies = (keys: string[]) =>
...
 export const createTokenCookiesSetOptions = (
 	{ domain, sameSite, expires, maxAge }: CookieStorage.SetCookieOptions,
-	overrides?: Pick<CookieStorage.SetCookieOptions, 'secure'>,
+	origin: string,
 ) => {
 	const result = {
 		domain,
 		path: '/',
 		httpOnly: true,
-		secure: overrides?.secure ?? true,
```
jjarvisp, line 64: I'm not certain how this will produce different behavior than the existing implementation. Also, do we have a unit test to cover this miss?

HuiSF: This answer applicable to @ashika112 question above, too:

Originally, `secure` is enforced as `true` by default, then added allowing `secure: false` for local development servers. 

Originally it's consumers of `createTokenCookiesSetOptions()` and `createAuthFlowProofCookiesSetOptions()` to pass in desired value of `secure` based on the result of `isSSLOrigin()`.

However, the consumer of `createTokenCookiesSetOptions()` missed passing `secure` overrides, so `secure` attribute is always set to true for local development servers. 

Now moving the `isSSLOrigin()` call into `createTokenCookiesSetOptions()` and `createAuthFlowProofCookiesSetOptions()` themselves with requiring a `origin` string can prevent that kind of miss from happening again. I.e. `secure` value is always being determined by `isSSLOrigin()` and there won't be a default fallback. This is working as these code is isolated within the server-side auth, which requires `origin` string to be always supplied.

jjarvisp: Thank for the explanation, this approach makes sense to prevent this miss from happening again.

jjarvisp (approved):

ashika112 (approved):

PR #14242, opened 2025-02-25
feat(data): add publish events over WebSocket

zxl629:
#### Description of changes

Add support to publish events over WebSocket connection for Amplify events client. Customers can now perform the following operations to publish their events.

```ts
// WSS Connect 
const channel = await events.connect("test/events");

// WSS publish events
await channel.publish(
      { key: "my event content" },
      { authMode: "identityPool" }
);

// WSS publish events batch
await channel.publish([1, 2, 3], { authMode: "identityPool" });
```

* Manual testing with a sample app
* E2Es: https://github.com/aws-amplify/amplify-js/actions/runs/13510524961
* Unit test added/ passed

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/src/internals/events/index.ts
```diff
@@ -70,7 +70,7 @@ async function connect(
 	};
 
 	// WS publish is not enabled in the service yet. It will be a follow up feature
```
blakedunson, line 72: should we adjust this comment?

zxl629: Yes, this comment has been removed.

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
```diff
@@ -263,22 +263,43 @@ export abstract class AWSWebSocketProvider {
...
+						cleanup();
 						resolve();
 					}
 
-					if (data.erroredEvents && data.erroredEvents.length > 0) {
-						// TODO: handle errors
+					if (data.errors && data.errors.length > 0) {
+						const errorTypes = data.errors.map((error: any) => error.errorType);
+						cleanup();
+						reject(new Error(`Publish errors: ${errorTypes.join(', ')}`));
```
iartemiev, line 273: Does this mean we'll close the WS connection if there's a publish error? If so, I don't think that's the behavior we want here. Publish errors would ideally be surfaced at the call site, similar to `events.connect`, e.g.
```js
try {
  await channel.publish({ some: 'data' })
} catch (err) {
  console.log('publish error', err)
}
```
This way, WS publish is self contained / decoupled from `channel.subscribe`.

iartemiev (changes requested):

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
zxl629: We will not close the WS connection if there's a publish error. However, there was a bug in the implementation causing `uncaught errors`, it should surface the publish errors at the call site now.

stocaaro (commented):
We discussed offline how I would expect `event.connect` should allow both publish and subscribe until the channel has been closed, after which neither publish nor subscribe should work.  With multi-channel, there are oddities around how this works in this implementation.

This change adds publish events to websockets. There should be tests added for this on the provider tests.

packages/api-graphql/src/Providers/AWSAppSyncEventsProvider/index.ts
```diff
@@ -73,7 +75,11 @@ export class AWSAppSyncEventProvider extends AWSWebSocketProvider {
 		options: AWSAppSyncEventProviderOptions,
 		customUserAgentDetails?: CustomUserAgentDetails,
 	) {
-		super.publish(options, customUserAgentDetails);
+		return super.publish(options, customUserAgentDetails);
+	}
+
+	public closeIfNoActiveChannel() {
```
stocaaro, line 81: This class/abstraction doesn't know about channels and the channel check is on the outer call. Can we rename this, maybe "closeIfNoActiveSubscription"?

stocaaro: Looking around, I suppose we do have channel params in the query params. I still think this is checking subs, not channels.

stocaaro (approved):
LGTM

iartemiev (approved):

PR #14290, opened 2025-03-13
refactor(api-graphql): set observers earlier to ensure handling unsubscription properly

no6v:
#### Description of changes

There are some cases unsubscription messages are not sent properly, e.g. in React (18+) development environment. This PR improves the handling of observers to be set earlier in the process to ensure the unsubscription operation can wait for the subscription is connected even if the `unsubscribe()` is called immediately.

#### Issue #, if available

aws-amplify/amplify-data#554

* added a unit test to verify that an unsubscription message is about to be sent

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
```diff
@@ -158,6 +158,16 @@ export abstract class AWSWebSocketProvider {
 
 			let subscriptionStartInProgress = false;
 			const subscriptionId = amplifyUuid();
+
+			const { query, variables } = options;
+			this.subscriptionObserverMap.set(subscriptionId, {
```
stocaaro, line 163: This behavior got pushed down to update the map during reconnection. Its interesting that removing the lower usage doesn't cause any test failures.

stocaaro:
I updated the PR with the variation I would propose we go with. Thoughts?

stocaaro (approved):

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
no6v: Nice catch. I didn't consider a reconnection situation.

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
```diff
@@ -399,14 +409,6 @@ export abstract class AWSWebSocketProvider {
 		const { additionalCustomHeaders, libraryConfigHeaders } =
 			await additionalHeadersFromOptions(options);
```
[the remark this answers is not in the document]
no6v: Actually, I moved the set call to the location your suggested at first, but I'm not sure when the thread (task?) switching happens.
I confirmed that my app works fine even in development environment with your update.

joon-won (approved):

Issue #14256, opened 2025-02-28
Guest Identity Inconsistency: Server Uses Cookies, Client Uses localStorage


MarekBodingerBA:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [x] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Next.js

### Amplify APIs

Authentication

### Amplify Version

v6

### Backend

None

### Environment information

<details>

```
  System:
    OS: Windows 11 10.0.26100
    CPU: (22) x64 Intel(R) Core(TM) Ultra 9 185H
    Memory: 7.52 GB / 31.43 GB
  Binaries:
    Node: 22.11.0 - ~\AppData\Local\Volta\tools\image\node\22.11.0\node.EXE
    npm: 10.9.2 - ~\AppData\Local\Volta\tools\image\npm\10.9.2\bin\npm.CMD
  Browsers:
    Edge: Chromium (131.0.2903.70)
  npmPackages:
    @ampproject/toolbox-optimizer:  undefined ()
    @aws-amplify/adapter-nextjs: 1.5.1 => 1.5.1
    @aws-amplify/adapter-nextjs/api:  undefined ()
    @aws-amplify/adapter-nextjs/data:  undefined ()
    @aws-sdk/client-cognito-identity: ^3.750.0 => 3.758.0
    @aws-sdk/client-sts: ^3.750.0 => 3.758.0 (3.621.0)
    @babel/core:  undefined ()
    @babel/runtime:  7.22.5
    @edge-runtime/cookies:  6.0.0
    @edge-runtime/ponyfill:  4.0.0
    @edge-runtime/primitives:  6.0.0
    @hapi/accept:  undefined ()
    @mswjs/interceptors:  undefined ()
    @napi-rs/triples:  undefined ()
    @next/font:  undefined ()
    @opentelemetry/api:  undefined ()
    @types/node: ^20 => 20.17.19
    @types/react: ^19 => 19.0.10
    @types/react-dom: ^19 => 19.0.4
    @vercel/nft:  undefined ()
    @vercel/og:  0.6.4
[...]

PR #14326, opened 2025-04-01
fix(aws-amplify): cookie store is not used for identity store in SSR

HuiSF:
#### Description of changes

When Amplify is configured on the client side with library options `{ ssr: true }`, create a new instance of `CognitoAWSCredentialsAndIdentityIdProvider` using cookie store based `DefaultIdentityIdStore` instead of using the default instance that uses local storage.

The unauthenticated ID may be generated from different origin depending on the use case when an unauthenticated identity ID has not been created:

1. when **_client-side_** `fetchAuthSession` gets called **first**, the unauthenticated identity ID gets generated on the client side.
2.  when **_server-side_** `fetchAuthSession` gets called **first** within a `runWithAmplifyServerContext` call specifying `nextServerContext` parameter with request/response or cookies, the unauthenticated identity ID gets generated on the server side, and set in cookie store via server response (for Next.js, this applicable to only responses returned from the middleware, an API route handler, or a Server Action)

Once a unauthenticated identity id is generated and while no end user signs in, regardless its origin, the client side and server side (when calling `runWithAmplifyServerContext` with specifying `nextServerContext`) should share the same identity ID.

**_Note:_** when calling `runWithAmplifyServerContext` with specifying `nextServerContext: null`, an unauthenticated identity ID is created and persisted in memory only on the server side, and this identity ID is not involved in the above two use cases.

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/14256

- unit tests
- manual tests with a Next.js sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/aws-amplify/__tests__/initSingleton.test.ts
```diff
@@ -345,7 +370,6 @@ describe('initSingleton (DefaultAmplify)', () => {
 					expect(
 						mockCognitoUserPoolsTokenProviderSetAuthConfig,
 					).not.toHaveBeenCalled();
-					expect(MockCookieStorage).not.toHaveBeenCalled();
```
HuiSF, line 348: We now always create an instance of CookieStorage for later use, depends on `libraryOptions.ssr`'s value, if it's not used, it's GC'ed eventually.

AllanZhengYP (dismissed):

packages/aws-amplify/src/initSingleton.ts
```diff
@@ -36,6 +38,11 @@ export const DefaultAmplify = {
 		libraryOptions?: LibraryOptions,
 	): void {
 		const resolvedResourceConfig = parseAmplifyConfig(resourceConfig);
+		const cookieBasedKeyValueStorage = new CookieStorage({ sameSite: 'lax' });
+		const cookieBasedCredentialsAndIdentityIdProvider =
+			new CognitoAWSCredentialsAndIdentityIdProvider(
+				new DefaultIdentityIdStore(cookieBasedKeyValueStorage),
+			);
```
calebpollman, line 45: For simpler readability/future maintenance would it make sense to predicate the initialization of client/SSR related configuration values on `libarayOptions.ssr`?  For example:
```ts
// somewhere near line 56
const isSSR = !!libraryOptions?.ssr;
// If ssr is provided through libraryOptions, we should respect the intentional reconfiguration.
const keyValueStorage = !isSSR
	? defaultStorage
	: new CookieStorage({ sameSite: 'lax' });

// somewhere near line 66
const credentialsProvider = !isSSR
	? cognitoCredentialsProvider
	: new CognitoAWSCredentialsAndIdentityIdProvider(
			new DefaultIdentityIdStore(keyValueStorage),
		);

Comment (HuiSF) [reply packages/aws-amplify/src/initSingleton.ts]: Updated 👍

File: packages/aws-amplify/src/initSingleton.ts [hunk 3]
```diff
@@ -36,6 +38,17 @@ export const DefaultAmplify = {
...
+		const resolvedKeyValueStorage = libraryOptions?.ssr
+			? cookieBasedKeyValueStorage
+			: defaultStorage;
+		const cookieBasedCredentialsAndIdentityIdProvider =
+			new CognitoAWSCredentialsAndIdentityIdProvider(
+				new DefaultIdentityIdStore(cookieBasedKeyValueStorage),
+			);
+		const resolvedCredentialsProvider = libraryOptions?.ssr
+			? cookieBasedCredentialsAndIdentityIdProvider
+			: cognitoCredentialsProvider;
```
calebpollman, line 51: Nit: an unknowing contributor could mistakenly provide `cookieBasedCredentialsAndIdentityIdProvider` to client lib options
```suggestion
const resolvedCredentialsProvider = libraryOptions?.ssr
			? new CognitoAWSCredentialsAndIdentityIdProvider(
					new DefaultIdentityIdStore(cookieBasedKeyValueStorage),
				)
			: cognitoCredentialsProvider;
```

calebpollman (approved):

packages/aws-amplify/src/initSingleton.ts
HuiSF: Good catch, updated.

blakedunson (approved):
lgtm

calebpollman (approved):
🚢

ashika112 (approved):
Nice :+1:

PR #14383, opened 2025-05-13
feat(storage): upload with fullbody checksum

AllanZhengYP:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/storage/src/providers/s3/apis/internal/uploadData/multipart/initialUpload.ts
+++ b/packages/storage/src/providers/s3/apis/internal/uploadData/multipart/initialUpload.ts
@@ -137,6 +137,7 @@ export const loadOrCreateMultipartUpload = async ({
 				ContentEncoding: contentEncoding,
 				Metadata: metadata,
 				ChecksumAlgorithm: finalCrc32 ? 'CRC32' : undefined,
+				ChecksumType: finalCrc32 ? 'FULL_OBJECT' : undefined,
 				ExpectedBucketOwner: expectedBucketOwner,
 			},
 		);
--- a/packages/storage/src/providers/s3/utils/client/s3data/createMultipartUpload.ts
+++ b/packages/storage/src/providers/s3/utils/client/s3data/createMultipartUpload.ts
@@ -32,7 +32,8 @@ import { defaultConfig, parseXmlError } from './base';
 export type CreateMultipartUploadInput = Extract<
 	CreateMultipartUploadCommandInput,
 	PutObjectInput
->;
+> &
+	Pick<CreateMultipartUploadCommandInput, 'ChecksumType'>;
 
 export type CreateMultipartUploadOutput = Pick<
 	CreateMultipartUploadCommandOutput,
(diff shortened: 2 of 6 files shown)


packages/storage/src/providers/s3/utils/client/s3data/createMultipartUpload.ts
```diff
@@ -32,7 +32,8 @@ import { defaultConfig, parseXmlError } from './base';
 export type CreateMultipartUploadInput = Extract<
 	CreateMultipartUploadCommandInput,
 	PutObjectInput
->;
+> &
+	Pick<CreateMultipartUploadCommandInput, 'ChecksumType'>;
```
jjarvisp, line 36: sanity check: `ChecksumType` is not part of `PutObjectCommandInput`?

packages/storage/src/providers/s3/apis/internal/uploadData/multipart/initialUpload.ts
```diff
@@ -137,6 +137,7 @@ export const loadOrCreateMultipartUpload = async ({
 				ContentEncoding: contentEncoding,
 				Metadata: metadata,
 				ChecksumAlgorithm: finalCrc32 ? 'CRC32' : undefined,
+				ChecksumType: finalCrc32 ? 'FULL_OBJECT' : undefined,
```
jjarvisp, line 140: mega nit: when creating camel case variable names using `CRC-32` we are not consistent. in some cases we use `CRC` others we use `Crc` e.g. `calculateContentCRC32` vs `finalCrc32`

packages/storage/src/providers/s3/apis/internal/uploadData/multipart/uploadHandlers.ts
```diff
@@ -263,7 +264,9 @@ export const getMultipartUploadHandlers = (
 			},
 		);
 
-		if (size) {
+		// If full-object CRC32 checksum is NOT enabled, we need to ensure the upload integrity by making extra HEAD call
+		// to verify the uploaded object size.
+		if (!inProgressUpload.finalCrc32) {
```
jjarvisp, line 269: It's safe to make the assumption that size > 0 here?

packages/storage/src/providers/s3/utils/client/s3data/createMultipartUpload.ts
AllanZhengYP: It isn't. PutObject always uses full-object checksum.

packages/storage/src/providers/s3/apis/internal/uploadData/multipart/uploadHandlers.ts
AllanZhengYP: The previous `if (size) {}` is not accurate, we should have always HEAD object to verify the integrity. But just to answer question -- yes, the object size has to be > 5MB to enable multipart upload.

packages/storage/src/providers/s3/apis/internal/uploadData/multipart/initialUpload.ts
AllanZhengYP: Look more into it I realize we used `CRC32` in all of internal variable names and file names. However, the cache value we stored is final*Crc32*. Since is already stored in the local storage, it's too late the change name.

jjarvisp (approved):

jjarvisp (approved):

Issue #14378, opened 2025-05-07
Guest identity handling inconsistent with auth flow: throws errors instead of regenerating identities


MarekBodingerBA:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [x] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Next.js

### Amplify APIs

Authentication

### Amplify Version

v6

### Backend

None

### Environment information

<details>

```
  System:
    OS: Windows 11 10.0.26100
    CPU: (22) x64 Intel(R) Core(TM) Ultra 9 185H
    Memory: 4.04 GB / 31.43 GB
  Binaries:
    Node: 20.9.0 - ~\AppData\Local\Volta\tools\image\node\20.9.0\node.EXE
    Yarn: 4.9.1-git.20250411.hash-1908ee79f - ~\AppData\Local\Volta\tools\image\yarn\4.9.1\bin\yarn.CMD
    npm: 10.8.1 - ~\AppData\Local\Volta\tools\image\npm\10.8.1\bin\npm.CMD
    pnpm: 10.10.0 - C:\Program Files\Volta\pnpm.EXE
  Browsers:
    Edge: Chromium (131.0.2903.70)
  npmPackages:
    @ampproject/toolbox-optimizer:  undefined ()
    @aws-amplify/adapter-nextjs: 1.6.2 => 1.6.2 
    @aws-amplify/adapter-nextjs/api:  undefined ()
    @aws-amplify/adapter-nextjs/data:  undefined ()
    @babel/core:  undefined ()
    @babel/runtime:  7.22.5 
    @edge-runtime/cookies:  6.0.0 
    @edge-runtime/ponyfill:  4.0.0 
    @edge-runtime/primitives:  6.0.0
    @grafana/faro-core: ^1.17.1 => 1.17.1
    @grafana/faro-web-sdk: ^1.17.1 => 1.17.1
    @graphql-codegen/cli: 5.0.5 => 5.0.5
    @graphql-codegen/client-preset: 4.7.0 => 4.7.0
[...]

Issue #13787, opened 2024-09-05
When a client made by generateServerClientUsingCookies is used in a server action it mutates cookies


Aldo-Garza:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Next.js

### Amplify APIs

GraphQL API, DataStore

### Amplify Version

v6

### Amplify Categories

storage

### Backend

Amplify Gen 2 (Preview)

### Environment information

<details>

```
# Put output below this line
System:
    OS: Windows 11 10.0.22631
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 2.85 GB / 15.79 GB
  Binaries:
    Node: 20.11.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.2.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (127.0.2651.98)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    @ampproject/toolbox-optimizer:  undefined ()
    @aws-amplify/adapter-nextjs: ^1.2.10 => 1.2.10 
    @aws-amplify/adapter-nextjs/api:  undefined ()
    @aws-amplify/adapter-nextjs/data:  undefined ()
    @aws-amplify/backend: ^1.0.4 => 1.0.4 
    @aws-amplify/backend-cli: ^1.2.2 => 1.2.2 
    @aws-amplify/ui-react: ^6.1.13 => 6.1.14 
    @aws-amplify/ui-react-internal:  undefined ()
    @aws-sdk/s3-presigned-post: ^3.623.0 => 3.623.0
    @aws-sdk/s3-request-presigner: ^3.623.0 => 3.623.0
    @azure/storage-blob: ^12.24.0 => 12.24.0
    @babel/core:  undefined ()
    @babel/runtime:  7.22.5
    @edge-runtime/cookies:  4.1.1
    @edge-runtime/ponyfill:  2.4.2
[...]

PR #14437, opened 2025-06-19
fix: always remove guestIdentityId from keyValueStorage on client-side

adrianjoshua-strutt:
#### Description of changes
Using SSR the guest token in the cookies does not get removed on login. This is because `storeIdentityId` is called on the server-side before it is called on the client-side, thereby setting `_hasGuestIdentityId` to false and hindering the client-side to remove the cookie in the browser. 

We do not always want to remove the guest identity cookie, because we want to prevent unnecessary cookie deletions when switching between identity types or during credential refreshes. (synced about this with @HuiSF - see https://github.com/aws-amplify/amplify-js/issues/13787)

We fix this by 

#### Issue 
https://github.com/aws-amplify/amplify-js/issues/14378

1. Ran next.js app provided by costumer
2. Ran react app to validate that it does not break none SSR flows
3. e2e tests see: https://github.com/aws-amplify/amplify-js/actions/runs/15782549397

#### Checklist
- [x] PR description included
- [x] `yarn test` passes
- [x] Unit Tests are changed

### DIFF
--- a/packages/auth/src/providers/cognito/credentialsProvider/IdentityIdStore.ts
+++ b/packages/auth/src/providers/cognito/credentialsProvider/IdentityIdStore.ts
@@ -7,7 +7,10 @@ import {
 	Identity,
 	KeyValueStorageInterface,
 } from '@aws-amplify/core';
-import { assertIdentityPoolIdConfig } from '@aws-amplify/core/internals/utils';
+import {
+	assertIdentityPoolIdConfig,
+	isBrowser,
+} from '@aws-amplify/core/internals/utils';
 
 import { getAuthStorageKeys } from '../tokenProvider/TokenStore';
 import { AuthKeys } from '../tokenProvider/types';
@@ -80,11 +83,9 @@ export class DefaultIdentityIdStore implements IdentityIdStore {
 		} else {
 			this._primaryIdentityId = identity.id;
 			// Clear locally stored guest id
-			this.keyValueStorage.getItem(this._authKeys.identityId).then(item => {
-				if (item) {
-					this.keyValueStorage.removeItem(this._authKeys.identityId);
-				}
-			});
+			if (this._hasGuestIdentityId || isBrowser()) {
+				this.keyValueStorage.removeItem(this._authKeys.identityId);
+			}
 			this._hasGuestIdentityId = false;
 		}
 	}


soberm (dismissed):

ahmedhamouda78 (dismissed):

sarayev (dismissed):

packages/auth/src/providers/cognito/credentialsProvider/IdentityIdStore.ts
```diff
@@ -80,8 +80,12 @@ export class DefaultIdentityIdStore implements IdentityIdStore {
 		} else {
 			this._primaryIdentityId = identity.id;
 			// Clear locally stored guest id
-			if (this._hasGuestIdentityId) {
-				this.keyValueStorage.removeItem(this._authKeys.identityId);
+			// On the server-side we use the _hasGuestIdentityId flag to avoid caching issues
+			const serverside = typeof window === 'undefined';
+			if (this._hasGuestIdentityId || !serverside) {
+				if (this.keyValueStorage.getItem(this._authKeys.identityId) !== null) {
```
pranavosu: `getItem` returns a Promise, so we need to await it before the null check else it will always be truthy. 

Unrelated to your change, we're not awaiting `removeItem` and `setItem` calls either, which seems wrong since this function will return before those are complete.

packages/auth/src/providers/cognito/credentialsProvider/IdentityIdStore.ts
```diff
@@ -80,8 +80,12 @@ export class DefaultIdentityIdStore implements IdentityIdStore {
 		} else {
 			this._primaryIdentityId = identity.id;
 			// Clear locally stored guest id
-			if (this._hasGuestIdentityId) {
-				this.keyValueStorage.removeItem(this._authKeys.identityId);
+			// On the server-side we use the _hasGuestIdentityId flag to avoid caching issues
+			const serverside = typeof window === 'undefined';
```
pranavosu: minor, but we have an isBrowser utility in amplify/core that we can use here.

packages/auth/src/providers/cognito/credentialsProvider/IdentityIdStore.ts
adrianjoshua-strutt: Thanks for the input! I will investigate. My current understanding is that `storeIdentityId` does not want to `await` the subsequent `keyValueStorage` (remove and get) calls and optimizes for speed. However, of course, not awaiting the `getItem` in my added if-statement is wrong and we should maybe remove that if-statement altogether, as I noticed that `removeItem` can be called, even when the item does not exist.

packages/auth/src/providers/cognito/credentialsProvider/IdentityIdStore.ts
adrianjoshua-strutt: Thanks! I will use that

packages/auth/src/providers/cognito/credentialsProvider/IdentityIdStore.ts
adrianjoshua-strutt: I investigated a bit more and I saw that for `loadIdentityId` we are actually using `await`. This seems to be a bug. I will raise a revision that uses `await` every time we call `storeIdentityId` and uses `await` for calls in the method itself.

adrianjoshua-strutt: I updated the code. We are using `await` in the unit tests now, to properly check the behavior of the method. I also decided to use `then` for non-blocking evaluation of whether the item exsists.

packages/auth/src/providers/cognito/credentialsProvider/IdentityIdStore.ts
adrianjoshua-strutt: I argue that we do not need the isBrowser utility method anymore.

PR #14465, opened 2025-07-11
feat(12820): implement crossTab auth events

bobbor:
#### Description of changes
Introduce the possibility of listening to crossTab auth events.

The changes introduce a listener for storage events on the DefaultStorage (`window.localStorage`) as well as on the SessionStorage (`window.sessionStorage`)

since the logic for adding and removing listeners is on the `KeyValueStorage` it is (theoretically) possible in future to add listeners for crossTab events on CookieStorage and CustomStorage.

since it listens to all storage events, it then filters the event based on token-naming convention to dispatch crossTab Hub events on the "auth" channel.

the typing and implementation of `dispatch` and `listen` on the hub makes sure, that only auth-channel events are allowed for crossTab communication.

#### Issue #, if available
[Issue 12820](https://github.com/aws-amplify/amplify-js/issues/12820)

I added unit-tests and did manual explorative testing.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
  - will be done in a follow-up PR

### DIFF
--- a/packages/auth/__tests__/providers/cognito/tokenProvider/tokenStore.test.ts
+++ b/packages/auth/__tests__/providers/cognito/tokenProvider/tokenStore.test.ts
@@ -1,10 +1,13 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { KeyValueStorageInterface } from '@aws-amplify/core';
-import { decodeJWT } from '@aws-amplify/core/internals/utils';
+import { Hub, KeyValueStorageInterface } from '@aws-amplify/core';
+import { AMPLIFY_SYMBOL, decodeJWT } from '@aws-amplify/core/internals/utils';
 
-import { DefaultTokenStore } from '../../../../src/providers/cognito/tokenProvider';
+import {
+	AUTH_KEY_PREFIX,
+	DefaultTokenStore,
+} from '../../../../src/providers/cognito/tokenProvider';
 
 const userPoolId = 'us-west-1:0000523';
 const userPoolClientId = 'mockCognitoUserPoolsId';
@@ -21,6 +24,9 @@ jest.mock(
 		TokenProviderErrorCode: 'mockErrorCode',
 	}),
 );
+jest.mock('../../../../src/providers/cognito/apis/getCurrentUser', () => ({
+	getCurrentUser: () => Promise.resolve({}),
+}));
 
 const mockedDecodeJWT = jest.mocked(decodeJWT);
 mockedDecodeJWT.mockReturnValue({
@@ -72,6 +78,7 @@ const mockKeyValueStorage: jest.Mocked<KeyValueStorageInterface> = {
 	getItem: jest.fn(),
 	removeItem: jest.fn(),
 	clear: jest.fn(),
+	addListener: jest.fn(),
 };
 
 describe('TokenStore', () => {
@@ -401,6 +408,85 @@ describe('TokenStore', () => {
 
 			const finalTokens = await tokenStore.loadTokens();
 			expect(finalTokens?.refreshToken).toBe(newMockAuthToken.refreshToken);
+		});
+	});
+
+	describe('setupNotify', () => {
+		it('should setup a KeyValueStorageEvent listener', async () => {
+			tokenStore.setupNotify();
+
+			const spy = jest.spyOn(keyValStorage, 'addListener');
+			const hubSpy = jest.spyOn(Hub, 'dispatch');
+
+			expect(spy).toHaveBeenCalledWith(expect.any(Function));
+
+			const listener = spy.mock.calls[0][0];
+
+			// does nothing if key does not match
+			await listener({
+				key: 'foo.bar',
+				oldValue: null,
+				newValue: null,
+			});
+
+			expect(hubSpy).not.toHaveBeenCalled();
+
+			// does nothing if both values are null
+			await listener({
+				key: `${AUTH_KEY_PREFIX}.someid.someotherId.refreshToken`,
+				oldValue: null,
+				newValue: null,
+			});
+
+			expect(hubSpy).not.toHaveBeenCalled();
+
+			// dispatches signedIn on new value
+			await listener({
+				key: `${AUTH_KEY_PREFIX}.someid.someotherId.refreshToken`,
+				newValue: '123',
+				oldValue: null,
+			});
+
+			expect(hubSpy).toHaveBeenCalledWith(
+				'auth',
+				{ event: 'signedIn', data: {} },
+				'Auth',
+				AMPLIFY_SYMBOL,
+				true,
+			);
+			hubSpy.mockClear();
+
+			// dispatches signedOut on null newValue
+			await listener({
+				key: `${AUTH_KEY_PREFIX}.someid.someotherId.refreshToken`,
+				newValue: null,
+				oldValue: '123',
+			});
+
+			expect(hubSpy).toHaveBeenCalledWith(
+				'auth',
+				{ event: 'signedOut' },
+				'Auth',
+				AMPLIFY_SYMBOL,
+				true,
+			);
+			hubSpy.mockClear();
+
+			// dispatches tokenRefresh for changed value
+			await listener({
+				key: `${AUTH_KEY_PREFIX}.someid.someotherId.refreshToken`,
+				newValue: '456',
+				oldValue: '123',
+			});
+
+			expect(hubSpy).toHaveBeenCalledWith(
+				'auth',
+				{ event: 'tokenRefresh' },
+				'Auth',
+				AMPLIFY_SYMBOL,
+				true,
+			);
+			hubSpy.mockClear();
 		});
 	});
 });
--- a/packages/auth/__tests__/providers/cognito/tokenProvider/tokenStore.test.ts
+++ b/packages/auth/__tests__/providers/cognito/tokenProvider/tokenStore.test.ts
@@ -1,7 +1,11 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { Hub, KeyValueStorageInterface } from '@aws-amplify/core';
+import {
+	Hub,
+	KeyValueStorageEvent,
+	KeyValueStorageInterface,
+} from '@aws-amplify/core';
 import { AMPLIFY_SYMBOL, decodeJWT } from '@aws-amplify/core/internals/utils';
 
 import {
@@ -78,7 +82,39 @@ const mockKeyValueStorage: jest.Mocked<KeyValueStorageInterface> = {
 	getItem: jest.fn(),
 	removeItem: jest.fn(),
 	clear: jest.fn(),
-	addListener: jest.fn(),
+	// addListener now returns an unsubscribe function.
+	addListener: jest.fn(() => jest.fn()),
+};
+
+/**
+ * Faithful KeyValueStorage double: `addListener` really tracks listeners and
+ * returns a working unsubscribe, and `emit` fans an event out to them. Lets us
+ * exercise the real TokenStore notify lifecycle without mocking its internals.
+ */
+const createFakeStorage = () => {
+	const listeners = new Set<(ev: KeyValueStorageEvent) => Promise<void>>();
+
+	return {
+		setItem: jest.fn(),
+		getItem: jest.fn(),
+		removeItem: jest.fn(),
+		clear: jest.fn(),
+		addListener: jest.fn(
+			(listener: (ev: KeyValueStorageEvent) => Promise<void>) => {
+				listeners.add(listener);
+
+				return () => {
+					listeners.delete(listener);
+				};
+			},
+		),
+		emit: async (ev: KeyValueStorageEvent) => {
+			for (const listener of listeners) {
+				await listener(ev);
+			}
+		},
+		listenerCount: () => listeners.size,
+	};
 };
 
 describe('TokenStore', () => {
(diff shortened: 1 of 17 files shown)


packages/auth/__tests__/providers/cognito/tokenProvider/tokenStore.test.ts
```diff
@@ -403,4 +410,83 @@ describe('TokenStore', () => {
 			expect(finalTokens?.refreshToken).toBe(newMockAuthToken.refreshToken);
 		});
 	});
+
+	describe('setupNotify', () => {
+		it('should setup a KeyValueStorageEvent listener', async () => {
```
ahmedhamouda78: Can we split this into multiple tests? It will make debugging easier in case of failures

packages/core/__tests__/Hub.test.ts
```diff
@@ -140,4 +140,76 @@ describe('Hub', () => {
...
+				true,
+			);
+
+			expect(crossTabListener).toHaveBeenCalled();
+			expect(sameTabListener).not.toHaveBeenCalled();
+		});
+
+		it('should not allow crossTab dispatch in other channels', () => {
+			Hub.dispatch(
+				// this looks weird but is only used to mute TS.
```
ahmedhamouda78, line 201: 🤔

ahmedhamouda78 (dismissed):
Just one minor comment on one of the tests. Other than that, looks good to me!

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
```diff
@@ -35,13 +43,80 @@ export class DefaultTokenStore implements AuthTokenStore {
...
 	setAuthConfig(authConfig: AuthConfig) {
 		this.authConfig = authConfig;
 	}
 
+	setupNotify() {
+		// Idempotent: a second call while already subscribed is a no-op rather
+		// than a second (leaked) listener.
+		if (this.stopNotify) {
+			return;
+		}
```
osama-rizk, line 66: **[blocking] Dotted usernames disable all cross-tab auth events.**

Parsing the key positionally (`const [key, , , id] = (e.key || '').split('.')`) assumes the username is a single dot-free segment. The real layout is `CognitoIdentityServiceProvider.<clientId>.<username>.<tokenType>`, and Cognito allows dots in usernames (email-style `john.doe`, `a.b@x.com`). For those, index 3 is a *fragment of the username*, so `id === 'refreshToken'` is false and `signedIn`/`signedOut`/`tokenRefresh` are never dispatched to other tabs.

Suggest matching `e.key?.startsWith(AUTH_KEY_PREFIX)` and checking the **last** dot-segment `=== 'refreshToken'` instead of a fixed index.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
```diff
@@ -35,13 +43,80 @@ export class DefaultTokenStore implements AuthTokenStore {
...
+					} else if (newValue === null && oldValue) {
+						Hub.dispatch(
+							'auth',
+							{
+								event: 'signedOut',
+							},
+							'Auth',
+							AMPLIFY_SYMBOL,
+							true,
+						);
```
osama-rizk, line 109: **[major] The `tokenRefresh` cross-tab branch is effectively dead.**

It's keyed on the refreshToken storage key changing (`newValue && oldValue`). For default pools (no refresh-token rotation), `storeTokens` rewrites the **same** refresh-token string on every refresh. Per the HTML storage-event spec, `setItem` with an identical value fires **no** cross-tab `storage` event, so other tabs never receive `tokenRefresh`.

The signal can't detect the event it names — consider hinging detection on the access/id-token key (which does change on refresh).

packages/core/src/storage/KeyValueStorage.ts
```diff
@@ -55,4 +57,44 @@ export class KeyValueStorage implements KeyValueStorageInterface {
...
+		return () => {
+			listeners.delete(listener);
+			// Detach once the last subscriber unsubscribes — real teardown, not
+			// just pruning the set.
+			if (isBrowser() && listeners.size === 0) {
+				window.removeEventListener('storage', this.storageListener, false);
+			}
+		};
+	}
+
```
osama-rizk: **[major] Storage events aren't filtered by `storageArea`.**

Both `DefaultStorage` (localStorage) and `SessionStorage` (sessionStorage) attach to the single window `storage` event, which fires for **any** Storage area on the origin. Because the handler never checks `e.storageArea === this.storage`, a store wrapping one area reacts to changes from the other; on any key-name collision this produces spurious or wrong-source cross-tab dispatches.

Compare `e.storageArea` to `this.storage` before notifying.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
```diff
@@ -35,13 +43,80 @@ export class DefaultTokenStore implements AuthTokenStore {
...
+	setupNotify() {
+		// Idempotent: a second call while already subscribed is a no-op rather
+		// than a second (leaked) listener.
+		if (this.stopNotify) {
+			return;
+		}
+		this.stopNotify = this.keyValueStorage?.addListener?.(
+			async (e: KeyValueStorageEvent) => {
+				const [key, , , id] = (e.key || '').split('.');
+				if (key === AUTH_KEY_PREFIX && id === 'refreshToken') {
```
osama-rizk: **[minor] First sign-in can be dropped when `oldValue` isn't strictly `null`.**

The `signedIn` guard is `newValue && oldValue === null`, while the `KeyValueStorageEvent` contract types `oldValue` as `any`. If an adapter reports `undefined` or `''`, a genuine first sign-in matches neither the `=== null` (signedIn) nor the truthy (tokenRefresh) branch, silently dropping the event. Prefer a falsy check over strict `=== null`.

packages/core/src/storage/KeyValueStorage.ts
```diff
@@ -55,4 +57,44 @@ export class KeyValueStorage implements KeyValueStorageInterface {
...
+	private storageListener = (e: StorageEvent) => {
+		this.listeners?.forEach(listener => {
```
osama-rizk: **[minor] Unhandled promise rejection from the async listener.**

`storageListener` invokes `listener({...})` fire-and-forget, but the TokenStore listener is `async` and awaits `getCurrentUser()`. If that rejects in the other tab during a `signedIn` storage event (transient session/config error), it surfaces as an unhandled rejection — noisy console errors and process-level warnings in strict environments. Wrap the invocation (catch) or have the listener swallow its own errors.

packages/core/src/Hub/index.ts
```diff
@@ -184,10 +217,16 @@ export class HubClass {
 	private _toListeners<Channel extends AmplifyChannel | string>(
 		capsule: HubCapsule<Channel, EventDataMap | AmplifyEventData[Channel]>,
 	) {
-		const { channel, payload } = capsule;
+		const { channel, payload, crossTab } = capsule;
 		const holder = this.listeners.get(channel);
 		if (holder) {
-			holder.forEach(listener => {
+			const crossTabHolders = holder.filter(listener => {
+				const sameCrossTabSpec = !!crossTab === !!listener.crossTab;
```
osama-rizk: **[design/footgun] `crossTab` listener partitioning silently drops same-tab events.**

`_toListeners` keeps only listeners where `!!crossTab === !!listener.crossTab`. A consumer who adds `{ enableCrossTabEvents: true }` to an existing `Hub.listen('auth', cb)` will silently **stop** receiving same-tab `signedIn`/`signedOut`/`tokenRefresh` (those dispatch with falsy `crossTab`). Getting both now requires two separate listeners, and nothing in the types or docs signals this. Worth either documenting prominently or having cross-tab listeners also receive same-tab events.

osama-rizk (commented):
Automated review of the cross-tab auth events. Six findings below, ordered by severity — #1 looks like a blocker; #2/#3 are functional gaps vs. what the feature advertises. Happy to discuss any of them.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
```diff
@@ -35,13 +43,80 @@ export class DefaultTokenStore implements AuthTokenStore {
...
 
+	setupNotify() {
+		// Idempotent: a second call while already subscribed is a no-op rather
+		// than a second (leaked) listener.
+		if (this.stopNotify) {
+			return;
+		}
+		this.stopNotify = this.keyValueStorage?.addListener?.(
+			async (e: KeyValueStorageEvent) => {
+				const [key, , , id] = (e.key || '').split('.');
```
soberm: [major] Positional key parse silently drops cross-tab events for dotted usernames.

`const [key, , , id] = (e.key || '').split('.')` assumes the refreshToken key is exactly 4 dot-separated segments: PREFIX, clientId, username, refreshToken. But the key is built from `createKeysForAuthStorage(AUTH_KEY_PREFIX, clientId + '.' + lastAuthUser)`, where `lastAuthUser` is the Cognito username -- which commonly contains `.` (email usernames, aliases, some federated identities). When it does, `refreshToken` shifts past index 3, `id` is no longer `'refreshToken'`, the guard fails, and no cross-tab event ever fires for those users -- a silent failure for a large user segment.

Suggest matching by prefix/suffix instead of position:
```ts
const key = e.key || '';
if (key.startsWith(`${AUTH_KEY_PREFIX}.`) && key.endsWith('.refreshToken')) {
  // ...
}
```
And add a test with a dotted username (e.g. a key ending in `.clientId.first.last@example.com.refreshToken`) asserting an event fires.

packages/core/src/storage/KeyValueStorage.ts
```diff
@@ -55,4 +57,44 @@ export class KeyValueStorage implements KeyValueStorageInterface {
...
+			listener({
```
soberm: [major] Floating promise -- listener rejections are unhandled.

`storageListener` invokes each listener inside `forEach` and ignores the returned `Promise<void>`. The TokenStore listener is `async` and awaits `getCurrentUser()`, which can reject (e.g. no current user / network). An unhandled rejection is noise in the browser and can crash under Node's `unhandledRejection` handler.

Suggest catching:
```ts
this.listeners?.forEach(listener => {
  Promise.resolve(
    listener({ key: e.key, oldValue: e.oldValue, newValue: e.newValue }),
  ).catch(err => { /* logger.error(err) */ });
});
```

packages/core/src/Hub/index.ts
soberm: [minor] Cross-tab listeners are *partitioned*, not additive. `sameCrossTabSpec = !!crossTab === !!listener.crossTab` means a listener registered with `enableCrossTabEvents: true` receives ONLY cross-tab events and will miss same-tab auth events (signedIn/signedOut/tokenRefresh dispatched in the same tab). A consumer wanting both has to register twice. If that's the intended contract, please document it on `Hub.listen`'s `enableCrossTabEvents` option; otherwise consider making cross-tab listeners a superset (receive both).

(Existing non-crossTab listeners are unaffected: a normal dispatch has `crossTab === undefined` and a normal listener has `crossTab === false`, so both are falsy and they still fire -- verified.)

packages/core/src/types/storage.ts
```diff
@@ -28,3 +37,9 @@ export interface SyncStorage {
 	removeItem(key: string): void;
 	clear(): void;
 }
+
+export interface KeyValueStorageEvent {
+	readonly key: string | null;
+	readonly oldValue: any;
```
soberm: [minor] `oldValue`/`newValue` typed as `any` defeats type-checking for consumers. The underlying `StorageEvent` fields are `string | null`, and `storageListener` forwards them verbatim. Suggest:
```ts
readonly oldValue: string | null;
readonly newValue: string | null;
```

packages/core/src/Hub/index.ts
```diff
@@ -152,17 +172,27 @@ export class HubClass {
...
-		listenerName = 'noname',
+		options:
+			| string
+			| { listenerName?: string; enableCrossTabEvents?: boolean } = {
+			listenerName: 'noname',
+		},
 	): StopListenerCallback {
-		let cb: HubCallback<string, EventDataMap>;
+		let cb: HubCallback;
+		let o;
```
soberm: [minor] `let o;` is implicitly `any`, forcing the repeated `(o as { listenerName: string; enableCrossTabEvents: boolean })` casts below. Declare the type once so the casts drop out:
```ts
const o: { listenerName: string; enableCrossTabEvents: boolean } =
  typeof options === 'string'
    ? { listenerName: options, enableCrossTabEvents: false }
    : { listenerName: 'noname', enableCrossTabEvents: false, ...options };
```

packages/core/src/Hub/types/HubTypes.ts
```diff
@@ -8,6 +8,7 @@ export type IListener<
 	EventData extends EventDataMap = EventDataMap,
 > = {
 	name: string;
+	crossTab: boolean;
```
soberm: [nit] `crossTab: boolean` is added as a *required* field on the exported `IListener` type. The only in-repo construction site (`Hub.listen`) is updated so nothing breaks internally, but a required field on an exported type is technically a breaking change for any external code that constructs an `IListener`. Consider `crossTab?: boolean`.

soberm (commented):
Reviewed the cross-tab auth events implementation. Solid, well-tested feature with good SSR guarding (window access lazily guarded by `isBrowser()`) and clean listener lifecycle (idempotent setup, teardown, re-register on storage swap). No self-trigger loop (storage events only fire in other tabs; `getCurrentUser()` doesn't write `refreshToken`), and the `signInWithRedirectStore` `name` -> `AUTH_KEY_PREFIX` swap is behavior-preserving.

Two correctness items are worth addressing before merge (both small, localized): the positional key parse drops cross-tab events for usernames containing dots, and the fire-and-forget async listener has no rejection handling. Remaining comments are type-tightening and test-coverage suggestions. Details inline.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
bobbor: Fixed in 0bd140fcb — key matching is now `startsWith(`${AUTH_KEY_PREFIX}.`)` + `endsWith('.refreshToken')`/`endsWith('.accessToken')`, no positional split. Added a dotted/email-username test.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
bobbor: Fixed in 0bd140fcb — `tokenRefresh` now hinges on the accessToken key (`oldValue && newValue && oldValue !== newValue`); refreshToken transitions drive only signedIn/signedOut. Tests cover no-dispatch on sign-in's null→value accessToken and exactly one tokenRefresh under rotation (both keys changing).

packages/core/src/storage/KeyValueStorage.ts
bobbor: Fixed in 9407043c4 — the handler early-returns unless `e.storageArea === this.storage`, with match/mismatch tests for both DefaultStorage and SessionStorage.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
bobbor: Fixed in 0bd140fcb — guards are now truthy/falsy (`newValue && !oldValue`), with tests for `undefined` and `''` as first-sign-in oldValue.

packages/core/src/storage/KeyValueStorage.ts
bobbor: Fixed in 9407043c4 — each listener call is wrapped in `Promise.resolve(...).catch()` logging via ConsoleLogger; a rejecting listener no longer surfaces as an unhandled rejection and doesn't stop the fan-out.

packages/core/src/Hub/index.ts
bobbor: Went with the superset in 9407043c4: same-tab dispatches now reach ALL listeners (incl. cross-tab-enabled ones); cross-tab dispatches reach only `enableCrossTabEvents` listeners on 'auth'. JSDoc on the option documents the contract.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
bobbor: Fixed in 0bd140fcb exactly as suggested (prefix + suffix match), incl. the dotted-username test asserting the event fires.

packages/core/src/storage/KeyValueStorage.ts
bobbor: Fixed in 9407043c4 with your suggested shape — `Promise.resolve(listener(...)).catch(err => logger.error(...))` per listener.

packages/core/src/Hub/index.ts
bobbor: Made it a superset in 9407043c4 (cross-tab listeners also receive same-tab events; plain listeners unaffected) and documented it in JSDoc on `enableCrossTabEvents`.

packages/core/src/types/storage.ts
bobbor: Fixed in 9407043c4 — `readonly oldValue: string | null; readonly newValue: string | null`.

packages/core/src/Hub/index.ts
bobbor: Fixed in 9407043c4 — `o` is a typed const via ternary, casts at the push site removed.

packages/core/src/Hub/types/HubTypes.ts
bobbor: Fixed in 9407043c4 — `crossTab?: boolean` is optional on IListener.

packages/auth/__tests__/providers/cognito/tokenProvider/tokenStore.test.ts
bobbor: Done in 0bd140fcb — split into focused it-cases (non-matching key, both-null no-op, signedIn, signedOut, tokenRefresh via accessToken, dotted username, falsy-oldValue variants) sharing setup in beforeEach.

packages/core/__tests__/Hub.test.ts
bobbor: This test changed intentionally in 9407043c4: cross-tab listeners are now a superset (also receive same-tab events), per the partition-vs-superset feedback from soberm/osama-rizk — the assertion now encodes the new contract.

soberm (approved):

osama-rizk (approved):

Issue #14455, opened 2025-07-03
instanceof Request check in @aws-amplify/adapter-nextjs breaks on AWS Lambda (OpenNext/SST)


thrawny:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [x] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Next.js

### Amplify APIs

Authentication

### Amplify Version

v6

### Amplify Categories

auth

### Backend

Other

### Environment information

 System:
    OS: macOS 15.5
    CPU: (8) arm64 Apple M3
    Memory: 114.50 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.3.0 - ~/.asdf/installs/nodejs/v24.3.0/bin/node
    Yarn: 1.22.22 - ~/.asdf/installs/nodejs/v24.3.0/bin/yarn
    npm: 11.4.2 - ~/.asdf/plugins/nodejs/shims/npm
    pnpm: 10.12.4 - ~/.asdf/installs/nodejs/v24.3.0/bin/pnpm
  Browsers:
    Chrome: 138.0.7204.50
    Safari: 18.5
  npmPackages:
    @ampproject/toolbox-optimizer:  undefined ()
    @aws-amplify/adapter-nextjs: ^1.6.4 => 1.6.4 
    @aws-amplify/adapter-nextjs/api:  undefined ()
    @aws-amplify/adapter-nextjs/data:  undefined ()
    @aws-amplify/auth: ^6.13.1 => 6.13.1 
    @aws-amplify/auth/cognito:  undefined ()
    @aws-amplify/auth/cognito/server:  undefined ()
    @aws-amplify/auth/enable-oauth-listener:  undefined ()
    @aws-amplify/auth/server:  undefined ()
    @aws-sdk/client-cognito-identity-provider: ^3.839.0 => 3.839.0 
    @babel/core:  undefined ()
    @babel/runtime:  7.22.5 
    @edge-runtime/cookies:  6.0.0 
    @edge-runtime/ponyfill:  4.0.0 
[...]

PR #14466, opened 2025-07-14
fix: next request check for serverless environments

soberm:
#### Description of changes

The instanceof Request type guard check fails in AWS Lambda environments (specifically with OpenNext/SST 
deployments). The problem causes authentication API routes to work on cold Lambda starts but fail on subsequent warm invocations with a 500 error: "Invalid request and context/response 
combination. The request cannot be handled." This PR changes the check to use duck typing to prevent possible constructor references to differ between invocations.

#### Issue #, if available
[14455](https://github.com/aws-amplify/amplify-js/issues/14455)

* Units tests
* Deployed sample app and verified manually
* [E2E test](https://github.com/aws-amplify/amplify-js/actions/runs/16265824550) run successfull
**  Failing tests unrelated

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/adapter-nextjs/__tests__/auth/utils/predicates.test.ts
+++ b/packages/adapter-nextjs/__tests__/auth/utils/predicates.test.ts
@@ -39,4 +39,124 @@ describe('isNextRequest', () => {
 
 		expect(isNextRequest(request)).toBe(true);
 	});
+
+	it('returns true when the request is like a next request', () => {
+		const mockNextRequest = {
+			nextUrl: {
+				pathname: '/api/auth',
+				search: '?param=value',
+				searchParams: new URLSearchParams('param=value'),
+				href: 'https://example.com/api/auth?param=value',
+			},
+			cookies: {
+				get: jest.fn(),
+				set: jest.fn(),
+				delete: jest.fn(),
+			},
+			url: 'https://example.com/api/auth?param=value',
+			headers: new Headers({
+				'content-type': 'application/json',
+				'user-agent': 'test-agent',
+			}),
+			method: 'POST',
+			body: null,
+			bodyUsed: false,
+		};
+
+		expect(isNextRequest(mockNextRequest)).toBe(true);
+	});
+
+	it('returns false for regular Request objects without NextRequest properties', () => {
+		const regularRequest = {
+			headers: new Headers(),
+			method: 'GET',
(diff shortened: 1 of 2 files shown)


packages/adapter-nextjs/src/auth/utils/predicates.ts
```diff
@@ -10,7 +10,25 @@ import { AuthRoutesHandlerContext } from '../types';
 export function isNextRequest(request: object): request is NextRequest {
```
adrianjoshua-strutt, line 10: Would it be better to remove the `as any` cast? I checked and it should be equivalent [code snippet](https://www.programiz.com/online-compiler/78MUI0hRkD8cd)

```
export function isNextRequest(request: object): request is NextRequest {
  return (
    typeof request === 'object' &&
    request !== null &&
    'nextUrl' in request &&
    typeof request.nextUrl === 'object' &&
    request.nextUrl !== null &&
    'cookies' in request &&
    'url' in request &&
    typeof request.url === 'string' &&
    'headers' in request &&
    'method' in request &&
    typeof request.method === 'string'
  );
}
```

packages/adapter-nextjs/__tests__/auth/utils/predicates.test.ts
```diff
@@ -39,4 +39,124 @@ describe('isNextRequest', () => {
 
 		expect(isNextRequest(request)).toBe(true);
 	});
+
```
adrianjoshua-strutt, line 42: While implicitly it is already covered, maybe we could add a check for an empty object

```
it('returns false for empty object', () => {
    expect(isNextRequest({})).toBe(false);
});
```

packages/adapter-nextjs/src/auth/utils/predicates.ts
soberm: Thx. Updated it.

packages/adapter-nextjs/__tests__/auth/utils/predicates.test.ts
soberm: Since this is already covered by all other test cases I think we can skip it.

adrianjoshua-strutt (approved):

pranavosu (approved):

PR #14478, opened 2025-07-22
chore: update token revocation tests to amplify v6

adrianjoshua-strutt:
Rewrote the token revocation tests to work with V6 Amplify's new architecture. The old tests were breaking because they relied on V5's exposed internal APIs that no longer exist.

This PR just adds the headed tests.

For the test implementation see: https://github.com/aws-amplify/amplify-js-samples-staging/pull/1087

### DIFF
--- a/.github/integ-config/integ-all-headless.yml
+++ b/.github/integ-config/integ-all-headless.yml
@@ -1,7 +1,2 @@
 tests:
   []
-  # integ_auth_test_cypress_no_ui
-  # - test_name: 'Token revocation'
-  #   category: 'auth'
-  #   spec: 'token-revocation'
-  #   is_headless: true


.github/integ-config/integ-all.yml
sarayev, line 1: question: Should we remove the old token-revocation tests from the [integ-all-headless](https://github.com/aws-amplify/amplify-js/blob/main/.github/integ-config/integ-all-headless.yml)

sarayev, line 1: question: Do we need to run those tests as canary as well

adrianjoshua-strutt: They were not part of the canaries during V5. As this change is mostly related to feature parity between V5 and V6 I would suggest not including them into the canaries (see https://github.com/aws-amplify/amplify-js/commit/46c2d2eab9d0bf11b07a38292da52ee1d54b65bb)

adrianjoshua-strutt: Yes. I will raise a revision.

PR #14491, opened 2025-07-31
fix: account for EVENT_SUBSCRIBE_ERROR in AppSync

adrianjoshua-strutt:
### Problem
AppSync Events subscribe_error messages (like authorization failures) weren't being handled, so error callbacks never fired. Subscriptions failed silently.

### Solution
Added EVENT_SUBSCRIBE_ERROR to the error handling logic so these messages properly trigger the observer's error callback.

Before:
```typescript
const sub = channel.subscribe({
  error: (err) => console.error('error', err), // Never called
})
```

After:
```typescript
const sub = channel.subscribe({
  error: (err) => console.error('error', err), // Now properly called on subscribe errors
})
```

e2e, yarn test

#### Checklist

- [ x] PR description included
- [x ] `yarn test` passes
- [ x] Unit Tests are added

### DIFF
--- a/packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
+++ b/packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
@@ -704,7 +704,7 @@ export abstract class AWSWebSocketProvider {
 
 				if (type === MESSAGE_TYPES.EVENT_SUBSCRIBE_ERROR) {
 					const { errors } = JSON.parse(String(message.data));
-					if (errors && Array.isArray(errors) && errors.length > 0) {
+					if (Array.isArray(errors) && errors.length > 0) {
 						const error = errors[0];
 						errorMessage = `${error.errorType}: ${error.message}`;
 					}
--- a/packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
+++ b/packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
@@ -684,7 +684,10 @@ export abstract class AWSWebSocketProvider {
 			return;
 		}
 
-		if (type === MESSAGE_TYPES.GQL_ERROR) {
+		if (
+			type === MESSAGE_TYPES.GQL_ERROR ||
+			type === MESSAGE_TYPES.EVENT_SUBSCRIBE_ERROR
+		) {
 			const subscriptionState = SUBSCRIPTION_STATUS.FAILED;
 			if (observer) {
 				this.subscriptionObserverMap.set(id, {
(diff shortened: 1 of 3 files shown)


packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
```diff
@@ -697,15 +700,23 @@ export abstract class AWSWebSocketProvider {
...
 				});
 
-				this.logger.debug(
-					`${CONTROL_MSG.CONNECTION_FAILED}: ${JSON.stringify(payload ?? data)}`,
-				);
+				let errorMessage = JSON.stringify(payload ?? data);
+
+				if (type === MESSAGE_TYPES.EVENT_SUBSCRIBE_ERROR) {
+					const { errors } = JSON.parse(String(message.data));
+					if (errors && Array.isArray(errors) && errors.length > 0) {
```
soberm: You can simplify this to `if (Array.isArray(errors) && errors.length > 0)`. `Array.isArray` returns false if errors is null or undefined.

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
```diff
@@ -697,15 +700,23 @@ export abstract class AWSWebSocketProvider {
 					subscriptionState,
 				});
 
-				this.logger.debug(
-					`${CONTROL_MSG.CONNECTION_FAILED}: ${JSON.stringify(payload ?? data)}`,
-				);
+				let errorMessage = JSON.stringify(payload ?? data);
+
+				if (type === MESSAGE_TYPES.EVENT_SUBSCRIBE_ERROR) {
+					const { errors } = JSON.parse(String(message.data));
```
soberm, line 706: `JSON.parse` can potentially throw an exception when the error message is malformed.

soberm (dismissed):

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
adrianjoshua-strutt: At the beginning of the method we are calling `const [isData, data] = this._handleSubscriptionData(message);`. All implementations of `_handleSubscriptionData` in our codebase already call `JSON.stringify` on the `message.data` (See: [AWSAppSyncEventProvider](https://github.com/aws-amplify/amplify-js/blob/2212704a2a35b40e60984a0a79770a4b59aae52d/packages/api-graphql/src/Providers/AWSAppSyncEventsProvider/index.ts#L168) and [AWSAppSyncRealTimeProvider](https://github.com/aws-amplify/amplify-js/blob/2212704a2a35b40e60984a0a79770a4b59aae52d/packages/api-graphql/src/Providers/AWSAppSyncRealTimeProvider/index.ts#L134)). So if the JSON would be malformed, we would see the error even without the newly added line. I will investigate if the method is guarded by a try/catch somewhere, or if we need to add it.

A bit off topic. I tried avoiding calling `JSON.parse` two times, as it is already called in `_handleSubscriptionData`. However, unfortunately, I would need to change the return signature of `_handleSubscriptionData` to receive the `errors` array.

packages/api-graphql/src/Providers/AWSWebSocketProvider/index.ts
adrianjoshua-strutt: Thanks! I will edit it

Issue #14421, opened 2025-06-06
Expo / React-Native [Android-Only]: `uploadData` function calculating `onProgress` incorrectly.


ChristopherGabba:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [x] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

Storage

### Amplify Version

v6

### Amplify Categories

storage

### Backend

Amplify Gen 2

### Environment information

<details>

```

  System:
    OS: macOS 15.5
    CPU: (10) arm64 Apple M2 Pro
    Memory: 571.27 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 23.4.0 - ~/.nvm/versions/node/v23.4.0/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 11.3.0 - ~/.nvm/versions/node/v23.4.0/bin/npm
    bun: 1.2.10 - /opt/homebrew/bin/bun
    Watchman: 2025.04.14.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 137.0.7151.69
    Safari: 18.5
  npmPackages:
    %name%:  0.1.0 
    @aws-amplify/backend: ^1.11.0 => 1.16.1 
    @aws-amplify/backend-cli: ^1.7.0 => 1.7.2 
    @aws-amplify/react-native: ^1.1.6 => 1.1.10 
    @aws-amplify/rtn-web-browser: ^1.1.3 => 1.1.4 
    @aws-appsync/utils: ^2.0.3 => 2.0.3 
    @aws-sdk/client-cognito-identity-provider: ^3.799.0 => 3.826.0 
    @aws-sdk/client-dynamodb: ^3.799.0 => 3.826.0 
    @aws-sdk/client-sso-oidc: ^3.799.0 => 3.825.0 (3.622.0, 3.637.0, 3.624.0, 3.621.0)
[...]

PR #14518, opened 2025-08-18
fix(storage): upload progress exceeds total in Expo/RN

pranavosu:
#### Description of changes

This PR attempts to mitigate an issue caused by a bug in Expo 53 ( https://github.com/expo/expo/issues/37526 ) by clamping the values for `event.loaded` so as to not exceed the total, while we dig deeper to investigate the root cause in Expo for Android.

Note: This can be reproduced [without using Amplify](https://github.com/pranavosu/expo-bug-onprogress-repro).

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/14421

Tested locally using an Android simulator, and added unit tests.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/storage/src/providers/s3/utils/client/runtime/xhrTransferHandler.ts
+++ b/packages/storage/src/providers/s3/utils/client/runtime/xhrTransferHandler.ts
@@ -190,7 +190,11 @@ export const xhrTransferHandler: TransferHandler<
 const convertToTransferProgressEvent = (
 	event: ProgressEvent,
 ): TransferProgressEvent => ({
-	transferredBytes: event.loaded,
+	// `loaded` can exceed the `total` in some cases due to platform issues/bugs e.g. React Native & Expo Android
+	// clamp `loaded` values down to `total` if possible.
+	transferredBytes: event.lengthComputable
+		? Math.min(event.loaded, event.total)
+		: event.loaded,
 	totalBytes: event.lengthComputable ? event.total : undefined,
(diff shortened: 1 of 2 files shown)


packages/storage/src/providers/s3/utils/client/runtime/xhrTransferHandler.ts
```diff
@@ -190,7 +190,11 @@ export const xhrTransferHandler: TransferHandler<
 const convertToTransferProgressEvent = (
 	event: ProgressEvent,
 ): TransferProgressEvent => ({
-	transferredBytes: event.loaded,
+	// `loaded` can exceed the `total` in some cases due to platform issues/bugs e.g. React Native & Expo Android
+	// clamp `loaded` values down to `total` if possible.
```
soberm, line 194: Will this create a new issue where we report the download as finished while its still downloading due to loaded exceeding total too early?

pranavosu: There is a separate event to handle [completion](https://github.com/aws-amplify/amplify-js/pull/14518/files#diff-34a89762b620e96a87412d2548e26fac58d7aea66666a8c6662d942af255e9abR113), so we still wait for a final 'ack' from the server in this case before resolving the promise.

ahmedhamouda78 (approved):

PR #14519, opened 2025-08-19
chore: enable retry for e2e tests

Simone319:
#### Description of changes

Enable retries for e2e tests, as well as yarn install step for sample staging repo, which fails occasionally.

Run e2e tests with the updated github workflow: https://github.com/aws-amplify/amplify-js/actions/runs/17043975986

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/.github/actions/setup-samples-staging/action.yml
+++ b/.github/actions/setup-samples-staging/action.yml
@@ -49,7 +49,7 @@ runs:
       working-directory: ./amplify-js-samples-staging
       run: |
         for i in {1..3}; do
-          echo "Starting $i attempt."
+          echo "Starting attempt $i."
           yarn && break
           echo "Attempt $i failed."
           sleep 5
--- a/.github/actions/setup-samples-staging/action.yml
+++ b/.github/actions/setup-samples-staging/action.yml
@@ -45,7 +45,12 @@ runs:
         fi
 
     - name: Install
-      run: |
-        yarn
       shell: bash
       working-directory: ./amplify-js-samples-staging
+      run: |
+        for i in {1..3}; do
+          echo "Starting $i attempt."
+          yarn && break
+          echo "Attempt $i failed."
+          sleep 5
+        done
--- a/.github/integ-config/integ-all.yml
+++ b/.github/integ-config/integ-all.yml
@@ -536,6 +536,7 @@ tests:
     category: auth
     sample_name: [javascript-auth]
     spec: functional-auth
+    retry_count: 3
     browser: *minimal_browser_list
     backend: gen2
   - test_name: integ_react_auth_1_guest_to_authenticated_user_gen2
@@ -544,6 +545,7 @@ tests:
     category: auth
     sample_name: [guest-to-auth-user]
     spec: guest-to-auth-user
+    retry_count: 3
     browser: *minimal_browser_list
     backend: gen2
   - test_name: integ_react_typescript_authentication_gen2
@@ -552,6 +554,7 @@ tests:
     category: auth
     sample_name: [typescript-auth]
     spec: functional-auth
+    retry_count: 3
     browser: *minimal_browser_list
     backend: gen2
   - test_name: integ_react_auth_2_sign_in_after_sign_up_gen2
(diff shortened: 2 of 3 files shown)


.github/actions/setup-samples-staging/action.yml
```diff
@@ -45,7 +45,12 @@ runs:
         fi
 
     - name: Install
-      run: |
-        yarn
       shell: bash
       working-directory: ./amplify-js-samples-staging
+      run: |
+        for i in {1..3}; do
+          echo "Starting $i attempt."
```
ahmedhamouda78: nit: this will output "Starting 2 attempt.", which sounds confusing. Maybe something like this would be better

"Starting attempt $i." i.e "Starting attempt 2." which sounds much better

ahmedhamouda78 (approved):

.github/actions/setup-samples-staging/action.yml
```diff
@@ -45,7 +45,12 @@ runs:
...
+          yarn && break
+          echo "Attempt $i failed."
+          sleep 5
+        done
```
bobbor, line 56: do we have issues, where installing project dependencies fails or times out? and a reinstall fixes it?

.github/integ-config/integ-all.yml
```diff
@@ -15,293 +15,335 @@ tests:
     category: datastore
     sample_name: owner-based-default
     spec: owner-based-default
+    retry_count: 3
```
sarayev: Is there any timeout per test or github action that we have to adjust to accommodate having those many retries?

.github/integ-config/integ-all.yml
```diff
@@ -15,293 +15,335 @@ tests:
...
     spec: static-user-pool-groups-operations
+    retry_count: 3
     browser: *minimal_browser_list
   - test_name: integ_datastore_auth-owner-and-group-different-models-default
     desc: 'DataStore Auth'
     framework: react
     category: datastore
     sample_name: owner-and-group-different-models-default
     spec: owner-and-group-different-models-default
+    retry_count: 3
```
adrianjoshua-strutt: Could we omit the default value? I think we already have a default value of `3` implemented in https://github.com/aws-amplify/amplify-js/blob/main/.github/workflows/callable-e2e-tests.yml#L45

adrianjoshua-strutt (changes requested):

.github/actions/setup-samples-staging/action.yml
Simone319: Yes, we occasionally have 5xx during install. See https://github.com/aws-amplify/amplify-js/actions/runs/17037734366/job/48294300286

.github/integ-config/integ-all.yml
Simone319: Good catch, we can take advantage of the default value.

.github/integ-config/integ-all.yml
Simone319: The default timeout is 35 minutes, and we don't have any tests that run over this limit even with retries, except for 1 test. I can increase the timeout for that.

.github/integ-config/integ-all.yml
```diff
@@ -1021,6 +1034,8 @@ tests:
     category: auth
     sample_name: [mfa]
     spec: refresh-token-auth
+    retry_count: 15
+    timeout_minutes: 90
```
adrianjoshua-strutt, line 1038: Do we know how that interacts with https://github.com/aws-amplify/amplify-js-samples-staging/pull/1091/files?

Simone319: In GitHub Actions, the timeout-minutes setting defines the maximum number of minutes a job or a step is allowed to run before GitHub automatically cancels it. That includes multiple times of retries within the step.

While the timeout in https://github.com/aws-amplify/amplify-js-samples-staging/pull/1091/files should be the timeout when running individual test.

ahmedhamouda78: The timeout from my PR is for Cypress commands. The timeout here is for the whole E2E test as defined in the workflow file.

ahmedhamouda78 (approved):

Issue #14517, opened 2025-08-18
TypeScript type mismatch: AWSAuthDevice.attributes definition doesn't match fetchDevices() runtime structure


kturczyski-lz:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [x] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React

### Amplify APIs

Authentication

### Amplify Version

v6

### Amplify Categories

auth

### Backend

None

### Environment information

<details>

```
# Put output below this line

 System:
    OS: macOS 15.6
    CPU: (12) arm64 Apple M2 Pro
    Memory: 84.53 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.20.5 - ~/.nvm/versions/node/v18.20.5/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v18.20.5/bin/npm
  Browsers:
    Chrome: 139.0.7258.128
    Edge: 139.0.3405.102
    Safari: 18.6
 npmPackages:
    aws-amplify: 6.15.5 => 6.15.5
    aws-amplify/adapter-core:  undefined ()
    aws-amplify/adapter-core/internals:  undefined ()
    aws-amplify/analytics:  undefined ()
    aws-amplify/analytics/kinesis:  undefined ()
    aws-amplify/analytics/kinesis-firehose:  undefined ()
    aws-amplify/analytics/personalize:  undefined ()
    aws-amplify/analytics/pinpoint:  undefined ()
    aws-amplify/api:  undefined ()
    aws-amplify/api/internals:  undefined ()
    aws-amplify/api/server:  undefined ()
    aws-amplify/auth:  undefined ()
    aws-amplify/auth/cognito:  undefined ()
    aws-amplify/auth/cognito/server:  undefined ()
    aws-amplify/auth/enable-oauth-listener:  undefined ()
    aws-amplify/auth/server:  undefined ()
    aws-amplify/data:  undefined ()
    aws-amplify/data/server:  undefined ()
    aws-amplify/datastore:  undefined ()
[...]

PR #14520, opened 2025-08-19
fix: device attributes type

soberm:
#### Description of changes

The runtime structure returned by `fetchDevices()` contains device attributes as a plain object with keys like `device_status`, `device_name`, and `last_ip_used`, but the TypeScript definition incorrectly defines attributes as `AuthUserAttribute<UserAttributeKey>` (a single object with `attributeKey` and `value` properties). This PR updates the type for device attributes.

#### Issue #, if available
#14517 

Manually verified that the types are correct.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/auth/src/providers/cognito/types/models.ts
+++ b/packages/auth/src/providers/cognito/types/models.ts
@@ -10,7 +10,6 @@ import {
 	AWSAuthUser,
 	AuthCodeDeliveryDetails,
 	AuthDevice,
-	AuthUserAttribute,
 } from '../../../types';
 import { AuthProvider } from '../../../types/inputs';
 
@@ -89,11 +88,14 @@ export type AutoSignInEventData =
 	| {
 			event: 'autoSignIn';
 	  };
+
+type DeviceAttributeKey = 'device_status' | 'device_name' | 'last_ip_used';
+
 /**
  * Holds the device specific information along with it's id and name.
  */
 export type AWSAuthDevice = AuthDevice & {
-	attributes: AuthUserAttribute<UserAttributeKey>;
+	attributes: Record<DeviceAttributeKey, string>;
 	createDate?: Date;
 	lastAuthenticatedDate?: Date;
 	lastModifiedDate?: Date;


packages/auth/src/providers/cognito/types/models.ts
```diff
@@ -89,11 +88,14 @@ export type AutoSignInEventData =
 	| {
 			event: 'autoSignIn';
 	  };
+
+type DeviceAttributeKey = 'device_status' | 'device_name' | 'last_ip_used';
```
kturczyski-lz, line 92: Thanks a lot @soberm  for looking into this so quickly. It looks like this may not be the full list of predefined attributes (based on the existing type). There also seem to be some custom attributes involved. Could you double check this with the reviewers?

https://github.com/aws-amplify/amplify-js/blob/a196b0cb4fcda92a535b85ec046eb03a562ac907/packages/auth/src/providers/cognito/types/models.ts#L57

https://github.com/aws-amplify/amplify-js/blob/a196b0cb4fcda92a535b85ec046eb03a562ac907/packages/core/src/singleton/Auth/types.ts#L81

https://github.com/aws-amplify/amplify-js/blob/a196b0cb4fcda92a535b85ec046eb03a562ac907/packages/auth/src/providers/cognito/types/models.ts#L68

soberm: Hi @kturczyski-lz, will double check but the previous type mistakenly used UserAttributes and not DeviceAttributes which are different.

kturczyski-lz: Ahh, yeah, sorry, I actually see only those 3 keys in the result. I assumed the old type wasn’t simply ported to the new structure.

packages/auth/src/providers/cognito/types/models.ts
```diff
@@ -89,11 +88,15 @@ export type AutoSignInEventData =
...
+
 /**
  * Holds the device specific information along with it's id and name.
  */
 export type AWSAuthDevice = AuthDevice & {
-	attributes: AuthUserAttribute<UserAttributeKey>;
+	attributes: Partial<Record<DeviceAttributeKey, string>> &
+		Record<string, string>;
```
adrianjoshua-strutt, line 99: Is there a reason why we would want to allow all attributes as possible return types, other than 'device_status' | 'device_name' | 'last_ip_used' ?

soberm: Cognito does not specify which device attributes they set. They only use a generic [AttributeType](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_DeviceType.html#CognitoUserPools-Type-DeviceType-DeviceAttributes). We only know for sure, that `'device_status' | 'device_name' | 'last_ip_used'` is used. So we provide Type completion for the device attributes that are known but leave room for additional device attributes.

bobbor (approved):

pranavosu (approved):

Issue #14459, opened 2025-07-03
rtn-web-browser does not open browser on Chromebook


caroldinh:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [x] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

React Native

### Amplify APIs

Authentication

### Amplify Version

v6

### Amplify Categories

auth

### Backend

Amplify CLI

### Environment information

<details>

```
  System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 13th Gen Intel(R) Core(TM) i7-13620H
    Memory: 3.00 GB / 15.69 GB
  Binaries:
    Node: 22.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 10.9.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (135.0.3179.66)
    Internet Explorer: 11.0.26100.1882
  npmPackages:
    @aws-amplify/cli: ^12.13.0 => 12.14.4 
    @aws-amplify/core: ^6.10.3 => 6.10.5 
    @aws-amplify/core/internals/adapter-core:  undefined ()
    @aws-amplify/core/internals/aws-client-utils:  undefined ()
    @aws-amplify/core/internals/aws-client-utils/composers:  undefined ()
    @aws-amplify/core/internals/aws-clients/cognitoIdentity:  undefined ()
    @aws-amplify/core/internals/aws-clients/pinpoint:  undefined ()
    @aws-amplify/core/internals/providers/pinpoint:  undefined ()
    @aws-amplify/core/internals/utils:  undefined ()
    @aws-amplify/core/server:  undefined ()
    @aws-amplify/react-native: ^1.1.7 => 1.1.7 
    @aws-amplify/rtn-web-browser: ^1.1.1 => 1.1.1 
    @aws-amplify/ui-react-native: ^2.4.4 => 2.4.4 
    @babel/core: ^7.26.9 => 7.26.10
[...]

PR #14523, opened 2025-08-26
feat(rtn-web-browser): add auth support for chromebook

ahmedhamouda78:
#### Description of changes

This PR adds authentication support for Chromebook devices in the `rtn-web-browser` package. The changes include:

- Added `isChromebook()` function to detect Chromebook devices using Expo Device API and native ChromeOS module
- Implemented `openAuthSessionChromeOs()` function to handle auth sessions on Chromebook using `Linking.openURL()` instead of native browser tabs
- Updated `openAuthSessionAsync()` to detect Chromebook devices and route to the appropriate auth session handler
- Enhanced Android auth flow to fallback gracefully when Chromebook detection fails

#### Issue #, if available

#14459

- Added Unit tests (I will raise a different PR for the tests as they are the first tests in the `rtn-web-browser` package)

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/rtn-web-browser/src/apis/openAuthSessionAsync.ts
+++ b/packages/rtn-web-browser/src/apis/openAuthSessionAsync.ts
@@ -5,6 +5,7 @@ import {
 	AppState,
 	Linking,
 	NativeEventSubscription,
+	NativeModules,
 	Platform,
 } from 'react-native';
 
@@ -12,6 +13,32 @@ import { nativeModule } from '../nativeModule';
 
 let appStateListener: NativeEventSubscription | undefined;
 let redirectListener: NativeEventSubscription | undefined;
+
+export async function isChromebook(): Promise<boolean> {
+	// expo go
+	try {
+		const Device = require('expo-device');
+		if (Device?.hasPlatformFeatureAsync) {
+			if (await Device.hasPlatformFeatureAsync('org.chromium.arc')) return true;
+			if (
+				await Device.hasPlatformFeatureAsync(
+					'org.chromium.arc.device_management',
+				)
+			)
+				return true;
+		}
+	} catch {
+		// not using Expo
+	}
+
+	// fallback to native module
+	try {
+		const nm = (NativeModules as any)?.ChromeOS;
+		if (nm?.isChromeOS) return !!(await nm.isChromeOS());
+	} catch {}
+
+	return false;
+}
 
 export const openAuthSessionAsync = async (
 	url: string,
@@ -25,6 +52,15 @@ export const openAuthSessionAsync = async (
 	}
 
 	if (Platform.OS === 'android') {
+		try {
+			const isChromebookRes = await isChromebook();
+			if (isChromebookRes) {
+				return openAuthSessionChromeOs(httpsUrl, redirectUrls);
+			}
+		} catch {
+			// ignore and fallback to android
+		}
+
 		return openAuthSessionAndroid(httpsUrl, redirectUrls);
 	}
 };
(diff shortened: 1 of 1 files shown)


packages/rtn-web-browser/src/apis/__tests__/openAuthSessionAsync.test.ts
```diff
@@ -0,0 +1,78 @@
...
+const mockPlatform = Platform as jest.Mocked<typeof Platform>;
+const mockNativeModule = nativeModule as jest.Mocked<typeof nativeModule>;
+
+describe('openAuthSessionAsync', () => {
+	beforeEach(() => {
+		jest.clearAllMocks();
+		mockPlatform.OS = 'ios';
+	});
+
+	describe('isChromebook', () => {
```
soberm, line 29: Should we also add tests for the other cases?

packages/rtn-web-browser/jest.config.js
```diff
@@ -2,10 +2,10 @@ module.exports = {
 	...require('../../jest.config'),
 	coverageThreshold: {
 		global: {
-			branches: 100,
-			functions: 100,
-			lines: 100,
-			statements: 100,
+			branches: 25,
```
soberm, line 5: Why reduce the coverage thresholds?

packages/rtn-web-browser/src/apis/__tests__/openAuthSessionAsync.test.ts
ahmedhamouda78: I added full coverage for this package in PR #14541

packages/rtn-web-browser/jest.config.js
ahmedhamouda78: The original values are incorrect as the package doesn't have any unit tests at all. So even though it says 100%, it's actually 0%. I reduced the coverage in this PR to match the actual coverage that was achieved after I added some unit tests in this PR. I added full coverage for this package in another PR though (#14541), which I will merge after this PR

packages/rtn-web-browser/src/apis/openAuthSessionAsync.ts
```diff
@@ -13,6 +14,32 @@ import { nativeModule } from '../nativeModule';
 let appStateListener: NativeEventSubscription | undefined;
 let redirectListener: NativeEventSubscription | undefined;
 
+export async function isChromebook(): Promise<boolean> {
+	// expo go
+	try {
+		const Device = require('expo-device');
```
bobbor, line 20: I don't understand.

do we expect users to install `expo-device` optionally on their own?

why not include it as a dependency? and if we cannot include it as a dependency, why is this logic needed at all?

packages/rtn-web-browser/src/apis/openAuthSessionAsync.ts
```diff
@@ -66,6 +102,27 @@ const openAuthSessionAndroid = async (url: string, redirectUrls: string[]) => {
...
+	try {
+		const [redirectUrl] = await Promise.all([
+			Promise.race([
+				// wait for app to redirect, resulting in a redirectUrl
+				getRedirectPromise(redirectUrls),
+				// wait for app to return some other way, resulting in null
+				getAppStatePromise(),
+			]),
+			Linking.openURL(url),
+		]);
```
bobbor, line 115: this is hard to follow.

this races between, 2 promises
1. whenever a "linking" event happens
2. when the app leaves inactive state and becomes active.

and then this race is part of an all, which resolves when the race resolved and an url has been opened.

----
doesn't `Linking.openURL` automatically mean that `getRedirectPromise` gets resolved.
as a result the same URL, which was just opened gets returned by this Promise construct 
because:
```
if (redirectUrls.some(url => event.url.startsWith(url))) {
				resolve(event.url);
}
```
as
1. `getRedirectPromise` sets up a listener on "Linking"-url change and resolves with that URL when this happens
2. `getAppStatePromise` sets up a listener on "going active" and resolves with `null`
3. openURL - navigates

so that.

1. navigate.
2. getRedirectPromise-emits
3. redirectUrl === url

and then this makes another call to `openURL` with the same URL

packages/rtn-web-browser/src/apis/openAuthSessionAsync.ts
ahmedhamouda78: **We** don't require `expo-device` as a dependency. However, customers are free to use it with their Amplify react-native apps, so this line is just to support device detection in case customers are using expo. If customers are not using expo, we fallback to NativeModules on line 36

bobbor: got it. thanks for the explanation

packages/rtn-web-browser/src/apis/openAuthSessionAsync.ts
bobbor: resolved this offline.

the solution uses `Linking` instead of native navigation if the latter fails to open the URL

PR #14536, opened 2025-09-01
fix(storage): add correct contentType

soberm:
#### Description of changes

Implements automatic content type detection for S3 uploads to replace the static `'application/octet-stream'` fallback. More specifically:

• Added `getContentType` utility function with extension-to-MIME-type mapping for common file formats
• Updated single-part uploads (`putObjectJob.ts`) to detect content type from File objects and extensions  
• Updated multipart uploads (`uploadHandlers.ts`) to use the same content type detection logic
• Updated upload cache (`uploadCache.ts`) to include detected content type in cache key generation
• Added comprehensive unit tests for content type detection and upload scenarios
• Maintains backward compatibility - explicit `contentType` option still takes precedence

Added unit tests and run [E2E tests](https://github.com/aws-amplify/amplify-js/actions/runs/17373680126)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/storage/src/providers/s3/apis/internal/uploadData/putObjectJob.ts
+++ b/packages/storage/src/providers/s3/apis/internal/uploadData/putObjectJob.ts
@@ -61,8 +61,8 @@ export const putObjectJob =
 		const {
 			contentDisposition,
 			contentEncoding,
-			contentType = uploadDataOptions?.contentType ||
-				getContentType(data, objectKey) ||
+			contentType = uploadDataOptions?.contentType ??
+				getContentType(data, objectKey) ??
 				'application/octet-stream',
 			preventOverwrite,
 			metadata,


padenn2:
There is a media type list published by mozilla that might offer a file extension -> mime type reference. https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types
It in turn references a very long iana list https://www.iana.org/assignments/media-types/media-types.xhtml which is beyond the scope of this fix.

soberm:
> There is a media type list published by mozilla that might offer a file extension -> mime type reference. https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types It in turn references a very long iana list https://www.iana.org/assignments/media-types/media-types.xhtml which is beyond the scope of this fix.

I updated the mapping to include the most important types that are mentioned there.

packages/storage/src/utils/contentType.ts
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Detect content type from file data or filename extension
+ */
+export const getContentType = (data: any, key: string): string | undefined => {
+	if (data instanceof File && data.type) {
+		return data.type;
+	}
+
+	const ext = key.split('.').pop()?.toLowerCase();
+	const mimeTypes: Record<string, string> = {
```
Simone319: Should we create a const outside of this function, so that it's not recreated every time?

soberm: Yes, changed it. Thx.

Simone319 (approved):

ahmedhamouda78 (approved):

packages/storage/__tests__/utils/contentType.test.ts
```diff
@@ -0,0 +1,37 @@
...
+		expect(getContentType('data', 'file.unknown')).toBeUndefined();
+		expect(getContentType('data', 'noextension')).toBeUndefined();
+	});
+
+	it('should prefer File.type over extension detection', () => {
+		const file = new File(['content'], 'test.png', { type: 'image/jpeg' });
+		expect(getContentType(file, 'test.png')).toBe('image/jpeg');
+	});
+
+	it('should fallback to extension when File has no type', () => {
```
pranavosu, line 39: Let's add a test for multiple `.` in a file name, e.g `myfile.tar.gz`

packages/storage/src/utils/contentType.ts
```diff
@@ -0,0 +1,99 @@
...
+ * Detect content type from file data or filename extension
+ */
+export const getContentType = (data: any, key: string): string | undefined => {
+	if (data instanceof File && data.type) {
+		return data.type;
+	}
+
+	const ext = key.split('.').pop()?.toLowerCase();
+
+	return ext ? MIME_TYPES[ext] : undefined;
```
pranavosu, line 98: I wonder if this should default to `'application/octet-stream'` since this util is storage specific

soberm: I think it is better to keep the fallback logic separate from the util. It's clearer and leaves more flexibility.

packages/storage/__tests__/utils/contentType.test.ts
soberm: Added a test case to verify that this works.

Issue #13379, opened 2024-05-14
id token doesn't contain ClientMetadata after refresh


rashidwiizb:
this way I m using signIn
"@aws-amplify/react-native": "^1.1.1",
"aws-amplify": "^6.3.1",
"react-native": "0.73.2",

  const userSignIn = await signIn({
        username: userName,
        password: password,
        options: {
          clientMetadata: {
            roleType: "Student"
          }
        }
      });
      
I already attache pre token lambda trigger on Cognito for customise the id token . so I get the roleType in it on idTojeb when signing success . But when this idToken expires I get the new id token from amplify itself so in that idToken I didn't get this roleType

cwomack:
Hello, @rashidwiizb 👋.  I think we may need to understand how your Auth flow is structured to better help here.  How are you persisting the `roleType` field and attaching it to the iDToken? Can you share the lambda hook implementation and how to reproduce this?  Thanks!

TomMuehlegger:
We are facing the same issue. When providing the ClientMetadata to the signIn method, we are getting the right JWT token from Cognito (adding a claim, depending on the ClientMetadata, to the token via the PreTokenGeneration trigger in Cognito). 

But when doing the fetchAuthSession({ forceRefresh: true }); via Amplify, the ClientMetadata is not provided to the PreTokenGeneration trigger and thus, this information is missing in the JWT token provided by the fetchAuthSession call.

We are using the latest "aws-amplify" version 6.3.2.
Thanks...

rashidwiizb:
Hi  @mattcreaser  thanks for the reply.

@cwomack my pre-token lambda triggers is 

const handler = async (event, context,callback) => {
  try {
        let roleType = '';
        if (event.triggerSource === 'TokenGeneration_Authentication') {
            roleType = event.request?.clientMetadata?.roleType;
            cachedRoleType = roleType;
        } else if (event.triggerSource === 'TokenGeneration_RefreshTokens') {
            roleType = cachedRoleType;
        }
        
        event.response = {
            claimsOverrideDetails: {
                claimsToAddOrOverride: {
                    'roleType': roleType
                },                
            },
        };

        return callback(null, event);
    } catch (error) {
        console.error('Error processing Pre Token Generation:', error);
        throw error;
    }
};

export { handler };

[...]

PR #14556, opened 2025-09-17
feat(auth): support clientmetadata for token refresh

soberm:
## Description of changes

Adds support for dynamically providing client metadata for Cognito token refresh operations through a `ClientMetadataProvider` function type. More specifically:

• Added `ClientMetadataProvider` function type to @aws-amplify/core types
• Added `setClientMetadataProvider()` method to `CognitoUserPoolsTokenProvider`  
• Integrated `ClientMetadataProvider` into `TokenOrchestrator` for token refresh operations
• Added async client metadata provider support to token refresh flow
• Updated tests for new functionality

## Usage

```typescript
// Set global clientMetadata (affects all token refreshes)
import { cognitoUserPoolsTokenProvider } from 'aws-amplify/auth/cognito';

const clientMetadataProvider = () => Promise.resolve({
  'app-version': '1.0.0',
  'device-type': 'mobile'
});

cognitoUserPoolsTokenProvider.setClientMetadataProvider(clientMetadataProvider);

// Per-call clientMetadata (overrides global)
import { fetchAuthSession } from 'aws-amplify/auth';

const session = await fetchAuthSession({
  forceRefresh: true,
  clientMetadata: { 
    'context': 'user-action',
    'operation': 'manual-refresh'
  }
});

// Existing usage remains unchanged (backward compatible)
const session = await fetchAuthSession({ forceRefresh: true });
```

#### Issue #, if available

https://github.com/aws-amplify/amplify-js/issues/13379

## Description of how you validated changes

• Added unit tests for setClientMetadataProvider functionality

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.github/workflows/callable-canary-sampleapp-tests.yml
+++ b/.github/workflows/callable-canary-sampleapp-tests.yml
@@ -125,7 +125,7 @@ jobs:
         run: npm install
         working-directory: amplify-js-samples-staging/samples/next/auth/new-next-app
       - name: Install latest stable amplify version
-        run: npm install aws-amplify@latest @aws-amplify/adapter-nextjs@latest
+        run: npm install aws-amplify@latest @aws-amplify/adapter-nextjs@latest @aws-amplify/auth@latest
         working-directory: amplify-js-samples-staging/samples/next/auth/new-next-app
       - name: Start application and run test
         run: |
--- a/README.md
+++ b/README.md
@@ -52,6 +52,14 @@ AWS Amplify is available as `aws-amplify` on [npm](https://www.npmjs.com/package
 
 To get started pick your platform from our [**Getting Started** home page](https://docs.amplify.aws/javascript/)
 
+## Version Support
+AWS Amplify JavaScript library v4 and below will end support on **April 13, 2026**, as documented in our [AWS Amplify JavaScript library version support calendar](https://github.com/aws-amplify/amplify-js/issues/14580).
+
+Effective immediately, the AWS Amplify JavaScript library v4 and below will enter **Maintenance Mode** until April 13, 2026 after which it will receive no more updates. While in Maintenance Mode, the libraries will only receive updates for **critical bug fixes and security vulnerabilities**. Refer to [Amplify Documentation](https://docs.amplify.aws/reference/maintenance-policy) for more information on the maintenance policy.
+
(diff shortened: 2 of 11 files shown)


packages/core/src/singleton/Auth/types.ts
```diff
@@ -4,6 +4,16 @@
 import { StrictUnion } from '../../types';
 import { AtLeastOne } from '../types';
 
+/**
+ * Arbitrary key/value pairs that may be passed as part of certain Cognito requests
+ */
+export type ClientMetadata = Record<string, string>;
```
sarayev, line 10: Should we have input validation for the ClientMetadata fields?

packages/core/src/singleton/Auth/types.ts
sarayev, line 1: Do we need to update our documentations to inform the developers about the handling of the ClientMetadata fields, similar to how Cognito is doing [here](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminRespondToAuthChallenge.html#API_AdminRespondToAuthChallenge_RequestSyntax)?

packages/auth/__tests__/providers/cognito/refreshToken.test.ts
```diff
@@ -60,6 +63,7 @@ describe('refreshToken', () => {
 		});
 
 		it('should refresh token', async () => {
```
sarayev, line 65: Do we need to have a test to verify the API behavior when both values (global clientMetadataProvider and API-level clientMetadata field) are configured, where per-call metadata takes precedence over metadata from provider?

packages/core/src/singleton/Auth/types.ts
soberm: ClientMetadata is unvalidated/unencrypted by design. Security responsibility lies with AWS service and developer education, not client-side validation.

packages/auth/__tests__/providers/cognito/refreshToken.test.ts
soberm: Added a test case to verify this behavior.

packages/core/src/singleton/Auth/types.ts
soberm: This is already mentioned in the docs. However, we should add that this is now also available for refreshing tokens.

PR #14423, opened 2025-06-09
Added support for idp_identifier query parameter in cognito authorizer endpoint

jon-armen:
#### Description of changes

An isIdpIdentifier was added as an optional field in the AuthSignInWithRedirectInput provider options to enable the custom provider to be supplied as an idp_identifier query parameter when sending the request URL to the Cognito authorize endpoint, as described in the [Amazon Cognito documentation](https://docs.aws.amazon.com/cognito/latest/developerguide/authorization-endpoint.html).

#### Issue #, if available

Closes #10226 

#### Checklist

- [X] PR description included
- [x] `yarn test` passes
- [X] Unit Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)
- [X] Relevant documentation is changed or added (and PR referenced)

#### Checklist for repo maintainers

- [ ] Verify E2E tests for existing workflows are working as expected or add E2E tests for newly added workflows
- [ ] New source file paths included in this PR have been added to CODEOWNERS, if appropriate

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #14573, opened 2025-09-29
feat: add support for idp_identifier query parameter when going through OAuth using signInWithRedirect function

sarayev:
#### Description of changes

Adds a new `idpIdentifier` parameter for the `signInWithRedirect` function's input definition in order to allow Amplify developers to use `idp_identifier` query param of the [authorization endpoint of Cognito](https://docs.aws.amazon.com/cognito/latest/developerguide/authorization-endpoint.html). 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/react-native/__tests__/apis.test.ts
+++ b/packages/react-native/__tests__/apis.test.ts
@@ -1,0 +1,58 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+	computeModPow,
+	computeS,
+	getDeviceName,
+	getOperatingSystem,
+} from '../src/apis';
+import { nativeModule } from '../src/nativeModule';
+
+jest.mock('react-native', () => ({
+	Platform: { OS: 'ios' },
+	NativeModules: {},
+}));
+
+jest.mock('../src/nativeModule', () => ({
+	nativeModule: {
+		computeModPow: jest.fn(),
+		computeS: jest.fn(),
+		getDeviceName: jest.fn(),
+	},
+}));
+
+describe('APIs', () => {
+	beforeEach(() => {
+		jest.clearAllMocks();
+	});
+
+	describe('getOperatingSystem', () => {
+		it('should return Platform.OS', () => {
+			expect(getOperatingSystem()).toBe('ios');
+		});
+	});
+
+	describe('computeModPow', () => {
+		it('should call nativeModule.computeModPow with payload', () => {
+			const payload = { base: '2', exponent: '3', divisor: '5' };
+			computeModPow(payload);
+			expect(nativeModule.computeModPow).toHaveBeenCalledWith(payload);
+		});
+	});
+
+	describe('computeS', () => {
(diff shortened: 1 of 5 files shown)


packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
```diff
@@ -117,10 +123,17 @@ const oauthSignIn = async ({
 		['redirect_uri', redirectUri],
 		['response_type', responseType],
 		['client_id', clientId],
-		['identity_provider', provider],
-		['scope', scopes.join(' ')],
 	]);
 
+	// Add either identity_provider or idp_identifier, but not both
```
soberm, line 128: Should this be reflected in the types as well? Should we also throw en error in this case since implicitly preferring idpIdentifier might not be clear?

packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
```diff
@@ -117,10 +123,17 @@ const oauthSignIn = async ({
...
+	if (idpIdentifier) {
+		params.append('idp_identifier', idpIdentifier);
+	} else {
+		params.append('identity_provider', provider);
+	}
+
+	params.append('scope', scopes.join(' '));
```
soberm, line 135: Is the order of appending important or why is scope the last here and not part of the array passed to URLSearchparams?

packages/auth/src/types/inputs.ts
```diff
@@ -69,7 +69,7 @@ export type AuthProvider = 'Amazon' | 'Apple' | 'Facebook' | 'Google';
 export type AuthPrompt = 'NONE' | 'LOGIN' | 'CONSENT' | 'SELECT_ACCOUNT';
 
 export interface AuthSignInWithRedirectInput {
-	provider?: AuthProvider | { custom: string };
+	provider?: AuthProvider | { custom: string } | { idpIdentifier: string };
```
soberm: This will also accept:
```typescript
const badProvider = { 
  custom: "MyCustomProvider",
  idpIdentifier: "example.com" 
};
```

Is this intended?

ahmedhamouda78: I changed the type to make them mutually exclusive.

packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
ahmedhamouda78: In general, the order of request params in oauth2 endpoints should not matter. However, since this is undocumented behavior for cognito, maintaining the same order we had before is the safest route we can take.

packages/auth/src/providers/cognito/apis/signInWithRedirect.ts
ahmedhamouda78: Same answer as the other comment; I modified the type to make the types mutually exclusive. Nevertheless, this check will make sure we are only setting one parameter at runtime in case users override the type-safety somehow.

annjawn:
@pranavosu wondering when will the release with #14423 be available for the idp_identifier parameter?

PR #14631, opened 2025-11-17
fix(rtn-web-browser): Don't use custom tabs on default and fallback to normal tabs

cadivus:
This is a follow-up of https://github.com/aws-amplify/amplify-js/pull/14523 to change the fallback order. The ChromeOS way works everywhere, the Android not. And the Android tab opening needs to be awaited so that we can catch it. It should also fix https://github.com/aws-amplify/amplify-js/issues/14459

#### Issue #, if available
https://github.com/aws-amplify/amplify-js/pull/14523
https://github.com/aws-amplify/amplify-js/issues/14459

I used a Chromebook and the Android emulator. Both do the expected.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/rtn-web-browser/src/apis/openAuthSessionAsync.ts
+++ b/packages/rtn-web-browser/src/apis/openAuthSessionAsync.ts
@@ -37,14 +37,16 @@ export const openAuthSessionAsync = async (
 	if (Platform.OS === 'android') {
 		try {
 			const isChromebookRes = await isChromebook();
-			if (isChromebookRes) {
-				return openAuthSessionChromeOs(httpsUrl, redirectUrls);
+			if (!isChromebookRes) {
+				// If it is not a Chromebook, it probably supports custom tabs
+				return await openAuthSessionAndroid(httpsUrl, redirectUrls);
 			}
 		} catch {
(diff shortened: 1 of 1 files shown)


packages/rtn-web-browser/src/apis/openAuthSessionAsync.ts
```diff
@@ -37,14 +37,16 @@ export const openAuthSessionAsync = async (
 	if (Platform.OS === 'android') {
 		try {
 			const isChromebookRes = await isChromebook();
-			if (isChromebookRes) {
-				return openAuthSessionChromeOs(httpsUrl, redirectUrls);
+			if (!isChromebookRes) {
+				// If it is not a Chromebook, it probably supports custom tabs
+				return await openAuthSessionAndroid(httpsUrl, redirectUrls);
```
osama-rizk, line 42: Why adding await to the Android path but not the ChromeOS fallback path.

cadivus: Because Android can work with the ChromeOS path but not the other way around. So we fall back to the more compatible path.

PR #14779, opened 2026-04-12
feat(adapter-nextjs): support dynamic redirect after sign-in using redirectTo query param

### LINKED ISSUE #14677: adapter-nextjs: redirect to the originally requested url after sign-in
### Is this related to a new or existing framework?

Next.js

### Is this related to a new or existing API?

Authentication

### Is this related to another service?

_No response_

### Describe the feature you'd like to request

Support dynamic redirect after sign-in to return users to their originally requested page.

Currently, `createAuthRouteHandlers` only supports a static `redirectOnSignInComplete` value:

```typescript
createAuthRouteHandlers({
  redirectOnSignInComplete: '/dashboard', // Always redirects here
})
```

This means users always land on the same page after sign-in, regardless of where they were trying to go. For example:
- User tries to access `/settings/profile`
- Gets redirected to sign-in
- After successful sign-in, lands on `/dashboard` instead of `/settings/profile`

This is a common UX pattern that most authentication libraries support out of the box (NextAuth.js `callbackUrl`, Auth0 `returnTo`, Clerk, etc.).

### Describe the solution you'd like

Support a `redirectTo` query parameter on the sign-in endpoint:

```
/api/auth/sign-in?redirectTo=/settings/profile
```

After successful authentication, the user should be redirected to `/settings/profile` instead of the static `redirectOnSignInComplete` value.

Usage in middleware would be:
```typescript
// middleware.ts
if (!authenticated) {
  return NextResponse.redirect(
    `/api/auth/sign-in?redirectTo=${encodeURIComponent(request.nextUrl.pathname)}`
  );
}
```

Security considerations:
- Only allow relative paths (reject absolute URLs to prevent open redirect attacks)
- Fall back to `redirectOnSignInComplete` if `redirectTo` is not provided or invalid

### Describe alternatives you've considered

1. **User-managed cookies**: Store the return URL in a cookie before redirecting to sign-in, then read it after sign-in and redirect manually. This works but requires users to handle cookie management (security attributes, expiration, cleanup) and implement open redirect protection themselves.

2. **Client-side redirect**: After sign-in, land on a page that reads a query parameter and does `router.replace()`. This adds an extra redirect hop and requires coordination between middleware and the landing page.

Both alternatives push complexity onto users for what is a very common use case.

### Additional context

_No response_

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

TiwariLokesh:
#### Description of changes

This PR adds support for dynamic redirects after sign-in in the Next.js adapter by introducing a `redirectTo` query parameter.

Currently, users are always redirected to a static `redirectOnSignInComplete` path after authentication. This change allows users to be redirected back to their originally requested page.

#### Issue #, if available

Fixes #14677

* Verified redirect logic by extracting and testing query parameter handling locally
* Confirmed that:

  * Valid relative paths (e.g. `/dashboard`) are correctly used for redirection
  * Invalid or external URLs (e.g. `https://example.com`) are ignored
  * Fallback to `redirectOnSignInComplete` works when `redirectTo` is absent or invalid
* Successfully built the project using `yarn build` with no errors

#### Checklist

* [x] PR description included
* [ ] `yarn test` passes
* [ ] Unit Tests are changed or added
* [ ] Relevant documentation is changed or added

### DIFF
--- a/packages/adapter-nextjs/src/auth/handlers/handleSignInCallbackRequest.ts
+++ b/packages/adapter-nextjs/src/auth/handlers/handleSignInCallbackRequest.ts
@@ -57,7 +57,6 @@ export const handleSignInCallbackRequest: HandleSignInCallbackRequest = async ({
 	const { [PKCE_COOKIE_NAME]: clientPkce, [STATE_COOKIE_NAME]: clientState } =
 		getCookieValuesFromRequest(request, [PKCE_COOKIE_NAME, STATE_COOKIE_NAME]);
 
-	// The state and pkce cookies are removed from cookie store after 5 minutes
 	if (!clientState || !clientPkce) {
 		const searchParamsString = createErrorSearchParamsString({
 			error: SIGN_IN_TIMEOUT_ERROR_CODE,
@@ -72,7 +71,6 @@ export const handleSignInCallbackRequest: HandleSignInCallbackRequest = async ({
 		});
 	}
 
-	// Most likely the cookie has been tampered
 	if (clientState !== state) {
 		return new Response(null, { status: 400 });
 	}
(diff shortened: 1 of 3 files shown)


bobbor:
hey @TiwariLokesh 

This PR looks good so far. While I can see the use-case for scenario-specific `redirectTo`-URL's, I cannot understand how you would specify them. 😕 

Can you quickly guide me through your flow? Ideally highlighting, when you would set the `redirectTo` at the beginning of signIn, and how you've configured (keeping out secrets) your cognito - should you use it - or another auth. Then maybe we can find a solution, which works now already or at least I'd understand and we can go ahead with your proposed changes.

PR #14788, opened 2026-04-24
fix: remove unused uuid dependency from api-graphql, predictions, and interactions

AdrianoNicolucci:
## Summary

Removes the direct `uuid` dependency from packages that don't need it, consolidating all UUID generation through `@aws-amplify/core`'s `amplifyUuid` wrapper. This addresses the security advisory [GHSA-w5hq-g745-h8pq](https://github.com/advisories/GHSA-w5hq-g745-h8pq).

## Motivation

Dependabot flagged `uuid` for an out-of-bounds write vulnerability in `v3()`, `v5()`, and `v6()` when a caller-supplied buffer has an invalid offset. While this project only uses `v4()` (which is **not affected**), the cleanest resolution is to remove the unnecessary direct dependency from packages that never use it, and consolidate through core.

## Changes

| Package | Change |
|---|---|
| `@aws-amplify/api-graphql` | Removed `uuid` from dependencies — was never imported directly (uses `amplifyUuid` from core) |
| `@aws-amplify/predictions` | Removed `uuid` from dependencies — was never imported anywhere in the package |
| `@aws-amplify/interactions` | Replaced direct `import { v4 as uuid } from 'uuid'` with `amplifyUuid` from `@aws-amplify/core/internals/utils`, then removed `uuid` from dependencies |

After these changes, `uuid` exists only as a dependency in `@aws-amplify/core` (`^11.0.0`), which is the single package that wraps `uuid.v4()` via `amplifyUuid`.

## Why not upgrade to uuid@14?

uuid@12+ dropped CommonJS support. Since amplify-js builds both CJS and ESM outputs with `uuid` kept as an external (not bundled), upgrading would break the CJS build at runtime. Staying on v11 is safe because only `v4()` is used, which is unaffected by the vulnerability.

## Testing

- No behavioral change — `amplifyUuid()` is already a direct re-export of `uuid.v4()`
- The interactions package already imports from `@aws-amplify/core/internals/utils` for `getAmplifyUserAgentObject`, so this follows the existing pattern

### DIFF
--- a/.changeset/remove-uuid-dependency.md
+++ b/.changeset/remove-uuid-dependency.md
@@ -1,0 +1,7 @@
+---
+'@aws-amplify/api-graphql': patch
+'@aws-amplify/interactions': patch
+'@aws-amplify/predictions': patch
+---
+
+Remove unused uuid dependency from @aws-amplify/api-graphql, @aws-amplify/interactions, and @aws-amplify/predictions packages. All UUID generation is now consolidated through @aws-amplify/core's amplifyUuid wrapper, addressing security advisory GHSA-w5hq-g745-h8pq.
--- a/packages/api-graphql/package.json
+++ b/packages/api-graphql/package.json
@@ -85,8 +85,7 @@
 		"@aws-sdk/types": "^3.973.6",
 		"graphql": "15.8.0",
 		"rxjs": "^7.8.1",
-		"tslib": "^2.5.0",
-		"uuid": "^11.0.0"
+		"tslib": "^2.5.0"
 	},
 	"size-limit": [
 		{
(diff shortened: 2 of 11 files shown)


soberm:
Thank you for your contribution! There is another file that still uses uuid:

https://github.com/aws-amplify/amplify-js/blob/8bf7c89282dfcdd00cb0cc56d21bf6ec52fcb21e/packages/interactions/__tests__/lex-v2/AWSLexV2Provider.test.ts#L13

Additionally, please make sure to add a changeset.

AdrianoNicolucci:
> Thank you for your contribution! There is another file that still uses uuid:
> 
> https://github.com/aws-amplify/amplify-js/blob/8bf7c89282dfcdd00cb0cc56d21bf6ec52fcb21e/packages/interactions/__tests__/lex-v2/AWSLexV2Provider.test.ts#L13
> 
> Additionally, please make sure to add a changeset.

Done

soberm (approved):

bobbor (approved):
LGTM. let's do this

PR #14805, opened 2026-05-08
feat(core): add AmplifyContext infrastructure

bobbor:
Add context module with AmplifyContext interface, context brand for runtime type checking, global context management, and resolveCtxArgs utility for optional context-first parameter pattern.

Also adds configurationBuilder, AmplifyServerContextError, and setGlobalContext call in Amplify.configure().

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/core/src/configurationBuilder/configurationBuilder.ts
```diff
@@ -0,0 +1,226 @@
...
+ * import outputs from './amplify_outputs.json';
+ *
+ * const config = createConfiguration({ from: outputs })
+ *   .add('Storage', { S3: { bucket: 'my-bucket', region: 'us-east-1' } })
+ *   .build();
+ *
+ * Amplify.configure(config);
+ * ```
+ */
+export function createConfiguration(
```
soberm: Should we name this `createConfigurationBuilder` to make it clearer?

packages/core/__tests__/configurationBuilder/createConfiguration.test.ts
```diff
@@ -0,0 +1,375 @@
...
+		it('creates category if it does not exist', () => {
+			const config = createConfiguration()
+				.patch('Auth', {
+					Cognito: { userPoolId: 'pool', userPoolClientId: 'client' },
+				})
+				.build();
+			expect(config.Auth).toBeDefined();
+		});
+
+		it('handles arrays by replacing (not merging)', () => {
```
soberm: Which arrays?

packages/core/__tests__/storage/SyncKeyValueStorage.test.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { SyncKeyValueStorage } from '../../src/storage/SyncKeyValueStorage';
+
+describe('SyncKeyValueStorage', () => {
```
soberm, line 6: Should we add more tests for edge cases?

packages/core/__tests__/errors/APIError.test.ts
```diff
@@ -0,0 +1,47 @@
...
+			body: '{"error":"Not found"}',
+		};
+		const error = new ApiError({
+			name: 'NotFoundError',
+			message: 'Resource not found',
+			response,
+		});
+		expect(error.response).toEqual(response);
+		expect(error.response).not.toBe(response);
+		expect(error.response?.headers).not.toBe(response.headers);
```
soberm: Do we actually need this?

packages/core/src/configurationBuilder/index.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export {
+	createConfiguration,
+	ConfigurationBuilder,
+	CreateConfigurationOptions,
```
soberm: Should then also be `CreateConfigurationBuilderOptions`

packages/core/src/context/globalContext.ts
```diff
@@ -0,0 +1,57 @@
...
+	return _globalContext;
+}
+
+/**
+ * Returns the global {@link AmplifyContext} set by `Amplify.configure()`.
+ * Alias for {@link getActiveContext} — provided for semantic clarity.
+ *
+ * @throws If `Amplify.configure()` has not been called yet.
+ */
+export function getGlobalContext(): AmplifyContext {
```
soberm: Do we need to differentiate between active and global context? Maybe its easier to understand if we just have `getContext`, `setContext`, `clearContext`, `hasContext` ?

packages/core/src/context/resolveCtxArgs.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export function resolveCtxArgs<T>(args: unknown[]): [AmplifyContext, T] {
+	if (args.length > 1 && args[0] === undefined) {
+		throw new Error(
+			'Undefined AmplifyContext passed. Call configure() first or omit the parameter.',
+		);
+	}
+
+	if (isAmplifyContext(args[0])) {
+		return [args[0], args[1] as T];
```
soberm: Does this only pass the second arg or the whole remaining input?

packages/core/src/index.ts
```diff
@@ -79,3 +102,4 @@ export { ConsoleLogger } from './Logger';
 
 // Service worker
 export { ServiceWorker } from './ServiceWorker';
+// Errors — AmplifyServerContextError is exported from adapterCore
```
soberm: ?

packages/core/__tests__/context/resolveCtxArgs.test.ts
```diff
@@ -0,0 +1,91 @@
...
+	return ctx as unknown as AmplifyContext;
+}
+
+describe('resolveCtxArgs', () => {
+	afterEach(() => {
+		clearGlobalContext();
+	});
+
+	describe('when first arg is a branded AmplifyContext', () => {
+		it('returns [ctx, secondArg] tuple', () => {
```
soberm, line 34: Why only second arg? Should we add tests for multiple?

soberm (changes requested):

packages/core/__tests__/context/resolveCtxArgs.test.ts
bobbor: good point, will add a (or more) additional test.

packages/core/src/configurationBuilder/configurationBuilder.ts
```diff
@@ -0,0 +1,226 @@
...
+	/**
+	 * Add or replace a category configuration.
+	 *
+	 * @example
+	 * ```ts
+	 * config.add('Auth', { Cognito: { userPoolId: '...', userPoolClientId: '...' } });
+	 * config.add('Storage', { S3: { bucket: '...', region: '...' } });
+	 * ```
+	 */
+	add<K extends CategoryName>(
```
osama-rizk, line 59: WDYT about simplifying this app for example 

```

 add(value: Partial<ResourcesConfig>): ConfigurationBuilder;

  // usage
  .add({ Auth: { Cognito: {...} } })
  .add({ Storage: { S3: {...} } })

  vs the current:

  .add('Auth', { Cognito: {...} })
  .add('Storage', { S3: {...} })
```

packages/core/src/configurationBuilder/configurationBuilder.ts
```diff
@@ -0,0 +1,226 @@
...
+	 *
+	 * @example
+	 * ```ts
+	 * // Base config from outputs, override just the user pool for dev
+	 * const dev = createConfiguration({ from: outputs })
+	 *   .patch('Auth', { Cognito: { userPoolId: 'us-east-1_DEV' } })
+	 *   .build();
+	 * ```
+	 */
+	patch<K extends CategoryName>(
```
osama-rizk, line 76: Same here, is it possible to simplify this API signature ?

packages/core/__tests__/errors/APIError.test.ts
bobbor: nope, will remove

packages/core/src/configurationBuilder/configurationBuilder.ts
```diff
@@ -0,0 +1,226 @@
...
+	return builder;
+}
+
+function deepMerge(target: unknown, source: unknown): unknown {
+	if (
+		typeof target !== 'object' ||
+		typeof source !== 'object' ||
+		target === null ||
+		source === null ||
+		Array.isArray(target) ||
```
osama-rizk: A few thoughts on this helper:

**1. The "is this mergeable?" check is duplicated.** The same predicate appears in the early-return guard and again inline before the recursive call. Worth extracting to a small type guard:

```ts
function isPlainMergeable(v: unknown): v is Record<string, unknown> {
  return typeof v === 'object' && v !== null && !Array.isArray(v);
}
```

That removes the type assertions too.

**2. The inline check before recursing is redundant.** Since `deepMerge` already guards its inputs at the top, the recursive call handles the primitive/array cases correctly on its own — it just returns `source`. So the loop body collapses to:

```ts
for (const key of Object.keys(source)) {
  result[key] = deepMerge(result[key], source[key]);
}
```

Together, the helper drops to ~10 lines:

```ts
function deepMerge(target: unknown, source: unknown): unknown {
  if (!isPlainMergeable(target) || !isPlainMergeable(source)) {
    return source;
  }
  const result: Record<string, unknown> = { ...target };
  for (const key of Object.keys(source)) {
    result[key] = deepMerge(result[key], source[key]);
  }
  return result;
}
```

**3. Could we use `lodash.merge` (or `deepmerge`) instead?** Genuine question — is there a reason we're hand-rolling this? My hesitation about pulling in a dep here is that `@aws-amplify/core` is foundational and ships in every consumer's bundle, so adding even a small runtime dep has a real cost. Also, `lodash.merge` mutates its target, which would be a behavior change. But if `lodash` is already a transitive dep of core, or there's an existing internal `deepMerge` util in `packages/core/src/utils/` we should be reusing, that'd be the cleaner path. Did you consider either?

packages/core/src/configurationBuilder/configurationBuilder.ts
osama-rizk: +1

packages/core/src/context/globalContext.ts
bobbor: the suggestions are ambiguous.

as soon as we introduce local context, `hasContext` and `getContext` will cause confusion.

we don't need to differentiate. and I will remove the active context to just have globalContext.

packages/core/src/singleton/Amplify.ts
```diff
@@ -82,6 +85,27 @@ export class AmplifyClass {
...
+		};
+
+		Object.defineProperty(ctx, AMPLIFY_CONTEXT_BRAND, {
+			value: true,
+			enumerable: false,
+			configurable: false,
+			writable: false,
+		});
+
+		Object.freeze(ctx);
```
osama-rizk, line 106: any reasons to use "Object.freeze" instead of "deepFreeze"

packages/core/src/libraryUtils.ts
```diff
@@ -147,3 +148,13 @@ export {
...
 	SESSION_STOP_EVENT,
 } from './utils/sessionListener';
+
+// Global context internals
+export { setGlobalContext, clearGlobalContext } from './context/globalContext';
+
+// Context argument resolution
+export { resolveCtxArgs } from './context/resolveCtxArgs';
+
+// Storage internals
```
osama-rizk: why we are exporting these storage two utils?

packages/core/src/context/globalContext.ts
```diff
@@ -0,0 +1,57 @@
...
+	return getActiveContext();
+}
+
+/**
+ * Stores the given context as the global {@link AmplifyContext}.
+ *
+ * @internal
+ */
+export function setGlobalContext(ctx: AmplifyContext): void {
```
osama-rizk, line 29: Do we need to add validation here ?

packages/core/__tests__/context/resolveCtxArgs.test.ts
bobbor: added multi-arg tests

packages/core/__tests__/storage/SyncKeyValueStorage.test.ts
bobbor: added

packages/core/src/configurationBuilder/configurationBuilder.ts
bobbor: why do you think this is more simplified?

packages/core/src/context/globalContext.ts
bobbor: what kind of validation do you mean?

the function is only used (for now) in the Amplify singleton.
there we know that this is valid context, as it is controlled there.

packages/core/src/singleton/Amplify.ts
bobbor: ~makes sense, let me change that.~
the props in the ctx are already deepFrozen, so a plain freeze is enough

packages/core/src/libraryUtils.ts
bobbor: true, these are unnecessary. will remove

packages/core/src/configurationBuilder/configurationBuilder.ts
bobbor: again, why do you think this is simpler?

packages/core/src/configurationBuilder/configurationBuilder.ts
bobbor: > But if lodash is already a transitive dep of core

it isn't.

> or there's an existing internal deepMerge util in `packages/core/src/utils/`

there isn't.


so, I created a small helper util.

soberm (approved):

packages/core/src/context/globalContext.ts
osama-rizk: run time validation if the "ctx" is defined..

osama-rizk (approved):

soberm (approved):

osama-rizk (approved):

Issue #14827, opened 2026-05-27
uuid, js-cookie and axios updates for v5


brendanqshuttleid:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [x] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Not applicable

### Amplify APIs

Not applicable

### Amplify Version

v5

### Amplify Categories

Not applicable

### Backend

None

### Environment information

<details>

```
# Put output below this line
  System:
    OS: macOS 26.5
    CPU: (10) arm64 Apple M4
    Memory: 218.36 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.13.0 - /Users/brendan/.nvm/versions/node/v24.13.0/bin/node
    Yarn: 1.22.22 - /Users/brendan/.nvm/versions/node/v24.13.0/bin/yarn
    npm: 11.12.1 - /Users/brendan/.nvm/versions/node/v24.13.0/bin/npm
  Browsers:
    Chrome: 148.0.7778.179
    Safari: 26.5
  npmPackages:
    aws-amplify: ^5.0.0 => 5.3.33
  npmGlobalPackages:
    corepack: 0.34.5
    http-server: 14.1.1
    npm: 11.12.1
    yarn: 1.22.22

```

</details>

### Describe the bug

I can see there are current PRs open for v6 to address the recent high severity CVEs for uuid ([GHSA-w5hq-g745-h8pq)](https://github.com/uuidjs/uuid/security/advisories/GHSA-w5hq-g745-h8pq)) and js-cookie ([GHSA-qjx8-664m-686j](https://github.com/js-cookie/js-cookie/security/advisories/GHSA-qjx8-664m-686j)), but not yet for v5.

### Expected behavior

As v5 is still actively supported, there should be corresponding updates for that version.

### Reproduction steps

A fresh install of aws-amplify v5 gives me insecure versions of the two named packages:

```
% cat package.json
{
  "name": "amplify-check",
  "version": "1.0.0",
  "description": "",
  "license": "ISC",
  "author": "",
[...]

PR #14839, opened 2026-06-10
fix(deps): resolve uuid to ^11.1.1 to address CVE-2026-41907 in v5

osama-rizk:
## Issue

Closes part of #14827.

A fresh install of `aws-amplify@^5` still pulls a vulnerable `uuid@3.4.0` transitively:

```
% npm why uuid | grep uuid
uuid@3.4.0
  uuid@"^3.2.1" from @aws-amplify/analytics
  uuid@"^3.2.1" from @aws-amplify/api-graphql
  uuid@"3.4.0"  from @aws-amplify/datastore
  uuid@"^3.2.1" from @aws-amplify/notifications
  uuid@"^3.2.1" from @aws-amplify/predictions
  uuid@"^3.2.1" from @aws-amplify/pubsub
  uuid@"^3.0.0" from @aws-sdk/client-comprehend / client-translate / middleware-retry
```

`uuid < 11.1.1` is affected by **[GHSA-w5hq-g745-h8pq](https://github.com/uuidjs/uuid/security/advisories/GHSA-w5hq-g745-h8pq) (CVE-2026-41907)** — the `v3()`/`v5()`/`v6()` methods skip bounds validation on caller-provided output buffers, allowing silent partial writes that can produce truncated/malformed identifiers.

The other items from #14827 — **axios, js-cookie, fast-xml-parser, qs** — were already addressed in #14832. **uuid** was the remaining gap.

## Change

Add a root `resolutions` entry forcing `uuid@^11.1.1`, mirroring the v6 fix in **#14828** (already released):

```diff
   "lodash": "^4.18.1",
-  "js-cookie": "^3.0.7"
+  "js-cookie": "^3.0.7",
+  "uuid": "^11.1.1"
```

After this, all `uuid` ranges collapse to a single resolved `11.1.1` in `yarn.lock` (the vulnerable 3.4.0 / 7.0.3 / 8.3.2 entries are removed).

## Compatibility

uuid changed its import surface in v7 (named exports only), so I verified every import site in v5 source:

```
packages/*/src: import { v1 as uuid } from 'uuid';
                import { v4 as uuid } from 'uuid';
                import { v4 as uuid4 } from 'uuid';
```

**All** usages are named imports of `v1`/`v4` — API-compatible from uuid v3 through v11. There are **no** default imports (`import uuid from 'uuid'`) or `uuid.v4()` member-access patterns that v7+ removed, so no source changes are required. Confirmed `require('uuid')` in the installed 11.1.1 exposes `v1`/`v3`/`v4`/`v5` as functions.

## Notes

- Scope is intentionally minimal: only `package.json` + `yarn.lock`, only `uuid` entries change.
- Following the convention of the prior v5 security PR (#14832), no changeset file is added (v5-stable does not use `.changeset/`).
- Pre-existing `pubsub`/`api-graphql` TypeScript build errors (`symbol` index types) on v5-stable are unrelated to this change — they reproduce on a pristine `v5-stable` checkout (introduced by the Node 24 toolchain bump in #14784) and are unaffected by this resolution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,8 +116,7 @@
 		"qs": "^6.15.2",
 		"**/form-data": "4.0.4",
 		"lodash": "^4.18.1",
-		"js-cookie": "^3.0.7",
-		"uuid": "^11.1.1"
+		"js-cookie": "^3.0.7"
 	},
 	"jest": {
 		"resetMocks": true,
--- a/package.json
+++ b/package.json
@@ -116,7 +116,8 @@
 		"qs": "^6.15.2",
 		"**/form-data": "4.0.4",
 		"lodash": "^4.18.1",
-		"js-cookie": "^3.0.7"
+		"js-cookie": "^3.0.7",
+		"uuid": "^11.1.1"
 	},
 	"jest": {
 		"resetMocks": true,
--- a/packages/analytics/package.json
+++ b/packages/analytics/package.json
@@ -56,7 +56,7 @@
 		"@aws-sdk/util-utf8-browser": "3.6.1",
 		"lodash": "^4.18.1",
 		"tslib": "^1.8.0",
-		"uuid": "^3.2.1"
+		"uuid": "^11.1.1"
 	},
 	"devDependencies": {
 		"@aws-sdk/types": "3.6.1"
--- a/packages/api-graphql/package.json
+++ b/packages/api-graphql/package.json
@@ -58,7 +58,7 @@
     "@aws-amplify/pubsub": "5.6.6",
     "graphql": "15.8.0",
     "tslib": "^1.8.0",
-    "uuid": "^3.2.1",
+    "uuid": "^11.1.1",
     "zen-observable-ts": "0.8.19"
   },
   "size-limit": [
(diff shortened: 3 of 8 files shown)


soberm (changes requested):
uuid is direct dependency of multiple Amplify packages. A resolution will not work.

osama-rizk:
Good catch, thanks @soberm — you're right. A root `resolutions` entry only affects this monorepo's own install and isn't published, so consumers running `npm install aws-amplify@5` would still resolve the vulnerable `uuid@3.4.0` via each package's declared range.

I've pushed a new commit that bumps `uuid` to `^11.1.1` as a **direct dependency** in every package that declares it — `analytics`, `api-graphql`, `datastore`, `notifications`, `predictions`, `pubsub` — and removed the root resolution. I also bumped `datastore`'s `@types/uuid` `3.4.6` → `^9.0.0` (the v3 types describe a default export that no longer exists in v11), matching what v6 does.

All source uses named imports (`{ v1 }`/`{ v4 }`), which are API-compatible from uuid v3 through v11, so no source changes were needed. PTAL 🙏

soberm (approved):

PR #14854, opened 2026-07-01
feat(analytics): add configureAutoTrack support for Kinesis and Kinesis Firehose

soberm:
#### Description of changes

Adds `configureAutoTrack` (auto-tracking) support to the **Kinesis** and **Kinesis Firehose** Analytics providers. Auto-tracking (session / pageView / event trackers) previously existed only for Pinpoint; this brings the same capability to the two stream providers.

This change is **purely additive** — Pinpoint behavior is unchanged.

**What it adds**

- `configureAutoTrack(input)` for both providers, mirroring the existing Pinpoint implementation (module-level `configuredTrackers` singleton, `updateProviderTrackers` + `validateTrackerConfiguration`, session/pageView/event tracker types).
- Each auto-tracked event is forwarded to the provider's `record()` API with the data shape `{ name: eventName, attributes }` (mirrors `PinpointAnalyticsEvent`).
- **Per-tracker-type stream configuration**: `streamName` (and `partitionKey` for Kinesis) are supplied in each `configureAutoTrack` call's `options`, so different tracker types can target different streams. Reconfiguring a tracker updates its stream coordinates.
- New input types: `KinesisConfigureAutoTrackInput` / `KinesisFirehoseConfigureAutoTrackInput` and their `Session/PageView/Event` tracking-option types, exported from each provider's public entry point.

**Kinesis Firehose difference**: Firehose has no `partitionKey`, so its options only carry `streamName` and its `record()` call omits `partitionKey`.

**Shared-infra decouple**

- `utils/trackerHelpers.ts` now types `updateProviderTrackers` against the generic `AnalyticsConfigureAutoTrackInput` (from `../types/inputs`) instead of the Pinpoint-specific `ConfigureAutoTrackInput`. Pinpoint's type is a structural alias of the generic one, so Pinpoint continues to compile and behave identically.

**Runtime validation** (options stay optional in the type; validated at runtime)

- Both providers throw `AnalyticsValidationErrorCode.NoStreamName` when `enable: true` and `streamName` is missing.
- Kinesis additionally throws `AnalyticsValidationErrorCode.NoPartitionKey` when `enable: true` and `partitionKey` is missing.
- No validation fires when `enable: false`.

React Native is unchanged — both providers inherit the existing shared validator that restricts auto-tracking to `session` on native.

#### Issue #, if available

N/A

- **Analytics unit tests**: full package suite passes — **29 suites, 102 tests**. New suites:
  - `__tests__/providers/kinesis/apis/configureAutoTrack.test.ts`
  - `__tests__/providers/kinesis-firehose/apis/configureAutoTrack.test.ts`
  - `__tests__/utils/trackerHelpers.test.ts` (proves `updateProviderTrackers` accepts a generic, non-Pinpoint input and forwards options; provider-specific fields are ignored by the shared trackers)
  - Coverage: `validateTrackerConfiguration` invoked; `NoStreamName` / `NoPartitionKey` thrown when enabling without the required fields; no throw when disabling; session/pageView/event trackers fire and call `record()` with the correct `{ streamName[, partitionKey], data: { name, attributes } }`; reconfigure updates stream coordinates; disable cleans trackers up.
- **`aws-amplify` exports test**: passes — **14 tests**. `exports.test.ts` updated so the Kinesis and Kinesis Firehose expected-export arrays include `configureAutoTrack`.
- **TypeScript**: `tsc --noEmit -p tsconfig.build.json` is clean (exit 0), no new errors and no Pinpoint regressions.
- **ESLint**: all new/modified files are clean.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/analytics/src/providers/kinesis/apis/configureAutoTrack.ts
```diff
@@ -0,0 +1,112 @@
...
+			partitionKey: input.options!.partitionKey,
+			data: {
+				name: eventName,
+				attributes,
+			},
+		});
+	};
+
+	// Initialize or update this provider's trackers
+	updateProviderTrackers(input, emitTrackingEvent, configuredTrackers);
```
osama-rizk: **`pageView` auto-track on two providers silently drops half the page views (shared `PageViewTracker` global state).**

Before this PR, auto-track existed only for Pinpoint, so at most one `PageViewTracker` could ever exist. This PR makes concurrent instances reachable — e.g. a consumer mid-migration enabling pageView on **both** Pinpoint and Kinesis. Each provider has its own `configuredTrackers` singleton, so two independent `PageViewTracker` objects are created, but the class holds **process-global state that isn't namespaced**:

1. **Shared `sessionStorage` key clobbers events.** `PageViewTracker` uses a module constant `PREV_URL_STORAGE_KEY = 'aws-amplify-analytics-prevUrl'`. On each navigation both instances' `handleLocationChange` fire; the first records and writes the key to `currentUrl`, so the second's `urlHasChanged()` returns false and it **skips recording**. Each page view lands in only one provider, chosen by listener order — the consumer expecting page views in both silently loses half.

2. **History-proxy stacking breaks teardown.** `setupSPATracking` captures `this.originalPushState = window.history.pushState` and installs a revocable proxy. The second instance captures the *first instance's proxy* as its "original"; on `cleanup()` (disable one provider) the restore/revoke order can leave a **revoked proxy** installed → `TypeError` on the next `pushState`.

`SessionTracker`/`EventTracker` are fine — they fire to distinct destinations, which is the intended "send to both" behavior. This is specific to `PageViewTracker`'s shared `sessionStorage` key + `window.history` patching. The per-provider unit tests can't catch it (they mock the trackers and run one provider in isolation).

Not necessarily a merge-blocker, but worth a conscious decision rather than an accident. Two options:
- **Deep fix:** namespace `PREV_URL_STORAGE_KEY` per provider/recorder and make the history patching reference-counted / instance-safe (in `PageViewTracker`).
- **Cheap + honest:** if multi-provider pageView isn't a supported scenario, document in the `configureAutoTrack` JSDoc that pageView auto-track is single-provider.

(Same concern applies to the Firehose `configureAutoTrack.ts`.)

soberm: Thanks for the careful review. Both issues were valid and are now fixed in commit `92d07a3ca`.

**1. Shared `sessionStorage` "previous URL" key.** The key is now namespaced per provider (`aws-amplify-analytics-prevUrl-<namespace>`, e.g. `...-kinesis` and `...-kinesis-firehose`), so each provider keeps its own previous URL and every provider records the page view instead of the first one clobbering a shared key. Pinpoint keeps the original un namespaced key, so its behavior is unchanged.

**2. Stacked per-instance `history` proxies.** Replaced the per-instance revocable proxies with a single reference counted global patch of `pushState`/`replaceState` that broadcasts a navigation event, which each tracker subscribes to independently. Teardown decrements the ref count and restores the native `pushState`/`replaceState` once the last tracker detaches, so it never leaves a revoked proxy installed and providers can be enabled or disabled in any order.

I also added a regression test at covering concurrent trackers, out of order teardown, and popstate. Beyond the unit tests, I verified this end to end in a real browser against real Kinesis and Firehose: a single SPA navigation is recorded by both providers (confirmed in the delivered data), and disabling providers in a non-LIFO order no longer throws the revoked proxy error.

osama-rizk (approved):

bobbor (approved):
Reviewed — looks good to me. The concurrent multi-provider PageViewTracker concern from the first round (shared sessionStorage key + stacked history proxies) is correctly resolved via per-provider key namespacing and the ref-counted global history patch, with solid regression coverage. Approving.

PR #14866, opened 2026-07-07
feat(notifications): add Amazon Connect Customer Profiles provider for Push Notifications

soberm:
#### Description of changes

Adds an **Amazon Connect Customer Profiles provider for Push Notifications**, available as a new opt-in subpath export alongside the existing Amazon Pinpoint provider:

- `aws-amplify/push-notifications/customer-profiles`
- `@aws-amplify/notifications/push-notifications/customer-profiles`

**What it adds**

- **Identify-user and device lifecycle.** The provider exposes `identifyUser`, `initializePushNotifications`, `registerDevice`, and `removeDevice`, backed by the Amazon Connect Customer Profiles REST endpoint (`/identify-user`, `/register-device`, `/remove-device`). Device registration uses a find-or-create pattern — a stable per-install `deviceId` upsert key plus inflight-registration serialization — so first-login concurrency resolves to a single profile. Because the backend authorizes each call against the caller's derived `principalId`, the device follows the signed-in principal: the provider re-registers the device on the `signedIn` Hub event, which re-homes an existing guest registration to the authenticated principal via the backend's idempotent last-writer-wins upsert on `deviceId` (a returning user's push token is unchanged, so the token listener alone never re-registers). De-registration is authorized the same way, so applications await `removeDevice()` **before** `signOut()` to stop delivery to a device — once sign-out completes the authenticated credentials are gone and the caller signs as a new guest identity. This contract is documented on the `removeDevice` API.
- **SigV4 request signing.** Every request is SigV4-signed for `execute-api` with Identity Pool credentials, so both Cognito user-pool users and guests are authorized through one path. The backend derives `principalId` from the signer identity, which means the client sends no identity field and a guest can register a device token and receive push before sign-in. An Amplify user-agent header is attached to each signed request.
- **`identifyUser` on both React Native and the browser.** On React Native it registers the device token together with any supplied user profile. On the browser, where there is no push device token, it performs a device-less profile identify — associating a `userId` and `userProfile` (for example an email address for contact) with the Customer Profile.
- **Client-side validation.** `UserProfile` and `customAttributes` are validated against the same bounds the backend enforces, and the resolved endpoint must be `https`, so a misconfigured endpoint cannot exfiltrate SigV4-signed tokens over plaintext.
- **Shared transport-agnostic push APIs.** The badge, permission, launch-notification, and notification/token listener APIs (`getBadgeCount`, `setBadgeCount`, `getPermissionStatus`, `requestPermissions`, `getLaunchNotification`, `onNotificationOpened`, `onNotificationReceivedInForeground`, `onNotificationReceivedInBackground`, `onTokenReceived`) live in an internal shared module that both providers consume, so each provider ships a single implementation of everything that is not transport-specific. Message parsing is transport-agnostic as well, so `@aws-amplify/rtn-push-notification` is untouched. `shared/` is internal and is not a published subpath.
- **Configuration.** `Amplify.configure` reads the `notifications.amazon_connect` key from `amplify_outputs.json` into `Notifications.PushNotification.CustomerProfiles = { endpoint, region }`. `@aws-amplify/core` carries the matching optional config type, and Pinpoint configuration remains supported side by side.
- **Deprecation notice on the default entry point.** The default `aws-amplify/push-notifications` entry point is backed by Amazon Pinpoint, for which AWS ends support on October 30, 2026. Each of its 11 runtime APIs now emits a one-time `ConsoleLogger` notice directing customers to the Customer Profiles subpath, which exposes an equivalent for every one of them. The notice names the entry point rather than the individual API, since 9 of the 11 are transport-agnostic and behave identically under either provider.

Every export keeps its name, type, and signature, so this is additive for existing applications: the default and `aws-amplify/push-notifications/pinpoint` entry points still resolve to the Amazon Pinpoint provider, including its engagement telemetry. The Customer Profiles provider emits no Pinpoint engagement telemetry (`_campaign.*`/`_journey.*`), as Amazon Connect Customer Profiles has no client-side event API. The React-Native-only `getDeviceId` is imported lazily so `@aws-amplify/react-native` stays out of the browser bundle on the device-less web path.

Base branch: `main`.

#### Issue #, if available

N/A

- `tsc --noEmit` passes against the build config for `@aws-amplify/notifications`.
- Full library build passes: rollup ESM + CJS plus the webpack UMD bundle. The `customer-profiles` and `pinpoint` subpath dist entries (`index.mjs`, `index.js`) and their `typesVersions` targets (`index.d.ts`) resolve, and the compiled providers require the shared module at runtime.
- `npm run lint` (eslint at `--max-warnings 0`, plus the TypeScript coverage threshold) and `prettier --check` pass on every touched file, and the license header check passes.
- A bidirectional type-assignability check (`Exact<A, B>` in both directions) confirms all 11 default entry point exports keep signatures identical to the underlying provider APIs, so the deprecation notice introduces no type-surface change.
- `npx changeset status` reports a **minor** bump for `@aws-amplify/core`, `@aws-amplify/notifications`, and `aws-amplify`.
- Jest unit tests exercise real code paths, mocking only the network and auth boundary. All **61 suites (261 tests) pass**, of which **47 suites (197 tests)** cover `pushNotifications`:
  - `resolveConfig` — reads `PushNotification.CustomerProfiles` and requires an `https` endpoint.
  - `resolveCredentials` — Identity Pool credentials for authenticated and guest callers, and the missing-credentials error.
  - `signedFetch` — SigV4 `execute-api` signing and the Amplify user-agent header.
  - `validateUserProfile` — `UserProfile` and `customAttributes` bounds matching the backend.
  - `identifyUserInternal` — the nested `options:{deviceId,address,channelType}` backend contract, find-or-create `deviceId` (resolved versus caller-supplied), device-less identify on both the authenticated and guest paths, and non-2xx plus network-failure handling; asserts no Pinpoint `UpdateEndpoint` call.
  - `identifyUser` (web) — device-less profile identify: forwards `userId`, `userProfile`, and `options` to the engine with no device token or channel type.
  - `identifyUser.native` — device and profile registration, inflight-token race, and device option passthrough.
  - `initializePushNotifications.native` — token to Connect registration, the inflight race, re-registration on `signedIn` (including the no-token-yet short circuit), and that `signedOut` attempts no removal.
  - `authStateTransitions.native` — guest/authenticated transitions driven through the REAL credential path: `resolveCredentials`, `signedFetch` and SigV4 `signRequest` are not mocked (only `fetchAuthSession` and `fetch` are stubbed), asserting which identity signs each register/remove call and therefore which `principalId` owns or gates the device row.
  - `registerDevice` / `removeDevice` — web and native.
  - shared APIs (`get*`, `request*`, `setBadgeCount`, `on*`, web and native) — reused by both providers.
  - `deprecatePinpoint` — one notice per wrapped API, independent per-API counters, transparent argument and return delegation, and both synchronous throws and rejected promises reaching the caller untouched.
  - the default entry point — every one of the 11 APIs is invoked for real and asserted to emit the notice, with a companion assertion that the Customer Profiles subpath exposes an equivalent for each.
  - `parseAmplifyOutputs` — the `amazon_connect` notifications key, both alongside Pinpoint and on its own.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/notifications/src/pushNotifications/providers/customer-profiles/utils/signedFetch.ts
```diff
@@ -0,0 +1,90 @@
...
+		{
+			credentials,
+			signingRegion: region,
+			signingService: SIGNING_SERVICE,
+		},
+	);
+
+	let response: Response;
+	try {
+		response = await fetch(url.toString(), {
```
bobbor: this uses the original URL.

but `signed` is an [`HTTRequest`](https://github.com/aws-amplify/amplify-js/blob/main/packages/core/src/clients/types/http.ts#L15) which contains `url` as key, which may or may not have been updated. and if not it's just the original URL (`url`).

IMHO it makes sense to use `signed.url` instead.

packages/notifications/src/pushNotifications/providers/customer-profiles/utils/signedFetch.ts
```diff
@@ -0,0 +1,90 @@
...
+export const signedFetch = async (
+	path: string,
+	body: unknown,
+	action: PushNotificationAction,
+): Promise<void> => {
+	const { endpoint, region } = resolveConfig();
+	const { credentials } = await resolveCredentials();
+
+	const serializedBody = JSON.stringify(body);
+	const url = new URL(`${endpoint}${path}`);
```
bobbor, line 41: nit: endpoint should be trimmed for a trailing `/`.

example: 
```ts
const endpoint = "http://example.com/";
const path = "/foo";

const url = new URL("http://example.com//foo")
// url.origin = "http://example.com"
// url.pathname = "//foo"
```

and this url may get rejected.

bobbor: alt: 

update this: https://github.com/aws-amplify/amplify-js/blob/983a7196d87214018ed305a81cadcb900607e09e/packages/notifications/src/pushNotifications/providers/customer-profiles/utils/resolveConfig.ts#L63

to

```ts
return { endpoint: parsedEndpoint.origin, region };
```

packages/notifications/src/pushNotifications/providers/customer-profiles/utils/getDeviceId.ts
```diff
@@ -0,0 +1,48 @@
...
+
+		return stored;
+	}
+
+	const deviceId = amplifyUuid();
+	await asyncStorage.setItem(DEVICE_ID_STORAGE_KEY, deviceId);
+	cachedDeviceId = deviceId;
+
+	return deviceId;
+};
```
bobbor: this whole function is prone to race-conditioning.
since the storage is async, it is possible to set multiple devices - on the same key though.

while this is theoretically possible, I don't think this is a realistic scenario, and I'll just leave this here as a note 😄 

no action needed

bobbor (commented):
two small things. not really blockers

let me know what you think.

packages/notifications/src/pushNotifications/providers/customer-profiles/apis/initializePushNotifications.native.ts
```diff
@@ -0,0 +1,184 @@
...
+				await registerReceivedDevice(token);
+			} catch (err) {
+				logger.error('Failed to register device for push notifications', err);
+				throw err;
+			}
+		},
+	);
+};
+
+const addAuthListener = (): void => {
```
osama-rizk, line 157: This listener handles `signedOut` → `removeDevice`, but there is no `signedIn` handler. Walking the guest→sign-in flow the PR describes: a guest registers a device (signed with the guest Identity Pool identity, so the backend stores it under `principalId = guest`); the user then signs in and their Identity Pool `identityId` flips to the authenticated identity — but the push token has not changed, so the token-received listener short-circuits (`getToken() === token`) and nothing re-registers. Does the device end up stranded under the guest `principalId`, so messages targeting the signed-in user's principal never reach it? If so, should `addAuthListener` also re-register on `signedIn` (and on token-less identity changes), or should we document that consumers must call `registerDevice` after sign-in? (This may already be handled by backend identity-linking — if so, worth stating.)

packages/notifications/src/pushNotifications/providers/customer-profiles/apis/initializePushNotifications.native.ts
```diff
@@ -0,0 +1,184 @@
...
+	// Deterministically de-register the device from the current principal at
+	// sign-out (the backend remove-device is principalId-gated), closing the
+	// logged-out-then-campaign-fires window. Best-effort — failures are logged.
+	Hub.listen('auth', ({ payload }) => {
+		if (payload.event === 'signedOut') {
+			removeDevice().catch(err => {
```
osama-rizk: `removeDevice()` here ultimately signs via `resolveCredentials()` → `fetchAuthSession()`. Since the `signedOut` Hub event fires after sign-out completes, does `fetchAuthSession` at this point return guest credentials (i.e. `principalId = guest`)? If the device was registered under the authenticated principal, a principal-gated backend removal would then not match, the request fails, and the error is swallowed by `.catch` — leaving the device registered under the logged-out user and keeping open the exact "logged-out-then-campaign-fires" window the comment says it closes. Is removal expected to run with the pre-sign-out (authenticated) credentials? If de-registration is genuinely best-effort, should the comment be softened?

packages/notifications/src/pushNotifications/providers/customer-profiles/utils/validateUserProfile.ts
```diff
@@ -0,0 +1,97 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { PushNotificationValidationErrorCode, assert } from '../../../errors';
+import { UserProfile } from '../types';
+
+// Mirrors amplify-backend validateIdentifyUser — keep in sync.
```
osama-rizk, line 7: These bounds (255-char limit, reserved `principalId` key) are a second copy of the backend `validateIdentifyUser` contract, kept in sync by comment. Is there a shared source of truth we could pull from, or a contract test that would fail if the two drift? Not blocking — defense-in-depth here is reasonable (and rejecting a client-supplied `principalId` is a nice touch) — just flagging the maintenance seam.

osama-rizk:
**E2E coverage before merge?** This adds 37 unit-test files but no end-to-end coverage, and the maintainer checklist item *"add E2E tests for newly added workflows"* is still unchecked. Since this introduces a new SigV4-signed network provider whose behavior spans the auth lifecycle (guest registration → sign-in → sign-out), could we add E2E tests for the Customer Profiles flow in the samples app — or at least stand up a branch with this feature name there to run the existing E2E suite against these changes before merging to main? The unit tests exercise each API in isolation, so the cross-auth-state transitions (see the sign-in / sign-out comments above) would not be caught by them.

packages/notifications/src/pushNotifications/providers/customer-profiles/utils/signedFetch.ts
soberm: Good catch — switched to `signed.url` so we fetch the exact URL the signer produced.

packages/notifications/src/pushNotifications/providers/customer-profiles/utils/signedFetch.ts
soberm: Fixed in resolveConfig by stripping trailing slashes. I kept the path rather than using `.origin`, since `.origin` would drop an API Gateway stage path (e.g. `/prod`). While here I also tightened endpoint validation to require an `execute-api.<region>.amazonaws.com` host, so SigV4-signed requests can only ever go to the real API Gateway host.

packages/notifications/src/pushNotifications/providers/customer-profiles/utils/getDeviceId.ts
soberm: Went ahead and addressed it — getDeviceId now memoizes an in-flight promise so concurrent first-calls share one resolution (single UUID, single write), with the cached promise cleared on error so a failure does not wedge future calls.

osama-rizk (dismissed):

packages/notifications/src/pushNotifications/providers/customer-profiles/apis/initializePushNotifications.native.ts
soberm: Good catch — added a signedIn handler that re-registers the device, which re-homes it to the authenticated principal (backend register-device is an idempotent last-writer-wins upsert on deviceId). Confirmed the returning-user case: the Identity Pool identityId flips guest→authenticated on sign-in while the push token is unchanged, so the token listener short-circuits and nothing re-registered without this.

packages/notifications/src/pushNotifications/providers/customer-profiles/apis/initializePushNotifications.native.ts
soberm: You're right, and it's actually a silent no-op rather than just a wrong-cred match: signOut() clears credentials before the signedOut Hub event fires, so the listener signs with a brand-new guest identity, and the backend's principal-gated delete returns 200 for a non-owner (the no-info-leak invariant) — so .catch never runs and the device stays registered. There's no reliable way to do this automatically today (no pre-signOut hook; and Hub.dispatch is fire-and-forget so a pre-event couldn't be awaited before creds clear). I removed the ineffective listener and documented that apps should call removeDevice() before signOut() to de-register.

packages/notifications/src/pushNotifications/providers/customer-profiles/utils/validateUserProfile.ts
soberm: Intentional defense-in-depth mirroring the backend validateIdentifyUser contract. There's no shared package to pull from (separate repos, and the client runtime shouldn't depend on the CDK package); the values match today (255 + reserved principalId) and the backend 400 is the authoritative gate, so drift only softens a client-side pre-check. Kept the keep-in-sync comment; a contract test is a reasonable future addition.

soberm:
@osama-rizk on E2E: the server contract is covered end-to-end by integration tests in amplify-backend (real SigV4 calls against a deployed Amazon Connect instance / Customer Profiles domain / DynamoDB / HTTP API: identify-user, register, remove, last-writer-wins re-home, and the ownership/leak gate). The gap here is amplify-js client e2e — the samples route needs a new RN push sample plus real APNs/FCM credentials in CI, which is heavy and flaky. To catch the cross-auth-state transitions you flagged, I added unit tests that exercise the real credential/signing path (no boundary mock) and assert the sign-in re-registration and that sign-out no longer attempts a (silently-failing) removal. Happy to file a follow-up for full client e2e.

osama-rizk (approved):

bobbor (approved):

pranavosu (approved):

PR #14867, opened 2026-07-07
feat(core): add ServiceWorker onStateChange telemetry hook

soberm:
#### Description of changes

Adds an optional, vendor-neutral telemetry hook to `ServiceWorker.register()` and deprecates the implicit Pinpoint auto-recording of service worker lifecycle events.

**Motivation**

Today `ServiceWorkerClass` hard-wires service worker `statechange` telemetry to Amazon Pinpoint (`Analytics.Pinpoint`). Consumers who use a different analytics/observability backend — or who don't use Pinpoint at all — have no way to observe SW lifecycle events without pulling in Pinpoint. This change decouples SW lifecycle telemetry from Pinpoint so consumers can emit vendor-neutral telemetry.

**What changed**

- New `register()` option: `register(filePath?, scope?, { onStateChange })`.
- New types exported from `@aws-amplify/core` (and re-exported from `aws-amplify/utils`):
  - `type ServiceWorkerStateChangeHandler = (state: ServiceWorkerState) => void`
  - `interface ServiceWorkerOptions { onStateChange?: ServiceWorkerStateChangeHandler }`
- When `onStateChange` is provided, it is invoked on every `statechange` event.

**Behavior (additive & backwards compatible)**

- **No hook provided → identical to today:** the built-in Pinpoint event is still recorded (when Pinpoint is configured and credentials are available).
- **Hook provided → the hook fires and the built-in Pinpoint path is suppressed**, so the same state change is never double-recorded (hook overrides built-in).
- The consumer hook is invoked inside a `try/catch`; a throwing handler is logged via `logger.error` and never rejects out of the async listener.
- The handler is captured per `register()` call, so re-registering with a different handler only affects its own listener.

**Deprecation**

The implicit Pinpoint auto-recording of SW lifecycle events is deprecated in favor of `onStateChange` and is planned for removal in a future major version. Only the implicit Pinpoint recording is deprecated — `register()` itself is not. This is documented in the `register()` JSDoc as prose (deliberately not a method-level `@deprecated` tag, to avoid IDEs striking through all `register()` usages).

**Versioning**

Minor bump for `@aws-amplify/core` and `aws-amplify` (new opt-in public API; no breaking changes). Changeset included: `.changeset/serviceworker-onstatechange-hook.md`.

#### Issue #, if available

N/A — no linked issue.

- **Unit tests** (`packages/core/__tests__/ServiceWorker.test.ts`) — 20/20 passing, including new cases:
  - records a Pinpoint event when no handler is provided (unchanged behavior)
  - invokes `onStateChange` and suppresses built-in recording when a handler is provided (no double-record)
  - does not record when Pinpoint is not configured and no handler is provided
  - logs and swallows errors thrown by the handler (`logger.error`, no reject, no `record`)
  - invokes the handler once per `statechange` event
  - captures the handler per registration (re-register does not re-target a prior listener)
- **Type check** — `tsc --noEmit -p packages/core/tsconfig.json` (src + `__tests__`): identical error set before/after this change (zero new type errors).
- **Lint** — `eslint --fix` on all changed files reports zero new errors/warnings.
- **Changeset** — `changeset status` confirms minor bumps for `@aws-amplify/core` and `aws-amplify`.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes — ran the affected `@aws-amplify/core` `ServiceWorker` suite (20/20 passing); the full monorepo `yarn test` was not executed in this environment
- [x] Relevant documentation is changed or added (JSDoc on `register()` and the new exported types)

### DIFF
--- a/packages/core/src/ServiceWorker/ServiceWorker.ts
+++ b/packages/core/src/ServiceWorker/ServiceWorker.ts
@@ -10,6 +10,28 @@ import { Amplify, fetchAuthSession } from '../singleton';
 import { ServiceWorkerErrorCode, assert } from './errorHelpers';
 
 /**
+ * Handler invoked on every service worker `statechange` event, receiving the
+ * worker's current lifecycle state.
+ */
+export type ServiceWorkerStateChangeHandler = (
+	state: ServiceWorkerState,
+) => void;
+
+/**
+ * Options for {@link ServiceWorker.register}.
+ */
+export interface ServiceWorkerOptions {
+	/**
+	 * Optional handler invoked on every service worker `statechange` event.
+	 *
+	 * When provided, this handler replaces the built-in Pinpoint auto-recording:
+	 * the built-in analytics event is only recorded when no handler is supplied,
+	 * which prevents duplicate telemetry for the same state change.
+	 */
+	onStateChange?: ServiceWorkerStateChangeHandler;
+}
+
+/**
  * Provides a means to registering a service worker in the browser
  * and communicating with it via postMessage events.
  * https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/
@@ -35,6 +57,9 @@ export class ServiceWorkerClass {
 	// push subscription
 	private _subscription?: PushSubscription;
 
+	// Optional handler invoked on every service worker state change
+	private _onStateChange?: ServiceWorkerStateChangeHandler;
+
 	// The AWS Amplify logger
 	private _logger: ConsoleLogger = new ConsoleLogger('ServiceWorker');
 
@@ -55,14 +80,29 @@ export class ServiceWorkerClass {
 	 * Make sure the service-worker.js is part of the build
 	 * for example with Angular, modify the angular-cli.json file
 	 * and add to "assets" array "service-worker.js"
+	 *
+	 * Note: when `options.onStateChange` is omitted, this method implicitly
+	 * records service worker lifecycle (`statechange`) events to Amazon
+	 * Pinpoint. That built-in auto-recording is deprecated and will be removed
+	 * in a future major version — only the implicit Pinpoint recording is
+	 * deprecated, not `register()` itself. Provide `options.onStateChange` to
+	 * observe lifecycle state changes and emit vendor-neutral telemetry instead.
 	 * @param {string} filePath Service worker file. Defaults to "/service-worker.js"
 	 * @param {string} scope The service worker scope. Defaults to "/"
 	 *  - API Doc: https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/register
+	 * @param {ServiceWorkerOptions} [options] Optional registration options. When
+	 * `onStateChange` is provided it is invoked on every service worker state
+	 * change and replaces the built-in Pinpoint auto-recording.
 	 * @returns {Promise}
 	 *	- resolve(ServiceWorkerRegistration)
 	 *	- reject(Error)
 	 **/
-	register(filePath = '/service-worker.js', scope = '/') {
+	register(
+		filePath = '/service-worker.js',
+		scope = '/',
+		options?: ServiceWorkerOptions,
+	) {
+		this._onStateChange = options?.onStateChange;
 		this._logger.debug(`registering ${filePath}`);
 		this._logger.debug(`registering service worker with scope ${scope}`);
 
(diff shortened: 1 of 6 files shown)


packages/core/src/ServiceWorker/ServiceWorker.ts
```diff
@@ -210,40 +250,60 @@ export class ServiceWorkerClass {
...
-			} = Amplify.getConfig().Analytics?.Pinpoint ?? {};
-			const { credentials } = await fetchAuthSession();
-
-			if (appId && region && credentials) {
-				// Pinpoint is configured, record an event
-				record({
+			// Notify a consumer-provided handler, isolating any error it throws so
+			// it cannot surface as an unhandled rejection from the listener.
+			try {
+				onStateChange?.(currentState);
```
osama-rizk: **Async throwing handler escapes this try/catch, defeating the "never rejects" guarantee.**

`ServiceWorkerStateChangeHandler = (state) => void` structurally accepts an `async` function (TS ignores a `=> void` return). Since `onStateChange?.(currentState)` isn't awaited, an async handler returns a *rejected promise* the synchronous try/catch can't see → unhandled rejection, contradicting the JSDoc's "never rejects out of the async listener."

The "logs and swallows errors" test only covers a **synchronous** `() => { throw }` — the one shape that already works — so the async case is untested.

Fix: `await onStateChange?.(currentState);` inside the try (safe for sync handlers too — `await undefined` resolves immediately), plus a regression test with `async () => { throw ... }`.

packages/core/src/ServiceWorker/ServiceWorker.ts
```diff
@@ -210,40 +250,60 @@ export class ServiceWorkerClass {
 	/**
 	 * Listen for service worker state change and message events
 	 * https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorker/state
+	 *
+	 * Each call to `register()` attaches its own `statechange` listener. The
+	 * `onStateChange` handler is captured in a local at listener-creation time,
+	 * so re-registering with a different handler only affects its own listener
+	 * and never re-targets a previously attached one.
 	 **/
 	_setupListeners() {
```
osama-rizk, line 263: **An already-active worker is never observed — hook/Pinpoint gap.**

This only wires `addEventListener('statechange', …)`, which fires only on *transitions*. On a repeat visit the SW is often already `activated`, so `register()` takes the `registration.active` branch, no `statechange` fires, and `onStateChange` (and Pinpoint) never sees the current state — a telemetry gap for exactly the returning-user population consumers most want to measure.

Pre-existing for Pinpoint, but the new public hook creates a reasonable expectation that an observer reports the current state. Worth closing while the API is young: after attaching the listener, if `this._serviceWorker?.state` is already set, invoke the same notify-or-record path once with that state (the "check readyState after attaching onopen" idiom).

packages/core/src/ServiceWorker/ServiceWorker.ts
soberm: Fixed in 9b5d357e5. The handler is now awaited inside the try/catch (`await onStateChange?.(currentState)`), so a rejected promise from an async handler is caught rather than escaping as an unhandled rejection; `await undefined` resolves immediately for sync/absent handlers. Added a regression test with `async () => { throw ... }` asserting the listener resolves, the error is logged via `logger.error`, and the built-in Pinpoint path stays suppressed. Verified the old code crashed the Jest worker on the async-throw case while the fix passes (21/21).

packages/core/src/ServiceWorker/ServiceWorker.ts
soberm: Good catch — fixed in d84e8fd11. After attaching the `statechange` listener, `register()` now reads the worker's current state and, if a handler was provided, emits it once to `onStateChange` (via a shared `_notifyStateChange` helper), so an already-`activated` worker on a repeat visit is still observed.

Deliberate scope decision: the initial emit fires for the `onStateChange` hook **only** — it does NOT invoke the built-in Pinpoint path. Recording a Pinpoint event on the already-active branch would emit telemetry that never existed before for existing consumers, breaking this PR's "no hook provided = identical to today" backwards-compatibility guarantee. The new public hook is where we can reasonably set the "observer reports current state" expectation.

Added regression tests: (1) the handler is emitted once with the current state on register with no `statechange` dispatched (and Pinpoint is not touched), and (2) no Pinpoint record occurs on register when no handler is provided. Full suite 23/23 passing, lint clean.

osama-rizk (dismissed):

bobbor (changes requested):
Thanks — the feature design is clean and additive, and the error-isolation and test coverage are strong. One change requested before merge:

**Tests mock `Amplify.getConfig` on the singleton** (`packages/core/__tests__/ServiceWorker.test.ts`):

```ts
jest.spyOn(Amplify, 'getConfig').mockReturnValue({
	Analytics: { Pinpoint: pinpointConfig },
} as any);
```

Our testing convention is to not mock `getConfig` on the Amplify singleton — mock the actual underlying modules/functions instead (real Amplify config approach). This also drags in an `as any`. The PR already mocks `../src/providers/pinpoint` and `fetchAuthSession` the right way — please provide the Pinpoint config through real configuration (e.g. `Amplify.configure(...)` with a test config) rather than stubbing `getConfig`.

soberm:
Addressed in 957b865da. The `State change telemetry` tests no longer mock `Amplify.getConfig` on the singleton — the Pinpoint config is now provided through real configuration via `Amplify.configure({ Analytics: { Pinpoint: pinpointConfig } })` in `beforeEach`, and the "not configured" case uses `Amplify.configure({})`. This also removes the `as any` casts. The existing `../src/providers/pinpoint` and `fetchAuthSession` mocks are unchanged. Full suite still 23/23 passing, lint clean.

osama-rizk (approved):

bobbor (approved):

osama-rizk (approved):

PR #14874, opened 2026-07-10
feat(storage): migrate storage APIs to AmplifyContext

bobbor:
## Description

Migrates the `@aws-amplify/storage` package from the global `Amplify` singleton to explicit `AmplifyContext` threading, following the same pattern as the landed auth migration (#14836). This is the **B-storage** step of the v6→v7 context migration, targeting the `feat/bobbor/v6-context` feature branch.

### What changed

- **Browser public APIs** (`copy`, `downloadData`, `getProperties`, `getUrl`, `list`, `remove`, `uploadData`): added `(ctx, input)` overloads alongside `(input)`, with a variadic impl using `resolveCtxArgs` (global-context fallback preserved for existing callers).
- **Internal workers, `resolveS3ConfigAndInput`, access-grant internals + `listPaths`**: now take `ctx: AmplifyContext` first; config via `ctx.resourcesConfig` (not `Amplify.getConfig()`), auth via `ctx.fetchAuthSession()`.
- **Server wrappers**: accept `AmplifyContext | AmplifyServer.ContextSpec` via a new `resolveServerContext`, preserving adapter-nextjs compatibility. Server impl files are **not** deleted.
- **Tests**: migrated to a branded mock `AmplifyContext` (`createMockAmplifyContext`); underlying modules mocked rather than `Amplify.getConfig`.

### Deliberately out of scope (vs v7-poc)

- The **endpoint-provider feature** (`endpointProvider`/`forcePathStyle`) — depends on core `Storage/types.ts` changes not yet on this base branch. `resolveS3ConfigAndInput`/`base.ts` retain the existing `LOCAL_TESTING_S3_ENDPOINT` logic.
- **Deleting the server impls** — v7-poc removes them assuming a later adapter-nextjs split supplies ctx; that split isn't landed, so they're preserved here.

### Notable fix

`resolveServerContext` adapts the unwrapped `AmplifyClass` (from `getAmplifyServerContext`) into a real `AmplifyContext` by bridging `fetchAuthSession`/`clearCredentials`/`getTokens` to `amplify.Auth.*`. A bare `AmplifyClass` lacks those top-level methods, so a plain cast would throw on the server path at runtime. Uses the `isAmplifyContext` brand check rather than a structural probe. A guarding unit test covers this.

> Note: the landed auth `resolveServerContext` (#14836) has the same latent server-path issue (`getCurrentUser` calls `ctx.getTokens` on a method-less `AmplifyClass`). Recommend a follow-up, ideally a core-level fix so all categories share one correct server bridge.

## Testing

- `yarn build --scope @aws-amplify/storage` → passes
- `yarn test --scope @aws-amplify/storage` → **86 suites / 858 tests pass**, lint clean, ts-coverage passes; `resolveServerContext.ts` at 100% coverage

## Checklist

- [x] Unit tests updated/added; coverage preserved
- [x] No changeset (targets a feature branch, not `main`)
- [x] No `tsconfig.tsbuildinfo` committed
- [x] Server/browser split preserved

### Note: `internals/` APIs take a required `ctx` (no global fallback)

Unlike the public S3 APIs (which keep the zero-ctx overload via `resolveCtxArgs` for backward compatibility), the `internals/apis/*` functions re-exported from `@aws-amplify/storage/internals` now take a **required** `AmplifyContext` first parameter. This is intentional: this surface is consumed by first-party packages (StorageBrowser / `ui-react-storage`) that migrate in lockstep on the `v6-context` feature branch.

### Note: includes a base-branch test fix (`test(auth): …signInWithRedirect`)
The first commit fixes 4 `prompt=NONE` tests in `packages/auth` that were broken on the base branch itself (they came from #14902 on main pre-migration and call `signInWithRedirect` without a context; the base rebase replayed the auth migration over them). The fix rides this PR and lands on the base when it merges — the base branch receives only PR merges.

### DIFF
--- a/packages/aws-amplify/package.json
+++ b/packages/aws-amplify/package.json
@@ -341,7 +341,7 @@
 			"name": "[Analytics] identifyUser (Pinpoint)",
 			"path": "./dist/esm/analytics/index.mjs",
 			"import": "{ identifyUser }",
-			"limit": "18 kB"
+			"limit": "18.1 kB"
 		},
 		{
 			"name": "[Analytics] enable",
--- a/packages/storage/__tests__/providers/s3/apis/copy.test.ts
+++ b/packages/storage/__tests__/providers/s3/apis/copy.test.ts
@@ -1,17 +1,32 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { Amplify } from '@aws-amplify/core';
+import {
+	clearGlobalContext,
+	setGlobalContext,
+} from '@aws-amplify/core/internals/utils';
 
 import { CopyInput, CopyWithPathInput } from '../../../../src';
 import { copy } from '../../../../src/providers/s3/apis';
 import { copy as internalCopyImpl } from '../../../../src/providers/s3/apis/internal/copy';
+import { createMockAmplifyContext } from '../../../testUtils/mockAmplifyContext';
 
 jest.mock('../../../../src/providers/s3/apis/internal/copy');
 
 const mockInternalCopyImpl = jest.mocked(internalCopyImpl);
+const mockCtx = createMockAmplifyContext();
 
 describe('client-side copy', () => {
+	beforeAll(() => {
+		// The public API falls back to the global AmplifyContext when no ctx is
+		// passed explicitly; establish it so resolveCtxArgs can resolve it.
+		setGlobalContext(mockCtx);
+	});
+
+	afterAll(() => {
+		clearGlobalContext();
+	});
+
 	beforeEach(() => {
 		jest.clearAllMocks();
 	});
@@ -28,7 +43,7 @@ describe('client-side copy', () => {
 			},
 		};
 		expect(copy(input)).toEqual(mockInternalResult);
-		expect(mockInternalCopyImpl).toBeCalledWith(Amplify, input);
+		expect(mockInternalCopyImpl).toBeCalledWith(mockCtx, input);
 	});
 
 	it('should pass through input with path and output to internal implementation', async () => {
@@ -39,6 +54,6 @@ describe('client-side copy', () => {
 			destination: { path: 'abc' },
 		};
 		expect(copy(input)).toEqual(mockInternalResult);
-		expect(mockInternalCopyImpl).toBeCalledWith(Amplify, input);
+		expect(mockInternalCopyImpl).toBeCalledWith(mockCtx, input);
 	});
 });
--- a/packages/storage/__tests__/providers/s3/apis/copy.test.ts
+++ b/packages/storage/__tests__/providers/s3/apis/copy.test.ts
@@ -56,4 +56,16 @@ describe('client-side copy', () => {
 		expect(copy(input)).toEqual(mockInternalResult);
 		expect(mockInternalCopyImpl).toBeCalledWith(mockCtx, input);
 	});
+
+	it('should pass explicit AmplifyContext to internal implementation when called with two args', () => {
+		const explicitCtx = createMockAmplifyContext();
+		const mockInternalResult = 'RESULT' as any;
+		mockInternalCopyImpl.mockReturnValue(mockInternalResult);
+		const input: CopyWithPathInput = {
+			source: { path: 'source-path' },
+			destination: { path: 'dest-path' },
+		};
+		expect(copy(explicitCtx, input)).toEqual(mockInternalResult);
+		expect(mockInternalCopyImpl).toBeCalledWith(explicitCtx, input);
+	});
 });
--- a/packages/storage/__tests__/providers/s3/apis/internal/copy.test.ts
+++ b/packages/storage/__tests__/providers/s3/apis/internal/copy.test.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { AWSCredentials } from '@aws-amplify/core/internals/utils';
-import { Amplify, StorageAccessLevel } from '@aws-amplify/core';
+import { AmplifyContext, StorageAccessLevel } from '@aws-amplify/core';
 
 import { StorageError } from '../../../../../src/errors/StorageError';
 import { StorageValidationErrorCode } from '../../../../../src/errors/types/validation';
@@ -18,20 +18,21 @@ import './testUtils';
 import { BucketInfo } from '../../../../../src/providers/s3/types/options';
 
 jest.mock('../../../../../src/providers/s3/utils/client/s3data');
-jest.mock('@aws-amplify/core', () => ({
-	ConsoleLogger: jest.fn().mockImplementation(function ConsoleLogger() {
-		return { debug: jest.fn() };
-	}),
-	Amplify: {
-		getConfig: jest.fn(),
-		Auth: {
-			fetchAuthSession: jest.fn(),
-		},
-	},
-}));
 const mockCopyObject = copyObject as jest.Mock;
-const mockFetchAuthSession = Amplify.Auth.fetchAuthSession as jest.Mock;
-const mockGetConfig = Amplify.getConfig as jest.Mock;
+const mockGetConfig = jest.fn();
+const mockFetchAuthSession = jest.fn();
+// Internal workers now receive an AmplifyContext. Back resourcesConfig with a
+// jest.fn so existing per-test config mocking keeps working, and expose
+// fetchAuthSession as a jest.fn for credential/identityId control.
+const mockCtx: AmplifyContext = {
+	get resourcesConfig() {
+		return mockGetConfig();
+	},
+	libraryOptions: {},
+	fetchAuthSession: mockFetchAuthSession,
+	clearCredentials: jest.fn(),
+	getTokens: jest.fn(),
+};
 
 const sourceKey = 'sourceKey';
 const destinationKey = 'destinationKey';
@@ -75,7 +76,7 @@ describe('copy API', () => {
 
 	describe('Happy Cases', () => {
 		describe('With key', () => {
-			const copyWrapper = async (input: CopyInput) => copy(Amplify, input);
+			const copyWrapper = async (input: CopyInput) => copy(mockCtx, input);
 			beforeEach(() => {
 				mockCopyObject.mockImplementation(() => {
 					return {
@@ -296,7 +297,7 @@ describe('copy API', () => {
 
 		describe('With path', () => {
 			const copyWrapper = async (input: CopyWithPathInput) =>
-				copy(Amplify, input);
+				copy(mockCtx, input);
 
 			beforeEach(() => {
 				mockCopyObject.mockImplementation(() => {
@@ -495,7 +496,7 @@ describe('copy API', () => {
 			expect.assertions(3);
 			const missingSourceKey = 'SourceKeyNotFound';
 			try {
-				await copy(Amplify, {
+				await copy(mockCtx, {
 					source: { key: missingSourceKey },
 					destination: { key: destinationKey },
 				});
@@ -518,7 +519,7 @@ describe('copy API', () => {
 			expect.assertions(2);
 			try {
 				// @ts-expect-error mismatch copy input not allowed
-				await copy(Amplify, {
+				await copy(mockCtx, {
 					source: { path: 'sourcePath' },
 					destination: { key: 'destinationKey' },
 				});
@@ -533,7 +534,7 @@ describe('copy API', () => {
 			expect.assertions(2);
 			try {
 				// @ts-expect-error mismatch copy input not allowed
-				await copy(Amplify, {
+				await copy(mockCtx, {
 					source: { key: 'sourcePath' },
 					destination: { path: 'destinationKey' },
 				});
@@ -546,7 +547,7 @@ describe('copy API', () => {
 		it('should throw an error when only source has bucket option', async () => {
 			expect.assertions(2);
 			try {
-				await copy(Amplify, {
+				await copy(mockCtx, {
 					source: { path: 'source', bucket: 'bucket-1' },
 					destination: {
 						path: 'destination',
@@ -563,7 +564,7 @@ describe('copy API', () => {
 		it('should throw an error when only one destination has bucket option', async () => {
 			expect.assertions(2);
 			try {
-				await copy(Amplify, {
+				await copy(mockCtx, {
 					source: { key: 'source' },
 					destination: {
 						key: 'destination',
--- a/packages/storage/__tests__/providers/s3/apis/internal/copy.test.ts
+++ b/packages/storage/__tests__/providers/s3/apis/internal/copy.test.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { AWSCredentials } from '@aws-amplify/core/internals/utils';
-import { AmplifyContext, StorageAccessLevel } from '@aws-amplify/core';
+import { StorageAccessLevel } from '@aws-amplify/core';
 
 import { StorageError } from '../../../../../src/errors/StorageError';
 import { StorageValidationErrorCode } from '../../../../../src/errors/types/validation';
@@ -16,23 +16,14 @@ import {
 } from '../../../../../src/providers/s3/types';
 import './testUtils';
 import { BucketInfo } from '../../../../../src/providers/s3/types/options';
+import { createMockAmplifyContext } from '../../../../testUtils/mockAmplifyContext';
 
 jest.mock('../../../../../src/providers/s3/utils/client/s3data');
 const mockCopyObject = copyObject as jest.Mock;
 const mockGetConfig = jest.fn();
-const mockFetchAuthSession = jest.fn();
-// Internal workers now receive an AmplifyContext. Back resourcesConfig with a
-// jest.fn so existing per-test config mocking keeps working, and expose
-// fetchAuthSession as a jest.fn for credential/identityId control.
-const mockCtx: AmplifyContext = {
-	get resourcesConfig() {
-		return mockGetConfig();
-	},
-	libraryOptions: {},
-	fetchAuthSession: mockFetchAuthSession,
-	clearCredentials: jest.fn(),
-	getTokens: jest.fn(),
-};
+// Live getter delegates to mockGetConfig so per-test config variance works.
+const mockCtx = createMockAmplifyContext({ getConfig: mockGetConfig });
+const mockFetchAuthSession = jest.mocked(mockCtx.fetchAuthSession);
 
 const sourceKey = 'sourceKey';
 const destinationKey = 'destinationKey';
--- a/packages/storage/__tests__/testUtils/mockAmplifyContext.ts
+++ b/packages/storage/__tests__/testUtils/mockAmplifyContext.ts
@@ -1,0 +1,30 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+	AMPLIFY_CONTEXT_BRAND,
+	AmplifyContext,
+	ResourcesConfig,
+} from '@aws-amplify/core';
+
+/**
+ * Creates a mock AmplifyContext for testing.
+ */
+export function createMockAmplifyContext(
+	resourcesConfig: ResourcesConfig = {},
+): AmplifyContext {
+	const ctx: AmplifyContext = {
+		resourcesConfig,
+		libraryOptions: {},
+		fetchAuthSession: jest.fn().mockResolvedValue({}),
+		clearCredentials: jest.fn().mockResolvedValue(undefined),
+		getTokens: jest.fn().mockResolvedValue(undefined),
+	};
+
+	Object.defineProperty(ctx, AMPLIFY_CONTEXT_BRAND, {
+		value: true,
+		enumerable: false,
+	});
+
+	return ctx;
+}
--- a/packages/storage/__tests__/testUtils/mockAmplifyContext.ts
+++ b/packages/storage/__tests__/testUtils/mockAmplifyContext.ts
@@ -7,14 +7,34 @@ import {
 	ResourcesConfig,
 } from '@aws-amplify/core';
 
+export interface MockAmplifyContextOptions {
+	/** When provided, resourcesConfig becomes a live getter delegating to this fn. */
+	getConfig?(): ResourcesConfig;
+}
+
 /**
  * Creates a mock AmplifyContext for testing.
  */
 export function createMockAmplifyContext(
-	resourcesConfig: ResourcesConfig = {},
+	resourcesConfigOrOpts?: ResourcesConfig | MockAmplifyContextOptions,
 ): AmplifyContext {
+	// Determine whether we received the options form or raw ResourcesConfig.
+	const isOpts =
+		resourcesConfigOrOpts != null &&
+		'getConfig' in resourcesConfigOrOpts &&
+		typeof (resourcesConfigOrOpts as MockAmplifyContextOptions).getConfig ===
+			'function';
+
 	const ctx: AmplifyContext = {
-		resourcesConfig,
+		// Use a live getter when getConfig is supplied so per-test config changes propagate.
+		get resourcesConfig() {
+			if (isOpts) {
+				return (resourcesConfigOrOpts as MockAmplifyContextOptions)
+					.getConfig!();
+			}
+
+			return (resourcesConfigOrOpts as ResourcesConfig) ?? {};
+		},
 		libraryOptions: {},
 		fetchAuthSession: jest.fn().mockResolvedValue({}),
 		clearCredentials: jest.fn().mockResolvedValue(undefined),
--- a/packages/storage/src/foundation/types/context.ts
+++ b/packages/storage/src/foundation/types/context.ts
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { AmplifyClassV6 } from '@aws-amplify/core';
+import { AmplifyContext } from '@aws-amplify/core';
 
 /**
  * Read a `Blob` (or `File`) into an `ArrayBuffer`. Implementation differs
@@ -23,7 +23,7 @@ export type ToBase64 = (input: string | ArrayBufferView) => string;
  * Per-invocation context passed into foundation-layer functions by the
  * client or server layer. It carries dependencies that are either
  * environment-specific or environment-scoped (e.g. a request-scoped
- * `AmplifyClassV6` on the server).
+ * `AmplifyContext` on the server).
  *
  * All environment-specific behavior the foundation needs MUST be injected
  * through this object so the foundation layer stays free of any
@@ -34,7 +34,7 @@ export interface FoundationContext {
 	 * The Amplify instance. On the client this is the global singleton;
 	 * on the server this is a request-scoped instance from the server adapter.
 	 */
-	amplify: AmplifyClassV6;
+	amplify: AmplifyContext;
 
 	/**
 	 * Reads a `Blob` into an `ArrayBuffer`.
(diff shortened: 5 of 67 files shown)


packages/storage/__tests__/providers/s3/apis/copy.test.ts
```diff
@@ -28,7 +43,7 @@ describe('client-side copy', () => {
 			},
 		};
 		expect(copy(input)).toEqual(mockInternalResult);
```
soberm, line 45: [major] All tests in this file (and the equivalent `downloadData`, `getUrl`, `getProperties`, `list`, `remove`, `uploadData` public-API suites) exercise only the single-argument overload via the global context fallback. The new explicit-context overloads `copy(ctx, input)` are never called directly anywhere in the public-API test suite. If `resolveCtxArgs` dispatch changes or the two-arg branch breaks, nothing here will catch it. Add at least one test per API that calls the two-arg form directly without relying on `setGlobalContext`.

packages/storage/src/providers/s3/apis/server/resolveServerContext.ts
```diff
@@ -0,0 +1,42 @@
...
+	// Legacy server ContextSpec: unwrap the `AmplifyClass` and adapt it to the
+	// `AmplifyContext` shape. `AmplifyClass` exposes resourcesConfig/libraryOptions
+	// and a cross-category `Auth` utility, but NOT the top-level context methods
+	// (fetchAuthSession/clearCredentials/getTokens), so bridge them to `Auth.*` here.
+	const { amplify } = getAmplifyServerContext(ctxOrContextSpec);
+
+	// Annotate the object so the bridged lambdas receive contextual parameter
+	// types (avoids implicit-any) and the shape is checked against AmplifyContext.
+	const resolved: AmplifyContext = {
+		resourcesConfig: amplify.getConfig(),
```
soberm: [minor] `resourcesConfig` is evaluated once via `amplify.getConfig()` at the time `resolveServerContext` is called and stored as a static value. Every other `AmplifyContext` consumer in this codebase (including the manually-built `mockCtx` objects in the test suite) exposes `resourcesConfig` as a live getter. On a server request this is harmless in practice, but the inconsistency could surprise callers who hold a resolved context across a config update. Consider: `get resourcesConfig() { return amplify.getConfig(); }`.

packages/storage/src/internals/apis/listPaths/listPaths.ts
```diff
@@ -1,22 +1,24 @@
...
 import { getHighestPrecedenceUserGroup } from './getHighestPrecedenceUserGroup';
 
-export const listPaths = async (): Promise<ListPathsOutput> => {
-	const { buckets } = Amplify.getConfig().Storage!.S3!;
-	const { groups } = Amplify.getConfig().Auth!.Cognito;
+export const listPaths = async (
+	ctx: AmplifyContext,
+): Promise<ListPathsOutput> => {
+	const { buckets } = ctx.resourcesConfig.Storage!.S3!;
+	const { groups } = ctx.resourcesConfig.Auth!.Cognito;
```
soberm: [minor] Both `Storage!.S3!` and `Auth!` are forced non-null assertions. If a caller passes a context that has `Storage` configured without an `S3` sub-key, or a storage-only context without `Auth`, these throw a bare `TypeError: Cannot read properties of undefined` instead of a descriptive `StorageError`. Now that `listPaths` accepts an explicit `AmplifyContext`, callers can supply partial configs more easily than with the global singleton. Guard with `assertValidationError` + an appropriate `StorageValidationErrorCode`, or use optional chaining before reading `.Cognito`.

packages/storage/__tests__/testUtils/mockAmplifyContext.ts
```diff
@@ -0,0 +1,30 @@
...
+	AMPLIFY_CONTEXT_BRAND,
+	AmplifyContext,
+	ResourcesConfig,
+} from '@aws-amplify/core';
+
+/**
+ * Creates a mock AmplifyContext for testing.
+ */
+export function createMockAmplifyContext(
+	resourcesConfig: ResourcesConfig = {},
```
soberm: [minor] The factory only accepts a static `resourcesConfig` object. The six internal worker test files (`copy`, `downloadData`, `getProperties`, `getUrl`, `list`, `remove`) cannot use it because they need `resourcesConfig` to be a getter backed by a per-test `jest.fn()` so they can vary config between cases. As a result, all six files duplicate the same manual `AmplifyContext` shape. Consider an options overload: `createMockAmplifyContext({ getConfig?: () => ResourcesConfig })` so the getter pattern is centralised and future worker tests have one place to update.

packages/storage/src/providers/s3/apis/server/resolveServerContext.ts
```diff
@@ -0,0 +1,42 @@
...
+		libraryOptions: amplify.libraryOptions,
+		fetchAuthSession: options => amplify.Auth.fetchAuthSession(options ?? {}),
```
soberm, line 40: [question] The `?? {}` default ensures `amplify.Auth.fetchAuthSession` always receives a non-undefined argument, and `resolveServerContext.test.ts` line 79 explicitly asserts this (`toHaveBeenCalledWith({})`), so the behavior is intentional. Add a short comment here explaining why the default is required (i.e., `AmplifyClass.Auth.fetchAuthSession` requires a non-optional first argument, unlike `AmplifyContext.fetchAuthSession`) so future readers do not silently drop it.

packages/storage/src/providers/s3/apis/copy.ts
```diff
@@ -36,7 +37,17 @@ export function copy(input: CopyWithPathInput): Promise<CopyWithPathOutput>;
  * source or destination key is not defined.
  */
 export function copy(input: CopyInput): Promise<CopyOutput>;
+export function copy(
```
soberm, line 44: [nit] The two new `(ctx, input)` overload signatures have no JSDoc. The zero-ctx overloads above them carry full `@param`/`@throws` documentation. Add at minimum a `@param ctx` line for the `AmplifyContext` parameter, consistent with the pattern in the landed auth migration.

packages/storage/__tests__/providers/s3/apis/internal/copy.test.ts
```diff
@@ -18,20 +18,21 @@ import './testUtils';
...
-}));
 const mockCopyObject = copyObject as jest.Mock;
-const mockFetchAuthSession = Amplify.Auth.fetchAuthSession as jest.Mock;
-const mockGetConfig = Amplify.getConfig as jest.Mock;
+const mockGetConfig = jest.fn();
+const mockFetchAuthSession = jest.fn();
+// Internal workers now receive an AmplifyContext. Back resourcesConfig with a
+// jest.fn so existing per-test config mocking keeps working, and expose
+// fetchAuthSession as a jest.fn for credential/identityId control.
+const mockCtx: AmplifyContext = {
```
soberm: [nit] This three-line comment narrates what the code already shows. The type annotation, getter, and `jest.fn` declarations are self-explanatory. The only non-obvious point worth noting is WHY a getter is used instead of a static value (per-test config variance). Either remove the comment block or collapse it to a single line covering that decision. The same block is duplicated verbatim in `downloadData.test.ts`, `getProperties.test.ts`, `getUrl.test.ts`, `list.test.ts`, and `remove.test.ts`.

packages/storage/src/internals/apis/getUrl.ts
```diff
@@ -10,8 +10,8 @@ import { GetUrlOutput } from '../types/outputs';
 /**
  * @internal
  */
-export const getUrl = (input: GetUrlInput) =>
-	getUrlInternal(Amplify, {
+export const getUrl = (ctx: AmplifyContext, input: GetUrlInput) =>
```
osama-rizk, line 13: These `internals/apis/*` functions (copy, downloadData, getProperties, getUrl, list, listPaths, remove, uploadData) now take a **required** `ctx` with no global fallback — unlike the public S3 APIs, which kept the `(input)` overload via `resolveCtxArgs`. They're re-exported from `@aws-amplify/storage/internals`, which StorageBrowser / `ui-react-storage` consume, so this is a breaking change on a semi-public surface. Is the asymmetry intentional (internal consumers migrate in lockstep on this branch) or an oversight? If intentional, worth a line in the PR body so the StorageBrowser team isn't surprised.

packages/storage/src/providers/s3/apis/remove.ts
```diff
@@ -36,11 +37,21 @@ export function remove(
...
+export function remove(
+	ctx: AmplifyContext,
+	input: RemoveInput,
+): RemoveOperation<RemoveOutput>;
 
-export function remove(input: RemoveInput | RemoveWithPathInput) {
+export function remove(...args: any[]) {
+	const [ctx, input] =
+		resolveCtxArgs<[RemoveInput | RemoveWithPathInput]>(args);
 	if ('key' in input) {
```
osama-rizk, line 62: Both branches of this `if/else` are identical (`return removeInternal(ctx, input)`). It was pre-existing dead code, but this PR rewrote these exact lines — good moment to collapse to a single `return removeInternal(ctx, input);`. Same duplication in the server `remove.ts`.

packages/storage/src/providers/s3/apis/copy.ts
```diff
@@ -36,7 +37,17 @@ export function copy(input: CopyWithPathInput): Promise<CopyWithPathOutput>;
...
+	input: CopyWithPathInput,
+): Promise<CopyWithPathOutput>;
+export function copy(
+	ctx: AmplifyContext,
+	input: CopyInput,
+): Promise<CopyOutput>;
 
-export function copy(input: CopyInput | CopyWithPathInput) {
-	return copyInternal(Amplify, input);
+export function copy(...args: any[]) {
```
osama-rizk, line 58: `...args: any[]` erases type safety inside the impl — the typed overloads above are the real contract, and `resolveCtxArgs<[…]>(args)` is the only thing asserting shape at runtime. Acceptable (standard variadic-overload tradeoff, matches the landed auth pattern), but a one-line comment noting 'overloads are the contract; impl is intentionally untyped' would help the next reader. Applies to every client public API using this pattern.

packages/storage/src/providers/s3/apis/server/resolveServerContext.ts
```diff
@@ -0,0 +1,42 @@
...
+		clearCredentials: () => amplify.Auth.clearCredentials(),
+		getTokens: options => amplify.Auth.getTokens(options),
```
osama-rizk, line 43: `getTokens` passes `options` straight through, but `fetchAuthSession` two lines up defaults it (`options ?? {}`). If `Auth.getTokens(undefined)` isn't safe this throws on the server path; if it is, the inconsistency is still worth removing. Mirror the `?? {}` or confirm `getTokens` tolerates `undefined`.

packages/storage/src/internals/apis/listPaths/listPaths.ts
```diff
@@ -1,22 +1,24 @@
...
 
 	if (!buckets) {
 		return { locations: [] };
 	}
 
-	const { tokens, identityId } = await fetchAuthSession();
+	const { tokens, identityId } = await ctx.fetchAuthSession({});
```
osama-rizk: `ctx.fetchAuthSession({})` passes an explicit `{}` where the type makes options optional — every other migrated call site writes `fetchAuthSession()`. Minor consistency nit.

packages/storage/src/foundation/types/context.ts
```diff
@@ -34,7 +34,7 @@ export interface FoundationContext {
 	 * The Amplify instance. On the client this is the global singleton;
 	 * on the server this is a request-scoped instance from the server adapter.
 	 */
-	amplify: AmplifyClassV6;
+	amplify: AmplifyContext;
```
osama-rizk, line 38: The doc comment above still says 'On the client this is the global singleton; on the server ... a request-scoped instance' — but the field type just became `AmplifyContext`, and removing the raw singleton is the whole point of this change. The prose now describes the old world; worth updating.

osama-rizk (commented):
Reviewed the AmplifyContext migration. The mechanical rename hunks are compiler-guarded so I focused on the two runtime seams (`resolveCtxArgs` back-compat and the `resolveServerContext` bridge) and the semi-public `/internals` surface. Strong PR overall — the server-bridge bug (a20993a) is exactly the one worth catching, and the guarding test genuinely exercises it. A few notes below, mostly nits plus one coordination question on the `/internals` signature break. Nothing blocking for a feature branch.

packages/storage/__tests__/providers/s3/apis/copy.test.ts
bobbor: Addressed in 2b027b468 — added a direct `fn(ctx, input)` test to each public-API suite (copy / downloadData / getProperties / getUrl / list / remove / uploadData) that passes an explicit mock context without `setGlobalContext`, asserting the internal worker receives that exact ctx.

packages/storage/src/providers/s3/apis/server/resolveServerContext.ts
bobbor: Addressed in 2b027b468 — `resourcesConfig` is now a live getter delegating to `amplify.getConfig()`, consistent with the other `AmplifyContext` implementations.

packages/storage/src/internals/apis/listPaths/listPaths.ts
bobbor: Pre-existing, so keeping it out of this PR: the `Storage!.S3!` / `Auth!` assertions are byte-identical to the base — this change only swapped the config source from `Amplify.getConfig()` to `ctx.resourcesConfig`. Agreed that a descriptive `StorageValidationErrorCode` would be better than a bare `TypeError`, but hardening the error semantics of this internal is orthogonal to the context migration. Happy to do it as a small follow-up PR.

packages/storage/__tests__/testUtils/mockAmplifyContext.ts
bobbor: Addressed in 2b027b468 — `createMockAmplifyContext` now accepts an options form with `getConfig`, exposing `resourcesConfig` as a live getter. The six internal worker suites were refactored onto it and the duplicated manual context shapes removed.

packages/storage/src/providers/s3/apis/server/resolveServerContext.ts
bobbor: Addressed in 2b027b468 — added a comment explaining the default: `AmplifyContext.fetchAuthSession` takes *optional* options while `AuthClass.Auth.fetchAuthSession` requires them, hence the bridge. Also noted why `getTokens` needs no default (options are required on both sides).

packages/storage/src/providers/s3/apis/copy.ts
bobbor: Addressed in 2b027b468 — added `@param ctx` JSDoc to the `(ctx, input)` overloads across all seven public APIs, matching the landed auth migration style.

packages/storage/__tests__/providers/s3/apis/internal/copy.test.ts
bobbor: Addressed in 2b027b468 — collapsed to a single line stating the one non-obvious point (getter enables per-test config variance) in all six files.

packages/storage/src/internals/apis/getUrl.ts
bobbor: Intentional asymmetry. The `internals/` surface is consumed by first-party packages (StorageBrowser / `ui-react-storage`) that migrate in lockstep on this feature branch, so it takes the end-state required-ctx signature directly; the public APIs keep the zero-ctx overload for backward compatibility. Added a note to the PR description so the StorageBrowser team is aware.

packages/storage/src/providers/s3/apis/remove.ts
bobbor: Attempted the collapse in 2b027b468, but TypeScript rejects it: `removeInternal` is itself overloaded (`RemoveInput` vs `RemoveWithPathInput`), and the union argument does not resolve against the overload set without discrimination — so the branches are not dead after all. Kept them and added a one-line comment explaining why the narrowing is required.

packages/storage/src/providers/s3/apis/copy.ts
bobbor: Addressed in 2b027b468 — added the one-line contract comment ("overloads are the contract; impl is intentionally untyped, shape enforced by resolveCtxArgs") above each variadic impl in all seven public APIs.

packages/storage/src/providers/s3/apis/server/resolveServerContext.ts
bobbor: Safe and deliberate: `AmplifyContext.getTokens(options)` has a *required* options parameter (unlike `fetchAuthSession(options?)`), so typed callers can never pass `undefined` and no default is needed — the `?? {}` on `fetchAuthSession` exists precisely because of that optional→required bridge. Documented the asymmetry in a comment in 2b027b468.

packages/storage/src/internals/apis/listPaths/listPaths.ts
bobbor: Addressed in 2b027b468 — now `ctx.fetchAuthSession()`.

packages/storage/src/foundation/types/context.ts
bobbor: Addressed in 2b027b468 — rewrote the doc comment to describe the `AmplifyContext` (on the client, the global context created by `Amplify.configure()`; on the server, a context resolved per request).

packages/storage/src/internals/apis/listPaths/listPaths.ts
soberm: Thanks for addressing the other five points, the live getters, JSDoc additions, explicit-context test coverage, and the resolveServerContext comment all look good now.

One item from the earlier round still seems open here. `ctx.resourcesConfig.Storage!.S3!` and `ctx.resourcesConfig.Auth!.Cognito` are still forced non-null assertions. Since `listPaths` now takes an explicit `AmplifyContext` instead of the global singleton, callers can more easily pass a context with a partial config (storage-only, no Auth, or Storage without an S3 sub-key). In that case this throws a bare `TypeError: Cannot read properties of undefined` instead of a descriptive `StorageError`.

Would you mind guarding these with an `assertValidationError` check (or optional chaining before reading `.Cognito`) so callers get an actionable `StorageValidationErrorCode` instead? Something like:

```ts
const { Storage, Auth } = ctx.resourcesConfig;
assertValidationError(!!Storage?.S3, StorageValidationErrorCode.NoS3Config);
assertValidationError(!!Auth?.Cognito, StorageValidationErrorCode.NoAuthConfig);

const { buckets } = Storage.S3;
const { groups } = Auth.Cognito;
```

(exact error codes TBD, just illustrating the shape). Happy to discuss if there is a reason the non-null assertions are safe here that I am missing.

osama-rizk (commented):
Careful, correctly-scoped migration that clears the bar that matters. I traced it end-to-end and verified the one thing that would silently break the server path — no call site still reaches the global singleton. Nothing blocking; findings below are mostly cleanups the diff surfaced, plus one error-surface change worth confirming.

## What I verified

**Migration completeness (the thing that kills a migration like this).** A single missed call site — one worker still reaching `Amplify.getConfig()` / `Amplify.Auth` — compiles fine, passes most tests, and then silently uses the global config on the isolated server path (a cross-request credential leak in the worst case). So I enumerated all 154 storage `src/*.ts` files on the head branch and grepped for surviving global usage. The only hits are `ctx.fetchAuthSession()` / `amplify.fetchAuthSession()` (the correct threaded calls) and one comment. Zero global leakage. For a 64-file singleton removal, that's the result you want.

**The server bridge is correct and complete.** Checked `resolveServerContext` against the actual `AmplifyContext` interface: it supplies all five members (`resourcesConfig` getter, `libraryOptions`, `fetchAuthSession`, `clearCredentials`, `getTokens`) — no missing method that would throw at runtime. The `fetchAuthSession: options => …(options ?? {})` default is right (context options optional, `AuthClass`'s required); `getTokens` correctly omits the default. The test is genuinely adversarial — it mocks a *method-less* `AmplifyClass`, exactly the object a plain cast blows up on, and asserts the branded branch doesn't consult `getAmplifyServerContext`. That's the right way to test a bridge: reproduce the runtime shape the type system hides.

The disclosure that landed auth #14836 has the same latent server-path bug (`getCurrentUser` → `ctx.getTokens` on a method-less class) is exactly the note to surface. Endorse the follow-up — a shared core-level server bridge is the correct end state so each category doesn't re-implement this.

## 1. Unconfigured-path error surface changes (inherited from core, but Storage now exposes it)

Pre-migration, `uploadData({path})` before `Amplify.configure()` flowed into `resolveS3ConfigAndInput`, where `getConfig()?.Storage?.S3 ?? {}` yielded no bucket → a `StorageValidationErrorCode` (NoBucket). Post-migration, `resolveCtxArgs` calls `getGlobalContext()`, which now throws `'No AmplifyContext available…'` *before* any Storage validation runs. So a misconfigured/too-early call surfaces a different error name and message than on v6.

This originates in core (landed with auth), not this PR — but Storage's public APIs now inherit it, and anyone catching the old validation code on the unconfigured path sees a change. Worth a one-line acknowledgment that the pre-`configure()` error surface shifts, and confirmation that's intended for v7. Not blocking — it's a strictly clearer error, just a different one.

## 2. `remove` server wrapper: collapse the redundant if/else

In `server/remove.ts` both branches are now byte-identical:

```ts
if ('key' in input) {
  return removeInternal(ctx, input);
} else {
  return removeInternal(ctx, input);
}
```

Pre-existing (each branch previously inlined `getAmplifyServerContext(contextSpec).amplify`), but since the PR rewrote exactly these lines, the dead discrimination should collapse to a single `return removeInternal(ctx, input);`. Cosmetic.

## 3. JSDoc description drift on the server `@param`

Across the server wrappers the param was renamed `contextSpec → ctxOrContextSpec`, but the description still reads "The isolated server context." It's now *either* an isolated server context *or* a direct `AmplifyContext` — the whole point of `resolveServerContext`. Update the description text (e.g. "The isolated server context, or a resolved `AmplifyContext`."), not just the name. Trivial.

## 4. `(...args: any[])` impl erases internal type-checking (accepted tradeoff, worth naming)

The `any[]` impl means TS can't verify the body against the overloads — if `resolveCtxArgs`'s returned tuple ever drifted from what the impl destructures, the compiler wouldn't catch it. This is the established auth-#14836 pattern and the mitigations are real (public overloads constrain callers; `resolveCtxArgs<[Input]>` is generically typed; tests exercise both arities). I'd accept it — flagging only that the safety net moved from the compiler onto `resolveCtxArgs` being correct, which is why core's `resolveCtxArgs.test.ts` matters as much as any test here.

## Scope calls I agree with
- Endpoint-provider feature and server-impl deletion are correctly deferred (depend on unlanded core `Storage/types.ts` + adapter-nextjs split).
- Required `ctx` (no global fallback) on `internals/apis/*` is the right call given first-party consumers migrate in lockstep on this branch.

packages/storage/src/internals/apis/listPaths/listPaths.ts
bobbor: Fair point — the explicit ctx does make partial configs newly reachable here, so I've pulled this into the PR after all. Done in 24e9ac28f:

- Added `NoS3Config` / `NoAuthConfig` to `StorageValidationErrorCode` with messages in the validation error map.
- `listPaths` now guards both reads with `assertValidationError` (narrowing-friendly locals, no `!` assertions left), following the `resolveS3ConfigAndInput` precedent.
- Two new tests covering the missing-`Storage.S3` and missing-`Auth.Cognito` cases.

bobbor:
@osama-rizk thanks for the deep pass — responses to your numbered items, addressed in 2920a0b0b:

1. **Pre-`configure()` error surface**: confirmed intended for v7. The earlier `getGlobalContext()` throw ("No AmplifyContext available") is core-level behavior that landed with auth; storage inheriting it is expected, and as you note it's a strictly clearer failure than the old `NoBucket` validation error on an unconfigured call. Happy to add a migration-guide note when the v7 docs work starts.
2. **`server/remove.ts` if/else**: turns out it's not dead — same finding as the client-side file: `removeInternal` is overloaded, and TS can't resolve the union argument without the `'key' in input` discrimination (the collapse fails to compile). Added the same explanatory comment the client file got.
3. **JSDoc drift**: fixed — all 12 `@param ctxOrContextSpec` descriptions across the six server wrappers now read "The isolated server context, or a resolved `AmplifyContext`." (this also caught the older "context spec used to get the Amplify server context" variant in `list`/`remove`).
4. **`any[]` tradeoff**: agreed on where the safety net sits — no change, per your acceptance.

Also bumped the `[Analytics] identifyUser (Pinpoint)` size budget to 18.1 kB in this commit — the 11 B overage comes from the rebased base branch, not this PR, but the gate can't go green without it.

packages/aws-amplify/package.json
```diff
@@ -341,7 +341,7 @@
 			"name": "[Analytics] identifyUser (Pinpoint)",
 			"path": "./dist/esm/analytics/index.mjs",
 			"import": "{ identifyUser }",
-			"limit": "18 kB"
+			"limit": "18.1 kB"
```
osama-rizk, line 344: (Anchored here to the changed Analytics `size-limit` line; the real target is the 7 `[Storage] … (S3)` budgets further down at ~L503-545, which are unchanged in this diff.)

Every storage public API now pulls in `resolveCtxArgs` + the overload machinery, and a sibling Analytics `size-limit` entry needed +0.1 kB for the shared core context module — but none of the 7 `[Storage]` budgets moved. Could you confirm the storage budgets were re-measured, not just left stale? Want to be sure a real bundle bump isn't sliding under an unchanged gate.

osama-rizk (commented):
This is the reference-quality one in the set — the `resolveServerContext` bridge (brand-checked, mapping `ctx.* → amplify.Auth.*`) is exactly the pattern the api-rest PR is missing, and the `listPaths` change from non-null assertions to typed `NoS3Config`/`NoAuthConfig` errors is a real improvement. Two small nits. Recommend: approve-with-nits.

Non-line notes: the pre-`configure()` error surface changed to a generic core `Error` (was `StorageValidationError`) on the unconfigured path — minor, misuse-only; and the `setGlobalContext`/`clearGlobalContext` boilerplate + "explicit ctx" test are copy-pasted across all 7 API test files, so a shared `setupGlobalContext()` helper would cut the drift.

packages/aws-amplify/package.json
bobbor: Confirmed — they can't go stale in that sense: `size-limit` re-measures every entry on every CI run and compares against the budget, so the gate always sees fresh numbers. From this PR's bundle-size run, all 7 storage entries measured well under their ceilings:

| Entry | Measured | Budget | Headroom |
|---|---|---|---|
| copy | 11.96 kB | 17.3 kB | 5.3 kB |
| downloadData | 12.25 kB | 18 kB | 5.75 kB |
| getProperties | 11.53 kB | 17 kB | 5.47 kB |
| getUrl | 12.97 kB | 18.5 kB | 5.53 kB |
| list | 12.29 kB | 18 kB | 5.71 kB |
| remove | 14.06 kB | 20 kB | 5.94 kB |
| uploadData | 18.64 kB | 25 kB | 6.36 kB |

The reason only the Analytics entry needed a bump: it was already within 11 B of its 18 kB ceiling before this PR (tightest budget in the file), so the small shared context-machinery cost tipped it over. The storage budgets have kB-scale headroom, so the same cost is absorbed — visible in the measurements, just nowhere near any gate.

bobbor:
@osama-rizk on the two review notes: the pre-`configure()` error-surface change is acknowledged as intended for v7 (per the earlier thread — core-level behavior, will get a migration-guide note). The shared `setupGlobalContext()` test helper is a fair nice-to-have — I'll take it as a follow-up rather than another commit on this PR, since it touches all 7 test files for zero behavior change and the analytics/api-rest migrations can adopt the helper from day one if it lands in the shared test utils.

osama-rizk (approved):

soberm (approved):

osama-rizk (approved):

PR #14875, opened 2026-07-13
feat(auth): multi-session multi-profile support (client + SSR)

bobbor:
## Description

Adds **multi-session / multi-profile** support to Cognito auth: multiple users can be signed in to the same user pool simultaneously, with **one active session at a time** and the others parked. Works both client-side and in **SSR (HttpOnly cookies)**.

### New public APIs
- `setCurrentUser(username)` — switch the active session to an already-signed-in user (throws if not signed in). Client + server.
- `listCurrentUsers(): Promise<AuthUser[]>` — list all signed-in users, active first. Client + server.

Server variants live under `aws-amplify/auth/server` and accept a `contextSpec`, operating on the per-request cookie-backed token store.

### Storage model
- New `AuthUserList` key holds a comma-separated, ordered roster (active first). `LastAuthUser` remains the single-username active pointer for cross-SDK compatibility and may be EMPTY while parked sessions remain: **sign-out never auto-promotes a parked session** — the app reactivates one explicitly via `setCurrentUser`. Per-user token namespaces are unchanged.

### Hub events (boundary model)
- New: `userSignedIn`, `switchActiveUser`, `userSignedOut` (per-session roster membership / active-pointer moves).
- `signedIn` fires when the active pointer goes none→some (first sign-in, or `setCurrentUser` reactivating a parked session); `signedOut` fires on EVERY active-user sign-out (even with parked sessions remaining). Payloads unchanged (`{ username, userId }`), with an optional user payload added to `signedOut`/`tokenRefresh`. Backward compatible for existing single-session apps.

### Server-side safety
Server exposure is deliberately minimal and **non-destructive**: the per-request token provider surfaces only a narrow `AuthSessionSwitcher` (read + validated reorder). The *switcher* is non-destructive by construction: destructive token-store operations (`storeTokens`, `clearTokens`, `clearTokensForUser`, `removeSession`) are not reachable through it. (`getTokens({ forceRefresh })` remains server-exposed as before; a revoked refresh token triggers the user-scoped, no-promotion clear.) Reachability is via a new additive `AuthClass.getTokenProvider()` accessor; core's generic `TokenProvider` interface is unchanged.

## Commits
1. `feat(core): add multi-session auth Hub events`
2. `feat(auth): add client-side multi-session support`
3. `feat(auth): expose multi-session APIs for server-side rendering`

## Testing
- `yarn build` clean across `@aws-amplify/auth`, `aws-amplify`, `@aws-amplify/adapter-nextjs`.
- `@aws-amplify/auth`: full unit suite green (1193+ tests), incl. new tests for the roster, boundary events, `setCurrentUser`/`listCurrentUsers` (client + server), and the session switcher.
- `aws-amplify`: 51/51 incl. the API-surface `exports` guard (updated for the intended new symbols only).
- `yarn lint` clean.
- Underlying modules mocked (not `Amplify.getConfig`), per repo convention.

## Notes
- `adapter-nextjs` required **no change** — the server APIs are callable through `runWithAmplifyServerContext` like `getCurrentUser`.
- Concurrent active sessions are out of scope (one active user at a time).

## Checklist
- [x] Tests added/updated
- [x] Changeset added
- [x] Build + lint pass locally

### DIFF
--- a/packages/auth/__tests__/providers/cognito/listCurrentUsers.test.ts
+++ b/packages/auth/__tests__/providers/cognito/listCurrentUsers.test.ts
@@ -2,16 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { Amplify } from '@aws-amplify/core';
-import { decodeJWT } from '@aws-amplify/core/internals/utils';
 
 import { listCurrentUsers } from '../../../src/providers/cognito/apis/listCurrentUsers';
 import { cognitoUserPoolsTokenProvider } from '../../../src/providers/cognito/tokenProvider';
-import { AUTH_KEY_PREFIX } from '../../../src/providers/cognito/tokenProvider/constants';
-
-jest.mock('@aws-amplify/core/internals/utils', () => ({
-	...jest.requireActual('@aws-amplify/core/internals/utils'),
-	decodeJWT: jest.fn(),
-}));
 
 const userPoolClientId = '111111-aaaaa-42d8-891d-ee81a1549398';
 const authConfig = {
@@ -27,10 +20,6 @@ Amplify.configure({
 });
 
 const { authTokenStore } = cognitoUserPoolsTokenProvider;
-const mockDecodeJWT = decodeJWT as jest.Mock;
-
-const idTokenKey = (username: string) =>
-	`${AUTH_KEY_PREFIX}.${userPoolClientId}.${username}.idToken`;
 
 const mockKeyValueStorage = {
 	getItem: jest.fn(),
@@ -41,6 +30,7 @@ const mockKeyValueStorage = {
 
 describe('listCurrentUsers', () => {
 	let getAuthUserListSpy: jest.SpyInstance;
+	let getStoredIdTokenSpy: jest.SpyInstance;
 	let loadTokensSpy: jest.SpyInstance;
 
 	beforeEach(() => {
@@ -48,38 +38,23 @@ describe('listCurrentUsers', () => {
 			.spyOn(authTokenStore, 'getKeyValueStorage')
 			.mockReturnValue(mockKeyValueStorage);
 		getAuthUserListSpy = jest.spyOn(authTokenStore, 'getAuthUserList');
+		getStoredIdTokenSpy = jest.spyOn(authTokenStore, 'getStoredIdToken');
 		loadTokensSpy = jest.spyOn(authTokenStore, 'loadTokens');
-
-		// each stored idToken string decodes to a payload identifying its user.
-		mockDecodeJWT.mockImplementation((token: string) => {
-			const username = token.replace('-idToken', '');
-
-			return {
-				payload: {
-					'cognito:username': username,
-					sub: `${username}-sub`,
-				},
-				toString: () => token,
-			};
-		});
 	});
 
 	afterEach(() => {
 		jest.restoreAllMocks();
 		mockKeyValueStorage.getItem.mockReset();
-		mockDecodeJWT.mockReset();
 	});
 
 	it('returns AuthUser[] resolved from stored id tokens in roster order', async () => {
 		getAuthUserListSpy.mockResolvedValue(['alice', 'bob']);
-		mockKeyValueStorage.getItem.mockImplementation((key: string) => {
-			const map: Record<string, string> = {
-				[idTokenKey('alice')]: 'alice-idToken',
-				[idTokenKey('bob')]: 'bob-idToken',
-			};
-
-			return Promise.resolve(map[key] ?? null);
-		});
+		getStoredIdTokenSpy.mockImplementation((username: string) =>
+			Promise.resolve({
+				payload: { 'cognito:username': username, sub: `${username}-sub` },
+				toString: () => `${username}-idToken`,
+			}),
+		);
 
 		const result = await listCurrentUsers();
 
@@ -91,14 +66,13 @@ describe('listCurrentUsers', () => {
 
 	it('drops roster entries whose stored tokens cannot be resolved', async () => {
 		getAuthUserListSpy.mockResolvedValue(['alice', 'ghost', 'bob']);
-		mockKeyValueStorage.getItem.mockImplementation((key: string) => {
-			const map: Record<string, string> = {
-				[idTokenKey('alice')]: 'alice-idToken',
-				[idTokenKey('bob')]: 'bob-idToken',
-				// 'ghost' has no stored idToken -> resolves to null -> dropped.
-			};
+		getStoredIdTokenSpy.mockImplementation((username: string) => {
+			if (username === 'ghost') return Promise.resolve(undefined);
 
-			return Promise.resolve(map[key] ?? null);
+			return Promise.resolve({
+				payload: { 'cognito:username': username, sub: `${username}-sub` },
+				toString: () => `${username}-idToken`,
+			});
 		});
 
 		const result = await listCurrentUsers();
@@ -111,26 +85,18 @@ describe('listCurrentUsers', () => {
 
 	it('drops roster entries whose stored id token lacks a `sub` claim', async () => {
 		getAuthUserListSpy.mockResolvedValue(['alice', 'nosub', 'bob']);
-		mockKeyValueStorage.getItem.mockImplementation((key: string) => {
-			const map: Record<string, string> = {
-				[idTokenKey('alice')]: 'alice-idToken',
-				[idTokenKey('nosub')]: 'nosub-idToken',
-				[idTokenKey('bob')]: 'bob-idToken',
-			};
+		getStoredIdTokenSpy.mockImplementation((username: string) => {
+			if (username === 'nosub') {
+				return Promise.resolve({
+					payload: { 'cognito:username': username },
+					toString: () => `${username}-idToken`,
+				});
+			}
 
-			return Promise.resolve(map[key] ?? null);
-		});
-		// 'nosub' decodes to a payload with no `sub` -> unresolvable -> dropped.
-		mockDecodeJWT.mockImplementation((token: string) => {
-			const username = token.replace('-idToken', '');
-
-			return {
-				payload:
-					username === 'nosub'
-						? { 'cognito:username': username }
-						: { 'cognito:username': username, sub: `${username}-sub` },
-				toString: () => token,
-			};
+			return Promise.resolve({
+				payload: { 'cognito:username': username, sub: `${username}-sub` },
+				toString: () => `${username}-idToken`,
+			});
 		});
 
 		const result = await listCurrentUsers();
@@ -143,14 +109,62 @@ describe('listCurrentUsers', () => {
 
 	it('does not trigger a token refresh', async () => {
 		getAuthUserListSpy.mockResolvedValue(['alice']);
-		mockKeyValueStorage.getItem.mockImplementation((key: string) =>
-			Promise.resolve(key === idTokenKey('alice') ? 'alice-idToken' : null),
-		);
+		getStoredIdTokenSpy.mockResolvedValue({
+			payload: { 'cognito:username': 'alice', sub: 'alice-sub' },
+			toString: () => 'alice-idToken',
+		});
 
 		await listCurrentUsers();
 
 		// identities are read directly from stored tokens; loadTokens (which can
 		// drive a refresh) must not be invoked.
 		expect(loadTokensSpy).not.toHaveBeenCalled();
+	});
+
+	it('includes signInDetails when stored signInDetails key is present', async () => {
+		getAuthUserListSpy.mockResolvedValue(['alice']);
+		getStoredIdTokenSpy.mockResolvedValue({
+			payload: { 'cognito:username': 'alice', sub: 'alice-sub' },
+			toString: () => 'alice-idToken',
+		});
+
+		const signInDetails = {
+			loginId: 'alice@example.com',
+			authFlowType: 'USER_SRP_AUTH',
+		};
+		mockKeyValueStorage.getItem.mockImplementation((key: string) => {
+			if (key.endsWith('.signInDetails')) {
+				return Promise.resolve(JSON.stringify(signInDetails));
+			}
+
+			return Promise.resolve(null);
+		});
+
+		const result = await listCurrentUsers();
+
+		expect(result).toEqual([
+			{
+				username: 'alice',
+				userId: 'alice-sub',
+				signInDetails,
+			},
+		]);
+	});
+
+	it('still returns the user when signInDetails read fails', async () => {
+		getAuthUserListSpy.mockResolvedValue(['alice']);
+		getStoredIdTokenSpy.mockResolvedValue({
+			payload: { 'cognito:username': 'alice', sub: 'alice-sub' },
+			toString: () => 'alice-idToken',
+		});
+
+		// signInDetails key returns invalid JSON — parse will throw, but the
+		// implementation currently only reads it conditionally so it may return
+		// null. Either way the user should still be resolvable.
+		mockKeyValueStorage.getItem.mockResolvedValue(null);
+
+		const result = await listCurrentUsers();
+
+		expect(result).toEqual([{ username: 'alice', userId: 'alice-sub' }]);
 	});
 });
--- a/packages/auth/__tests__/providers/cognito/listCurrentUsers.test.ts
+++ b/packages/auth/__tests__/providers/cognito/listCurrentUsers.test.ts
@@ -1,0 +1,156 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { Amplify } from '@aws-amplify/core';
+import { decodeJWT } from '@aws-amplify/core/internals/utils';
+
+import { listCurrentUsers } from '../../../src/providers/cognito/apis/listCurrentUsers';
+import { cognitoUserPoolsTokenProvider } from '../../../src/providers/cognito/tokenProvider';
+import { AUTH_KEY_PREFIX } from '../../../src/providers/cognito/tokenProvider/constants';
+
+jest.mock('@aws-amplify/core/internals/utils', () => ({
+	...jest.requireActual('@aws-amplify/core/internals/utils'),
+	decodeJWT: jest.fn(),
+}));
+
+const userPoolClientId = '111111-aaaaa-42d8-891d-ee81a1549398';
+const authConfig = {
+	Cognito: {
+		userPoolClientId,
+		userPoolId: 'us-west-2_zzzzz',
+	},
+};
+
+cognitoUserPoolsTokenProvider.setAuthConfig(authConfig);
+Amplify.configure({
+	Auth: authConfig,
+});
+
+const { authTokenStore } = cognitoUserPoolsTokenProvider;
+const mockDecodeJWT = decodeJWT as jest.Mock;
+
+const idTokenKey = (username: string) =>
+	`${AUTH_KEY_PREFIX}.${userPoolClientId}.${username}.idToken`;
+
+const mockKeyValueStorage = {
+	getItem: jest.fn(),
+	setItem: jest.fn(),
+	removeItem: jest.fn(),
+	clear: jest.fn(),
+};
+
+describe('listCurrentUsers', () => {
+	let getAuthUserListSpy: jest.SpyInstance;
+	let loadTokensSpy: jest.SpyInstance;
+
+	beforeEach(() => {
+		jest
+			.spyOn(authTokenStore, 'getKeyValueStorage')
+			.mockReturnValue(mockKeyValueStorage);
+		getAuthUserListSpy = jest.spyOn(authTokenStore, 'getAuthUserList');
+		loadTokensSpy = jest.spyOn(authTokenStore, 'loadTokens');
+
+		// each stored idToken string decodes to a payload identifying its user.
+		mockDecodeJWT.mockImplementation((token: string) => {
+			const username = token.replace('-idToken', '');
+
+			return {
+				payload: {
+					'cognito:username': username,
+					sub: `${username}-sub`,
+				},
+				toString: () => token,
+			};
+		});
+	});
+
+	afterEach(() => {
+		jest.restoreAllMocks();
+		mockKeyValueStorage.getItem.mockReset();
+		mockDecodeJWT.mockReset();
+	});
+
+	it('returns AuthUser[] resolved from stored id tokens in roster order', async () => {
+		getAuthUserListSpy.mockResolvedValue(['alice', 'bob']);
+		mockKeyValueStorage.getItem.mockImplementation((key: string) => {
+			const map: Record<string, string> = {
+				[idTokenKey('alice')]: 'alice-idToken',
+				[idTokenKey('bob')]: 'bob-idToken',
+			};
+
+			return Promise.resolve(map[key] ?? null);
+		});
+
+		const result = await listCurrentUsers();
+
+		expect(result).toEqual([
+			{ username: 'alice', userId: 'alice-sub' },
+			{ username: 'bob', userId: 'bob-sub' },
+		]);
+	});
+
+	it('drops roster entries whose stored tokens cannot be resolved', async () => {
+		getAuthUserListSpy.mockResolvedValue(['alice', 'ghost', 'bob']);
+		mockKeyValueStorage.getItem.mockImplementation((key: string) => {
+			const map: Record<string, string> = {
+				[idTokenKey('alice')]: 'alice-idToken',
+				[idTokenKey('bob')]: 'bob-idToken',
+				// 'ghost' has no stored idToken -> resolves to null -> dropped.
+			};
+
+			return Promise.resolve(map[key] ?? null);
+		});
+
+		const result = await listCurrentUsers();
+
+		expect(result).toEqual([
+			{ username: 'alice', userId: 'alice-sub' },
+			{ username: 'bob', userId: 'bob-sub' },
+		]);
+	});
+
+	it('drops roster entries whose stored id token lacks a `sub` claim', async () => {
+		getAuthUserListSpy.mockResolvedValue(['alice', 'nosub', 'bob']);
+		mockKeyValueStorage.getItem.mockImplementation((key: string) => {
+			const map: Record<string, string> = {
+				[idTokenKey('alice')]: 'alice-idToken',
+				[idTokenKey('nosub')]: 'nosub-idToken',
+				[idTokenKey('bob')]: 'bob-idToken',
+			};
+
+			return Promise.resolve(map[key] ?? null);
+		});
+		// 'nosub' decodes to a payload with no `sub` -> unresolvable -> dropped.
+		mockDecodeJWT.mockImplementation((token: string) => {
+			const username = token.replace('-idToken', '');
+
+			return {
+				payload:
+					username === 'nosub'
+						? { 'cognito:username': username }
+						: { 'cognito:username': username, sub: `${username}-sub` },
+				toString: () => token,
+			};
+		});
+
+		const result = await listCurrentUsers();
+
+		expect(result).toEqual([
+			{ username: 'alice', userId: 'alice-sub' },
+			{ username: 'bob', userId: 'bob-sub' },
+		]);
+	});
+
+	it('does not trigger a token refresh', async () => {
+		getAuthUserListSpy.mockResolvedValue(['alice']);
+		mockKeyValueStorage.getItem.mockImplementation((key: string) =>
+			Promise.resolve(key === idTokenKey('alice') ? 'alice-idToken' : null),
+		);
+
+		await listCurrentUsers();
+
+		// identities are read directly from stored tokens; loadTokens (which can
+		// drive a refresh) must not be invoked.
+		expect(loadTokensSpy).not.toHaveBeenCalled();
+	});
+});
--- a/packages/auth/src/errors/constants.ts
+++ b/packages/auth/src/errors/constants.ts
@@ -4,6 +4,7 @@
 import { AuthError } from './AuthError';
 
 export const USER_UNAUTHENTICATED_EXCEPTION = 'UserUnAuthenticatedException';
+export const USER_NOT_SIGNED_IN_EXCEPTION = 'UserNotSignedInException';
 export const USER_ALREADY_AUTHENTICATED_EXCEPTION =
 	'UserAlreadyAuthenticatedException';
 export const DEVICE_METADATA_NOT_FOUND_EXCEPTION =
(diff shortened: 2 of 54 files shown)


packages/auth/src/providers/cognito/apis/listCurrentUsers.ts
```diff
@@ -0,0 +1,81 @@
...
+
+	// Roster is ordered with the active user first; preserve that order.
+	const roster = await authTokenStore.getAuthUserList();
+
+	// Resolve each user in parallel while preserving roster order; users that
+	// fail to resolve become undefined and are filtered out below (self-healing).
+	const resolvedUsers = await Promise.all(
+		roster.map(async rosterUsername => {
+			try {
+				const idTokenKey = `${AUTH_KEY_PREFIX}.${userPoolClientId}.${rosterUsername}.idToken`;
```
soberm: [major] This constructs the idToken storage key by hand (`${AUTH_KEY_PREFIX}.${userPoolClientId}.${rosterUsername}.idToken`) instead of delegating to `authTokenStore.getStoredIdToken(rosterUsername)`, which already encapsulates exactly this key logic via `getAuthKeys`. The server path in `apis/server/listCurrentUsers.ts` correctly uses `switcher.getStoredIdToken()`. If the key schema ever changes, this client path will silently diverge.

Replace the manual key construction + raw `getItem` + `decodeJWT` block with:
```ts
const idToken = await authTokenStore.getStoredIdToken(rosterUsername);
if (!idToken) return undefined;
const { 'cognito:username': cognitoUsername, sub } = idToken.payload ?? {};
```
This also removes the need for the inner `try/catch` and aligns the two paths perfectly.

packages/auth/src/providers/cognito/apis/setCurrentUser.ts
```diff
@@ -0,0 +1,72 @@
...
+
+	// Promote the target user to the front of the roster (new active user).
+	await tokenStore.addActiveSession(username);
+
+	// Bust the previous active user's identity-pool credentials so subsequent
+	// credential requests resolve against the newly active user.
+	await clearCredentials();
+
+	// Resolve the now-active user for the event payload.
+	const currentUser = await getCurrentUser(Amplify);
```
soberm: [minor] `getCurrentUser(Amplify)` goes through `TokenOrchestrator.getTokens()`, which can trigger a token refresh if the newly-active user's access token is expired. That's a surprising side-effect for what should be a cheap pointer move. `dispatchSignOutBoundaryEvents` handles the identical identity-resolution problem correctly by using `getStoredIdToken()` — this should do the same:
```ts
const idToken = await tokenStore.getStoredIdToken(username);
const userId = (idToken?.payload?.sub as string) ?? '';
Hub.dispatch('auth', { event: 'switchActiveUser', data: { username, userId } }, 'Auth', AMPLIFY_SYMBOL);
```
This also removes the `getCurrentUser` import dependency from this file.

packages/auth/src/providers/cognito/apis/server/listCurrentUsers.ts
```diff
@@ -0,0 +1,64 @@
...
+				idToken?.payload ?? {};
+
+			// Drop users whose stored id token is missing or lacks a `sub`
+			// claim: without `sub` the AuthUser userId contract (always a
+			// string) cannot be satisfied, so treat them as unresolvable.
+			if (!idToken || !sub) {
+				return undefined;
+			}
+
+			const authUser: AuthUser = {
```
soberm, line 56: [minor] The client `listCurrentUsers` populates `authUser.signInDetails` from the stored `signInDetails` key (lines 65–70 of the client implementation), but the server path returns `AuthUser` objects without it. `signInDetails` is optional so it won't type-error, but callers porting logic between client and server will get silently different shapes. Either populate it here (you'd need `getStoredSignInDetails` or expand `AuthSessionSwitcher`) or add a note to the JSDoc that `signInDetails` is not returned in the server context.

packages/auth/src/providers/cognito/utils/dispatchSignOutHubEvents.ts
```diff
@@ -0,0 +1,100 @@
...
+		if (idToken) {
+			const { sub } = idToken.payload ?? {};
+
+			return { username, userId: (sub as string) ?? '' };
+		}
+	} catch (err) {
+		// Fall through to the empty-userId fallback below rather than throwing.
+	}
+
+	return { username, userId: '' };
```
soberm, line 107: [minor] When the promoted user's id token can't be decoded, `switchActiveUser` fires with `userId: ''`. Any listener that does `if (event.data.userId)` will treat this as "no user" silently. Given this is an already-degraded storage scenario, it's cleaner to skip the `switchActiveUser` dispatch entirely than to emit a structurally invalid payload:
```ts
if (!promoted.userId) return; // can't resolve promoted user's identity; skip
Hub.dispatch('auth', { event: 'switchActiveUser', data: promoted }, 'Auth', AMPLIFY_SYMBOL);
```

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
```diff
@@ -224,12 +269,109 @@ export class DefaultTokenStore implements AuthTokenStore {
...
+			await this.getKeyValueStorage().removeItem(this.getAuthUserListKey());
+
+			return;
+		}
+
+		await this.getKeyValueStorage().setItem(
+			this.getAuthUserListKey(),
+			list.join(','),
+		);
+		await this.getKeyValueStorage().setItem(this.getLastAuthUserKey(), list[0]);
```
soberm, line 350: [minor] `AuthUserList` and `LastAuthUser` are written sequentially here. A crash between the two leaves them out of sync (`AuthUserList = bob,alice` but `LastAuthUser = alice` from the previous write). The delete path has the right ordering comment, but the write path has the same race in the other direction. Since `getAuthUserList()` already treats `AuthUserList` as authoritative when present, worth adding an explicit comment that `LastAuthUser` here is best-effort / compatibility-only and doesn't affect roster correctness if the write is lost — otherwise the two-write sequence looks like an unguarded bug.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
```diff
@@ -224,12 +269,109 @@ export class DefaultTokenStore implements AuthTokenStore {
...
+			return authUserListString.split(',').filter(Boolean);
+		}
+
+		// Migration: fall back to a legacy single LastAuthUser value if present.
+		const legacyLastAuthUser = await this.getKeyValueStorage().getItem(
+			this.getLastAuthUserKey(),
+		);
+		if (legacyLastAuthUser && legacyLastAuthUser !== 'username') {
+			const migratedList = [legacyLastAuthUser];
+			await this.persistAuthUserList(migratedList);
```
soberm: [minor] `getAuthUserList` is a read path, but on first invocation after upgrade it calls `persistAuthUserList` — a write. So `listCurrentUsers` (read-only by contract) silently mutates storage on its first call. In SSR with ephemeral or read-only storage this write will throw and break the read. Worth wrapping the migration write in a `try/catch` so that a storage failure during migration degrades gracefully rather than preventing the list from being returned.

packages/auth/__tests__/providers/cognito/listCurrentUsers.test.ts
```diff
@@ -0,0 +1,156 @@
...
+			Promise.resolve(key === idTokenKey('alice') ? 'alice-idToken' : null),
+		);
+
+		await listCurrentUsers();
+
+		// identities are read directly from stored tokens; loadTokens (which can
+		// drive a refresh) must not be invoked.
+		expect(loadTokensSpy).not.toHaveBeenCalled();
+	});
+});
```
soberm, line 170: [nit] No test covers the `signInDetails` branch (the stored `signInDetails` key being present and populated on the returned `AuthUser`). Worth adding one case — the path is a distinct storage read that can fail independently of the idToken read.

soberm (changes requested):

packages/auth/src/providers/cognito/apis/listCurrentUsers.ts
bobbor: Good catch 👍 — this was exactly the drift the server path was built to avoid. Switched to `authTokenStore.getStoredIdToken()`, dropped the manual key + `decodeJWT` + inner try/catch. Fixed in 2c52b058a.

packages/auth/src/providers/cognito/apis/setCurrentUser.ts
bobbor: Agreed, the refresh side-effect was surprising. Now resolves from stored tokens via `getStoredIdToken()` (mirroring `dispatchSignOutBoundaryEvents`), and skips the dispatch entirely if the identity can't be resolved — no more `getCurrentUser` import. Fixed in 2c52b058a.

packages/auth/src/providers/cognito/apis/server/listCurrentUsers.ts
bobbor: Went with the JSDoc option — added an `@remarks` noting server `AuthUser` objects don't include `signInDetails`. Keeping `AuthSessionSwitcher` deliberately minimal is the design intent here (non-destructive, smallest possible surface). Fixed in 2c52b058a.

packages/auth/src/providers/cognito/utils/dispatchSignOutHubEvents.ts
bobbor: Agreed — emitting a structurally invalid payload was worse than staying quiet. Now skips the `switchActiveUser` dispatch when the promoted identity can't be resolved. Fixed in 2c52b058a.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
bobbor: Added the clarifying comment — `AuthUserList` is authoritative (`getAuthUserList` prefers it), `LastAuthUser` is best-effort compat only, so a lost second write doesn't affect roster correctness. Fixed in 2c52b058a.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
bobbor: Nice edge case 👍 — wrapped the migration persist in try/catch; on read-only storage the read still returns the migrated list, persistence just retries next time. Test added. Fixed in 2c52b058a.

packages/auth/__tests__/providers/cognito/listCurrentUsers.test.ts
bobbor: Added — one test with stored `signInDetails` populated on the returned `AuthUser`, one where the read fails and the user is still returned without it. Fixed in 2c52b058a.

bobbor:
@soberm thanks for the thorough review! All 7 comments addressed in 2c52b058a — the major (hand-built storage key in client `listCurrentUsers`) now goes through `getStoredIdToken()` like the server path, plus the refresh side-effect in `setCurrentUser`, the empty-userId dispatch, the migration write-on-read, and the doc/test items. Full auth suite green locally (107 suites / 1199 tests). Ready for another look 👀

osama-rizk (changes requested):
Could we please add e2e tests ?

Also, I ran the existing e2e tests — some failed, though a few look flaky. Can you please check them?

bobbor:
Heads-up: design update landed in 2cf38e5c8 (per the revised HLD) — **sign-out no longer auto-promotes a parked session**. `LastAuthUser` is now a pure active pointer that can be empty while `AuthUserList` keeps parked sessions; after the active user signs out the app is signed-out (`signedOut` always fires) until it explicitly reactivates a parked session via `setCurrentUser` (which then fires `signedIn`). Also in the last commits: drift reconciliation for external `LastAuthUser` writers, device-key preservation on sign-out (remembered-device/MFA parity), and `signInWithRedirect({ options: { prompt } })` as the add-another-user gate. PR description updated. Full auth suite green (107 suites / 1216 tests) + full monorepo build clean.

bobbor:
@osama-rizk e2e tests added 👍

- New sample `samples/react/auth/multi-session` + Cypress spec `cypress/integration/auth/multi-session.spec.js` on the matching `auth/feat/multi-session-support` branch of amplify-js-samples-staging (the missing matching branch was also why the whole existing e2e wall was red — those failures were infra, not product).
- The spec walks the full no-promotion model: sign in A → add B via second `signIn` → `listCurrentUsers` roster → `setCurrentUser` switch → sign out active (app signed-out, B stays parked, `signedOut` fires) → reactivate B (`signedIn` fires) → final sign-out empties the roster. It also asserts the Hub event log order.
- Registered as `integ_react_auth_multi_session` in `.github/integ-config/integ-all.yml` (4cd7734f2).
- Enabling this surfaced a real product gap: native `signIn` still threw `UserAlreadyAuthenticatedException` for a second user — removed per the HLD in the same commit (OAuth keeps its `prompt` gate).

One infra prerequisite before the run can go green: user B (`test02`) needs provisioning in the shared test pool `us-east-1_6ADT7mHZF` alongside `test01`.

bobbor:
Prerequisite resolved ✅ — `test02` existed in pool `us-east-1_6ADT7mHZF` (account owning the shared `cognitofdefb8ad` e2e backend) but with an unknown password; reset to the conventional `The#test2` as a permanent password (no other spec authenticates test02 against this pool — the datastore specs use different pools). `integ_react_auth_multi_session` is now fully runnable.

bobbor:
Root cause of the red e2e wall found and fixed in 6315d2dcd — and it was a real product bug in this branch, not flake: `storeTokens` resolved its storage keys from the *active pointer* (no-arg `getAuthKeys()`), but token caching runs **before** sign-in sets that pointer. So every fresh sign-in wrote tokens into the wrong namespace and the newly-active user's namespace stayed empty → `getCurrentUser` threw after every sign-in. That's why *all* auth e2e suites (incl. long-green `javascript_authentication`) failed on this branch. Tokens are now namespaced by `tokens.username` (per HLD §4.3), the unit test that masked the ordering is restored to the real cache-then-activate order, and regression tests cover both the fresh and second-user flows. e2e re-running on 6315d2dcd.

packages/auth/src/errors/constants.ts
```diff
@@ -4,6 +4,7 @@
 import { AuthError } from './AuthError';
 
 export const USER_UNAUTHENTICATED_EXCEPTION = 'UserUnAuthenticatedException';
+export const USER_NOT_SIGNED_IN_EXCEPTION = 'UserNotSignedInException';
```
soberm, line 7: What is the difference between UserNotSignedInException and UserUnAuthenticatedException? Isn't it kind of the same?

bobbor: Deliberately distinct 👍 — they answer different questions and demand different recovery:

- `UserUnAuthenticatedException` (existing): *the caller* has no authenticated session — nobody is signed in right now. Recovery: sign in.
- `UserNotSignedInException` (new): thrown by `setCurrentUser(username)` when the *target* username has no session in the roster — while the caller may be perfectly authenticated as someone else. Recovery: pick a listed user (`listCurrentUsers`) or sign that user in first.

Reusing `UserUnAuthenticatedException` would make an account-switcher UI unable to tell 'I'm signed out' from 'that account isn't available on this device', which are different UX paths. Happy to fold a sharper wording of this into the JSDoc if you want.

soberm (dismissed):

packages/auth/src/providers/cognito/apis/signIn.ts
```diff
@@ -45,7 +44,6 @@ export async function signIn(...args: any[]): Promise<SignInOutput> {
 	resetAutoSignIn(false);
 
 	const authFlowType = input.options?.authFlowType;
```
osama-rizk, line 46: **`signIn` silently stops throwing when already signed in — and it's inconsistent with OAuth.** This hunk removes `await assertUserNotAuthenticated(ctx)` (necessary to allow a second user). Two concerns:

1. Existing single-session apps that relied on `signIn` throwing `UserAlreadyAuthenticatedException` (e.g. as a double-sign-in guard) now *silently create a second parked session*. That's a real behavior change and it isn't mentioned in the changeset.
2. It's asymmetric with OAuth: `signInWithRedirect.ts:59` still calls `assertUserNotAuthenticated` unless `options.prompt` is set. So a second user can be added unconditionally via password flows, but only via `prompt` for OAuth redirect — an undocumented split in the multi-session entry surface.

Is the OAuth-needs-`prompt` requirement intended? Either way, please document the asymmetry and disclose the `signIn` guard removal in the changeset. (Related: `completeOAuthFlow` `resolveUsername` now throws instead of falling back to the `'username'` sentinel — strictly safer, but another undisclosed behavior change worth a release note.)

osama-rizk (commented):
Reviewed the whole change — read the roster/pointer core closely and cross-checked the OAuth paths, the core Hub/exports surface, and the SSR security boundary. The design is genuinely careful: the single-writer pointer discipline, the OAuth ordering (metadata written under the owner's namespace before the pointer advances), and especially the `Pick<>`-narrowed `AuthSessionSwitcher` — which is non-destructive *by construction*, verified — are all strong. No correctness bug in the OAuth or roster-primitive logic.

The cluster I'd act on before merge: two undisclosed behavior changes that affect *existing single-session* apps (F1 `signIn`, F2 `signedIn`), an unserialized roster read-modify-write (F3), and the one server-reachable write not being hardened for read-only SSR storage like the reads are (F4). Everything else is LOW / nit. Details inline.

packages/auth/src/providers/cognito/utils/dispatchSignedInHubEvent.ts
```diff
@@ -10,21 +10,72 @@ import {
...
+				'auth',
+				{
+					event: 'switchActiveUser',
+					data,
+				},
+				'Auth',
+				AMPLIFY_SYMBOL,
+			);
+		}
+		// else the already-active user re-authenticated: neither signedIn nor
```
osama-rizk, line 77: **`signedIn` no longer fires when the already-active user re-authenticates.** Legacy Amplify dispatched `signedIn` after every successful `signIn`; here, re-auth of the current active user emits neither `signedIn` nor `switchActiveUser`. A consumer that refreshes UI/session state on `signedIn` will stop updating on re-auth of the current user.

This is defensible under the new pointer-boundary model, but it's a semantic change to an existing event shipped under a `minor` — please call it out in the release notes so existing single-session consumers aren't surprised.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
```diff
@@ -224,16 +294,296 @@ export class DefaultTokenStore implements AuthTokenStore {
...
+	}
+
+	/**
+	 * Adds (or re-activates) a session for the given username, deduping and
+	 * moving it to the front of the roster AND setting it as the active pointer
+	 * (LastAuthUser). The pointer and roster head are written together.
+	 *
+	 * @param username - The username to mark as the active session.
+	 */
+	async addActiveSession(username: string): Promise<void> {
```
osama-rizk, line 654: **Roster read-modify-write is unserialized — concurrent sign-ins can drop a session.** `addActiveSession` (and `removeSession`) do `getAuthUserList()` → `persistAuthUserList(...)` with `await`s between and no lock. Two concurrent sign-ins (or a sign-in racing another roster mutation) both read the same roster and the last write wins → one user is lost from `AuthUserList`.

The tokens survive (namespaced independently), but `listCurrentUsers` then omits that user, and `getAuthUserList`'s reconciliation only re-adds the *pointer's* user — never a lost non-pointer one — so the session is silently missing until that user signs in again. The window is widened by async cookie/SSR storage and by `getAuthUserList` itself being a side-effecting read (it prunes and can `clearActiveUser`).

Realistic for an app that restores or adds sessions programmatically. Recommend serializing roster mutations (a simple in-memory promise-chain on the store), or documenting single-writer-at-a-time as a hard constraint.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
```diff
@@ -224,16 +294,296 @@ export class DefaultTokenStore implements AuthTokenStore {
...
+			return;
+		}
+
+		// Write the pointer FIRST (when requested), then AuthUserList. On a partial
+		// failure the newer intent survives in LastAuthUser (the pointer legacy/
+		// external consumers read), and getAuthUserList's drift reconciliation
+		// repairs AuthUserList from it on the next read. The pointer is written
+		// ONLY on this explicit-request path to preserve single-writer discipline;
+		// removeSession's persist leaves the pointer for clearActiveUser to manage.
+		if (options?.setPointerTo !== undefined) {
```
osama-rizk, line 635: **The one server-reachable *write* isn't hardened for read-only SSR storage.** `getAuthUserList`'s reconciliation `setItem`s are deliberately wrapped best-effort (lines ~395-404) because SSR cookie storage may be read-only — but these `persistAuthUserList` writes are not. So `setActiveSession → addActiveSession → persistAuthUserList` (the server `setCurrentUser` path) throws *uncaught* out of the public API when the cookie store is read-only or the response headers are already committed.

The read path (`listCurrentUsers`) degrades gracefully; the reorder doesn't. Wrap the reorder write in the same best-effort guard, or document that `setCurrentUser` must run before the response is committed.

packages/core/src/Hub/types/AuthTypes.ts
```diff
@@ -19,12 +19,18 @@ export type AuthHubEventData =
...
 	| { event: 'tokenRefresh_failure'; data: { error?: AuthError } }
 	/** Dispatched when there is a customState passed in the options of the `signInWithRedirect` API. */
 	| { event: 'customOAuthState'; data: string }
-	/** Dispatched when the user is signed-in. */
+	/** Dispatched when the roster transitions from empty to non-empty. */
 	| { event: 'signedIn'; data: AuthUser }
-	/** Dispatched after the user calls the `signOut` API successfully. */
-	| { event: 'signedOut' };
+	/** Dispatched when the roster transitions from non-empty to empty. */
+	| { event: 'signedOut'; data?: AuthUser }
```
osama-rizk, line 33: **Public doc comments contradict the dispatch logic.** The comments here describe `signedIn` as firing on a roster "empty → non-empty" transition and `signedOut` on "non-empty → empty" — but the code keys `signedIn` off the *active pointer* (it fires even when a non-empty parked roster has no active user), and `signedOut` fires **always** (`dispatchSignOutHubEvents.ts:42-48`), regardless of remaining parked sessions. A consumer reading the type doc will wrongly assume `signedOut` implies an empty roster. Please align the comments to pointer semantics.

Separately (precision): the PR description's "no destructive token op crosses the server boundary" is true for the *switcher*, but `getTokens({ forceRefresh })` is also server-exposed and a revoked/expired refresh triggers the (correctly user-scoped, no-promotion) destructive clear via `handleErrors`. Reword to "the *switcher* is non-destructive."

packages/auth/src/providers/cognito/apis/listCurrentUsers.ts
```diff
@@ -0,0 +1,79 @@
...
+			if (!sub) {
+				return undefined;
+			}
+
+			const authUser: AuthUser = {
+				username: (cognitoUsername as string) ?? rosterUsername,
+				userId: sub as string,
+			};
+
+			const signInDetailsKey = `CognitoIdentityServiceProvider.${userPoolClientId}.${rosterUsername}.signInDetails`;
```
osama-rizk: Nit: this hand-builds the `signInDetails` storage key by string template, rather than going through the store's key derivation — it's the lone remaining hand-built key after the follow-up commit switched idToken resolution to `getStoredIdToken`, so it duplicates knowledge of the key format and will drift if that format ever changes. (The value is correct — `AUTH_KEY_PREFIX === 'CognitoIdentityServiceProvider'`.) The server variant omits `signInDetails` entirely, so there's also a documented client/server parity gap. A `getStoredSignInDetails(username)` store method used in both places would fix both.

packages/auth/src/providers/cognito/apis/setCurrentUser.ts
```diff
@@ -0,0 +1,105 @@
...
+	// context-scoped clearCredentials, NOT the global import.
+	await ctx.clearCredentials();
+
+	// Resolve the now-active user's identity from their stored id token (no
+	// refresh). Skip the dispatch if the id token is undecodable — emitting an
+	// event with userId:'' would violate the AuthUser contract.
+	const idToken = await tokenStore.getStoredIdToken(username);
+	const userId = (idToken?.payload?.sub as string) ?? '';
+
+	if (!userId) {
```
osama-rizk: Edge case: the switch is already committed above (roster reordered + `clearCredentials()`), then this returns with **no Hub event** if the target's idToken is undecodable. Consumers listening for `switchActiveUser` never learn the active user changed — though it did. Same "mutate before confirming a resolvable identity" ordering exists in `dispatchSignedInHubEvent` (`addActiveSession` runs before `getCurrentUser`, which can throw, leaving the pointer advanced on a *failed* sign-in). Consider resolving identity first, or emitting a payload-less boundary so state and events don't diverge.

packages/auth/src/providers/cognito/apis/signOut.ts
```diff
@@ -85,9 +84,28 @@ export async function signOut(...args: any[]): Promise<void> {
 		}
 	} else {
 		// complete sign out
-		tokenOrchestrator.clearTokens();
+		// Resolve the signed-out user identity from STORED tokens (no refresh)
+		// before any local mutation, so the boundary events can carry it.
+		const tokenStore = tokenOrchestrator.getTokenStore();
+		const activeUsername = await tokenStore.getLastAuthUser();
```
osama-rizk: Minor: when `signOut()` runs with a parked-only roster (no active pointer), `getLastAuthUser()` returns the `'username'` sentinel, the clears become no-ops, and `dispatchSignOutBoundaryEvents(undefined)` still fires `signedOut` **always** — so a sign-out when nobody is active emits a `signedOut` for a user that wasn't signed in. Consider guarding the dispatch on an actual resolved active user.

packages/aws-amplify/__tests__/exports.test.ts
```diff
@@ -189,6 +189,8 @@ describe('aws-amplify Exports', () => {
 					'associateWebAuthnCredential',
 					'listWebAuthnCredentials',
 					'deleteWebAuthnCredential',
+					'listCurrentUsers',
```
osama-rizk, line 193: Test coverage: this guard imports only the client entries, so the **server** `setCurrentUser`/`listCurrentUsers` exports aren't covered — a future accidental drop of a server export wouldn't be caught here. Two smaller gaps while here: the sign-out tests assert the right *calls* but not their *order* (a mutate-before-resolve-identity regression would slip through — the sign-in test already asserts `invocationCallOrder`, so mirror it), and widening the `AuthHubEventData` union source-breaks exhaustive `switch(event){ default: assertNever }` consumers (acceptable under a minor, just worth noting).

packages/auth/src/providers/cognito/apis/signIn.ts
bobbor: Both points taken 👍 The OAuth `prompt` requirement is intended — without it the Hosted UI cookie silently SSOs back as the existing user, so an ungated redirect can't actually add anyone; password flows have no such cookie. Documented the asymmetry on the `prompt` option in 292104e2b, and the changeset now discloses all three behavior changes (signIn guard removal, the OAuth prompt split, and `resolveUsername` throwing instead of the sentinel fallback).

packages/auth/src/providers/cognito/utils/dispatchSignedInHubEvent.ts
bobbor: Disclosed in the changeset in 292104e2b — signedIn now fires only on none→some pointer transitions, called out explicitly for single-session consumers.

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
bobbor: Good catch — real race. Roster mutations are now serialized through an in-store promise chain (`serializeRoster`), so concurrent `addActiveSession`/`removeSession`/reconciliation writes queue instead of last-write-wins. Had to split `getAuthUserList` into a pure reader + serialized persist to avoid re-entrant deadlock. Concurrency test added (two racing sign-ins → both users survive). 292104e2b

packages/auth/src/providers/cognito/tokenProvider/TokenStore.ts
bobbor: Agreed, but went stricter than best-effort: a switch that didn't persist shouldn't look successful. `addActiveSession` now wraps storage write failures in a descriptive `SessionPersistenceException`, and the server `setCurrentUser` JSDoc states it needs a writable context (Route Handler / Server Action / Middleware). Reads stay best-effort as before. 292104e2b

packages/core/src/Hub/types/AuthTypes.ts
bobbor: Both fixed in 292104e2b — event docs rewritten to pointer semantics (signedOut explicitly does NOT imply an empty roster), and I'll reword the PR description to scope the non-destructive claim to the switcher 👍

packages/auth/src/providers/cognito/apis/listCurrentUsers.ts
bobbor: Added `getStoredSignInDetails(username)` on the store and listCurrentUsers now uses it — last hand-built key gone. Kept the server parity gap as documented (deliberate minimal switcher surface). 292104e2b

packages/auth/src/providers/cognito/apis/setCurrentUser.ts
bobbor: Reordered: setCurrentUser resolves the target's identity first and throws UserNotSignedInException before touching anything — an unresolvable session is effectively not switchable, so no half-committed state and no silent event gap. Sign-in dispatch ordering covered by the existing order-asserting test. 292104e2b

packages/auth/src/providers/cognito/apis/signOut.ts
bobbor: Guarded — signOut (native + OAuth) now no-ops when no user is active, so no phantom signedOut for a parked-only roster. Tests added for the case. 292104e2b

packages/aws-amplify/__tests__/exports.test.ts
bobbor: All three: server-entry export assertions added to the guard, sign-out tests now assert invocationCallOrder (mirroring sign-in), and agreed the union widening is acceptable-under-minor — it's noted in the changeset disclosures. 292104e2b

bobbor:
@osama-rizk thanks for the thorough pass — especially the roster race and the read-only-SSR write, both real. All 9 comments addressed in 292104e2b: roster mutations serialized, persistence failures surface as `SessionPersistenceException`, setCurrentUser resolves-before-mutating, parked-only signOut guarded, event docs aligned to pointer semantics, the changeset now discloses the three single-session behavior changes, plus the signInDetails store method and the test-coverage gaps. Full suites green (auth 1243, aws-amplify, adapter-nextjs) + monorepo build clean. Ready for another look 👀

osama-rizk (dismissed):

PR #14882, opened 2026-07-17
ci(e2e): apply retry wrapper to prod cypress runs

bobbor:
# Description of changes

Follow-up to #14880. While analyzing e2e flakiness, we found a retry gap in the Cypress runner: in `callable-e2e-test.yml`, the "Run cypress tests for … prod" step's specialized-`yarn_script` branch invoked `yarn "$E2E_YARN_SCRIPT" …` directly — with **no retry** — while the dev path and the non-specialized prod path both go through `scripts/retry-yarn-script.sh`.

Observed impact (run 29569697344): `integ_next-use-cases-15` failed on a single prod attempt (3/75 tests, cascading from one transient 500 in the sample app's API route) with zero retry protection, while its dev counterpart passed 75/75 under the retry wrapper.

This change routes the specialized prod invocation through the same `retry-yarn-script.sh -n $E2E_RETRY_COUNT` wrapper the dev path uses. Word-splitting semantics were verified against all current `yarn_script` consumers in `.github/integ-config/integ-all.yml` (all single-token args/specs), matching the invocation pattern already used elsewhere in the same file.

## Testing

- YAML validated (`yaml.safe_load`)
- Verified `E2E_RETRY_COUNT` is present in the step's env block (same as dev path)
- Wrapper path matches the sibling invocations under the same `working-directory`


soberm (approved):

osama-rizk (approved):

soberm:
Retry change is not necessary. Prod already used the retry script.

PR #14912, opened 2026-08-19
feat(interactions): migrate Lex providers to AmplifyContext

bobbor:
## Description

B-interactions split (Task 5) of the v6 → v7 AmplifyContext migration. Migrates the `@aws-amplify/interactions` package (both Lex v1 and Lex v2 providers) from the singleton-based `Amplify.getConfig()` pattern to explicit `AmplifyContext` threading.

### Changes

- **Public APIs** `send` / `onComplete` (lex-v1 + lex-v2 subpaths): dual overloads — `fn(input)` (backward compatible) and `fn(ctx, input)` — resolved via `resolveCtxArgs` with global-context fallback
- **`resolveBotConfig`** (both versions): now takes `(ctx, botName)` and reads from `ctx.resourcesConfig.Interactions`
- **Providers**: `sendMessage(ctx, botConfig, message)` — credentials resolved via `ctx.fetchAuthSession()`; no direct singleton imports remain
- **Types**: `IInteractions` gains the dual overload signatures
- **Tests**: all 8 test files migrated to `createMockAmplifyContext()` + `setGlobalContext`/`clearGlobalContext` (no `Amplify.getConfig` mocking); added explicit-ctx test cases

### Design note

The `lexProvider` singleton is kept (ctx threaded per call) rather than adopting v7-poc's `createLexProvider(ctx)` per-call factory — the factory would break the stateful `onComplete` callback registry (`_botsCompleteCallback`), since a callback registered on one instance would never fire from a later `send`.

No server wrappers (interactions is client-only). No changeset (feature-branch target).

## Validation

- `yarn turbo run test --filter=@aws-amplify/interactions` — 8 suites, 34 tests pass (lint runs first in the package test script)
- `yarn turbo run build --filter=@aws-amplify/interactions` — pass
- `npx eslint '**/*.{ts,tsx}'` in package — clean
- `yarn test:size` — within budget (45.87/57 kB default+v2, 39.57/49 kB v1)

### DIFF
--- a/packages/core/src/context/resolveCtxArgs.ts
+++ b/packages/core/src/context/resolveCtxArgs.ts
@@ -35,5 +35,15 @@ export function resolveCtxArgs<T extends unknown[]>(
 		return [args[0], ...args.slice(1)] as [AmplifyContext, ...T];
 	}
 
+	// Guard against mis-ordered calls (e.g. `send(input, ctx)` from untyped JS):
+	// a context anywhere but the first position would otherwise be silently
+	// treated as a regular argument while the call falls back to the global context.
+	if (args.some(isAmplifyContext)) {
+		throw new Error(
+			'AmplifyContext must be passed as the first argument. ' +
+				'Found an AmplifyContext in a later position — check the argument order.',
+		);
+	}
+
 	return [getGlobalContext(), ...args] as [AmplifyContext, ...T];
 }
--- a/packages/core/src/context/resolveCtxArgs.ts
+++ b/packages/core/src/context/resolveCtxArgs.ts
@@ -25,12 +25,6 @@ import { getGlobalContext } from './globalContext';
 export function resolveCtxArgs<T extends unknown[]>(
 	args: unknown[],
 ): [AmplifyContext, ...T] {
-	if (args.length > 1 && args[0] === undefined) {
-		throw new Error(
-			'Undefined AmplifyContext passed. Call configure() first or omit the parameter.',
-		);
-	}
-
 	if (isAmplifyContext(args[0])) {
 		return [args[0], ...args.slice(1)] as [AmplifyContext, ...T];
 	}
@@ -38,10 +32,18 @@ export function resolveCtxArgs<T extends unknown[]>(
 	// Guard against mis-ordered calls (e.g. `send(input, ctx)` from untyped JS):
 	// a context anywhere but the first position would otherwise be silently
 	// treated as a regular argument while the call falls back to the global context.
+	// Checked before the undefined-first-arg guard so the more specific message
+	// wins whenever a context actually exists somewhere in `args`.
 	if (args.some(isAmplifyContext)) {
 		throw new Error(
 			'AmplifyContext must be passed as the first argument. ' +
 				'Found an AmplifyContext in a later position — check the argument order.',
+		);
+	}
+
+	if (args.length > 1 && args[0] === undefined) {
+		throw new Error(
+			'Undefined AmplifyContext passed. Call configure() first or omit the parameter.',
 		);
 	}
 
--- a/packages/interactions/__tests__/lex-v1/apis/send.test.ts
+++ b/packages/interactions/__tests__/lex-v1/apis/send.test.ts
@@ -1,10 +1,15 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { amplifyUuid } from '@aws-amplify/core/internals/utils';
+import {
+	amplifyUuid,
+	clearGlobalContext,
+	setGlobalContext,
+} from '@aws-amplify/core/internals/utils';
 import { lexProvider } from '../../../src/lex-v1/AWSLexProvider';
 import { send } from '../../../src/lex-v1/apis';
 import { generateRandomLexV1Config } from '../../testUtils/randomConfigGeneration';
+import { createMockAmplifyContext } from '../../testUtils/mockAmplifyContext';
 import { resolveBotConfig } from '../../../src/lex-v1/utils';
 import { InteractionsError } from '../../../src/errors/InteractionsError';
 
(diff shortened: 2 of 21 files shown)


packages/interactions/src/lex-v1/AWSLexProvider.ts
```diff
@@ -69,13 +69,14 @@ class AWSLexProvider {
 	}
 
 	async sendMessage(
+		ctx: AmplifyContext,
```
osama-rizk, line 72: Keeping the `lexProvider` singleton (to preserve the `onComplete` registry) means the callback registry is process-global and keyed by bot name, while `ctx` is threaded per call. Two different `AmplifyContext`s using the same bot name would share one callback slot. Fine for single-context apps; worth a note now that multi-context is the point of the migration.

(Same applies to `lex-v2/AWSLexV2Provider.ts`.)

osama-rizk (commented):
Mechanically clean, symmetric across lex-v1/v2, and both dispatch branches are tested. Two notes — one inline, one below; no functional break. Recommend: approve-with-nits.

Non-line note: the overload impls use `...args: any[]`, so external callers keep type safety via the declared overloads, but an internal mis-ordered call (`send(input, ctx)`) would type-check and silently fall through to the global context. Contained, but a latent foot-gun.

bobbor:
Re the overload-impl note (`...args: any[]`): the variadic implementation signature isn't callable from the outside — TS only exposes the two declared overloads, so a mis-ordered `send(input, ctx)` fails to type-check for TS consumers (internal callers included; they import the same overloaded declaration). The exposure is limited to untyped JS callers, and the shape is the established Pattern 1 from Split A (auth landed identically in #14836). Agreed it's a foot-gun worth hardening — but centrally: a runtime guard in `resolveCtxArgs` (reject extra args when the first isn't a context) covers every category at once. We'll provide a follow-up there rather than diverging in this package.

packages/interactions/src/lex-v1/AWSLexProvider.ts
bobbor: Validated — the sharing is real: the registry is process-global and keyed by bot name, so two contexts using the same bot name would share one callback slot. It's the deliberate tradeoff in this split: the per-call factory (v7-poc shape) drops registered callbacks entirely, and a proper per-context registry needs an identity-keyed store (e.g. `WeakMap<AmplifyContext, registry>`), which changes the pattern for all stateful providers — that's Phase C scope. Since interactions is client-only in the B-splits, the single-global-context path is the only one exercisable today. We'll provide a follow-up (per-context registry + code note) rather than expanding this PR.

bobbor:
Update: the `resolveCtxArgs` hardening landed in this PR after all (33d518c6d) rather than as a follow-up — `resolveCtxArgs` now throws when a branded `AmplifyContext` appears in any non-first position, so a mis-ordered `send(input, ctx)` from untyped JS fails loudly instead of silently falling back to the global context. Covers all category packages centrally. Verified: core 103 suites / 723 tests green, interactions 8 suites / 34 tests green. The per-context `onComplete` registry note remains a Phase C follow-up.

osama-rizk (approved):

packages/core/src/context/resolveCtxArgs.ts
```diff
@@ -35,5 +35,15 @@ export function resolveCtxArgs<T extends unknown[]>(
 		return [args[0], ...args.slice(1)] as [AmplifyContext, ...T];
 	}
 
+	// Guard against mis-ordered calls (e.g. `send(input, ctx)` from untyped JS):
+	// a context anywhere but the first position would otherwise be silently
+	// treated as a regular argument while the call falls back to the global context.
+	if (args.some(isAmplifyContext)) {
```
soberm, line 37: [minor] This new check gets shadowed for `resolveCtxArgs([undefined, ctx])`: the `args[0] === undefined` guard above (line 28) fires first and throws the generic "Undefined AmplifyContext passed" message, so a caller who happens to have `undefined` first but a real context later still gets the less specific error instead of this one. Swapping the order (this mis-order check before the undefined check) would make the more helpful message win whenever a context actually exists somewhere in `args`.

packages/interactions/__tests__/lex-v1/apis/send.test.ts
```diff
@@ -29,8 +43,28 @@ describe('Interactions LexV1 API: send', () => {
...
+			v1BotConfig.name,
+		);
 		expect(mockLexProvider).toHaveBeenCalledTimes(1);
-		expect(mockLexProvider).toHaveBeenCalledWith(v1BotConfig, message);
+		expect(mockLexProvider).toHaveBeenCalledWith(
+			explicitCtx,
+			v1BotConfig,
+			message,
+		);
 	});
```
soberm, line 68: [minor] Now that `resolveCtxArgs` throws on mis-ordered args, might be worth a test here that exercises it through `send` itself instead of only at the `resolveCtxArgs` unit level in `packages/core`. Something like `await expect((send as any)({ botName: v1BotConfig.name, message: 'hi' }, mockCtx)).rejects.toThrow('AmplifyContext must be passed as the first argument')`. Right now nothing confirms the guard is actually reachable through the public entry point (same gap applies to the v2 and `onComplete` test files).

packages/core/src/context/resolveCtxArgs.ts
bobbor: Good catch — fixed in d782c2bfb. The mis-order check now runs before the undefined-first-arg guard, so `resolveCtxArgs([undefined, ctx])` gets the specific message. Added a unit test covering exactly that shadowing case.

packages/interactions/__tests__/lex-v1/apis/send.test.ts
bobbor: Done in d782c2bfb — added a mis-ordered-args test through the public entry point in all four API test files (send + onComplete, lex-v1 + lex-v2), asserting both the specific error and that the provider is never invoked. Used a typed variadic cast instead of `as any` (lint disallows explicit any).

soberm (approved):

PR #14913, opened 2026-08-19
feat(analytics): migrate analytics APIs to AmplifyContext

bobbor:
## Description

Migrates the `@aws-amplify/analytics` package from the global `Amplify` singleton to explicit `AmplifyContext` threading, following the same pattern as the landed auth migration (#14836). This is the **B-analytics** step of the v6→v7 context migration, targeting the `feat/bobbor/v6-context` feature branch.

### What changed

- **Public APIs across all four providers** got `(ctx, input)` overloads alongside `(input)`, with a variadic impl using `resolveCtxArgs` (global-context fallback preserved for existing callers):
  - **pinpoint**: `record`, `identifyUser`, `flushEvents`, `configureAutoTrack`
  - **kinesis** / **kinesis-firehose**: `record`, `flushEvents`, `configureAutoTrack` (incl. the auto-track APIs added in #14854)
  - **personalize**: `record`, `flushEvents`
- **`resolveConfig` (per provider)**: now takes `ctx: AmplifyContext` and reads `ctx.resourcesConfig.Analytics?.<Provider>` instead of `Amplify.getConfig()`.
- **`resolveCredentials` (shared + pinpoint)**: uses `ctx.fetchAuthSession()` — no direct singleton imports remain.
- **`deprecatePinpoint` wrapper**: retyped as `<TFn extends (...args: any[]) => any>(fn: TFn): TFn` so the context overloads survive on the wrapped default exports (#14877 had erased overload signatures). A compile-time test asserts both call signatures surface on the wrapped `record` from the package index.
- **No server wrappers** — analytics is a client-only package.

### Tests

- Added `__tests__/testUtils/mockAmplifyContext.ts` (branded factory, same as auth).
- Migrated 19 test files off `Amplify.getConfig` spies to `createMockAmplifyContext` + `setGlobalContext`/`clearGlobalContext` (with `afterAll` cleanup). Notes for reviewers:
  - The `jest.isolateModules`-based `configureAutoTrack` tests register the mock context from inside each isolated registry (the global-context store is per-module-registry).
  - `pinpoint/apis/record.test.ts` switched from a full `@aws-amplify/core` automock to a partial mock (`jest.requireActual` + `Hub.dispatch: jest.fn()`) so the real `AMPLIFY_CONTEXT_BRAND` stays intact for `resolveCtxArgs`.
  - Kinesis/firehose `configureAutoTrack` tests now assert `record` is called with `(ctx, payload)`.

## Testing

- `yarn turbo run build --filter=@aws-amplify/analytics` → passes
- `yarn turbo run test --filter=@aws-amplify/analytics` → **30 suites / 112 tests pass**, lint clean, ts-coverage passes
- No `tsconfig.tsbuildinfo` committed

## Checklist

- [x] Unit tests updated/added; coverage preserved
- [x] No changeset (targets a feature branch, not `main`)
- [x] Scope limited to `packages/analytics/`
- [x] Client-only package — no server wrapper changes needed

### DIFF
--- a/packages/analytics/__tests__/providers/pinpoint/apis/configureAutoTrack.test.ts
+++ b/packages/analytics/__tests__/providers/pinpoint/apis/configureAutoTrack.test.ts
@@ -47,10 +47,6 @@ describe('Pinpoint API: configureAutoTrack', () => {
 			const {
 				configureAutoTrack,
 			} = require('../../../../src/providers/pinpoint/apis');
-			// The isolated registry has its own global context storage; register the
-			// mock context so `resolveCtxArgs` can resolve it.
-			const { setGlobalContext } = require('@aws-amplify/core/internals/utils');
-			setGlobalContext(mockCtx);
 
 			try {
 				configureAutoTrack({
@@ -68,10 +64,6 @@ describe('Pinpoint API: configureAutoTrack', () => {
 			const {
 				configureAutoTrack,
 			} = require('../../../../src/providers/pinpoint/apis');
-			// The isolated registry has its own global context storage; register the
-			// mock context so `resolveCtxArgs` can resolve it.
-			const { setGlobalContext } = require('@aws-amplify/core/internals/utils');
-			setGlobalContext(mockCtx);
 
 			configureAutoTrack(MOCK_INPUT);
 		});
@@ -92,10 +84,6 @@ describe('Pinpoint API: configureAutoTrack', () => {
 			const {
 				configureAutoTrack,
 			} = require('../../../../src/providers/pinpoint/apis');
-			// The isolated registry has its own global context storage; register the
-			// mock context so `resolveCtxArgs` can resolve it.
-			const { setGlobalContext } = require('@aws-amplify/core/internals/utils');
-			setGlobalContext(mockCtx);
 
 			configureAutoTrack(testInput);
 		});
@@ -116,10 +104,6 @@ describe('Pinpoint API: configureAutoTrack', () => {
 			const {
 				configureAutoTrack,
 			} = require('../../../../src/providers/pinpoint/apis');
-			// The isolated registry has its own global context storage; register the
-			// mock context so `resolveCtxArgs` can resolve it.
-			const { setGlobalContext } = require('@aws-amplify/core/internals/utils');
-			setGlobalContext(mockCtx);
 
 			configureAutoTrack(testInput);
 		});
@@ -130,15 +114,40 @@ describe('Pinpoint API: configureAutoTrack', () => {
 		);
 	});
 
+	it('accepts an explicitly provided context', () => {
+		jest.isolateModules(() => {
+			const {
+				configureAutoTrack,
+			} = require('../../../../src/providers/pinpoint/apis');
+
+			configureAutoTrack(mockCtx, MOCK_INPUT);
+		});
+
+		expect(MockEventTracker).toHaveBeenCalledWith(
+			expect.any(Function),
+			MOCK_INPUT.options,
+		);
+	});
+
+	it('can be configured before Amplify.configure (no global context)', () => {
+		jest.isolateModules(() => {
+			const {
+				configureAutoTrack,
+			} = require('../../../../src/providers/pinpoint/apis');
+
+			// The isolated registry has NO global context set — tracker setup must
+			// not resolve the context eagerly, so this must not throw.
+			expect(() => {
+				configureAutoTrack(MOCK_INPUT);
+			}).not.toThrow();
+		});
+	});
+
 	it('Reconfigures an existing tracker', () => {
 		jest.isolateModules(() => {
 			const {
 				configureAutoTrack,
 			} = require('../../../../src/providers/pinpoint/apis');
-			// The isolated registry has its own global context storage; register the
-			// mock context so `resolveCtxArgs` can resolve it.
-			const { setGlobalContext } = require('@aws-amplify/core/internals/utils');
-			setGlobalContext(mockCtx);
 
 			// Enable the tracker
 			configureAutoTrack(MOCK_INPUT);
@@ -165,10 +174,6 @@ describe('Pinpoint API: configureAutoTrack', () => {
 			const {
 				configureAutoTrack,
 			} = require('../../../../src/providers/pinpoint/apis');
-			// The isolated registry has its own global context storage; register the
-			// mock context so `resolveCtxArgs` can resolve it.
-			const { setGlobalContext } = require('@aws-amplify/core/internals/utils');
-			setGlobalContext(mockCtx);
 
 			// Enable the tracker
 			configureAutoTrack(MOCK_INPUT);
--- a/packages/analytics/__tests__/providers/pinpoint/apis/configureAutoTrack.test.ts
+++ b/packages/analytics/__tests__/providers/pinpoint/apis/configureAutoTrack.test.ts
@@ -7,9 +7,11 @@ import {
 	PageViewTracker,
 	SessionTracker,
 } from '../../../../src/trackers';
+import { record } from '../../../../src/providers/pinpoint/apis/record';
 import { createMockAmplifyContext } from '../../../testUtils/mockAmplifyContext';
 
 jest.mock('../../../../src/trackers');
+jest.mock('../../../../src/providers/pinpoint/apis/record');
 
 const mockCtx = createMockAmplifyContext();
 
@@ -24,6 +26,7 @@ const MOCK_INPUT = {
 } as ConfigureAutoTrackInput;
 
 describe('Pinpoint API: configureAutoTrack', () => {
+	const mockRecord = record as jest.Mock;
 	const MockEventTracker = EventTracker as jest.MockedClass<
 		typeof EventTracker
 	>;
@@ -38,6 +41,7 @@ describe('Pinpoint API: configureAutoTrack', () => {
 		MockEventTracker.mockClear();
 		MockPageViewTracker.mockClear();
 		MockSessionTracker.mockClear();
+		mockRecord.mockClear();
 	});
 
 	it('Validates the tracker configuration', () => {
@@ -114,7 +118,7 @@ describe('Pinpoint API: configureAutoTrack', () => {
 		);
 	});
 
-	it('accepts an explicitly provided context', () => {
+	it('accepts an explicitly provided context and threads it through to record', () => {
 		jest.isolateModules(() => {
 			const {
 				configureAutoTrack,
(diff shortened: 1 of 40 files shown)


packages/analytics/src/providers/pinpoint/apis/configureAutoTrack.ts
```diff
@@ -46,9 +49,19 @@ const emitTrackingEvent = (
...
+export function configureAutoTrack(...args: any[]): void {
+	const [ctx, input] = resolveCtxArgs<[ConfigureAutoTrackInput]>(args);
 	validateTrackerConfiguration(input);
 
 	// Initialize or update this provider's trackers
-	updateProviderTrackers(input, emitTrackingEvent, configuredTrackers);
-};
+	updateProviderTrackers(
+		input,
+		emitTrackingEvent.bind(null, ctx),
```
osama-rizk, line 80: `configureAutoTrack` now resolves the context eagerly and captures it (`.bind(null, ctx)` / closure), where the old code resolved config lazily at each event emit. Two effects: (a) calling it before `Amplify.configure()` — common during module-init tracker setup — now throws; (b) since reconfigure is supported, auto-tracked events keep emitting against the appId/region captured at setup time after a later `configure()`. Recommend resolving `ctx` lazily inside `emitTrackingEvent` so auto-track follows live config like the one-shot APIs do. Same shape applies to the kinesis and kinesis-firehose `configureAutoTrack`.

packages/analytics/src/providers/pinpoint/apis/record.ts
```diff
@@ -51,9 +52,12 @@ const logger = new ConsoleLogger('Analytics');
  * })
  * ```
  */
-export const record = (input: RecordInput): void => {
+export function record(input: RecordInput): void;
+export function record(ctx: AmplifyContext, input: RecordInput): void;
+export function record(...args: any[]): void {
+	const [ctx, input] = resolveCtxArgs<[RecordInput]>(args);
```
osama-rizk, line 58: Minor contract note: on the unconfigured path this now throws a generic `Error('No AmplifyContext available')` synchronously, where callers previously saw a typed `AnalyticsError` (`NoAppId`/`NoRegion`). Code doing `catch (e) { if (e instanceof AnalyticsError) … }` won't match. Low severity (misuse path) — flagging for awareness / a possible typed re-wrap.

osama-rizk (commented):
Consistent migration, the `deprecatePinpoint` overload fix is careful, and the happy paths are well-tested. One functional finding on `configureAutoTrack` (inline) now that reconfigure is supported, plus a small error-surface note. Recommend: approve-with-nits once the `configureAutoTrack` capture is addressed.

packages/analytics/src/providers/pinpoint/apis/configureAutoTrack.ts
bobbor: Valid catch — confirmed both effects against core's `Amplify.configure()` (each call publishes a fresh frozen ctx via `setGlobalContext`, so a captured ctx goes stale, and eager resolution throws pre-configure). Fixed in 14aca2d59: all three `configureAutoTrack`s now resolve only an *explicitly passed* ctx (via `isAmplifyContext`, no global fallback at setup time); when none is supplied, `emitTrackingEvent` calls `record(input)` so the global context is resolved lazily at emit time — pre-`configure()` tracker setup works again and auto-tracked events follow live config. Added regression tests for both (setup-without-global-context, explicit-ctx passthrough).

packages/analytics/src/providers/pinpoint/apis/record.ts
bobbor: Agreed it's a real contract difference, but low severity (unconfigured-misuse path) and it equally affects the already-landed auth/storage splits — the same generic error comes from core's `getGlobalContext()`. We'll provide a follow-up: a core-level typed re-wrap (or category-level wrapping of the ctx-resolution error into `AnalyticsError`-style typed errors) so all categories fix this consistently, rather than special-casing analytics in this PR.

osama-rizk (approved):

packages/analytics/src/providers/kinesis-firehose/apis/configureAutoTrack.ts
```diff
@@ -73,9 +75,25 @@ const configuredTrackers: Partial<Record<TrackerType, TrackerInterface>> = {};
...
 	input: KinesisFirehoseConfigureAutoTrackInput,
-): void => {
+): void;
+export function configureAutoTrack(...args: any[]): void {
+	// Resolve the optional leading context WITHOUT falling back to the global
+	// context. The context is only needed when events are emitted; resolving it
+	// eagerly would (a) throw when trackers are configured before
+	// `Amplify.configure()` and (b) pin auto-tracked events to the configuration
+	// snapshot captured at setup time after a later `configure()` call.
+	const [ctx, input] = isAmplifyContext(args[0])
```
soberm, line 91: [minor] This ctx-peeking ternary is copy-pasted into `kinesis/apis/configureAutoTrack.ts` and `pinpoint/apis/configureAutoTrack.ts` too, only the input type changes between the three. Since all three got touched for the same reason, might be worth factoring this into one small helper in `src/utils` (something like `resolvePeekedCtxArgs<TInput>(args: any[])`), so a future tweak to this logic doesn't need three synchronized edits.

packages/analytics/src/providers/pinpoint/apis/configureAutoTrack.ts
```diff
@@ -20,15 +21,25 @@ import { record } from './record';
 // Configured Tracker instances for Pinpoint
 const configuredTrackers: Partial<Record<TrackerType, TrackerInterface>> = {};
 
-// Callback that will emit an appropriate event to Pinpoint when required by the Tracker
+// Callback that will emit an appropriate event to Pinpoint when required by the Tracker.
+// When no explicit context was supplied to `configureAutoTrack`, the context is
+// deliberately NOT captured here — `record` resolves the global context lazily at
+// emit time, so auto-tracked events follow the live configuration and trackers can
+// be set up before `Amplify.configure()` is called.
 const emitTrackingEvent = (
```
soberm, line 29: [nit] Kinesis and Kinesis Firehose build `emitTrackingEvent` as a closure inside `configureAutoTrack` that captures `ctx` directly. This one keeps it at module scope and threads `ctx` in via `.bind(null, ctx)` at the call site instead. Both work, but given all three got the same treatment in this PR, might be worth picking one style. The closure is a bit easier to follow than tracing the bound arg back to the `updateProviderTrackers` call below.

packages/analytics/__tests__/providers/pinpoint/apis/configureAutoTrack.test.ts
```diff
@@ -111,6 +114,35 @@ describe('Pinpoint API: configureAutoTrack', () => {
...
+	it('accepts an explicitly provided context', () => {
+		jest.isolateModules(() => {
+			const {
+				configureAutoTrack,
+			} = require('../../../../src/providers/pinpoint/apis');
+
+			configureAutoTrack(mockCtx, MOCK_INPUT);
+		});
+
+		expect(MockEventTracker).toHaveBeenCalledWith(
```
soberm, line 130: [minor] This only checks that `EventTracker` gets constructed, it doesn't actually confirm `mockCtx` reaches `record()` when the tracker fires. The kinesis and kinesis-firehose versions of this same test capture the emitted callback and assert `mockRecord` was called with the ctx; this one can't because `record` isn't mocked in this file. Might be worth adding `jest.mock('../../../../src/providers/pinpoint/apis/record')` here too so this test actually proves the ctx makes it through, not just that it's accepted without throwing.

packages/analytics/src/index.ts
```diff
@@ -4,8 +4,9 @@
 // This entry point intentionally re-exports the deprecated Pinpoint provider
 // APIs as the default Analytics surface. Removing them would be a breaking
 // change; instead each is wrapped to emit a one-time runtime deprecation
-// warning. The deprecated imports below are therefore expected.
-/* eslint-disable import/no-deprecated */
+// warning. The deprecated imports below are therefore expected. (Note: the
+// `import/no-deprecated` rule no longer detects these since the APIs became
```
soberm, line 8: [question] `import/no-deprecated` matches based on the `@deprecated` tag on the resolved symbol, so I wouldn't have expected switching from a const arrow function to an overloaded function declaration to change whether it fires. Did you verify this against an actual lint run with the disable removed, or is this more of a reasoned guess? If the rule just isn't covering this file for an unrelated reason rather than genuinely not matching overloads, we'd want the disable comment back before it silently stops warning on some future change here.

packages/analytics/__tests__/providers/pinpoint/apis/configureAutoTrack.test.ts
bobbor: Fair — the test proved the ctx was accepted but not that it arrived. Fixed in 178d24300: `record` is now mocked in this file too, the explicit-ctx test fires the recorder callback and asserts `record(mockCtx, payload)`, and the pre-configure test additionally asserts the global-fallback path emits `record(payload)` (no captured ctx). Mirrors the kinesis/firehose coverage.

packages/analytics/src/index.ts
bobbor: Verified against an actual lint run, not a guess: with the directive still present, eslint reported `8:1 warning Unused eslint-disable directive (no problems were reported from 'import/no-deprecated')` for this file — in the same run where the rule actively fired on the `record`/`flushEvents` imports in the test files (which import the wrapped consts, whose `@deprecated` JSDoc it does resolve). So the rule covers the file and matches const exports, but doesn't resolve the `@deprecated` tag through overloaded function declarations (JSDoc sits on the first overload signature; the plugin's ExportMap resolves the binding to a declaration without it). Agreed there's a residual risk if a future non-overloaded deprecated import lands here — happy to re-add a targeted `eslint-disable-next-line` per import in the follow-up if you prefer belt-and-suspenders.

packages/analytics/src/providers/kinesis-firehose/apis/configureAutoTrack.ts
bobbor: Agreed — will provide a follow-up factoring the ctx-peeking ternary into a shared `src/utils` helper (e.g. `peekCtxArgs<TInput>(args)`) used by all three providers. Keeping it out of this PR to avoid churning all three files again while it's approved+green; the follow-up will also unify the emit-callback style (see other thread).

packages/analytics/src/providers/pinpoint/apis/configureAutoTrack.ts
bobbor: Agreed the closure reads better. Will unify all three on the closure style in the same follow-up that extracts the shared ctx-peeking helper — pinpoint's module-scope + bind shape came from mirroring v7-poc, but there's no reason to keep two styles.

soberm (approved):

PR #14914, opened 2026-08-19
feat(api-rest): migrate REST API to AmplifyContext

bobbor:
#### Description of changes

B-api-rest split of the v6 context migration (Task 3 of the migration plan). Migrates `@aws-amplify/api-rest` from the `AmplifyClassV6` singleton pattern to `AmplifyContext`:

- **Public APIs** (`get`, `post`, `put`, `del`, `head`, `patch` in `src/apis/index.ts`): overloaded signatures `fn(input)` + `fn(ctx, input)` via `resolveCtxArgs` with global-context fallback (Pattern 1)
- **Internals** (`publicApis.ts`, `transferHandler.ts`, `internalPost.ts`): first param type changed to `AmplifyContext`; auth resolved via `ctx.fetchAuthSession()` instead of `.Auth.fetchAuthSession()`
- **Utils** (`resolveApiUrl`, `parseSigningInfo`, `resolveLibraryOptions`): read from `ctx.resourcesConfig` / `ctx.libraryOptions` instead of `getConfig()` (Pattern 4)
- **Server wrappers kept** (`src/apis/server.ts`, `src/internals/server.ts`): accept `AmplifyContext | AmplifyServer.ContextSpec`, normalized via new cast-based `resolveServerContext` adapter mirroring the landed B-auth version (Pattern 5 — full server-path bridging deferred to Phase C)
- **Tests**: new branded `createMockAmplifyContext` factory (with typed overrides); migrated tests use `setGlobalContext`/`clearGlobalContext` — no `Amplify.getConfig` mocking

Follows the same shape as B-auth (#14836) and B-storage (#14874). Unblocks B-api-graphql.

#### Issue #, if available

N/A — part of the v6 context migration plan.

- `yarn turbo run build --filter=@aws-amplify/api-rest` — exit 0
- `npx eslint '**/*.{ts,tsx}'` in the package — exit 0 (one pre-existing jsdoc warning)
- `yarn turbo run test --filter=@aws-amplify/api-rest` — 10 suites, 220 tests passed

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/api-rest/__tests__/server.test.ts
+++ b/packages/api-rest/__tests__/server.test.ts
@@ -27,64 +27,89 @@ const contextSpec = { token: { value: 'token' } } as any;
 const mockGetAmplifyServerContext = getAmplifyServerContext as jest.Mock;
 
 describe('REST API server handlers', () => {
+	const mockAmplify = {
+		getConfig: jest.fn().mockReturnValue({ API: { REST: {} } }),
+		libraryOptions: {},
+		Auth: {
+			fetchAuthSession: jest.fn().mockResolvedValue({}),
+			clearCredentials: jest.fn().mockResolvedValue(undefined),
+			getTokens: jest.fn().mockResolvedValue(undefined),
+		},
+	};
+
 	beforeEach(() => {
 		jest.clearAllMocks();
 		mockGetAmplifyServerContext.mockReturnValue({
-			amplify: 'mockedAmplifyServerSideContext',
+			amplify: mockAmplify,
 		});
 	});
 
 	describe('legacy ContextSpec path', () => {
-		it('get should call common get API with server-side Amplify context', () => {
+		it('get should call common get API with bridged server-side Amplify context', () => {
 			get(contextSpec, input);
 			expect(mockGetAmplifyServerContext).toHaveBeenCalledWith(contextSpec);
 			expect(commonGet).toHaveBeenCalledWith(
-				'mockedAmplifyServerSideContext',
+				expect.objectContaining({
+					libraryOptions: mockAmplify.libraryOptions,
+					fetchAuthSession: expect.any(Function),
+					clearCredentials: expect.any(Function),
+					getTokens: expect.any(Function),
+				}),
 				input,
 			);
 		});
 
-		it('post should call common post API with server-side Amplify context', () => {
+		it('post should call common post API with bridged server-side Amplify context', () => {
 			post(contextSpec, input);
 			expect(mockGetAmplifyServerContext).toHaveBeenCalledWith(contextSpec);
 			expect(commonPost).toHaveBeenCalledWith(
-				'mockedAmplifyServerSideContext',
+				expect.objectContaining({
+					fetchAuthSession: expect.any(Function),
+				}),
 				input,
 			);
 		});
 
-		it('put should call common put API with server-side Amplify context', () => {
+		it('put should call common put API with bridged server-side Amplify context', () => {
 			put(contextSpec, input);
 			expect(mockGetAmplifyServerContext).toHaveBeenCalledWith(contextSpec);
 			expect(commonPut).toHaveBeenCalledWith(
-				'mockedAmplifyServerSideContext',
+				expect.objectContaining({
+					fetchAuthSession: expect.any(Function),
+				}),
 				input,
 			);
 		});
 
-		it('del should call common del API with server-side Amplify context', () => {
+		it('del should call common del API with bridged server-side Amplify context', () => {
 			del(contextSpec, input);
 			expect(mockGetAmplifyServerContext).toHaveBeenCalledWith(contextSpec);
 			expect(commonDel).toHaveBeenCalledWith(
-				'mockedAmplifyServerSideContext',
+				expect.objectContaining({
+					fetchAuthSession: expect.any(Function),
+				}),
 				input,
 			);
 		});
 
-		it('patch should call common patch API with server-side Amplify context', () => {
+		it('patch should call common patch API with bridged server-side Amplify context', () => {
 			patch(contextSpec, input);
 			expect(mockGetAmplifyServerContext).toHaveBeenCalledWith(contextSpec);
 			expect(commonPatch).toHaveBeenCalledWith(
-				'mockedAmplifyServerSideContext',
+				expect.objectContaining({
+					fetchAuthSession: expect.any(Function),
+				}),
 				input,
 			);
 		});
 
-		it('head should call common head API with server-side Amplify context', () => {
+		it('head should call common head API with bridged server-side Amplify context', () => {
 			head(contextSpec, input);
 			expect(mockGetAmplifyServerContext).toHaveBeenCalledWith(contextSpec);
 			expect(commonHead).toHaveBeenCalledWith(
-				'mockedAmplifyServerSideContext',
+				expect.objectContaining({
+					fetchAuthSession: expect.any(Function),
+				}),
 				input,
 			);
 		});
--- a/packages/api-rest/__tests__/server.test.ts
+++ b/packages/api-rest/__tests__/server.test.ts
@@ -13,6 +13,8 @@ import {
 	put as commonPut,
 } from '../src/apis/common/publicApis';
 
+import { createMockAmplifyContext } from './testUtils/mockAmplifyContext';
+
 jest.mock('../src/apis/common/publicApis');
 jest.mock('@aws-amplify/core/internals/adapter-core');
 
@@ -24,7 +26,7 @@ const input = {
 const contextSpec = { token: { value: 'token' } } as any;
 const mockGetAmplifyServerContext = getAmplifyServerContext as jest.Mock;
 
-describe('REST API handlers', () => {
+describe('REST API server handlers', () => {
 	beforeEach(() => {
 		jest.clearAllMocks();
 		mockGetAmplifyServerContext.mockReturnValue({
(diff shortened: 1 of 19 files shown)


packages/api-rest/src/apis/common/transferHandler.ts
```diff
@@ -139,10 +139,10 @@ const getRetryDeciderFromStrategy = (
 };
 
 const resolveCredentials = async (
-	amplify: AmplifyClassV6,
+	amplify: AmplifyContext,
 ): Promise<AWSCredentials | null> => {
 	try {
-		const { credentials } = await amplify.Auth.fetchAuthSession();
+		const { credentials } = await amplify.fetchAuthSession();
```
osama-rizk, line 145: On the legacy server (`ContextSpec`) path this crashes. `resolveServerContext` returns `getAmplifyServerContext(spec).amplify`, which is a bare `AmplifyClass` — and `AmplifyClass` has no top-level `fetchAuthSession` (it only exposes `.Auth.fetchAuthSession()`; the top-level methods are built onto the *context* object `configure()` publishes, core `Amplify.ts:94-96`). So any default-auth SSR REST call (the default — the guard is `defaultAuthMode !== 'none'`) throws `TypeError: amplify.fetchAuthSession is not a function`. This path worked before the PR. It also propagates through `internalPost`, which api-graphql and storage SSR route through. Fix: bridge in `resolveServerContext` the way storage's version does — map `ctx.fetchAuthSession → amplify.Auth.fetchAuthSession` (and `clearCredentials`/`getTokens`) rather than a bare cast.

packages/api-rest/src/apis/server/resolveServerContext.ts
```diff
@@ -0,0 +1,24 @@
...
+export const resolveServerContext = (
+	ctxOrContextSpec: AmplifyContext | AmplifyServer.ContextSpec,
+): AmplifyContext => {
+	const ctx =
+		'resourcesConfig' in ctxOrContextSpec
+			? ctxOrContextSpec
+			: getAmplifyServerContext(ctxOrContextSpec).amplify;
+
+	// `'resourcesConfig' in x` doesn't narrow the union; AmplifyClass satisfies AmplifyContext structurally.
+	return ctx as AmplifyContext;
```
osama-rizk: The comment "AmplifyClass satisfies AmplifyContext structurally" isn't accurate — `AmplifyClass` has no top-level `fetchAuthSession`/`clearCredentials`/`getTokens`, so the cast silences a real mismatch rather than reflecting one. Bridging the three auth methods here (as storage does) makes the cast honest and fixes the server crash.

packages/api-rest/__tests__/server.test.ts
```diff
@@ -13,6 +13,8 @@ import {
 	put as commonPut,
 } from '../src/apis/common/publicApis';
 
+import { createMockAmplifyContext } from './testUtils/mockAmplifyContext';
+
 jest.mock('../src/apis/common/publicApis');
```
osama-rizk, line 18: This mock is why the suite is green while SSR is broken: stubbing `publicApis` means `transferHandler → resolveCredentials` never runs against the real server object, so the exact regression above is never exercised. Worth one server test that drives a real request end-to-end (`createAmplifyServerContext → get → transferHandler`) — it would fail today with the `TypeError`.

osama-rizk (commented):
Client migration is clean and well-tested across both overloads. One blocker on the server path — details inline. Recommend: needs-changes. The fix is already proven in-repo (storage's `resolveServerContext` bridges the auth methods).

packages/api-rest/src/apis/common/transferHandler.ts
bobbor: Validated and fixed in a57689c49 — `resolveServerContext` now bridges `fetchAuthSession`/`clearCredentials`/`getTokens` to `amplify.Auth.*` (ported from the storage PR's adapter). One nuance: the TypeError was actually swallowed by `resolveCredentials`'s try/catch, so the observable failure was IAM requests silently going out unsigned (then 403 at the service) rather than a crash — worse to debug, same root cause. The new e2e test fails against the old adapter and passes with the bridge.

packages/api-rest/src/apis/server/resolveServerContext.ts
bobbor: Fixed in a57689c49 — replaced the cast (and its inaccurate comment) with the bridged adapter matching storage's version, including the `isAmplifyContext` brand check instead of the structural probe.

packages/api-rest/__tests__/server.test.ts
bobbor: Added in a57689c49: `__tests__/apis/server.e2e.test.ts` drives the real chain (server `get` → publicApis → transferHandler → resolveCredentials) via `createAmplifyServerContext`, with only the low-level HTTP handlers mocked. Verified it fails against the old cast-based adapter (request fell through to unauthenticatedHandler) and passes with the bridge, asserting the signed request carries credentials/region/service.

osama-rizk (approved):

packages/api-rest/src/utils/bridgeAmplifyClass.ts
```diff
@@ -0,0 +1,34 @@
...
+ * This is used:
+ * - In the server wrapper where a legacy `ContextSpec` provides an `AmplifyClass` instance.
+ * - In the internals `post` function where dependents (e.g. api-graphql) still pass
+ *   an `AmplifyClassV6` until they migrate to `AmplifyContext`.
+ *
+ * @internal
+ */
+export const bridgeAmplifyClass = (
+	amplify: AmplifyClassV6,
+): AmplifyContext => ({
```
soberm: [blocker] This object never gets branded with `AMPLIFY_CONTEXT_BRAND` — compare with `createMockAmplifyContext` in the tests or `AmplifyClass.configure()` in core, both of which call `Object.defineProperty(ctx, AMPLIFY_CONTEXT_BRAND, ...)` on their output.

That matters here because this bridge runs *twice* on the legacy-`ContextSpec` server path: `resolveServerContext()` bridges the raw `AmplifyClassV6` and hands the (unbranded) result to `internalPost.post()`, which does its own `isAmplifyContext(amplify) ? amplify : bridgeAmplifyClass(amplify)` check. Since the first output isn't branded, it gets treated as a raw `AmplifyClassV6` and bridged again — but the second bridge's `fetchAuthSession` closure reads `.Auth` off an object that doesn't have one, so it throws `TypeError: Cannot read properties of undefined (reading 'fetchAuthSession')`.

That error never surfaces though: `transferHandler`'s `resolveCredentials()` catches it, logs a debug line, and returns `null`. So any server-side IAM-signed request through `@aws-amplify/api-rest/internals/server`'s `post()` with a legacy `ContextSpec` silently drops its credentials and goes out unsigned. I reproduced the bridging logic standalone to confirm — the `TypeError` fires every time.

One-line fix: brand the object this returns, the same way `createMockAmplifyContext` does, so the second `isAmplifyContext` check downstream recognizes it as already-bridged.

packages/api-rest/src/internals/server.ts
```diff
@@ -22,13 +21,19 @@ import { InternalPostInput } from '../types';
...
+ *
  * @internal
  */
 export const post = (
-	contextSpec: AmplifyServer.ContextSpec,
+	ctxOrContextSpec: AmplifyContext | AmplifyServer.ContextSpec,
 	input: InternalPostInput,
 ) => {
-	return internalPost(getAmplifyServerContext(contextSpec).amplify, input);
+	return internalPost(resolveServerContext(ctxOrContextSpec), input);
```
soberm, line 36: [blocker] This is the call site that actually hits the double-bridging issue from `bridgeAmplifyClass.ts` (left a detailed comment there): `resolveServerContext()` bridges a legacy `ContextSpec` into an unbranded context, then `internalPost.post()` re-checks `isAmplifyContext` on that result, fails, and bridges it a second time — which breaks `fetchAuthSession` and silently drops credentials for the request.

Also worth flagging: there's no test file for this module at all, before or after this PR. `server.e2e.test.ts` covers the equivalent legacy-`ContextSpec` flow for `apis/server.ts`'s `get`, but that path doesn't go through `internalPost`'s bridging so it wouldn't catch this. An e2e test here in the same style (real `createAmplifyServerContext` + spied `Auth.fetchAuthSession`) asserting the outgoing request is actually IAM-signed would catch this class of bug.

packages/api-rest/src/utils/bridgeAmplifyClass.ts
```diff
@@ -0,0 +1,34 @@
...
+	get resourcesConfig() {
+		return amplify.getConfig();
+	},
+	libraryOptions: amplify.libraryOptions,
```
soberm: [minor] `resourcesConfig` above is a live getter that re-reads `amplify.getConfig()` on every access, but `libraryOptions` is captured once at bridge-creation time. `AmplifyClass.libraryOptions` gets reassigned wholesale on every `configure()` call (not mutated in place), so if a bridged context were ever retained across a reconfigure, this field would go stale while `resourcesConfig` stays current. Doesn't bite today since every call site here re-bridges per request, but making this a getter too (`get libraryOptions() { return amplify.libraryOptions; }`) would remove the inconsistency.

soberm (changes requested):
Requesting changes for a bug in the server-side context bridging that silently drops IAM credentials on signed requests (see inline comments on `bridgeAmplifyClass.ts` and `internals/server.ts`). Everything else in the migration — the public API overloads, `publicApis`/`transferHandler` updates, and test coverage — looks correct.

packages/api-rest/src/utils/bridgeAmplifyClass.ts
bobbor: Confirmed and fixed in f49e45370 — the bridge output is now branded via `Object.defineProperty(resolved, AMPLIFY_CONTEXT_BRAND, ...)` exactly like `createMockAmplifyContext`/`configure()`, so `internalPost`'s `isAmplifyContext` check recognizes an already-bridged context and double-bridging can't occur. Verified the failure mode both directions with the new e2e test: with the branding stripped, the legacy-ContextSpec request falls through to `unauthenticatedHandler` (your exact scenario); branded, it's signed. Good catch — this was introduced when the internals widening added a second brand checkpoint downstream of `resolveServerContext`.

packages/api-rest/src/internals/server.ts
bobbor: Added in f49e45370: `__tests__/internals/server.e2e.test.ts` mirrors the apis/server e2e style — real `createAmplifyServerContext` + spied `Auth.fetchAuthSession`, driving internals `post` through the actual bridging path and asserting the outgoing request is IAM-signed (plus a branded direct-AmplifyContext case). It fails without the branding fix and passes with it.

packages/api-rest/src/utils/bridgeAmplifyClass.ts
bobbor: Done in f49e45370 — `libraryOptions` is now a live getter like `resourcesConfig`, with a comment noting configure() reassigns it wholesale. Note storage's `resolveServerContext` (PR #14874) snapshots `libraryOptions` the same way — worth a small follow-up there.

osama-rizk (approved):

soberm (approved):

PR #14915, opened 2026-08-19
feat(notifications): migrate notifications APIs to AmplifyContext

bobbor:
#### Description of changes

B-notifications split of the v6 context migration (Task 6 of the migration plan): migrates `@aws-amplify/notifications` from the `Amplify.getConfig()` / singleton `fetchAuthSession` pattern to explicit `AmplifyContext`, with backward-compatible overloads.

- **In-app messaging (Pinpoint)**: `identifyUser`, `syncMessages`, `initializeInAppMessaging` gain `fn(input)` / `fn(ctx, input)` overloads (Pattern 1/2, variadic impl via `resolveCtxArgs`). Event-listener-only APIs (`onMessage*`, `dispatchEvent`, `clearMessages`, `setConflictHandler`) are deliberately not overloaded — they have no config/credential dependency.
- **Push notifications (Pinpoint + Amazon Connect Customer Profiles, incl. #14866 scope)**: `identifyUser`, `initializePushNotifications`, `registerDevice`, `removeDevice` and all `.native` variants gain overloads; web not-supported stubs use the canonical variadic signature.
- **Utils (Pattern 4)**: all six singleton read sites (`resolveConfig`/`resolveCredentials` across the three providers) now read `ctx.resourcesConfig` and call `ctx.fetchAuthSession()`; `helpers.ts`, `createMessageEventRecorder.ts`, `signedFetch.ts`, `identifyUserInternal.ts` take `ctx` as first param. Zero direct singleton reads remain in `src/`.
- **Types**: `types/apis.ts` in pinpoint, customer-profiles, and shared expose both call signatures; the `deprecatePinpoint` wrapper is typed generically so overloads survive wrapping.
- **Tests**: new `__tests__/testUtils/createMockAmplifyContext.ts` (branded factory matching `packages/auth`); ~28 test files migrated off `@aws-amplify/core` getConfig/fetchAuthSession mocking to `setGlobalContext`/`clearGlobalContext` with cleanup; explicit `(ctx, ...)` overload coverage added.

No server wrappers (client-only package). No changeset (feature-branch target).

#### Issue #, if available

Part of the v6 context migration (base: Split A #14805, reference: B-auth #14836).

- `yarn turbo run build --filter=@aws-amplify/notifications` — exit 0
- `npx eslint '**/*.{ts,tsx}'` in `packages/notifications` — 0 errors
- `yarn turbo run test --filter=@aws-amplify/notifications` — 61/61 suites, 263/263 tests passing (includes new context-overload tests)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/identifyUser.test.ts
+++ b/packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/identifyUser.test.ts
@@ -131,6 +131,8 @@ describe('InAppMessaging Pinpoint Provider API: identifyUser', () => {
 			},
 		};
 		await identifyUser(ctx, input);
+		expect(mockResolveConfig).toHaveBeenCalledWith(ctx);
+		expect(mockResolveCredentials).toHaveBeenCalledWith(ctx);
 		expect(mockUpdateEndpoint).toHaveBeenCalledWith({
 			...input,
 			...credentials,
--- a/packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/identifyUser.test.ts
+++ b/packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/identifyUser.test.ts
@@ -2,6 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { updateEndpoint } from '@aws-amplify/core/internals/providers/pinpoint';
+import {
+	clearGlobalContext,
+	setGlobalContext,
+} from '@aws-amplify/core/internals/utils';
 
 import {
 	identifyUser,
@@ -15,6 +19,7 @@ import {
 	resolveCredentials,
 } from '../../../../../src/inAppMessaging/providers/pinpoint/utils';
 import { IdentifyUserInput } from '../../../../../src/inAppMessaging/providers/pinpoint/types';
+import { createMockAmplifyContext } from '../../../../testUtils/createMockAmplifyContext';
 
 jest.mock('@aws-amplify/core/internals/providers/pinpoint');
 jest.mock('../../../../../src/inAppMessaging/providers/pinpoint/utils');
@@ -37,10 +42,15 @@ describe('InAppMessaging Pinpoint Provider API: identifyUser', () => {
 	const mockResolveCredentials = resolveCredentials as jest.Mock;
 
 	beforeAll(() => {
+		setGlobalContext(createMockAmplifyContext());
 		initializeInAppMessaging();
 		mockgetInAppMessagingUserAgentString.mockReturnValue(userAgentValue);
 		mockResolveConfig.mockReturnValue(config);
 		mockResolveCredentials.mockResolvedValue(credentials);
+	});
+
+	afterAll(() => {
+		clearGlobalContext();
 	});
 
 	beforeEach(() => {
@@ -101,5 +111,33 @@ describe('InAppMessaging Pinpoint Provider API: identifyUser', () => {
 			userProfile: {},
 		};
 		await expect(identifyUser(input)).rejects.toBeDefined();
+	});
+
+	it('passes through parameters when called with explicit context', async () => {
+		const ctx = createMockAmplifyContext({
+			Notifications: {
+				InAppMessaging: { Pinpoint: { appId: 'app-id', region: 'region' } },
+			},
+		});
+		const input: IdentifyUserInput = {
+			userId: 'user-id',
+			userProfile: {
+				customProperties: {
+					hobbies: ['biking', 'climbing'],
+				},
+				email: 'email',
+				name: 'name',
+				plan: 'plan',
+			},
+		};
+		await identifyUser(ctx, input);
+		expect(mockUpdateEndpoint).toHaveBeenCalledWith({
+			...input,
+			...credentials,
+			...config,
+			channelType: CHANNEL_TYPE,
+			category: CATEGORY,
+			userAgentValue,
+		});
 	});
 });
--- a/packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/interactionEvents.test.ts
+++ b/packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/interactionEvents.test.ts
@@ -1,6 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { setGlobalContext } from '@aws-amplify/core/internals/utils';
+
 import { inAppMessages } from '../../../../testUtils/data';
 import {
 	addEventListener,
@@ -14,6 +16,7 @@ import {
 	onMessageDisplayed,
 	onMessageReceived,
 } from '../../../../../src/inAppMessaging/providers/pinpoint/apis';
+import { createMockAmplifyContext } from '../../../../testUtils/createMockAmplifyContext';
 
 jest.mock('../../../../../src/eventListeners');
 
@@ -23,6 +26,7 @@ const mockAddEventListener = addEventListener as jest.Mock;
 describe('Interaction events', () => {
 	const handler = jest.fn();
 	beforeAll(() => {
+		setGlobalContext(createMockAmplifyContext());
 		initializeInAppMessaging();
 	});
 	it('can be listened to by onMessageReceived', () => {
--- a/packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/interactionEvents.test.ts
+++ b/packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/interactionEvents.test.ts
@@ -1,7 +1,10 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { setGlobalContext } from '@aws-amplify/core/internals/utils';
+import {
+	clearGlobalContext,
+	setGlobalContext,
+} from '@aws-amplify/core/internals/utils';
 
 import { inAppMessages } from '../../../../testUtils/data';
 import {
@@ -28,6 +31,10 @@ describe('Interaction events', () => {
 	beforeAll(() => {
 		setGlobalContext(createMockAmplifyContext());
 		initializeInAppMessaging();
+	});
+
+	afterAll(() => {
+		clearGlobalContext();
 	});
 	it('can be listened to by onMessageReceived', () => {
 		onMessageReceived(handler);
--- a/packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/syncMessages.test.ts
+++ b/packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/syncMessages.test.ts
@@ -134,6 +134,8 @@ describe('syncMessages', () => {
 		});
 		await syncMessages(ctx);
 
+		expect(mockResolveConfig).toHaveBeenCalledWith(ctx);
+		expect(mockResolveCredentials).toHaveBeenCalledWith(ctx);
 		expect(mockDefaultStorage.setItem).toHaveBeenCalledWith(
 			expect.stringContaining(STORAGE_KEY_SUFFIX),
 			JSON.stringify(simpleInAppMessages),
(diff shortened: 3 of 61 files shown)


osama-rizk (commented):
Clean, type-safe migration; the native-interop tests are the strongest in the set. One functional finding on the listener-registering init APIs (inline) now that reconfigure is supported, plus a stub-consistency nit. Recommend: approve-with-nits once the init-API capture is addressed.

packages/notifications/src/pushNotifications/providers/pinpoint/apis/initializePushNotifications.native.ts
```diff
@@ -40,18 +43,40 @@ const logger = new ConsoleLogger('Notifications.PushNotification');
...
+	const [ctx] = resolveCtxArgs<[]>(args);
+
 	if (isInitialized()) {
 		logger.info('Push notifications have already been enabled');
 
 		return;
 	}
-	addNativeListeners();
-	addAnalyticsListeners();
+	addNativeListeners(ctx);
```
osama-rizk: These capture the context once at init and close over it for the native/auth/analytics listeners; the old code read config/creds lazily per event. Given reconfigure is supported, after a later `Amplify.configure()` the token listener, analytics recorders, and Hub sign-in re-registration all keep firing against the stale captured ctx (events to the old appId, device registration signed via the old session). One-shot APIs are fine (they re-resolve per call). Recommend resolving per event/emit for the long-lived listener paths, or scoping the reconfigure contract explicitly.

Same pattern applies to `customer-profiles/apis/initializePushNotifications.native.ts` and to `inAppMessaging/providers/pinpoint/apis/initializeInAppMessaging.ts` (ctx captured for the `recordAnalyticsEvent` closures).

packages/notifications/src/pushNotifications/providers/customer-profiles/apis/registerDevice.ts
```diff
@@ -12,6 +16,15 @@ import { RegisterDevice } from '../types';
...
+export async function registerDevice(input: RegisterDeviceInput): Promise<void>;
+/**
+ * @param ctx - The {@link AmplifyContext} to use for config and credentials.
+ */
+export async function registerDevice(
+	ctx: AmplifyContext,
+	input: RegisterDeviceInput,
+): Promise<void>;
+export async function registerDevice(...args: any[]): Promise<void> {
+	resolveCtxArgs<[RegisterDeviceInput]>(args);
```
osama-rizk, line 28: This stub changed from a sync `throw` to an `async function` that rejects, while its siblings (`removeDevice`, both `initializePushNotifications`) stayed sync throws. A web caller doing `try { registerDevice(...) }` without `await` used to catch it; now it escapes as an unhandled rejection. The async form may be the better target — just make the not-supported stubs consistent. (It also calls `resolveCtxArgs(args)` then discards the result, which the siblings don't.)

bobbor: Confirmed — registerDevice is the only web stub that rejects asynchronously (siblings throw synchronously), and the resolveCtxArgs call result is unused. Agreed the async/rejecting form is probably the right end state for promise-returning APIs, so rather than churn all four stubs in this PR we'll align them (and drop the dead resolveCtxArgs call) in a follow-up.

packages/notifications/src/pushNotifications/providers/pinpoint/apis/initializePushNotifications.native.ts
bobbor: Good catch — confirmed and fixed in 9aa661556. Validation detail: the global context is a frozen snapshot (configure() creates a new ctx and swaps it via setGlobalContext), so config reads via a captured ctx do go stale; ctx.fetchAuthSession is a live closure over the singleton's Auth, so credentials were never affected — the staleness was config-only (appId/region).

Fix: all three init APIs (both initializePushNotifications.native variants + initializeInAppMessaging) now build a resolveCtx closure — an explicitly passed context stays pinned (caller owns its lifecycle), while legacy no-ctx callers get getGlobalContext() resolved per event in the token, Hub auth, and analytics listener paths. Covered by reconfigure regression tests per API (swap global ctx A→B mid-stream, assert listeners resolve B by identity) plus explicit-ctx pinning tests. 64 suites / 270 tests green.

osama-rizk (approved):

packages/notifications/src/pushNotifications/providers/pinpoint/apis/identifyUser.ts
```diff
@@ -60,6 +64,15 @@ import { IdentifyUser, IdentifyUserInput } from '../types';
...
+export async function identifyUser(input: IdentifyUserInput): Promise<void>;
+/**
+ * @param ctx - The {@link AmplifyContext} to use for config and credentials.
+ */
+export async function identifyUser(
+	ctx: AmplifyContext,
+	input: IdentifyUserInput,
+): Promise<void>;
+export async function identifyUser(...args: any[]): Promise<void> {
+	resolveCtxArgs<[IdentifyUserInput]>(args);
```
soberm, line 76: [minor] Same dead-call pattern as the one flagged on `registerDevice.ts` (customer-profiles): `resolveCtxArgs` only runs here for its validation side effect, the result gets discarded, and the stub throws unconditionally right after anyway. Probably worth folding into that same follow-up cleanup.

packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/interactionEvents.test.ts
```diff
@@ -23,6 +26,7 @@ const mockAddEventListener = addEventListener as jest.Mock;
 describe('Interaction events', () => {
 	const handler = jest.fn();
 	beforeAll(() => {
+		setGlobalContext(createMockAmplifyContext());
```
soberm, line 32: [minor] This is the one test file in the PR that calls `setGlobalContext` without a matching `clearGlobalContext` (it's not even imported here). Probably harmless since Jest isolates modules per file, but it breaks the pattern every other suite in this PR follows, so worth adding an `afterAll` for consistency.

packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/identifyUser.test.ts
```diff
@@ -102,4 +112,32 @@ describe('InAppMessaging Pinpoint Provider API: identifyUser', () => {
...
+			userProfile: {
+				customProperties: {
+					hobbies: ['biking', 'climbing'],
+				},
+				email: 'email',
+				name: 'name',
+				plan: 'plan',
+			},
+		};
+		await identifyUser(ctx, input);
```
soberm, line 133: [minor] This doesn't really prove `ctx` gets forwarded. `resolveConfig`/`resolveCredentials` are mocked to fixed values regardless of their argument, so the assertion below would pass identically even if `identifyUser` ignored `ctx` and fell back to the global context. Adding `expect(mockResolveConfig).toHaveBeenCalledWith(ctx)` and `expect(mockResolveCredentials).toHaveBeenCalledWith(ctx)` would actually pin down the thing the test name claims to cover.

packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/syncMessages.test.ts
```diff
@@ -115,4 +125,18 @@ describe('syncMessages', () => {
...
 		);
 	});
+
+	it('Gets in-app messages and stores them when called with explicit context', async () => {
+		const ctx = createMockAmplifyContext({
+			Notifications: {
+				InAppMessaging: { Pinpoint: { appId: 'app-id', region: 'region' } },
+			},
+		});
+		await syncMessages(ctx);
```
soberm, line 135: [minor] Same gap as the `identifyUser` explicit-context test: `resolveConfig`/`resolveCredentials` are stubbed to fixed values, so this passes whether or not `ctx` was actually threaded through. `expect(mockResolveConfig).toHaveBeenCalledWith(ctx)` (and the same for credentials) would make it a real regression check.

packages/notifications/src/pushNotifications/providers/pinpoint/utils/createMessageEventRecorder.ts
```diff
@@ -20,15 +20,22 @@ const logger = new ConsoleLogger('PushNotification.recordMessageEvent');
...
 export const createMessageEventRecorder =
 	(
+		ctxOrGetter: AmplifyContext | (() => AmplifyContext),
 		event: PinpointMessageEvent,
 		callback?: () => void,
 	): OnPushNotificationMessageHandler =>
 	async message => {
-		const { credentials } = await resolveCredentials();
-		const { appId, region } = resolveConfig();
+		const ctx = typeof ctxOrGetter === 'function' ? ctxOrGetter() : ctxOrGetter;
```
soberm, line 36: [question] This getter branch is what the reconfigure fix relies on for the analytics/message-event listeners, but `createMessageEventRecorder.test.ts` only ever passes a resolved `AmplifyContext` (`mockCtx`), never a getter, and the pinpoint reconfigure suite only covers the token/native-listener path, not `addAnalyticsListeners`. Could be worth a test that passes a getter returning different contexts across two calls, to confirm it re-resolves each time rather than caching the first result.

packages/notifications/src/inAppMessaging/providers/pinpoint/apis/initializeInAppMessaging.ts
```diff
@@ -26,7 +35,24 @@ import { dispatchEvent } from './dispatchEvent';
  * initializeInAppMessaging();
  * ```
  */
-export function initializeInAppMessaging(): void {
+export function initializeInAppMessaging(): void;
+/**
+ * @param ctx - The {@link AmplifyContext} to use for config and credentials.
```
soberm, line 42: [question] Since `isInitialized()` gates the whole body, calling `initializeInAppMessaging(ctx)` after a prior no-arg call already ran is a silent no-op, the passed `ctx` never gets pinned and you just get the "already enabled" log. Might be worth a callout in the docs here, since someone reaching for this overload specifically to pin a context could reasonably expect a second call to take effect.

packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/identifyUser.test.ts
bobbor: Done in 2f920b0fd — added toHaveBeenCalledWith(ctx) identity assertions for both resolveConfig and resolveCredentials.

packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/syncMessages.test.ts
bobbor: Done in 2f920b0fd — same identity assertions added here.

packages/notifications/src/pushNotifications/providers/pinpoint/utils/createMessageEventRecorder.ts
bobbor: Good call — added a getter re-resolution test in 2f920b0fd: getter returns ctxA then ctxB across two recorded events, asserting the downstream utils received each in turn (proves per-call resolution, no caching).

packages/notifications/__tests__/inAppMessaging/providers/pinpoint/apis/interactionEvents.test.ts
bobbor: Done in 2f920b0fd — added the clearGlobalContext import and afterAll cleanup to match the rest of the PR.

packages/notifications/src/pushNotifications/providers/pinpoint/apis/identifyUser.ts
bobbor: Agreed — folding this into the same stub-cleanup follow-up already promised on the customer-profiles registerDevice thread (align all not-supported stubs and drop the dead resolveCtxArgs calls).

packages/notifications/src/inAppMessaging/providers/pinpoint/apis/initializeInAppMessaging.ts
bobbor: Fair point — documented in 2f920b0fd with an @remarks note: initialization runs once per session; subsequent calls are no-ops, so a context passed on a later call won't replace the one captured at first init. Kept as docs-only since changing it would alter the established initialize-once semantics; if we want re-init or ctx re-pinning that feels like a deliberate design change for Phase C.

osama-rizk (approved):

soberm (approved):

PR #14916, opened 2026-08-19
feat(geo): migrate Geo to AmplifyContext

bobbor:
## Description

B-geo split (Task 4) of the v6 context migration. Migrates `@aws-amplify/geo` from the Amplify singleton to the explicit `AmplifyContext` pattern (Pattern 7, class-based):

- `GeoClass` stores the context; constructor accepts an optional `AmplifyContext` and the module-level `Geo` singleton lazily falls back to the global context on first use (`isAmplifyContext`/`getGlobalContext`) — fully backward compatible.
- `AmazonLocationServiceProvider` accepts a context; config is read from `ctx.resourcesConfig.Geo` and credentials via `ctx.fetchAuthSession()` (direct `fetchAuthSession` import removed).
- `index.ts` additionally exports `GeoClass`.
- Tests rewritten to use a branded `createMockAmplifyContext` factory + `setGlobalContext`/`clearGlobalContext` with cleanup — no `Amplify.getConfig` mocking.

No server wrappers (geo has none). No changeset (feature-branch target).

## Validation

- `yarn turbo run build --filter=@aws-amplify/geo` — pass
- `eslint '**/*.{ts,tsx}'` in packages/geo — pass
- `yarn turbo run test --filter=@aws-amplify/geo` — 113 tests / 3 suites pass, ts-coverage threshold met
- `yarn turbo run test:size --filter=@aws-amplify/geo` — 40.75 kB vs 46 kB limit

## Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/geo/__tests__/Geo.test.ts
+++ b/packages/geo/__tests__/Geo.test.ts
@@ -669,4 +669,127 @@ describe('Geo', () => {
 			);
 		});
 	});
+
+	describe('lazy context resolution', () => {
+		afterEach(() => {
+			clearGlobalContext();
+		});
+
+		test('reconfigure honored: global context change is picked up by the same GeoClass/provider instance', async () => {
+			// Restore the LocationClient mock for this test
+			const sendMock = jest.fn(async command => {
+				if (command instanceof SearchPlaceIndexForTextCommand) {
+					return {
+						Results: [{ Place: TestPlacePascalCase }],
+					};
+				}
+			});
+			LocationClient.prototype.send = sendMock;
+
+			const ctxA = createMockAmplifyContext(awsConfigGeoV4);
+			(ctxA.fetchAuthSession as jest.Mock).mockResolvedValue({ credentials });
+
+			setGlobalContext(ctxA);
+			const geo = new GeoClass();
+
+			// First operation uses ctxA's config
+			const resultsA = await geo.searchByText('test');
+			expect(resultsA).toEqual([testPlaceCamelCase]);
+			expect(ctxA.fetchAuthSession).toHaveBeenCalled();
+
+			// Reconfigure with a different context (different search index)
+			const altConfig = {
+				...awsConfigGeoV4,
+				Geo: {
+					LocationService: {
+						...awsConfigGeoV4.Geo!.LocationService,
+						searchIndices: {
+							items: ['altSearchIndex'],
+							default: 'altSearchIndex',
+						},
+					},
+				},
+			};
+			const ctxB = createMockAmplifyContext(altConfig);
+			(ctxB.fetchAuthSession as jest.Mock).mockResolvedValue({ credentials });
+
+			setGlobalContext(ctxB);
+
+			// Same geo instance's next operation should use ctxB's config
+			const resultsB = await geo.searchByText('test2');
+			expect(resultsB).toEqual([testPlaceCamelCase]);
+			expect(ctxB.fetchAuthSession).toHaveBeenCalled();
+
+			// Second call (index 1) should use ctxB's search index
+			const { input } = sendMock.mock.calls[1][0];
+			expect(input).toHaveProperty('IndexName', 'altSearchIndex');
+		});
+
+		test('explicit ctx pinned: GeoClass constructed with explicit ctx ignores global context changes', async () => {
+			// Restore the LocationClient mock for this test
+			LocationClient.prototype.send = jest.fn(async command => {
+				if (command instanceof SearchPlaceIndexForTextCommand) {
+					return {
+						Results: [{ Place: TestPlacePascalCase }],
+					};
+				}
+			});
+
+			const explicitCtx = createMockAmplifyContext(awsConfigGeoV4);
+			(explicitCtx.fetchAuthSession as jest.Mock).mockResolvedValue({
+				credentials,
+			});
+
+			const geo = new GeoClass(explicitCtx);
+
+			// Set a different global context
+			const altConfig = {
+				...awsConfigGeoV4,
+				Geo: {
+					LocationService: {
+						...awsConfigGeoV4.Geo!.LocationService,
+						searchIndices: {
+							items: ['globalSearchIndex'],
+							default: 'globalSearchIndex',
+						},
+					},
+				},
+			};
+			const globalCtx = createMockAmplifyContext(altConfig);
+			(globalCtx.fetchAuthSession as jest.Mock).mockResolvedValue({
+				credentials,
+			});
+			setGlobalContext(globalCtx);
+
+			// Operation should still use the explicit ctx
+			const results = await geo.searchByText('test');
+			expect(results).toEqual([testPlaceCamelCase]);
+			expect(explicitCtx.fetchAuthSession).toHaveBeenCalled();
+			expect(globalCtx.fetchAuthSession).not.toHaveBeenCalled();
+		});
+
+		test('no eager throw: constructing GeoClass and addPluggable before configure does not throw', () => {
+			clearGlobalContext();
+
+			// These should not throw even though global context is not set
+			expect(() => new GeoClass()).not.toThrow();
+			expect(() => new AmazonLocationServiceProvider()).not.toThrow();
+
+			const geo = new GeoClass();
+			const provider = new AmazonLocationServiceProvider();
+			expect(() => {
+				geo.addPluggable(provider);
+			}).not.toThrow();
+		});
+
+		test('no eager throw: operation throws when global context is not set', async () => {
+			clearGlobalContext();
+
+			const geo = new GeoClass();
+
+			await expect(geo.searchByText('test')).rejects.toThrow(
+				'No AmplifyContext available',
+			);
+		});
+	});
 });
--- a/packages/geo/__tests__/Geo.test.ts
+++ b/packages/geo/__tests__/Geo.test.ts
@@ -1,6 +1,5 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { Amplify, fetchAuthSession } from '@aws-amplify/core';
 import {
 	GetPlaceCommand,
 	LocationClient,
@@ -8,6 +7,11 @@ import {
 	SearchPlaceIndexForSuggestionsCommand,
 	SearchPlaceIndexForTextCommand,
 } from '@aws-sdk/client-location';
+import { AmplifyContext } from '@aws-amplify/core';
+import {
+	clearGlobalContext,
+	setGlobalContext,
+} from '@aws-amplify/core/internals/utils';
 import camelcaseKeys from 'camelcase-keys';
 
 import { GeoClass } from '../src/Geo';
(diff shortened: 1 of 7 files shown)


packages/geo/src/Geo.ts
```diff
@@ -33,17 +38,62 @@ export class GeoClass {
...
+	 *
+	 * Note: The provider captures the resolved context at first use.
+	 * A full reconfiguration with a NEW context object requires constructing
+	 * a new GeoClass instance (matches previous singleton behavior).
+	 * @private
+	 */
+	private _initialize(): void {
+		this._pluggables = [];
+		const ctx = this._getCtx();
+		const amplifyConfig = ctx.resourcesConfig ?? {};
```
osama-rizk: Two things here. (1) The provider captures one frozen `ctx` at first use and every op reads `this._ctx.resourcesConfig` — a snapshot. `Amplify.configure()` builds a *new* frozen context and replaces the global, but the Geo singleton keeps its old reference, so `Amplify.configure(B)` after first use is silently ignored (`searchByText` still uses config A). The old `_refreshConfig` read `Amplify.getConfig()` live per op, so it picked up reconfigure. (2) The inline comment claiming this "matches previous singleton behavior" is therefore inaccurate. Since reconfigure/tenant-switch is supported, please re-resolve the context per operation (or via the global each time) rather than caching one, and correct the comment.

packages/geo/src/providers/location-service/AmazonLocationServiceProvider.ts
```diff
@@ -70,13 +75,20 @@ export class AmazonLocationServiceProvider implements GeoProvider {
...
 	 * @param {Object} config - Configuration object for Geo
+	 * @param {AmplifyContext} ctx - The AmplifyContext to use for auth and config
 	 */
-	constructor(config?: GeoConfig) {
+	constructor(config?: GeoConfig, ctx?: AmplifyContext) {
 		this._config = config || {};
+		if (isAmplifyContext(ctx)) {
+			this._ctx = ctx;
+		} else {
+			this._ctx = getGlobalContext();
```
osama-rizk: The constructor now eagerly calls `getGlobalContext()`, which throws if `configure()` hasn't run. With `GeoClass` newly public-exported, `new GeoClass(); g.addPluggable(...)` before `configure()` now throws a generic "No AmplifyContext available" where it previously deferred. Lower blast radius (the provider isn't re-exported from the package entry), but worth a conscious, cross-category-consistent choice — the `GeoClass` fallback is lazy while the provider's is eager, which is the asymmetry behind this.

osama-rizk (commented):
Mechanically consistent migration and the tests improved (real `test()` for `getModuleName`, added `deleteGeofences` coverage). One blocker: config is now captured as a snapshot, which drops post-`configure()` changes — and reconfigure is a supported scenario. Details inline. Recommend: needs-changes.

packages/geo/src/Geo.ts
bobbor: Good catch — validated and fixed in 5b5421307. The provider no longer captures a context: it stores only an explicitly-passed ctx and otherwise resolves `getGlobalContext()` fresh at every use site (config read + credential fetch), so a later `Amplify.configure()` is honored again, matching the old live `getConfig()` behavior. The inaccurate comment is corrected: explicit ctx = pinned by design, global fallback = re-resolved per operation. Added regression tests for both (reconfigure honored on the same instance; explicit ctx pinned after global changes).

packages/geo/src/providers/location-service/AmazonLocationServiceProvider.ts
bobbor: Fixed in 5b5421307 together with the snapshot issue — the constructor no longer resolves any context, which removes the eager throw and the lazy/eager asymmetry: both GeoClass and the provider now defer global-context resolution to operation time. `new GeoClass()` / `addPluggable()` before `configure()` work again; only actual operations throw `No AmplifyContext available` when unconfigured. Regression test added for the construct-before-configure path.

osama-rizk (approved):

packages/geo/src/Geo.ts
```diff
@@ -33,19 +37,61 @@ export class GeoClass {
 	 * @private
 	 */
 	private _config?: GeoConfig;
-	private _pluggables: GeoProvider[];
+	private _pluggables: GeoProvider[] | undefined;
+	private _ctx: AmplifyContext | undefined;
 
-	constructor() {
+	constructor(ctx?: AmplifyContext) {
```
soberm, line 43: [question] One thing I'm curious about: the sibling migrations (storage's `getUrl(ctx, input)`, interactions' `send(ctx, input)`, analytics' `record(ctx, input)`) all thread `ctx` through `resolveCtxArgs` on every call, so a caller can point a single call at a specific context without building a new object. Here `ctx` only exists at construction time, so the exported `Geo` singleton can never target an explicit context per call — you'd need `new GeoClass(ctx)` (and you'd lose whatever pluggables were registered on the shared instance via `addPluggable`).

Is that a deliberate scope decision for Geo (no server wrapper, per the PR description), or should `searchByText`/`getGeofence`/etc. eventually get the same `(ctx, input)` overload treatment for consistency with the rest of the migration series?

packages/geo/src/Geo.ts
```diff
@@ -33,19 +37,61 @@ export class GeoClass {
 	 * @private
 	 */
 	private _config?: GeoConfig;
-	private _pluggables: GeoProvider[];
+	private _pluggables: GeoProvider[] | undefined;
```
soberm, line 40: [minor] Since `_pluggables` is now optional, every read after `_ensureInitialized()` needs a `!` (lines 114, 124, 139). `AmazonLocationServiceProvider` solves basically the same shape of problem a few files over with a private getter (`_ctx`) that lazily resolves and always returns a non-optional value — might be worth doing the same here, e.g. a `_pluggablesList` getter that calls `_ensureInitialized()` internally and returns `GeoProvider[]`. Would drop the three `!` assertions and keep both classes consistent with each other.

packages/geo/src/Geo.ts
```diff
@@ -33,19 +37,61 @@ export class GeoClass {
...
+	private _ctx: AmplifyContext | undefined;
 
-	constructor() {
+	constructor(ctx?: AmplifyContext) {
 		this._config = undefined;
+		this._pluggables = undefined;
+
+		if (isAmplifyContext(ctx)) {
```
soberm, line 47: [question] `isAmplifyContext(ctx)` silently treats anything that isn't a branded context — including a non-context object someone accidentally passes — the same as "no ctx was given", and falls back to the global context. For a class that's now public API, that seems like it could mask a real bug: caller thinks they're pinning an explicit context, a typo or stale reference means the brand check fails, and it quietly starts reading the global config/credentials instead of throwing.

`resolveCtxArgs` over in core has a narrower version of this same problem but at least special-cases `undefined` explicitly and throws for it. Worth throwing here for a defined-but-invalid `ctx` rather than silently falling through? Same question applies to `AmazonLocationServiceProvider`'s constructor (`AmazonLocationServiceProvider.ts` line 104).

packages/geo/__tests__/Geo.test.ts
```diff
@@ -123,22 +126,28 @@ describe('Geo', () => {
...
+
+	describe('global context fallback', () => {
+		test('falls back to global context when no ctx is provided', () => {
+			setGlobalContext(mockCtx);
 			const geo = new GeoClass();
+
 			expect(geo.getPluggable('AmazonLocationService')).toBeInstanceOf(
 				AmazonLocationServiceProvider,
 			);
+			clearGlobalContext();
```
soberm, line 144: [minor] This clears the global context inline at the end of the test. If the `expect` above it throws, `clearGlobalContext()` never runs and `mockCtx` leaks into whatever test runs next. The `lazy context resolution` block near the bottom of this file handles it more safely with a scoped `afterEach(() => clearGlobalContext())` — might be worth the same pattern here.

packages/geo/src/index.ts
```diff
@@ -1,4 +1,4 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-export { Geo } from './Geo';
+export { Geo, GeoClass } from './Geo';
```
soberm, line 3: [nit] Now that `GeoClass` is exported from the package root, it's directly constructible by consumers for the first time. Might be worth a class-level doc comment on `GeoClass` (in `Geo.ts`) explaining when you'd reach for `new GeoClass(ctx)` versus the default `Geo` singleton, since that's now part of the public contract rather than an implementation detail.

packages/geo/__tests__/testUtils/mockAmplifyContext.ts
```diff
@@ -0,0 +1,31 @@
...
+	AmplifyContext,
+	ResourcesConfig,
+} from '@aws-amplify/core';
+
+/**
+ * Creates a mock AmplifyContext for testing.
+ * Accepts partial/incomplete configs to test error paths.
+ */
+export function createMockAmplifyContext(
+	resourcesConfig?: ResourcesConfig | object,
```
soberm, line 15: [nit] `ResourcesConfig | object` is effectively just `object` for type-checking purposes, since `object` swallows the more specific type — so this doesn't really constrain callers. `Partial<ResourcesConfig>` would keep the "partial configs for testing error paths" intent from the docstring above while still giving useful autocomplete on the config shape.

packages/geo/src/Geo.ts
bobbor: Deliberate, per the migration plan: class-based categories (geo, pubsub, predictions) use Pattern 7 — ctx is bound at construction and the instance is the context binding. Per-call targeting is served by constructing `new GeoClass(ctx)` (now exported) instead of overloading every method; function-based categories use `resolveCtxArgs` because they have no instance to bind to. If we want per-call overloads on class methods too for cross-category symmetry, that's a Phase C design question — happy to raise it there.

packages/geo/src/Geo.ts
bobbor: Agreed it'd read better — we'll provide a follow-up applying the lazy-getter pattern here rather than churning this PR further.

packages/geo/src/Geo.ts
bobbor: Valid concern, and it's the same failure mode as mis-ordered ctx args in the function-based categories — which we're fixing centrally with a runtime guard in `resolveCtxArgs` (see #14912). We'll extend that follow-up to constructor paths (throw on a defined-but-unbranded ctx argument) so all categories behave consistently, rather than adding a one-off check here.

packages/geo/__tests__/Geo.test.ts
bobbor: Agreed — we'll provide a follow-up moving this to a scoped afterEach like the lazy-context block does.

packages/geo/src/index.ts
bobbor: Fair point — we'll provide a follow-up adding a class-level doc comment covering when to use `new GeoClass(ctx)` vs the `Geo` singleton.

packages/geo/__tests__/testUtils/mockAmplifyContext.ts
bobbor: Agreed, `Partial<ResourcesConfig>` keeps the intent with better autocomplete — we'll provide a follow-up (it also touches the copies of this test util in the sibling packages, so we'll do them together).

soberm (approved):

PR #14917, opened 2026-08-21
feat(pubsub): migrate PubSub AWSIoT provider to AmplifyContext

bobbor:
## Description

B-pubsub split (Task 7, Phase 2) of the v6 context migration. Migrates `@aws-amplify/pubsub`'s `AWSIoT` provider from the module-level `fetchAuthSession` import to the explicit `AmplifyContext` pattern (Pattern 7, class-based):

- `AWSIoT` constructor gains an overload accepting an optional `AmplifyContext` as first argument (`new AWSIoT(ctx, options)`), discriminated via `isAmplifyContext` — plain options objects are never mistaken for a context, so existing `new AWSIoT(options)` callers are fully backward compatible.
- The explicit context is stored only when branded; a private `_ctx` getter lazily resolves `explicitCtx ?? getGlobalContext()` fresh per operation (no constructor-time global-context snapshot, per the correction that landed in #14916).
- WebSocket URL signing in `get endpoint` now uses `this._ctx.fetchAuthSession()`; the direct `fetchAuthSession` import from `@aws-amplify/core` is removed.
- Adds `__tests__/testUtils/mockAmplifyContext.ts` (branded mock factory, same as sibling packages) and explicit-ctx tests: ctx-based signing, missing-credentials error path, and a negative assertion that the global context is not consulted when an explicit ctx is provided. No `Amplify.getConfig` mocking.

Config (region/endpoint) remains constructor-options-based — pubsub reads no `resourcesConfig` category; only the auth call migrates to ctx. No server wrappers (pubsub has none). No changeset (feature-branch target).

Depends on B-auth (#14836) — already merged into the base branch.

## Validation

- `yarn turbo run build --filter=@aws-amplify/pubsub` — pass
- `yarn turbo run test --filter=@aws-amplify/pubsub` — 20 tests / 1 suite pass
- `yarn turbo run lint --filter=@aws-amplify/pubsub` — pass
- `yarn test:size` (packages/pubsub) — IoT 88 B vs 1.2 kB limit, Mqtt 89 B vs 1.07 kB limit

## Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/pubsub/__tests__/PubSub.test.ts
+++ b/packages/pubsub/__tests__/PubSub.test.ts
@@ -18,13 +18,34 @@ jest.mock('@aws-amplify/core', () => ({
 			},
 		};
 	},
+	// Provides the global-ctx fallback for tests that construct providers
+	// without an explicit AmplifyContext (lazy ctx resolution path).
+	// References mockGlobalCtxFetchAuthSession lazily (at call time) so
+	// explicit-ctx tests can assert it is NOT called.
+	getGlobalContext() {
+		return {
+			fetchAuthSession: mockGlobalCtxFetchAuthSession,
+		};
+	},
+}));
+
+const mockGlobalCtxFetchAuthSession = jest.fn(() => ({
+	credentials: {
+		accessKeyId: 'accessKeyId',
+		sessionToken: 'sessionToken',
+		secretAccessKey: 'secretAccessKey',
+		identityId: 'identityId',
+		authenticated: true,
+	},
 }));
 
 import { Reachability } from '@aws-amplify/core/internals/utils';
 import * as Paho from '../src/vendor/paho-mqtt';
 import { ConnectionState, PubSub as IotPubSub, mqttTopicMatch } from '../src';
 import { PubSub as MqttPubSub } from '../src/clients/mqtt';
+import { AWSIoT } from '../src/Providers/AWSIot';
 import { HubConnectionListener } from './helpers';
+import { createMockAmplifyContext } from './testUtils/mockAmplifyContext';
 import { Observable, Observer } from 'rxjs';
 import * as constants from '../src/Providers/constants';
 
@@ -568,5 +589,68 @@ describe('PubSub', () => {
 				spyDisconnect.mockClear();
 			},
 		);
+	});
+
+	describe('AWSIoT with explicit AmplifyContext', () => {
+		test('uses ctx.fetchAuthSession() for URL signing when constructed with explicit ctx', async () => {
+			const mockCtx = createMockAmplifyContext();
+			(mockCtx.fetchAuthSession as jest.Mock).mockResolvedValue({
+				credentials: {
+					accessKeyId: 'ctxAccessKeyId',
+					secretAccessKey: 'ctxSecretAccessKey',
+					sessionToken: 'ctxSessionToken',
+				},
+			});
+
+			const provider = new AWSIoT(mockCtx, {
+				region: 'us-east-1',
+				endpoint: 'wss://iot.test.org:443/mqtt',
+			});
+
+			// Access the protected endpoint getter via bracket notation
+			await provider['endpoint'];
+
+			expect(mockCtx.fetchAuthSession).toHaveBeenCalledTimes(1);
+		});
+
+		test('throws when explicit ctx returns no credentials', async () => {
+			const mockCtx = createMockAmplifyContext();
+			(mockCtx.fetchAuthSession as jest.Mock).mockResolvedValue({
+				credentials: undefined,
+			});
+
+			const provider = new AWSIoT(mockCtx, {
+				region: 'us-east-1',
+				endpoint: 'wss://iot.test.org:443/mqtt',
+			});
+
+			await expect(provider['endpoint']).rejects.toThrow(
+				'No auth session credentials',
+			);
+		});
+
+		test('does not fall back to global fetchAuthSession when explicit ctx is provided', async () => {
+			mockGlobalCtxFetchAuthSession.mockClear();
+			const mockCtx = createMockAmplifyContext();
+			(mockCtx.fetchAuthSession as jest.Mock).mockResolvedValue({
+				credentials: {
+					accessKeyId: 'ctxAccessKeyId',
+					secretAccessKey: 'ctxSecretAccessKey',
+					sessionToken: 'ctxSessionToken',
+				},
+			});
+
+			const provider = new AWSIoT(mockCtx, {
+				region: 'us-east-1',
+				endpoint: 'wss://iot.test.org:443/mqtt',
+			});
+
+			await provider['endpoint'];
+
+			// The global-ctx fetchAuthSession must NOT have been called because
+			// the explicit ctx path is used instead
+			expect(mockCtx.fetchAuthSession).toHaveBeenCalled();
+			expect(mockGlobalCtxFetchAuthSession).not.toHaveBeenCalled();
+		});
 	});
 });
--- a/packages/pubsub/__tests__/PubSub.test.ts
+++ b/packages/pubsub/__tests__/PubSub.test.ts
@@ -649,7 +649,7 @@ describe('PubSub', () => {
 
 			// The global-ctx fetchAuthSession must NOT have been called because
 			// the explicit ctx path is used instead
-			expect(mockCtx.fetchAuthSession).toHaveBeenCalled();
+			expect(mockCtx.fetchAuthSession).toHaveBeenCalledTimes(1);
 			expect(mockGlobalCtxFetchAuthSession).not.toHaveBeenCalled();
 		});
 	});
--- a/packages/pubsub/__tests__/PubSub.test.ts
+++ b/packages/pubsub/__tests__/PubSub.test.ts
@@ -39,7 +39,7 @@ const mockGlobalCtxFetchAuthSession = jest.fn(() => ({
 	},
 }));
 
-import { Reachability } from '@aws-amplify/core/internals/utils';
+import { Reachability, Signer } from '@aws-amplify/core/internals/utils';
 import * as Paho from '../src/vendor/paho-mqtt';
 import { ConnectionState, PubSub as IotPubSub, mqttTopicMatch } from '../src';
 import { PubSub as MqttPubSub } from '../src/clients/mqtt';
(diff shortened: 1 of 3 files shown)


soberm (commented):
Took a pass through this. Overall the approach (constructor overload + a `_ctx` resolver that pins an explicit context but re-resolves the global one lazily) looks solid and the new tests cover the three cases that matter (explicit ctx used, explicit ctx takes priority over global.

Nothing here blocks merging as far as I can tell — the core migration logic and test coverage look correct. The constructor overload style is the main thing I'd want a second opinion on before this pattern gets replicated across other providider migrations in this series.

packages/pubsub/src/Providers/AWSIot.ts
```diff
@@ -13,8 +17,34 @@ export interface AWSIoTOptions extends MqttOptions {
 }
 
 export class AWSIoT extends MqttOverWS {
-	constructor(options: AWSIoTOptions = {}) {
+	private _explicitCtx: AmplifyContext | undefined;
+
+	constructor(options?: AWSIoTOptions);
+	constructor(ctx: AmplifyContext, options?: AWSIoTOptions);
+	constructor(...args: unknown[]) {
```
soberm: [minor] Using `constructor(...args: unknown[])` with a runtime `isAmplifyContext(args[0])` check works, but it means the two declared overloads above aren't actually enforced by the compiler for anything calling the constructor dynamically (spread args, `.apply`, etc). Since `MqttOverWS`'s own constructor still only takes `options`, would a static factory like `AWSIoT.withContext(ctx, options)` be cleaner here? Keeps the public constructor single-shape and avoids the `unknown[]` cast — worth considering especially if this pattern gets copied to the other provider migrations in this series.

packages/pubsub/src/Providers/AWSIot.ts
```diff
@@ -13,8 +17,34 @@ export interface AWSIoTOptions extends MqttOptions {
...
+	 * - Otherwise, the global context is resolved fresh per access so that reconfiguration
+	 *   (setGlobalContext with a new AmplifyContext) is honored across operations.
+	 * @private
+	 */
+	private get _ctx(): AmplifyContext {
+		if (this._explicitCtx) {
+			return this._explicitCtx;
+		}
+
+		return getGlobalContext();
```
soberm, line 48: [question] Nice doc comment above explaining the pinned-vs-re-resolved design. What happens here if `getGlobalContext()` throws or is unset because no explicit ctx was passed and `configure()` hasn't run yet? Is that an existing/expected failure mode from `getGlobalContext()` itself, or should `_ctx` wrap it with a clearer error message specific to this provider?

packages/pubsub/src/Providers/AWSIot.ts
```diff
@@ -13,8 +17,34 @@ export interface AWSIoTOptions extends MqttOptions {
 }
 
 export class AWSIoT extends MqttOverWS {
-	constructor(options: AWSIoTOptions = {}) {
+	private _explicitCtx: AmplifyContext | undefined;
```
soberm: [nit] `_explicitCtx` is only ever set once, in the constructor, and never reassigned — could this be `private readonly _explicitCtx: AmplifyContext | undefined;` to signal intent?

packages/pubsub/__tests__/PubSub.test.ts
```diff
@@ -569,4 +590,67 @@ describe('PubSub', () => {
...
+			const provider = new AWSIoT(mockCtx, {
+				region: 'us-east-1',
+				endpoint: 'wss://iot.test.org:443/mqtt',
+			});
+
+			await provider['endpoint'];
+
+			// The global-ctx fetchAuthSession must NOT have been called because
+			// the explicit ctx path is used instead
+			expect(mockCtx.fetchAuthSession).toHaveBeenCalled();
```
soberm: [minor] `expect(mockCtx.fetchAuthSession).toHaveBeenCalled()` could be `toHaveBeenCalledTimes(1)` here to match the precision of the earlier test and catch an accidental double-call regression.

packages/pubsub/__tests__/testUtils/mockAmplifyContext.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export function createMockAmplifyContext(
+	resourcesConfig?: ResourcesConfig | object,
+): AmplifyContext {
+	const ctx: AmplifyContext = {
+		resourcesConfig: (resourcesConfig || {}) as ResourcesConfig,
+		libraryOptions: {},
+		fetchAuthSession: jest.fn().mockResolvedValue({}),
+		clearCredentials: jest.fn().mockResolvedValue(undefined),
+		getTokens: jest.fn().mockResolvedValue(undefined),
```
soberm, line 22: [question] `getTokens: jest.fn().mockResolvedValue(undefined)` — does this match the real `AmplifyContext.getTokens` signature (looks like it should resolve to `AuthTokens | undefined`, so `undefined` is a valid default)? Just flagging in case a non-undefined default would be useful for other tests that exercise the tokens path.

packages/pubsub/src/Providers/AWSIot.ts
bobbor: Good question — on the no-ctx path this surfaces `getGlobalContext()`'s own unconfigured error, which is the same (accepted) behavior as in every other migrated package (geo, analytics, storage, …). Rather than wrapping it per-provider, there is already a promised core-level follow-up from #14913 (raised by osama-rizk) to re-wrap the `getGlobalContext()` unconfigured error with a clearer typed error centrally in core — pubsub will pick that up automatically. Keeping this provider aligned with the siblings until that lands.

packages/pubsub/__tests__/testUtils/mockAmplifyContext.ts
bobbor: Correct — `AmplifyContext.getTokens` resolves `AuthTokens | undefined`, so `undefined` is a valid (and the most neutral) default. The helper is a 1:1 copy of the sibling packages' testUtil (geo et al.), and pubsub's tests never exercise the tokens path (IoT URL signing only consumes `fetchAuthSession().credentials`). Improvements to this shared mock helper across all package copies (incl. the `Partial<ResourcesConfig>` typing) are tracked as follow-up (5) from #14916 — a richer tokens default can be added there if a package needs it.

packages/pubsub/src/Providers/AWSIot.ts
bobbor: Done in b761366ba.

packages/pubsub/__tests__/PubSub.test.ts
bobbor: Done in b761366ba.

packages/pubsub/src/Providers/AWSIot.ts
bobbor: Good catch on the `unknown[]` impl — addressed in e317eb350 with a middle ground: the implementation signature is now the typed union `(ctxOrOptions?: AmplifyContext | AWSIoTOptions, maybeOptions?: AWSIoTOptions)`. Since `isAmplifyContext` is a type guard, the compiler narrows both branches and all three `as` casts are gone — the body is fully type-checked now, while the two public overloads (and the ctx-first call shape) stay as-is.

On `withContext(ctx, options)`: considered it, but leaning against — (a) every function API in this series uses the ctx-first overload shape (`fn(input)` / `fn(ctx, input)` via `resolveCtxArgs`), and geo (#14916) already shipped ctx-in-constructor, so a static factory would make class packages the one place where 'ctx as first arg' doesn't apply; (b) static factories don't inherit cleanly for subclasses. The residual runtime concern (an *unbranded* ctx-like first arg silently treated as options) is the same mis-ordered/unbranded-ctx issue we're centralizing in the `resolveCtxArgs` guard follow-up from #14912 — the constructor path is covered there.

Proposing the typed-union impl as the replicable pattern for the remaining class-based providers (predictions is the other ctx+options constructor in this series) — flagging for @osama-rizk as the second opinion you asked for.

soberm (approved):

packages/pubsub/__tests__/PubSub.test.ts
```diff
@@ -569,4 +590,67 @@ describe('PubSub', () => {
...
+
+			const provider = new AWSIoT(mockCtx, {
+				region: 'us-east-1',
+				endpoint: 'wss://iot.test.org:443/mqtt',
+			});
+
+			// Access the protected endpoint getter via bracket notation
+			await provider['endpoint'];
+
+			expect(mockCtx.fetchAuthSession).toHaveBeenCalledTimes(1);
```
osama-rizk: This test mocks the ctx to return specific credentials, accesses `endpoint`, then asserts only that `fetchAuthSession` was called once — it never checks those credentials actually reached the signed URL. A future refactor that called `ctx.fetchAuthSession()` and then signed with wrong/stale creds would still pass. Consider asserting the signed URL (or a spied `Signer.signUrl` argument) carries the ctx credentials — closing the loop on the one thing this migration changed.

osama-rizk (commented):
approve-with-nits — clean migration, backward compat structurally sound (the `isAmplifyContext` brand-symbol guard can't misclassify a plain options object), and lazy-ctx correct: signing already resolves per-access inside `get endpoint`, so no stale-cred risk on long-lived subscriptions (verified in `AWSIot.ts` — `fetchAuthSession()` is re-invoked on every endpoint access; even a pinned explicit ctx yields fresh creds because only the ctx object is fixed, not the auth call). One test note inline. Minor: the unconfigured-path error changes to a generic `Error('No AmplifyContext available')` — this is series-wide across the v6-context migration, worth one release note at the epic level rather than per-PR; and the 3 new tests repeat the mockCtx/credentials boilerplate, so a shared helper would cut drift. The `createMockAmplifyContext` `resourcesConfig` param is unused here (fine to keep for cross-package mock parity).

packages/pubsub/__tests__/PubSub.test.ts
bobbor: Good catch — addressed in 6b47570cb: the test now spies `Signer.signUrl` and asserts it receives exactly the ctx credentials (`access_key`/`secret_key`/`session_token` from the mock ctx) plus the expected endpoint and service info, closing the loop from `ctx.fetchAuthSession()` to the signed URL.

bobbor:
Re the two minors from @osama-rizk's review: (1) the generic `Error('No AmplifyContext available')` on the unconfigured path is indeed series-wide — a typed re-wrap of the `getGlobalContext()` error centrally in core is already a tracked follow-up (from #14913), and I agree the release note belongs at the epic level, will fold it into the v6-context epic notes; (2) the mockCtx/credentials test boilerplate — a shared `createMockAmplifyContext` exposed via a core testing entry point is also on the follow-up ledger (from #14916/#14918), pubsub's copy will be replaced when that lands. Thanks for the pass! Note: the review shows as COMMENTED — mind flipping it to an actual Approve so the merge gate reflects it?

osama-rizk (approved):

PR #14918, opened 2026-08-21
feat(predictions): migrate Predictions to AmplifyContext

bobbor:
## Description

Task 8 / B-predictions split of the v6 AmplifyContext migration: PredictionsClass + 3 providers accept optional explicit AmplifyContext via `isAmplifyContext`-guarded constructor, lazy private `ctx` getter falling back to `getGlobalContext()` per operation — no constructor snapshot; config via `ctx.resourcesConfig`, credentials via `ctx.fetchAuthSession()`, storage `getUrl(ctx, ...)` threading in Identify provider; branded `createMockAmplifyContext` testUtil replaces core-module mocks.

## Related

- Depends on storage #14874 (merged)
- Follows geo #14916 pattern incl. lazy-ctx fix

## Validation

- [x] `yarn turbo run build --filter=@aws-amplify/predictions` — green
- [x] `yarn turbo run test --filter=@aws-amplify/predictions` — 4 suites, 53 passed
- [x] lint-staged (eslint) green on commit
- [x] aws-amplify dependent build green

> No changeset — PR targets feature branch `feat/bobbor/v6-context`

---

- [x] Tests added/updated
- [x] By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/packages/predictions/__tests__/Predictions.test.ts
+++ b/packages/predictions/__tests__/Predictions.test.ts
@@ -96,6 +96,23 @@ describe('Predictions test', () => {
 		afterEach(() => {
 			jest.restoreAllMocks();
 		});
+
+		test('all three providers receive the same ctx instance', () => {
+			const ctx = createMockAmplifyContext();
+
+			const predictions = new PredictionsClass(ctx);
+
+			// Access the private providers and verify each received the exact same ctx (identity check)
+			type ProviderWithCtx = { _explicitCtx: unknown };
+			const convertProvider = (predictions as unknown as { convertProvider: ProviderWithCtx }).convertProvider;
+			const identifyProvider = (predictions as unknown as { identifyProvider: ProviderWithCtx }).identifyProvider;
+			const interpretProvider = (predictions as unknown as { interpretProvider: ProviderWithCtx }).interpretProvider;
+
+			expect(convertProvider._explicitCtx).toBe(ctx);
+			expect(identifyProvider._explicitCtx).toBe(ctx);
+			expect(interpretProvider._explicitCtx).toBe(ctx);
+		});
+
 		test('convert delegates to provider constructed with ctx', async () => {
 			const ctx = createMockAmplifyContext();
 			const input: TranslateTextInput = {
--- a/packages/predictions/__tests__/Predictions.test.ts
+++ b/packages/predictions/__tests__/Predictions.test.ts
@@ -113,7 +113,7 @@ describe('Predictions test', () => {
 			expect(interpretProvider._explicitCtx).toBe(ctx);
 		});
 
-		test('convert delegates to provider constructed with ctx', async () => {
+		test('convert delegates to the convert provider', async () => {
 			const ctx = createMockAmplifyContext();
 			const input: TranslateTextInput = {
 				translateText: { source: { text: 'sourceText' } },
@@ -133,7 +133,7 @@ describe('Predictions test', () => {
 			expect(convertSpy).toHaveBeenCalledTimes(1);
 		});
 
-		test('identify delegates to provider constructed with ctx', async () => {
+		test('identify delegates to the identify provider', async () => {
 			const ctx = createMockAmplifyContext();
 			const input: IdentifyTextInput = {
 				text: { source: { key: 'key' }, format: 'PLAIN' },
@@ -157,7 +157,7 @@ describe('Predictions test', () => {
 			expect(identifySpy).toHaveBeenCalledTimes(1);
 		});
 
-		test('interpret delegates to provider constructed with ctx', async () => {
+		test('interpret delegates to the interpret provider', async () => {
 			const ctx = createMockAmplifyContext();
 			const input: InterpretTextInput = {
 				text: {
--- a/packages/predictions/__tests__/Predictions.test.ts
+++ b/packages/predictions/__tests__/Predictions.test.ts
@@ -13,6 +13,8 @@ import {
 	TranslateTextOutput,
 } from '../src/types';
 
+import { createMockAmplifyContext } from './testUtils';
+
 describe('Predictions test', () => {
 	describe('getModuleName tests', () => {
 		test('happy and the only case', () => {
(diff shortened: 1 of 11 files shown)


packages/predictions/src/providers/AmazonAIConvertPredictionsProvider.ts
```diff
@@ -50,6 +55,33 @@ const LANGUAGES_CODE_IN_8KHZ = ['fr-FR', 'en-AU', 'en-GB', 'fr-CA'];
 export class AmazonAIConvertPredictionsProvider {
 	private translateClient?: TranslateClient;
 	private pollyClient?: PollyClient;
+	private _explicitCtx: AmplifyContext | undefined;
```
soberm, line 58: [minor] Same `private _explicitCtx` + `_ctx` getter + constructor block is duplicated verbatim across all three providers plus mirrored in `PredictionsClass`. Might be worth pulling into a small shared mixin/base class or helper (`resolveCtx(explicit?)`) in a follow-up, since any future tweak now needs updating in 3-4 places identically.

packages/predictions/src/providers/AmazonAIConvertPredictionsProvider.ts
```diff
@@ -50,6 +55,33 @@ const LANGUAGES_CODE_IN_8KHZ = ['fr-FR', 'en-AU', 'en-GB', 'fr-CA'];
...
+		return getGlobalContext();
+	}
+
+	/**
+	 * @param ctx - The AmplifyContext to use for auth and config.
+	 *   When provided, the provider is pinned to this context.
+	 *   When omitted, the provider resolves the global context lazily per operation.
+	 */
+	constructor(ctx?: AmplifyContext) {
+		if (isAmplifyContext(ctx)) {
```
soberm, line 81: [question] If a caller passes a truthy value that isn't a valid `AmplifyContext`, this silently falls back to global context instead of throwing. Intended for this migration, or should an invalid-but-defined ctx fail loudly? Worth confirming it matches the #14916 pattern intentionally.

packages/predictions/src/providers/AmazonAIConvertPredictionsProvider.ts
```diff
@@ -84,14 +116,14 @@ export class AmazonAIConvertPredictionsProvider {
 		logger.debug('Starting translation');
 
 		const { translateText = {} } =
-			Amplify.getConfig().Predictions?.convert ?? {};
+			this._ctx.resourcesConfig.Predictions?.convert ?? {};
```
soberm, line 119: [minor] Optional chaining on `.Predictions` but not on `resourcesConfig` itself. Fine for the mock, but confirm `getGlobalContext()`'s real implementation guarantees `resourcesConfig` is never undefined/null — otherwise an unconfigured path would throw rather than surface the existing NoRegion/validation error.

packages/predictions/src/providers/AmazonAIIdentifyPredictionsProvider.ts
```diff
@@ -110,7 +142,7 @@ export class AmazonAIIdentifyPredictionsProvider {
 					targetIdentityId: source.identityId,
 				};
 
-				getUrl({ key: source.key, options: storageConfig })
+				getUrl(this._ctx, { key: source.key, options: storageConfig })
```
soberm, line 145: [major] This is the one call site threading ctx into a cross-package call (`@aws-amplify/storage` `getUrl`). Confirm it's only reachable once storage's ctx-aware `getUrl` overload is merged (depends on #14874). There's also no local test asserting an explicit ctx flows through to `getUrl` — only the fallback path is exercised.

packages/predictions/__tests__/providers/AWSAIIdentifyPredictionsProvider.test.ts
```diff
@@ -272,37 +261,38 @@ const options = {
...
-});
-mockGetConfig.mockReturnValue({
-	Predictions: {
-		identify: options,
-	},
-});
-mockGetUrl.mockImplementation(({ key, options }) => {
-	console.log(key, options);
-	const level = options?.accessLevel || 'guest';
+mockGetUrl.mockImplementation((...args: any[]) => {
```
soberm, line 264: [minor] Handles both the old and ctx-aware signatures, but since this PR's point is the ctx-aware path, assert at least once that `args[0]` is the ctx passed to the constructor — a regression dropping the ctx arg wouldn't be caught.

packages/predictions/__tests__/providers/AWSAIIdentifyPredictionsProvider.test.ts
```diff
@@ -272,37 +261,38 @@ const options = {
...
-	credentials,
-	identityId,
-});
-mockGetConfig.mockReturnValue({
-	Predictions: {
-		identify: options,
-	},
-});
-mockGetUrl.mockImplementation(({ key, options }) => {
-	console.log(key, options);
```
soberm, line 285: [nit] Nice catch removing the stray `console.log(key, options)` — noisy test output unrelated to assertions.

packages/predictions/src/Predictions.ts
```diff
@@ -23,10 +25,29 @@ import {
...
 export class PredictionsClass {
-	private convertProvider = new AmazonAIConvertPredictionsProvider();
-	private identifyProvider = new AmazonAIIdentifyPredictionsProvider();
-	private interpretProvider = new AmazonAIInterpretPredictionsProvider();
+	private convertProvider: AmazonAIConvertPredictionsProvider;
+	private identifyProvider: AmazonAIIdentifyPredictionsProvider;
+	private interpretProvider: AmazonAIInterpretPredictionsProvider;
+
+	constructor(ctx?: AmplifyContext) {
+		if (isAmplifyContext(ctx)) {
```
soberm: [minor] The branches are identical except for the constructor args. Could simplify to `resolvedCtx = isAmplifyContext(ctx) ? ctx : undefined` passed to all three providers. Style call.

packages/predictions/src/index.ts
```diff
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-export { Predictions } from './Predictions';
+export { Predictions, PredictionsClass } from './Predictions';
```
soberm, line 3: [question] Newly exporting `PredictionsClass` is a public API addition. No changeset in this PR (intentional for the feature branch), just flagging it'll need a changeset/release note when it lands on main.

packages/predictions/__tests__/testUtils/mockAmplifyContext.ts
```diff
@@ -0,0 +1,31 @@
...
+): AmplifyContext {
+	const ctx: AmplifyContext = {
+		resourcesConfig: (resourcesConfig || {}) as ResourcesConfig,
+		libraryOptions: {},
+		fetchAuthSession: jest.fn().mockResolvedValue({}),
+		clearCredentials: jest.fn().mockResolvedValue(undefined),
+		getTokens: jest.fn().mockResolvedValue(undefined),
+	};
+
+	Object.defineProperty(ctx, AMPLIFY_CONTEXT_BRAND, {
```
soberm, line 25: [minor] This manually stamps `AMPLIFY_CONTEXT_BRAND`, duplicating a private branding mechanism from `@aws-amplify/core`. If the brand check changes, this mock could silently fail `isAmplifyContext()` and the explicit-ctx tests would fall through to the no-ctx branch without an obvious signal. Consider importing from a shared core testing package if one exists.

packages/predictions/__tests__/testUtils/mockAmplifyContext.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Accepts partial/incomplete configs to test error paths.
+ */
+export function createMockAmplifyContext(
+	resourcesConfig?: Partial<ResourcesConfig>,
+): AmplifyContext {
+	const ctx: AmplifyContext = {
+		resourcesConfig: (resourcesConfig || {}) as ResourcesConfig,
+		libraryOptions: {},
+		fetchAuthSession: jest.fn().mockResolvedValue({}),
+		clearCredentials: jest.fn().mockResolvedValue(undefined),
```
soberm, line 21: [nit] `clearCredentials`/`getTokens` are stubbed but unused by the exercised predictions paths. Fine as boilerplate, just flagging.

packages/predictions/__tests__/providers/AWSAIIdentifyPredictionsProvider.test.ts
```diff
@@ -716,7 +706,7 @@ describe('Predictions identify provider test', () => {
 		test('error case invalid input source', () => {
 			const detectLabelInput = {
 				labels: { source: null, type: 'LABELS' },
-			};
+			} as unknown as IdentifyLabelsInput;
```
soberm, line 721: [nit] Reasonable to force a type-incompatible test input, but a one-line comment on why (testing the runtime validation error path with an invalid source) would help the next reader.

packages/predictions/__tests__/providers/AWSAIIdentifyPredictionsProvider.test.ts
```diff
@@ -735,7 +725,7 @@ describe('Predictions identify provider test', () => {
 			await predictionsProvider.identify(detectLabelInput);
 
 			expect(
-				predictionsProvider.rekognitionClient.config.customUserAgent,
+				(predictionsProvider as any).rekognitionClient.config.customUserAgent,
```
soberm, line 740: [nit] Casting to `any` to reach the private `rekognitionClient` — this PR introduced the explicit type creating the need. Worth an eslint-disable note, or accept it; flagging as a new occurrence, three times over.

packages/predictions/__tests__/Predictions.test.ts
```diff
@@ -13,75 +13,155 @@ import {
...
+		});
 	});
 
-	test('interpret test', async () => {
-		const predictions = new PredictionsClass();
-		const input: InterpretTextInput = {
-			text: {
-				source: {
-					text: 'Test text',
+	describe('explicit-ctx construction', () => {
```
soberm, line 95: [minor] Good addition, but none of the three tests assert the provider received the same ctx instance — only that the spy method resolved and was called once. A bug passing `undefined` instead of `ctx` wouldn't be caught. Consider a constructor spy on the provider.

soberm (commented):
Summary: This PR migrates the predictions package's three providers (`AmazonAIConvertPredictionsProvider`, `AmazonAIIdentifyPredictionsProvider`, `AmazonAIInterpretPredictionsProvider`) plus the `PredictionsClass` facade to accept an optional explicit `AmplifyContext`, falling back to a lazily-resolved global context (`getGlobalContext()`) when no context is passed. Config access moves from `Amplify.getConfig()` to `ctx.resourcesConfig`, and auth moves from the module-level `fetchAuthSession()` to `ctx.fetchAuthSession()`. It also threads ctx into the Storage `getUrl` call in the Identify provider. Tests are reworked to use a new `createMockAmplifyContext` test util instead of mocking `@aws-amplify/core` wholesale. This follows an established pattern from a prior Geo PR (#14916) in the same migration series.

Overall the change is mechanical and consistent across the three providers, and the lazy-getter approach for the fallback context is a reasonable design (avoids snapshotting stale global context at construction time). A few things worth raising before merge.

Positive notes:
- The `_explicitCtx` / lazy `_ctx` getter pattern is well-reasoned and documented in the JSDoc — it explicitly explains why it's lazy (to honor `setGlobalContext` reconfiguration).
- Consistent, mechanical application of the same pattern across all three providers.
- Replacing the `jest.mock('@aws-amplify/core', ...)` module-level mock with a proper `createMockAmplifyContext()` factory is a real testability improvement.
- Good test structure split (global-ctx fallback vs explicit-ctx construction describe blocks).
- Type-tightening in `AWSAIIdentifyPredictionsProvider.test.ts` is a nice incidental improvement.

Recommended vote: Comment. Nothing here is a functional blocker for the feature branch. Main things to address before it lands on main: (1) the `isAmplifyContext` silent-fallback behavior for invalid-but-defined ctx (confirm intentional), (2) tightening the `getUrl(this._ctx, ...)` test coverage, and (3) the changeset note for the new `PredictionsClass` export once this leaves the feature branch.

packages/predictions/src/providers/AmazonAIConvertPredictionsProvider.ts
bobbor: Intentional for now and consistent with geo/pubsub: `isAmplifyContext` is the brand check, and a truthy-but-unbranded value falling back to global matches the ctx-arg behavior elsewhere. Making this throw is the constructor-path guard we agreed to fold into the central `resolveCtxArgs` guard follow-up from #14912 (same as geo #14916) — will be addressed there so all Pattern-7 packages change together.

packages/predictions/src/providers/AmazonAIIdentifyPredictionsProvider.ts
bobbor: Good catch — added an identity assertion: the test now verifies the exact ctx instance passed to the provider is the first arg storage `getUrl` receives (`mock.calls[0][0] toBe ctx`). And yes — depends on #14874, which is merged in the base branch. Fixed in ec47b09.

packages/predictions/src/providers/AmazonAIConvertPredictionsProvider.ts
bobbor: Agreed — will factor the explicit-ctx + lazy getter into a shared helper in a follow-up PR; same dedup applies to geo/pubsub, so doing it centrally rather than per-package here.

packages/predictions/src/providers/AmazonAIConvertPredictionsProvider.ts
bobbor: Confirmed: `getGlobalContext()` either throws (pre-configure) or returns a valid ctx — it can never yield `resourcesConfig === undefined`. `AmplifyContext` declares `resourcesConfig` as required readonly, and `configure()` always assigns it (worst case `{}`). The optional part is the category sub-key (`Predictions?`), which is why the chaining sits there.

packages/predictions/__tests__/providers/AWSAIIdentifyPredictionsProvider.test.ts
bobbor: 🙏 yeah, drive-by cleanup.

packages/predictions/src/index.ts
bobbor: No changeset needed on this PR since it targets the feat/bobbor/v6-context feature branch; noted for the changeset that accompanies v6-context landing on main — `PredictionsClass` export will be called out as a public API addition (same treatment as `GeoClass`).

packages/predictions/__tests__/providers/AWSAIIdentifyPredictionsProvider.test.ts
bobbor: Covered by the new identity assertion added for the thread above — the mock now asserts `args[0]` is the exact ctx instance, so dropped ctx-forwarding fails the suite. Fixed in ec47b09.

packages/predictions/src/Predictions.ts
bobbor: Done — collapsed to `resolvedCtx = isAmplifyContext(ctx) ? ctx : undefined` with single construction. Fixed in ec47b09.

packages/predictions/__tests__/providers/AWSAIIdentifyPredictionsProvider.test.ts
bobbor: Acknowledged — these are rewrites of the pre-existing user-agent assertions (the suite already peeked at the private clients pre-migration). A clean fix (asserting via SDK client constructor args or a test-visible accessor) is a test-architecture change across the provider suites — bundling it with the shared mock-context/test-helper follow-up rather than growing this PR.

packages/predictions/__tests__/testUtils/mockAmplifyContext.ts
bobbor: Yep — kept as boilerplate so the mock satisfies the full `AmplifyContext` surface; keeps the util copy-paste-consistent across packages until the shared helper lands.

packages/predictions/__tests__/testUtils/mockAmplifyContext.ts
bobbor: Agreed — this is the same concern as the geo follow-up (shared `createMockAmplifyContext`, e.g. in a core testing entry point) so the brand mechanics live next to `isAmplifyContext`. Will cover predictions/geo/pubsub copies in that follow-up.

packages/predictions/__tests__/providers/AWSAIIdentifyPredictionsProvider.test.ts
bobbor: Added a comment explaining the double cast intentionally bypasses compile-time checks to exercise the runtime validation path. Fixed in ec47b09.

packages/predictions/__tests__/Predictions.test.ts
bobbor: Added the identity test, though via typed structural access to the providers' stored `_explicitCtx` rather than constructor spies — spying on the provider constructors would mean mocking package internals, which we're trying to keep out of these suites (mock only globals/boundaries). Asserts `toBe(ctx)` for all three providers. Fixed in ec47b09.

bobbor:
Thanks for the thorough review @soberm!

**Fixed in ec47b09:**
- Ctx-identity coverage for `getUrl` + all three providers (`toBe(ctx)` assertions)
- Constructor simplification in `Predictions.ts`
- Runtime-validation double-cast comment explaining intent

**Answered inline:**
- Silent fallback → intentional, central guard follow-up from #14912/#14916
- `resourcesConfig` guarantee — required readonly, never undefined post-configure
- Changeset deferred to v6-context → main landing

**Follow-ups owed:**
- Shared explicit-ctx + lazy-getter helper (dedup across geo/pubsub/predictions)
- Shared `createMockAmplifyContext` test util (covers the `as any` private-field test-architecture cleanup)

packages/predictions/__tests__/providers/AWSAIInterpretPredictionsProvider.test.ts
```diff
@@ -232,7 +210,16 @@ describe('Predictions interpret provider test', () => {
 	});
 	describe('interpretText tests', () => {
 		test('happy case credentials exist detectEntities', async () => {
-			const predictionsProvider = new AmazonAIInterpretPredictionsProvider();
+			const ctx = createMockAmplifyContext({
```
osama-rizk: Convert and Identify each have a credential-failure test exercising `fetchAuthSession() → {}` → NoCredentials against the new ctx; Interpret's suite is happy-path + user-agent only (no error / no-credentials test in this file), so its migrated `this._ctx.fetchAuthSession()` NoCredentials branch is unverified. Looks pre-existing, but the migration is a good moment to close it.

osama-rizk (commented):
approve-with-nits — all 4 classes migrated uniformly and faithfully to the lazy-ctx pattern; storage `getUrl(ctx, …)` threading verified safe (the mock handles the 2-arg overload, so a wrong-arg-order regression would surface). Facade ctx-threading IS covered — the "all three providers receive the same ctx instance" test asserts `provider._explicitCtx === ctx` for all three. One test-coverage note inline. Minor: the three "delegates … constructed with ctx" tests actually assert only delegation (the spied method's call count), not ctx-threading — that coverage lives in the separate identity test above them — so consider renaming them for clarity; the unconfigured-path error-surface change (series-wide across the v6-context migration, worth one epic-level release note); and Interpret builds its ctx mock per-test (7×) where Identify uses `beforeAll` — a shared helper would cut ~40 lines.

packages/predictions/__tests__/providers/AWSAIInterpretPredictionsProvider.test.ts
bobbor: Agreed — good moment to close it. Added the credential-failure test for Interpret mirroring the Convert/Identify pattern (explicit ctx with fetchAuthSession → {} asserting the NoCredentials rejection) in 288e8be93. Verified the assertion bites — with valid credentials mocked the test fails.

bobbor:
Thanks for the review, @osama-rizk — appreciate the careful pass.

Summary of 288e8be93 (test-only):
- (a) Added the Interpret NoCredentials test (explicit ctx, fetchAuthSession → {}, asserting the rejection).
- (b) Renamed the three delegation tests to drop the ctx-threading claim — they assert delegation only; ctx-threading is covered by the dedicated identity test.
- (c) Hoisted the Interpret suite's per-test ctx mock into beforeAll per the Identify structure (net -33 lines).

On the release note: agreed the unconfigured-path error-surface change (an Error from getGlobalContext instead of a silent empty config) deserves an epic-level release note. Rather than doing it per-package, I'll include it in the notes/changeset when feat/bobbor/v6-context lands on main, so it covers the whole series.

If there's nothing further, could both reviewers submit a formal Approve? GitHub still shows both reviews in Comment state — @osama-rizk your approve-with-nits didn't register as an Approval — and the branch-protection gate needs the explicit approved state to unblock.

osama-rizk (approved):

soberm (approved):

PR #14919, opened 2026-08-21
feat(api-graphql): migrate GraphQL API, realtime providers, and events to AmplifyContext

bobbor:
#### Description of changes

B-api-graphql split of the v6 context migration (Task 9 of the migration plan): migrates `@aws-amplify/api-graphql` and the `@aws-amplify/api` umbrella from the `Amplify.getConfig()` / singleton `fetchAuthSession` pattern to explicit `AmplifyContext`, with backward-compatible overloads.

- **GraphQL core**: `InternalGraphQLAPI` / `GraphQLAPI` accept `AmplifyContext | AmplifyClassV6` at the entry point and bridge legacy instances via a branded `bridgeAmplifyClass` (mirroring the corrected storage #14874 bridge pattern — not api-rest's original bare-cast version); all internals (`graphqlAuth`, `graphqlRequest`, `resolveConfig`, `resolveLibraryOptions`) take `ctx: AmplifyContext` first and read `ctx.resourcesConfig` / call `ctx.fetchAuthSession()` (Patterns 3/4).
- **Realtime / WebSocket auth**: `awsRealTimeHeaderBasedAuth` takes an optional `ctx`, and the explicit context is threaded per-call (never stored on the long-lived provider singletons) from `_graphqlSubscribe` and the events APIs through `AWSAppSyncRealTimeProvider`, `AWSAppSyncEventsProvider`, and `AWSWebSocketProvider` to **both** auth points — the subscription registration payload and the WebSocket connection handshake. Falls back to `getGlobalContext()` when no ctx is supplied (unchanged singleton behavior).
- **Events**: `events.connect()` / `events.post()` gain `fn(...)` / `fn(ctx, ...)` overloads via the shared `resolveCtxArgs` helper (inherits the mis-ordered-ctx guard from #14912); `appsyncRequest` and `events/utils` take ctx.
- **generateClient**: `ClientGenerationParams.amplify` widened to `AmplifyContext | AmplifyClassV6`; `internals/generateClient.ts` bridges legacy singleton callers (so `packages/api` `API.ts` keeps its call shape). The SSR closure path in `internals/server/generateClientWithAmplifyInstance.ts` is intentionally unchanged (`null | closure`).
- **Server (Pattern 5)**: `server/generateClient.ts` kept as a dedicated wrapper accepting `AmplifyContext | AmplifyServer.ContextSpec`, normalizing via a new `resolveServerContext` (bridged + branded, matching the corrected storage #14874 pattern). Full legacy server-path auth alignment remains deferred to Phase C.
- **api umbrella**: `internals/InternalAPI.ts` migrated; `API.ts` unchanged by design (bridging happens inside api-graphql).
- **Tests**: new branded `__tests__/testUtils/mockAmplifyContext.ts`; 7 test files migrated off singleton-shape assertions; new non-tautological tests prove an explicitly-passed ctx (not the global one) reaches the realtime auth layer on both the RealTime and Events provider paths.

No changeset (feature-branch target).

#### Issue #, if available

Part of the v6 context migration (base: Split A #14805; dependency: B-api-rest #14914).

- `yarn turbo run build test lint --filter=@aws-amplify/api-graphql --filter=@aws-amplify/api --force` — exit 0; api-graphql 8/8 suites, 153/153 tests; api 2/2 suites, 86/86 tests; lint clean
- `yarn turbo run build --filter=aws-amplify` — exit 0 (dependent umbrella build, cross-package type check)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/api-graphql/src/Providers/AWSWebSocketProvider/authHeaders.ts
+++ b/packages/api-graphql/src/Providers/AWSWebSocketProvider/authHeaders.ts
@@ -22,10 +22,10 @@ type AWSAppSyncRealTimeAuthInput =
 	};
 
 const awsAuthTokenHeader = async (
-	ctx: AmplifyContext,
+	resolveCtx: () => AmplifyContext,
 	{ host }: AWSAppSyncRealTimeAuthInput,
 ) => {
-	const session = await ctx.fetchAuthSession();
+	const session = await resolveCtx().fetchAuthSession();
 
 	return {
 		Authorization: session?.tokens?.accessToken?.toString(),
@@ -34,7 +34,7 @@ const awsAuthTokenHeader = async (
 };
 
 const awsRealTimeApiKeyHeader = async (
-	_ctx: AmplifyContext,
+	_resolveCtx: () => AmplifyContext,
 	{ apiKey, host }: AWSAppSyncRealTimeAuthInput,
 ) => {
 	const dt = new Date();
@@ -48,7 +48,7 @@ const awsRealTimeApiKeyHeader = async (
 };
 
 const awsRealTimeIAMHeader = async (
-	ctx: AmplifyContext,
+	resolveCtx: () => AmplifyContext,
 	{
 		payload,
 		canonicalUri,
@@ -61,7 +61,7 @@ const awsRealTimeIAMHeader = async (
 		service: 'appsync',
 	};
 
-	const creds = (await ctx.fetchAuthSession()).credentials;
+	const creds = (await resolveCtx().fetchAuthSession()).credentials;
 
 	const request = {
 		url: `${appSyncGraphqlEndpoint}${canonicalUri}`,
@@ -88,7 +88,7 @@ const awsRealTimeIAMHeader = async (
 };
 
 const customAuthHeader = async (
-	_ctx: AmplifyContext,
+	_resolveCtx: () => AmplifyContext,
 	{ host, additionalCustomHeaders }: AWSAppSyncRealTimeAuthInput,
 ) => {
 	/**
@@ -120,9 +120,13 @@ export const awsRealTimeHeaderBasedAuth = async (
 		payload,
 	} = authInput;
 
-	// Resolve context: use the explicitly provided ctx, or fall back to global context.
-	// NOTE: we do NOT capture getGlobalContext() eagerly — it's resolved fresh per call.
-	const resolvedCtx = ctx ?? getGlobalContext();
+	// Resolve context LAZILY. Only the auth handlers that actually consume the
+	// context (iam / oidc / userPool) invoke this thunk. apiKey / none / lambda
+	// ignore it, so they must keep working with no explicit ctx AND no global
+	// context — resolving (and potentially throwing) here would break them.
+	// NOTE: we also do NOT capture getGlobalContext() eagerly — it's resolved
+	// fresh, per invocation of the thunk.
+	const resolveCtx = (): AmplifyContext => ctx ?? getGlobalContext();
 
 	const headerHandler = {
 		apiKey: awsRealTimeApiKeyHeader,
@@ -148,7 +152,7 @@ export const awsRealTimeHeaderBasedAuth = async (
 
 		logger.debug(`Authenticating with ${JSON.stringify(authenticationType)}`);
 
-		const result = await handler(resolvedCtx, {
+		const result = await handler(resolveCtx, {
 			payload,
 			canonicalUri,
 			appSyncGraphqlEndpoint,
--- a/packages/api-graphql/src/Providers/AWSWebSocketProvider/authHeaders.ts
+++ b/packages/api-graphql/src/Providers/AWSWebSocketProvider/authHeaders.ts
@@ -1,7 +1,11 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { ConsoleLogger, fetchAuthSession } from '@aws-amplify/core';
+import {
+	AmplifyContext,
+	ConsoleLogger,
+	getGlobalContext,
+} from '@aws-amplify/core';
 import { signRequest } from '@aws-amplify/core/internals/aws-client-utils';
 import { AmplifyUrl } from '@aws-amplify/core/internals/utils';
 
@@ -17,8 +21,11 @@ type AWSAppSyncRealTimeAuthInput =
 		host?: string | undefined;
 	};
 
-const awsAuthTokenHeader = async ({ host }: AWSAppSyncRealTimeAuthInput) => {
-	const session = await fetchAuthSession();
+const awsAuthTokenHeader = async (
+	ctx: AmplifyContext,
+	{ host }: AWSAppSyncRealTimeAuthInput,
+) => {
+	const session = await ctx.fetchAuthSession();
 
 	return {
 		Authorization: session?.tokens?.accessToken?.toString(),
@@ -26,10 +33,10 @@ const awsAuthTokenHeader = async ({ host }: AWSAppSyncRealTimeAuthInput) => {
 	};
 };
 
-const awsRealTimeApiKeyHeader = async ({
-	apiKey,
-	host,
-}: AWSAppSyncRealTimeAuthInput) => {
+const awsRealTimeApiKeyHeader = async (
+	_ctx: AmplifyContext,
+	{ apiKey, host }: AWSAppSyncRealTimeAuthInput,
+) => {
 	const dt = new Date();
 	const dtStr = dt.toISOString().replace(/[:-]|\.\d{3}/g, '');
 
@@ -40,18 +47,21 @@ const awsRealTimeApiKeyHeader = async ({
 	};
 };
 
-const awsRealTimeIAMHeader = async ({
-	payload,
-	canonicalUri,
-	appSyncGraphqlEndpoint,
-	region,
-}: AWSAppSyncRealTimeAuthInput) => {
+const awsRealTimeIAMHeader = async (
+	ctx: AmplifyContext,
+	{
+		payload,
+		canonicalUri,
+		appSyncGraphqlEndpoint,
+		region,
+	}: AWSAppSyncRealTimeAuthInput,
+) => {
 	const endpointInfo = {
 		region,
 		service: 'appsync',
 	};
 
-	const creds = (await fetchAuthSession()).credentials;
+	const creds = (await ctx.fetchAuthSession()).credentials;
 
 	const request = {
 		url: `${appSyncGraphqlEndpoint}${canonicalUri}`,
@@ -77,10 +87,10 @@ const awsRealTimeIAMHeader = async ({
 	return signedParams.headers;
 };
 
-const customAuthHeader = async ({
-	host,
-	additionalCustomHeaders,
-}: AWSAppSyncRealTimeAuthInput) => {
+const customAuthHeader = async (
+	_ctx: AmplifyContext,
+	{ host, additionalCustomHeaders }: AWSAppSyncRealTimeAuthInput,
+) => {
 	/**
 	 * If `additionalHeaders` was provided to the subscription as a function,
 	 * the headers that are returned by that function will already have been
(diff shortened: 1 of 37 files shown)


packages/api-graphql/src/utils/bridgeAmplifyClass.ts
```diff
@@ -0,0 +1,51 @@
...
+			return amplify.getConfig();
+		},
+		// Live getter: AmplifyClass.libraryOptions is REASSIGNED wholesale on
+		// configure(), so a snapshot would go stale.
+		get libraryOptions() {
+			return amplify.libraryOptions;
+		},
+		// AmplifyContext.fetchAuthSession has OPTIONAL options while
+		// AuthClass.Auth.fetchAuthSession requires it, hence the `?? {}` default.
+		fetchAuthSession: options => amplify.Auth.fetchAuthSession(options ?? {}),
```
osama-rizk, line 36: The bridge here is correct and does not inherit the api-rest #14914 bare-cast bug (this maps to `amplify.Auth.*`, verified). But it has no direct test — the exact path that shipped broken in that sibling PR is only exercised indirectly, so a future revert to `return ctx as AmplifyContext` would pass every existing test and re-break all default-auth SSR GraphQL (Cognito/IAM). For the one PR in this series carrying that landmine, a direct unit test — feed a bare `AmplifyClass` (only `.Auth.*`), assert the bridged `ctx.fetchAuthSession()` delegates to `Auth.fetchAuthSession` with the `{}` default — is the regression guard worth adding before merge.

packages/api-graphql/src/Providers/AWSWebSocketProvider/authHeaders.ts
```diff
@@ -96,17 +106,24 @@ const customAuthHeader = async ({
...
+		canonicalUri,
+		appSyncGraphqlEndpoint,
+		region,
+		additionalCustomHeaders,
+		payload,
+	} = authInput;
+
+	// Resolve context: use the explicitly provided ctx, or fall back to global context.
+	// NOTE: we do NOT capture getGlobalContext() eagerly — it's resolved fresh per call.
+	const resolvedCtx = ctx ?? getGlobalContext();
```
osama-rizk: `resolvedCtx = ctx ?? getGlobalContext()` is computed unconditionally here, before the handler dispatch below — so it's evaluated for every auth type, including apiKey/none/lambda whose handlers don't use the ctx for auth. Since `getGlobalContext()` throws when unconfigured, a direct no-ctx provider `subscribe`/`connect` with `authenticationType: 'apiKey'` before `configure()` could now throw where the old code proceeded with the API key. (The line-above comment addresses not *caching* it at singleton scope, which is a separate concern.) Narrow — the normal client + events paths always supply a ctx — but resolving ctx lazily inside only the handlers that need it (userPool/oidc/iam) avoids it.

osama-rizk (commented):
approve-with-nits — the riskiest PR in the series, and the high-risk parts are sound. The server bridge is correct: `bridgeAmplifyClass` maps `fetchAuthSession`/`clearCredentials`/`getTokens` to `amplify.Auth.*` (verified lines 36-39 + traced the full SSR call path), so it does NOT inherit api-rest #14914's bare-cast `TypeError`. Realtime ctx is threaded per-call to both websocket auth points and never stored on the singletons — well tested. Two nits inline. Also: the description says this "mirrors the api-rest helper," but api-rest's bridge was the broken one — it actually mirrors the correct storage #14874 pattern; worth fixing the description so it doesn't point future readers at the buggy sibling. Minor: unconfigured-path error-surface change (series-wide).

packages/api/src/internals/InternalAPI.ts
```diff
@@ -89,7 +89,7 @@ export class InternalAPIClass {
 		};
 
 		return this._graphqlApi.graphql(
-			Amplify,
+			getGlobalContext(),
```
soberm: [blocker] `getGlobalContext()` throws synchronously if `Amplify.configure()`/`setGlobalContext()` hasn't run yet, and this method isn't `async` and has no try/catch around it — so the exception now propagates straight out of `.graphql()` before a `Promise` or `Observable` is ever returned. That's a real change from passing `Amplify` directly here before (always truthy, never throws).

The riskiest caller is DataStore's subscription processor (`packages/datastore/src/sync/processors/subscription.ts`), which calls this path directly and expects an `Observable` back — RxJS convention is that setup errors surface via `.subscribe()`'s error channel, not a synchronous throw at call time. This PR's own test changes are evidence the assumption doesn't always hold: five datastore test files needed `setGlobalContext()`/`clearGlobalContext()` added specifically because this call now throws where it didn't before.

Worth deferring the context resolution into the async/Observable boundary rather than resolving it eagerly here as an argument expression, plus a regression test that exercises the no-global-context case for this method specifically (a query/mutation and a subscription) and asserts it fails as a rejection/Observable error rather than a synchronous throw.

packages/datastore/__tests__/helpers/datastoreFactory.ts
```diff
@@ -204,6 +204,34 @@ export function getDataStore({
 		DataStore: DataStoreType;
 	} = require('../../src/datastore/datastore');
 
+	// After jest.resetModules(), the fresh module graph has no global context.
+	// InternalAPI.graphql() now calls getGlobalContext() which throws unless
+	// a context exists. Establish one via setGlobalContext() on the FRESH core
+	// module (not Amplify.configure — that dispatches Hub events which can
+	// auto-start DataStore before its storage adapter is configured).
+	const { setGlobalContext } = require('@aws-amplify/core/internals/utils');
+	const testCtx = {
```
soberm, line 213: [minor] This exact `testCtx` shape (plus the `Symbol.for('amplify.context')` branding boilerplate) is duplicated near-verbatim in `commonAdapterTests.ts` (twice), `observeQuery.test.ts`, and `SQLiteAdapter.test.ts` — only the endpoint URL differs between them. `api-graphql` solved the same problem in this PR with a shared `__tests__/testUtils/mockAmplifyContext.ts` helper; datastore could use the same pattern here.

Also, since this file already imports from `@aws-amplify/core/internals/utils`, might as well pull in `AMPLIFY_CONTEXT_BRAND` from `@aws-amplify/core` instead of hardcoding `Symbol.for('amplify.context')`. Works today since `Symbol.for` is registry-backed, but it duplicates an implementation detail that's supposed to live behind the exported constant.

packages/api-graphql/src/utils/bridgeAmplifyClass.ts
```diff
@@ -0,0 +1,51 @@
...
+ * (`fetchAuthSession` / `clearCredentials` / `getTokens`). Those live under its
+ * cross-category `Auth` utility, so this helper maps them through.
+ *
+ * This is used:
+ * - In the server wrapper where a legacy `ContextSpec` provides an `AmplifyClass` instance.
+ * - In the internals where dependents still pass an `AmplifyClassV6` until they migrate.
+ *
+ * @internal
+ */
+export const bridgeAmplifyClass = (amplify: AmplifyClassV6): AmplifyContext => {
```
soberm, line 24: [minor] Neither this file nor `resolveServerContext.ts` has a dedicated test file — both are only exercised indirectly through other tests (e.g. `resolveConfig.test.ts` builds an `AmplifyContext` directly via `createMockAmplifyContext`, bypassing this bridge entirely). Given this is the backward-compat path basically every pre-migration caller hits, a focused test would be good here: the live-getter behavior of `resourcesConfig`/`libraryOptions` tracking the underlying `AmplifyClassV6` after it mutates, the `options ?? {}` default in `fetchAuthSession`, and that branding actually gets applied to the result.

packages/api-graphql/src/internals/InternalGraphQLAPI.ts
```diff
@@ -158,7 +183,7 @@ export class InternalGraphQLAPIClass {
 			}
 			case 'subscription':
 				return this._graphqlSubscribe(
-					amplify as AmplifyClassV6,
+					ensureContext(amplify as AmplifyContext | AmplifyClassV6),
```
soberm, line 186: [nit] This skips the `isAmplifyInstance()` check the query/mutation branch above uses before calling `ensureContext`. If `amplify` were ever the SSR closure form here, this would throw a confusing `TypeError` out of `bridgeAmplifyClass`'s getter instead of a clear error. Pre-existing though — the old code did the same unchecked cast for the subscription case — so not introduced by this PR, just flagging since you're touching this block anyway.

packages/api/src/internals/InternalAPI.ts
bobbor: Fixed in ec5c3e64e. `graphql()` now detects the operation type first (touches no context), then defers resolution to the async boundary: query/mutation surfaces a missing global context as a rejected Promise (`try/catch → Promise.reject`, and on success the original cancellable promise is returned unchanged so `cancel()` token identity is preserved); subscriptions wrap resolution in rxjs `defer()` so the error arrives on the Observable's error channel — exactly the contract DataStore's subscription processor expects. Added `packages/api/__tests__/internals/InternalAPI.test.ts` asserting both paths: the call expression does not throw synchronously, the query path rejects with the no-context error, and the subscription path errors via `.subscribe({ error })`.

packages/api-graphql/src/Providers/AWSWebSocketProvider/authHeaders.ts
bobbor: Good catch — fixed in ec5c3e64e. The handlers now receive a lazy `resolveCtx` thunk (`() => ctx ?? getGlobalContext()`) and only the userPool/oidc/iam handlers invoke it; apiKey/none/lambda never touch the context, so a pre-configure `subscribe`/`connect` with `authenticationType: 'apiKey'` proceeds exactly as the old code did. Explicit-ctx forwarding for the credentialed auth types is unchanged (existing ctx-threading tests still pass unmodified). Added a test covering apiKey auth with no ctx and no global context.

packages/api-graphql/src/utils/bridgeAmplifyClass.ts
bobbor: Agreed on the landmine framing — added `__tests__/utils/bridgeAmplifyClass.test.ts` in ec5c3e64e: feeds a bare AmplifyClass-shaped object exposing the auth methods only under `.Auth`, asserts the bridged `ctx.fetchAuthSession()` delegates to `Auth.fetchAuthSession` with the `{}` default (same for getTokens/clearCredentials), so a revert to `return ctx as AmplifyContext` now fails these tests immediately. Also covers live `resourcesConfig`/`libraryOptions` getters after config mutation and non-enumerable branding. Also fixed the PR description — the bridge mirrors the corrected storage #14874 pattern, not api-rest's original.

packages/api-graphql/src/utils/bridgeAmplifyClass.ts
bobbor: Done in ec5c3e64e — `__tests__/utils/bridgeAmplifyClass.test.ts` covers the live-getter behavior (config mutated after bridging is reflected), the `options ?? {}` default in fetchAuthSession, and branding applied non-enumerably; `__tests__/utils/resolveServerContext.test.ts` covers the branded pass-through branch (same ref, adapter-core not called) and the ContextSpec branch via a mocked adapter-core boundary (same pattern as the api-rest suite).

packages/datastore/__tests__/helpers/datastoreFactory.ts
bobbor: Valid on both counts — deferring to a follow-up rather than this PR: I'll extract a shared `createTestCtx()`/`injectNoOpInternalAPI()` helper for the datastore suites and switch the hand-rolled `Symbol.for('amplify.context')` to the exported `AMPLIFY_CONTEXT_BRAND`. Folding it into the shared createMockAmplifyContext consolidation follow-up already promised on #14916 (the same helper duplication exists across sibling packages), so it lands once, series-wide.

packages/api-graphql/src/internals/InternalGraphQLAPI.ts
bobbor: Agreed it's pre-existing (the old code did the same unchecked cast on the subscription branch). Tracking it as a follow-up together with the central resolveCtxArgs unbranded-ctx guard from #14912 — that guard is the right single place to give this a clear error message rather than adding a one-off check in this block.

bobbor:
@osama-rizk thanks for the deep trace on the SSR path. Both review-body items handled: PR description corrected (the bridge mirrors the corrected storage #14874 pattern — updated so nobody copies the api-rest original), and the unconfigured-path error-surface change is tracked as the existing series-wide follow-up (typed re-wrap of the getGlobalContext error at core level, promised on #14913 — affects auth/storage/analytics equally, so it lands once in core rather than per-package). All inline items addressed in ec5c3e64e; CI re-running.

osama-rizk (approved):

soberm (approved):

PR #14923, opened 2026-08-26
feat(api): add AmplifyContext overload to generateClient

bobbor:
## Description

Task 10 (final Phase B step) of the v6 `AmplifyContext` migration: migrates the `@aws-amplify/api` umbrella package so its public `generateClient` accepts an explicit `AmplifyContext`, while preserving the singleton path for backward compatibility.

### Changes
- `packages/api/src/API.ts` — adds a second public overload `generateClient(ctx: AmplifyContext, options?)` alongside the existing `generateClient(options?)`. Implementation uses the typed-union signature (`ctxOrOptions?: AmplifyContext | Options, maybeOptions?: Options`) narrowed via `isAmplifyContext` (per the pubsub #14917 pattern) — no `...args`, no `as any` (the pre-existing `options || ({} as any)` fallback was removed in favor of `...(options ?? {})` — behavior-neutral, since `options` is either an object or `undefined`, never another falsy value). With no ctx, the `Amplify` singleton is passed through unchanged and bridged downstream by api-graphql's `bridgeAmplifyClass` (#14919), so the legacy path is untouched.
- `packages/api/__tests__/testUtils/mockAmplifyContext.ts` — Pattern-8 branded `createMockAmplifyContext` test util (same shape as api-graphql's).
- `packages/api/__tests__/generateClientContext.test.ts` — real-config tests (no mocking of api/core internals): global-singleton path, explicit branded-ctx path (asserts config resolves from ctx and NOT the global), and ctx+options forwarding.

Not in scope (already landed in #14919): `InternalAPI.graphql` lazy `getGlobalContext`, `server.ts`/`internals` re-exports. Server-context bridging remains a Phase C concern.

No changeset — PR targets the feature branch `feat/bobbor/v6-context`, not `main`.

## Validation
- `yarn turbo run build --filter=@aws-amplify/api` ✅
- `yarn turbo run test --filter=@aws-amplify/api` ✅ (4 suites, 97 tests, lint + 100% ts-coverage)
- Dependent builds: `yarn turbo run build --filter=aws-amplify --filter=@aws-amplify/adapter-nextjs` ✅
- `yarn turbo run test:size --filter=aws-amplify` ✅ (all bundles within limits)

## Checklist
- [x] PR description included
- [x] `yarn test` passes for the affected package
- [ ] Relevant documentation is changed or added (N/A — feature-branch migration work)

### DIFF
--- a/packages/api/__tests__/generateClientContext.test.ts
+++ b/packages/api/__tests__/generateClientContext.test.ts
@@ -26,6 +26,10 @@ const _postSpy = jest.spyOn((GraphQLAPI as any)._api, 'post');
  * GraphQL POST.
  */
 function getFirstPost() {
+	// Guard against silently indexing into an empty `mock.calls` array: assert a
+	// single POST was recorded before reading `calls[0]`.
+	expect(_postSpy).toHaveBeenCalledTimes(1);
+
 	const postOptions = _postSpy.mock.calls[0][1] as {
 		url: URL;
 		options: {
@@ -60,6 +64,9 @@ function configureGlobalSingleton() {
 describe('generateClient - AmplifyContext overload', () => {
 	beforeEach(() => {
 		configureGlobalSingleton();
+		// The mocked POST returns a plain (non-promise) object. Downstream code
+		// `await`s the response, and `await` resolves non-thenable values as-is,
+		// so `mockReturnValue` (rather than `mockResolvedValue`) is intentional.
 		_postSpy.mockReturnValue({
 			body: {
 				json() {
@@ -81,8 +88,26 @@ describe('generateClient - AmplifyContext overload', () => {
 		jest.clearAllMocks();
 	});
 
+	afterAll(() => {
+		// Restore the module-scope spy so it does not leak into other suites.
+		_postSpy.mockRestore();
+	});
+
 	test('generateClient() binds to the global singleton path', async () => {
 		const client = generateClient();
+
+		await client.graphql({ query: `query A { queryA { a b c } }` });
+
+		const { endpoint, apiKey } = getFirstPost();
+		expect(endpoint).toEqual(GLOBAL_ENDPOINT);
+		expect(apiKey).toEqual(GLOBAL_API_KEY);
+	});
+
+	test('generateClient(options) with a plain options object resolves config from the global singleton (not mistaken for a context)', async () => {
+		// A plain options object is not a branded AmplifyContext, so the client
+		// must fall back to the global singleton for config resolution and must
+		// not throw or treat the options object as a context.
+		const client = generateClient({ authMode: 'apiKey' });
 
 		await client.graphql({ query: `query A { queryA { a b c } }` });
 
--- a/packages/api/__tests__/generateClientContext.test.ts
+++ b/packages/api/__tests__/generateClientContext.test.ts
@@ -1,0 +1,147 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { enableFetchMocks } from 'jest-fetch-mock';
+import { Amplify } from '@aws-amplify/core';
+import { GraphQLAPI } from '@aws-amplify/api-graphql';
+import { generateClient } from '@aws-amplify/api';
+
+import { createMockAmplifyContext } from './testUtils/mockAmplifyContext';
+
+// Make global `Request` available.
+enableFetchMocks();
+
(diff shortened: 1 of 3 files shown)


osama-rizk (approved):
Approving — cleanest step of the series. The overload is behavior-preserving on the legacy path (verified all call shapes: `generateClient(options)` and `generateClient()` produce the same `{...options, amplify: Amplify}` as before; the `({} as any)` → `{}` swap is behavior-neutral), and the new `generateClient(ctx)` path binds to the branded context, which api-graphql's bridge no-ops correctly.

The tests are a model for the rest of the migration: distinct global/ctx/options configs + a spy on the real GraphQL POST prove config resolves from the context and explicitly NOT the global singleton (and options override both) — real request inspection rather than call-count proxies. Mock is correctly branded and resourcesConfig-wired. No findings.

soberm (commented):
Automated code review (comment-only — no blocking issues). Verdict: **comment-only**.

**Summary:** Adds a 2nd `generateClient()` overload to `@aws-amplify/api` accepting an explicit `AmplifyContext` instead of the global `Amplify` singleton (Task 10 of the v6 context migration, mirroring #14917/#14919). Adds a mock-context test util + 3 tests (singleton / ctx / ctx+options paths).

**Comments:**

- **[major] `packages/api/src/API.ts` — `isAmplifyContext` robustness:** The overload's correctness hinges on `isAmplifyContext`'s brand check safely returning `false` for arbitrary plain options objects (e.g. `{ authMode: 'apiKey' }`) without throwing. That guard is out-of-diff (from `@aws-amplify/core`). Please confirm it's tested upstream and consider adding a negative test here (`generateClient({ authMode: 'apiKey' })` still resolves the global singleton).
- **[minor] `packages/api/src/API.ts` — `options ?? {}`:** Switching `|| {}` → `?? {}` is a safe/intentional improvement; worth calling out explicitly in the PR description.
- **[minor] `packages/api/__tests__/generateClientContext.test.ts` — `getFirstPost()`:** It indexes `_postSpy.mock.calls[0][1]` without first asserting the spy was called → cryptic TypeError on failure. Add `expect(_postSpy).toHaveBeenCalledTimes(1)` before indexing.
- **[minor] test `_postSpy`:** The module-scope spy is never `mockRestore()`'d — consider `afterAll(() => _postSpy.mockRestore())` to avoid cross-file leakage.
- **[nit] JSDoc example:** `listTodos` isn't defined/imported in the new overload's JSDoc snippet — match the sibling overload's style or add a one-line note that it's a generated GraphQL document.
- **[nit] mock `post` return shape:** `mockReturnValue({ body: { json() {...} } })` returns a non-promise; add a one-line comment noting `await` handles both so nobody "fixes" it into `mockResolvedValue` incorrectly.
- **[nit] `mockAmplifyContext.ts`:** `resourcesConfig: ResourcesConfig = {}` default is unused in practice — consider making it required to force explicit config.

Overall: small, well-tested, consistent with the established #14917/#14919 pattern. Nothing blocks merging into the feature branch.

bobbor:
@soberm thanks for the thorough pass — addressed in 767a238df:

- **[major] `isAmplifyContext` robustness**: it's tested upstream in `packages/core/__tests__/context/contextBrand.test.ts` — explicit cases for `null`, `undefined`, plain `{}` (no brand), and non-objects, all returning `false` without throwing (the guard is a null/undefined/typeof-object check + `AMPLIFY_CONTEXT_BRAND in value`). Added the requested negative test here anyway: `generateClient({ authMode: 'apiKey' })` with a plain options object resolves config from the global singleton. (No cast needed — it type-routes to the options overload cleanly.)
- **[minor] `?? {}` callout**: added to the PR description.
- **[minor] `getFirstPost()`**: now asserts `expect(_postSpy).toHaveBeenCalledTimes(1)` before indexing.
- **[minor] spy leakage**: added `afterAll(() => _postSpy.mockRestore())`.
- **[nit] JSDoc example**: added a note that `listTodos` is a generated GraphQL query document.
- **[nit] mock `post` return shape**: added a comment explaining `await` resolves non-thenables as-is, so `mockReturnValue` is intentional.
- **[nit] `resourcesConfig = {}` default**: keeping as-is for now — the util is byte-consistent with api-graphql's reference copy; consolidating these into a shared `createMockAmplifyContext` in a core testing entry (where we can tighten the signature once) is already on the Phase-C follow-up ledger.

All checks green locally (98 tests passing).

osama-rizk (approved):

soberm (approved):

PR #14819, opened 2026-05-16
fix(aws-amplify): sync Cognito auth config and merge libraryOptions on reconfigure

ShrutiPundir17:
Solves issue : https://github.com/aws-amplify/amplify-js/issues/14820

#### Description of changes
When DefaultAmplify.configure runs again with a new resourcesConfig.Auth but without libraryOptions.Auth, the default cognitoUserPoolsTokenProvider was not always refreshed. TokenStore storage keys depend on authConfig.Cognito.userPoolClientId, so tokens could read/write under stale keys after reconfigure (related to #13707).

Change (packages/aws-amplify/src/initSingleton.ts):

Pass-through to core when !resolvedResourceConfig.Auth or libraryOptions?.Auth is provided (caller owns Auth wiring).
Singleton-managed path (Auth in config, no libraryOptions.Auth): on every configure/reconfigure:
cognitoUserPoolsTokenProvider.setAuthConfig(resolvedResourceConfig.Auth)
cognitoUserPoolsTokenProvider.setKeyValueStorage(...) (SSR-aware: cookies vs defaultStorage)
Amplify.configure with default token + credentials providers
Out of scope: merging libraryOptions across configure calls (discussed on #14820; tracked separately in #14815 / #14816).

Issue #, if available
Fixes #14820
Related: #13707, #14815, #14816

Description of how you validated changes
cd packages/aws-amplify
..\..\node_modules\.bin\jest.cmd initSingleton -w 1
Updated initSingleton.test.ts and initSingleton.integration.test.ts for pass-through vs refresh behavior.

Checklist

 PR description included

 yarn test passes (full monorepo — rely on CI)

 Unit Tests are [changed or added](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#steps-towards-contributions)

 Relevant documentation is changed or added
Checklist for repo maintainers

 Verify E2E tests for existing workflows are working as expected or add E2E tests for newly added workflows

 New source file paths included in this PR have been added to CODEOWNERS, if appropriate
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #14925, opened 2026-08-27
feat(core): Phase C server-path alignment — context-native server APIs

bobbor:
## Description

**Phase C of the v6→v7 context migration: server-path alignment.** The one non-additive split of the series — aligns the server path with the `AmplifyContext` model and removes the legacy singleton/ServerContext machinery. Lands as 5 stacked commits (C1 core → C4 categories → C2 aws-amplify → C3 adapter-nextjs → review polish) that must merge together: removing `ContextSpec`/registry from core breaks every server wrapper simultaneously, so the destructive core change and its dependents land as one coordinated PR.

### C1 — core (`6b4c2206f`)
- NEW `createAmplifyContext(resourcesConfig, libraryOptions?)` → branded + frozen `AmplifyContext` (core main entry)
- PROMOTED shared branded `bridgeAmplifyClass` into core internals (replaces the api-rest/api-graphql copies + storage inline bridge)
- REMOVED: `serverContext.ts`, `serverContextRegistry.ts`, `AmplifyServer.ContextSpec` (deprecated type alias `ContextSpec = AmplifyContext` kept on internals/adapter-core), server `fetchAuthSession`, `ADD_OAUTH_LISTENER`; `AmplifyServerContextError` moved to `core/src/errors` as `@deprecated`
- **F4.1**: unconfigured `getGlobalContext()` path now throws typed `NoAmplifyContextError`
- **F3.1**: central ctor guard (`InvalidAmplifyContextError`) applied to geo/predictions/pubsub Pattern-7 constructors
- **F3.2**: api-graphql `InternalGraphQLAPI` subscription-branch unchecked cast routed through the checked guard path

### C4 — category server-wrapper collapse (`68dc0e1a7`)
- Deleted all 4 per-package `resolveServerContext` files; server entries collapsed Pattern 5 → Pattern 6 (bare re-export of the ctx-native main APIs) across auth/storage/api-rest/api-graphql
- auth pure-cast bridging bug closed by deletion; **F5.1** storage `libraryOptions` snapshot bug dies with the inline bridge (core `bridgeAmplifyClass` is the single live-getter impl)
- auth `enableOAuthListener` rewritten to Hub `configure`-event wiring + global-context catch-up (no `ADD_OAUTH_LISTENER`)
- **F6.4**: interactions callback registry is now per-context (`WeakMap<AmplifyContext, …>`) with documented re-configure semantics + isolation test

### C2 — aws-amplify (`f9dade402`)
- NEW `src/configure.ts`: public `createAmplifyContext` — Gen2 outputs parsing + factory-style per-context Cognito providers (cookie storage for `ssr: true`), delegates to core for branding/freezing
- NEW `src/Amplify.ts`: singleton facade — builds branded ctx, `setGlobalContext`, dispatches the Hub `configure` event auth listens for
- REMOVED `initSingleton.ts` + `adapter-core/runWithAmplifyServerContext.ts` (deprecated shim kept inline on adapter-core resolving the global context)

### C3 — adapter-nextjs (`8730cbf7b`)
- `createRunWithAmplifyServerContext` builds a **per-request branded + frozen `AmplifyContext`** via `createAmplifyContext(resourcesConfig, { Auth: cookie-backed providers })` — no registry, no create/destroy
- `NextServer` operations receive `AmplifyContext` (param name `contextSpec` kept for source back-compat via the deprecated alias); `generateServerClient` + auth route handlers updated
- New isolation test: two concurrent requests get distinct branded contexts with zero cross-talk

### Review polish (`40c811cd5`)
- `bridgeAmplifyClass` output frozen (producer parity); mis-ordered-ctx guard throws typed error; stale comment + dead test mock removed

## Validation
- `yarn build`: 20/20 tasks green, no duplicated Amplify dependencies
- `yarn test`: **all 20 packages, 538 suites, 5580 passed / 23 skipped, 0 failures**
- `yarn test:size`: 27/27, all 53 entries under budget (no budget changes needed)
- `yarn lint`: clean on all 11 touched packages (react-native/rtn-passkeys config errors are pre-existing on base, untouched by this PR)
- No `as any`/`@ts-ignore` added in source; guard tests assert typed errors
- Post-implementation architectural review: APPROVE, no critical findings (context isolation, branding invariants, removal completeness, OAuth-listener races, back-compat aliases all verified)

## Documented deviations
- api-graphql server `generateClient` keeps a thin ctx-binding wrapper (bare re-export would break its server suite; wrapper already binds the branded ctx) — post-C follow-up
- storage server `uploadData` keeps its server-specific task type (unified type is an out-of-scope refactor)
- Reconfigure with only a resource config preserves previously-resolved libraryOptions (v7-poc shape)
- adapter-nextjs auth route handlers still use `AmplifyServerContextError` via the deprecated re-export — post-C follow-up to migrate to typed `AmplifyError` names

## Checklist
- [x] PR targets the `feat/bobbor/v6-context` feature branch (no changeset required)
- [x] Unit tests added/updated for all changed code
- [x] Full monorepo build + test + size validation green


packages/aws-amplify/src/Amplify.ts
```diff
@@ -0,0 +1,200 @@
...
+	 * `amplifyconfiguration.json` / `amplify_outputs.json` file. It can also be an object literal
+	 * created inline when calling `Amplify.configure`.
+	 * @param libraryOptions The {@link LibraryOptions} additional options for the library.
+	 *
+	 * @example
+	 * import config from './amplifyconfiguration.json';
+	 *
+	 * Amplify.configure(config);
+	 */
+	configure(
```
soberm, line 73: [question] Now that `configure()` builds the context directly via `createAmplifyContext()` instead of delegating to `@aws-amplify/core`'s own `AmplifyClass.configure()` (like the old `DefaultAmplify` did), the core singleton's own `resourcesConfig`/`libraryOptions`/`isConfigured` fields never get populated when an app only configures through this facade. Both paths still write the same shared `_globalContext`, so anything resolving through `getGlobalContext()`/`AmplifyContext` is fine — but is there any remaining code path (internal or third-party) that reads `AmplifyClassV6` instance fields directly rather than going through the context? Worth confirming explicitly before merge, given how many places historically imported `Amplify` straight from `@aws-amplify/core`.

packages/interactions/__tests__/lex-v1/AWSLexProvider.test.ts
```diff
@@ -368,20 +368,28 @@ describe('Interactions', () => {
...
+				completeSuccessResp,
+				botConfig.BookTrip,
+			);
 			jest.runAllTimers();
 			expect(completeSuccessCallback).toHaveBeenCalledTimes(1);
 
 			// 3. task complete; error, callback be called with error
-			provider.onComplete(botConfig.BookTrip, completeFailCallback);
-			provider.reportBotStatus(completeFailResp, botConfig.BookTrip);
+			provider.onComplete(mockCtx, botConfig.BookTrip, completeFailCallback);
```
soberm, line 392: [minor] `AWSLexV2Provider.test.ts` got two new tests for this exact fix ("callbacks are isolated per resolved context (F6.4)" and "re-configure replaces the context object..."), but this file only threads `mockCtx` through the existing tests — the v1 provider has the identical `WeakMap`-based isolation fix but no isolation-specific coverage. Might be worth copying the two new v2 tests over to this describe block too, since it's the same bug/fix in a near-identical implementation.

packages/aws-amplify/src/configure.ts
```diff
@@ -0,0 +1,121 @@
...
+ * const ctx = createAmplifyContext(outputs);
+ * // Pass ctx explicitly to category APIs:
+ * await signIn(ctx, { username, password });
+ * ```
+ */
+export function createAmplifyContext(
+	resourceConfig: ResourcesConfig | LegacyConfig | AmplifyOutputsUnknown,
+	libraryOptions?: LibraryOptions,
+): AmplifyContext {
+	const resolvedResourceConfig = parseAmplifyConfig(resourceConfig);
```
soberm, line 66: [nit] This parses `resourceConfig` here, then hands the already-parsed config to `createCoreAmplifyContext()` below, which parses it again in `packages/core/src/context/createAmplifyContext.ts`. Probably harmless if `parseAmplifyConfig` is idempotent, but it's redundant work on every call — and `createRunWithAmplifyServerContext` in adapter-nextjs calls this per-request, so it's now parsing the same config twice per SSR request instead of once.

packages/aws-amplify/src/adapter-core/index.ts
```diff
@@ -26,3 +24,38 @@ export {
...
+ * Runs an operation with an {@link AmplifyContext}.
+ *
+ * @deprecated The server-context registry has been removed in favor of the
+ * branded {@link AmplifyContext} model. Prefer building a context explicitly
+ * with `createAmplifyContext()` (for per-request isolation) or calling category
+ * APIs directly against the global context. This shim is retained for
+ * backwards compatibility and simply resolves the current global context
+ * before invoking `operation`.
+ */
+export async function runWithAmplifyServerContext<T>(input: {
```
soberm, line 59: [minor] This keeps the `runWithAmplifyServerContext` name but the behavior is now the opposite of what it used to do — the old version built a fresh, isolated context per call from explicit config; this one just grabs whatever the current global context happens to be (`getGlobalContext()` below), so there's no per-call isolation anymore. It's clearly marked `@deprecated` with an accurate comment, and the signature change means old callers fail loudly rather than silently losing isolation, so I don't think it's actively dangerous — but reusing the same name for inverted semantics feels like a foot-gun for anyone hand-rolling a custom SSR integration without reading the deprecation note closely.

packages/auth/src/providers/cognito/utils/oauth/enableOAuthListener.ts
```diff
@@ -1,21 +1,46 @@
...
+// Attach the side effect for handling the completion of an inflight OAuth flow.
+// This side effect works only on Web.
+//
+// Phase C4: the singleton `ADD_OAUTH_LISTENER` hook was removed from core. The
+// listener is now driven entirely by the `configure` Hub event that
+// `Amplify.configure()` publishes (after it sets the global context), which
+// preserves the previous browser behavior: the OAuth flow is completed as soon
+// as Amplify is configured with an OAuth-enabled Cognito config.
+if (isBrowser()) {
+	Hub.listen('core', ({ payload }) => {
```
soberm, line 64: [minor] Behavior change worth calling out: the old `ADD_OAUTH_LISTENER` hook fired `attemptCompleteOAuthFlow` at most once ever (it self-cleared after the first oauth-enabled configure). This `Hub.listen` subscription is permanent, so it'll fire on every single `configure()` call that has oauth enabled, not just the first. Checked `attemptCompleteOAuthFlow` — it no-ops via `oAuthStore.loadOAuthInFlight()` when nothing's in flight, so repeat calls across reconfigures look safe, but might be worth a test covering the multi-reconfigure case since it's a real change from the old "fires once" contract.

packages/api-graphql/src/internals/InternalGraphQLAPI.ts
```diff
@@ -182,8 +179,19 @@ export class InternalGraphQLAPIClass {
 				return responsePromise;
 			}
 			case 'subscription':
+				// Use the same checked guard as the query/mutation branch instead of
+				// an unchecked cast. The callback-function form (used only by the
+				// server context manager for await-able query/mutation operations) is
+				// not supported for subscriptions, so reject it explicitly rather than
+				// casting it into the context union.
+				if (!isAmplifyInstance(amplify)) {
```
soberm, line 188: [nit] Nice fix replacing the unchecked cast with this guard, but the thrown error here is a bare `Error` rather than one of the new typed `AmplifyError` subclasses this same PR introduces elsewhere (`InvalidAmplifyContextError`, `NoAmplifyContextError`). Consistent with the `default:` case a few lines down, just not with the broader "typed, catchable errors" theme of the PR.

packages/aws-amplify/src/adapter-core/AmplifyServer.ts
```diff
@@ -0,0 +1,14 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { AmplifyContext } from '@aws-amplify/core';
+
+/**
+ * @deprecated Use {@link AmplifyContext} directly instead. This namespace is
+ * retained only for backwards compatibility with consumers (e.g. adapter
+ * packages) that still reference `AmplifyServer.ContextSpec`.
+ */
+export declare namespace AmplifyServer {
```
soberm, line 18: [nit] This only preserves `AmplifyServer.ContextSpec` for back-compat. The deleted `types/amplifyServer.ts` (core) also had `Context`, `ContextToken`, and `RunOperationWithContext` — anyone who imported those specific type names directly gets a hard compile break with no alias. Probably a narrow/rare case, but the PR description calls out "back-compat aliases... verified" so flagging the gap.

packages/core/src/context/bridgeAmplifyClass.ts
```diff
@@ -0,0 +1,63 @@
...
+		clearCredentials: () => amplify.Auth.clearCredentials(),
+		// getTokens needs no default — options is required on both interfaces.
+		getTokens: options => amplify.Auth.getTokens(options),
+	};
+
+	// Branding makes downstream isAmplifyContext checks (e.g. internalPost)
+	// recognize an already-bridged context, preventing double-bridging which
+	// would break the Auth.* closures. Explicit descriptor (non-writable /
+	// non-configurable) mirrors the other AmplifyContext producers.
+	Object.defineProperty(resolved, AMPLIFY_CONTEXT_BRAND, {
```
soberm, line 48: [nit] Worth noting the trade-off baked into `AMPLIFY_CONTEXT_BRAND`: it's `Symbol.for('amplify.context')` now (a well-known global-registry symbol) rather than a per-call unique `Symbol()`, so any code in the same realm can construct an object that passes `isAmplifyContext()` without ever holding a real context. That's a deliberate, reasonable trade-off to survive duplicate module instances in a bundle, and exploitability is low given what else you'd need in-realm to make it matter — just flagging it as a trade-off rather than a free lunch.

packages/adapter-nextjs/src/api/generateServerClient.ts
```diff
@@ -43,12 +40,13 @@ export function generateServerClientUsingCookies<
 		CookiesClientParams = DefaultCommonClientOptions & CookiesClientParams,
 >(options: Options): V6ClientSSRCookies<T, Options> {
 	if (typeof options.cookies !== 'function') {
-		throw new AmplifyServerContextError({
+		throw new AmplifyError({
```
soberm, line 43: [nit] Small breaking change: this used to throw `AmplifyServerContextError`, now it's a generic `AmplifyError` with `name: 'InvalidCookiesError'`. Reasonable cleanup since the old type was a semantic mismatch anyway (and it's `@deprecated` now), but anyone narrowly catching `instanceof AmplifyServerContextError` at this specific call site will stop matching.

packages/interactions/src/lex-v1/AWSLexProvider.ts
```diff
@@ -34,21 +34,37 @@ type AWSLexProviderSendResponse =
...
+	//
+	// NOTE (re-configure semantics): `Amplify.configure()` publishes a NEW frozen
+	// global context object on every call. A callback registered against the
+	// global context before a re-configure is therefore keyed on the OLD context
+	// object and will NOT be found after re-configure (the new global context is
+	// a different object). Callers relying on the global context must re-register
+	// `onComplete` after reconfiguring; callers passing an explicit context keep
+	// their callbacks for that context's lifetime. Using a `WeakMap` lets the
+	// per-context records be garbage-collected once a context is unreferenced.
+	private readonly _botsCompleteCallbackByCtx = new WeakMap<
```
soberm, line 51: [nit] This WeakMap isolation fix and its explanatory comment are duplicated close to verbatim in `AWSLexV2Provider.ts`. Not new duplication introduced by this PR necessarily (v1/v2 have always been parallel implementations), but this was a chance to pull the per-context callback registry into one shared helper instead of growing the copy-paste further.

soberm (commented):
Reviewed the full diff (110 files) plus several unchanged files it touches indirectly (`AmplifyContext.ts`, `contextBrand.ts`, `InternalGraphQLAPI.ts`, `api-rest/apis/index.ts`, `attemptCompleteOAuthFlow.ts`) to check call chains rather than trust hunks in isolation.

No blocking issues. This is a clean, well-tested consolidation — branding/freezing is handled correctly (including the freeze-after-getters subtlety in `bridgeAmplifyClass`), the "Pattern 6" server-wrapper collapse doesn't lose the ctx-aware overloads (verified against `api-rest/apis/index.ts`), and the refactor closes a couple of real bugs along the way (the Lex `onComplete` cross-context callback leak, the storage `libraryOptions` staleness bug). The adapter-nextjs concurrency test is exactly the kind of coverage this change needs.

Left a handful of inline comments below — one open architectural question worth confirming before merge, one real test-coverage gap, and some smaller nits.

packages/aws-amplify/src/Amplify.ts
```diff
@@ -0,0 +1,200 @@
...
+ *   and inject the singleton token/credentials providers.
+ */
+function resolveLibraryOptions(
+	resourceConfig: ResourcesConfig,
+	libraryOptions?: LibraryOptions,
+	previousLibraryOptions?: LibraryOptions,
+): LibraryOptions {
+	// If no new libraryOptions were provided, preserve the previously configured
+	// ones (the core AmplifyClass singleton used to perform this merge).
+	if (!libraryOptions && previousLibraryOptions) {
```
osama-rizk: When `configure()` is called a second time with a new `resourcesConfig.Auth` (e.g. a different `userPoolId`) but no `libraryOptions`, this early-return preserves the previous options and skips the `cognitoUserPoolsTokenProvider.setAuthConfig(resourceConfig.Auth)` call below (line 187).

The fresh `AuthClass` gets the new config, but `AuthClass.getTokens` delegates straight to `tokenProvider.getTokens(options)` without passing `authConfig` (`core/src/singleton/Auth/index.ts:105-111`) — the token provider reads its own `authConfig`, which is only ever set by that skipped `setAuthConfig`. Net effect: after `configure(A); configure(B)`, user-pool **token refresh still targets pool A**, while `fetchAuthSession`'s credentials path and `getConfig()` both reflect B (which masks the problem).

The base behavior (`initSingleton.ts` on the base branch) always ran `setAuthConfig` when `Auth` was present and `libraryOptions.Auth` was absent, so this is a behavior change. Suggestion: when preserving previous `libraryOptions`, still push the new `resourceConfig.Auth` into the singleton token provider (`setAuthConfig` + `setKeyValueStorage`). If configure-once is the intended v7 semantics instead, worth stating explicitly since it drops a v6 capability.

Note: `Amplify.test.ts:201` exercises this exact path but only asserts `getConfig()`/`libraryOptions`, so it passes without catching the stale-pool case — worth extending to assert the token provider received the new pool.

packages/auth/src/providers/cognito/utils/oauth/enableOAuthListener.ts
osama-rizk, line 64: This rewrite fires `attemptCompleteOAuthFlow` on every `configure` event and also once at import time (catch-up), with no fire-once guard. `attemptCompleteOAuthFlow` has no synchronous re-entry lock — its only guard is `await oAuthStore.loadOAuthInFlight()`, and both that flag and the URL `code` are cleared only at the end of `completeOAuthFlow` (after the token-endpoint fetch).

Concurrent window: the catch-up call runs (config was set before this module imported) and is mid-fetch when a second `Amplify.configure()` (framework double-config, React StrictMode, HMR) fires the Hub listener. Both invocations pass the inflight guard and POST the same authorization code. OAuth2 codes are single-use, so the second exchange returns `invalid_grant` → `handleFailure` dispatches a `signInWithRedirect` **failure** Hub event even though sign-in actually succeeded — apps listening for that event surface a spurious error.

The previous core singleton fired the listener at most once per lifecycle (`notifyOAuthListener` nulled `oAuthListener` — "should only be notified once"). Suggestion: an in-memory handling/handled flag, or claim/clear the inflight marker synchronously before the fetch. Steady-state re-fires after completion are already harmless.

Coverage: `enableOAuthListener.test.ts` covers only the single happy-path listener — neither the catch-up branch nor a multi-`configure` sequence is tested.

packages/aws-amplify/src/adapter-core/index.ts
osama-rizk, line 59: This retained shim resolves the process-wide global context (`getGlobalContext()`) and its overloads dropped the authenticated `{ nextServerContext: reqRes }` shape. That's fine for the deprecation — real per-request auth is correctly forced onto the adapter-nextjs path, and this never stores or mutates request state. But a legacy JS caller who previously drove per-request auth through this entry will now silently share one global context across requests. Worth an explicit note in the `@deprecated` block that the request-scoped shape is gone and this returns the global context only. (TS already blocks the authenticated overload, so this is about the untyped-JS caller.)

packages/pubsub/src/Providers/AWSIot.ts
```diff
@@ -29,6 +32,14 @@ export class AWSIoT extends MqttOverWS {
 			super(maybeOptions ?? {});
 			this._explicitCtx = ctxOrOptions;
 		} else {
+			// When a second argument is supplied the caller used the
+			// `(ctx, options)` overload, so the first argument is required to be a
+			// branded AmplifyContext. Guard against a defined-but-unbranded value
+			// being silently swallowed as `options`. In the single-argument form
+			// `ctxOrOptions` is legitimately the options object, so no assertion.
+			if (maybeOptions !== undefined) {
```
osama-rizk, line 40: Minor edge case: `assertOptionalCtxArg` allows `undefined` in the context slot, so `new AWSIoT(undefined, options)` passes the guard and then `super(ctxOrOptions ?? {})` discards `maybeOptions`. Not a documented call shape (options-first is the real API), so low priority — but silently dropping the second arg could confuse. Consider forwarding `maybeOptions` when the first arg is `undefined`.

osama-rizk (commented):
Nicely structured migration — net reduction in code and duplication (one promoted `bridgeAmplifyClass`, the deleted per-package resolvers), and the description matches the implementation.

I verified the properties this PR rests on and they hold: SSR per-request **context isolation** is sound (no server path sets or falls back to global state; per-request `AuthClass` and token/credentials providers are freshly instantiated with no shared store captured across contexts; the concurrency test genuinely exercises this). **Removal is complete** with no dangling runtime references, and the `ContextSpec`/`AmplifyServerContextError` back-compat re-exports resolve consistently. The `Symbol.for`-based branding is a good choice for surviving duplicate-core bundling.

Two logic regressions are worth resolving before this rides into v7 consumers (inline): a stale-user-pool case on reconfigure, and a double auth-code-redemption race in the OAuth listener. Both are timing/logic issues the current unit tests don't cover, so green CI doesn't clear them.

packages/aws-amplify/src/Amplify.ts
bobbor: Great catch — and it's worse than a deviation: this re-introduced the exact pre-#14819 bug that mainline already fixed (`6f5a076ee`, "sync Cognito auth config on reconfigure", which IS in this branch's base). Fixed in 13b971b56 mirroring the #14819 contract: when `resourceConfig.Auth` is present and the caller doesn't supply `libraryOptions.Auth`, `setAuthConfig` + `setKeyValueStorage` now run on **every** configure call (storage resolved from the effective `ssr` flag), while non-Auth libraryOptions are still preserved across reconfigures. Also ported #14819's regression tests: the unit test you flagged (Amplify.test.ts:201) now asserts the token provider received the new pool, and the integration suite covers configure(poolA)→configure(poolB) against the real singleton plus a non-Auth-options-preservation case. `createAmplifyContext` (per-context path) was audited and is unaffected — it builds fresh providers per call with no preservation/early-return.

packages/auth/src/providers/cognito/utils/oauth/enableOAuthListener.ts
bobbor: Fixed in b6f80f83f with the localized guard you suggested: a module-level `isHandlingOAuthFlow` flag claimed **synchronously** in a shared `attemptCompleteOAuthFlowOnce()` wrapper (released in `.finally()`), which both the Hub listener and the import-time catch-up funnel through — concurrent triggers collapse to one execution, so the single-use code is exchanged at most once and no spurious `signInWithRedirect` failure fires. Chose once-per-**cycle** rather than once-ever: after a completed flow the re-fire path stays possible and no-ops via `loadOAuthInFlight()` as before, which preserves the old observable at-most-once-completion contract without permanently disabling the listener. Added the concurrency tests: overlapping double/triple configure with a controllable pending promise → exactly one execution; catch-up + configure overlap; sequential multi-reconfigure. `attemptCompleteOAuthFlow` itself untouched.

bobbor: Multi-reconfigure test added in b6f80f83f (sequential 3× configure case plus overlapping-concurrent variants). The fire-once question osama raised on the same lines is addressed there too — see the guard + contract note in the code comment: concurrent triggers collapse to one execution, steady-state re-fires stay harmless no-ops.

packages/pubsub/src/Providers/AWSIot.ts
bobbor: Fixed in 7c1b79adb: the else branch now forwards `super(ctxOrOptions ?? maybeOptions ?? {})`, so `new AWSIoT(undefined, options)` honors the options (overload widened to accept the explicit-undefined form; test added). Checked geo/predictions for the same edge — their constructors take only a single ctx param, no options slot to drop.

packages/interactions/__tests__/lex-v1/AWSLexProvider.test.ts
bobbor: Done in 7c1b79adb — both F6.4 isolation tests ported to the lex-v1 suite (adapted to v1's reportBotStatus/onComplete API), in a dedicated context-isolation describe block.

packages/aws-amplify/src/configure.ts
bobbor: Fixed in 7c1b79adb: core's `createAmplifyContext` now accepts an internal `skipConfigParse` option; `configure.ts` passes it since it already parses for provider injection. Parse now runs once per call (added a core test spying on `parseAmplifyConfig` to prove it), so the adapter-nextjs per-request path no longer double-parses.

packages/aws-amplify/src/adapter-core/index.ts
bobbor: Agreed on the foot-gun. Strengthened the @deprecated JSDoc in 7c1b79adb to state explicitly: semantics are inverted vs the old API (resolves the process-wide global context, no per-call isolation), the request-scoped `nextServerContext` shape is gone, and adapter-nextjs `createRunWithAmplifyServerContext` is the correct per-request path. (Also covers osama's note on the same lines.)

bobbor: Done in 7c1b79adb — the @deprecated block now explicitly warns that the request-scoped `{ nextServerContext }` shape is gone and this resolves the process-wide global context only, with a pointer to the adapter-nextjs per-request path for untyped-JS callers.

packages/api-graphql/src/internals/InternalGraphQLAPI.ts
bobbor: Fixed in 7c1b79adb — the subscription-branch guard now throws `InvalidAmplifyContextError` (message unchanged), consistent with the typed-error theme; test updated to assert the type.

packages/aws-amplify/src/adapter-core/AmplifyServer.ts
bobbor: Good catch — the alias gap was real. Fixed in 7c1b79adb: `AmplifyServer.Context`, `ContextToken`, and `RunOperationWithContext` are back as @deprecated type aliases mapped onto the ctx model (shapes faithful to the deleted originals), reachable from the same adapter-core entry.

packages/aws-amplify/src/Amplify.ts
bobbor: Excellent question — we grepped the whole monorepo and the concern was REAL. Confirmed the facade never writes the core singleton (by design: the global context is the single source of truth), and found six src paths still reading the singleton's instance state, which would see empty config under facade-only configuration: (1) `generateClient()`'s no-context fallback in `api/src/API.ts` — the most user-facing one, (2–4) the three WebAuthn auth APIs (`associate`/`delete`/`listWebAuthnCredentials`) and their foundation fns, (5) `cognitoUserPoolTransferHandler`'s custom `Auth.headers` read, (6) datastore (deprecated — left as-is with a comment). Existing tests masked this by configuring the core singleton directly. All fixed in 99ca761b2: these paths now resolve `getGlobalContext()` (lazily, per operation), with new facade-parity regression tests that set ONLY a global context and assert the APIs resolve config (incl. a reconfigure case for generateClient). So: not safe before, safe now.

packages/core/src/context/bridgeAmplifyClass.ts
bobbor: Agreed and acknowledged — `Symbol.for` is the deliberate trade-off to survive duplicate module instances in bundled apps (the alternative, a unique Symbol, broke brand checks across duplicated core copies). The brand is an integrity marker, not a security boundary; anything in-realm that can forge the brand can already reach the real context. Keeping as-is.

packages/adapter-nextjs/src/api/generateServerClient.ts
bobbor: Acknowledged — intended cleanup: `AmplifyServerContextError` was a semantic mismatch there and is @deprecated now. Callers narrowly catching it at this call site will need to match `AmplifyError` with `name: 'InvalidCookiesError'`; we'll include this in the Phase C migration notes.

packages/interactions/src/lex-v1/AWSLexProvider.ts
bobbor: Agreed — v1/v2 have always been parallel copies and this grew the duplication by one more block. Extracting the per-context callback registry into a shared helper is queued in the post-C fit-and-finish ledger (alongside the shared lazy-ctx getter and createMockAmplifyContext consolidation) rather than this PR, to keep the destructive split reviewable.

osama-rizk (approved):

soberm (approved):

PR #14929, opened 2026-08-31
fix: mark v5 public APIs as deprecated for maintenance mode

soberm:
## What

Amplify JavaScript **v5 is in maintenance mode**. This PR annotates the **public API surface of every publishable package** with a TSDoc `@deprecated` tag pointing users to the v5 → v6 migration guide:

> https://docs.amplify.aws/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/

The notice surfaces directly in users' editors (IDE strikethrough + hover text) and in the generated `.d.ts` files, so consumers learn about the upgrade path at the point of use rather than having to find a release note.

Every tag uses the same wording, applied consistently:

```ts
/**
 * @deprecated Amplify JavaScript v5 is in maintenance mode. Upgrade to v6.
 * See the migration guide:
 * https://docs.amplify.aws/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/
 */
export class StorageHelper { ... }
```

## Scope of the change

**Comment-only. No runtime behavior, signatures, or logic are modified.**

- `117 files changed, 2526 insertions(+), 2 deletions(-)`
- **520 annotations** — 516 new `@deprecated` tags, plus 4 merged into pre-existing `@deprecated` tags
- The only two non-additive edits expand single-line JSDoc blocks (`/** @class */`, `/** @enum {string} */`) into multi-line blocks so a tag could be added without corrupting them.

Where a symbol already had an `@deprecated` tag (e.g. the Amazon Pinpoint end-of-support notices), the migration pointer was **appended to the existing tag** rather than added as a second tag — a duplicate tag would have been ignored by TypeScript and would have hidden the original note:

```ts
/**
 * @deprecated AWS will end support for Amazon Pinpoint on October 30, 2026.
 * Amplify JavaScript v5 is in maintenance mode. Upgrade to v6.
 * See the migration guide:
 * https://docs.amplify.aws/...
 */
```

## How the public API surface was determined

None of the packages declare an `exports` map, so for each package the public surface is the set of symbols re-exported from its public entry point (`main`/`module`/`typings` all resolve to `src/index.ts`). That set was resolved with the TypeScript compiler API rather than by grepping, then each symbol was annotated **at its original declaration site** so the tag flows through re-export chains. Internal modules that are not reachable from an entry point were deliberately left alone.

Two packages needed special handling:

- **`amazon-cognito-identity-js`** ships JavaScript sources with a hand-written `index.d.ts` that wraps its whole surface in `declare module 'amazon-cognito-identity-js' { ... }`. That ambient declaration file *is* the consumer-facing type surface, so the tags were added there.
- **`aws-amplify`** (the umbrella package) is almost entirely re-exports of its siblings. Only its own `withSSRContext` is annotated locally; the other 31 symbols inherit the tag from the sibling declarations. This was verified, not assumed (see below).

## Coverage

| Package | Public exports | Annotated |
|---|---|---|
| `amazon-cognito-identity-js` | 30 | 30 |
| `@aws-amplify/analytics` | 6 | 6 |
| `@aws-amplify/api` | 8 | 8 |
| `@aws-amplify/api-graphql` | 10 | 9 ¹ |
| `@aws-amplify/api-rest` | 3 | 3 |
| `@aws-amplify/auth` | 9 | 9 |
| `aws-amplify` | 32 | 32 (inherited) |
| `aws-amplify-react-native` | 29 | 29 ² |
| `@aws-amplify/cache` | 4 | 4 |
| `@aws-amplify/core` | 80 | 80 |
| `@aws-amplify/datastore` | 149 | 149 |
| `@aws-amplify/datastore-storage-adapter` | 1 | 1 |
| `@aws-amplify/geo` | 42 | 42 |
| `@aws-amplify/interactions` | 15 | 15 |
| `@aws-amplify/notifications` | 19 | 19 |
| `@aws-amplify/predictions` | 62 | 62 |
| `@aws-amplify/pubsub` | 13 | 13 |
| `@aws-amplify/pushnotification` | 1 | 1 |
| `@aws-amplify/rtn-push-notification` | 2 | 2 |
| `@aws-amplify/storage` | 48 | 48 |

All 20 publishable packages are covered; **no package was skipped.**

¹ `OperationTypeNode` is re-exported straight from the third-party `graphql` package. It is intentionally **not** annotated, since that would mean editing `node_modules`.

² `aws-amplify-react-native` is Babel-built and publishes no type declarations (no `typings` field), so its tags are verified at the source level rather than through emitted `.d.ts`.

## Verification

Run with Node 24 (matching CI) and driven through the `yarn` scripts:

- **`yarn build`** — exit 0, all 20 packages built, "No duplicated Amplify dependencies detected."
- **`lerna run test`** — exit 0, all 20 packages, **2212 tests passed, 0 failures**.
- **`lerna run lint`** — exit 0, clean in all 19 linted packages.
- **Consumer-perspective type check** — a synthetic consumer (`import * as ns from '<pkg>'`) was compiled against each package's **built** declarations and every resolved export was asserted to carry a `@deprecated` tag containing the migration URL: **533/534**, the single exception being the third-party `OperationTypeNode` above. This is what confirms the `aws-amplify` umbrella re-exports genuinely inherit the tag after a real build.

Two things worth calling out, both checked against a clean baseline and **pre-existing, not caused by this PR**:

- `yarn test:license` reports failures only for files inside gitignored local build/tooling directories (`.turbo/`, `.husky/_/`); the count is identical before and after this change. No source file changed here is missing its license header.
- `prettier --check` flags 20 of the touched files — the identical set and count fails on `v5-stable` before this change. The lines added by this PR are Prettier-clean; formatting those files would have added unrelated churn, so it was left out.

All added lines were also kept within the 120-character `max-line-length` that the packages' `tslint.json` enforces, which is why the message is wrapped across three comment lines.

## Release type

The commit is typed `fix:`, so Lerna will cut a **patch version bump** for the affected packages on the next `v5-stable` release — meaning **these deprecation notices will publish to npm and reach consumers**, rather than sitting unreleased.

`AGENTS.md` forbids `feat:` on this maintenance branch; `fix:` is permitted, and is the correct type here given the intent is for users to actually receive the upgrade notice. The change itself remains comment-only — no runtime behavior, signatures, or logic are modified.

## Deferred scope

- **`OperationTypeNode`** (`@aws-amplify/api-graphql`) — owned by the third-party `graphql` package; cannot be annotated from this repo.
- Only **public entry-point exports** are annotated. Internal/private modules and individual class *members* are not tagged; the deprecation is applied at the level of each exported symbol, which is what consumers import.
- Generated API docs under `docs/api` were not regenerated, as that is produced by the release pipeline.

### DIFF
--- a/packages/amazon-cognito-identity-js/index.d.ts
+++ b/packages/amazon-cognito-identity-js/index.d.ts
@@ -1,24 +1,54 @@
 declare module 'amazon-cognito-identity-js' {
 	//import * as AWS from "aws-sdk";
 
+	/**
+	 * @deprecated Amplify JavaScript v5 is in maintenance mode. Upgrade to v6.
+	 * See the migration guide:
+	 * https://docs.amplify.aws/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/
+	 */
 	export type NodeCallback<E, T> = (err?: E, result?: T) => void;
+	/**
+	 * @deprecated Amplify JavaScript v5 is in maintenance mode. Upgrade to v6.
+	 * See the migration guide:
+	 * https://docs.amplify.aws/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/
+	 */
 	export type UpdateAttributesNodeCallback<E, T, K> = (
 		err?: E,
 		result?: T,
 		details?: K
 	) => void;
+	/**
+	 * @deprecated Amplify JavaScript v5 is in maintenance mode. Upgrade to v6.
+	 * See the migration guide:
+	 * https://docs.amplify.aws/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/
+	 */
 	export namespace NodeCallback {
 		export type Any = NodeCallback<Error | undefined, any>;
 	}
 
+	/**
+	 * @deprecated Amplify JavaScript v5 is in maintenance mode. Upgrade to v6.
+	 * See the migration guide:
+	 * https://docs.amplify.aws/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/
+	 */
 	export interface CodeDeliveryDetails {
 		AttributeName: string;
 		DeliveryMedium: string;
 		Destination: string;
 	}
 
+	/**
+	 * @deprecated Amplify JavaScript v5 is in maintenance mode. Upgrade to v6.
+	 * See the migration guide:
+	 * https://docs.amplify.aws/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/
+	 */
 	export type ClientMetadata = { [key: string]: string } | undefined;
 
+	/**
+	 * @deprecated Amplify JavaScript v5 is in maintenance mode. Upgrade to v6.
+	 * See the migration guide:
+	 * https://docs.amplify.aws/gen1/javascript/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/
+	 */
 	export interface IAuthenticationCallback {
 		onSuccess: (
 			session: CognitoUserSession,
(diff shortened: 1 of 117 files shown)


osama-rizk (approved):

bobbor (approved):
Approving after mechanical verification of the diff:

- **Comment-only confirmed**: all 2,526 added lines are JSDoc-shaped (`/**`, `*`, `*/`); zero code lines added. The only 2 deletions are the stated single-line JSDoc expansions (`/** @class */`, `/** @enum {string} */`).
- **Tag accounting matches**: 516 new `@deprecated` tags + exactly 4 sentences merged into pre-existing tags (3× Pinpoint end-of-support, 1× bare `@deprecated`) = 520.
- **Consistency**: exactly 1 wording variant and 1 URL variant across all 520; longest added line is 110 chars (< 120 lint limit).
- **Special handling checks out**: `amazon-cognito-identity-js` touches only `index.d.ts`; `aws-amplify` only `withSSRContext.ts`. Spot-checked `api-rest`: all 3 public exports tagged at their declaration sites.
- **Release**: single `fix:` commit → Lerna patch bump on `v5-stable` (changeset-bot warning is a false positive; this branch doesn't use changesets). All 28 CI checks green.

One heads-up for the release notes: consumers running `eslint-plugin-deprecation`-style rules in CI will start flagging every Amplify v5 usage after this publishes — intended, but worth calling out.

PR #14930, opened 2026-09-01
refactor(core): post-migration fit-and-finish — shared helpers, typed errors, test dedup

bobbor:
## Description

**Post-migration fit-and-finish** — the complete follow-up ledger promised across the v6-context migration review threads (#14836, #14912, #14913, #14915–#14919, #14925), shipped as one PR now that Phase C is merged. No functional surface changes except those disclosed below; everything else is deduplication onto shared helpers and test-suite cleanup.

### New shared core helpers (wave 1)
- **`@aws-amplify/core/internals/testing`** (F1.1/F1.2/F1.3): single `createMockAmplifyContext` — branded, not frozen, `Partial<ResourcesConfig>` config, auth fns typed as `jest.Mock` (type-only, no runtime jest dep), `resourcesConfig` is getter-only (documented) — plus `withTokens(ctx, accessToken)` and brand/`isAmplifyContext` re-exports
- **`createCtxResolver(explicitCtx?)`** (F2.1): the explicit-ctx + lazy-global-resolution pattern, resolved fresh per call (no constructor capture)
- **`selectSsrKeyValueStorage(ssr?)`** (P3): single source of truth for the ssr→CookieStorage selection, extracted faithfully from `resolveLibraryOptions`
- analytics **`peekCtxArgs<TInput>`** (F2.2) + interactions **`perContextCallbackRegistry`** (P4)

### Adoption + test migration (wave 2)
- geo/pubsub/predictions adopt `createCtxResolver` (constructor guards, overloads, and `(undefined, options)` forwarding preserved verbatim); `Geo.ts` intentionally untouched (explicit-pin field, not a lazy getter)
- aws-amplify `Amplify.ts` + `configure.ts` adopt `selectSsrKeyValueStorage` — factory-vs-singleton provider split preserved; **the #14819 reconfigure regression tests are unmodified in this diff**
- ~140 test files across 10 packages migrated to the shared testing entry; 8 local `createMockAmplifyContext` copies deleted, zero forks/wrappers needed
- auth: 49 redundant `as jest.Mock` casts removed; `withTokens` adopted at 31 sites — **audited**: only one internal code path reads `ctx.getTokens` and none of the affected APIs reach it, so priming both fns does not change what any test proves
- datastore shared test helpers (F1.4); predictions user-agent assertions via public `config.customUserAgent` (F1.5)

### Behavior changes — bug fixes only (no breaking public contracts)

Per maintainer decision (`e9b7a089b`), this PR carries **no breaking public-API changes**. Earlier revisions of this branch changed three public contracts; all three were made backward-compatible:

- **notifications**: the "not supported" stub unification was reverted — `initializePushNotifications` stubs are byte-identical to base (`(): void`, sync throw; native sync). Only kept: dead `resolveCtxArgs` removal (internal) and the `removeDevice` web stub now *rejecting* instead of sync-throwing, which matches its **declared** base type `Promise<void>` (the sync throw was a base bug — a `.catch`-chaining caller got an uncaught exception).
- **adapter-nextjs**: the typed-error swap was reverted — the three auth/cookie sites throw `AmplifyServerContextError` exactly as base (`instanceof`/name matchers unaffected).
- **`fetchAuthSession()`/`clearCredentials()` no-arg before `configure()`**: restored base behavior exactly (empty session `{tokens, credentials, identityId, userSub: undefined}` / harmless resolve). The new explicit-ctx overloads (`fetchAuthSession(ctx)`) are purely additive; the typed guards still fire on misuse (unbranded/mis-ordered ctx args).

**Retained as bug fixes:**
1. **api-graphql server request-clients**: client-instance-level `authMode`/`endpoint`/`apiKey`/`authToken`/`headers` now take effect (previously silently dropped by the old wrapper's `this`-binding) — matching browser and cookie-SSR clients.
2. **api-graphql events error messages**: reworded to accurately describe the failure (plain `Error` retained; message text is not API).

Also: analytics per-import `import/no-deprecated` disables (F6.3). Storage F6.1 was found already resolved on the branch (landed with #14874).

## Validation
- `yarn build` 20/20 · `yarn test` all packages 0 failures (auth 1157, datastore 1160, storage 859, core 770, adapter-nextjs 303, notifications 276, api-rest 218, api-graphql 161, geo 124, analytics 123, api 101, aws-amplify 71, predictions 59, interactions 47, pubsub 25, + RN packages) · `yarn test:size` 27/27 under budget · lint clean on touched packages
- Pre-PR LLM review: one critical (native/web async mismatch) fixed in `bdab8e6dd`; review-round fixes (await pass, doc comments, plain-language comments) in `77ebbcbfd`

## Checklist
- [x] PR targets the `feat/bobbor/v6-context` feature branch (no changeset required)
- [x] Unit tests added/updated; no assertions weakened
- [x] Full monorepo build + test + size validation green


packages/api-graphql/src/server/generateClient.ts
```diff
@@ -34,34 +26,15 @@ export function generateClient<
...
-	const client = generateClientWithAmplifyInstance<T, V6ClientSSRRequest<T>>({
+	// Pattern 6 (post-Phase C3): `client.graphql` natively accepts the
+	// per-request branded `AmplifyContext` as its first argument (see
+	// `internals/v6.ts`), so the per-client `graphql` wrapper this entry used
+	// to carry is gone — this is a bare pass-through, mirroring the other
+	// categories' server entries.
+	// Passing `null` instance because each method must receive a valid context
+	// from the per-request argument.
+	return generateClientWithAmplifyInstance<T, V6ClientSSRRequest<T>>({
 		amplify: null,
```
osama-rizk, line 37: Behavior change worth confirming + noting in the description — this is not purely a "bare pass-through." The old per-client `graphql` wrapper called `prevGraphql.call({ [__amplify]: contextSpec }, ...)`, binding `this` to an object carrying **only** the context — so `getInternals(this)` saw `undefined` for client-instance-level `authMode` / `endpoint` / `apiKey` / `authToken` / `headers`, silently dropping them on SSR request clients.

Collapsing to `amplify: null` + passing the ctx explicitly keeps `this = client`, so those client-level options now actually take effect (matching the browser and cookie-SSR clients). That's a correctness improvement, but it's a real behavior change for anyone who set client-level options on a server request client and had them silently ignored — worth calling out in the PR description rather than filing it under test dedup.

packages/notifications/src/pushNotifications/providers/pinpoint/apis/initializePushNotifications.native.ts
```diff
@@ -63,13 +63,22 @@ const BACKGROUND_TASK_TIMEOUT = 25; // seconds
  * initializePushNotifications();
  * ```
  */
-export function initializePushNotifications(): void;
+export async function initializePushNotifications(): Promise<void>;
```
osama-rizk: The sync→async unification is correct (body is synchronous listener wiring, so wiring still completes before the resolved Promise). One exposure to be aware of: this native `initializePushNotifications` (and `removeDevice`) previously **threw synchronously** on a misconfigured app; now they **reject**. A fire-and-forget caller that neither awaits nor `.catch`es will get an unhandled promise rejection where it used to get a catchable synchronous throw. Consistent with the web stubs and documented in-code, so this is an accepted tradeoff — flagging for awareness.

osama-rizk (commented):
Reviewed with the src/test split — of 211 files only 18 carry behavior; the ~140 migrated test files got a faithfulness check rather than a line read. Findings: **no defects.**

Verified sound:
- Core helpers (`createCtxResolver`, `selectSsrKeyValueStorage`, `peekCtxArgs`) are faithful extractions; the `resolveLibraryOptions` reconfigure path and its regression tests are preserved unchanged.
- Test dedup weakens nothing: net assertions rose, the shared `createMockAmplifyContext` is a faithful superset of the 8 deleted copies, and the disclosed `withTokens` double-priming is inert — every migrated auth suite reads the session via `fetchAuthSession`, none via `getTokens`, so no suite silently switched branches.
- Typed-error migration (adapter-nextjs) preserves `instanceof AmplifyError` + the exact messages, diverging only on `.name` (the intent); zero references to the deprecated class remain. The `events/utils.ts` rewording is message-only and additionally hardens against `Symbol` interpolation.
- api-graphql null-instance server path is guarded (`InvalidAmplifyContextError` fires before the only `amplify` deref); no NPE, and SSR request clients expose no subscription path.

Two low-severity behavior changes inline — both intended/improvements, just worth confirming and noting in the description since the framing reads as test-only dedup.

packages/notifications/src/pushNotifications/providers/customer-profiles/apis/initializePushNotifications.native.ts
```diff
@@ -48,13 +48,22 @@ const BACKGROUND_TASK_TIMEOUT = 25; // seconds
...
-	// Validate that config is available (throws if not configured yet)
+export async function initializePushNotifications(
+	ctx: AmplifyContext,
+): Promise<void>;
+// The body is synchronous listener wiring; the function is `async` so the
+// return type matches the web stub overloads (`Promise<void>`, F6.2 unified
+// async-rejecting contract). This deliberately surfaces the not-configured
+// `resolveCtxArgs` throw as a rejection, consistent with the web stubs which
+// reject with `PlatformNotSupportedError`.
+export async function initializePushNotifications(
```
soberm: [major] Heads up: this flips `initializePushNotifications()` from a synchronous throw to an async rejection when called before `Amplify.configure()`. Anyone with existing `try { initializePushNotifications() } catch {}` code (pretty natural given the old signature was `void`) will now get an unhandled promise rejection instead. Same change in the pinpoint provider. Given this is a public API on a widely used SDK, might be worth a changelog/breaking-change note rather than just the code comment here, is that tracked elsewhere in the v6-context migration notes?

packages/notifications/src/pushNotifications/providers/pinpoint/apis/initializePushNotifications.native.ts
```diff
@@ -63,13 +63,22 @@ const BACKGROUND_TASK_TIMEOUT = 25; // seconds
...
-	// Validate that config is available (throws if not configured yet)
+export async function initializePushNotifications(
+	ctx: AmplifyContext,
+): Promise<void>;
+// The body is synchronous listener wiring; the function is `async` so the
+// return type matches the web stub overloads (`Promise<void>`, F6.2 unified
+// async-rejecting contract). This deliberately surfaces the not-configured
+// `resolveCtxArgs` throw as a rejection, consistent with the web stubs which
+// reject with `PlatformNotSupportedError`.
+export async function initializePushNotifications(
```
soberm: [major] Same async/reject-vs-throw contract change as the customer-profiles provider, see my comment over there. Flagging here too since it's an independent public export.

packages/notifications/__tests__/pushNotifications/providers/customer-profiles/apis/initializePushNotifications.native.test.ts
```diff
@@ -161,6 +161,21 @@ describe('initializePushNotifications (customer-profiles, native)', () => {
...
+		try {
+			await expect(initializePushNotifications()).rejects.toThrow();
+		} finally {
+			setGlobalContext(createMockAmplifyContext());
+		}
+	});
+
 	describe('background notification', () => {
 		it('registers a headless task if able', () => {
 			initializePushNotifications();
```
soberm, line 166: [minor] Now that this returns a real promise, most of the call sites in this file are unawaited floating promises — this one plus lines 160, 194, 204, 226, 241, 260, 271, 282, 317, 330, 345, 367, 380, 393, 403, 416, 434 (only the two tests right above this one actually await it). Harmless today since the body has no internal `await` so side effects still land synchronously, but it's fragile: any future `await` added inside `initializePushNotifications`, or any of these running in a state where it actually rejects, becomes a silent unhandled rejection instead of a test failure. Same pattern in the pinpoint sibling test (15 sites) and `initializePushNotifications.native.reconfigure.test.ts` (2 sites). Might be worth an `await` pass across all three files while this is fresh.

packages/core/src/testing/createMockAmplifyContext.ts
```diff
@@ -0,0 +1,144 @@
...
+		(configIsOptions
+			? (resourcesConfig as CreateMockAmplifyContextOptions)
+			: options) ?? {};
+	const staticConfig = configIsOptions
+		? undefined
+		: (resourcesConfig as ResourcesConfig | undefined);
+
+	const ctx: MockAmplifyContext = {
+		// Live getter so `getConfig`-driven per-test config changes propagate.
+		get resourcesConfig(): ResourcesConfig {
```
soberm, line 88: [minor] `resourcesConfig` used to be a plain writable property in basically all the per-package mocks this replaces (checked auth's and analytics' old copies, both had it as a plain data field). Here it's getter-only with no setter. If any migrated test does `ctx.resourcesConfig = ...` directly instead of going through `getConfig`, it'll throw a TypeError now. Doesn't look like anything hits this today given the green tests, but probably worth a line in the doc comment above so the next person doesn't assume it's assignable like the old copies were.

packages/core/src/testing/createMockAmplifyContext.ts
```diff
@@ -0,0 +1,144 @@
...
+	ctx: MockAmplifyContext,
+	accessToken: string | JWT,
+): MockAmplifyContext {
+	const jwt: JWT =
+		typeof accessToken === 'string'
+			? { payload: {}, toString: () => accessToken }
+			: accessToken;
+	const tokens: AuthTokens = { accessToken: jwt };
+
+	ctx.getTokens.mockResolvedValue(tokens);
```
soberm, line 133: [minor] Since this is called out in the PR description as a known risk: might be worth actually auditing the ~31 `withTokens` call sites for any that previously mocked only `fetchAuthSession` and independently asserted `getTokens` as unset, rather than leaning on tests being green as the only signal. Priming both here could be masking a spot where a code path should look unauthenticated but now silently sees valid tokens.

packages/adapter-nextjs/src/auth/createAuthRouteHandlersFactory.ts
```diff
@@ -97,15 +97,17 @@ export const createAuthRouteHandlersFactory = ({
 		// origin validation should happen when createAuthRouteHandlers is being called to create
 		// Auth API routes.
 		if (!amplifyAppOrigin) {
-			throw new AmplifyServerContextError({
+			throw new AmplifyError({
```
soberm: [minor] This (and the two similar swaps in `resolveRedirectUrl.ts` and `createCookieStorageAdapterFromNextServerContext.ts`) replaces `AmplifyServerContextError` with newly-named `AmplifyError` variants. Messages are preserved and the tests are updated so this reads as intentional, but it's breaking for anyone doing `instanceof AmplifyServerContextError` or matching on `error.name === 'AmplifyServerContextError'`. Worth a changelog line alongside the other disclosed changes in the description?

packages/api-graphql/src/internals/v6.ts
```diff
@@ -104,10 +106,65 @@ export function graphql<
...
+	this: V6Client,
+	ctxOrOptions:
+		| AmplifyContext
+		| GraphQLOptionsV6<FALLBACK_TYPES, TYPED_GQL_STRING, Options>,
+	optionsOrHeaders?:
+		| GraphQLOptionsV6<FALLBACK_TYPES, TYPED_GQL_STRING, Options>
+		| CustomHeaders,
+	maybeHeaders?: CustomHeaders,
 ): GraphQLResponseV6<FALLBACK_TYPES, TYPED_GQL_STRING> {
+	// Pattern 6 (post-Phase C3): server req/res clients pass the per-request
```
soberm: [nit] "Pattern 6 (post-Phase C3)" (and "F6.2" over in the notifications files) are internal migration-phase labels that won't mean much to anyone outside this refactor reading the history later. Might read better as a plain description of why the ctx-first overload exists.

packages/api-graphql/src/internals/events/utils.ts
```diff
@@ -30,7 +30,7 @@ export const configure = (ctx: AmplifyContext) => {
 
 	if (!eventsConfig) {
 		throw new Error(
-			'Amplify configuration is missing. Have you called Amplify.configure()?',
+			'Events API configuration is missing. Have you called Amplify.configure() with an `API.Events` configuration?',
```
soberm, line 33: [nit] Nicer message, but since these stay plain `Error` rather than a typed `AmplifyError` (here and the two similar ones below), anyone matching the old exact string breaks with no type signal to catch it. Probably fine, just flagging since it's a public-facing message.

packages/notifications/src/pushNotifications/providers/customer-profiles/apis/removeDevice.ts
```diff
@@ -22,6 +22,6 @@ export function removeDevice(): Promise<void>;
  * @param ctx - The {@link AmplifyContext} to use for config and credentials.
  */
 export function removeDevice(ctx: AmplifyContext): Promise<void>;
-export function removeDevice(..._args: any[]): Promise<void> {
+export async function removeDevice(..._args: any[]): Promise<void> {
```
soberm, line 28: [minor] This one's already fixed correctly (`async` added), it was declared `Promise<void>` but threw synchronously, so a `.catch()`-chaining caller would've gotten an uncaught exception instead of a rejection. Since the native `initializePushNotifications` files in this same PR reintroduce the identical class of mistake at call sites that don't await it (see my comment on the test file), might be worth `@typescript-eslint/no-floating-promises` or `promise-function-async` repo-wide instead of catching this by hand each time.

packages/api-graphql/src/server/generateClient.ts
bobbor: Confirmed — your mechanism analysis is exactly right: the old wrapper's `this`-binding carried only the context, so client-level options were silently dropped on SSR request clients, and the collapse makes them take effect (matching browser/cookie-SSR clients). Promoted to the PR description as disclosed behavior change #1 rather than test dedup.

packages/notifications/src/pushNotifications/providers/pinpoint/apis/initializePushNotifications.native.ts
bobbor: Acknowledged — the throw→reject exposure for fire-and-forget callers is the accepted trade-off of the unified contract. It's now disclosed in the PR description (change #2) and tracked in the v6-context migration notes (P5).

packages/notifications/src/pushNotifications/providers/customer-profiles/apis/initializePushNotifications.native.ts
bobbor: Yes — now tracked: the sync-throw→async-rejection contract change is disclosed in the PR description (change #2) and recorded in the migration-notes task (P5) alongside the other public-contract changes from the series (InvalidCookiesError, the deprecated runWithAmplifyServerContext shim semantics, and the typed adapter-nextjs errors from this PR). Correction from earlier wording: there is no v7 — where/how these notes ship depends on the release posture of the feature branch, which is an open question we're resolving (this change may yet be made backward-compatible instead).

packages/notifications/src/pushNotifications/providers/pinpoint/apis/initializePushNotifications.native.ts
bobbor: Same treatment as the customer-profiles thread — disclosed in the description and tracked in the migration notes (P5).

packages/notifications/__tests__/pushNotifications/providers/customer-profiles/apis/initializePushNotifications.native.test.ts
bobbor: Done in 77ebbcbfd — full await pass across both native suites, both reconfigure test files, and the customer-profiles authStateTransitions test (~37 sites total); the six done-callback tests were converted to async with a deferred promise (Jest forbids async+done). Assertions unchanged.

packages/core/src/testing/createMockAmplifyContext.ts
bobbor: Doc comment added in 77ebbcbfd: resourcesConfig is documented as getter-only/read-only, pointing tests at the getConfig option instead of assignment.

packages/core/src/testing/createMockAmplifyContext.ts
bobbor: Audited properly in the 77ebbcbfd round rather than leaning on green suites: all 31 withTokens sites across 17 auth test files, each compared against its pre-migration inline mock. Result: ctx.getTokens is read by exactly one internal code path (getCurrentUser), which none of the 17 audited APIs' implementations reach; zero test files reference getTokens; and both old and new mocks default getTokens to resolve undefined, so the priming is inert. Every test still proves what it proved before — no site needed changing.

packages/adapter-nextjs/src/auth/createAuthRouteHandlersFactory.ts
bobbor: Agreed it's breaking for instanceof/name matchers — now disclosed in the PR description (change #3) and tracked in the migration notes (P5) together with Phase C's InvalidCookiesError. The old class was already @deprecated since Phase C, so this PR removes its last consumers.

packages/api-graphql/src/internals/v6.ts
bobbor: Fixed in 77ebbcbfd — replaced all seven phase/ledger labels in src comments with plain-language rationale (this one now explains the ctx-first overload exists so server callers can pass a per-request context for request-scoped signing); test files keep their references.

packages/api-graphql/src/internals/events/utils.ts
bobbor: Acknowledged — kept as plain Error deliberately (typing the events errors is a bigger surface decision than this cleanup PR should make), and the exact-string break is now disclosed in the description (change #4). If we want typed events errors, that's a good candidate for the events API's own hardening pass.

packages/notifications/src/pushNotifications/providers/customer-profiles/apis/removeDevice.ts
bobbor: Good idea — agreed that catching this class of bug by hand doesn't scale. Adding @typescript-eslint/no-floating-promises (or promise-function-async) repo-wide touches every package's lint config and existing violations, so proposing it as a standalone repo-hygiene follow-up (DXE bucket) rather than folding it in here. The await pass in 77ebbcbfd covers the sites this PR introduced.

osama-rizk (approved):

bobbor:
Heads-up for re-review: pushed `51917e9b6` adding the `fetchAuthSession(ctx, options?)` and `clearCredentials(ctx)` overloads — this restores the pre-context SSR calling pattern (`operation: ctx => fetchAuthSession(ctx)`) that the Phase C removal of the spec-taking server variant left without a direct-context equivalent. Covered by ctx-isolation, global-fallback, typed-guard, and an end-to-end SSR-shaped test via `aws-amplify/auth/server`; legacy `fetchAuthSession(contextSpec)` compiles unchanged (type-level test). One disclosed behavior note added to the description (#5): pre-configure no-arg calls now throw the typed `NoAmplifyContextError` instead of returning an empty session.

packages/notifications/src/pushNotifications/providers/customer-profiles/apis/initializePushNotifications.native.ts
bobbor: Update — maintainer decision: this branch must not carry breaking public contracts (there is no committed major release). The stub unification is REVERTED in `e9b7a089b`: `initializePushNotifications` is byte-identical to base (sync `(): void`, native sync). Only `removeDevice` keeps the async form, since its *declared* base type was already `Promise<void>` and the sync throw was the base bug you called out on that thread. Your `[major]` flag was the right call — thanks for insisting.

packages/notifications/src/pushNotifications/providers/pinpoint/apis/initializePushNotifications.native.ts
bobbor: Update: reverted in `e9b7a089b` per maintainer decision — no breaking contracts on this branch. The throw→reject exposure you flagged no longer exists; the stubs match base exactly.

packages/adapter-nextjs/src/auth/createAuthRouteHandlersFactory.ts
bobbor: Update: reverted in `e9b7a089b` per maintainer decision — the three sites throw `AmplifyServerContextError` exactly as base, so `instanceof`/name matchers are unaffected. The typed-error migration is deferred until there's an actual major-release vehicle to carry it.

soberm (approved):

osama-rizk (approved):

